PR #4191, opened 2022-02-11
[MAINTENANCE] Refactor RuleBasedProfiler toolkit pattern

cdkini:
- Move toolkit methods to RBP class as static methods

### Definition of Done


great_expectations/data_context/data_context.py
```diff
@@ -941,6 +936,17 @@ def profiler_store(self) -> ProfilerStore:
...
+    @staticmethod
+    def default_profilers_exist(directory_path: Optional[str]) -> bool:
+        if not directory_path:
+            return False
+
+        profiler_directory_path: str = os.path.join(
+            directory_path,
+            DataContextConfigDefaults.DEFAULT_PROFILER_STORE_BASE_DIRECTORY_RELATIVE_NAME.value,
+        )
+        return os.path.isdir(profiler_directory_path)
```
cdkini: This method was in the RBP toolkit but makes more sense in DataContext since it is directly asking the question of whether or not the context's profiler directory is present.

donaldheppner: I'd be tempted to put this on the RulesBasedProfiler as a static method along with the constant; there's no reason the component cannot own this IMHO

tests/rule_based_profiler/test_rule_based_profiler.py
```diff
@@ -430,3 +443,261 @@ def test_reconcile_profiler_rules_existing_rule_full_rule_override(
...
+):
+    with pytest.raises(AssertionError) as e:
+        RuleBasedProfiler.get_profiler(
+            data_context=mock_data_context,
+            profiler_store=populated_profiler_store,
+            name="my_profiler",
+            ge_cloud_id="my_ge_cloud_id",
+        )
+
+    assert "either name or ge_cloud_id" in str(e.value)
```
donaldheppner, line 514: I like this

great_expectations/data_context/data_context.py
donaldheppner: Can this be made "private" with an _? I consider anything without an underscore to be part of the public API, and I want to keep that surface as small as possible. Unless we expect GE users to use it, it should be private

donaldheppner (changes requested):

great_expectations/data_context/data_context.py
```diff
@@ -941,6 +936,17 @@ def profiler_store(self) -> ProfilerStore:
                 f'Attempted to access the Profiler store named "{profiler_store_name}", which is not a configured store.'
             )
 
+    @staticmethod
+    def default_profilers_exist(directory_path: Optional[str]) -> bool:
```
anthonyburdi:
```suggestion
    def _default_profilers_exist(directory_path: Optional[str]) -> bool:
```

alexsherstinsky (changes requested):
Looks great provided the recommended changes by @donaldheppner @anthonyburdi are incorporated.  Thanks for doing this!

alexsherstinsky (approved):
LGTM

donaldheppner (approved):

PR #4194, opened 2022-02-14
[MAINTENANCE] Update `suite new` workflow to better integrate with RuleBasedProfiler 

cdkini:
- Expose `Validator._expectation_suite` through setter to ensure that RBP results are recognized in `suite new` workflow
- Reduce log level of `setattr` to reduce the visual clutter within the workflow

### Definition of Done


great_expectations/render/renderer/v3/suite_profile_notebook_renderer.py
```diff
@@ -213,6 +213,8 @@ def _add_rule_based_profiler_cells(self) -> None:
     name="{self._profiler_name}",
     expectation_suite_name=expectation_suite_name,
 )
+
+validator.expectation_suite = suite
```
cdkini, line 217: Per discussion with @alexsherstinsky, we need a way to associate our RBP result suite with our validator in the CLI workflow.

While we might want to refine this later on, the agreed-upon approach is to expose the suite attr through a setter so we can manually make that connection.

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -99,7 +99,7 @@ def __init__(
         self.batch_request = batch_request
         for k, v in kwargs.items():
             setattr(self, k, v)
-            logger.warning(
+            logger.debug(
                 "Setting unknown kwarg (%s, %s) provided to config as attr", k, v
             )
```
cdkini, line 104: Between numerous logging statements and metric calc progress bars, the notebook is super difficult to decipher. Let's keep things simple.

PR #4216, opened 2022-02-16
[BUGFIX] Update validate_configuration for core Expectations that don't return…

kenwade4:
The `validate_configuration` methods defined on several Expectations do not `return True` at the end, as specified in their docstrings. Exceptions should be raised if there is anything wrong with provided configs (before `return True`).

### Definition of Done

### DIFF
--- a/great_expectations/expectations/core/expect_column_max_to_be_between.py
+++ b/great_expectations/expectations/core/expect_column_max_to_be_between.py
@@ -115,7 +115,9 @@ class ExpectColumnMaxToBeBetween(ColumnExpectation):
 
     """ A Column Map MetricProvider Decorator for the Maximum"""
 
-    def validate_configuration(self, configuration: Optional[ExpectationConfiguration]):
+    def validate_configuration(
+        self, configuration: Optional[ExpectationConfiguration]
+    ) -> bool:
         """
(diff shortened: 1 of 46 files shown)


alexsherstinsky (approved):
LGTM

great_expectations/expectations/core/expect_column_max_to_be_between.py
```diff
@@ -129,6 +129,8 @@ def validate_configuration(self, configuration: Optional[ExpectationConfiguratio
         super().validate_configuration(configuration)
         self.validate_metric_value_between_configuration(configuration=configuration)
 
+        return True
+
     @classmethod
```
cdkini, line 136: Would you mind type annotating the return value of all these Expectations? Thanks so much for fixing this 🙇🏽

kenwade4: For all Expectations or just the ones that were modified in this branch?

cdkini: Just the ones modified in this branch! I think we can incrementally refactor and improve our code but I don't want to increase the scope of your work here.

As we touch the other files, we can update those accordingly.

kenwade4: Too late, haha.

PR #4226, opened 2022-02-17
[MAINTENANCE] Type hint usage stats

cdkini:
- Type hints!

### Definition of Done

### DIFF
--- a/great_expectations/core/usage_statistics/usage_statistics.py
+++ b/great_expectations/core/usage_statistics/usage_statistics.py
@@ -11,7 +11,8 @@ import threading
 import time
 from functools import wraps
 from queue import Queue
-from typing import Optional
+from types import FrameType
+from typing import Callable, Optional
 
 import jsonschema
 import requests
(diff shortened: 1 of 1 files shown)


anthonyburdi (approved):
LGTM! Thank you for adding these type hints!

great_expectations/core/usage_statistics/usage_statistics.py
```diff
@@ -343,11 +352,11 @@ def usage_statistics_wrapped_method_partial(func):
 
 # noinspection PyUnusedLocal
 def run_validation_operator_usage_statistics(
-    data_context,
-    validation_operator_name,
-    assets_to_validate,
+    data_context: "DataContext",  # noqa: F821
```
NathanFarmer, line 355: Why wouldn't you import `DataContext` here and use the object? Just to avoid the import? Is that why we see "noqa"? Should we be doing this everywhere that an import isn't really required?

NathanFarmer (approved):
LGTM! One question though...

great_expectations/core/usage_statistics/usage_statistics.py
cdkini: X-posting for visibility:

The quotes are necessary to deal with cyclic imports: https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports

The # noqa: F821 is a linter directive to ignore flake8 checks for PEP compliance. A little unfortunate but as long as we have cycles in our dependency graph, we require quoted annotations.

I think 3.7+ has a better way to do this but but we’re bound by our lowest supported version. Alex Sherstinsky might be able to provide a bit more clarity here!

We do not actually import DataContext here because the context references this file (creating a cycle that the interpreter cannot resolve). To still get that relevant annotation, we settle for this string version.

PR #4247, opened 2022-02-22
[MAINTENANCE] Update contrib expectations to use print_diagnostic_checklist method

kenwade4:
The object returned when calling the Expectation's run_diagnostics method is not JSON serializable so `print(json.dumps(diagnostics_report, indent=2)) ` would error out.

Updating to `print(json.dumps(diagnostics_report.to_json_dict(), indent=2))` works, but produces a LOT of unhelpful noise (since the report is much more robust now).

- Update contrib expectations to use `print_diagnostic_checklist` method
- Update last template that was not previously updated

### Definition of Done


austiezr (approved):
LGTM!

PR #4251, opened 2022-02-22
[MAINTENANCE] Resolve cyclic import issue with usage stats

cdkini:
- Import anonymizers at runtime when needed ("lazy load" of sorts) instead of at the top of the file - as these classes interact with many parts of the GE codebase, their inclusion as global imports causes a multitude of import issues.
- As a way to circumvent this issue, I've attempted to make the imports work in the constructor. This shouldn't impact performance but let me know if you think otherwise!

### Definition of Done

Thank you for submitting!


great_expectations/core/usage_statistics/usage_statistics.py
```diff
@@ -84,6 +57,22 @@ def __init__(
...
+            BatchAnonymizer,
+            BatchRequestAnonymizer,
+            CheckpointRunAnonymizer,
+            DataDocsSiteAnonymizer,
+            DatasourceAnonymizer,
+            ExecutionEngineAnonymizer,
+            ExpectationSuiteAnonymizer,
+            StoreAnonymizer,
+            ValidationOperatorAnonymizer,
+        )
```
cdkini, line 74: This resolves any import issues but let me know if this doesn't feel right.

Additionally, let me know if you think any of these are viable additions:
* Singleton pattern through a class-level attr so we only ever need to walk through this code once
* Usage of `importlib` instead of the keyword `import`

setup.cfg
```diff
@@ -14,3 +14,4 @@ parentdir_prefix = great_expectations-
 [flake8]
 max-line-length = 200
 exclude = tests/*,docs/*
+per-file-ignores = */__init__.py: F401 # Allow unused imports in __init__'s
```
cdkini: F401 is the warning for unused imports - I've decided to silence this globally since we never actually use the imported class/function in the init.

donaldheppner (approved):
I think this is fine for now; our usage statistics need to be revisited because of a dependency on DataContext anyway

Issue #123, opened 2017-11-17
Use cleaner solution for non-truncating division in python 2


jpcampb2:
Prefer `from __future__ import division` to `1.*x/y`

jpcampb2:
Done inline during improve_partitions push.

PR #4258, opened 2022-02-22
[FEATURE] Self-Initializing Expectations (implemented for three example expectations).

alexsherstinsky:
### Scope
The feature "Self-Initializing Expectations" is an application of the Rule-Based Profiler component.

* Three example Expectations,  `expect_column_values_to_be_between`, `expect_table_row_count_to_be_between`, and `expect_column_quantile_values_to_be_between expectations` can now estimate their own parameters (`kwargs`).
* Exhaustive integration tests are provide for all three of the above expectations, exploring such scenarios as:
  - Only default profiler is configured;
  - Only custom profiler is specified;
  - Both default and custom profiler configurations are specified;
  - Any explicit user-provided overrides are specified.

With the use of the `auto` and `profiler` config, one can now explore data without knowing expectation values a-priori.

For instance, using `expect_column_values_to_be_between` and the "NYC Taxi" dataset for an illustration, one can utilize the default Rule-Based Profiler, configured for that expectation, as follows:
```
    result = validator.expect_column_values_to_be_between(
        column="fare_amount",
        result_format="SUMMARY",
        include_config=True,
        auto=True,
    )
```
 
Then `result.expectation_config["kwargs"]` will contain the entries
```
        "min_value": -52.0,
        "max_value": 3004.0,
```
(which correspond to the `batch_request` used to obtain the `validator: Validator` object above).

In addition, one can define a custom Rule-Based Profiler configuration for this expectation:
```
    custom_profiler_config = {
        "name": "expect_column_values_to_be_between",  # Convention: use "expectation_type" as profiler name.
        "config_version": 1.0,
        "rules": {
            "custom_column_values_between_rule": {
                "parameter_builders": [
                    {
                        "name": "my_min_estimator",
                        "class_name": "MetricMultiBatchParameterBuilder",
                        "module_name": "great_expectations.rule_based_profiler.parameter_builder",
                        "metric_name": "column.min",
                        "metric_domain_kwargs": "$domain.domain_kwargs",
                        "enforce_numeric_metric": True,
                        "replace_nan_with_zero": True,
                    },
                ],
                "expectation_configuration_builders": [
                    {
                        "class_name": "DefaultExpectationConfigurationBuilder",
                        "module_name": "great_expectations.rule_based_profiler.expectation_configuration_builder",
                        "expectation_type": "expect_column_values_to_be_between",
                        "column": "$domain.domain_kwargs.column",
                        "min_value": "$parameter.my_min_estimator.value[0]",
                        "mostly": "$variables.mostly",
                        "strict_min": "$variables.strict_min",
                        "meta": {
                            "details": {
                                "my_min_estimator": "$parameter.my_min_estimator.details",
                            },
                        },
                    },
                ],
            },
        },
    }
```

The custom Rule-Based Profiler configuration, merged with the default Rule-Based Profiler configuration, cause the `min_value` to be estimated, while leaving the `max_value` unconstrained (as per the default behavior of the expectation):
```
        column="fare_amount",
        result_format="SUMMARY",
        include_config=True,
        auto=True,
        profiler_config=custom_profiler_config,
    )
```

Then `result.expectation_config["kwargs"]` will contain the entry
```
        "min_value": -52.0,
```
(which corresponds to the `batch_request` used to obtain the `validator: Validator` object above) as well as the custom Rule-Based Profiler configuration, while "max_value" will be absent.

Moreover, the user can supply any Expectation `kwarg` key-value pairs explicitly, thereby overriding the Rule-Based Profiler (both default and custom configurations, whichever are present):

```
    result = validator.expect_column_values_to_be_between(
        column="fare_amount",
        min_value=0.0,
        mostly=8.75e-1,
        result_format="SUMMARY",
        include_config=True,
        auto=True,
        profiler_config=custom_profiler_config,
    )
```

Then `result.expectation_config["kwargs"]` will contain the entries
```
        "min_value": 0.0,
        "mostly": 8.75e-1,
```
(which correspond to the `batch_request` used to obtain the `validator: Validator` object above) as well as the custom Rule-Based Profiler configuration.

These scenarios are now applicable to the other two expectations, mentioned above.

- JIRA: GREAT-497
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/checkpoint/util.py
```diff
@@ -266,25 +267,18 @@ def get_substituted_batch_request(
         batch_request=substituted_runtime_batch_request
     )
 
-    effective_batch_request: dict = dict(
```
alexsherstinsky: Note: These changes have been taken care of by PR #4259

great_expectations/core/batch.py
```diff
@@ -605,13 +605,40 @@ def head(self, n_rows=5, fetch_all=False):
         return self._data.execution_engine.resolve_metrics((metric,))[metric.id]
 
 
+def materialize_batch_request(
```
alexsherstinsky: Note: These changes have been taken care of by PR #4259

great_expectations/rule_based_profiler/util.py
```diff
@@ -113,21 +118,23 @@ def build_batch_request(
     domain: Optional[Domain] = None,
     variables: Optional[ParameterContainer] = None,
     parameters: Optional[Dict[str, ParameterContainer]] = None,
-) -> Optional[BatchRequest]:
+) -> Optional[Union[BatchRequest, RuntimeBatchRequest]]:
```
alexsherstinsky: Note: These changes have been taken care of by PR #4259

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -99,20 +96,26 @@ def to_json_dict(self) -> dict:
         return convert_to_json_serializable(data=self.to_dict())
 
 
-DEFAULT_RECONCILATION_DIRECTIVES: ReconciliationDirectives = ReconciliationDirectives(
```
alexsherstinsky: Note: These changes have been taken care of by PR #4259

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -381,7 +381,7 @@ def reconcile_profiler_variables(
         :return: reconciled variables in their canonical ParameterContainer object form
         """
         effective_variables: ParameterContainer
-        if variables is not None and isinstance(variables, dict):
+        if variables and isinstance(variables, dict):
```
alexsherstinsky: Note: These changes have been taken care of by PR #4261

tests/test_fixtures/rule_based_profiler/alice_user_workflow_verbose_profiler_config.yml
```diff
@@ -36,7 +36,7 @@ rules:
       - expectation_type: expect_column_values_to_be_between
```
alexsherstinsky: Note: These changes have been taken care of by PR #4261

great_expectations/validator/validator.py
```diff
@@ -766,6 +1073,7 @@ def resolve_validation_graph(
         graph: ValidationGraph,
         metrics: Dict[Tuple[str, str, str], Any],
         runtime_configuration: Optional[dict] = None,
+        min_graph_edges_pbar_enable: int = 0,  # Set to low number (e.g., 3) to suppress progress bar for small graphs.
```
alexsherstinsky: Note: These changes have been taken care of by PR #4261

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
```diff
@@ -159,6 +221,14 @@ def validate_configuration(
             assert isinstance(
                 configuration.kwargs["quantile_ranges"], dict
             ), "quantile_ranges should be a dictionary"
+
+            if isinstance(
+                configuration.kwargs["quantile_ranges"]["value_ranges"], np.ndarray
+            ):
+                configuration.kwargs["quantile_ranges"][
+                    "value_ranges"
+                ] = configuration.kwargs["quantile_ranges"]["value_ranges"].tolist()
```
jcampbell: This surprises me a bit to be here instead of managed by the configuration itself during serialization. Seems like the longer we can keep things in numpy the better.

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
```diff
@@ -131,16 +131,78 @@ class ExpectColumnQuantileValuesToBeBetween(ColumnExpectation):
...
+                            },
+                            "allow_relative_error": "$variables.allow_relative_error",
+                            "meta": {
+                                "profiler_details": "$parameter.quantile_value_ranges.details"
+                            },
+                        }
+                    ],
+                }
+            },
+        },
```
jcampbell: Any reason this is a dict not a "ProfilerConfiguration"?

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
```diff
@@ -131,16 +131,78 @@ class ExpectColumnQuantileValuesToBeBetween(ColumnExpectation):
...
+        "auto",
+        "profiler_config",
     )
     default_kwarg_values = {
         "row_condition": None,
-        "allow_relative_error": None,
         "condition_parser": None,
         "quantile_ranges": None,
         "result_format": "BASIC",
         "allow_relative_error": False,
```
jcampbell, line 205: What's the significance of None->False?

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
```diff
@@ -131,16 +131,78 @@ class ExpectColumnQuantileValuesToBeBetween(ColumnExpectation):
...
+                "truncate_values": {
+                    "lower_bound": None,
+                    "upper_bound": None,
+                },
+                "round_decimals": 1,
+            },
+            "rules": {
+                "column_quantiles_rule": {
+                    "domain_builder": {
+                        "class_name": "ColumnDomainBuilder",
```
jcampbell: Does this domain builder see a pared-down version of the batch with only the column provided?

jcampbell (commented):
So excited!!

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @jcampbell Unfortunately, the `all()` operator insists on lists and does not work on `numpy.ndarray` -- the conversion to list was in response to this error showing up when `numpy.ndarray` valued results were encountered by this configuration change.

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @jcampbell In an earlier conversation with @anthonyburdi we agreed that it would be better to defer the instantiation until needed (because the `auto` flag can be `False`, meaning that the user does not want the automatic expectation kwargs estimation feature to be turned on).  Either will work -- the code would have to convert at some point in order to reconcile custom and default configurations.

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @jcampbell A column-style domain builder gets the active batch.

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @jcampbell There seemed to have been an inconsistency.  First, the docstring of `ExpectColumnQuantileValuesToBeBetween` says:
```
               allow_relative_error (boolean): \
                   Whether to allow relative error in quantile communications on backends that support or require it.
``` 

Hence, one is led to believe that `allow_relative_error` must be of type `bool`.  However, there was no `Optional` type hint and it was set to `None`.  Furthermore, there is still the following section in the code:
```
        if "allow_relative_error" in configuration.kwargs:
            allow_relative_error = configuration.kwargs["allow_relative_error"]
        else:
            allow_relative_error = False
```
However, this variable, `allow_relative_error`, is set, but not used; `PyCharm` flags this as a warning (we have not taken care of it as of yet).

Any guidance for dealing with the above you can offer will be appreciated.  Thanks!

jcampbell: Hmmm, I *think* that the non-bool values were used by redshift, and it looks to me like below you've got it set to "linear"?

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
jcampbell: So, how does this rule know to only operate on the user's selected column?

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
jcampbell: Got it. Totally a nitpick, but you could instead just check the resulting list for the check instead of overwriting the actual parameter.

great_expectations/expectations/core/expect_column_values_to_be_between.py
```diff
@@ -107,6 +109,60 @@ class ExpectColumnValuesToBeBetween(ColumnMapExpectation):
         "include_config": True,
         "catch_exceptions": False,
         "meta": None,
+        "auto": False,
+        "profiler_config": {
```
jcampbell: Why are these dictionaries instead of typed profiler configs?

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @jcampbell The column name is obtained from domain keys of the Expectation at hand and set (using a proper setter method) on the domain builder.  The code for that appears in `Validator` starting line 633.  Thanks.

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @jcampbell Yes, I use "linear".  So should we change it back to `None` and not expect it to be strictly boolean?  Perhaps `Union[bool, str]`?  Also, do you mind if I delete the code regarding the unused variable?  Thanks.

great_expectations/expectations/core/expect_column_values_to_be_between.py
alexsherstinsky: @jcampbell The reasoning is the same as in the case of the other two expectations, now imbued with the default profiler.  Specifically, in an earlier conversation with @anthonyburdi we agreed that it would be better to defer the instantiation until needed (because the auto flag can be False, meaning that the user does not want the automatic expectation kwargs estimation feature to be turned on). Either will work -- the code would have to convert at some point in order to reconcile custom and default configurations.

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @jcampbell Nice catch -- fixed (and committed).  Thanks!

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @jcampbell I made the change I proposed.  Thanks.

alexsherstinsky: @jcampbell Actually, this change causes downstream errors (Spark expects boolean).  I will change the type back.  Thanks.

Shinnnyshinshin (commented):
Thank you very much for this @alexsherstinsky. Going forward, are we planning on building a `profiler_config` for each Expectation to enable default self-initialization? Or is a subset that we would be adding this to?   I ask because the use-case and benefit of these configs is clear (in the scope of this PR) but wanting to get a better picture of our expected endpoint

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
```diff
@@ -131,16 +131,78 @@ class ExpectColumnQuantileValuesToBeBetween(ColumnExpectation):
...
         "quantile_ranges": None,
         "result_format": "BASIC",
-        "allow_relative_error": False,
+        "allow_relative_error": None,
         "include_config": True,
         "catch_exceptions": False,
+        "meta": None,
+        "auto": False,
+        "profiler_config": {
+            "name": "expect_column_quantile_values_to_be_between",  # Convention: use "expectation_type" as profiler name.
```
cdkini: Should this convention be documented elsewhere as well? Should it be enforced in any way?

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
```diff
@@ -159,9 +221,12 @@ def validate_configuration(
...
+
             assert all(
                 [
-                    True if None in x or x == sorted(x) else False
+                    True
+                    if isinstance(x, np.ndarray) or None in x or x == sorted(x)
+                    else False
                     for x in configuration.kwargs["quantile_ranges"]["value_ranges"]
                 ]
             ), "quantile_ranges must consist of ordered pairs"
```
cdkini: We could make this a generator - probably not a noticeable performance upgrade but no need to make an additional list if we can avoid it.

great_expectations/validator/validator.py
```diff
@@ -289,8 +301,11 @@ def inst_expectation(*args, **kwargs):
                         f"Invalid positional argument: {arg}"
                     )
 
-            configuration = ExpectationConfiguration(
-                expectation_type=name, kwargs=expectation_kwargs, meta=meta
+            configuration = self._build_expectation_configuration(
```
cdkini: Could we add back the type hint?

great_expectations/validator/validator.py
```diff
@@ -360,6 +375,298 @@ def inst_expectation(*args, **kwargs):
...
+        expectation_type: str,
+        expectation_kwargs: dict,
+        meta: dict,
+        expectation_impl: "Expectation",  # noqa: F821
+    ) -> ExpectationConfiguration:
+        success_keys: Tuple[str] = (
+            expectation_impl.success_keys
+            if hasattr(expectation_impl, "success_keys")
+            else None
+        ) or tuple()
```
cdkini:
```suggestion
        success_keys: Tuple[str] = (
            expectation_impl.success_keys
            if hasattr(expectation_impl, "success_keys")
            else tuple()
        )
```
Is this a viable change? If so, please implement throughout.

great_expectations/validator/validator.py
```diff
@@ -360,6 +375,298 @@ def inst_expectation(*args, **kwargs):
...
+                Dict[
+                    str,
+                    Union[
+                        str,
+                        float,
+                        Optional[Dict[str, Any]],
+                        Optional[Dict[str, Dict[str, Any]]],
+                    ],
+                ],
+            ]
```
cdkini: Do we have a cleaner approach to annotate the dict representation of configs? Can't immediately think of one but I'd really like to avoid super complex types like this if possible (I realize this isn't entirely feasible here so do you what you think is best).

great_expectations/validator/validator.py
```diff
@@ -360,6 +375,298 @@ def inst_expectation(*args, **kwargs):
...
+            """
+            profiler_config = default_profiler_config or profiler_config
+
+            # Roundtrip through schema validation to remove any illegal fields add/or restore any missing fields.
+            validated_profiler_config: dict = ruleBasedProfilerConfigSchema.load(
+                profiler_config
+            )
+            profiler_config = ruleBasedProfilerConfigSchema.dump(
+                validated_profiler_config
+            )
```
cdkini: Do we have a ticket to abstract that into a schema-level method? I think it could make for a nice refactor.

great_expectations/validator/validator.py
```diff
@@ -360,6 +375,298 @@ def inst_expectation(*args, **kwargs):
...
+                    str,
+                    Union[
+                        str,
+                        float,
+                        Optional[Dict[str, Any]],
+                        Optional[Dict[str, Dict[str, Any]]],
+                    ],
+                ],
+            ]
+        ] = None,
```
cdkini: Actually, given that this is so prevalent, I'd love to discuss a way to possibly circumvent this before we merge.

great_expectations/validator/validator.py
```diff
@@ -360,6 +375,298 @@ def inst_expectation(*args, **kwargs):
...
+            column_name = expectation_kwargs["column"]
+            rule.domain_builder.column_names = [column_name]
+            if ignore_batch_request or rule.domain_builder.batch_request is None:
+                # A DomainBuilder that emits MetricDomainTypes.COLUMN type Domain object needs exactly 1 Batch of data.
+                rule.domain_builder.batch = self.active_batch
+        elif domain_type == MetricDomainTypes.TABLE:
+            pass  # No action is needed.
+        else:
+            pass  # No action is needed.
+
```
cdkini, line 597: Any logging?

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
```diff
@@ -232,6 +236,705 @@ def test_bobby_profiler_user_workflow_multi_batch_row_count_range_rule_and_colum
...
+    try:
+        # noinspection PyUnusedLocal
+        suite = context.get_expectation_suite(
+            expectation_suite_name=expectation_suite_name
+        )
+    except ge_exceptions.DataContextError:
+        suite = context.create_expectation_suite(
+            expectation_suite_name=expectation_suite_name
+        )
+        print(f'Created ExpectationSuite "{suite.expectation_suite_name}".')
```
cdkini, line 259: Why are we doing a try/except here? We should test both paths right?

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
```diff
@@ -232,6 +236,705 @@ def test_bobby_profiler_user_workflow_multi_batch_row_count_range_rule_and_colum
...
+        # noinspection PyUnusedLocal
+        result = validator.expect_column_values_to_be_between(
+            column=column_name,
+            mostly=1.0,
+            result_format="SUMMARY",
+            include_config=True,
+            auto=False,
+            profiler_config=None,
+        )
+    assert "min_value and max_value cannot both be None" in str(e.value)
```
cdkini, line 357: ❤️

great_expectations/validator/validator.py
alexsherstinsky: @cdkini Logging is not needed here, because it is open to new domains to be needed; until then, no action.

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @cdkini This is a typical "get or create" pattern from our notebooks (I used it in tests for convenience).

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @cdkini It will be documented, and is enforced in the code (in `Validator._build_rule_based_profiler()` at the top).  Thanks.

great_expectations/validator/validator.py
alexsherstinsky: @cdkini Done -- good catch (committed).

great_expectations/validator/validator.py
alexsherstinsky: @cdkini -- Done (committed) -- thanks!

great_expectations/validator/validator.py
alexsherstinsky: @cdkini Perhaps for a later refactor (it is just me being explicit here; we could of course just say `dict`, but I wanted to be more descriptive so that in a future refactor we will immediately know "the type that is actually there".  Thanks.

great_expectations/validator/validator.py
alexsherstinsky: @cdkini Not yet -- we should do it throughout.

great_expectations/validator/validator.py
alexsherstinsky: This can be a follow-on improvement.  Perhaps define the types -- but this is just our profiler/rules configuration structure (nothing new, I am just "explicitly showing what is in there so that in the future we can refactor).

alexsherstinsky:
> Thank you very much for this @alexsherstinsky. Going forward, are we planning on building a `profiler_config` for each Expectation to enable default self-initialization? Or is a subset that we would be adding this to? I ask because the use-case and benefit of these configs is clear (in the scope of this PR) but wanting to get a better picture of our expected endpoint

@Shinnnyshinshin I am hoping that we can -- together -- build as many of default profilers for the expectations as our time allows; getting a prioritized list of most relevant/popular expectations would be great.  I think that "in value set" would be the next great candidate. /cc @anthonyburdi -- thanks!

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @jcampbell Actually, I had to revert -- by leaving as `ndarray` I was getting serialization issues in the notebook.  Thanks.

great_expectations/validator/validator.py
cdkini: Yup don't block merge on this!

great_expectations/validator/validator.py
cdkini: Will resolve this comment upon writing or identifying a ticket - I can add one in the backlog when I'm on tomorrow (unless you want to cover it).

alexsherstinsky: @cdkini I created a ticket in JIRA -- thanks!

great_expectations/expectations/core/expect_column_values_to_be_between.py
donaldheppner: +1 on using real types here that can encapsulate the reconciliation logic and even pair down the default values substantially. For example, you can imagine not having to serialize any default values.

As it stands, this bucket of hardcoded strings will be impossible to maintain and a source for many future defects. Functionally what is happening here is amazing, but I'd rather move slowly and create something maintainable than adopt this pattern over the rest of our expectations.

great_expectations/validator/validator.py
donaldheppner: Please remove the Dict here, we don't need to support both.

great_expectations/validator/validator.py
```diff
@@ -360,6 +377,304 @@ def inst_expectation(*args, **kwargs):
...
+                        Optional[Dict[str, Any]],
+                        Optional[Dict[str, Dict[str, Any]]],
+                    ],
+                ],
+            ]
+        ] = expectation_kwargs.get("profiler_config")
+        default_profiler_config: Optional[
+            Union[
+                RuleBasedProfilerConfig,
+                Dict[
```
donaldheppner: No Dict please, if at all possible. This doubles the code paths and more than doubles the defects

great_expectations/validator/validator.py
```diff
@@ -360,6 +377,304 @@ def inst_expectation(*args, **kwargs):
...
+
+        configuration: ExpectationConfiguration
+
+        if auto:
+            # Save custom Rule-Based Profiler configuration for reconciling it with optionally-specified default
+            # Rule-Based Profiler configuration as an override argument to "BaseRuleBasedProfiler.run()" method.
+            override_profiler_config: Optional[
+                Union[
+                    RuleBasedProfilerConfig,
+                    Dict[
```
donaldheppner: No Dict please

great_expectations/validator/validator.py
```diff
@@ -360,6 +377,304 @@ def inst_expectation(*args, **kwargs):
...
+    def _build_rule_based_profiler(
+        self,
+        expectation_type: str,
+        expectation_kwargs: dict,
+        success_keys: Tuple[str],
+        profiler_config: RuleBasedProfilerConfig,
+        override_profiler_config: Optional[
+            Union[
+                RuleBasedProfilerConfig,
+                Dict[
```
donaldheppner: No Dict please

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
```diff
@@ -232,6 +236,705 @@ def test_bobby_profiler_user_workflow_multi_batch_row_count_range_rule_and_colum
...
+    parameter_builder_batch_request: dict = {
+        "datasource_name": "taxi_pandas",
+        "data_connector_name": "monthly",
+        "data_asset_name": "my_reports",
+        "data_connector_query": {
+            "index": ":-1",
+        },
+    }
+
+    custom_profiler_config = {
```
donaldheppner: this should be a config type

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
```diff
@@ -404,3 +1174,279 @@ def test_quentin_profiler_user_workflow_multi_batch_quantiles_value_ranges_rule(
...
+    parameter_builder_batch_request = {
+        "datasource_name": "taxi_pandas",
+        "data_connector_name": "monthly",
+        "data_asset_name": "my_reports",
+        "data_connector_query": {
+            "index": ":-1",
+        },
+    }
+
+    custom_profiler_config = {
```
donaldheppner: this should be a config type

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
```diff
@@ -404,3 +1174,279 @@ def test_quentin_profiler_user_workflow_multi_batch_quantiles_value_ranges_rule(
...
+            )
+
+    # Use all batches, loaded by Validator, for estimating Expectation argument values.
+    parameter_builder_batch_request = {
+        "datasource_name": "taxi_pandas",
+        "data_connector_name": "monthly",
+        "data_asset_name": "my_reports",
+    }
+
+    custom_profiler_config = {
```
donaldheppner: this should be a config type

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
```diff
@@ -232,6 +236,705 @@ def test_bobby_profiler_user_workflow_multi_batch_row_count_range_rule_and_colum
...
+
+    batch_request: dict
+
+    validator: Validator
+
+    result: ExpectationValidationResult
+
+    custom_profiler_config: dict
+
+    custom_profiler_config = {
```
donaldheppner: this should be a config type

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
```diff
@@ -232,6 +236,705 @@ def test_bobby_profiler_user_workflow_multi_batch_row_count_range_rule_and_colum
...
+
+    # Use one batch (at index "1"), loaded by Validator as active_batch, for estimating Expectation argument values.
+    batch_request = {
+        "datasource_name": "taxi_pandas",
+        "data_connector_name": "monthly",
+        "data_asset_name": "my_reports",
+        "data_connector_query": {"index": 1},
+    }
+
+    custom_profiler_config = {
```
donaldheppner: this should be a config type

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
```diff
@@ -232,6 +236,705 @@ def test_bobby_profiler_user_workflow_multi_batch_row_count_range_rule_and_colum
...
+        "data_asset_name": "my_reports",
+    }
+
+    validator = context.get_validator(
+        batch_request=BatchRequest(**batch_request),
+        expectation_suite_name=expectation_suite_name,
+    )
+    assert len(validator.batches) == 3
+
+    custom_profiler_config = {
```
donaldheppner: this should be a config type

donaldheppner (changes requested):

great_expectations/validator/validator.py
alexsherstinsky: @donaldheppner The dictionary is offered here under the assumption that the user can supply either a dictionary or an instantiated typed object.  I do not think that at this point we should require users to go an extra step and instantiate the object -- we have not held this stance thus far, because of maximal convenience we want to offer in our main user-facing API -- `validator.expect_*` (for other APIs, I agree that we should consider more canonical ways).  Thanks.

great_expectations/validator/validator.py
alexsherstinsky: @donaldheppner -- this is the same situation.  The `custom_profiler_config` is externally facing -- and is available for users to be able override the default `profiler_config` of an Expectation.

great_expectations/validator/validator.py
alexsherstinsky: @donaldheppner -- this is the same situation.  The `custom_profiler_config` is externally facing -- and is available for users to be able override the default `profiler_config` of an Expectation.

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner -- same reason -- users may not want to go through the extra step of creating the config object -- we can give them both options.

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner Again -- this is a custom override, provided by users (who may find plain dictionary more convenient -- at least we have held this pattern for user-facing `batch_request` API).

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner Again -- this is a custom override, provided by users (who may find plain dictionary more convenient -- at least we have held this pattern for user-facing `batch_request` API).

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner Again -- this is a custom override, provided by users (who may find plain dictionary more convenient -- at least we have held this pattern for user-facing `batch_request` API).

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner Again -- this is a custom override, provided by users (who may find plain dictionary more convenient -- at least we have held this pattern for user-facing `batch_request` API).

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner Again -- this is a custom override, provided by users (who may find plain dictionary more convenient -- at least we have held this pattern for user-facing `batch_request` API).

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @cdkini This is an old code -- so I would rather defer this enhancement to a future work.  Thanks.

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
```diff
@@ -74,11 +74,11 @@ class ExpectColumnQuantileValuesToBeBetween(ColumnExpectation):
            <great_expectations.execution_engine.MetaExecutionEngine.column_aggregate_expectation>`.
 
            Args:
-               column (str): \
+               :column (str): \
```
anthonyburdi: Just a nitpick, but we are standardizing on the google python styleguide which doesn't have preceding colons and uses the `Attributes` style : https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
anthonyburdi: While I think we should not instantiate a `Profiler` object here we should instantiate a `ProfilerConfig` object instead of a dictionary to take advantage of the validation, documentation and guiding parameters of the ProfilerConfig class. Alex and I discussed and this will be changed before merge.

great_expectations/expectations/core/expect_column_values_to_be_between.py
anthonyburdi: Reiterating here: While I think we should not instantiate a `Profiler` object here we should instantiate a `ProfilerConfig` object instead of a dictionary to take advantage of the validation, documentation and guiding parameters of the `ProfilerConfig` class. Alex and I discussed and this will be changed before merge.

great_expectations/validator/validator.py
```diff
@@ -360,6 +377,304 @@ def inst_expectation(*args, **kwargs):
 
         return inst_expectation
 
+    def _build_expectation_configuration(
```
anthonyburdi, line 378: As discussed, it would be great for code organization to split the self-building expectation functionality out into a separate class in a new module and invoke it here. The Validator already has many responsibilities. Choices include a new module / class in the expectation module or the rule_based_profiler module. I'm actually leaning towards the former but happy to discuss.
In addition, breaking this method up into separate methods describing what is happening would be helpful for understanding, e.g. (pseudocode, return values, type hints, etc omitted):
```
def _build_expectation_configuration(...args...):
    _validate_input_params(...args...)
    if auto:
        if user_submitted_profiler_config is not None:
            if not _check_user_submitted_profiler_config_is_complete(...args...):
                _merge_user_profiler_config_with_default(...args...)
        expectation_configuration = profiler.run(...args...)
    else:
        expectation_configuration = ExpectationConfiguration(...args...)

    return expectation_configuration

```

great_expectations/validator/validator.py
anthonyburdi: See my suggestion above about isolating the logic merging different types of user input into a separate class and separate private methods so that all other methods and classes can use the typed ProfilerConfig object. This can be similar to the reconciliation methods on the RuleBasedProfiler.

great_expectations/validator/validator.py
```diff
@@ -360,6 +377,304 @@ def inst_expectation(*args, **kwargs):
...
+        else:
+            configuration = ExpectationConfiguration(
+                expectation_type=expectation_type,
+                kwargs=expectation_kwargs,
+                meta=meta,
+            )
+
+        return configuration
+
+    def _build_rule_based_profiler(
```
anthonyburdi, line 481: Can we move this to the RuleBasedProfiler class or utilize some of the logic already in the RuleBasedProfiler here?

great_expectations/validator/validator.py
```diff
@@ -360,6 +377,304 @@ def inst_expectation(*args, **kwargs):
...
+            self._validate_rule_and_update_rule_properties(
+                rule=rule,
+                expectation_type=expectation_type,
+                expectation_kwargs=expectation_kwargs,
+                success_keys=success_keys,
+            )
+
+        return profiler
+
+    def _validate_rule_and_update_rule_properties(
```
anthonyburdi, line 572: Can we move this to the RuleBasedProfiler class or utilize some of the logic already in the RuleBasedProfiler here?

anthonyburdi (dismissed):
This is a huge step forward. Just some code organization comments and clarification on using typed objects.

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @anthonyburdi I will check -- PyCharm was complaining.

alexsherstinsky: @anthonyburdi I handled it (in the next commit) by rolling back to the way it was before my changes and disabled the PyCharm warning (so it will be compliant with our style guide).

great_expectations/validator/validator.py
alexsherstinsky: @anthonyburdi How about we get this PR into the mergeable state and tackle this one as a maintenance-style enhancement shortly after?  I have other ideas for (could be same) maintenance-level PR, once this PR and PR #4269 
 are merged.  Thanks!

great_expectations/validator/validator.py
alexsherstinsky: @anthonyburdi All can be done -- proposing as part of separate maintenance PR.  Thanks.

great_expectations/validator/validator.py
alexsherstinsky: @anthonyburdi All can be done -- proposing as part of separate maintenance PR. Thanks.

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @jcampbell This got done in a better way to use typed configuration object (hat tip to @donaldheppner and @anthonyburdi ) as well as per your suggestion to define the default profiler above.  Thanks.

great_expectations/expectations/core/expect_column_values_to_be_between.py
alexsherstinsky: @jcampbell This got done in a better way to use typed configuration object (hat tip to @donaldheppner and @anthonyburdi ) as well as per your suggestion to define the default profiler above.  Thanks.

great_expectations/validator/validator.py
alexsherstinsky: @donaldheppner This got done, based on our conversation.  Thanks!

great_expectations/validator/validator.py
alexsherstinsky: @donaldheppner This got done, based on our conversation.  Thanks!

great_expectations/validator/validator.py
alexsherstinsky: @donaldheppner This got done, based on our conversation.  Thanks!

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner This got done, based on our conversation.  Thanks!

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner This got done, based on our conversation.  Thanks!

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner This got done, based on our conversation.  Thanks!

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner This got done, based on our conversation.  Thanks!

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner This got done, based on our conversation.  Thanks!

tests/integration/profiling/rule_based_profilers/test_profiler_user_workflows.py
alexsherstinsky: @donaldheppner This got done, based on our conversation.  Thanks!

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
```diff
@@ -131,16 +133,81 @@ class ExpectColumnQuantileValuesToBeBetween(ColumnExpectation):
...
+            "num_bootstrap_samples": 9999,
+            "false_positive_rate": 0.05,
+            "truncate_values": {
+                "lower_bound": None,
+                "upper_bound": None,
+            },
+            "round_decimals": 1,
+        },
+        rules={
+            "column_quantiles_rule": {
```
donaldheppner, line 159: are there types we can use here, too?

great_expectations/expectations/core/expect_column_values_to_be_between.py
```diff
@@ -91,6 +92,62 @@ class ExpectColumnValuesToBeBetween(ColumnMapExpectation):
...
+            "strict_min": False,
+            "strict_max": False,
+        },
+        rules={
+            "default_column_values_between_rule": {
+                "domain_builder": {
+                    "class_name": "ColumnDomainBuilder",
+                    "module_name": "great_expectations.rule_based_profiler.domain_builder",
+                },
+                "parameter_builders": [
```
donaldheppner, line 113: can we use types all through this config block?

great_expectations/validator/validator.py
donaldheppner: Can we now replace this type hint with `Optional[RuleBasedProfilerConfig]`?

great_expectations/expectations/core/expect_table_row_count_to_be_between.py
```diff
@@ -71,14 +72,64 @@ class ExpectTableRowCountToBeBetween(TableExpectation):
...
+            "num_bootstrap_samples": 9999,
+            "false_positive_rate": 0.05,
+            "truncate_values": {
+                "lower_bound": 0,
+                "upper_bound": None,
+            },
+            "round_decimals": 0,
+        },
+        rules={
+            "row_count_range_rule": {
```
donaldheppner, line 92: can we type this as well?

great_expectations/validator/validator.py
donaldheppner: Can we now replace this type hint with `Optional[RuleBasedProfilerConfig]`?

donaldheppner (changes requested):

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @donaldheppner Not yet -- for now, we follow exactly the same pattern for "Base" level business objects' configuration objects as established in `DataContextConfig` and `CheckpointConfig`.  Potentially can be done in the future, but a bit too many changes if attempted to do here.

great_expectations/expectations/core/expect_column_values_to_be_between.py
alexsherstinsky: @donaldheppner Not yet -- for now, we follow exactly the same pattern for "Base" level business objects' configuration objects as established in `DataContextConfig` and `CheckpointConfig`.  Potentially can be done in the future, but a bit too many changes if attempted to do here.

great_expectations/expectations/core/expect_table_row_count_to_be_between.py
alexsherstinsky: @donaldheppner Not yet -- for now, we follow exactly the same pattern for "Base" level business objects' configuration objects as established in `DataContextConfig` and `CheckpointConfig`.  Potentially can be done in the future, but a bit too many changes if attempted to do here.

alexsherstinsky: @donaldheppner Also, we are keeping the option for users to specify either dictionary or config object in the calls to `validator.expect_*`.  If/when the product decision is made to remove this flexibility, then it would be a very quick fix.  Thanks.

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
donaldheppner: Agreed, I've added a task to the sprint for a refactor of this config type.

alexsherstinsky:
Based on several conversations with @donaldheppner and @anthonyburdi the chosen direction is to require that the (optional) user-specified `profiler_config` parameter of the `validator.expect_*()`  method be a typed `RuleBasedProfilerConfig` config object, even though it entails more keystrokes than if we also allowed the Python `dict` equivalent as an alternative.  Even though this will be the first time that `validator.expect_*()` accepts a typed object, rather than a primitive type, we are making this decision, speaking in the same voice, as a directional move towards canonical APIs. 

The next update to this PR will reflect the above change.  Thank you.

/cc @jcampbell @abegong

donaldheppner (approved):
Very cool work, Alex, and the beginning of great things for GE!!

great_expectations/expectations/core/expect_column_quantile_values_to_be_between.py
alexsherstinsky: @jcampbell In the end, I was actually able to remove this conversion and get this validation to work the way it was written originally.  Thanks!

PR #4248, opened 2022-02-22
[BUGFIX] Improve pandas version checking in test_expectations[_cfe].py files

kenwade4:
There is an Expectation (`expect_column_values_to_be_of_type`) with some tests that are "only_for" specific versions of pandas. The code in the `tests/test_definitions/test_expectations[_cfe].py` files were lazily parsing the pandas version (which was fine back when pandas only had versions matching 0.x.x; current latest is 1.4.1)

Changes proposed in this pull request:
- Improve pandas version checking in `test_expectations[_cfe].py` files

**NOT changing in `self_check/util.py`** since it has been handled in a large refactor that's already merged to the hackathon-docs branch (which should be merged into develop sometime soon) https://github.com/great-expectations/great_expectations/blob/953c96be8366bef476364f82f7adb467d07a236b/great_expectations/self_check/util.py#L1620-L1630

### Definition of Done
Please delete options that are not relevant.

- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code

PR #4263, opened 2022-02-23
[BUGFIX] Check for pandas>=024 not pandas>=24

kenwade4:
In a fast follow to [#4248](https://github.com/great-expectations/great_expectations/pull/4248) it appears that the `test_expectations[_cfe].py` files were always wrong since [expect_column_values_to_be_of_type](https://github.com/great-expectations/great_expectations/blob/7fb3877c5a342bf4a55863cb9880677619c34525/tests/test_definitions/column_map_expectations/expect_column_values_to_be_of_type.json#L401) always specified that one test was `"only_for": ["pandas>=024"]` (not `pandas>=24`)


Shinnnyshinshin (approved):
👍🏼

PR #4264, opened 2022-02-23
[FEATURE] ProfilerRunAnonymizer

cdkini:
- PR 1 of 3 to implement usage stats for the RBP
- This PR focuses on the ProfilerRunAnonymizer class and is heavily inspired by the CheckpointRunAnonymizer

### Definition of Done


great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,205 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class ProfilerRunAnonymizer(Anonymizer):
+    def __init__(self, salt: Optional[str] = None) -> None:
+        super().__init__(salt=salt)
+
+        self._salt = salt
+        self._batch_request_anonymizer = BatchRequestAnonymizer(self._salt)
```
cdkini, line 19: Since this is used in the nested builder configs, I figured it made sense to just set a private attr and leverage that - this is a slight deviation from the CheckpointRunAnonymizer.

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,205 @@
...
+        variables: Optional[Dict[str, Any]] = None,
+        rules: Optional[Dict[str, Dict[str, Any]]] = None,
+    ) -> dict:
+        runtime_config: dict = profiler.config.to_dict()
+
+        # If applicable, override config attributes with runtime args
+        if variables:
+            runtime_config["variables"] = variables
+        if rules:
+            runtime_config["rules"] = rules
```
cdkini, line 188: This might need a bit of work? If the user has supplied runtime overrides to the `run` method, we want to ensure that our payload accounts for them.

The actual override happens in the method but we also need to account for them here if we want accurate usage stats.

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,205 @@
...
+
+        # Ensure that the list of rules contains all relevant information (including rule names)
+        resolved_rules: List[dict] = []
+        for name, rule in runtime_config["rules"].items():
+            rule["name"] = name
+            resolved_rules.append(rule)
```
cdkini: In order to get the name of each rule, we need to ensure the resulting dictionary representation of the rule has it's name as a field. If we don't care about the anonymized name of the rule, we can forgo this.

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        Traverse the entire RuleBasedProfiler configuration structure (as per its formal, validated Marshmallow schema) and
+        anonymize every field that can be customized by a user (public fields are recorded as their original names).
+        """
+
+        name: Optional[str] = kwargs.get("name")
+        anonymized_name: Optional[str] = self.anonymize(name)
+
+        config_version: Union[Number, str] = kwargs.get("config_version", 1.0)
+        variable_count: Number = kwargs.get("variable_count", 0)
```
Shinnnyshinshin: I thought the choice for `Number` type here was super-interesting. What was the reasoning behind it?

Shinnnyshinshin (approved):
Really great @cdkini. Thank you very much for this. Beautifully scoped, with separation of concerns and very easy to follow.

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
cdkini: that's a great question! I stole this bit of code from the `CheckpointRunAnonymizer` 😬 

I think float makes more sense!

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,196 @@
...
+            "anonymized_expectation_configuration_builders"
+        ] = self._anonymize_expectation_configuration_builders(
+            expectation_configuration_builders
+        )
+
+        return anonymized_rule
+
+    def _anonymize_domain_builder(self, domain_builder: dict) -> dict:
+        anonymized_domain_builder: dict = {}
+        anonymized_domain_builder["class_name"] = domain_builder.get("class_name")
```
anthonyburdi: This class name will need to be anonymized if it is not one of the GE provided class names. In other anonymizers we've used a constant to hold a list of class names to check. That might be difficult to keep in sync, maybe there is a better way or at least an opportunity to add a test for completeness on that constant e.g. that all subclasses of DomainBuilder in a vanilla GE install are captured in the list.

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,196 @@
...
+        return anonymized_parameter_builders
+
+    def _anonymize_parameter_builder(self, parameter_builder: dict) -> dict:
+        anonymized_parameter_builder: dict = {}
+
+        anonymized_parameter_builder["anonymized_name"] = self.anonymize(
+            parameter_builder.get("name")
+        )
+
+        anonymized_parameter_builder["class_name"] = parameter_builder.get("class_name")
```
anthonyburdi: This class name will also need to be anonymized if it is not one of the GE provided class names. See note above.

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,196 @@
...
+    def _anonymize_domain_builder(self, domain_builder: dict) -> dict:
+        anonymized_domain_builder: dict = {}
+        anonymized_domain_builder["class_name"] = domain_builder.get("class_name")
+
+        batch_request: Optional[dict] = domain_builder.get("batch_request")
+        if batch_request:
+            anonymized_domain_builder[
+                "anonymized_batch_request"
+            ] = self._batch_request_anonymizer.anonymize_batch_request(
+                *(), **batch_request
```
anthonyburdi: Just for my understanding, is the `*()` necessary because the method signature has `args*` e.g.  `anonymize_batch_request(*args, **kwargs)`?

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,196 @@
...
+        return anonymized_expectation_configuration_builders
+
+    def _anonymize_expectation_configuration_builder(
+        self, expectation_configuration_builder: dict
+    ) -> dict:
+        anonymized_expectation_configuration_builder: dict = {}
+
+        anonymized_expectation_configuration_builder[
+            "class_name"
+        ] = expectation_configuration_builder.get("class_name")
```
anthonyburdi: This class name will also need to be anonymized if it is not one of the GE provided class names. See note above.

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,196 @@
...
+        anonymized_expectation_configuration_builder[
+            "expectation_type"
+        ] = expectation_configuration_builder.get("expectation_type")
```
anthonyburdi: This expectation_type will also need to be anonymized if it is not one of the GE provided expectation types. See note above.

tests/rule_based_profiler/test_profiler_run_anonymizer.py
```diff
@@ -0,0 +1,328 @@
...
+@pytest.fixture
+def profiler_run_anonymizer() -> ProfilerRunAnonymizer:
+    # Standardize the salt so our tests are deterinistic
+    salt: str = "00000000-0000-0000-0000-00000000a004"
+    anonymizer: ProfilerRunAnonymizer = ProfilerRunAnonymizer(salt=salt)
+    return anonymizer
+
+
+@pytest.fixture
+def usage_stats_profiler_config() -> dict:
```
anthonyburdi: These fixtures are great - can you add a version of the single and multiple rule config that also has custom class names in addition to GE supplied classes for DomainBuilder, ParameterBuilder, ExpectationConfigurationBuilder and custom Expectations - and test that the custom names are anonymized and the GE supplied are not?

tests/rule_based_profiler/test_profiler_run_anonymizer.py
```diff
@@ -0,0 +1,328 @@
...
+            {
+                "anonymized_domain_builder": {
+                    "anonymized_batch_request": {
+                        "anonymized_batch_request_required_top_level_properties": {
+                            "anonymized_data_asset_name": "eac128c5824b698c22b441ada61022d4",
+                            "anonymized_data_connector_name": "123a3221fc4b65014d061cce4a71782e",
+                            "anonymized_datasource_name": "df78ebde1957385a02d8736cd2c9a6d9",
+                        },
+                        "batch_request_optional_top_level_keys": [
+                            "data_connector_query"
```
anthonyburdi: Interesting that this `data_connector_query` is showing up even though you didn't provide it in the batch_request - do you have an idea of why that might be?

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,196 @@
...
+            runtime_config["rules"] = rules
+
+        runtime_config["variable_count"] = len(runtime_config["variables"])
+        runtime_config["rule_count"] = len(runtime_config["rules"])
+
+        for attr in ("class_name", "module_name", "variables"):
+            runtime_config.pop(attr)
+            logger.debug("Removed unnecessary attr %s from profiler config", attr)
+
+        return runtime_config
```
anthonyburdi: Is there a better location to put this method? Perhaps on the ProfilerConfig or RuleBasedProfiler rather than the anonymizer?

anthonyburdi (changes requested):
Very easy to follow code, thank you for spending the time to make this implementation so clean :) I do have some requested changes and a few questions.

tests/rule_based_profiler/test_profiler_run_anonymizer.py
cdkini: I am not! That's a great point though.

If we look at the call stack, here's what goes on:
```python
# batch_request_anonymizer - get_batch_request_from_acceptable_arguments
{
    "datasource_name": "datasource_name",
    "data_connector_name": "my_basic_data_connector",
    "data_asset_name": "data_asset_name",
    "data_connector_query": {"batch_filter_parameters": {}},
    "limit": None,
    "batch_spec_passthrough": {}
}
```
```python
# batch_request_anonymizer - anonymize_batch_request_properties
{
    "datasource_name": ...,
    "data_connector_name": ...,
    "data_asset_name": ...,
    "data_connector_query": {},
}
```

Is this unexpected behavior?

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
cdkini: great idea! moved it to the config class since we're "resolving" config. Note that the equivalent anonymizer for checkpoint runs has the same static method - if we like this pattern, I can write a ticket to refactor that into the config as well.

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
cdkini: This was just a pattern I copied over from the checkpoint run anonymizer without thinking 😅 

The `*()` is not necessary here; I've gone ahead and removed it.

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
cdkini: I am attempting to resolve this across the codebase in another PR - will address all of these through that (if possible) or will individually update these.

Open for my implementation, I'll reach out when this is in a good spot.

tests/rule_based_profiler/test_profiler_run_anonymizer.py
cdkini: Added!

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,251 @@
...
+        self._ge_expectation_configuration_builders = [
+            DefaultExpectationConfigurationBuilder,
+            ExpectationConfigurationBuilder,
+        ]
+        self._ge_expectation_types = set(list_registered_expectation_implementations())
+
+        self._salt = salt
+        self._batch_request_anonymizer = BatchRequestAnonymizer(self._salt)
+
+    def anonymize_profiler_run(self, *args: List[Any], **kwargs: dict) -> dict:
```
anthonyburdi: Would it make more sense here to take the ProfilerConfig or BaseRuleBasedProfiler object instead of a dict? That way we can solidify the interface and not have to pull data via named kwargs.

tests/rule_based_profiler/test_profiler_run_anonymizer.py
anthonyburdi: It is unexpected - when `data_connector_query` is not provided, I would expect it to not appear in the `batch_request_optional_top_level_keys`. Any idea why it is getting added?

tests/rule_based_profiler/test_profiler_run_anonymizer.py
```diff
@@ -0,0 +1,525 @@
...
+                ],
+            }
+        ],
+        "config_version": 1.0,
+        "rule_count": 1,
+        "variable_count": 1,
+    }
+
+
+def test_resolve_config_using_acceptable_arguments(
```
anthonyburdi: Thank you for these tests. Though tests related to `resolve_config_using_acceptable_arguments` should go into the main RuleBasedProfilerConfig test file, not the anonymizer. Would you mind moving?

tests/rule_based_profiler/test_profiler_run_anonymizer.py
```diff
@@ -0,0 +1,525 @@
...
+    )
+
+    domain_builder: dict = config["rules"]["rule_1"]["domain_builder"]
+    converted_batch_request: dict = domain_builder["batch_request"]
+    assert converted_batch_request["datasource_name"] == datasource_name
+    assert converted_batch_request["data_connector_name"] == data_connector_name
+    assert converted_batch_request["data_asset_name"] == data_asset_name
+
+
+def test_all_builder_classes_are_recognized_as_profiler_run_anonymizer_attrs(
```
anthonyburdi, line 300: ❤️❤️❤️❤️❤️❤️❤️
Love this test. Until we can refactor this will help us not forget to add new builder classes to the anonymizer. Thank you for writing this, our future selves will also thank you.

anthonyburdi (commented):
Thank you for following up on the changes requested. I left a few additional questions and requests.

tests/rule_based_profiler/test_profiler_run_anonymizer.py
cdkini: ah great call! moving now

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
cdkini: I like that a lot! I believe we can access the config on a Profiler instance through `.config` so I'll use that so we get stricter types.

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,253 @@
...
+        self._ge_expectation_configuration_builders = [
+            DefaultExpectationConfigurationBuilder,
+            ExpectationConfigurationBuilder,
+        ]
+        self._ge_expectation_types = set(list_registered_expectation_implementations())
+
+        self._salt = salt
+        self._batch_request_anonymizer = BatchRequestAnonymizer(self._salt)
+
+    def anonymize_profiler_run(self, profiler_config: RuleBasedProfilerConfig) -> dict:
```
cdkini, line 94: Rather than `*args` and `**kwargs`, this now just takes a config object - this contains everything we need and allows us to leverage types.

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -303,6 +303,29 @@ def get_config_class(cls) -> Type["RuleBasedProfilerConfig"]:  # noqa: F821
...
         return RuleBasedProfilerConfigSchema
 
+    # noinspection PyUnusedLocal,PyUnresolvedReferences
+    @classmethod
+    def resolve_config_using_acceptable_arguments(
+        cls,
+        profiler: "RuleBasedProfiler",  # noqa: F821
+        variables: Optional[Dict[str, Any]] = None,
+        rules: Optional[Dict[str, Dict[str, Any]]] = None,
+    ) -> "RuleBasedProfilerConfig":
```
cdkini: Although our configs are not currently immutable through frozen dataclasses, I like the idea that configs are not to be touched after instantiation.

As such, I've made this resolution method a class method and ensured that it returned an entirely new instance of the config - it may or may not contain new values depending on runtime overrides.

Thoughts here? I could as easily modify the existing object so I'd love to know the pros/cons.

tests/rule_based_profiler/config/test_base.py
```diff
@@ -279,3 +283,65 @@ def test_rule_based_profiler_from_commented_map():
...
+    config: RuleBasedProfilerConfig = (
+        RuleBasedProfilerConfig.resolve_config_using_acceptable_arguments(
+            profiler=profiler_with_placeholder_args
+        )
+    )
+
+    # The resolved config is equivalent to the old one but is a new instance
+    old_config: RuleBasedProfilerConfig = profiler_with_placeholder_args.config
+    assert id(old_config) != id(config)
+    assert old_config.__dict__ == config.__dict__
```
cdkini: This extends that idea of immutability - the config is equivalent in value but it is an entirely different instance. This allows us to maintain the idealized version of a config (a snapshot of the arguments used to instantiate the business logic object).

great_expectations/core/usage_statistics/anonymizers/batch_request_anonymizer.py
```diff
@@ -47,11 +47,14 @@ def anonymize_batch_request(
...
-                inplace=True,
-            )
+
+            # Clean object until all falsy values are removed
+            while any(not attr for attr in anonymized_batch_request_dict.values()):
+                deep_filter_properties_iterable(
+                    properties=anonymized_batch_request_dict,
+                    clean_falsy=True,
+                    inplace=True,
+                )
```
cdkini: @anthonyburdi this is the solution I came up with based on the issue with the top-level key showing up. This may not be the best solution.

cc @alexsherstinsky will link you to an additional thread for more context in case you know (since you wrote this filtering function).

tests/rule_based_profiler/test_profiler_run_anonymizer.py
cdkini: It isn't being removed by `deep_filter_properties_iterable` - I've added a workaround by essentially running that filter function as many times as needed to remove all falsy values but I think this may be a bug in our code.

great_expectations/core/usage_statistics/anonymizers/batch_request_anonymizer.py
```diff
@@ -47,11 +47,14 @@ def anonymize_batch_request(
             anonymized_batch_request_dict = standardize_batch_request_display_ordering(
                 batch_request=anonymized_batch_request_dict
             )
-            deep_filter_properties_iterable(
-                properties=anonymized_batch_request_dict,
-                clean_falsy=True,
-                inplace=True,
-            )
+
+            # Clean object until all falsy values are removed
```
alexsherstinsky: @cdkini Could you please explain why the original single-statement approach does not accomplish the needed goals, while the "while loop" (no pun intended) does?  Perhaps an example or a test (e.g., that endorses the new approach) would help.  Thanks!

great_expectations/rule_based_profiler/config/base.py
alexsherstinsky: @cdkini Whatever we do here, I would offer that we take inspiration from websites and databases.  For example, the web page retrieves the user's account from a database table, and it becomes the instantiated account object.  The user makes a change (e.g., specifies a communication preference).  This causes the Account object to become "dirty" in memory.  Then the user saves the change, which causes the account object to update the database entry.  So in this example, the Account object was never frozen.  We have a pattern in DataContext (which is really a filesystem-backed DataContext).  We load great_expections.yml into DataContextConfig, make changes (e.g., update a store setting), and then run _save_project_config to persist it back to a file.  Happy to discuss. /cc @donaldheppner

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -303,6 +303,29 @@ def get_config_class(cls) -> Type["RuleBasedProfilerConfig"]:  # noqa: F821
...
+        runtime_config: RuleBasedProfilerConfig = profiler.config
+
+        # If applicable, override config attributes with runtime args
+        runtime_rules: dict = rules if rules is not None else runtime_config.rules
+        runtime_variables: Optional[dict] = (
+            variables if variables is not None else runtime_config.variables
+        )
+
+        return cls(
```
alexsherstinsky, line 327: @cdkini I may be missing the crux of the matter here -- I was hoping to see something like one of our reconciliation methods run taking in the existing configuration (or parts thereof) and the supplied overrides (e.g., like we have in the tests of the reconciliation logic as well as, most recently, utilized for self-initializing expectations).  Happy to discuss.

tests/rule_based_profiler/config/test_base.py
alexsherstinsky: @cdkini Despite my previous example with the website/database example, this approach is totally fine (and can be better from code maintenance perspective, which is the main thrust of functional programming style).  Thanks. /cc @donaldheppner

tests/rule_based_profiler/conftest.py
```diff
@@ -1067,3 +1071,32 @@ def profiler_with_placeholder_args(
         **profiler_config_dict, data_context=empty_data_context
     )
     return profiler
+
+
+@pytest.fixture
+def usage_stats_profiler_config() -> RuleBasedProfilerConfig:
```
alexsherstinsky: @cdkini This seems close to our other profiler configuration fixture(s) -- any chance we can use that (with a potential modification)?  Even if we have to update a lot of existing tests a little bit, I feel that it could be worth it.  Thanks!

alexsherstinsky (changes requested):
@cdkini Strong stuff as anonymizers, especially for a complex function, such as `Checkpoint.run()` and `RuleBasedProfiler.run()` are very laborious to develop and test.  I found some things that compelled me to ask for clarification, so happy to discuss.  Thank you!

great_expectations/rule_based_profiler/config/base.py
cdkini: So all of these methods noted herein are actually run as part of the usage stats decorator on the run method - that reconciliation will still happen during the run but for purposes of usage statistics, we need a much simpler set of details.

tests/rule_based_profiler/test_profiler_run_anonymizer.py
cdkini: It looks like it doesn't get cleared out from `deep_filter_properties_iterable`.

We take in our `**kwargs` and create a batch request using `get_batch_request_from_acceptable_arguments`. This sets:
```python
    "data_connector_query": {"batch_filter_parameters": {}},
```

The filtering function results in:
```python
    "data_connector_query": {},
```

Since this key is part of the top-level (even though it has no associated value), it gets included. This seems to be a bug with the filtering function or the actual BatchRequestAnonymizer.

great_expectations/rule_based_profiler/config/base.py
cdkini: UPDATE - per our discussion, I've leveraged the existing logic to aid with reconciliation. I've also reconciled the variables since it makes sense to keep the logic the same between usage stats and the actual run method.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -404,6 +393,30 @@ def reconcile_profiler_variables(
 
         return effective_variables
 
+    def reconcile_profiler_variables_as_dict(
```
cdkini, line 396: Since we can use these in usage stats, making it public and refactoring primary reconciliation methods.

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -0,0 +1,262 @@
...
+            ge_classes=self._ge_expectation_configuration_builders,
+            runtime_environment={
+                "module_name": "great_expectations.rule_based_profiler.expectation_configuration_builder"
+            },
+        )
+
+        expectation_type: Optional[str] = expectation_configuration_builder.get(
+            "expectation_type"
+        )
+        if expectation_type in self._ge_expectation_types:
```
anthonyburdi: Non blocking & up to you, but you might be able to refactor expectation anonymization into a public `ExpectationSuiteAnonymizer.anonymize_expectation()` method and then use here and in `ExpectationSuiteAnonymizer` to reduce duplication.

tests/rule_based_profiler/config/test_base.py
```diff
@@ -279,3 +283,117 @@ def test_rule_based_profiler_from_commented_map():
...
+        RuleBasedProfilerConfig.resolve_config_using_acceptable_arguments(
+            profiler=profiler_with_placeholder_args
+        )
+    )
+
+    old_config: RuleBasedProfilerConfig = profiler_with_placeholder_args.config
+
+    # Brand new config is created but existing attributes are unchanged
+    assert id(old_config) != id(config)
+    assert all(
```
anthonyburdi, line 301: 👍

tests/rule_based_profiler/config/test_base.py
```diff
@@ -279,3 +283,117 @@ def test_rule_based_profiler_from_commented_map():
...
+        old_config[attr] == config[attr]
+        for attr in ("class_name", "config_version", "module_name", "name")
+    )
+
+
+def test_resolve_config_using_acceptable_arguments_with_runtime_overrides(
+    profiler_with_placeholder_args: RuleBasedProfiler,
+) -> None:
+    rule_name: str = "my_new_rule"
+    assert all(rule.name != rule_name for rule in profiler_with_placeholder_args.rules)
```
anthonyburdi: 👍  Nice

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
cdkini: might as well sneak it in here! let me respond to this once the change is up.

tests/rule_based_profiler/config/test_base.py
```diff
@@ -279,3 +283,117 @@ def test_rule_based_profiler_from_commented_map():
...
+    assert all(
+        old_config[attr] == config[attr]
+        for attr in ("class_name", "config_version", "module_name", "name")
+    )
+
+
+def test_resolve_config_using_acceptable_arguments_with_runtime_overrides(
+    profiler_with_placeholder_args: RuleBasedProfiler,
+) -> None:
+    rule_name: str = "my_new_rule"
```
anthonyburdi: Non blocking nitpick, up to you - I would name this something more obvious like `runtime_override_rule_name` and rename `rule` to `runtime_override_rule`, `rules` to `runtime_override_rules`, etc.

tests/rule_based_profiler/test_profiler_run_anonymizer.py
```diff
@@ -0,0 +1,338 @@
...
+    anonymized_result: dict = profiler_run_anonymizer.anonymize_profiler_run(
+        profiler_config_with_placeholder_args_custom_values
+    )
+    assert anonymized_result == {
+        "anonymized_name": "5b6c98e19e21e77191fb071bb9e80070",
+        "anonymized_rules": [
+            {
+                "anonymized_domain_builder": {
+                    "anonymized_class": "d2972bccf7a2a0ff91ba9369a86dcbe1",
+                    "parent_class": "__not_recognized__",
```
anthonyburdi, line 67: I was going to ask for one more test with actual custom classes that are subclassed from GE classes (so this would show the GE parent class), but we already test that code works in test_anonymizer.py `test_anonymizer__is_parent_class_recognized`.

anthonyburdi (approved):
LGTM - some optional refactor steps if you want but otherwise 👍

tests/rule_based_profiler/config/test_base.py
cdkini: done!

great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
cdkini: done!

tests/rule_based_profiler/test_profiler_run_anonymizer.py
cdkini: ah that makes a lot of sense! I think I'll defer this for the time being since we have coverage through the parent but will put this on my radar.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -382,20 +382,9 @@ def reconcile_profiler_variables(
...
-                variables_configs = nested_update(
-                    variables_configs,
-                    variables,
-                )
-            elif reconciliation_strategy == ReconciliationStrategy.REPLACE:
-                variables_configs = variables
-            elif reconciliation_strategy == ReconciliationStrategy.UPDATE:
-                variables_configs.update(variables)
-
+            variables_configs: dict = self.reconcile_profiler_variables_as_dict(
```
alexsherstinsky, line 385: @cdkini Personal preference, if you do not mind -- use "keyword style" (rather than positional) arguments.  Thanks!

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -303,6 +303,34 @@ def get_config_class(cls) -> Type["RuleBasedProfilerConfig"]:  # noqa: F821
     def get_schema_class(cls) -> Type["RuleBasedProfilerConfigSchema"]:  # noqa: F821
         return RuleBasedProfilerConfigSchema
 
+    # noinspection PyUnusedLocal,PyUnresolvedReferences
+    @classmethod
+    def resolve_config_using_acceptable_arguments(
```
alexsherstinsky, line 308: @cdkini I do not see `resolve_config_using_acceptable_arguments()` being used anywhere (other than in tests).  Is this here for a next PR?  Thanks!

alexsherstinsky (changes requested):
@cdkini Solid work -- just a question and a tiny change request.  Thank you!

great_expectations/rule_based_profiler/config/base.py
cdkini: yes it is to come in the final PR.

great_expectations/rule_based_profiler/rule_based_profiler.py
cdkini: sure thing!

cdkini: resolved

alexsherstinsky (approved):
LGTM

PR #4269, opened 2022-02-23
[FEATURE] ValueSetMultiBatchParameterBuilder and CategoricalColumnDomainBuilder

anthonyburdi:
- This feature adds cardinality related functionality to the rule based profiler:
  - CategoricalColumnDomainBuilder allows filtering of domains by cardinality (number of unique values)
  - ValueSetMultiBatchParameterBuilder allows building the unique value_set from the filtered domains for use in creating Expectations. This unique value_set is the unique values from ALL batches accessible to the parameter builder. For example, if batch 1 has the unique values {1, 4, 8} and batch 2 {2, 8, 10} the unique values are the set union, or {1, 2, 4, 8, 10}
  - It also adds a separate CardinalityChecker class which isolates the logic related to determining if measured cardinality meets the configured and supplies 25+ presets along with the flexibility of specifying a custom limit of either the number of unique values or the proportion of unique values.
- It also adds tests for the above functionality, including example configurations in the existing alice & bobby integration tests.

### Definition of Done

Thank you for submitting!


great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
```diff
@@ -0,0 +1,211 @@
...
+class CardinalityLimitMode(enum.Enum):
+    """Preset limits based on unique values (cardinality)
+
+    Defines relative (ratio) and absolute number of records (table rows) that
+    correspond to each cardinality category.
+
+    Used to determine appropriate Expectation configurations based on data.
+    """
+
+    ZERO = AbsoluteCardinalityLimit("ZERO", 0)
```
alexsherstinsky, line 35: @anthonyburdi Type hints for each, perhaps?.. not sure if overkill (I put them in for the profiler reconciliation strategy).  Thanks.

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
```diff
@@ -0,0 +1,211 @@
...
+    related_metric_name: str = "column.unique_proportion"
+
+
+@dataclass
+class AbsoluteCardinalityLimit(CardinalityLimit):
+    max_unique_values: int
+    related_metric_name: str = "column.distinct_values.count"
+
+
+class CardinalityLimitMode(enum.Enum):
```
alexsherstinsky, line 26: @anthonyburdi Any objections to import `Enum` from `enum`?..

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+)
+from great_expectations.rule_based_profiler.domain_builder.domain_builder import (
+    build_domains_from_column_names,
+)
+from great_expectations.rule_based_profiler.types import Domain, ParameterContainer
+from great_expectations.validator.metric_configuration import MetricConfiguration
+from great_expectations.validator.validator import Validator
+
+
+class CategoricalColumnDomainBuilder(DomainBuilder):
```
alexsherstinsky, line 17: @anthonyburdi Can this domain builder allow columns to be set on it just like `ColumnDomainBuilder`?

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+        Categorical generally refers to columns with relatively limited
+        number of unique values.
+        Limit mode can be absolute (number of unique values) or relative
+        (proportion of unique values). You can choose one of: limit_mode,
+        max_unique_values or max_proportion_unique to specify the cardinality
+        limit.
+        Note that the limit must be met for each batch separately that is
+        supplied in the batch_request or the column domain will not be included.
+        Note that the columns used will be from the first batch retrieved
+        via the batch_request. If other batches contain additional columns,
```
alexsherstinsky, line 43: @anthonyburdi I am not seeing `batch` and `batch_request` property getters/setters -- I think that we need them for Self-Initializing Expectations (please see `ColumnDomainBuilder`).  Thanks!

great_expectations/rule_based_profiler/domain_builder/domain_builder.py
```diff
@@ -144,3 +144,26 @@ def get_batch_id(
             )
 
         return batch_ids[0]
+
+
+def build_domains_from_column_names(column_names: List[str]) -> List[Domain]:
```
alexsherstinsky: @anthonyburdi Do we want a column-domain specific logic in the general `DomainBuilder` module, or should it go to `util.py`?  My vote is for the latter.

Shinnnyshinshin: I second @alexsherstinsky's thought here. I think this function would be better-placed in a `util.py`

great_expectations/rule_based_profiler/domain_builder/__init__.py
```diff
@@ -1,9 +1,18 @@
-from .domain_builder import DomainBuilder  # isort:skip
-from .table_domain_builder import TableDomainBuilder  # isort:skip
-from .column_domain_builder import ColumnDomainBuilder  # isort:skip
-from .simple_column_suffix_domain_builder import (  # isort:skip
+from great_expectations.rule_based_profiler.domain_builder.domain_builder import (  # isort:skip
```
Shinnnyshinshin, line 1: Thank you so much for turning these into absolute imports :)

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
```diff
@@ -0,0 +1,130 @@
...
+)
+from great_expectations.rule_based_profiler.types.parameter_container import (
+    ParameterNode,
+)
+
+
+class ValueSetMultiBatchParameterBuilder(MetricMultiBatchParameterBuilder):
+    """Build a set of unique values across all specified batches.
+
+    This parameter can be used in building the unique value_set from the
```
Shinnnyshinshin: Absolutely LOVE the detailed docstring.

tests/rule_based_profiler/domain_builder/test_cardinality_checker.py
```diff
@@ -0,0 +1,192 @@
...
+
+@pytest.mark.parametrize(
+    "input_limit_mode",
+    [member for member in CardinalityLimitMode]
+    + [member.name for member in CardinalityLimitMode],
+)
+def test_exhaustively_cardinality_within_limit_for_all_supported_cardinality_limits(
+    input_limit_mode,
+):
+    """What does this test and why?
```
Shinnnyshinshin, line 129: <3

tests/rule_based_profiler/domain_builder/test_categorical_column_domain_builder.py
```diff
@@ -0,0 +1,306 @@
...
+            },
+        )
+        for column_name in alice_all_column_names
+    ]
+    assert domains == alice_all_column_domains
+
+    assert len(domains) == 1
+
+
+@mock.patch("great_expectations.data_context.data_context.DataContext")
```
Shinnnyshinshin, line 91: Absolutely love the unittesting

tests/rule_based_profiler/parameter_builder/test_value_set_multi_batch_parameter_builder.py
```diff
@@ -0,0 +1,171 @@
...
+        [[{1, 2, 3}, {1, 4, 5}], {1, 2, 3, 4, 5}],
+        [[[1], [2, 3]], {1, 2, 3}],
+        [[{1}, {2, 3}], {1, 2, 3}],
+        [[[1], [1, 2]], {1, 2}],
+        [[{1}, {1, 2}], {1, 2}],
+        [[[1], [1]], {1}],
+        [[{1}, {1}], {1}],
+    ],
+)
+def test__get_unique_values_from_iterable_of_iterables(test_input, expected):
```
Shinnnyshinshin: love this test

Shinnnyshinshin (approved):
Absolutely amazing @anthony, really well described, documented and tested (including unittesting patterns that I'm going to borrow heavily on).

I'm going to approve, but would like to  ask that:

1. @alexsherstinsky 's comments are addressed 
2. Other MarioPod members have had their say

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
```diff
@@ -0,0 +1,211 @@
+import abc
+import enum
+from dataclasses import dataclass
+from typing import Optional, Tuple, Union
+
+from great_expectations.exceptions import ProfilerConfigurationError
+
+
+@dataclass
```
alexsherstinsky, line 9: @anthonyburdi Some docstrings about what these classes are for would be nice...  Thanks!

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
```diff
@@ -0,0 +1,211 @@
...
+class RelativeCardinalityLimit(CardinalityLimit):
+    max_proportion_unique: Union[int, float]
+    related_metric_name: str = "column.unique_proportion"
+
+
+@dataclass
+class AbsoluteCardinalityLimit(CardinalityLimit):
+    max_unique_values: int
+    related_metric_name: str = "column.distinct_values.count"
```
alexsherstinsky: @anthonyburdi `related_metric_name` or `relevant_metric_name`? :)

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
```diff
@@ -0,0 +1,211 @@
...
+            limit_mode=limit_mode,
+            max_unique_values=max_unique_values,
+            max_proportion_unique=max_proportion_unique,
+        )
+
+    @property
+    def limit_mode(self) -> CardinalityLimitMode:
+        return self._limit_mode
+
+    def cardinality_within_limit(self, metric_value: Union[int, float]) -> bool:
```
alexsherstinsky: @anthonyburdi how about using `Number` type instead of `Union[int, float]` -- might it work?  Thanks!

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        self._validate_metric_value(metric_value=metric_value)
+        if isinstance(self._limit_mode, AbsoluteCardinalityLimit):
+            return metric_value <= self._limit_mode.max_unique_values
+        elif isinstance(self._limit_mode, RelativeCardinalityLimit):
+            return metric_value <= self._limit_mode.max_proportion_unique
+
+    @staticmethod
+    def _validate_metric_value(metric_value: Union[int, float]) -> None:
+        if not isinstance(metric_value, (int, float)):
```
alexsherstinsky, line 133: @anthonyburdi We have an `is_numeric()` checker in great_expectations/util.py -- perhaps it might work here?  Thanks!

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
```diff
@@ -0,0 +1,211 @@
...
+                max_proportion_unique=max_proportion_unique,
+            )
+
+    def _validate_input_parameters(
+        self,
+        limit_mode: Optional[Union[CardinalityLimitMode, str]] = None,
+        max_unique_values: Optional[int] = None,
+        max_proportion_unique: Optional[int] = None,
+    ) -> None:
+        params = (
```
alexsherstinsky: @anthonyburdi -- type hint for params?

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
```diff
@@ -0,0 +1,211 @@
...
+            max_unique_values,
+            max_proportion_unique,
+        )
+        num_params: int = sum([0 if param is None else 1 for param in params])
+        if num_params != 1:
+            raise ProfilerConfigurationError(
+                f"Please pass ONE of the following parameters: limit_mode, max_unique_values, max_proportion_unique, you passed {num_params} parameters."
+            )
+
+        if limit_mode is not None:
```
alexsherstinsky, line 197: @anthonyburdi Can this be accomplished with one `if` statement using and/or logic therein (and thus without the `pass`)?

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+        """Create column domains where cardinality is within the specified limit.
+
+        Cardinality refers to the number of unique values in a given domain.
+        Categorical generally refers to columns with relatively limited
+        number of unique values.
+        Limit mode can be absolute (number of unique values) or relative
+        (proportion of unique values). You can choose one of: limit_mode,
+        max_unique_values or max_proportion_unique to specify the cardinality
+        limit.
+        Note that the limit must be met for each batch separately that is
```
alexsherstinsky, line 40: @anthonyburdi In addition to satisfying cardinality requirements/limits at a `Batch` level, do we also have aggregated across all `Batch` objects requirements for cardinality (i.e., no more than X for all batches)?  Thanks.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+
+class CategoricalColumnDomainBuilder(DomainBuilder):
+    """
+    This DomainBuilder uses column cardinality to identify domains.
+    """
+
+    def __init__(
+        self,
+        data_context: "DataContext",  # noqa: F821
+        batch_request: Optional[Union[BatchRequest, RuntimeBatchRequest, dict]] = None,
```
alexsherstinsky, line 25: @anthonyburdi Constructor must also take in `batch: Optional[Batch] = None,` as an argument (please see `ColumnDomainBuilder`) -- this is important for Self-Initializing Expectations.  Thanks.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+        """Return domains matching the selected limit_mode.
+
+        Args:
+            variables: Optional variables to substitute when evaluating.
+
+        Returns:
+            List of domains that match the desired cardinality.
+        """
+
+        batch_ids: List[str] = self._get_batch_ids(variables=variables)
```
alexsherstinsky: @anthonyburdi I did not envision `get_batch_ids()` for `DomainBuilder` and hence made it private (as an internal-use-only method).  But your use case indicates that it should be made public -- might it make sense then to change `_get_batch_ids()` to `get_batch_ids()` throughout?  If you agree, then could you please affect the change?  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+        Returns:
+            List of column names from the first loaded batch.
+        """
+
+        if len(batch_ids) == 0:
+            raise ProfilerConfigurationError(
+                "No batch_ids were available, please check your configuration."
+            )
+
+        first_batch_id: str = batch_ids[0]
```
alexsherstinsky: @anthonyburdi Please note that the first `Batch` should probably be the `active_batch` -- and the index for it is: -1 (negative 1) -- thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+
+        table_column_names: List[str] = validator.get_metric(
```
alexsherstinsky, line 135: @anthonyburdi We should probably make this code that gets the columns available to all `ColumnDomainBuilder` types...  Thanks.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+                metric_value_kwargs=None,
+                metric_dependencies=None,
+            )
+        )
+
+        if self._exclude_columns is not None:
+            table_column_names = [
+                colname
+                for colname in table_column_names
+                if colname not in self._exclude_columns
```
alexsherstinsky: @anthonyburdi For things like `exclude_columns`, I would prefer it to be turned into a property and accessed as such herein.  At least that is the style I have been following -- and it plays well with the `Builder` parent class (and above).  Thanks.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+            List of dicts of the form [{column_name: List[MetricConfiguration]},...]
+        """
+
+        limit_mode: CardinalityLimitMode = self._cardinality_checker.limit_mode
+
+        metric_name: str = limit_mode.related_metric_name
+
+        metric_configurations: List[Dict[str, List[MetricConfiguration]]] = []
+
+        for column_name in column_names:
```
alexsherstinsky: @anthonyburdi type hint for `column_name` please?  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+
+        Returns:
+            List of dicts of the form [{column_name: List[MetricConfiguration]},...]
+        """
+
+        limit_mode: CardinalityLimitMode = self._cardinality_checker.limit_mode
+
+        metric_name: str = limit_mode.related_metric_name
+
+        metric_configurations: List[Dict[str, List[MetricConfiguration]]] = []
```
alexsherstinsky: @anthonyburdi The `metric_configurations` can be more elegantly gathered as a comprehension.  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+        Args:
+            validator: Validator used to compute column cardinality.
+            table_column_names: column names to verify cardinality.
+            metrics_for_cardinality_check: metric configurations used to
+                compute cardinality.
+
+        Returns:
+            List of column names meeting cardinality.
+        """
+        candidate_column_names: List[str] = table_column_names.copy()
```
alexsherstinsky, line 210: @anthonyburdi Maybe a better pattern would be instead of copying the entire list and then filtering out columns build the list out of columns that fit the criteria (using the `Python` `filter()` method)?  That is the pattern I followed in other `DomainBuilder` classes.  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+        #  by first building up MetricConfigurations to compute. See GREAT-584
+        for metric_for_cardinality_check in metrics_for_cardinality_check:
+            for (
+                column_name,
+                metric_config_batch_list,
+            ) in metric_for_cardinality_check.items():
+                if column_name not in candidate_column_names:
+                    continue
+                else:
+                    for metric_config in metric_config_batch_list:
```
alexsherstinsky, line 222: @anthonyburdi It does not seem to me that `metric_config_batch_list` matches the actual type -- which is `MetricConfig`, I think -- so should it not be called `metric_config_list`?  Thanks.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+                        metric_value = validator.get_metric(metric=metric_config)
+
+                        batch_cardinality_within_limit: bool = (
+                            self._cardinality_checker.cardinality_within_limit(
+                                metric_value=metric_value
+                            )
+                        )
+
+                        if batch_cardinality_within_limit is False:
+                            candidate_column_names.remove(column_name)
```
alexsherstinsky, line 233: @anthonyburdi Maybe a better pattern would be instead of copying the entire list and then filtering out columns build the list out of columns that fit the criteria (using the `Python` `filter()` method)?  That is the pattern I followed in other `DomainBuilder` classes.  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -0,0 +1,247 @@
...
+                            break
```
alexsherstinsky, line 234: @anthonyburdi Why do we `break` out of the loop -- might there be other columns to filter out?..  Also, please see an earlier comment about building columns list, instead of starting the the maximum possible list and then filtering it out.  Thanks!

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
```diff
@@ -0,0 +1,130 @@
...
+    """Get unique values from an iterable of iterables e.g. a list of sets.
+
+    Args:
+        iterable: List, Set containing iterables of values.
+
+    Returns:
+        Single flattened set containing unique values.
+    """
+
+    return set().union(*iterable)
```
alexsherstinsky: @anthonyburdi Using `chain` might be more elegant here -- care to give it a try?  We have examples in the codebase.  Thanks!

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
```diff
@@ -0,0 +1,130 @@
...
+from great_expectations.rule_based_profiler.types.parameter_container import (
+    ParameterNode,
+)
+
+
+class ValueSetMultiBatchParameterBuilder(MetricMultiBatchParameterBuilder):
+    """Build a set of unique values across all specified batches.
+
+    This parameter can be used in building the unique value_set from the
+    filtered domains for use in creating Expectations.
```
alexsherstinsky: @anthonyburdi I believe that the statement `from the filtered domains` is not correct.  All `ParameterBuilder` implementations will execute their computations for each and every `Domain` object.  For this reason, the `batch_request` (or `batch_list`) property of each `ParameterBuilder` has to be considered carefully.  Happy to discuss.  Thanks!

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # Build the list of unique values for each batch
+        super()._build_parameters(
+            parameter_container=parameter_container,
+            domain=domain,
+            variables=variables,
+            parameters=parameters,
+        )
+
+        # Retrieve and replace the list of unique values for each batch with
```
alexsherstinsky, line 102: @anthonyburdi I am thinking of a small refactor focused on making the code segment below dealing with setting the parameter to be an interface that all `ParameterBuilder` classes implement (after a few current PRs get merged).

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
```diff
@@ -0,0 +1,130 @@
...
+        # the set of unique values for all batches in the given domain.
+        fully_qualified_parameter_name: str = f"$parameter.{self.name}"
+        parameter_value_node: ParameterNode = (
+            get_parameter_value_by_fully_qualified_parameter_name(
+                fully_qualified_parameter_name=fully_qualified_parameter_name,
+                domain=domain,
+                parameters={domain.id: parameter_container},
```
alexsherstinsky, line 109: @anthonyburdi What if we just say here `parameters=parameters` -- will it work?  Thanks.

alexsherstinsky: @anthonyburdi Also, can we please try the "whole method" 
```
 build_parameter_container(
            parameter_container=parameter_container, parameter_values=parameter_values
        )
```
using the new values?  It might be a safer way as compared to updating a specific value.  Thanks!

tests/conftest.py
```diff
@@ -6588,239 +6636,6 @@ def bobster_columnar_table_multi_batch_normal_mean_5000_stdev_1000_data_context(
     return context
 
 
-@pytest.fixture
```
alexsherstinsky, line 6591: @anthonyburdi I cannot tell for sure -- is this fixture being deleted, or is it being moved?  Thanks.

alexsherstinsky (changes requested):
@anthonyburdi Lovely work -- I added thoughts/comments in a number of places -- happy to discuss.  Thanks!

great_expectations/rule_based_profiler/domain_builder/__init__.py
cdkini: Do we want absolute imports in our `__init__.py`s? I don't think we have a reason not to but we should probably make the change universally if that's the case!

The `absolufy-imports` package will do this for the full codebase.
```bash
pip install absolufy-imports
find great_expectations -name "*.py" | xargs absolufy-imports    
```

tests/rule_based_profiler/domain_builder/test_categorical_column_domain_builder.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+
+    domain_builder: DomainBuilder = CategoricalColumnDomainBuilder(
+        data_context=data_context,
+        batch_request=batch_request,
+        limit_mode="very_few",
+    )
+    domains: List[Domain] = domain_builder.get_domains()
+
+    alice_all_column_names = [
```
NathanFarmer: Type hint please

tests/rule_based_profiler/domain_builder/test_categorical_column_domain_builder.py
```diff
@@ -0,0 +1,306 @@
...
+        "id",
+        "event_type",
+        "user_id",
+        "event_ts",
+        "server_ts",
+        "device_ts",
+        "user_agent",
+    ]
+    alice_all_column_domains = [
```
NathanFarmer: Type hint please

tests/rule_based_profiler/domain_builder/test_categorical_column_domain_builder.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+
+    domain_builder: DomainBuilder = CategoricalColumnDomainBuilder(
+        data_context=data_context,
+        batch_request=batch_request,
+        limit_mode="ONE",
+    )
+    domains: List[Domain] = domain_builder.get_domains()
+
+    alice_all_column_names = [
```
NathanFarmer: Type hint

tests/rule_based_profiler/domain_builder/test_categorical_column_domain_builder.py
```diff
@@ -0,0 +1,306 @@
...
+        "user_agent",
+    ]
+    alice_all_column_domains = [
```
NathanFarmer: Type hint

tests/rule_based_profiler/domain_builder/test_categorical_column_domain_builder.py
```diff
@@ -0,0 +1,306 @@
...
+            "id",
+            "event_type",
+            "user_id",
+            "event_ts",
+            "server_ts",
+        ],
+    )
+    domains: List[Domain] = domain_builder.get_domains()
+
+    alice_all_column_names = [
```
NathanFarmer: Type hint

tests/rule_based_profiler/domain_builder/test_categorical_column_domain_builder.py
```diff
@@ -0,0 +1,306 @@
...
+        "device_ts",
+        "user_agent",
+    ]
+    alice_all_column_domains = [
```
NathanFarmer: Type hint

tests/rule_based_profiler/domain_builder/test_categorical_column_domain_builder.py
```diff
@@ -0,0 +1,306 @@
...
+
+    domain_builder: DomainBuilder = CategoricalColumnDomainBuilder(
+        data_context=data_context,
+        batch_request=batch_request,
+        limit_mode="VERY_FEW",
+        exclude_columns=[],
+    )
+    domains: List[Domain] = domain_builder.get_domains()
+
+    alice_all_column_names = [
```
NathanFarmer: Type hint

tests/rule_based_profiler/domain_builder/test_categorical_column_domain_builder.py
```diff
@@ -0,0 +1,306 @@
...
+        "id",
+        "event_type",
+        "user_id",
+        "event_ts",
+        "server_ts",
+        "device_ts",
+        "user_agent",
+    ]
+    alice_all_column_domains = [
```
NathanFarmer: Type hint

tests/rule_based_profiler/parameter_builder/test_value_set_multi_batch_parameter_builder.py
```diff
@@ -0,0 +1,171 @@
...
+            "num_batches": 1,
+        },
+    }
+
+    assert (
+        get_parameter_value_by_fully_qualified_parameter_name(
+            fully_qualified_parameter_name=fully_qualified_parameter_name_for_value,
+            domain=domain,
+            parameters={domain.id: parameter_container},
+        )
```
NathanFarmer: `great_expectations.rule_based_profiler.util.get_parameter_value_and_validate_return_type` may be a better option here as it also accepts fully qualified parameter names and validates that a `dict` is returned.

tests/rule_based_profiler/parameter_builder/test_value_set_multi_batch_parameter_builder.py
```diff
@@ -0,0 +1,171 @@
...
+            "num_batches": 3,
+        },
+    }
+
+    assert (
+        get_parameter_value_by_fully_qualified_parameter_name(
+            fully_qualified_parameter_name=fully_qualified_parameter_name_for_value,
+            domain=domain,
+            parameters={domain.id: parameter_container},
+        )
```
NathanFarmer: `great_expectations.rule_based_profiler.util.get_parameter_value_and_validate_return_type`

tests/rule_based_profiler/test_user_workflow_fixtures.py
```diff
@@ -41,5 +41,5 @@ def test_bobby_fixture_generation(
                 "test_configuration_oneshot_sampling_method"
             ]["expected_expectation_suite"].expectations
         )
-        == 37
+        == 39
```
NathanFarmer, line 44: ❤️ I love watching this number increase

NathanFarmer (changes requested):
This was an absolute joy to review! I learned a lot!

great_expectations/rule_based_profiler/domain_builder/__init__.py
anthonyburdi: Thanks Will! I needed to change one to fix an issue and ended up changing them all.
@cdkini That sounds great to me! Absolute is more verbose but more readable and explicit about intent so I would recommend always using them unless there is a clear reason not to for a particular case (and recommended in [pep8](https://www.python.org/dev/peps/pep-0008/) and the [google style guide](https://google.github.io/styleguide/pyguide.html#224-decision)). Let's do that in a separate focused PR and maybe add a check that they don't creep back in unless we explicitly flag them, maybe using something like https://pypi.org/project/flake8-absolute-import/#description. How about we add a cleanup task in the backlog for this?

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
anthonyburdi: I was following the style guide: https://google.github.io/styleguide/pyguide.html#22-imports "Use import statements for packages and modules only, not for individual classes or functions. Classes imported from the [typing module](https://google.github.io/styleguide/pyguide.html#typing-imports), [typing_extensions module](https://github.com/python/typing/tree/master/typing_extensions), and redirects from the [six.moves module](https://six.readthedocs.io/#module-six.moves) are exempt from this rule."

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
anthonyburdi: @alexsherstinsky good call, but looks like the recommendation is to use `float` instead: https://www.python.org/dev/peps/pep-0484/#the-numeric-tower

anthonyburdi: Changed in a local commit, will push after resolving other comments.

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
anthonyburdi: Thank you for the reminder about our util methods @alexsherstinsky ! In this case it's not exactly equivalent, here I'm just checking the type of `metric_value`. In `is_numeric()` it tries to coerce the value and check for errors - so for example a string would be considered numeric e.g. `is_numeric("1") == True` whereas here we just want to ensure an int or float type. If we want to handle numeric strings we would do the conversion outside of the responsibility of the `CardinalityChecker`.

Interestingly, we turned `Union[int, float]` to `float` based on pep484 in another comment, but `isinstance()` behaves differently e.g. `isinstance(1, float) == False`

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
anthonyburdi: Changed in a local commit, will push after resolving other comments.

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
anthonyburdi: Of course 🤦  good catch. Changed in a local commit, will push after resolving other comments.

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
anthonyburdi: Feels like overkill in this particular enum (assuming you mean e.g. `ZERO: AbsoluteCardinalityLimit = AbsoluteCardinalityLimit("ZERO", 0)`). If we want 100% type coverage though I agree then we will have to do it.

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
anthonyburdi: I don't like either upon reflection. Changing to `metric_name_defining_limit` - thoughts?

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: Not currently, checking with Eugene on requirements. Will report back.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: @alexsherstinsky What is the purpose of this for the CategoricalColumnDomainBuilder? I see that it is used in the proposed `_validate_rule_and_update_rule_properties` but the `ValueSetMultiBatchParameterBuilder` builds its own set of parameters from all batches provided (not just the active batch).

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: Not currently, checking with Eugene on requirements. Will report back.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: Could we split that into a Self-initializing expectations PR? I want to understand it better before adding.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: Since it's inherited from the parent class, I think we can keep it private since we do not envision using it elsewhere in the codebase. Python doesn't set limits on inherited private methods like e.g. Java. If we do need to use it elsewhere, I do agree with making it public.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: Confirmed that if a user wants to specify specific columns they would use a ColumnDomainBuilder specifying those specific columns in a different rule.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: In the current PR we do not, but in the future we can further extend CardinalityLimit with additional subclasses that handle appearance and recency thresholds e.g. cardinality limit checked against "num of unique values that appear in x% of batches" and "num of unique values that appear in the latest 20% of batches"

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: Changed in a local commit, will push after resolving other comments. I am now using the validator.active_batch

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: @alexsherstinsky I agree, though perhaps better to tackle separately as a refactor, what do you think?

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: @alexsherstinsky I hear you, but I would like to keep this private unless/until we have a good reason to access it outside of the class. Currently it is created at instantiation time and not modified which is simpler until we have a reason to make it accessible. Please let me know if I'm missing something.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: Good call. I didn't do it initially because it's a large nested block of code but you are right. Changed!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: `metric_config_batch_list` is a list of MetricConfig so I think `metric_config` is named appropriately. `metric_config_batch_list` is so named because it is metric configs for each batch in batch_list. But open to suggestions for improvement.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: I did this purposely to avoid computing metrics once a domain in a particular batch fails, then we stop computing for the rest of the batches.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: We are breaking out of the inner loop, so we stop considering that column since it fails the cardinality test. Then we continue with the other columns in the outer loop. I hope soon we can use `validator.get_metrics()` to compute these all at once. It will depend on the data / data store whether one method is faster than another.

great_expectations/rule_based_profiler/domain_builder/domain_builder.py
anthonyburdi: I understand the logic, but I disagree. The rationale is that this functionality is specific and tightly coupled to domains so I would look for it in domain.py rather than a general `util.py` module. That being said, I do agree it should be moved from it's current home in `domain_builder.py` to `domain.py`.

tests/conftest.py
anthonyburdi: Apologies - meant to add a note in the PR. This fixture was duplicated so the duplicate is being removed.

great_expectations/rule_based_profiler/domain_builder/__init__.py
anthonyburdi: Added GREAT-612

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
anthonyburdi: Thank you!! :heart:

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
anthonyburdi: Agree. Changed @alexsherstinsky! Please let me know if it is more clear and precise now.:
```python
This parameter builder can be used to build a unique value_set for each 
    of the domains specified by the DomainBuilder from all of the batches 
    specified. This value_set can be used to create Expectations.
    
    This unique value_set...
```

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
anthonyburdi: Agree - that makes a lot of sense to me! And this PR is getting a little long in the tooth so a separate PR makes more sense.

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
anthonyburdi: Oooh that is a nice built-in @alexsherstinsky. But here we are only looking for unique values rather than all values so I think a set union is more appropriate. I did not find a corresponding itertools method. Also, I broke this into it's own function though so that if we do find a better more performant implementation it is easy to change.

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
anthonyburdi: `parameters=parameters` does not work because here parameters is None if not supplied. Also I looked at that `build_parameter_container()` method but it seems like it will not overwrite the node but please let me know if I am misunderstanding. Trying it now to double check.

anthonyburdi: I have not been able to get this to work as you've described, perhaps we can discuss.

tests/rule_based_profiler/domain_builder/test_categorical_column_domain_builder.py
anthonyburdi: Thank you!! :heart:

tests/rule_based_profiler/parameter_builder/test_value_set_multi_batch_parameter_builder.py
anthonyburdi: I tried this using the and was unable to get it to work properly. : ```assert get_parameter_value_and_validate_return_type(
        domain=domain,
        parameter_reference=fully_qualified_parameter_name_for_value,
        expected_return_type=dict
    ) == expected_parameter_value```
I am getting the following error in `get_parameter_value_by_fully_qualified_parameter_name` which is called by `get_parameter_value_and_validate_return_type`, any ideas?:
```
>           parameter_container = parameters[domain.id]
E           TypeError: 'NoneType' object is not subscriptable
```

tests/rule_based_profiler/parameter_builder/test_value_set_multi_batch_parameter_builder.py
anthonyburdi: Thank you! I needed it to make sure this method did what I expected, so I left it in the test suite to avoid regressions and help explain what the method does.

tests/rule_based_profiler/test_user_workflow_fixtures.py
anthonyburdi: Me too!

tests/rule_based_profiler/parameter_builder/test_value_set_multi_batch_parameter_builder.py
NathanFarmer: You are missing `parameters={domain.id: parameter_container},` in `get_parameter_value_and_validate_return_type`.

anthonyburdi: 🤦  Fixed!

tests/rule_based_profiler/parameter_builder/test_value_set_multi_batch_parameter_builder.py
anthonyburdi: Fixed! Thank you for the suggestion.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi What if one fails, but next one does not?..

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi So it cannot un-fail for the next batch?

alexsherstinsky: @anthonyburdi Please wait for #4282 to get merged, if you can -- you get parallel execution!  Thanks!

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
alexsherstinsky: @anthonyburdi I was thinking about something like this:
```
                selected_elements = list(
                    itertools.chain.from_iterable(
                        [all_elements]
                    )
                )
```
Just an idea -- up to you!

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
alexsherstinsky: @anthonyburdi Sounds good!  Thanks!

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
alexsherstinsky: @anthonyburdi Could we please make a (very time-boxed) attempt together?  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi Sorry cannot see -- was a type hint added?  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi Then we would need to merge #4282 and #4258 prior to this PR -- which I would be grateful for if you can hang on.  Thanks!

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
alexsherstinsky: @anthonyburdi There is a chance that we can get some help for this code segment by merging #4282 first.  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi It would be wonderful (as a nice-to-have) to document future possibility, such as this, succinctly, in the code.  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi This is important for Self-Initializing Expectations whereby there is no batch_request, just Batch or Batch list.  Happy to discuss.  Thanks.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi Yes -- now quite a few ideas for a code beautification/refactor PR!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi The problem is that once we start using this and try to serialize it, the output may not catch/show it, unless there is an accessor (this is per contract with `Builder` and its parents).

alexsherstinsky (changes requested):
@anthonyburdi Just a few more little things to iron out and/or try (please see comments) -- can be accomplished by a quick paring, and then it will be done!  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi and @alexsherstinsky discussed and agreed to add capabilities needed to enable Self-Initializing expectations that use `ValueSet` in a separate work item.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi and @alexsherstinsky discussed and agreed to add capabilities needed to enable Self-Initializing expectations that use ValueSet in a separate work item.

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
alexsherstinsky: @anthonyburdi and @alexsherstinsky discussed and agreed to add this in a separate work item.

NathanFarmer (approved):
LGTM pending the resolution of @alexsherstinsky's comments

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: Done!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: Done!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: It will continue computing with the next column domain.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
anthonyburdi: @alexsherstinsky correct - it can't un-fail for the next batch

tests/rule_based_profiler/domain_builder/test_cardinality_checker.py
anthonyburdi: Thank you!

alexsherstinsky (approved):
LGTM!

great_expectations/rule_based_profiler/domain_builder/cardinality_checker.py
alexsherstinsky: @anthonyburdi Instead of my preference, I will trust @cdkini 's intuition for what is sufficient to do here.  Thanks!

PR #4321, opened 2022-03-02
[FEATURE] Decorate RuleBasedProfiler.run() with usage statistics

cdkini:
- Write payload fn
- Add decorator to run method
- Update existing tests and write new ones

### Definition of Done


great_expectations/core/usage_statistics/anonymizers/profiler_run_anonymizer.py
```diff
@@ -172,9 +172,18 @@ def _anonymize_domain_builder(self, domain_builder: dict) -> dict:
...
         if batch_request:
+            anonymized_batch_request: Optional[Union[str, dict]]
+            if isinstance(batch_request, str):
+                anonymized_batch_request = self.anonymize(batch_request)
+            else:
+                anonymized_batch_request = (
+                    self._batch_request_anonymizer.anonymize_batch_request(
+                        **batch_request
+                    )
+                )
```
cdkini: Due to not performing variable substitution at this point, we must account for the case where the batch request is a `$variable` string.

great_expectations/core/usage_statistics/schemas.py
```diff
@@ -762,7 +762,12 @@
     "properties": {
         "parent_class": {"type": "string", "maxLength": 256},
         "anonymized_class": {"$ref": "#/definitions/anonymized_string"},
-        "anonymized_batch_request": {"$ref": "#/definitions/anonymized_batch_request"},
+        "anonymized_batch_request": {
+            "oneOf": [
+                {"$ref": "#/definitions/anonymized_string"},
+                {"$ref": "#/definitions/anonymized_batch_request"},
+            ]
+        },
```
cdkini: another round of staging/migration is necessary due to this change.

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -340,10 +340,8 @@ def __init__(
         self.name = name
         self.config_version = config_version
         self.rules = rules
-        if class_name is not None:
-            self.class_name = class_name
-        if module_name is not None:
-            self.module_name = module_name
+        self.class_name = class_name
+        self.module_name = module_name
```
cdkini: This may have been done for a particular reason but I really don't think we should have variable attributes like this. This, in addition to setattr, further detracts from the purpose of classes.

I've opted to add these back in but I'm happy to discuss how to strike a balance.

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -408,29 +406,23 @@ def resolve_config_using_acceptable_arguments(
...
+            variables=variables
         )
 
-        rule: "Rule"  # noqa: F821
+        effective_rules: Dict[
+            str, "Rule"  # noqa: F821
+        ] = profiler.reconcile_profiler_rules_as_dict(rules=rules)
         runtime_rules: Dict[str, dict] = {
-            rule.name: rule.to_json_dict() for rule in effective_rules
+            name: rule.to_dict() for name, rule in effective_rules.items()
```
cdkini: This was all changed in a prior PR but we should be leveraging the `as_dict` method as opposed to the full reconciliation.

great_expectations/types/__init__.py
```diff
@@ -135,7 +135,7 @@ def to_raw_dict(self):
 
     def to_dict(self) -> dict:
         new_dict = {
-            key: self[key]
+            key: self.get(key)
```
cdkini: Without this:
```python
AttributeError: ‘CategoricalColumnDomainBuilder’ object has no attribute ‘cardinality_checker’
```

We need to ensure we actually have the attr before trying to convert to dict - Alice breaks without this change.

tests/conftest.py
```diff
@@ -6428,6 +6429,7 @@ def bobby_columnar_table_multi_batch_deterministic_data_context(
 
     # Re-enable GE_USAGE_STATS
     monkeypatch.delenv("GE_USAGE_STATS")
+    monkeypatch.setattr(AnonymizedUsageStatisticsConfig, "enabled", True)
```
cdkini, line 6439: Maybe I'm missing something but this `delenv` statement did not enable usage stats - the statement added here ensures that we create a handler.

I've added this to all Alice/Bobby/Bobster/Quentin fixtures so we can determine their payloads.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -145,6 +149,11 @@ def __init__(
         if variables is None:
             variables = {}
 
+        if data_context:
+            self._usage_statistics_handler = data_context.usage_statistics_handler
+        else:
+            self._usage_statistics_handler = None
+
```
cdkini: Thoughts on passing the handler itself to the constructor as opposed to relying on the data context property? If we want to eventually get away from the context as sort of a catch-all, we could supply the handler from the context and propagate it downstream.

We've done something similar for the ProfilerStore on the context.

tests/integration/usage_statistics/test_usage_statistics_messages.py
```diff
@@ -2626,6 +2626,48 @@ def test_generate_messages_with_defaults():
             "ge_version": "0.14.7.manual_testing",
             "x-forwarded-for": "00.000.00.000, 00.000.000.000",
         },
+        {
```
cdkini: Additional payload to test the possibility of batch request being a `$variable` string.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -401,9 +414,14 @@ def reconcile_profiler_variables_as_dict(
...
-            self.variables.to_dict()["parameter_nodes"]["variables"]["variables"] or {}
-        )
+        variables_configs: dict
+        try:
+            variables_configs = self.variables.to_dict()["parameter_nodes"][
+                "variables"
+            ]["variables"]
+        except (TypeError, KeyError) as e:
+            variables_configs = {}
+            logger.warning("Could not convert existing variables to dict: %s", e)
```
cdkini, line 423: If we instantiate without any variables, we hit a type error here.

I think it's a little risky to be retrieving values so deeply nested in a dict without strict types so this is a sanity check.

great_expectations/rule_based_profiler/config/base.py
alexsherstinsky: @cdkini It is critical to have those as before (with the "if" statements); otherwise, the "missing" directive in the Marshmallow schema is ineffective, and we will get validation errors.  Happy to discuss.  Thank you!

great_expectations/rule_based_profiler/config/base.py
alexsherstinsky: @cdkini My preference is to always use the highest possible level method (within reason).  Here, by using the higher level method, we can -- in the future -- pass the non-default reconciliation directives.  There is no need to leverage "as_dict" -- it is really a "protected" method (that does not exist in Python, and we did not want to make it private, because it is used outside of the module, where it is defined).  Happy to discuss.  Thank you.

great_expectations/types/__init__.py
alexsherstinsky: @cdkini This could be due to a bad merge -- there is the `cardinality_checker` property now.  Happy to discuss.  Thanks!

alexsherstinsky (changes requested):
@cdkini A few changes requested -- thanks (getting close!).

great_expectations/rule_based_profiler/config/base.py
cdkini: to consider implications of both approach - both have value but we need to see what makes the most sense here.

great_expectations/rule_based_profiler/rule_based_profiler.py
anthonyburdi: I like passing the handler, but we might want to instead create a registry where we can retrieve the same handler instance based on data_context_id of the data_context requesting run of the RBP.

cdkini: We have GREAT-613 (a spike to determine the viability of a registry) in our backlog. I think for now I'll pass the handler and revisit this down the line?

I love the idea but don't want scope to increase if that's okay with you.

anthonyburdi (approved):
LGTM!

great_expectations/core/usage_statistics/usage_statistics.py
```diff
@@ -566,7 +571,55 @@ def get_checkpoint_run_usage_statistics(
...
+
+            resolved_runtime_config: "RuleBasedProfilerConfig" = (  # noqa: F821
+                RuleBasedProfilerConfig.resolve_config_using_acceptable_arguments(
+                    profiler=profiler,
+                    variables=variables,
+                    rules=rules,
+                )
+            )
+
+            payload: dict = profiler_run_anonymizer.anonymize_profiler_run(
```
alexsherstinsky, line 617: @cdkini positional vs. keyword arguments... I prefer keyword ones for easier readability (just a thought)...

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -408,32 +415,58 @@ def resolve_config_using_acceptable_arguments(
     ) -> "RuleBasedProfilerConfig":  # noqa: F821
         runtime_config: RuleBasedProfilerConfig = profiler.config
 
-        runtime_variables: Optional[
-            ParameterContainer
-        ] = profiler.reconcile_profiler_variables(variables)
-        runtime_variables_configs: dict = (
-            runtime_variables.to_dict()["parameter_nodes"]["variables"]["variables"]
-            or {}
+        runtime_variables: dict = profiler.reconcile_profiler_variables_as_dict(
```
alexsherstinsky, line 432: @cdkini Small worry here that this way only default reconciliation directives will be applied.  What if we want to customizer them later (that is why I am in favor of calling highest-level public methods)...

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -408,32 +415,58 @@ def resolve_config_using_acceptable_arguments(
...
+            build_parameter_container_for_variables(variables_configs=runtime_variables)
         )
 
-        rule: "Rule"  # noqa: F821
-        runtime_rules: Dict[str, dict] = {
-            rule.name: rule.to_json_dict() for rule in effective_rules
-        }
+        effective_rules: Dict[
+            str, "Rule"  # noqa: F821
+        ] = profiler.reconcile_profiler_rules_as_dict(rules=rules)
```
alexsherstinsky, line 441: @cdkini Same small worry as above.

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -408,32 +415,58 @@ def resolve_config_using_acceptable_arguments(
...
             module_name=profiler.__class__.__module__,
             name=runtime_config.name,
             config_version=runtime_config.config_version,
-            variables=runtime_variables_configs,
+            variables=runtime_variables,
             rules=runtime_rules,
         )
 
+    @staticmethod
+    def _interpolate_variables_in_config(
```
alexsherstinsky: @cdkini A little docstring for people not on this project might help for how `_traverse_and_interpolate` works and why it is there.  I also feel that the name "interpolate" does not fit (to me "expand" or "substitute" would fit better).  Thanks.

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -408,32 +415,58 @@ def resolve_config_using_acceptable_arguments(
...
+                        node[key] = get_parameter_value_and_validate_return_type(
+                            domain=None,
+                            parameter_reference=val,
+                            variables=variables_container,
+                            parameters=None,
+                        )
+                    _traverse_and_interpolate(val)
+            elif isinstance(node, list):
+                for val in node:
+                    _traverse_and_interpolate(val)
```
alexsherstinsky: @cdkini positional vs. keyword arguments ... what is best?  I like keyword ones...  Sometimes, of course, positional is the only way.  Just a thought.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -120,6 +125,7 @@ def __init__(
         self,
         profiler_config: RuleBasedProfilerConfig,
         data_context: Optional["DataContext"] = None,  # noqa: F821
+        usage_statistics_handler: Optional[UsageStatisticsHandler] = None,
```
alexsherstinsky, line 128: ❤️

great_expectations/rule_based_profiler/config/base.py
cdkini: don't think it's strictly necessary for 1-3 arg funcs but I agree in general

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -145,6 +151,8 @@ def __init__(
         if variables is None:
             variables = {}
 
+        self._usage_statistics_handler = usage_statistics_handler
```
alexsherstinsky, line 154: @cdkini Non-blocking question -- should we be grabbing the main `DataContext` handler (please see `Checkpoint`).

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -401,9 +413,14 @@ def reconcile_profiler_variables_as_dict(
         if variables is None:
             variables = {}
 
-        variables_configs: dict = (
-            self.variables.to_dict()["parameter_nodes"]["variables"]["variables"] or {}
-        )
+        variables_configs: dict
```
alexsherstinsky, line 416: ❤️

great_expectations/rule_based_profiler/config/base.py
cdkini: I'm basing "interpolate" on how it's denoted in JavaScript when using template literals but I'm happy to change it to substitute.

alexsherstinsky (approved):
@cdkini Just requesting a docstring and added some questions (so pre-approving).  Thanks!

great_expectations/rule_based_profiler/rule_based_profiler.py
cdkini: it's being supplied from the context in the other constructor! This is to decouple the two objects so we can eventually remove the context as a dep

great_expectations/core/usage_statistics/usage_statistics.py
cdkini: will change in the next commit!

great_expectations/rule_based_profiler/config/base.py
cdkini: I agree but I think our tests provide sufficient coverage to ensure we catch any breaking changes here.

I think the pros outweigh the cons but open to possibly re-evaluate down the line.

PR #4327, opened 2022-03-03
[FEATURE] Use more granular requirements-dev-xxx.txt files

kenwade4:
- Use more granular requirements-dev-xxx.txt files

### Definition of Done


setup.py
```diff
@@ -6,6 +6,57 @@
...
+req_sqlalchemy = [req for req in dev_lite if req.startswith("sqlalchemy")]
+
+with open("requirements-dev-airflow.txt") as f:
+    req_airflow = f.read().splitlines()
+
+with open("requirements-dev-arrow.txt") as f:
+    req_arrow = f.read().splitlines()
+
+with open("requirements-dev-athena.txt") as f:
+    req_athena = f.read().splitlines()
```
cdkini: Could this logic be encapsulated in a for-loop or function of sorts?

tests/test_packaging.py
```diff
@@ -16,8 +16,18 @@ def test_requirements_files():
...
-        requirements_dev_base = {
+    with open(file_relative_path(__file__, "../requirements-dev-lite.txt")) as req:
+        requirements_dev_lite = {
+            f'{line.name}{"".join(line.specs[0])}' for line in rp.parse(req)
+        }
+
+    with open(file_relative_path(__file__, "../requirements-dev-contrib.txt")) as req:
+        requirements_dev_contrib = {
+            f'{line.name}{"".join(line.specs[0])}' for line in rp.parse(req)
+        }
```
cdkini: Same note here - the granularity we're adding adds a lot of verbosity and I'd love to simply things into a loop/function/dict if possible.

requirements.txt
```diff
@@ -1,25 +1,20 @@
...
+ipywidgets>=7.5.1
+jinja2>=2.10
+jsonpatch>=1.22
+jsonschema>=2.5.1
+mistune>=0.8.4,<2.0.0
+numpy>=1.14.1
+pandas>=0.23.0
+pyparsing>=2.4,<3
+python-dateutil>=2.8.1
+pytz>=2015.6  # <---- should we update to something in the recent few years?
```
cdkini: Can we resolve this before merging?

requirements-dev-lite.txt
```diff
@@ -1,17 +1,12 @@
-
---requirement requirements.txt
-
-black==22.1.0  # lint
-boto3>=1.9  # all_tests
-flake8==3.8.3  # lint
+boto3>=1.9 # <------ should this be 1.14 (see extras_require["s3"] in setup.py)
+           # <------ should this be 1.17.106 (see constraints-dev.txt)
```
cdkini: Please resolve these comments before merging.

.github/workflows/test-sqlalchemy-latest.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies including latest sqlalchemy version
         run: |
-          pip install --requirement requirements-dev-base.txt --requirement requirements-dev-sqlalchemy.txt -c constraints-dev.txt
+          pip install --requirement requirements-dev-test.txt --requirement requirements-dev-sqlalchemy.txt -c constraints-dev.txt
```
cdkini, line 40: Please do a grep and make sure we've covered all instances. I think there are a few in the `azure` directory that was added last week.

docs/contributing/contributing_setup.md
```diff
@@ -58,7 +58,7 @@ source ge_dev/bin/activate
 
 pip install --upgrade pip==21.3.1
 
-pip install -r requirements-dev-lite.txt -c constraints-dev.txt -e .
+pip install -r requirements.txt -r requirements-dev-lite.txt -c constraints-dev.txt -e .
```
cdkini, line 61: Is this the only instance we need to change in the docs?

cdkini (changes requested):
Requesting a few changes but nothing too significant.

requirements.txt
kenwade4: Comment was added so it could be discussed in PR

requirements-dev-lite.txt
kenwade4: Comment was added so it could be discussed in PR

docs/contributing/contributing_setup.md
kenwade4: Yea, when I first added that "easy" section to the contributing doc, the dev-lite file initially had a `--requirement requirements.txt` at the top. This is the only place in docs where the `requirements-dev-lite.txt` file is referenced

.github/workflows/test-sqlalchemy-latest.yml
kenwade4: There are no more references to requirements-dev-base.txt any where

tests/test_packaging.py
kenwade4: Not changing this file

requirements-dev-airflow.txt
```diff
@@ -0,0 +1 @@
+apache-airflow[s3]>=1.9.0
```
kenwade4: I think we should get rid of the airflow file... I only added it because there was an extras_require entry in setup.py, but it's not documented and nobody would miss it.

tests/test_packaging.py
cdkini: Any reason why? I'm okay leaving it be but this seems like a good candidate for a refactor.

requirements-dev-airflow.txt
cdkini: Sounds good to me but maybe reach out to @talagluck first just to confirm!

cdkini (approved):

setup.py
```diff
@@ -1,7 +1,40 @@
...
+        key = rx.match(fname).group(1)
+        if key in ignore_keys:
+            continue
+        with open(fname) as f:
+            results[key] = f.read().splitlines()
+
+    lite = results.pop("lite")
+    results["boto"] = [req for req in lite if req.startswith("boto")]
+    results["sqlalchemy"] = [req for req in lite if req.startswith("sqlalchemy")]
+    results["airflow"] += results["boto"]
```
kenwade4: This line would be deleted too

tests/test_packaging.py
kenwade4: I ended up refactoring it...

PR #4329, opened 2022-03-03
[BUGFIX] Only validate row_condition if not None

roblim:
- `row_condition` validation should only occur if not `None`, otherwise exception is thrown and serialization fails (`TypeError: argument of type 'NoneType' is not iterable `)


alexsherstinsky (approved):
LGTM

PR #4340, opened 2022-03-05
[FEATURE] ParameterBuilder for Computing Average Unexpected Values Fractions for any Map Metric

alexsherstinsky:
### Scope
The new `MeanUnexpectedMapMetricMultiBatchParameterBuilder` accepts any map metric as well as the required name of the `ParameterBuilder` that computes the total count metric and optionally the name of the `ParameterBuilder` that computes the number of rows with the empty value for the domain (e.g., an empty column).  It then computes the unexpected counts of the map metric for each batch (e.g., `"column_values.nonnull"`, `""column_values.unique"`, etc.) and then the ratios of the unexpected counts to the counts of non-null records, and finally outputs the mean of these ratios.

This result can then be used by an ExpectationConfigurationBuilder to determine whether or not this mean ratio is sufficiently low for an expectation to be emitted, which is beneficial in avoiding "validation error noise" subsequently.

### Note
`MeanUnexpectedMapMetricMultiBatchParameterBuilder` requires the `ParameterBuilder` configurations for the total count and (optionally) the null count metric to appear earlier, since their results are pre-requisite for this parameter builder.

- JIRA: GREAT-464/GREAT-498/GREAT-636
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


Shinnnyshinshin (commented):
Thank you so much for this work @alexsherstinsky . the code + tests look great, but given the importance of this PR in the direction we are "nudging' it to go (ie more like a formal standard machine learning component), I think a synchronous discussion would really help. Would we be able to schedule some time on Monday?

alexsherstinsky:
> Thank you so much for this work @alexsherstinsky . the code + tests look great, but given the importance of this PR in the direction we are "nudging' it to go (ie more like a formal standard machine learning component), I think a synchronous discussion would really help. Would we be able to schedule some time on Monday?

@Shinnnyshinshin Happy to offer my half-baked ideas.  Although this PR in itself is only a hint at what a proper data pipeline should be.  In particular, this is the first time we have a `ParameterBuilder`, which requires other parameters to be executed first, because it uses the values that they compute.  As such, this is the first clear illustration of the capabilities of the current architecture, where `ParameterContainter` with its `ParameterNode` constituents acts as the holder of the State of a Rule during its execution lifecycle.  This was envisioned almost a year ago, implemented last summer, and finally utilized.  Thanks!

great_expectations/rule_based_profiler/parameter_builder/mean_unexpected_metric_multi_batch_parameter_builder.py
```diff
@@ -0,0 +1,191 @@
...
+                parameter_reference=fully_qualified_null_count_parameter_builder_name,
+                expected_return_type=None,
+                variables=variables,
+                parameters=parameters,
+            )
+            null_count_values = null_count_parameter_node.value
+
+        nonnull_count_values: np.ndarray = total_count_values - null_count_values
+
+        # Compute "unexpected_count" corresponding to "map_metric_name" (given as argument to this "ParameterBuilder").
```
anthonyburdi, line 164: Thank you for these helpful comments.

anthonyburdi (approved):
LGTM!

great_expectations/rule_based_profiler/parameter_builder/mean_unexpected_metric_multi_batch_parameter_builder.py
```diff
@@ -0,0 +1,191 @@
...
+        self,
+        name: str,
+        map_metric_name: str,
+        total_count_parameter_builder_name: str,
+        null_count_parameter_builder_name: Optional[str] = None,
+        metric_domain_kwargs: Optional[Union[str, dict]] = None,
+        metric_value_kwargs: Optional[Union[str, dict]] = None,
+        batch_list: Optional[List[Batch]] = None,
+        batch_request: Optional[Union[BatchRequest, RuntimeBatchRequest, dict]] = None,
+        json_serialize: Union[str, bool] = True,
```
cdkini, line 39: Why can this be a string as well?

great_expectations/rule_based_profiler/parameter_builder/parameter_builder.py
```diff
@@ -137,9 +141,20 @@ def build_parameters(
...
+            expected_return_type=bool,
+            variables=variables,
+            parameters=parameters,
+        )
+
         parameter_values: Dict[str, Any] = {
             self.fully_qualified_parameter_name: {
-                "value": convert_to_json_serializable(data=computed_parameter_value),
+                "value": convert_to_json_serializable(data=computed_parameter_value)
+                if json_serialize
```
cdkini, line 156: Can this be a string? If so, it'll always result in True. Are we okay with that?

great_expectations/rule_based_profiler/parameter_builder/mean_unexpected_metric_multi_batch_parameter_builder.py
alexsherstinsky: @cdkini Because of the `$parameter` syntax, potentially used.

great_expectations/rule_based_profiler/parameter_builder/parameter_builder.py
```diff
@@ -157,6 +172,14 @@ def fully_qualified_parameter_name(self) -> str:
     def name(self) -> str:
         return self._name
 
+    @property
+    def json_serialize(self) -> bool:
+        return self._json_serialize
```
cdkini, line 177: The input type is `Union[bool ,str]`. If we pass in a string, won't this be erroneous?

alexsherstinsky: @cdkini The `$parameter` syntax requires this Union for now -- hope we can simplify later.

tests/rule_based_profiler/parameter_builder/test_mean_unexpected_metric_multi_batch_parameter_builder.py
```diff
@@ -0,0 +1,279 @@
...
+    )
+
+    _: MeanUnexpectedMapMetricMultiBatchParameterBuilder = (
+        MeanUnexpectedMapMetricMultiBatchParameterBuilder(
+            name="my_name",
+            map_metric_name="column_values.nonnull",
+            total_count_parameter_builder_name="my_total_count",
+            data_context=data_context,
+        )
+    )
```
cdkini, line 35: Nitpick - we don't even need assignment here right? Perfectly fine to keep this but we should probably have a consistent approach since I tend to omit assignment when I'm just testing if the behavior errors/logs/etc.

great_expectations/rule_based_profiler/parameter_builder/mean_unexpected_metric_multi_batch_parameter_builder.py
cdkini: So it's the responsibility of the parameter builder to determine the true value here? That value can't be pre-computed so we can narrow the type to bool?

great_expectations/rule_based_profiler/parameter_builder/parameter_builder.py
cdkini: But what happens if I pass in a string value from the constructor? Per the `__init__`, it seem as though we're assigning the raw value to the attribute without any additional logic.

Doesn't this create a situation where we could possibly be returning a string from this bool property?

PR #4352, opened 2022-03-08
[BUGFIX] BatchRequest serialization (CLOUD-743)

roblim:
- Fixes issue where BatchRequest serialization for usage statistics resulted in serialization of entire Pandas or Spark DF, with commensurate increase in execution time


great_expectations/core/batch.py
```diff
@@ -261,7 +271,18 @@ def to_json_dict(self) -> dict:
         make this refactoring infeasible at the present time.
         """
         dict_obj: dict = self.to_dict()
+        batch_data = dict_obj.get("runtime_parameters", {}).get("batch_data")
+
+        if isinstance(batch_data, pd.DataFrame):
+            dict_obj["runtime_parameters"]["batch_data"] = "PandasDataFrame"
+
+        if pyspark and isinstance(batch_data, pyspark.sql.DataFrame):
```
rreinoldsc: It would be helpful to have a note that pyspark may not have been imported, and we're checking that here

great_expectations/core/batch.py
```diff
@@ -13,6 +15,14 @@
 
 logger = logging.getLogger(__name__)
 
+try:
+    import pyspark
+except ImportError:
+    pyspark = None
+    logger.debug(
```
rreinoldsc, line 20: I'd have suspected this would be a 'warn' level, but I trust you call on this. We don't want unnecessary noisy stdouts

great_expectations/core/batch.py
roblim: Sure thing.

great_expectations/core/batch.py
roblim: I was just following pattern in rest of GE.

rreinoldsc (approved):

great_expectations/core/batch.py
```diff
@@ -261,7 +271,22 @@ def to_json_dict(self) -> dict:
         make this refactoring infeasible at the present time.
         """
         dict_obj: dict = self.to_dict()
+
+        # if a Pandas or Spark DF appears in BatchRequest, temporarily replace it with
+        # str placeholder before calling convert_to_json_serializable so that
+        # DF is not serialized
+        batch_data = dict_obj.get("runtime_parameters", {}).get("batch_data")
```
alexsherstinsky: @roblim The idea is good, but we have a better way!  In `batch.py`, there is `batch_request_contains_batch_data()`, which you can use (please use the keyword-style argument to it, thanks!).  Also, please provide a type hint for `batch_data` (e.g., `Union[BatchRequest, RuntimeBatchRequest, dict]` -- thanks.

alexsherstinsky (changes requested):
Small change request.  Thanks!

great_expectations/core/batch.py
alexsherstinsky: @roblim In addition, if all you care about is to prevent serializing a dataframe, why not just say:
```
if batch_request_contains_batch_data(batch_request=batch_request):
   batch_data: Union[BatchRequest, RuntimeBatchRequest, dict] = batch_request["runtime_parameters"]["batch_data"] = str(type(batch_request["runtime_parameters"]["batch_data"]))
   serializeable_dict: dict = convert_to_json_serializable(data=get_batch_request_as_dict(batch_request=batch_request))
   batch_request["runtime_parameters"]["batch_data"] = batch_data

return serializeable_dict
```
the point being is that you should not even have to run `self.to_dict()` -- please double-check me on that by getting the code snippet above to work for your purposes.  Thanks!

alexsherstinsky (changes requested):
I made a couple of suggestions.  Thanks, @roblim !

anthonyburdi (approved):
LGTM pending resolution of Alex's comments.

great_expectations/core/batch.py
roblim: Thanks for the excellent feedback @alexsherstinsky! Comment addressed in [1448454](https://github.com/great-expectations/great_expectations/pull/4352/commits/14484543d78d95a7a195674055c2365e3baec203).

great_expectations/checkpoint/checkpoint.py
```diff
@@ -717,7 +717,11 @@ def instantiate_from_config_with_runtime_args(
         **runtime_kwargs,
     ) -> "Checkpoint":
         config: dict = checkpoint_config.to_json_dict()
-        config.update(runtime_kwargs)
+
+        for k, v in runtime_kwargs.items():
```
alexsherstinsky: @roblim Sorry, one minor thing -- please -- for style -- key: str and value: Any -- if not a problem (we now do type hints consistently -- at least trying to).  Thanks!

alexsherstinsky (approved):
LGTM (as long as the requested type hints are added prior to merge).  Thank you!

great_expectations/checkpoint/checkpoint.py
roblim: Added in https://github.com/great-expectations/great_expectations/pull/4352/commits/074633aac5191357a6b4b17acdfb8c3762641d84

PR #4358, opened 2022-03-08
[FEATURE] `add_rule()` method for RuleBasedProfilers and tests

Shinnnyshinshin:
- PR 1 of 2 for enabling non-YAML based `RuleBasedProfilers` workflow
- `add_rule()` method to support adding a Rule object to an existing Profiler object. 
- Integration tests that live in `tests/integration/profiling/rule_based_profilers/test_profiler_basic_workflows.py`
- Unittests that live in `tests/rule_based_profiler/test_rule_based_profiler.py`

Previous Design Review notes:
- GREAT-500
-
-

### Definition of Done

Thank you for submitting!


great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -154,12 +154,20 @@ def __init__(
         self._usage_statistics_handler = usage_statistics_handler
 
         # Necessary to annotate ExpectationSuite during `run()`
-        self._citation = {
-            "name": name,
-            "config_version": config_version,
-            "variables": variables,
-            "rules": rules,
-        }
+        if isinstance(rules, dict) or rules is None:
```
Shinnnyshinshin, line 157: check None first.. then check `isinstance()`

Shinnnyshinshin: do the conversion it would need to.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -829,13 +837,15 @@ def rules(self, value: List[Rule]):
 
     def to_json_dict(self) -> dict:
         rule: Rule
+
+        variables_dict: dict = self.variables.to_dict()
```
Shinnnyshinshin, line 841: place in temp variable.. and check None first.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -829,13 +837,15 @@ def rules(self, value: List[Rule]):
...
+        if variables_dict["parameter_nodes"] is not None:
```
Shinnnyshinshin, line 842: If this is a paramterContainer then it would exist, but if not..we would have to check above.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -1019,6 +1048,16 @@ def run_profiler_on_data(
...
         return result
 
+    def add_rule(self, rule: Rule) -> None:
+        """
+        Add Rule object to existing profiler object by appending to self._rules. In cases where a Rule with the same name exists, it is updated.
+        """
+        for index in range(len(self._rules)):
+            existing_rule: Rule = self.rules[index]
+            if existing_rule.name == rule.name:
+                self._rules.pop(index)
```
Shinnnyshinshin, line 1058: we will have to delete keys... add this

Shinnnyshinshin: does the order matter?

Shinnnyshinshin: Or maybe just reference directly.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -153,12 +153,17 @@ def __init__(
 
         self._usage_statistics_handler = usage_statistics_handler
 
-        # Necessary to annotate ExpectationSuite during `run()`
+        # Necessary to annotation ExpectationSuite during `run()`
```
alexsherstinsky: @Shinnnyshinshin Previous wording seemed better (unless you meant to "add annotation").  Thanks.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -153,12 +153,17 @@ def __init__(
...
+        rule_to_load_into_citation: dict = {}
```
alexsherstinsky: @Shinnnyshinshin The initialization should not be needed, because `rule_to_load_into_citation` is set for every condition.  Thanks.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -153,12 +153,17 @@ def __init__(
...
+        if isinstance(rules, Rule):
+            rule_to_load_into_citation = rules.to_json_dict()
```
alexsherstinsky: @Shinnnyshinshin This is not only applying the singular `Rule` type to the plural `rules` list, but is also not needed, because `rules: Optional[Dict[str, Dict[str, Any]]]` (i.e., it is already a dictionary).  Thanks.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -153,12 +153,17 @@ def __init__(
...
+        elif isinstance(rules, dict):
+            rule_to_load_into_citation = rules
         self._citation = {
             "name": name,
             "config_version": config_version,
             "variables": variables,
-            "rules": rules,
+            "rules": rule_to_load_into_citation,
```
alexsherstinsky: @Shinnnyshinshin Based on the above comments, it seems that this can be left as it was before.  Happy to discuss. 
 Thanks.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -563,6 +574,48 @@ def _reconcile_rule_config(
 
         return effective_rule_config
 
+    @staticmethod
```
alexsherstinsky: @Shinnnyshinshin I am not seeing why this is needed.  Is there a problem with the existing `domain_builder_config: dict = rule_config.get("domain_builder", {})`?  Thanks.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -521,8 +526,9 @@ def _reconcile_rule_config(
         effective_rule_config: Dict[str, Any]
         if rule_name in existing_rules:
             rule: Rule = existing_rules[rule_name]
-
-            domain_builder_config: dict = rule_config.get("domain_builder", {})
+            domain_builder_config: dict = RuleBasedProfiler._get_domain_builder_configs(
```
alexsherstinsky: @Shinnnyshinshin Just by looking at the declared types of `variables` and `rules`, this is not needed.  However, if the declared types are no longer current with the intended usage, then they should be changed.  Thanks.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -829,13 +882,18 @@ def rules(self, value: List[Rule]):
 
     def to_json_dict(self) -> dict:
         rule: Rule
+        variables_dict: dict = {}
```
alexsherstinsky: @Shinnnyshinshin Not sure why this is needed.  Thanks.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -961,6 +1019,34 @@ def __init__(
             usage_statistics_handler=usage_statistics_handler,
         )
 
+    def save_profiler(
```
alexsherstinsky: @Shinnnyshinshin I like this approach (creates a new configuration after reconciling it).  I do, however, feel a little uneasy about not using the top-lever reconciliation method calls.  If that can be done, that would be much easier to maintain long term.  This is important, hence I would be happy to discuss (because it requires a tiny bit more work).  Thanks.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -1019,6 +1105,23 @@ def run_profiler_on_data(
         )
         return result
 
+    def add_rule(self, rule: Rule) -> None:
+        """
+        Add Rule object to existing profiler object by appending to self._rules. In cases where a Rule with the same name exists, it is updated.
+        """
+        if not isinstance(rule, Rule):
```
alexsherstinsky: @Shinnnyshinshin If `rule` is already declared as `Rule` type, then this check should not be here.  If, however, the type were a `Union`, that would be a different case.  Thanks.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -1019,6 +1105,23 @@ def run_profiler_on_data(
...
+            raise TypeError("add_rule method requires a Rule to be passed in.")
+
+        index_to_delete: Optional[int] = None
+        for index in range(len(self._rules)):
+            existing_rule: Rule = self.rules[index]
+            if existing_rule.name == rule.name:
+                index_to_delete = index
```
alexsherstinsky: @Shinnnyshinshin This should not be needed -- there is a reconciliation strategy that would accomplish the same -- and using a high-level methods.  Might you be able to look into them?  Happy to help, if needed.  Thanks.

alexsherstinsky (changes requested):
@Shinnnyshinshin I made a few notes -- happy to go over them live.  Thanks!

great_expectations/rule_based_profiler/rule_based_profiler.py
cdkini: I agree I think we actually will have just dicts here because it is coming from the config object (thus removing the need for the conditional conversion).

great_expectations/rule_based_profiler/rule_based_profiler.py
cdkini: I like adding some safety around the `self.variables.to_dict()["parameters_nodes"]["variables"]["variables"]` retrieval.

There are situations where variables are not set and we don't actually have those keys right?

I don't think we need the `isinstance` here though because the property annotation is `Optional[ParameterContrainer]`.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -829,13 +882,18 @@ def rules(self, value: List[Rule]):
 
     def to_json_dict(self) -> dict:
         rule: Rule
+        variables_dict: dict = {}
+        if self.variables and isinstance(self.variables, ParameterContainer):
+            variables_dict = self.variables.to_dict()
+            if variables_dict.get("parameter_nodes"):
+                variables_dict = variables_dict["parameter_nodes"]["variables"][
+                    "variables"
+                ]
```
cdkini: Nitpick - we could make a slight change to reuse the result of the `get` here:

```python
parameter_nodes: Optional[dict] = variables_dict.get("parameter_nodes")
if parameter_nodes:
    variables_dict = parameter_nodes["variables"]["variables"]
```

great_expectations/rule_based_profiler/rule_based_profiler.py
cdkini: @alexsherstinsky you and I had this same discussion around the usage stats implementation.

I think using the top level reconciliation and then having to go back feels particularly wasteful since we're just backtracking a step or two to get the result in our desired format. I see the value in what you're saying though.

There are tests around the class method we're invoking here so I think we have the coverage we need to get by. Let's discuss this synchronously if that would be helpful.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -961,6 +1019,34 @@ def __init__(
...
+        self._profiler_config = updated_profiler_config
+
+        key: Union[GeCloudIdentifier, ConfigurationIdentifier]
+        if ge_cloud_id:
+            key = GeCloudIdentifier(resource_type="contract", ge_cloud_id=ge_cloud_id)
+        else:
+            key = ConfigurationIdentifier(
+                configuration_key=updated_profiler_config.name,
+            )
+        profiler_store.set(key=key, value=updated_profiler_config)
```
cdkini: This logic is encapsulated in the CheckpointStore's `_determine_key` method. Would you mind promoting that to public and using that here?

great_expectations/rule_based_profiler/rule_based_profiler.py
cdkini: Definitely agree - while our type annotations are just hints, they should act as the contract for what we expect; deviations from that will result in erroneous behavior.

I think we're okay with completely omitting this check. Do we want to give users the ability to pass in dicts? I'm leaning towards no because the method says "rule".

great_expectations/rule_based_profiler/rule_based_profiler.py
cdkini: @alexsherstinsky do we really need true "reconciliation" here? The logic there is far more complex than what we're actually trying to do here.

We're not doing nested updates or anything super complicated; rather, we're just deleting the rule if the name already exists. While I see the value in using reconciliation, I think it's too heavyweight for what is a trivial usecase.

Nitpick - @Shinnnyshinshin while this implementation definitely gets the job done, I think we could probably leverage `enumerate` or `filter` to tighten things up a bit.
```python
for idx, existing_rule in enumerate(self._rules):
    if existing_rule.name == rule.name:
        self._rules[idx] = rule
        return

self._rules.append(rule)
```
```python

self._rules = list(filter(lambda r: r.name != rule.name), self._rules)
self._rules.append(rule)
```

great_expectations/data_context/data_context.py
```diff
@@ -3317,6 +3317,15 @@ def add_profiler(
...
 
+    def save_profiler(
+        self,
+        profiler: RuleBasedProfiler,
+        ge_cloud_id: Optional[str] = None,
+    ):
+        return profiler.save_profiler(
+            profiler_store=self.profiler_store, ge_cloud_id=ge_cloud_id
+        )
+
```
cdkini: Do we want to add usage stats to this method? @anthonyburdi @NathanFarmer

great_expectations/rule_based_profiler/rule_based_profiler.py
cdkini: Actually since we're dealing with Profilers, perhaps let's move that helper to the parent class `ConfigurationStore`?

great_expectations/data_context/data_context.py
```diff
@@ -3317,6 +3317,15 @@ def add_profiler(
...
         )
 
+    def save_profiler(
+        self,
+        profiler: RuleBasedProfiler,
+        ge_cloud_id: Optional[str] = None,
+    ):
+        return profiler.save_profiler(
+            profiler_store=self.profiler_store, ge_cloud_id=ge_cloud_id
+        )
```
cdkini: I may be wrong here but I don't think the profiler should necessarily have access to the store here. I'd be inclined to split this into two methods:
1. One to update the profiler's config attr - this should be an instance method on the RBP.
2. Another to save the config to disk - this should be an instance method on the store.

```python
profiler.save_profiler()
self.profiler_store.add_profiler(profiler)
```

If I run `save_profiler` on my profiler instance, am I expected to have access to a store at that point? I think the updating of the config makes sense at that level but if I want to persist my changes and have them written to disk, I think the flow through the data context makes more sense.

See `add_checkpoint` on `CheckpointStore` for an analogous implementation.

tests/integration/profiling/rule_based_profilers/test_profiler_basic_workflows.py
```diff
@@ -0,0 +1,305 @@
+from unittest import mock
```
cdkini, line 1: Would you mind adding unit tests for the following?
* `add_rule`
  - Situation where we delete an existing rule and append a new one
  - Situation where simply append a new one
* `save_profiler`
  - I would probably mock the store and make sure the appropriate `set` call is being made with the update config.

I think you've got great coverage here through really robust integration tests but just trying to add some security around the individual methods. Thanks!

cdkini (changes requested):
Really solid work! Requesting some changes but I think the core logic is sound and makes a lot of sense. There are a few pieces that might warrant a discussion between you, me, and Alex.

Really appreciated the concerted effort to make the user experience a bit cleaner. Let me know if you have any questions around my comments; I think we should be close to merging since most of my items are small.

great_expectations/data_context/data_context.py
anthonyburdi: Good question. I think since we already capture running a profiler with anonymized info about the configuration then that should be sufficient unless we really need to dive into save workflows e.g. to identify user issues. I can't @ Tanner here so will connect with him in slack for his opinion.

anthonyburdi: @Super-Tanner

tannerbeam: my general philosophy is that if the marginal cost of implementing collection on new data points is low and there is clear business need for it, then go ahead. however, that does not seem to be the case at the moment: i get the impression it's a lot more work than we'd like it to be. i can't really speak to the business need. do you need to be able to measure if/when/how many contexts are using this method? i've worked with product managers who like to know how quickly new features are being adopted, for example, so they want to implement collection on literally every new event.

great_expectations/data_context/data_context.py
Shinnnyshinshin: This makes a TON of sense. Thank you for this feedback @cdkini

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -156,26 +156,11 @@ def get_checkpoint(
...
-            ge_cloud_id
-        ), "Must provide either name or ge_cloud_id."
-
-        key: Union[GeCloudIdentifier, ConfigurationIdentifier]
-        if ge_cloud_id:
-            key = GeCloudIdentifier(resource_type="contract", ge_cloud_id=ge_cloud_id)
-        else:
-            key = ConfigurationIdentifier(configuration_key=name)
-
-        return key
```
Shinnnyshinshin, line 181: Raised this up to `configuration_store.py` and made public. Now a public method.

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: the `add_rule()` method and reconciliation made this part unnecessary. Thanks @alexsherstinsky

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: agreed @cdkini. This was here because we do an extra check for whether `self.variables` is None (in cases where a `Rule` has no variables)

Shinnnyshinshin: Removed the `isinstance()`

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: This was just typo on my part :)

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: This was negated by the fact that `add_rule()` now does the reconciliation step and loads the proper dict-formatted rule into the `profiler_config`. There is no need to check for the type :)

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: Also unnecessary after the changes to `add_rule()`

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: removed :)

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: Removed as well :)

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: So this was also negated by a modification to the `add_rule()` method. Basically the`add_rule()` method does all th conversion back and forth between `dict` and `Rule` and ensures that both the profiler's `.rule` field and `._profiler_config.rules` are updated. That means the `save_profiler()` method in `DataContext` turns into 

```
key: Union[
            GeCloudIdentifier, ConfigurationIdentifier
        ] = self.profiler_store.determine_key(name=name, ge_cloud_id=ge_cloud_id)
self.profiler_store.set(key=key, value=profiler.config)
```
These two lines

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: Done :) Thank you

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: Thank you again for this. The update to `add_rule()` has resulted in a true reconciliation of the Rule config, and update to both the `.rule` field  and `._profiler_config`

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: I really like it. @cdkini ... added

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: Done! thank you for this suggestion. I wouldn't have known about the `determine_key()` method otherwise.

tests/integration/profiling/rule_based_profilers/test_profiler_basic_workflows.py
Shinnnyshinshin: Thank you for this suggestion 

* `add_rule()`
   *  There is a test under `tests/rule_based_profiler/test_rule_based_profiler.py`, that adds 3 rules sequentially and examines how many rules are in `profiler.rules`. The first 2 are identical, so the number of rules is 1, and the 3rd one gets added so the final number is 2. 
   * I tried thinking about `remove_rule()`, but it goes involves some reconciliation logic that maybe be out of scope for this PR. would that be ok? 
*  `save_profiler()` 
    *  This test was also added under `tests/rule_based_profiler/test_rule_based_profiler.py`

Shinnnyshinshin:
@cdkini  and @alexsherstinsky  this PR has been updated. The most significant change was around `add_rule()` which now handles the recociliation, conversation from `Dict` to `Rule` and ensuring the update of `self.rules` property and `self._profiler_config.rules` field of the RBP. 

This has cleaned up DataContext logic **significantly** and gotten rid of a lot of the typechecks that were previously needed. Thank you so much @alexsherstinsky  for the pairing time today on this.

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -104,9 +104,9 @@ def delete_checkpoint(
         name: Optional[str] = None,
         ge_cloud_id: Optional[str] = None,
     ) -> None:
-        key: Union[GeCloudIdentifier, ConfigurationIdentifier] = self._determine_key(
-            name=name, ge_cloud_id=ge_cloud_id
-        )
+        key: Union[
+            GeCloudIdentifier, ConfigurationIdentifier
+        ] = CheckpointStore.determine_key(name=name, ge_cloud_id=ge_cloud_id)
```
alexsherstinsky: @Shinnnyshinshin How come it is not `self.determine_key()`?..

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -117,9 +117,9 @@ def delete_checkpoint(
     def get_checkpoint(
         self, name: Optional[str], ge_cloud_id: Optional[str]
     ) -> CheckpointConfig:
-        key: Union[GeCloudIdentifier, ConfigurationIdentifier] = self._determine_key(
-            name=name, ge_cloud_id=ge_cloud_id
-        )
+        key: Union[
+            GeCloudIdentifier, ConfigurationIdentifier
+        ] = CheckpointStore.determine_key(name=name, ge_cloud_id=ge_cloud_id)
```
alexsherstinsky: @Shinnnyshinshin How come it is not self.determine_key()?..

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -156,26 +156,11 @@ def get_checkpoint(
     def add_checkpoint(
         self, checkpoint: "Checkpoint", name: Optional[str], ge_cloud_id: Optional[str]
     ) -> None:
-        key: Union[GeCloudIdentifier, ConfigurationIdentifier] = self._determine_key(
-            name=name, ge_cloud_id=ge_cloud_id
-        )
+        key: Union[
+            GeCloudIdentifier, ConfigurationIdentifier
+        ] = CheckpointStore.determine_key(name=name, ge_cloud_id=ge_cloud_id)
```
alexsherstinsky: @Shinnnyshinshin How come it is not self.determine_key()?..

great_expectations/rule_based_profiler/rule/rule.py
```diff
@@ -64,13 +64,14 @@ def generate(
...
             parameter_builder: ParameterBuilder
-            for parameter_builder in self._parameter_builders:
-                parameter_builder.build_parameters(
-                    parameter_container=parameter_container,
-                    domain=domain,
-                    variables=variables,
-                    parameters=self._parameters,
-                )
+            if self._parameter_builders:
+                for parameter_builder in self._parameter_builders:
```
alexsherstinsky: @Shinnnyshinshin We are changing to property: `self.parameter_builders` and similar for `ExpectationConfigurationBuilder` list.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -152,8 +152,6 @@ def __init__(
             variables = {}
 
         self._usage_statistics_handler = usage_statistics_handler
-
-        # Necessary to annotate ExpectationSuite during `run()`
```
alexsherstinsky: @Shinnnyshinshin How come this comment is being deleted?..

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -376,6 +374,20 @@ def run(
...
 
+    def add_rule(self, rule: Rule) -> None:
+        """
+        Add Rule object to existing profiler object by reconciling profiler rules and updating _profiler_config.
+        """
+        rules_dict: dict = {rule.name: rule.to_json_dict()}
+        effective_rules: List[Rule] = self.reconcile_profiler_rules(
+            rules=rules_dict,
+        )
+        updated_rules: Optional[Dict[str, Dict[str, Any]]] = {
```
alexsherstinsky, line 377: @Shinnnyshinshin we need to add `rule: Rule` (style).  Thanks!

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -376,6 +374,20 @@ def run(
...
+            rule.name: rule.to_json_dict() for rule in effective_rules
+        }
+        self.rules = effective_rules
+        self._profiler_config.rules = updated_rules
```
alexsherstinsky, line 381: @Shinnnyshinshin Let us try `self.profiler_config.rules` (instead of `self._profiler_config.rules`) -- thanks!

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -829,13 +841,20 @@ def rules(self, value: List[Rule]):
 
     def to_json_dict(self) -> dict:
         rule: Rule
+        variables_dict: dict = {}
```
alexsherstinsky: @Shinnnyshinshin Let us declare `Optional[Dict[str, Dict[str, Any]]] = None,` and then either set it -- or if not, check for `None` and set it to `{}` if `None`.  This would be a more economical coding pattern.  Happy to discuss.  Thanks!

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -829,13 +841,20 @@ def rules(self, value: List[Rule]):
...
+
+        if self.variables:
+            variables_dict = self.variables.to_dict()
+
+        parameter_nodes: Optional[dict] = variables_dict.get("parameter_nodes")
```
alexsherstinsky: @Shinnnyshinshin Let us use `convert_variables_to_dict()` instead -- it will be available once my PR 4375 is merged.  Thanks.

tests/integration/profiling/rule_based_profilers/test_rule_based_profiler.py
```diff
@@ -1,14 +1,17 @@
...
 from ruamel.yaml import YAML
 
+import great_expectations.rule_based_profiler.domain_builder
 from great_expectations import DataContext
 from great_expectations.core import ExpectationSuite
 from great_expectations.core.batch import BatchRequest
 from great_expectations.rule_based_profiler.config.base import (
     ruleBasedProfilerConfigSchema,
 )
+from great_expectations.rule_based_profiler.rule import Rule
```
alexsherstinsky: @Shinnnyshinshin are these new imports needed?

alexsherstinsky (changes requested):
Various suggestions made -- happy to discuss.  Thanks, @Shinnnyshinshin !

great_expectations/data_context/data_context.py
```diff
@@ -3317,6 +3318,17 @@ def add_profiler(
...
+        self,
+        profiler: RuleBasedProfiler,
+        name: Optional[str] = None,
+        ge_cloud_id: Optional[str] = None,
+    ):
+        key: Union[
+            GeCloudIdentifier, ConfigurationIdentifier
+        ] = self.profiler_store.determine_key(name=name, ge_cloud_id=ge_cloud_id)
+        self.profiler_store.set(key=key, value=profiler.config)
+
```
cdkini: ❤️ 

Nitpick - please add `-> None`

great_expectations/data_context/store/configuration_store.py
```diff
@@ -149,3 +150,19 @@ def self_check(self, pretty_print: bool = True) -> dict:
...
+            ge_cloud_id
+        ), "Must provide either name or ge_cloud_id."
+
+        key: Union[GeCloudIdentifier, ConfigurationIdentifier]
+        if ge_cloud_id:
+            key = GeCloudIdentifier(resource_type="contract", ge_cloud_id=ge_cloud_id)
+        else:
+            key = ConfigurationIdentifier(configuration_key=name)
+
+        return key
```
cdkini, line 168: love it!

great_expectations/rule_based_profiler/rule/rule.py
cdkini: Yup this is something I'm often guilty of. Let's utilize the property (even within the class) as a general best practice.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -376,6 +374,20 @@ def run(
...
         return expectation_suite
 
+    def add_rule(self, rule: Rule) -> None:
+        """
+        Add Rule object to existing profiler object by reconciling profiler rules and updating _profiler_config.
+        """
+        rules_dict: dict = {rule.name: rule.to_json_dict()}
+        effective_rules: List[Rule] = self.reconcile_profiler_rules(
+            rules=rules_dict,
+        )
```
cdkini: Ah I feel a little iffy about this.

In `reconcile_profiler_rules`, we do a bunch of work, get to a point where we create dicts, and then convert them to Rules. Here, we're taking that work and then going back a step to convert it back to dicts.

I see @alexsherstinsky's concern about maintainability but I really would prefer not to backtrack like that. It feels wasteful no? Performance isn't everything so I can see an argument being made against it - I do think we can find a happy medium though.

On a semi-related note, do we even need "reconciliation" here? It's this heavyweight, complex function that makes a lot of sense for runtime overrides but adding a rule is **so much** simpler right? Why do I need to do this complex reconciliation if all I'm doing is conditionally popping a value and appending? It feels slightly overengineered.

tests/integration/profiling/rule_based_profilers/test_rule_based_profiler.py
cdkini: nice catch Alex!

tests/rule_based_profiler/test_rule_based_profiler.py
```diff
@@ -1132,6 +1163,36 @@ def test_get_profiler(
...
+    ):
+        profiler = RuleBasedProfiler.get_profiler(
+            data_context=mock_data_context,
+            profiler_store=populated_profiler_store,
+            name="my_profiler",
+            ge_cloud_id=None,
+        )
+    assert isinstance(profiler, RuleBasedProfiler)
+
+
```
cdkini: If this is a context level method, I think we should move it to the context tests right?

What is it that we're aiming to test here? Given a profiler and name/ge_cloud_id, we can determine a key and use it to store the config on that profiler object.

As such, I think we might want to move away from the mocks we have here.

tests/rule_based_profiler/test_rule_based_profiler.py
```diff
@@ -1132,6 +1163,36 @@ def test_get_profiler(
...
+    populated_profiler_store: ProfilerStore,
+    profiler_config_with_placeholder_args: RuleBasedProfilerConfig,
+):
+    with mock.patch(
+        "great_expectations.data_context.store.profiler_store.ProfilerStore.set",
+        return_value=profiler_config_with_placeholder_args,
+    ):
+        mock_data_context.save_profiler(
+            profiler=profiler_config_with_placeholder_args,
+            profiler_store=populated_profiler_store,
```
cdkini: This is not actually part of the signature right? I think we might want to brainstorm how best to test this functionality.

tests/rule_based_profiler/test_rule_based_profiler.py
cdkini: since the store deals with filesystem interaction, I think it may make sense to actually mock that object? we can then assert that `set()` was invoked from the store with the appropriate config arg? I think there are a few ways to go about getting coverage.

As this is designed to be a unit test, I'd like us to isolate just the `save` method; could we remove the call to `get_profiler`?

tests/rule_based_profiler/test_rule_based_profiler.py
```diff
@@ -1211,3 +1272,48 @@ def test_list_profilers_in_cloud_mode(mock_profiler_store: mock.MagicMock):
...
+@mock.patch(
+    "great_expectations.rule_based_profiler.domain_builder.SimpleColumnSuffixDomainBuilder"
+)
+@mock.patch(
+    "great_expectations.rule_based_profiler.expectation_configuration_builder.DefaultExpectationConfigurationBuilder"
+)
+def test_add_rule(
+    mock_data_context: mock.MagicMock,
+    mock_domain_builder: mock.MagicMock,
+    mock_expectation_configuration_builder: mock.MagicMock,
```
cdkini: Mocks are a little strange - they work as follows:
```python
@mock.patch(A)
@mock.patch(B)
@mock.patch(C)
def my_func(
    mock_C
    mock_B
    mock_A
)
```

As it stands, you `mock_data_context` is actually mocking the `DefaultExpectationConfigurationBuilder`. I only know this after hours of very painful debugging 😢

tests/rule_based_profiler/test_rule_based_profiler.py
```diff
@@ -1211,3 +1272,48 @@ def test_list_profilers_in_cloud_mode(mock_profiler_store: mock.MagicMock):
...
+    sample_rule_dict: dict,
+):
+
+    profiler: RuleBasedProfiler = RuleBasedProfiler(
+        name="my_mocked_rbp", data_context=mock_data_context, config_version=1.0
```
cdkini: The RBP is not actually mocked right? Could we change the name just to prevent any confusion?

tests/rule_based_profiler/test_rule_based_profiler.py
```diff
@@ -1211,3 +1272,48 @@ def test_list_profilers_in_cloud_mode(mock_profiler_store: mock.MagicMock):
...
+    assert len(profiler.rules) == 1
+
+    second_rule: Rule = Rule(
+        name="second_rule",
+        domain_builder=mock_domain_builder,
+        expectation_configuration_builders=mock_expectation_configuration_builder,
+    )
+    second_rule.to_json_dict = MagicMock(return_value=sample_rule_dict)
+    profiler.add_rule(second_rule)
+    assert len(profiler.rules) == 2
```
cdkini: I think we should separate this out into 2 tests - one for the case where we remove a rule with the same name and one where we simply append.

As a general rule of thumb, I try to write a unit test for every possible code path. So with something like this:
```python
if x:
    ...
elif y:
    ...
else:
    raise Exception(...)
```
I'd write 3 separate tests!

tests/integration/profiling/rule_based_profilers/test_profiler_basic_workflows.py
```diff
@@ -0,0 +1,306 @@
...
+        expectation_type="expect_column_values_to_not_be_null",
+        column="$domain.domain_kwargs.column",
+    )
+    simple_rule: Rule = Rule(
+        name="rule_with_no_variables_no_parameters",
+        domain_builder=domain_builder,
+        expectation_configuration_builders=[default_expectation_configuration_builder],
+    )
+    print(simple_rule.to_json_dict())
+    print("###")
```
cdkini: Should we remove these?

cdkini (changes requested):
Looks very close but I have two points I'd like to address:
* This whole reconciliation usage - I don't think it is entirely necessary but this is perhaps a conversation I should have with Alex
* I think we should further split up our unit tests to isolate the behavior under test. Really appreciate the thorough testing here!

tests/integration/profiling/rule_based_profilers/test_profiler_basic_workflows.py
Shinnnyshinshin: shooot this is my fault 😓

great_expectations/data_context/store/checkpoint_store.py
Shinnnyshinshin: This was a mis-understanding on my part. Changed back to `self.`

great_expectations/data_context/store/checkpoint_store.py
Shinnnyshinshin: This was a mis-understanding on my part. Changed back to `self.`

great_expectations/data_context/store/checkpoint_store.py
Shinnnyshinshin: This was a mis-understanding on my part. Changed back to `self.`

great_expectations/rule_based_profiler/rule/rule.py
Shinnnyshinshin: Done :) Thank you for catching this @alexsherstinsky

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: Done

great_expectations/data_context/data_context.py
Shinnnyshinshin: Done!

tests/data_context/test_data_context_profilers.py
```diff
@@ -115,3 +115,33 @@ def test_run_profiler_on_data_emits_proper_usage_stats(
...
+        "great_expectations.data_context.store.profiler_store.ProfilerStore.get",
+        return_value=profiler_config_with_placeholder_args,
+    ):
+        profiler = RuleBasedProfiler.get_profiler(
+            data_context=mock_data_context,
+            profiler_store=populated_profiler_store,
+            name="my_profiler",
+            ge_cloud_id=None,
+        )
+    assert isinstance(profiler, RuleBasedProfiler)
```
cdkini, line 147: I would probably omit this portion because this test is now reliant on two units `save_profiler` and `get_profiler` - I think you could get away with the `save_profiler` command and perhaps asserting a mock call.

tests/rule_based_profiler/test_rule_based_profiler.py
```diff
@@ -1211,3 +1243,118 @@ def test_list_profilers_in_cloud_mode(mock_profiler_store: mock.MagicMock):
...
+
+
+@mock.patch("great_expectations.data_context.data_context.DataContext")
+def test_add_rule_bad_rule(
```
cdkini, line 1346: 😆 love the name

cdkini (approved):
One minor comment about a unit test but otherwise LGTM! Not a blocking comment so happily signing off.

Really strong work here 🎉 a real pleasure to see each iteration getting closer and closer to this refined end product.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -807,13 +826,15 @@ def rules(self, value: List[Rule]):
 
     def to_json_dict(self) -> dict:
         rule: Rule
+        variables_dict: Optional[Dict[str, Any]] = convert_variables_to_dict(
```
alexsherstinsky, line 831: ❤️

alexsherstinsky (approved):
LGTM!!!

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: actually this `rule` is a parameter not a local variable. Resolved synchronously

PR #4384, opened 2022-03-11
[MAINTENANCE] Change schedule for `packaging_and_installation` pipeline to run at off-hours

cdkini:
- Forgot that cron works on UTC - let's change this so it now runs at 12am PST or 3am EST


Shinnnyshinshin (approved):
Agreed :) LGTM @cdkini

PR #4387, opened 2022-03-11
[MAINTENANCE] Implicitly anonymize object based on __module__

cdkini:
- Refactor usage stats so we don't need to declare `self._ge_classes` in each Anonymizer subclass.
- This is PR 1 of 2 - the feature in question is very susceptible to breakage so we're focusing on simply removing the reliance here and then refactoring out the actually attributes in a subsequent PR.
- Various unit tests to demonstrate the different code paths we can take through the Anonymizer.
- An interesting dilemma around multiple-inheritance - I've opted to get information on the first parent available but an argument can be made to panic and exit early in the face of ambiguous parents.

### Definition of Done

Thank you for submitting!

### DIFF
--- a/great_expectations/core/usage_statistics/anonymizers/anonymizer.py
+++ b/great_expectations/core/usage_statistics/anonymizers/anonymizer.py
@@ -76,6 +76,7 @@ class Anonymizer:
             if Anonymizer._is_core_great_expectations_class(object_module_name):
                 anonymized_info_dict["parent_class"] = object_class_name
             elif len(parents) != 1:
+                # __bases__ provides us with the inheritance hierarchy - all GE core objects subclass exactly ONE parent class
                 if len(parents) == 0:
                     logger.info(
                         "Could not find any parent classes when anonymizing payload"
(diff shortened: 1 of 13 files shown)


great_expectations/core/usage_statistics/anonymizers/anonymizer.py
```diff
@@ -41,12 +44,11 @@ def anonymize(self, string_):
 
     def anonymize_object_info(
```
alexsherstinsky, line 45: @cdkini This is odd.  The self._ge_classes is still defined in the parent class, but not used any more.  So what happens to that list of action classes?  Thanks.

alexsherstinsky (changes requested):
Not sure I follow one aspect of the change (per comment).  Thank you.

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
cdkini: `self._ge_classes` still exists - this PR is one of two or three. The purpose of this PR is to remove `ge_classes` as an arg from this `anonymize_object_info` method, not to remove it entirely.

They're still strongly coupled with the Anonymizer as a result of other utilities like `_is_parent_class_recognized`. Due to the fact that usage stats is very easy to break, I want to do this in steps. That make sense?

alexsherstinsky: 🙇

alexsherstinsky (approved):
LGTM -- thanks!

PR #4444, opened 2022-03-17
Feature/cloud 385/mask cloud creds

joshua-stauffer:

- expands PasswordMasker functionality to handle cloud tokens
- adds a new method to DataContextConfig: `to_sanitized_json_dict`, which uses the new PasswordMasker functionality
- updates `DataContext.list_datasources()` and `DataContext.list_stores()` to use the same functionality
- resolves CLOUD-385

### Definition of Done

Thank you for submitting!


joshua-stauffer:
Looks like two tests are failing due to something regarding bigquery -- it might make sense to remove bigquery from the test data, since there isn't a password to be masked anyways.

great_expectations/data_context/types/base.py
```diff
@@ -1259,7 +1259,7 @@ def to_json_dict(self):
         return {
             "base_url": self.base_url,
             "organization_id": self.organization_id,
-            "access_token": self.access_token,
+            "access_token": "xxx",  # self.access_token,
```
alexsherstinsky: @joshua-stauffer Could we please pick a different string?  For example, "abc".  Thanks!

alexsherstinsky (changes requested):
@joshua-stauffer requesting a small change -- otherwise, happy to approve -- thanks!

great_expectations/data_context/types/base.py
```diff
@@ -2085,6 +2085,16 @@ def to_json_dict(self) -> dict:
         serializeable_dict: dict = convert_to_json_serializable(data=dict_obj)
         return serializeable_dict
 
+    def to_sanitized_json_dict(self) -> dict:
+        """
+        Wrapper for `to_json_dict` which ensures sensitive fields are properly masked.
+        """
+        # postpone importing to avoid circular imports
+        from great_expectations.data_context.util import PasswordMasker
```
cdkini, line 2096: ❤️

great_expectations/data_context/util.py
```diff
@@ -567,3 +570,79 @@ def mask_db_url(url: str, use_urlparse: bool = False, **kwargs) -> str:
                 )
 
             return masked_url
+
+    @classmethod
+    def sanitize_data_context_config(cls, config: dict) -> dict:
```
cdkini: I think this file is great but is there a way we could possibly genericize this whole utility? I'm wary of having to constantly expand this class to fit different core GE objects. I think it's probably fine as is but what are your thoughts around a recursive method that anonymizes anything in a list of keywords?

Something like:
```python
SENSITIVE_STRINGS = set(...)

@classmethod
def sanitize(cls, config: dict) -> dict:
    if not isinstance(config, dict):
        raise

    config_copy = safe_deep_copy(config)
    
    def _recursive_cleaner_method(config: Any) -> None:
        if isinstance(config, dict):
            for key, val in config.items():
                if key in self.SENSITIVE_STRINGS and isinstance(val, str):
                    # obfuscate
                _recursive_cleaner_method(val)
        elif isinstance(config, list):
            for val in config:
                _recursive_cleaner_method(val)

    _recursive_cleaner_method(config_config) # Perform anonymization in place
    
    return config_copy
```

tests/data_context/test_data_context_utils.py
```diff
@@ -286,6 +287,174 @@ def test_password_masker_mask_db_url(monkeypatch, tmp_path):
...
+            assert (
+                store_config["store_backend"]["ge_cloud_credentials"]["access_token"]
+                != ge_cloud_access_token
+            )
+        except KeyError:
+            # expect config to not be changed
+            assert (
+                store_config
+                == data_context_config_dict_with_cloud_backed_stores["stores"][name]
+            )
```
cdkini: Although this is definitely a unit test, I feel like it's pretty complicated and is testing a lot of different things. Could we simplify this at all? Maybe a snapshot test or splitting it up into a few different tests?

Having `try/except` and `if/else` for various asserts makes debugging this a little difficult imo.

cdkini (changes requested):
Really nice work! Two minor comments:
* Suggestion about possible way to make this more generic (not blocking).
* Request to split up/simplify the unit tests if at all possible.

Let me know your thoughts and I'd be more than happy to approve!

tests/data_context/test_data_context_utils.py
joshua-stauffer: thanks for the feedback -- i'll break these cases out into separate tests.

great_expectations/data_context/types/base.py
joshua-stauffer: hi @alexsherstinsky -- sure, I'll update this to match `PasswordMasker.MASKED_PASSWORD_STRING`. Thanks for the feedback!

great_expectations/data_context/util.py
joshua-stauffer: thanks @cdkini, this is a great idea -- I had originally decided against this because there is different processing required depending on the key value, but i believe integrating our approaches works well.

tests/data_context/test_data_context_utils.py
joshua-stauffer: I've factored the tests out into their own functions, but have left the try/except, since it's the most straightforward way to iterate through unknown values and check their configuration. Let me know if this is easier to parse!

great_expectations/data_context/types/base.py
```diff
@@ -1256,10 +1256,13 @@ def account_id(self):
...
     def to_json_dict(self):
+        # postpone importing to avoid circular imports
+        from great_expectations.data_context.util import PasswordMasker
+
         return {
             "base_url": self.base_url,
             "organization_id": self.organization_id,
-            "access_token": self.access_token,
+            "access_token": PasswordMasker.MASKED_PASSWORD_STRING,  # self.access_token,
             "account_id": self.account_id,  # TODO: remove when account_id is deprecated
```
cdkini: Are either of these comments needed? For the deprecation, do we have a plan in place to remove that?

tests/data_context/test_data_context_utils.py
```diff
@@ -286,6 +287,222 @@ def test_password_masker_mask_db_url(monkeypatch, tmp_path):
...
+                != ge_cloud_access_token
+            )
+        except KeyError:
+            # a field in store_config["store_backend"]["ge_cloud_credentials"]["access_token"]
+            # doesn't exist, so we expect this config to be unchanged
+            assert (
+                store_config
+                == data_context_config_dict_with_cloud_backed_stores["stores"][name]
+            )
+
```
cdkini: I still am a bit iffy on this general pattern as it introduces some tricky conditional logic in a test, which should be as simple and straightforward as possible (imo).

I won't block on this but I'm definitely of the opinion that the fewer asserts and overall conditional branches, the easier the test is to read and maintain.

tests/data_context/test_data_context_utils.py
```diff
@@ -286,6 +287,222 @@ def test_password_masker_mask_db_url(monkeypatch, tmp_path):
...
+            else:
+                # it doesn't contain a password, so make sure it's unaltered
+                assert processed_config == unaltered_datasources[name]
+
+        # processed_config either doesn't have an `execution_engine` field,
+        # or a `connection_string` field
+        else:
+            # expect this config to be unaltered
+            assert processed_config == unaltered_datasources[name]
+
```
cdkini, line 386: Same general thought here! Could we just assert against a dict so we know the full expected vs actual?

tests/data_context/test_data_context_utils.py
```diff
@@ -286,6 +287,222 @@ def test_password_masker_mask_db_url(monkeypatch, tmp_path):
...
+    config = {"credentials": {"url": "bigquery://foo/bar"}}
+    config_copy = safe_deep_copy(config)
+    res = PasswordMasker.sanitize_config(config_copy)
+    assert res == config
+
+
+def test_sanitize_config_with_nested_url_field(
+    conn_string_password, conn_string_with_embedded_password
+):
+    """Test case for PasswordPasker.sanitize_config"""
```
cdkini: Typo in all docstrings. Please fix before merging 😄 

Nitpick - since the name of the test specifies the method in question, is this necessary? I think you could probably get by without it since these are very simple (and well written!) unit test.

cdkini (approved):
LGTM but I am still a little iffy on the following tests:
* test_sanitize_config_masks_execution_engine_connection_strings
* test_sanitize_config_masks_cloud_store_backend_access_tokens

I'd love it if we could simplify the logic there just to make it easier to debug and work with. That being said, I definitely won't block on that. Really nice work here, thanks so much for resolving my comments!

great_expectations/data_context/types/base.py
joshua-stauffer: good point -- I'll remove the access token comment, that was just a reminder for myself, but the account_id comment is outside of the scope of this PR, so I'll leave it. I believe it's scheduled to be removed on the next major release, but I don't have the latest context on that.

tests/data_context/test_data_context_utils.py
joshua-stauffer: I hear your point on the complexity of the test! I refactored the try/except to make it more straightforward to read. I personally find this approach to be preferable to a snapshot-style test since I see it as less fragile. From my perspective, this test shouldn't care about the rest of the contents of the config dict, only the fields under test.

alexsherstinsky (approved):
LGTM

PR #4446, opened 2022-03-17
[FEATURE] Rule-Based Profiler: Implement semantic type logic for all DomainBuilder

alexsherstinsky:
### Scope
* We remove the `SimpleSemanticTypeDomainBuilder` and instead, add semantic type capabilities to all column style `DomainBuilder` classes.
* Moreover, we allow both include and exclude semantic types to be configured (in addition to explicit include and exclude column lists).
* Using the `$variable` syntax to specify include and exclude semantic types is supported.
* Existing tests have been updated (using `ColumnDomainBuilder` instead of `SimpleSemanticTypeDomainBuilder`).

- JIRA: GREAT-464/GREAT-680
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -33,6 +37,12 @@ def __init__(
         batch_list: Optional[List[Batch]] = None,
         batch_request: Optional[Union[BatchRequest, RuntimeBatchRequest, dict]] = None,
         data_context: Optional["DataContext"] = None,  # noqa: F821
+        include_semantic_types: Optional[
+            Union[str, SemanticDomainTypes, List[Union[str, SemanticDomainTypes]]]
+        ] = None,
+        exclude_semantic_types: Optional[
+            Union[str, SemanticDomainTypes, List[Union[str, SemanticDomainTypes]]]
+        ] = None,
```
anthonyburdi, line 45: We should add some defaults here to exclude columns e.g. unique columns

great_expectations/rule_based_profiler/domain_builder/column_domain_builder.py
```diff
@@ -131,9 +179,225 @@ def _get_domains(
         """
         Obtains and returns domains for all columns of a table (or for configured columns, if they exist in the table).
         """
-        return build_simple_domains_from_column_names(
-            column_names=self.get_effective_column_names(
-                variables=variables,
-            ),
-            domain_type=self.domain_type,
+        batch_ids: List[str] = self.get_batch_ids(variables=variables)
```
anthonyburdi, line 230: Not blocking but this method is getting pretty large, could be split up in a future refactor.

great_expectations/rule_based_profiler/domain_builder/column_domain_builder.py
```diff
@@ -131,9 +179,225 @@ def _get_domains(
...
+                },
+            )
+            for column_name in candidate_column_names
+        ]
+
+        return domains
+
+    # This method (default implementation) can be overwritten (with different implementation mechanisms) by subclasses.
+    @staticmethod
+    def infer_semantic_domain_type_from_table_column_type(
```
anthonyburdi: In a future refactor could some of this functionality be moved to a separate `ColumnSemanticTypeFilter` or similar?

anthonyburdi (changes requested):
Looks good other than adding semantic type exclusion defaults to `CategoricalColumnDomainBuilder`

great_expectations/rule_based_profiler/domain_builder/column_domain_builder.py
alexsherstinsky: @anthonyburdi Yes.  Since this did not make the release, I might just follow up on all of your feedback carrying action items this morning...  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: Great idea, @anthonyburdi !!!  Adding (in next commit).

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -82,6 +86,13 @@ def __init__(
             max_proportion_unique: proportion of unique values for a
                 custom cardinality limit to use when filtering columns.
         """
+        if exclude_semantic_types is None:
+            exclude_semantic_types = [
```
anthonyburdi, line 95: Awesome! Thank you for adding these defaults. I think we also want to add a few more:
- split numeric into integer and float. Maybe by adding `SemanticDomainTypes.FLOAT` and exclude that?
- I think BINARY is an interesting case. I think we want to include it, but not do the processing on it to figure out how many unique values there are
- STRING is also an interesting case. Do you think it would make sense to split shorter strings (char, varchar) from longer (text)? For example we scan shorter and skip scanning longer types?

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -45,6 +45,8 @@ def __init__(
         ] = None,
         include_column_names: Optional[Union[str, Optional[List[str]]]] = None,
         exclude_column_names: Optional[Union[str, Optional[List[str]]]] = None,
```
anthonyburdi, line 47: How about making the overridable default here to exclude `id` column names?

anthonyburdi: We also discussed excluding columns ending with `_id` but I think that one is less clear. It might be a foreign key to a small table with a set of options.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
```diff
@@ -45,6 +45,8 @@ def __init__(
...
+        semantic_type_filter_module_name: str = "great_expectations.rule_based_profiler.helpers.simple_semantic_type_filter",
+        semantic_type_filter_class_name: str = "SimpleSemanticTypeFilter",
```
anthonyburdi: I like that we include defaults here for the module/class name, thank you! I also wasn't expecting this to be overridable via plugin - I'd love to hear @donaldheppner's feedback on what we are planning to maintain as user configurable options (for example a user could build their own domain builder if they wanted to change the semantic type filtering functionality.

anthonyburdi (commented):
Thank you for making these changes! Apologies but I have a few more requests.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi this would be for a different PR, because these types came from before.  Do you agree?  Happy to discuss.  Thanks.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi this is for a separate work effort as well.  Shall we discuss to agree on what goes where and when?  Thanks.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi This is critical -- and stems from the requirements from last Spring/Summer work on the RBP.  Thanks.

cdkini: any chance we can use the actual type of the class (as opposed to the string)? perhaps not in this context but I'd love to use types instead of strings where possible because they are susceptible to typos and other misc errors.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi and I decided to include `FLOAT` in "excluded semantic types".

alexsherstinsky: @anthonyburdi and I decided to make "LOGIC" included.

anthonyburdi: (Also I misspoke above in my first comment when I wrote `BINARY` I meant `BOOLEAN`)

alexsherstinsky: @anthonyburdi How about we tackle strings as part of a future work?  Thanks.

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @anthonyburdi How about in the next PR, I fold the `SimpleColumnSuffixDomainBuilder` into the `ColumnDomainBuilder`, which will enable us to handle this case and more?  Thanks!

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @talagluck Do you have any thoughts regarding splitting `numeric` semantic type into "integral" (for integers) and fractional (for floats, etc.) and making the latter automatically excluded from the categorical domain builder?  If so, which names would you propose for these "split" semantic types?

great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py
alexsherstinsky: @cdkini We cannot -- for the reasons of `Marshmallow` `Schema` compatibility.

anthonyburdi (approved):
LGTM!

Issue #4407, opened 2022-03-14
Error: No module named 'packaging'


shpolina:
**Describe the bug**

We are running GE using pure python and started to experience an import error `ModuleNotFoundError: No module named 'packaging'` recently.

```
 from great_expectations.data_context.data_context import DataContext
  File "\lib\site-packages\great_expectations\__init__.py", line 7, in <module>    
    from great_expectations.data_context import DataContext
  File "\lib\site-packages\great_expectations\data_context\__init__.py", line 1, in <module>
    from .data_context import BaseDataContext, DataContext, ExplorerDataContext
  File "\lib\site-packages\great_expectations\data_context\data_context.py", line 23, in <module>
    from great_expectations.core.config_peer import ConfigPeer
  File "\lib\site-packages\great_expectations\core\config_peer.py", line 6, in <module>
    from great_expectations.data_context.types.base import BaseYamlConfig
  File "\lib\site-packages\great_expectations\data_context\types\base.py", line 29, in <module>
    from great_expectations.util import deep_filter_properties_iterable
  File "\lib\site-packages\great_expectations\util.py", line 32, in <module>       
    from packaging import version
ModuleNotFoundError: No module named 'packaging'
```
It looks like `packaging` is not explicitly declared as a dependency in [requirements.txt](https://github.com/great-expectations/great_expectations/blob/86583dc5a073f6e769356e39482c43d1e2976359/requirements.txt) and we were
previously getting it thanks to the explicitly declared `ipywidgets` dependency. But `ipywidgets` started to pick
up a later version which no longer brings `packaging` with it.  

1. in clean environment pip install great-expectations
2. import `from great_expectations.data_context import DataContext`
3. Try to invoke the code - and get the error

**Environment (please complete the following information):**
 - Operating System: Windows
 - Great Expectations Version: 0.14.10

austiezr:
@shpolina Thank you for surfacing; we'll investigate ASAP.

austiezr:
Closed by #4447 ; thank you @shpolina !

shpolina:
Thank you very much for the prompt fix.

PR #4447, opened 2022-03-17
[BUGFIX] Add any dependencies we import directly, but don't have as explicit requirements

kenwade4:
These are things that other explicit requirements may have as their dependencies.

Issue https://github.com/great-expectations/great_expectations/issues/4407 pointed out that `packaging` was previously provided by `ipywidgets` but it is not any more (which raised `ModuleNotFoundError` in their installation).

The output of the following shell command was filtered (via piping to `grep -vE '\b(thing1|thing2|...)\b'`) to determine what packages should be added (after manual checks).

```
grep -R --no-filename --exclude-dir=docs_rtd --exclude-dir=docs --exclude-dir=contrib --exclude-dir=tests --exclude-dir=venv -E '^ *(from|import)\b' . | sort | uniq -c | sort -k1,1nr -k2
```

- Add any dependencies we import directly, but don't have as explicit requirements

### Definition of Done

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,16 @@
 altair>=4.0.0,<5
 Click>=7.1.2
+colorama>=0.4.3
+cryptography>=3.2,<4.0.0
 importlib-metadata>=1.7.0 # (included in Python 3.8 by default.)
 ipywidgets>=7.5.1
 jinja2>=2.10
 jsonpatch>=1.22
 jsonschema>=2.5.1
 mistune>=0.8.4,<2.0.0
+nbformat>=5.0
 numpy>=1.14.1
+packaging
 pandas>=0.23.0
 pyparsing>=2.4,<3
 python-dateutil>=2.8.1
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,12 +1,16 @@
...
+cryptography>=3.2,<4.0.0
 importlib-metadata>=1.7.0 # (included in Python 3.8 by default.)
 ipywidgets>=7.5.1
 jinja2>=2.10
 jsonpatch>=1.22
 jsonschema>=2.5.1
 mistune>=0.8.4,<2.0.0
+nbformat>=5.0
 numpy>=1.14.1
+packaging
```
cdkini, line 13: should we pin this? or set a bound? just asking because we tend to do it for the others!

cdkini (approved):

requirements.txt
kenwade4: Only reason I didn't was that every one of the dev dependencies that had some requirement that required packaging required "any"

PR #4478, opened 2022-03-20
[DOCS] Add missing word and fix wrong dataset reference

strickvl:
If I'm not mistaken, we're missing an `on` here, and we're generating the reasonable range based on the dataset which we're feeding into the Profiler, which is January (not February).

- fix missing word
- fix dataset reference

### Definition of Done


austiezr (approved):
LGTM! Thank you @strickvl !

PR #4485, opened 2022-03-21
[MAINTENANCE] Refactor Anonymizer utilizing the Strategy design pattern

cdkini:
- Refactor Anonymizer using the Strategy design pattern to create a single, uniform interface for callers to interact with.
  - This abstracts away the need to know about the child classes.

Here are the following layers to be aware of:
* `BaseAnonymizer`
  * Contains generic behavior that applies to all or most `Anonymizer` child classes. Additionally, contains abstract methods `anonymize` and `can_handle` that are to be leveraged by children to conditionally anonymize input objects.
* `ProfilerRunAnonymizer`/`CheckpointRunAnonymizer`/etc
  * Fulfills the contract set by `BaseAnonymizer` by inheriting the class and fleshing out the aforementioned methods.
  * Contain a reference to the primary `Anonymizer` - see below for what that does!
* `Anonymizer`
  * Wrapper class that conducts the "strategy"
  * Iterates through possible specific `Anonymizer` classes, sees if one `can_handle`, and conditionally determines which "strategy" to use for anonymization.
  * In the case a specific anonymizer class needs behavior it isn't responsible for (i.e. `ProfilerAnonymizer` needs an anonymized batch request so it has to talk to `Anonymizer` to get those results).

This allows us to make all anonymizer behavior private and encapsulated from external users. If you want to anonymize something, you need to call `Anonymizer.anonymize()`.

The diff is HUGE but that's mainly because there are a lot of moving parts. Let me detail what's important to review:
* base.py - The abstract class with shared behavior
* anonymizer.py - The main class that conducts the "strategy"
  - Note that this implementation has been impacted by a whole lot of cyclic import issues. I've made design decisions that I wouldn't otherwise have made (using instance attributes when class attributes would make a lot more sense).
 
Outside of that, everything is a specific object anonymizer implementation (which has existed in the past) or changes to existing code to use this new, uniform API (`Anonymizer.anonymize()`).

### Definition of Done


great_expectations/core/usage_statistics/anonymizers/base.py
```diff
@@ -0,0 +1,692 @@
...
+    RUNTIME_PARAMETERS_KEYS,
+    BATCH_SPEC_PASSTHROUGH_KEYS,
+    BATCH_REQUEST_FLATTENED_KEYS,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+class BaseAnonymizer(ABC):
```
cdkini, line 14: Take all shared behavior from `Anonymizer` and place here - this is subclassed throughout. All child classes must fulfill the two abstract methods.

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
```diff
@@ -1,765 +1,136 @@
...
+            return self._anonymize_data_connector_info(**kwargs)
+        if self._is_batch_info(obj=obj):
+            return self._anonymize_batch_info(batch=obj)
+        if self._is_store_info(info=kwargs):
+            return self._anonymize_store_info(**kwargs)
+        if isinstance(obj, str):
+            return self._anonymize_string(string_=obj)
+        raise TypeError(
+            f"The type {type(obj)} cannot be handled by the Anonymizer; no suitable strategy found."
         )
```
cdkini: This is the core logic - check all each anonymizer strategy in `self.strategies`.

If that strategy `can_handle`, use that implementation's `anonymize` method. If no specialized `Anonymizer` child class works, we revert to generic strategies as noted in `self._anonymize`.

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
```diff
@@ -1,765 +1,136 @@
...
+    @staticmethod
+    def _is_store_info(info: dict):
+        return "store_name" in info or "store_obj" in info
 
-        expectations = expectation_suite.expectations
-        expectation_types = [
-            expectation.expectation_type for expectation in expectations
+    @staticmethod
+    def can_handle(obj: object, **kwargs) -> bool:
+        return isinstance(obj, object)
```
cdkini: Base class can handle everything!

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
```diff
@@ -1,765 +1,136 @@
...
+
+    def build_init_payload(self, data_context: "DataContext") -> dict:  # noqa: F821
+        """Adds information that may be available only after full data context construction, but is useful to
+        calculate only one time (for example, anonymization)."""
```
cdkini: Move logic from `usage_statistics` to `anonymizer` - most of the method is anonymizing behavior so it makes more sense here. Additionally, this lets us keep all methods private.

anthonyburdi: Totally works - this feels a bit weird to me though from a responsibility standpoint to include in any of the Anonymizer classes. I don't have a better idea though so just thinking out loud :)

great_expectations/core/usage_statistics/anonymizers/checkpoint_run_anonymizer.py
```diff
@@ -3,60 +3,92 @@
...
+    def anonymize(self, obj: object = None, **kwargs) -> Any:
+        assert self.can_handle(
+            obj=obj
+        ), "CheckpointRunAnonymizer can only handle objects of type Checkpoint or CheckpointConfig"
+        if "config" in kwargs:
+            return self._anonymize_checkpoint_info(**kwargs)
+        return self._anonymize_checkpoint_run(obj=obj, **kwargs)
 
-        self._salt = salt
+    def _anonymize_checkpoint_info(self, name: str, config: dict) -> dict:
```
anthonyburdi: Just an optional suggestion - could this be `_anonymize_checkpoint_config()` just to be more descriptive that it acts on checkpoint config dicts?

great_expectations/core/usage_statistics/usage_statistics.py
```diff
@@ -63,21 +62,7 @@ def __init__(
         # the risk of cyclic import issues. If these anonymizers have been imported at any earlier point
         # in the program's lifetime, retrieval of the import will be O(1) and not impact performance.
```
anthonyburdi: Looks like you can remove these 3 comments

great_expectations/data_context/types/base.py
```diff
@@ -2547,6 +2557,112 @@ def __str__(self) -> str:
         """
         return self.__repr__()
 
+    # noinspection PyUnusedLocal,PyUnresolvedReferences
+    @staticmethod
+    def resolve_config_using_acceptable_arguments(
```
anthonyburdi, line 2575: 👏 👏 👏 👏 👏  I love this method here :)

anthonyburdi (commented):
LGTM! I'll hold off on approving so @donaldheppner can take a look.

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
cdkini: Ah yes this is a very valid point. To be perfectly honest, the primary purpose here is to prevent having to expose functionality from the anonymizer.

By keeping the logic here, we can keep all helpers private and maintain the ideal that `anonymize` is the single public entrypoint into anonymizer behavior.

If you feel strongly about this (and I think that's totally justifiable), let me know and I can figure something out.

great_expectations/core/usage_statistics/usage_statistics.py
cdkini: done! great catch 🙇🏽

great_expectations/core/usage_statistics/anonymizers/checkpoint_run_anonymizer.py
cdkini: I agree - done!

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
```diff
@@ -1,765 +1,202 @@
...
+            StoreAnonymizer,
+            StoreBackendAnonymizer,
+        ]
 
-        Args:
-            name (str): The name of the given profiler.
-            config (dict): The dictionary configuration of the given profiler.
+        # Instead of instantiating all child anonymizers, we perform JIT or on-demand instantiation.
+        # While this has some performance benefits, the primary reason is due to cyclic imports and recursive overflow.
+        self._cache: Dict[Type[BaseAnonymizer], BaseAnonymizer] = {}
```
cdkini: We have a few problems that this implementation solves:
1. Cyclic imports - Our "strategies" should definitely be class level attributes but there's no feasible way to make that work with the current state of the codebase. I've opted to place them in the constructor here to shield them from the global scope, allowing us to get by without issues.
2. Infinite recursion - if I were to instantiate all these child anonymizers right here and then use an instance of this anonymizer in the child classes, there would be infinite recursion.

Let me try to explain this as best as I can. Our primary `Anonymizer` class is sort of the web between all anonymizers in GE. While this class references specific implementations, those specific implementations also reference back here. This is deemed necessary because if a particular class requires behavior that is outside the responsibility of its class, it requests that behavior from this primary `Anonymizer` class. 

If my `DatasourceAnonymizer` has a constructor like this:
```python
def __init__(self, salt, aggregate_anonymizer):
    ...
```
it becomes evident that part of the definition of the class is its relationship with the primary `Anonymizer`. This bidrectional link causes recursion issues because both classes are defined in terms of each other. Instead, I've opted to only instantiate what we need at a given time and cache those results. This could probably be explained better live but this solution here seems to have mitigated much of the pain of this refactor.

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
```diff
@@ -1,765 +1,202 @@
...
 
-    def _anonymize_execution_engine_info(self, name: str, config: dict) -> dict:
-        anonymized_info_dict = {}
-        anonymized_info_dict["anonymized_name"] = self.anonymize(name)
+        if anonymizer_type is not None:
+            anonymizer: Optional[BaseAnonymizer] = self._cache.get(anonymizer_type)
+            if not anonymizer:
+                anonymizer = anonymizer_type(salt=self._salt, aggregate_anonymizer=self)
+                self._cache[anonymizer_type] = anonymizer
+            return anonymizer.anonymize(obj=obj, **kwargs)
```
cdkini: 1. Look through our strategies and see if we have a solution for the problem at hand
2. If we do, look at the cache and try to get an already instantiated object to do the work for us
  - If we have a cache miss, create a new instance and add to the cache
3. Run the anonymize method on the instance that's responsible for the object we're currently working with

great_expectations/core/usage_statistics/anonymizers/base.py
```diff
@@ -0,0 +1,217 @@
...
+        else:
+            self._salt = salt
+
+    @abstractmethod
+    def anonymize(self, obj: Optional[object], **kwargs) -> Any:
+        raise NotImplementedError
+
+    @abstractstaticmethod
+    def can_handle(obj: Optional[object], **kwargs) -> bool:
+        raise NotImplementedError
```
cdkini: This is still the core part of this class - what does a class need to be able to do to be considered an anonymizer?

Note that all the other methods (outside a few key ones) have been refactored out into separate anonymizer classes like before.

great_expectations/core/usage_statistics/anonymizers/batch_anonymizer.py
```diff
@@ -0,0 +1,112 @@
...
+        super().__init__(salt=salt)
+
+        if aggregate_anonymizer:
+            self._aggregate_anonymizer = aggregate_anonymizer
+        else:
+            from great_expectations.core.usage_statistics.anonymizers.anonymizer import (
+                Anonymizer,
+            )
+
+            self._aggregate_anonymizer = Anonymizer(salt=salt)
```
cdkini: Let's dissect this a bit:

We take in `salt` and an `aggregate_anonymizer`.

The former is also defined in the base class and is used to offset hashes when anonymizing payloads. The latter is a reference to the primary `Anonymizer` class - the actual `Anonymizer` will generally pass itself here so we can traverse back and forth from `BatchAnonymizer` to `Anonymizer` (and vice-versa).

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
```diff
@@ -1,765 +1,202 @@
...
-        anonymized_info_dict[
-            "anonymized_store_backend"
-        ] = self._anonymize_store_backend_info(store_backend_obj=store_backend_obj)
-
-        return anonymized_info_dict
-
-    def anonymize_profiler_info(self, name: str, config: dict) -> dict:
-        """Anonymize RuleBasedProfiler objs from the 'great_expectations.rule_based_profiler' module.
+        self._strategies: List[Type[BaseAnonymizer]] = [
+            CheckpointAnonymizer,
```
donaldheppner, line 51: I think you can make these instances and replace `CheckpointAnonymizer` with `CheckpointAnonymizer(self)`

great_expectations/core/usage_statistics/anonymizers/action_anonymizer.py
```diff
@@ -0,0 +1,45 @@
...
+
+class ActionAnonymizer(BaseAnonymizer):
+    def __init__(
+        self,
+        salt: Optional[str],
+        aggregate_anonymizer: Optional["Anonymizer"] = None,  # noqa: F821
+    ) -> None:
+        super().__init__(salt=salt)
+
+        if aggregate_anonymizer:
```
donaldheppner: Do we need this reference to be optional?

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
donaldheppner: This would also allow can_handle to be an instance method

donaldheppner (changes requested):

donaldheppner (approved):

PR #4487, opened 2022-03-21
[BUGFIX] Removes cryptography upper bound for general reqs

### LINKED ISSUE #4480: Requirement range for cryptography package prevents using recent versions
Hi there,

A [recent commit](https://github.com/great-expectations/great_expectations/commit/607c9129ff9c9a6008ac6146c5a61c1d3f24d676) introduced a hard range on the cryptography module <4.0 which effectively locks it to [3.4.8](https://github.com/pyca/cryptography/releases/tag/3.4.8) in August last year.

After 3.4.8, crytography releases are numbered 35.x, 36.x and so on.

My project will defer upgrading GE until cryptography can be upgraded to more recent versions - and I suspect others might too.

Is there a specific reason why the <4.0 was chosen as an upper bound? And can it be lifted to take advantage of recent cryptography package releases?

Thanks in advance!

austiezr:

- Removes `cryptography` upper bound from general reqs as it is only needed for our azure reqs

Previous Design Review notes:
- closes #4480 

### Definition of Done

Thank you for submitting!


requirements-dev-azure.txt
```diff
@@ -1,3 +1,4 @@
 azure-identity>=1.0.0
 azure-keyvault-secrets>=4.0.0
 azure-storage-blob>=12.5.0
+cryptography>=3.2,<4.0.0
```
kenwade4: don't add that there.... this file is for folks using azure secrets, not for our azure CI/CD pipelines

austiezr: Right; my understanding is that `azure-identity` relies on `msal` which relies on `cryptography<4.0.0`, meaning for our pipelines the upper bound shouldn't matter, but for folks using azure secrets, it does. Does that track with what you've seen?

kenwade4: To be honest, this file would only be used if someone did `pip install great_expectations[azure]` or if the repo was cloned and doing `pip install -r requirements.txt -r requirements-dev-lite.txt -r requirements-dev-azure.txt` (or `pip install -r requirements-dev.txt`)

Since the dependencies would be resolved at once, they would be getting whatever bound version of cryptography without specifying it here at all

austiezr: Love it. Happy to remove 🙇

kenwade4 (approved):

PR #4488, opened 2022-03-21
[FEATURE] Remove conflict between filter directives and row_conditions

anthonyburdi:
- Remove conflict of row_conditions and filter directives by enabling the use of `row_conditions` as well as `filter_null` and `filter_nan` directives where supported via `filter_conditions` in domain_kwargs
- This PR handles the filter row conditions as a separate step in `get_domain_records()` after handling `row_conditions`
- This PR does not change the existing experimental functionality of row_condition as described here: https://docs.greatexpectations.io/docs/reference/expectations/conditional_expectations

### Definition of Done

- [NA] I have made corresponding changes to the documentation

Thank you for submitting!


great_expectations/execution_engine/execution_engine.py
```diff
@@ -483,19 +487,16 @@ def add_column_row_condition(
...
-
         new_domain_kwargs = copy.deepcopy(domain_kwargs)
         assert "column" in domain_kwargs or column_name is not None
         if column_name is not None:
             column = column_name
         else:
             column = domain_kwargs["column"]
-        new_domain_kwargs["condition_parser"] = "great_expectations__experimental__"
-        new_domain_kwargs["row_condition"] = f'col("{column}").notnull()'
+        row_condition: RowCondition = RowCondition(
```
alexsherstinsky, line 495: @anthonyburdi How come this step should precede the `row_conditions` one?  If possible, could it be happen immediately after?  Thanks!

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -400,6 +404,13 @@ def get_domain_records(
             else:
                 raise ValidationError(f"Unable to find batch with batch_id {batch_id}")
 
+        filter_conditions: List[RowCondition] = domain_kwargs.get(
+            "filter_conditions", []
```
alexsherstinsky: @anthonyburdi Is `filter_condition` a new key, which replaces some older keys?

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -480,6 +491,31 @@ def get_domain_records(
...
+        assert all(
+            [
+                condition.type_ == RowConditionParserType.SPARK_SQL
+                for condition in row_conditions
+            ]
+        ), "All row conditions must have type SPARK_SQL"
+        raw_conditions: List[str] = [
+            row_condition.condition for row_condition in row_conditions
+        ]
+        joined_condition: str = " AND ".join(raw_conditions)
```
alexsherstinsky, line 514: @anthonyburdi `raw_conditions` or `row_conditions` for this one?  Thanks.

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -480,6 +491,31 @@ def get_domain_records(
...
+        Returns:
+            Single Row Condition combined
+        """
+        assert all(
+            [
+                condition.type_ == RowConditionParserType.SPARK_SQL
+                for condition in row_conditions
+            ]
+        ), "All row conditions must have type SPARK_SQL"
+        raw_conditions: List[str] = [
```
alexsherstinsky: @anthonyburdi Minor -- any chance to come up with a slightly different sounding/spelling variable name?  This one seems too close and thus might be error prone for potential typos.

great_expectations/execution_engine/sparkdf_execution_engine.py
alexsherstinsky, line 514: @anthonyburdi If possible, could we use the `SQLAlchemy` way?  We have a few examples; happy to discuss.  Thanks.

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -523,31 +559,37 @@ def add_column_row_condition(
...
-        if filter_null and filter_nan:
-            new_domain_kwargs[
-                "row_condition"
-            ] = f"NOT isnan({column}) AND {column} IS NOT NULL"
-        elif filter_null:
-            new_domain_kwargs["row_condition"] = f"{column} IS NOT NULL"
-        elif filter_nan:
-            new_domain_kwargs["row_condition"] = f"NOT isnan({column})"
-        else:
+
```
alexsherstinsky, line 569: @anthonyburdi If possible, we should try to use the `SQLAlchemy` way.  Happy to discuss.  Thanks.

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -523,31 +559,37 @@ def add_column_row_condition(
         self, domain_kwargs, column_name=None, filter_null=True, filter_nan=False
     ):
         # We explicitly handle filter_nan & filter_null for spark using a spark-native condition
-        if "row_condition" in domain_kwargs and domain_kwargs["row_condition"]:
```
alexsherstinsky, line 526: @anthonyburdi Where does `row_condition` processing go now?  Thanks.

great_expectations/execution_engine/execution_engine.py
```diff
@@ -483,19 +487,16 @@ def add_column_row_condition(
...
+            condition=f'col("{column}").notnull()', type_=RowConditionParserType.GE
+        )
+        new_domain_kwargs.setdefault("filter_conditions", []).append(row_condition)
```
cdkini, line 499: I can't believe I didn't know about this method 👀

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -480,6 +491,31 @@ def get_domain_records(
...
+
+        Returns:
+            Single Row Condition combined
+        """
+        assert all(
+            [
+                condition.type_ == RowConditionParserType.SPARK_SQL
+                for condition in row_conditions
+            ]
+        ), "All row conditions must have type SPARK_SQL"
```
cdkini: Nitpick - could we get rid of the list here?

```python
all(condition.type == RowConditionParserType.SPARK_SQL for condition in row_conditions)
```

Does that work? Always get confused here but we can possibly skip creating a list.

great_expectations/execution_engine/sparkdf_execution_engine.py
anthonyburdi: It's a new key that doesn't replace any existing keys that I am aware of or able to find. My hope is that we can eventually do away with this method and eventually specify RowConditions directly but this PR is meant as a first pass at enabling the functionality.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -509,6 +513,26 @@ def get_domain_records(
         if TextClause and isinstance(selectable, TextClause):
             selectable = selectable.columns().subquery()
 
+        filter_conditions: List[RowCondition] = domain_kwargs.get(
+            "filter_conditions", []
+        )
+        # For SqlAlchemyExecutionEngine only one filter condition is allowed
+        if len(filter_conditions) == 1:
+            filter_condition = filter_conditions[0]
```
alexsherstinsky: @anthonyburdi So `filter_conditions` will now absorb `row_condition`?  If so, what about backward compatibility?  Thanks.

great_expectations/execution_engine/sparkdf_execution_engine.py
cdkini: +1

I assumed it was the same variable but only caught it after Alex's comment. Maybe just `conditions`?

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -509,6 +513,26 @@ def get_domain_records(
...
+
+            selectable = (
+                sa.select([sa.text("*")])
+                .select_from(selectable)
+                .where(parse_condition_to_sqlalchemy(filter_condition.condition))
+            )
+        elif len(filter_conditions) > 1:
+            raise GreatExpectationsError(
+                "SqlAlchemyExecutionEngine currently only supports a single filter condition."
+            )
```
cdkini: If we only support a single condition, is there a reason we have a list here?

great_expectations/expectations/row_conditions.py
```diff
@@ -56,6 +61,43 @@ class ConditionParserError(GreatExpectationsError):
     pass
 
 
+class RowConditionParserType(enum.Enum):
+    GE = "ge"  # GE intermediate language
+    SPARK = "spark"
+    SPARK_SQL = "spark_sql"
+    PANDAS = "pandas"
+    PYTHON = "python"
+    SQL = "sql"
```
cdkini: ❤️

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -509,6 +513,26 @@ def get_domain_records(
...
+        )
+        # For SqlAlchemyExecutionEngine only one filter condition is allowed
+        if len(filter_conditions) == 1:
+            filter_condition = filter_conditions[0]
+            assert (
+                filter_condition.type_ == RowConditionParserType.GE
+            ), "filter_condition must be of type GE for SqlAlchemyExecutionEngine"
+
+            selectable = (
+                sa.select([sa.text("*")])
```
alexsherstinsky: @anthonyburdi a couple of things here: a) Do we want a type hint for selectable here?  And b) There might be a way to avoid `sa.text("*")` and use the more `SQLAlchemy`-like way -- if we want to pursue it, happy to discuss.

great_expectations/expectations/row_conditions.py
```diff
@@ -56,6 +61,43 @@ class ConditionParserError(GreatExpectationsError):
...
+    @property
+    def type_(self):
+        return self._type_
+
+    def __eq__(self, other: "RowCondition"):
+        if not isinstance(other, self.__class__):
+            raise GreatExpectationsTypeError(
+                "Only comparison of the same type of object is supported."
+            )
+        return all([self.condition == other.condition, self.type_ == other.type_])
```
cdkini: Could this be a prime candidate for a dataclass? We get `__eq__` for free then right?

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -509,6 +513,26 @@ def get_domain_records(
...
+                .select_from(selectable)
+                .where(parse_condition_to_sqlalchemy(filter_condition.condition))
+            )
+        elif len(filter_conditions) > 1:
+            raise GreatExpectationsError(
+                "SqlAlchemyExecutionEngine currently only supports a single filter condition."
+            )
+
```
alexsherstinsky: @anthonyburdi I see the `row_condition` now; the question is: how come it is not the first filter (like it is currently)?  Thanks.

cdkini (commented):
lovely! some minor comments but nothing blocking!

great_expectations/expectations/row_conditions.py
```diff
@@ -10,7 +12,10 @@
     alphas,
 )
 
-from great_expectations.exceptions import GreatExpectationsError
+from great_expectations.exceptions import (
```
alexsherstinsky: @anthonyburdi Minor: my personal preference is to import all exceptions as `ge_exceptions`. :)

great_expectations/expectations/row_conditions.py
```diff
@@ -56,6 +61,43 @@ class ConditionParserError(GreatExpectationsError):
     pass
 
 
+class RowConditionParserType(enum.Enum):
+    GE = "ge"  # GE intermediate language
+    SPARK = "spark"
+    SPARK_SQL = "spark_sql"
```
alexsherstinsky: @anthonyburdi Do we support both `SPARK` and `SPARK_SQL`, or is one of them a placeholder for the future?  Thanks.

tests/execution_engine/test_sparkdf_execution_engine.py
```diff
@@ -184,6 +188,153 @@ def test_get_domain_records_with_column_domain(
...
+        pd_df.columns.tolist(),
+    )
+    engine = basic_spark_df_execution_engine
+    engine.load_batch_data(batch_id="1234", batch_data=df)
+    data = engine.get_domain_records(
+        domain_kwargs={
+            "column": "a",
+            "row_condition": 'col("b")<5',
+            "condition_parser": "great_expectations__experimental__",
+            "filter_conditions": [
```
alexsherstinsky, line 216: @anthonyburdi I can see this programmatic way of specifying the filter conditions; how would a user specify them?  Is there a text/dictionary-only syntax?  Thanks.

alexsherstinsky (changes requested):
@anthonyburdi I entered a few questions and remarks -- thanks!

great_expectations/execution_engine/execution_engine.py
anthonyburdi: It shouldn't matter but happy to swap them!

alexsherstinsky: @anthonyburdi I feel that swapping has some interpretive advantages, because of thinking of "coarse" (e.g., entire row level) first, specific ones (e.g., column) second.  Thanks!

great_expectations/execution_engine/sparkdf_execution_engine.py
anthonyburdi: Good call - will change 👍

great_expectations/execution_engine/sparkdf_execution_engine.py
anthonyburdi: Yep! Good call - `all()` takes an iterable.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: Good point - in the future I'm hoping to pass a list of RowCondition objects to the execution engine to unify the interface across all execution engines. Figure this was a good starting step, but happy to remove if too forward looking.

great_expectations/expectations/row_conditions.py
anthonyburdi: Good point, I went back and forth on this. At one point we were discussing the RowCondition having it's own logic for conversion but decided against it. I'll revert to dataclass.

great_expectations/expectations/row_conditions.py
anthonyburdi: Good catch - didn't mean to break the pattern.

great_expectations/expectations/row_conditions.py
anthonyburdi: Currently `SPARK` is a placeholder for the future, it would refer to the `pyspark.sql.Column` type.

tests/execution_engine/test_sparkdf_execution_engine.py
anthonyburdi: Great question! These are only specified via using the engine's `add_column_row_condition` method, which is accessed via expectation params or set to defaults internally. A user cannot specify them using yaml for example, the interface for user input row conditions is the `row_conditions` and `condition_parser` parameters.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: Row conditions are still processed, and I moved them now to above the `filter_conditions` as discussed.

great_expectations/execution_engine/execution_engine.py
anthonyburdi: Done!

great_expectations/execution_engine/execution_engine.py
anthonyburdi: I just found it for this PR - super useful!

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -1,4 +1,3 @@
-import copy
```
anthonyburdi, line 1: This is just me cleaning up an unused import.

great_expectations/execution_engine/sparkdf_execution_engine.py
anthonyburdi: Done!

great_expectations/execution_engine/sparkdf_execution_engine.py
anthonyburdi: It stays in `get_domain_records`, and moved above `filter_conditions` handling as discussed.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: a) good call, moved it up to the beginning of the method though since it is used in many places (this method could be split in a future PR).
b) Interesting - is it using the sa.Table object? Or something like `selectable = sa.select(selectable).where(parsed_condition)`?

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: Updated & now it is the first filter per your note above :)

great_expectations/expectations/row_conditions.py
anthonyburdi: Actually let me check with @donaldheppner - I think there is a reason I'm forgetting. If not happy to revert.

anthonyburdi: Reverted after discussion - we can always change from a dataclass if we add methods in the future.

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -480,6 +493,29 @@ def get_domain_records(
 
         return data
 
+    def _combine_row_conditions(
```
alexsherstinsky, line 496: @anthonyburdi is this method static?

anthonyburdi: Effectively yes, but it's called only within the context of the instance so I've left it an instance method.

great_expectations/expectations/row_conditions.py
alexsherstinsky: @anthonyburdi If there is no comment about this, can we please add it?  Thanks!

alexsherstinsky (approved):
@anthonyburdi LGTM -- pending answers to a couple of questions/suggestions and/or affecting the response in the code as you deem appropriate.  Thanks!

tests/execution_engine/test_execution_engine.py
```diff
@@ -1,62 +1,158 @@
...
+    and use that fixture in tests.
+    """
+
+    class TestExecutionEngine(ExecutionEngine):
+        def get_batch_data_and_markers(
+            self, batch_spec
+        ) -> Tuple[BatchData, BatchMarkers]:
+            raise NotImplementedError
+
+    return TestExecutionEngine()
```
cdkini, line 35: *chefs kiss* 👨🏽‍🍳

tests/execution_engine/test_execution_engine.py
```diff
@@ -1,62 +1,158 @@
...
 # Edge cases
-def test_add_column_row_condition_with_unsupported_conditions():
-    e = PandasExecutionEngine()
+def test_add_column_row_condition_with_unsupported_conditions(test_execution_engine):
+    e = test_execution_engine
 
     # Ensuring that an attempt to filter nans within base class yields an error
-    with pytest.raises(GreatExpectationsError) as error:
+    with pytest.raises(ge_exceptions.GreatExpectationsError) as error:
         new_domain_kwargs = e.add_column_row_condition({}, "a", filter_nan=True)
```
cdkini: When we use context managers around errors and name them with `as`, can we add an additional assert around their error message?

```python
with pytest.raises(...) as error:
    ...

assert "my error message" in str(error.value)
```

This is only relevant if we have a consistent error message. Otherwise, I don't think we even need to name the error since it is unused.

tests/execution_engine/test_sparkdf_execution_engine.py
```diff
@@ -184,6 +188,157 @@ def test_get_domain_records_with_column_domain(
...
+    expected_column_df = spark_session.createDataFrame(
+        [
+            tuple(
+                None if isinstance(x, (float, int)) and np.isnan(x) else x
+                for x in record.tolist()
+            )
+            for record in expected_column_pd_df.to_records(index=False)
+        ],
+        expected_column_pd_df.columns.tolist(),
+    )
```
cdkini: Nitpick - seems like this is used a few times. I'm totally cool with verbosity to illustrate how it all works but could this be placed into a fixture or lambda? I think we can leave it as is but just thinking out loud.

great_expectations/expectations/row_conditions.py
anthonyburdi: Added!

tests/execution_engine/test_execution_engine.py
anthonyburdi: Good call, done!

tests/execution_engine/test_sparkdf_execution_engine.py
anthonyburdi: Agreed! This is used quite a lot! A good refactor. Changing now.

tests/execution_engine/test_sparkdf_execution_engine.py
```diff
@@ -100,6 +104,35 @@ def generate_ascending_list_of_datetimes(
...
+@pytest.fixture
+def spark_df_from_pandas_df():
+    """
+    Construct a spark dataframe from pandas dataframe.
+    Returns:
+        Function that can be used in your test e.g.:
+        spark_df = spark_df_from_pandas_df(spark_session, pandas_df)
+    """
+
+    def _construct_spark_df_from_pandas(
```
anthonyburdi, line 116: For you @cdkini :)

great_expectations/execution_engine/sparkdf_execution_engine.py
alexsherstinsky: @anthonyburdi Doesn't PyCharm give us warnings?  Perhaps we can turn it off, because the next person will almost certainly add the "@staticmethod" decorator, thereby contravening your rationale.  I feel like something has to be done.  Thanks.

donaldheppner (approved):

tests/execution_engine/test_execution_engine.py
anthonyburdi: Don's idea!

PR #4491, opened 2022-03-21
[FEATURE] Update diagnostic checklist to do linting checks

kenwade4:
- Update diagnostic checklist to do linting checks
- Make linting check be a part of experimental checks instead of production checks
- Also move camel_to_snake func to great_expectations/util.py

### Definition of Done

Thank you for submitting!

### DIFF
--- a/great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
+++ b/great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
@@ -1,3 +1,6 @@
+import inspect
+import os.path
+import re
 from collections import defaultdict
 from dataclasses import asdict, dataclass
 from typing import List, Tuple
@@ -18,7 +21,21 @@ from great_expectations.core.expectation_diagnostics.supporting_types import (
     ExpectationTestDiagnostics,
 )
 from great_expectations.core.util import convert_to_json_serializable
+from great_expectations.expectations.registry import get_expectation_impl
 from great_expectations.types import SerializableDictDot
+from great_expectations.util import camel_to_snake, lint_code
+
+try:
+    import black
+except ImportError:
+    black = None
+
+try:
+    import isort
+except ImportError:
+    isort = None
+
+rx_expectation_instance_repr = re.compile(r"<.*\.([^\.]*) object at .*")
 
 # from pydantic.dataclasses import dataclass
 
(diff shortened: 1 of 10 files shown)


great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
```diff
@@ -1,3 +1,6 @@
+import inspect
+import os.path
```
cdkini: Nitpick - could we just import `os` to be consistent with the rest of the codebase?

great_expectations/expectations/expectation.py
```diff
@@ -1546,6 +1536,7 @@ def _get_maturity_checklist(
                 tests
             )
         )
+        experimental_checks.append(ExpectationDiagnostics._check_linting(self))
```
cdkini, line 1539: Should `check_linting` be a public method? I feel little iffy invoking a private method if we can avoid it.

great_expectations/util.py
```diff
@@ -95,6 +95,9 @@
     "rendered_data_docs": "rendered_data_doc",
 }
 
+p1 = re.compile(r"(.)([A-Z][a-z]+)")
+p2 = re.compile(r"([a-z0-9])([A-Z])")
```
cdkini, line 99: Could we give these more descriptive names? What do `p1` and `p2` do here?

great_expectations/util.py
```diff
@@ -122,6 +125,11 @@ def singularize(plural_ge_noun):
         )
 
 
+def camel_to_snake(name):
+    name = p1.sub(r"\1_\2", name)
+    return p2.sub(r"\1_\2", name).lower()
```
cdkini, line 130: Could we type annotate the args and return here? Also, could we add a few unit tests? I trust your regex skills a lot more than I do mine but I think a few tests to gain coverage would be great!

great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
```diff
@@ -378,10 +395,73 @@ def _check_renderer_methods(
     @staticmethod
     def _check_linting(expectation_instance) -> ExpectationDiagnosticCheckMessage:
         """Check if linting checks pass for Expectation"""
-        # TODO: Perform linting checks instead of just giving thumbs down
+        sub_messages = []
+        passed = False
+        black_ok = False
+        isort_ok = False
+        file_and_class_names_ok = False
+
```
cdkini: Could we add type annotations to all variable declarations throughout?
```suggestion
        sub_messages: List[dict] = []
        passed: bool = False
        black_ok: bool = False
        isort_ok: bool = False
        file_and_class_names_ok: bool = False
```

great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
```diff
@@ -378,10 +395,73 @@ def _check_renderer_methods(
...
+        expectation_camel_name = rx_expectation_instance_repr.match(
+            repr(expectation_instance)
+        ).group(1)
```
cdkini: Will this ever fail? What if group doesn't have 1?

Nitpick - could we split up this logic into separate lines? Perhaps the match on one and the group on the other?

great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
```diff
@@ -378,10 +395,73 @@ def _check_renderer_methods(
...
+            isort_ok = isort.check_code(code, **isort.profiles.black)
+            if not isort_ok:
+                sub_messages.append(
+                    {
+                        "message": "Your code would be reformatted with isort",
+                        "passed": False,
+                    }
+                )
+
+        passed = black_ok and isort_ok and file_and_class_names_ok
```
cdkini, line 488: love it!

great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
```diff
@@ -18,7 +21,21 @@
...
+    import black
+except ImportError:
+    black = None
+
+try:
+    import isort
+except ImportError:
+    isort = None
+
+rx_expectation_instance_repr = re.compile(r"<.*\.([^\.]*) object at .*")
```
cdkini: If this regex is only used in the single method, could we perhaps do this within the function? I'd rather clean up the global namespace if at all possible, even if it means being slightly less performant.

great_expectations/util.py
cdkini: Oops now seeing you just ported this over - will defer to you here.

cdkini (approved):
LGTM - left comments around type annotations and general style that I'd love resolved but I don't want to block on that.

great_expectations/expectations/expectation.py
kenwade4: All the staticmethod/classmethod decorated methods in the ExpectationDiagnostics class start with `_check_`

great_expectations/util.py
kenwade4: Just moved as-is from great_expectations/expectations/expectation.py

great_expectations/util.py
kenwade4: Just moved as-is from great_expectations/expectations/expectation.py

PR #4510, opened 2022-03-22
[FEATURE] `YAMLHandler` Class

Shinnnyshinshin:
- This PR introduces a `YAMLHandler` class, which is designed to be a lightweight wrapper around YAML serialization.

- A large number of warnings in OSS GE came from `ruamel`, which started to make major changes around `dump()` and `load()` starting with version 0.15.0. More information can be found in the associated documentation here : [Departure from previous API](https://yaml.readthedocs.io/en/latest/api.html). 

```
# The old API
res = yaml.dump(data) # This is deprecated and will raise errors down the line

----------

# The new API
yaml = YAML(typ="safe")

out = io.StringIO() # Can be a file, stdout, or any other valid stream
yaml.dump(data, out)
res = out.getvalue()
```

- This PR allows for the new `ruamel` API to be used through the `YAMLHandler`, which greatly simplifies the code changes needed to get rid of these errors. 

```
YAMLHandler.load(my_yaml)
YAMLHandler.dump(my_dict)
```

- Also interesting : According to the `ruamel` documentation, outputting the results of `dump()` as a string without specifying a stream parameter is  [incorrect usage](https://yaml.readthedocs.io/en/latest/example.html#output-of-dump-as-a-string) and inefficient.  However, `YAMLHandler` will default to `StringIO()` when the stream is not specified, which preserves existing Great Expectations behavior. 

- GREAT-575

### Definition of Done

Thank you for submitting!


great_expectations/core/yaml_handler.py
```diff
@@ -0,0 +1,68 @@
...
+    def _dump(data: dict, stream, **kwargs) -> None:
+        """If an input stream has been provided, modify it in place."""
+        YAMLHandler._handler.dump(data=data, stream=stream, **kwargs)
+
+    @staticmethod
+    def _dump_and_return_value(data: dict, **kwargs) -> str:
+        """If an input stream hasn't been provided, generate one and return the value."""
+        stream = io.StringIO()
+        YAMLHandler._handler.dump(data=data, stream=stream, **kwargs)
+        return stream.getvalue()
```
cdkini: I cleaned this up a little but let me know if you want me to revert it to the example noted in the docs.

great_expectations/cli/checkpoint.py
```diff
@@ -3,7 +3,6 @@
 from typing import List
 
 import click
-from ruamel.yaml import YAML
```
cdkini, line 6: Unused in the file but imported and instantiated. All such instances have been deleted.

great_expectations/cli/datasource.py
```diff
@@ -319,7 +319,7 @@ def yaml_snippet(self) -> str:
     class_name: InferredAssetFilesystemDataConnector
     base_directory: {self.base_path}
     default_regex:
-      group_names: 
+      group_names:
```
cdkini, line 322: My editor likes to remove whitespace on save, apologies for the confusion!

tests/core/test_yaml_handler.py
```diff
@@ -0,0 +1,71 @@
+import os.path
+import sys
+from pathlib import Path
+from typing import AnyStr, Optional
+
+import pytest
+from _pytest.capture import CaptureResult
```
cdkini: Since this is the internal API, what are the chances this breaks? I love the diligence around type annotations but I'd probably just slap an `Any` on this and not feel too bad.

tests/core/test_yaml_handler.py
```diff
@@ -0,0 +1,71 @@
...
+    simplest_dict: dict = dict(abc=1)
+    dumped: Optional[str] = YAMLHandler.dump(simplest_dict)
+    assert dumped == "abc: 1\n"
+
+
+def test_dump_stdout_specified(capsys) -> None:
+    # ruamel documentation recommends that we specify the stream as stdout when we are using YAML to return a string.
+    simplest_dict: dict = dict(abc=1)
+    YAMLHandler.dump(simplest_dict, stream=sys.stdout)
+    captured: CaptureResult[AnyStr] = capsys.readouterr()
```
cdkini: Love it!

great_expectations/cli/datasource.py
Shinnnyshinshin: ah I think there were 2 tests that were failiing as a result of this. They've been updated

tests/core/test_yaml_handler.py
Shinnnyshinshin: works for me :)  That's a good point about it being part of an internal API

great_expectations/core/yaml_handler.py
```diff
@@ -0,0 +1,69 @@
+import io
+from pathlib import Path
+from typing import Optional, Union
+
+from ruamel.yaml import YAML
+
+
+class YAMLHandler:
```
donaldheppner, line 8: I'd rather this not be a static class, so that we can declare it as a dependency and so that it is easier to mock

donaldheppner: for now, we can initialize the instance in the constructor of the objects that need it

donaldheppner (changes requested):
per comments, yaml handler should not be static

great_expectations/core/yaml_handler.py
```diff
@@ -0,0 +1,69 @@
+import io
```
anthonyburdi, line 1: Love this. I saw in our new style guide an example of a module docstring with "Typical Usage Example:" https://google.github.io/styleguide/pyguide.html#382-modules - does that seem appropriate here?

anthonyburdi (commented):
Looks great pending resolution of Don's comments

great_expectations/core/yaml_handler.py
Shinnnyshinshin: This is a great suggestion :) Thank you @anthonyburdi

great_expectations/core/yaml_handler.py
Shinnnyshinshin: Totally makes sense :) Will update the PR now

Shinnnyshinshin: @donaldheppner could you explain this part a little bit more? 

"we can initialize the instance in the constructor of the objects that need it"?

cdkini: I think what Don is alluding to here is the idea of dependency injection - while I could instantiate a `YAMLHandler` within my constructor, it makes a lot more sense to do it outside and "inject" it in when instantiating the given object.

This allows my class a lot of flexibility - because I'm passing things into the constructor, I'm decoupling the dependencies between the handler and the class that's using it.

```python
class MyClass:
    def __init__(self):
        self._handler = YAMLHandler()

class MyClass:
    def __init__(self, handler):
        self._handler = handler

"""
The former forces me to use YAMLHandler no matter what but the latter let's me inject whatever I want.
Ex: MyClass(handler=MagicMock()) in a unit tests so I'm only testing the behavior of MyClass and separating it from YAMLHandler.
"""
```

but @donaldheppner let me know if I'm missing anything.

Shinnnyshinshin: Got it got it.

donaldheppner (approved):

PR #4516, opened 2022-03-23
[BUGFIX] Account for case where SQLAlchemy dialect is not downloaded when masking a given URL

cdkini:
- Our current conditional checks for SQLAlchemy being installed but does not look for specific dialects; this causes a failure within SQLAlchemy's URL generation code.
- My proposed solution here is if we happen to hit that error, default to a non-SQLAlchemy style of obfuscation.

### Definition of Done


donaldheppner (approved):

great_expectations/data_context/util.py
```diff
@@ -542,39 +544,46 @@ def mask_db_url(cls, url: str, use_urlparse: bool = False, **kwargs) -> str:
...
-
-            # Do not parse sqlite
-            if parsed_url.scheme == "sqlite":
-                return url
+    @classmethod
+    def _mask_db_url_no_sa(cls, url: str) -> str:
+        # oracle+cx_oracle does not parse well using urlparse, parse as oracle then swap back
+        replace_prefix = None
+        if url.startswith("oracle+cx_oracle"):
+            replace_prefix = {"original": "oracle+cx_oracle", "temporary": "oracle"}
```
kenwade4, line 564: Do you know if there is the same issue with `mysql+pymysql://`? Doesn't matter for now since this looks like code ported as-is to a new func

kenwade4 (approved):

great_expectations/data_context/util.py
cdkini: I'm not sure! This was introduced in a change made yesterday and only fails in a weird edge case. Don't have a lot of familiarity here just patching the build.

Let me know if you think anything should be changed (going to merge as is for now just to unblock people).

joshua-stauffer (approved):
thanks @cdkini!

PR #4517, opened 2022-03-23
[MAINTENANCE] Add prod metadata and remove package attribute from library_metadata

kenwade4:
This will allow many core Expectations to have every single diagnostic checklist item verified in the gallery!

- Remove package attribute from all Expectation's library_metadata and add prod metadata to many core Expectations
- Update _check_linting to ignore whitespace an recognize local folder great_expectations with isort
- Rename expect_column_pair_values_a_to_be_greater_than_b.json
- Add basic validate_configuration method to core Expectations that pass everything but input validation

### Definition of Done


rdodev:
My big concern here is we're adding metadata to the expectations themselves. This may have been a long-ago decision, but it would seem to me that it would be far better to decouple metadata from code, especially since the point is to drive the gallery. Would it not be better to have a one-time or manual script to scrape the expectations directory(ies) and generate a metadata file and let that be the source of truth for the gallery?

azure-pipelines.yml
```diff
@@ -390,7 +390,7 @@ stages:
             displayName: 'pytest'
 
   - stage: deploy_gallery
-    condition: and(succeeded(), eq(variables.isMain, true))
+    # condition: and(succeeded(), eq(variables.isMain, true))
```
cdkini: Please revert before merging.

great_expectations/expectations/core/expect_column_values_to_be_null.py
```diff
@@ -62,15 +62,34 @@ class ExpectColumnValuesToBeNull(ColumnMapExpectation):
...
+        Args:
+            configuration (OPTIONAL[ExpectationConfiguration]): \
+                An optional Expectation Configuration entry that will be used to configure the expectation
+        Returns:
+            True if the configuration has been validated successfully. Otherwise, raises an exception
+        """
+        super().validate_configuration(configuration)
+        self.validate_metric_value_between_configuration(configuration=configuration)
+
+        return True
```
cdkini, line 91: The other validate methods don't return anything right? Just simply raise if something breaks?

I think we should keep that consistent since `bool` here denotes we can possible return `False` (but I don't think that's possible).

great_expectations/expectations/core/expect_column_values_to_not_be_null.py
```diff
@@ -64,17 +64,36 @@ class ExpectColumnValuesToNotBeNull(ColumnMapExpectation):
...
+        "has_full_test_suite": True,
+        "manually_reviewed_code": True,
     }
 
     map_metric = "column_values.nonnull"
     args_keys = ("column",)
 
+    def validate_configuration(
+        self, configuration: Optional[ExpectationConfiguration]
+    ) -> bool:
```
cdkini, line 81: Same note here and throughout!

great_expectations/expectations/core/expect_column_values_to_not_be_null.py
```diff
@@ -64,17 +64,36 @@ class ExpectColumnValuesToNotBeNull(ColumnMapExpectation):
...
+        """
+        Validates that a configuration has been set, and sets a configuration if it has yet to be set. Ensures that
+        necessary configuration arguments have been provided for the validation of the expectation.
+
+        Args:
+            configuration (OPTIONAL[ExpectationConfiguration]): \
+                An optional Expectation Configuration entry that will be used to configure the expectation
+        Returns:
+            True if the configuration has been validated successfully. Otherwise, raises an exception
```
cdkini, line 90: Really nice docstrings 🙇🏽

cdkini (commented):
Quick comments regarding validate method and reverting changes to pipeline but will accept after. Please ping me when I should look again!

great_expectations/expectations/core/expect_column_values_to_be_null.py
kenwade4: All the validate_configuration methods either raise an exception or return True. The validate_configuration methods in this PR were copy/pasted.

great_expectations/expectations/core/expect_column_values_to_not_be_null.py
kenwade4: Not taking credit for that since they were copy/pasted haha

great_expectations/expectations/core/expect_column_values_to_be_null.py
cdkini: Ah okay fair enough. Disregard this then.

great_expectations/expectations/core/expect_column_values_to_not_be_null.py
cdkini: aha didn't read closely enough - hey kudos for porting everything over so nicely 👀

cdkini (approved):

kenwade4:
@rdodev the metadata for contributed Expectations also includes things like who the contributors are and the tags they specify. You can bring up the issue of including metadata in Expectation defs at the next arch review or triage meeting

PR #4546, opened 2022-03-24
[MAINTENANCE] Package dependencies usage stats instrumentation - part 1

anthonyburdi:
- Add utilities to identify installed package dependencies to help us know when to upgrade GE required and dev packages.
- Part 2 of this effort will be the instrumentation and schema changes in a separate PR.

### Definition of Done

Thank you for submitting!


great_expectations/core/usage_statistics/execution_environment.py
```diff
@@ -0,0 +1,125 @@
...
+
+This module contains utilities for identifying installed package dependencies
+to help enable the core team to safely upgrade package dependencies to gain
+access to features of new package versions.
+
+    Typical usage example:
+        ge_execution_environment: GEExecutionEnvironment = GEExecutionEnvironment()
+        dependencies: List[PackageInfo] = ge_execution_environment.dependencies
+
+"""
```
cdkini, line 11: Truly a thing of beauty 😍 

Really loving this adherence to the style guide.

great_expectations/core/usage_statistics/execution_environment.py
```diff
@@ -0,0 +1,125 @@
...
+    DEV = "dev"
+    REQUIRED = "required"
+
+
+@dataclass
+class PackageInfo:
+    package_name: str
+    installed: bool
+    install_environment: InstallEnvironment
+    version: Optional[version.Version]
```
cdkini, line 41: Should this be set to a default value of `None`?

great_expectations/core/usage_statistics/execution_environment.py
```diff
@@ -0,0 +1,125 @@
...
+        """Get all installed packages. Used to check if dependency is installed or not.
+
+        Returns:
+            list of string names of packages that are installed
+        """
+        if self._all_installed_packages is None:
+            # Only retrieve once
+            self._all_installed_packages = [
+                item.metadata.get("Name") for item in metadata.distributions()
+            ]
```
cdkini: Is "Name" guaranteed? If not, should we perhaps filter out nulls before returning?

great_expectations/core/usage_statistics/execution_environment.py
```diff
@@ -0,0 +1,125 @@
...
+        Returns:
+            List of PackageInfo objects for each dependency including version,
+            install status, install environment.
+        """
+        dependencies: List[PackageInfo] = []
+        for dependency_name in dependency_names:
+
+            package_version: Optional[version.Version]
+            installed: bool
+            if dependency_name in self._get_all_installed_packages():
```
cdkini, line 120: will we ever run this method without having first run `self._get_all_installed_packages` in the constructor? If not, could we just access `self._all_installed_packages`?

great_expectations/core/usage_statistics/execution_environment.py
cdkini: I see we're setting a value even if `None` so this is perfect!

great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -0,0 +1,229 @@
...
+class GEDependencies:
+    """Store and provide dependencies when requested.
+
+    Also acts as a utility to check stored dependencies match our
+    library requirements.
+
+    Attributes: None
+    """
+
+    """This list should be kept in sync with our requirements.txt file."""
```
cdkini, line 32: Is there any way we can enforce this? Maybe parsing these deps and the deps in the requirements and making sure all the items noted herein are reflected in the files?

great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -0,0 +1,229 @@
...
+        "scipy",
+        "termcolor",
+        "tqdm",
+        "typing-extensions",
+        "urllib3",
+        "tzlocal",
+    ]
+
+    """This list should be kept in sync with our requirements-dev*.txt files."""
+    GE_DEV_DEPENDENCIES: List[str] = [
```
cdkini: How do we determine what is important to track? Could an argument be made to gather as much information as possible and put the burden of parsing/munging on analytics?

I know we have concerns about PII but what if we parsed all our requirements files and then only provided information about those libraries? That would allow us a lot of flexibility, even if the requirements rapidly changed.

great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -0,0 +1,229 @@
...
+        "sqlalchemy-bigquery",
+        "sqlalchemy-dremio",
+        "sqlalchemy-redshift",
+        "teradatasqlalchemy",
+        "xlrd",
+    ]
+
+    def __init__(self):
+        self._requirements_relative_base_dir = "../../../"
+        self._dev_requirements_prefix: str = "requirements-dev"
```
cdkini: Could `file_relative_path` come in handy here?

great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -0,0 +1,229 @@
...
+        """
+        return sorted(
+            list(
+                set(
+                    self._get_dependency_names_from_requirements_file(
+                        self.required_requirements_path
+                    )
+                )
+            )
+        )
```
cdkini: Nitpick - `sorted` already generates a list from an iterable so I think we can get rid of that right?

great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -0,0 +1,229 @@
...
+        self._dev_requirements_prefix: str = "requirements-dev"
+
+    def get_required_dependency_names(self) -> List[str]:
+        """Sorted list of required GE dependencies"""
+        return sorted(self.GE_REQUIRED_DEPENDENCIES)
+
+    def get_dev_dependency_names(self) -> List[str]:
+        """Sorted list of dev GE dependencies"""
+        return sorted(self.GE_DEV_DEPENDENCIES)
+
```
cdkini: Aren't these already sorted?

great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -0,0 +1,229 @@
...
+        for dev_dependency_filename in self.dev_requirements_paths:
+            dependency_names: List[
+                str
+            ] = self._get_dependency_names_from_requirements_file(
+                os.path.join(
+                    self._requirements_relative_base_dir, dev_dependency_filename
+                )
+            )
+            dev_dependency_names.update(dependency_names)
+        return sorted(list(dev_dependency_names))
```
cdkini: Same nit here - not a blocker but if decide to update, please apply throughout 🙇🏽

tests/core/usage_statistics/test_execution_environment.py
```diff
@@ -0,0 +1,105 @@
...
+        "dev-package-2",
+    ]
+
+    ge_execution_environment: GEExecutionEnvironment = GEExecutionEnvironment()
+    expected_dependencies: List[PackageInfo] = [
+        PackageInfo(
+            package_name="req-package-1",
+            installed=True,
+            install_environment=InstallEnvironment.REQUIRED,
+            version=version.Version("8.8.8"),
```
cdkini: Are we sure this works again Git hashes or other edge cases like `-e` installation? Could we perhaps add some items to test that?

tests/core/usage_statistics/test_package_dependencies.py
```diff
@@ -0,0 +1,78 @@
+from typing import List
+
+import pytest
+
+from great_expectations.core.usage_statistics.package_dependencies import GEDependencies
+
+
+def test__get_dependency_names():
+    """Test that the regex in _get_dependency_names() parses a requirements file correctly."""
```
cdkini, line 9: https://xkcd.com/1171/

great_expectations/core/usage_statistics/package_dependencies.py
cdkini: With all the data, we can always backtrack and say "oh hey that library is actually more important to track than we thought"

cdkini: I'm now realizing this is an executable script and you're actually verifying against the actual files - I'm still going to keep these comments here but feel free to resolve if they aren't applicable.

donaldheppner: I don't think we need to track dev dependencies; certainly not wholesale. I would be interested in tracking packages that are used to integrate in specific environments though. Ken did some work to break these out, maybe they would be useful here.

great_expectations/core/usage_statistics/package_dependencies.py
donaldheppner: if we need them sorted, we should do it at declaration, not every time the property is invoked

donaldheppner (changes requested):

great_expectations/core/usage_statistics/execution_environment.py
anthonyburdi: Good call though, this might not be the only caller. Added the default. Thanks for the catch!

anthonyburdi: Actually I'm getting errors when importing in tests when using this default. Will have to think on it a bit more - leaving as-is for now.

great_expectations/core/usage_statistics/execution_environment.py
anthonyburdi: It is from the output of `metadata.distributions()` but better safe than sorry, added to the list comp.

great_expectations/core/usage_statistics/execution_environment.py
anthonyburdi: We wouldn't but I was just hoping to have a single access point for the installed packages which enforces the singleton-like behavior (only retrieve once otherwise return cached value)

great_expectations/core/usage_statistics/package_dependencies.py
anthonyburdi: Yep! These are enforced in CI through the `test_required_dependency_names_match_requirements_file` and `test_dev_dependency_names_match_requirements_file` integration tests. If there is a mismatch between these lists and the requirements files at test time then we fail the test. The utility in the package_dependencies file is to help identify missing packages, we can improve it to highlight differences if it becomes burdensome.

great_expectations/core/usage_statistics/package_dependencies.py
anthonyburdi: Great points @cdkini . We don't want to track the full environment since we only want to know what version of a GE dependency is installed so that we can adjust our requirements bounds, and we don't want to inadvertently track a non-public package for example that pierces anonymization. And correct - we are verifying against the files in CI and via this utility so hopefully it is easy and enforced to keep these lists in sync. Very open to changing this if there is a better way that does not require manual update but I have not yet found one.

anthonyburdi: Very good point @donaldheppner. Those dependencies for specific environments are captured here since they fit the `requirements-dev-*.txt` naming pattern e.g. `requirements-dev-mysql.txt`. If we want to separate dev vs specific environment requirements via a different naming pattern it would be simple to update this script. cc @kenwade4 
Also we are pretty permissive with our dev requirements, it may be good to track these as well so that we can update the bounds to more recent versions.

great_expectations/core/usage_statistics/package_dependencies.py
anthonyburdi: Possibly - I did want to hard code this path specifically though. How would you suggest using it?

great_expectations/core/usage_statistics/package_dependencies.py
anthonyburdi: Good catch, done!

great_expectations/core/usage_statistics/package_dependencies.py
anthonyburdi: They are but manually sorted by hand. I wanted to make this a bit more lenient to someone inserting a package in the wrong order for internal comparison, especially since the case handling of sorted() and requirements.txt are different (sorted puts caps ahead: ```p = ["a", "Ba", "b", "c"]
sorted(p)
Out[8]: ['Ba', 'a', 'b', 'c']```

great_expectations/core/usage_statistics/package_dependencies.py
anthonyburdi: Done!

tests/core/usage_statistics/test_package_dependencies.py
anthonyburdi: Exactly :)

tests/core/usage_statistics/test_execution_environment.py
anthonyburdi: version.parse() should handle those cases, but good call on adding some test cases.

great_expectations/core/usage_statistics/package_dependencies.py
anthonyburdi: Good point about sorting at declaration, I had it that way in an earlier version. Changing now.

tests/core/usage_statistics/test_execution_environment.py
anthonyburdi: Added!

great_expectations/core/usage_statistics/execution_environment.py
cdkini: I'm totally cool with this, thanks for the explanation! Feel free to resolve.

great_expectations/core/usage_statistics/package_dependencies.py
donaldheppner: I'd like to remove the path reference here, especially since the file reference is parameratized in most methods. The path to the file can be specified in main, either as a method or constructor parameter, not hardcode in this class

cdkini: I think the usage is this: `file_relative_path(__file__, "../../../requirements-dev")`

That being said, what you have seems to work fine! Will defer to you here.

cdkini (approved):
LGTM!

great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -0,0 +1,231 @@
...
+            "termcolor",
+            "tqdm",
+            "typing-extensions",
+            "urllib3",
+            "tzlocal",
+        ]
+    )
+
+    """This list should be kept in sync with our requirements-dev*.txt files."""
+    GE_DEV_DEPENDENCIES: List[str] = sorted(
```
donaldheppner: I don't think we need to track dev dependencies

donaldheppner (changes requested):

great_expectations/core/usage_statistics/package_dependencies.py
anthonyburdi: Done - moved to constructor with default and added use of `file_relative_path()`

great_expectations/core/usage_statistics/package_dependencies.py
anthonyburdi: Updated to exclude dependencies that are only relevant in dev environments. In a future refactor, we discussed programatically creating requirements files via a script to have more control here but decided it is out of scope of the current work.

donaldheppner (approved):

great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Also acts as a utility to check stored dependencies match our
+    library requirements.
+
+    Attributes: None
+    """
+
+    """This list should be kept in sync with our requirements.txt file."""
+    GE_REQUIRED_DEPENDENCIES: List[str] = sorted(
+        [
```
kenwade4, line 34: Can the contents of these lists read in their data from the requirements files similarly to https://github.com/great-expectations/great_expectations/blob/develop/tests/test_packaging.py

kenwade4:
The `requirements-parser` package, which is already a part of requirements-dev-lite.txt file does this kind of stuff.

See https://github.com/great-expectations/great_expectations/blob/develop/tests/test_packaging.py#L4-L22

Pretty sure it strips out in-line comments and loads in other requirements referenced if there are any `--requirement blah.txt` lines in the requirements file

great_expectations/core/usage_statistics/package_dependencies.py
anthonyburdi: Good call - I tried that first but the requirements files are not available in the packaged version at runtime. These lists take their place at runtime, and hopefully the tests + utility help us keep these lists up to date. Very open to ideas though if there is a better way.

anthonyburdi:
> The `requirements-parser` package, which is already a part of requirements-dev-lite.txt file does this kind of stuff.
> 
> See https://github.com/great-expectations/great_expectations/blob/develop/tests/test_packaging.py#L4-L22
> 
> Pretty sure it strips out in-line comments and loads in other requirements referenced if there are any `--requirement blah.txt` lines in the requirements file

Good point - found that earlier when reviewing our dependencies for the list. Would you mind if that is in a future refactor PR?

PR #4582, opened 2022-03-29
[MAINTENANCE] Better NaN handling in tests

kenwade4:
(This should be merged to develop right after maintenance/great-640/great-684/eol-python36 gets merged in)

- Add support for testing lists that may have NaN in them

### Definition of Done


great_expectations/util.py
```diff
@@ -1295,14 +1296,64 @@ def is_nan(value: Any) -> bool:
...
+
+def compare_two_lists_or_items_that_might_have_nan(first: Any, second: Any) -> bool:
+    """Did you know that `float("nan") != float("nan")`?
+
+    Only 1 level of nesting is supported by this func
+    """
+    result = False
+    if first == second:
+        result = True
+    elif type(first) != list:
```
cdkini: Could we use `isinstance` here?

great_expectations/util.py
```diff
@@ -1295,14 +1296,64 @@ def is_nan(value: Any) -> bool:
...
+
+
+def compare_two_lists_or_items_that_might_have_nan(first: Any, second: Any) -> bool:
+    """Did you know that `float("nan") != float("nan")`?
+
+    Only 1 level of nesting is supported by this func
+    """
+    result = False
+    if first == second:
+        result = True
```
cdkini: Could we just return here? And at the end of each conditional branch? We don't need to evaluate the rest of the function of `result` gets set right?

great_expectations/util.py
```diff
@@ -1295,14 +1296,64 @@ def is_nan(value: Any) -> bool:
...
+    elif type(first) != list:
+        first_is_nan_or_none = _is_nan_or_none(first)
+        second_is_nan_or_none = _is_nan_or_none(second)
+        if first_is_nan_or_none and second_is_nan_or_none:
+            result = True
```
cdkini: Perhaps we could consolidate:
```python
elif not isinstance(first, list) and (_is_nan_or_none(first) and _is_nan_or_none(second)):
    return True
```

great_expectations/util.py
```diff
@@ -1295,14 +1296,64 @@ def is_nan(value: Any) -> bool:
...
+def _is_nan_or_none(scalar: Any) -> bool:
+    """Return True if scalar is NaN or None"""
+    try:
+        result = isnan(scalar)
+    except TypeError:
+        result = scalar == None
+    return result
+
+
+def compare_two_lists_or_items_that_might_have_nan(first: Any, second: Any) -> bool:
```
cdkini: Are these truly `Any`? Aren't they either NaN or lists?

great_expectations/util.py
```diff
@@ -1295,14 +1296,64 @@ def is_nan(value: Any) -> bool:
...
+                        pass
+                    else:
+                        first_is_nan_or_none = _is_nan_or_none(val)
+                        second_is_nan_or_none = _is_nan_or_none(second[i])
+                        if first_is_nan_or_none and second_is_nan_or_none:
+                            pass
+                        else:
+                            raise ValueError("These are not the same")
+            result = True
+        except ValueError:
```
cdkini: Could we make this whole thing recursive? Something like this?

```python
if type(first) != type(second):
    return False

if isinstance(first, list):
    return all(func(first=f, second=s) for f, s in zip(first, second)

return first == second    
```

great_expectations/util.py
cdkini: and throughout where applicable?

cdkini (changes requested):

great_expectations/util.py
cdkini: Here's a more fleshed out idea of what I'm thinking:

```python
def my_func(first: Any, second: Any) -> bool:
    if first == second:
        return True
    elif isinstance(first, list) and isinstance(second, list):
        return all(my_func(first=f, second=s) for f, s in zip(first, second)))
    else:
        return _is_nan_or_none(first) and _is_nan_or_none(second)

great_expectations/util.py
kenwade4: Other things can be (and are) passed to this func, but the NaN stuff will be done for simple values and lists

cdkini (approved):

PR #4557, opened 2022-03-26
[WIP] Demo of RBP Hero Experience 

Shinnnyshinshin:

PR #4590, opened 2022-03-29
[BUGFIX] Adjust output of datetime `ParameterBuilder` to match Expectation

Shinnnyshinshin:
- Follow-up to #4557 
- addresses issue where the `ParameterBuilder` for datetime was outputting a dictionary (containing successful string and candidate strings), when the `Expectation` was expecting a string (only the successful string). 
- Configuration has been adjusted so `strftime_format` is a string, and additional candidate strings are placed under `meta`

**Why weren't the existing tests catching this?**
- Fixes bug with `profiler_user_workflow` integration test where `sorted()` was being used to compare `ExpectationSuites`
- This sorting was unnecessary and has been removed. 
- Closes GREAT-754

### Definition of Done

Thank you for submitting!


tests/conftest.py
```diff
@@ -2838,6 +2838,8 @@ def bobby_columnar_table_multi_batch(empty_data_context):
                         "metric_configuration": {
                             "metric_name": "table.row_count",
                             "domain_kwargs": {},
+                            "metric_value_kwargs": None,
+                            "metric_dependencies": None,
```
Shinnnyshinshin, line 2842: These are parameter that are now outputted by the `profiler`. Was not caught by our previous test, because of how we were doing `sort()`

great_expectations/core/expectation_configuration.py
```diff
@@ -1268,6 +1268,12 @@ def __eq__(self, other):
             )
         )
 
+    def __lt__(self, other):
+        # Basic comparator for ExpectationType, so that ExpectationSuites can be compared
+        # in a deterministic way. Less-than is determined from Expectation-type
+        # TODO: ensure comparator takes into account kwargs and meta
+        return self.expectation_type < other.expectation_type
```
alexsherstinsky: @Shinnnyshinshin This worries me a little -- what is being compared if both are strings?  I am not sure what the meaning of comparison would be as measured in the units of string (lexicographical?) comparison of their `expectation_type` values.

great_expectations/core/expectation_suite.py
```diff
@@ -161,6 +161,11 @@ def __eq__(self, other):
         if not isinstance(other, self.__class__):
             # Delegate comparison to the other instance's __eq__.
             return NotImplemented
+
+        # sort expectations to be deterministic
+        self.expectations.sort()
```
alexsherstinsky: @Shinnnyshinshin I feel that we should use the `Python` `sorted()` because it does not modify the original list.  Do we really want to modify the `self.expectations` list?..

alexsherstinsky (changes requested):
@Shinnnyshinshin I had some concerns about comparison and sorting.  Even if those approaches are justified, we should probably add some illustrative tests for their use, I feel.  Thank you.

great_expectations/core/expectation_suite.py
Shinnnyshinshin: Great point. After discussion, we have determined that this sorting step is not necessary. PR has been updated to reflect this clearer understanding.

Shinnnyshinshin:
Thank you for the review @alexsherstinsky  After discussion, we have determined that this sorting step is not necessary. PR has been updated to reflect this clearer understanding.

alexsherstinsky (approved):
LGTM!

PR #4618, opened 2022-04-01
[BUGFIX] Correctly handle SQLA unexpected count metric for empty tables

### LINKED ISSUE #4603: SQLAlchemy `expect_column_values_to_be_unique` errors out when run on an empty table
**Describe the bug**
Using an `InferredAssetSqlDataConnector` with a SQLAlchemy connection to a Postgres DB and running `expect_column_values_to_be_unique` fails with a `TypeError` when run against an empty table. 

The error occurs in the `map_metric_provider` module:

```
  File "/Users/douglas/.virtualenvs/dags/lib/python3.7/site-packages/great_expectations/execution_engine/execution_engine.py", line 391, in resolve_metrics
    **metric_provider_kwargs
  File "/Users/douglas/.virtualenvs/dags/lib/python3.7/site-packages/great_expectations/expectations/metrics/map_metric_provider.py", line 1994, in _sqlalchemy_map_condition_unexpected_count_value
    unexpected_count = int(unexpected_count)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

It looks like this would happen with any DB, but I have only tested it with Postgres.

**To Reproduce**
1. Create a table in a Postgres DB without any rows populated
2. Configure an `InferredAssetSqlDataConnector` pointed to the DB
3. Add an `expect_column_values_to_be_unique` using the table
4. Run the expectation and the above error will be returned in the `exception_info`

**Expected behavior**
If the data asset is empty then the expectation should pass by default - it cannot contain any unexpected rows if there are no rows.

 - Operating System: MacOS
 - Python 3.7.13
 - Great Expectations version 0.14.11
 - SQLAlchemy version 1.4.9

douglascook:
### Changes proposed in this pull request
Fix a bug in `_sqlalchemy_map_condition_unexpected_count_value` to correctly handle edge case where the table being checked is empty. Closes #4603 

In this case, the unexpected count query may return None rather than a scalar value, resulting in a `TypeError` when trying to cast the result as an `int`. The error is now caught and handled by defaulting the unexpected count value to zero (since a table containing zero rows can only ever have zero unexpected rows).

### Definition of Done


jeffkpayne:
Awesome work!  We were just about to file an issue around this.  Thanks!  All you need to do is sign the CLA 😄

douglascook:
@cla-bot check

talagluck (approved):
LGTM - thanks so much for the contribution!

PR #4660, opened 2022-04-04
[BUGFIX] Update helper to add explicit alias to subqueries for SQLA version < 1.4

### LINKED ISSUE #4659: `_sqlalchemy_map_condition_unexpected_count_value`  fails for SQLA version < 1.4
**Describe the bug**
Certain expectations that generate a subquery fail with `sqlalchemy.exc.ProgrammingError: (psycopg2.errors.SyntaxError) subquery in FROM must have an alias` when run against Postgres on SQLAlchemy version earlier than 1.4.

The problem was partially identified in https://github.com/great-expectations/great_expectations/pull/3651, where a helper function was added to explicitly create a subquery with alias in order to avoid the deprecation warnings when running on SQLA >= 1.4. However, the fix is only applied on versions >= 1.4, so doesn't resolve the problem for the earlier versions where it actually results in the error above rather than a deprecation warning

**To Reproduce**
1. Install a version of `SQLAlchemy` < 1.4
2. Create an expectation to eg. `expect_column_values_to_be_unique`
3. Run the expectation against a Postgres DB (not sure if a similar error may occur with other DBs)

**Expected behavior**
Expectations using subqueries should run successfully on SQLA versions earlier than 1.4

 - Operating System: macOS
 - Great Expectations Version 0.14.13
 - SQLAlchemy version < 1.4

douglascook:
### Proposed Changes

Implicit conversion of a nested select into a subquery failed when running on `SQLAlchemy` 1.3 against Postgres. 

This PR updates the existing helper to also handle older supported versions of SQLA. Closes #4659
 
I'm not sure how best to test this, since the existing tests use a version of `SQLAlchemy` >= 1.4, but I've confirmed with some manual testing that it resolves my problem.

### Definition of Done


austiezr (approved):
LGTM! Thanks for this resolution @douglascook ! 🚀

PR #4661, opened 2022-04-04
[MAINTENANCE] Improve get validator functionality

abegong:
- Makes it possible to use `context.get_validator` without specifying an `ExpectationSuite` (by name, id, or otherwise.) This behavior was already supported by `Validator.__init__`. It just hadn't been enabled for `DataContext.get_validator`.
- Makes it possible to pass a single `Batch` to `context.get_validator`

Both of these additions will improve usability of the top-level API in DataContext.

Each is tested with a single unit test.


tests/data_context/test_data_context.py
```diff
@@ -1788,6 +1788,112 @@ def test_get_validator_with_attach_expectation_suite(
...
+        data_asset_name="A",
+        batch_identifiers={
+            "alphanumeric": "some_file",
+        },
+    )
+    assert type(my_validator.get_expectation_suite()) == ExpectationSuite
+    assert my_validator.expectation_suite_name == "default"
+
+def test_get_validator_with_batch(
+    empty_data_context, tmp_path_factory
```
cdkini: `tmp_path_factory` is session-scoped, which can cause state to bleed into other tests when the whole suite is run. I think we've been making a move to `tmp_path` for that reason.

tests/data_context/test_data_context.py
```diff
@@ -1788,6 +1788,112 @@ def test_get_validator_with_attach_expectation_suite(
     )
     assert my_validator.expectation_suite_name == "A_expectation_suite"
 
+def test_get_validator_without_expectation_suite(
+    empty_data_context_stats_enabled, tmp_path_factory
```
cdkini: do we want stats enabled here? I think we should refrain from using this fixture here since we don't test analytics.

tests/data_context/test_data_context.py
```diff
@@ -1788,6 +1788,112 @@ def test_get_validator_with_attach_expectation_suite(
...
+):
+    context: DataContext = empty_data_context_stats_enabled
```
cdkini: Instead of using a context and updating the filesystem, could we instantiate a `BaseDataContext` like we do here: https://github.com/great-expectations/great_expectations/blob/develop/tests/expectations/test_expectation_arguments.py#L39-L86

An approach like this would remove the need for I/O and improve performance. We might need to do some mocking though happy to dig in further as needed.

tests/data_context/test_data_context.py
```diff
@@ -1788,6 +1788,112 @@ def test_get_validator_with_attach_expectation_suite(
...
+
+    my_validator = context.get_validator(
+        datasource_name="my_directory_datasource",
+        data_connector_name="my_filesystem_data_connector",
+        data_asset_name="A",
+        batch_identifiers={
+            "alphanumeric": "some_file",
+        },
+    )
+    assert type(my_validator.get_expectation_suite()) == ExpectationSuite
```
cdkini: Nitpick - Could we use `isinstance` just to be consistent? This method accounts for things like inheritance hierarchies and is a bit safer.

tests/data_context/test_data_context.py
```diff
@@ -1788,6 +1788,112 @@ def test_get_validator_with_attach_expectation_suite(
...
+    assert my_validator.expectation_suite_name == "default"
+
+def test_get_validator_with_batch(
+    empty_data_context, tmp_path_factory
+):
+    context = empty_data_context
```
cdkini: Same general note about context - I think we could probably use `BaseDataContext` and mock any I/O.

great_expectations/data_context/data_context.py
```diff
@@ -1759,6 +1759,7 @@ def get_validator(
         data_connector_name: Optional[str] = None,
         data_asset_name: Optional[str] = None,
         *,
+        batch: Optional[Batch] = None,
```
alexsherstinsky, line 1762: @abegong I believe that this should be `batch_list: List[Batch]` -- do you agree?  At least some way to accept list of `Batch` objects, even if we must support the passing of a single `Batch` object.  Thanks

alexsherstinsky (changes requested):
@abegong Small change requested (happy to discuss) -- I really like what this PR will provide (I believe that we even had a JIRA item for it in the Backlog)!  Thanks!

great_expectations/data_context/data_context.py
abegong: For usability sake I think it's important to take a single `Batch` as an argument.

We could also add the option to pass a `batch_list`.

alexsherstinsky: @abegong I agree -- and, if possible, would prefer for both to be supported in the same pull request.  Thanks!

abegong: I've made this change

tests/data_context/test_data_context.py
```diff
@@ -1788,6 +1788,107 @@ def test_get_validator_with_attach_expectation_suite(
...
+                "id_key_0": "id_0_value_a",
+                "id_key_1": "id_1_value_a",
+            },
+        )
+    )
+
+    my_validator = context.get_validator(
+        batch=batch
+    )
+    assert type(my_validator.get_expectation_suite()) == ExpectationSuite
```
abegong:
```suggestion
    assert isinstance(my_validator.get_expectation_suite(), ExpectationSuite)
```

abegong:
@cdkini , @alexsherstinsky , I've responded to all comments.

alexsherstinsky (approved):
LGTM!

PR #4736, opened 2022-04-06
Add checks for mostly=1.0 for all renderers

abegong:
- Adds checks to see if `mostly==1.0` for all renderers. This avoids cases like this: "id values must be unique, at least 100 % of the time."
- I've added a single snapshot test to verify that this worked correctly in one instance. Note: I haven't used snapshot tests before, so it's possible that I implemented something incorrectly.
- Some of the multicolumn renderers had convoluted logic for deciding how to render Expectations with/without mostly. I've also cleaned that up a little.


alexsherstinsky (approved):
LGTM!

This is very nice -- and is an easy pattern to follow for adding "mostly" to the expectations that do not have it (but should).  Thanks!

anthonyburdi (approved):
LGTM!

abegong:
Thanks, @alexsherstinsky and @anthonyburdi !

PR #4862, opened 2022-04-14
[FEATURE] RuleBasedProfiler/DataAssistant/MetricMultiBatchParameterBuilder: Enable Returning Metric Computation Results with batch_id Attribution

alexsherstinsky:
### Scope
* Using the argument `include_batch_id_with_metric_value` (default is `False`), enable the metrics computation results by `MetricMultiBatchParameterBuilder` (and its subclasses) to return the `batch_id`, corresponding to the value of the specified metric, computed for the `Batch` object with that `batch_id` value.  This is useful for graphic the metrics.
* Update the existing unit tests for the case of `include_batch_id_with_metric_value = False`.
* Update the `VolumeDataAssistant` test of `metrics` to account for the ability to include `batch_id` values in the results.

- JIRA: GREAT-735/GREAT-798
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/expectations/core/expect_column_max_to_be_between.py
```diff
@@ -132,6 +132,7 @@ class ExpectColumnMaxToBeBetween(ColumnExpectation):
             bootstrap_random_seed=f"{VARIABLES_KEY}bootstrap_random_seed",
             round_decimals=f"{VARIABLES_KEY}round_decimals",
             truncate_values=f"{VARIABLES_KEY}truncate_values",
+            include_batch_id_with_metric_value=False,
```
donaldheppner: I would rather _remove_ this option and _always_ include batch ID when we can. It simplifies the API and code

anthonyburdi (commented):
Looks good pending resolution of Don's comment.

great_expectations/expectations/core/expect_column_max_to_be_between.py
alexsherstinsky: @donaldheppner @anthonyburdi @talagluck @NathanFarmer Currently, results are returned using the reserved keyword `value` and the metadata using the reserved keyword `detail`.  Hence, for using `value` as Expectation `kwargs`, only parameter computation results must be stored in `value`.  If we are to remove this option and always include `batch_id`, then we have two alternatives: store it in `details` (as a separate key) or create another reserved keyword, such as `value_with_batch_id`.  Of all three alternatives (leave as in this PR, store in `details`, and add the new keyword `value_with_batch_id`), the latter is the most appealing to me (i.e., add the new keyword `value_with_batch_id`).  Please let me know what you think -- happy to discuss.  Let us make the decision within an hour.  I will implement the above suggestion in the mean time so that we have it ready.  Thanks!

anthonyburdi: I think that suggestion makes sense. I thought about instead just including the list of batch_id corresponding to each value (in the same sorted order) but I think though there is duplication it is more clear as you have implemented it with {batch_id: value} just under a different key.

alexsherstinsky: This has been implemented.

anthonyburdi (approved):
LGTM!

PR #4879, opened 2022-04-16
[MAINTENANCE] split data context files

abegong:
- Non-breaking change to split the monster `data_context.py` module into a package with separate files for `BaseDataContext`, `DataContext`, and `ExplorerDataContext`
- Along the way, fix some minor bugs in imports and class inheritance that were obscured by the classes being in the same file.


cdkini (approved):
LGTM!

PR #4883, opened 2022-04-18
Experiment/lite data context

abegong:
- Add a `LiteDataContext` class that doesn't require pre-configuration, and has some syntactic sugar for better developer experience.
- Add a `PandasReaderDatasource` class with thin wrappers for all of pandas' `read_*` methods.

More context here (19min): https://www.loom.com/share/71cfd69e7aaf4fac82b345069cf8c359

This PR is a draft, not yet ready to merge. Known questions for arch discussion are flagged with `#!!!` and should be resolved before merging.


abegong:
One test is failing because `pyperclip` (needed to test `read_clipboard`) isn't included in requirements.

How do I know which of the many flavors of `requirements-dev` this should be included in?

great_expectations/datasource/configured_pandas_data_asset.py
```diff
@@ -0,0 +1,214 @@
...
+
+        self._method = method
+        self._base_directory = base_directory
+        self._regex = regex
+
+        super().__init__(
+            datasource=datasource, name=name, batch_identifiers=batch_identifiers
+        )
+
+    def update_configuration(
```
donaldheppner, line 49: I'd prefer to think of this as just updating the asset, and I think instead of using this method, this items could be properties that we assign. Using properties also makes it more extensible (fancy derived types can easily add properties/fields)

great_expectations/datasource/configured_pandas_data_asset.py
```diff
@@ -0,0 +1,214 @@
...
+        if base_directory != None:
+            self._base_directory = base_directory
+
+        if regex != None:
+            self._regex = regex
+
+        return self
+
+    def get_batch_request(
+        self, *batch_identifier_args, **batch_identifier_kwargs
```
donaldheppner, line 81: This is unnecessarily general I think. I also generally dislike kwargs from a usability stand-point since they are not self-documenting like named parameters or fields.

great_expectations/datasource/base_data_asset.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        self._name = name
+
+    def update_configuration(self, **kwargs) -> Self:
+        raise NotImplementedError
+        
+    def list_batches(self) -> List[NewBatchRequestBase]:
+        raise NotImplementedError
+
+    def get_batch(self, *args, **kwargs) -> Batch:
```
donaldheppner, line 51: These method signatures are not helpful ... how do I get a batch?

great_expectations/datasource/misc_types.py
```diff
@@ -0,0 +1,77 @@
...
+
+class PassthroughParameters(dict, SerializableDictDot):
+    pass
+    """
+    !!! For full feature parity with previous versions, this will need to support these 5 (or maybe 4) parameters.
+    batch_filter_parameters : {
+        "airflow_run_id": "string_airflow_run_id_that_was_provided",
+        "other_key": "string_other_key_that_was_provided",
+    limit: 10
+    index: Optional[Union[int, list, tuple, slice, str]],  # examples: 0; "-1"; [3:7]; "[2:4]"
```
donaldheppner, line 26: this could be better represented by a self-documenting class

great_expectations/datasource/misc_types.py
```diff
@@ -0,0 +1,77 @@
...
+
+class BatchIdentifiers(dict, SerializableDictDot):
+    pass
+
+
+class PassthroughParameters(dict, SerializableDictDot):
+    pass
+    """
+    !!! For full feature parity with previous versions, this will need to support these 5 (or maybe 4) parameters.
+    batch_filter_parameters : {
```
donaldheppner, line 22: What are these used for? Should they be at the batch level or the asset level?

great_expectations/datasource/misc_types.py
```diff
@@ -0,0 +1,77 @@
...
+        "airflow_run_id": "string_airflow_run_id_that_was_provided",
+        "other_key": "string_other_key_that_was_provided",
+    limit: 10
```
donaldheppner, line 25: this should be a limit sampler

great_expectations/datasource/misc_types.py
```diff
@@ -0,0 +1,77 @@
...
+@dataclass
+class NewConfiguredBatchRequest(NewBatchRequestBase):
+    passthrough_parameters: PassthroughParameters
+
+    def __str__(self) -> str:
+        return f"""NewConfiguredBatchRequest(
+    datasource_name='{self.datasource_name}',
+    data_asset_name='{self.data_asset_name}',
+    batch_identifiers='{self.batch_identifiers.__str__()}',
+    passthrough_parameters='{self.passthrough_parameters.__str__()}',
```
donaldheppner, line 52: rather than passthrough parameters, we should consider what parameters make sense to include in a batch request per execution engine and model specific batch request types to handle it.

I'm also not generally a fan of expressing these parameters here because of the implications on batch metrics for a given asset; we should consider moving these up to the asset level so we can better compare batches

great_expectations/datasource/new_new_new_datasource.py
```diff
@@ -0,0 +1,67 @@
...
+        #     GxExperimentalWarning,
+        # )
+
+        self._name = name
+        self._assets = DictDot()
+
+    def add_asset(self, name:str, *args, **kwargs) -> BaseDataAsset:
+        raise NotImplementedError
+
+    def rename_asset(self, old_name: str, new_name: str) -> None:
```
donaldheppner, line 39: IMHO this should be an asset operation, not a datasource one.

ex: person.name = "Gabe"

We can add magic to the assignment property to check for collisions

great_expectations/datasource/new_new_new_datasource.py
```diff
@@ -0,0 +1,67 @@
...
+        if new_name in self._assets.keys():
+            raise KeyError(f"An asset named {new_name} already exists.")
+
+        self._assets[new_name] = self._assets.pop(old_name)
+        asset = self._assets[new_name]
+        asset.set_name(new_name)
+
+        return asset
+
+    def list_asset_names(self) -> List[str]:
```
donaldheppner, line 52: let's just get assets here; we are in a runtime environment and can use objects over keys

great_expectations/datasource/misc_types.py
abegong: Agreed. This was just copied from elsewhere in the codebase.

great_expectations/datasource/new_new_new_datasource.py
abegong: Sure---the important thing is that the list of asset names on the Datasource must be kept in sync with the actual names of the DataAssets.

I don't have a strong opinion on which object is the driver for that.

great_expectations/datasource/new_new_new_datasource.py
abegong: I'm not sure what you mean. In this prototype, NewNewNewDatasource also supports and `assets` property. Is that what you're looking for?

I think it's important to have a convenience method that can return a list of asset names as strings. Hence: `list_asset_names`.

(`get_asset_name_list` would also work, if we make the style guide decision that we don't like `list_` as a verb.)

great_expectations/datasource/base_data_asset.py
abegong: I know you probably felt hurt and betrayed to see args and kwargs in method signatures, but hear me out.

Assuming we want to go down the road of Datasource proliferation---which I think is the right decision---then we're going to have:
1. many Datasources
2. each with a `get_batch` method,
3. each taking a different set of arguments.

In that world, it makes sense for the base class `get_batch` method to have an extremely general method declaration, so that other classes can have more specific signatures.

Note: The base class method docstring should explain all this. Then the child class methods can have actual arguments and informative docstrings. IMO, that's the best division of labor here.

Note: I also considered the alternative of having each child class declare its own `get_batch_from_*something*` method(s). This would avoid the args/kwargs issues, but it has other problems.

great_expectations/datasource/configured_pandas_data_asset.py
abegong: Maybe.... In this case, I'm using args and kwargs to (1) avoid having to enumerate every argument that can be accepted by any pandas read_ method, and (2) to allow users to specify custom names for batch identifiers and pass them in as arguments.

I agree that it makes the docstring unhelpful.

PR #4880, opened 2022-04-17
[BUGFIX] `great_expectations` import fails when SQL Alchemy is not installed

abegong:
Yikes! Without this line, `import great_expectations as gx` fails unless sqlalchemy is installed.

I'm scared that this bug made it to `develop`---really glad it's not on `main` yet. Do we have any test environments that don't include SQLalchemy?

```
NameError                                 Traceback (most recent call last)
Input In [1], in <cell line: 1>()
----> 1 import great_expectations as gx

File ~/Documents/great_expectations/great_expectations/__init__.py:7, in <module>
      4 __version__ = get_versions()["version"]  # isort:skip
      5 del get_versions  # isort:skip
----> 7 from great_expectations.data_context import DataContext
      9 from .util import (
     10     from_pandas,
     11     get_context,
   (...)
     20     validate,
     21 )
     23 # from great_expectations.expectations.core import *
     24 # from great_expectations.expectations.metrics import *

File ~/Documents/great_expectations/great_expectations/data_context/__init__.py:1, in <module>
----> 1 from great_expectations.data_context.data_context import (
      2     BaseDataContext,
      3     DataContext,
      4     ExplorerDataContext,
      5     LiteDataContext,
      6 )

File ~/Documents/great_expectations/great_expectations/data_context/data_context/__init__.py:1, in <module>
----> 1 from great_expectations.data_context.data_context.base_data_context import (
      2     BaseDataContext,
      3 )
      4 from great_expectations.data_context.data_context.data_context import DataContext
      5 from great_expectations.data_context.data_context.explorer_data_context import (
      6     ExplorerDataContext,
      7 )

File ~/Documents/great_expectations/great_expectations/data_context/data_context/base_data_context.py:21, in <module>
     18 from ruamel.yaml.comments import CommentedMap
     20 from great_expectations.core.config_peer import ConfigPeer
---> 21 from great_expectations.execution_engine import ExecutionEngine
     22 from great_expectations.rule_based_profiler.config.base import (
     23     ruleBasedProfilerConfigSchema,
     24 )
     26 try:

File ~/Documents/great_expectations/great_expectations/execution_engine/__init__.py:4, in <module>
      2 from .pandas_execution_engine import PandasExecutionEngine
      3 from .sparkdf_execution_engine import SparkDFExecutionEngine
----> 4 from .sqlalchemy_execution_engine import SqlAlchemyExecutionEngine

[...]

PR #4884, opened 2022-04-18
[MAINTENANCE] Add CI stage importing GE with only required dependencies installed 

anthonyburdi:
- We noticed a bug that only shows up in environments without sqlalchemy installed (fixed in: https://github.com/great-expectations/great_expectations/pull/4880), this PR adds a new CI stage that runs with only required dependencies installed (i.e. without sqlalchemy installed) so that bugs like this will be caught before they can be merged into `develop`. The test adds a new `import_ge` stage which installs GE from the branch and imports it in supported python versions. It precedes the `required` stage since it runs quickly and if there are issues in this stage they should be fixed before running the full test suite.

This branch was tested in the following CI runs:
- Failing with bug re-introduced: https://dev.azure.com/great-expectations/great_expectations/_build/results?buildId=47711&view=logs&j=2b99e989-08bc-5401-cdad-c2e6353fb293&t=af523853-95e2-5ff4-7108-1c77c9b4f35c
- Passing (in this PR)

### Definition of Done

Thank you for submitting!

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -81,8 +81,42 @@ stages:
               pyupgrade --py3-plus || EXIT_STATUS=$?
               exit $EXIT_STATUS
 
-  - stage: required
+  - stage: import_ge
     dependsOn: [lint]
+    pool:
+      vmImage: 'ubuntu-18.04'
+    jobs:
+      - job: import_ge
+
+        strategy:
+         matrix:
+           Python37:
+             python.version: '3.7'
+           Python38:
+             python.version: '3.8'
+           Python39:
+             python.version: '3.9'
+
+
+        steps:
+         - task: UsePythonVersion@0
+           inputs:
+             versionSpec: '$(python.version)'
+           displayName: 'Use Python $(python.version)'
+
+         - bash: python -m pip install --upgrade pip==21.3.1
+           displayName: 'Update pip'
+
+         - script: |
+             pip install  .
+           displayName: 'Install GE and required dependencies (i.e. not sqlalchemy)'
+
+         - script: |
+             python -c "import great_expectations as ge; print('GE Version:', ge.__version__)"
+           displayName: 'Import Great Expectations'
+
+  - stage: required
+    dependsOn: [lint, import_ge]
     pool:
       vmImage: 'ubuntu-18.04'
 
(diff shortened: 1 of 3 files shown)


NathanFarmer (commented):
Rather than test "no sqlalchemy" specifically, isn't this a problem with testing only dependencies in `requirements.txt` (it is missing `sqlalchemy`)? I seem to remember us adding a step that attempts to `import great_expectations`, but it was probably only for `requirements-dev.txt` because we were dealing with a broken dependency and couldn't merge into `develop`. This would do the same thing, but for `requirements.txt` and so it would test all dependencies and not just `sqlalchemy`.

azure-pipelines.yml
```diff
@@ -170,6 +170,64 @@ stages:
...
               reportDirectory: '$(System.DefaultWorkingDirectory)/**/htmlcov'
 
+      - job: no_sqlalchemy
+        condition: or(eq(variables.isScheduled, true), eq(variables.isMain, true), eq(variables.isManual, true))
+
+        services:
+          postgres: postgres
+
+        variables:
+          GE_pytest_opts: ''
```
cdkini: If opts are empty, do we need this section?

azure-pipelines.yml
```diff
@@ -170,6 +170,64 @@ stages:
...
+
+        strategy:
+          matrix:
+            Python38:
+              python.version: '3.8'
+              GE_pytest_pip_opts: '--requirement requirements-dev.txt --constraint constraints-dev.txt'
+              pip_uninstall_packages: 'sqlalchemy'
```
cdkini: Do we plan on testing more versions? If not, what are your thoughts on moving away from a matrix and just using the single version?

azure-pipelines.yml
```diff
@@ -170,6 +170,64 @@ stages:
...
+              versionSpec: '$(python.version)'
+            displayName: 'Use Python $(python.version)'
+
+          - bash: python -m pip install --upgrade pip==21.3.1
+            displayName: 'Update pip'
+
+          - script: |
+              sudo apt-get install -y pandoc
+              pip install pypandoc
+            displayName: 'Install pandoc'
```
cdkini: More of a general question but what is pandoc doing here? I've seen it a few times and haven't figured out where it's used.

cdkini (approved):
Minor comments but LGTM.

Should this also be run on `develop` or on on scheduled runs/release cuts?

azure-pipelines.yml
anthonyburdi: I just copied from comprehensive - I'm not sure, and also the napoleon-docstrings and other params in the pytest command. I think these may be related to code coverage / old rst docs. At some point worth investigating and removing.

azure-pipelines.yml
anthonyburdi: I'm not sure - was leaving in case we want to also add 3.9. Let me think on it.

azure-pipelines.yml
anthonyburdi: Good call - was just mirroring the comprehensive section but based on our conversation in slack this is probably better done with just a script that imports GE instead of running the full pytest suite anyway.

azure-pipelines.yml
```diff
@@ -81,8 +81,42 @@ stages:
...
+         - script: |
+             pip install  .
+           displayName: 'Install GE and required dependencies (i.e. not sqlalchemy)'
+
+         - script: |
+             python -c "import great_expectations as ge; print('Successfully imported GE Version:', ge.__version__)"
+           displayName: 'Import Great Expectations'
+
+  - stage: required
+    dependsOn: [lint, import_ge]
```
cdkini, line 119: Should we reflect this change in `dependsOn` in the other pipeline as well?

azure-pipelines.yml
```diff
@@ -81,8 +81,42 @@ stages:
...
+         - task: UsePythonVersion@0
+           inputs:
+             versionSpec: '$(python.version)'
+           displayName: 'Use Python $(python.version)'
+
+         - bash: python -m pip install --upgrade pip==21.3.1
+           displayName: 'Update pip'
+
+         - script: |
+             pip install  .
```
NathanFarmer, line 111: Won't this install anything that is imported in the codebase rather than `requirements.txt`?

NathanFarmer (changes requested):
Looking more like what I was thinking, but I have one question.

azure-pipelines.yml
cdkini: it'll use the `setup.py` to download a local installation so I think this is okay?

cdkini: https://github.com/great-expectations/great_expectations/blob/develop/setup.py#L37-L39 shows that it is equivalent to `requirements.txt`

anthonyburdi: 👀 Checking

azure-pipelines.yml
anthonyburdi: Good catch - fixing!

azure-pipelines.yml
anthonyburdi: (It did not install sqlalchemy locally but checking in-pipeline and in the pip docs again)

NathanFarmer (approved):
LGTM!

azure-pipelines.yml
anthonyburdi: Looks like we are good, using pip to install from a directory e.g. `pip install .` makes pip look for a setup.py file (https://pip.pypa.io/en/stable/cli/pip_install/#argument-handling) which in our case specifies requirements.txt as our requirements and other requirements as `extras_require`.

NathanFarmer: Some research confirms what you say @cdkini and `setup.py` does not allow `sqlalchemy` unless you use the "extra" e.g. `great_expectations[sqlalchemy]`

PR #4898, opened 2022-04-19
[MAINTENANCE] Add condition to primary pipeline to ensure `import_ge` stage doesn't cause misleading Slack notifications

cdkini:
- Without the condition proposed herein, any failures will trigger a Slack notification. Since this is commonplace in the development process, we should only run this stage when we're in a scheduled run or are performing a release cut (the rest of the pipeline mirrors this logic).

### Definition of Done


anthonyburdi (approved):
LGTM!

PR #4940, opened 2022-04-25
[MAINTENANCE] Add condition for custom checks in `great_expectations` pipelines

cdkini:
- Before this PR, these checks would get run in both the `dependency_graph` and `great_expectations` pipelines.
- We should have parity between the two pipelines but only `dependency_graph` should run on each commit. `great_expectations` is reserved for scheduled runs and release cutting (and therefore is more extensive).


NathanFarmer (approved):
LGTM!

PR #4954, opened 2022-04-26
[BUGFIX] Fix Splitter Integration Test for BigQuery

Shinnnyshinshin:
- Fixes integration test for Splitters that was causing our `docs-integration` stage to fail. 

## What was the cause? 
- Using the `to_sql()` method in `Pandas` appears to have an insert query for every row [in the DataFrame](https://stackoverflow.com/questions/29706278/python-pandas-to-sql-with-sqlalchemy-how-to-speed-up-exporting-to-ms-sql), so for the ~300 line test df we were loading for the splitter tests, this was taking > 15 min and causing our tests to time out and fail.
- [The BigQuery Documentation](https://cloud.google.com/bigquery/docs/samples/bigquery-load-table-dataframe) recommends using the Google BigQuery client for these kinds of situations, which will load the Dataframe as part of 1 query. 
- Therefore, the primary change for this PR is the `load_data_into_test_bigquery_database_with_bigquery_client()` which is used by the `split_data_on_datetime_bigquery()` test.  Test now passes and runs in ~ 1 min. 

### Definition of Done

### DIFF
--- a/tests/integration/db/test_sql_data_splitting.py
+++ b/tests/integration/db/test_sql_data_splitting.py
@@ -70,13 +70,11 @@ if __name__ == "test_script_module":
 
     dialect, connection_string = _get_connection_string_and_dialect()
     print(f"Testing dialect: {dialect}")
+
     print("Preemptively cleaning old tables")
-    # see if expiration can be set with BigQuery backend?
-    # or if this is being cleaned up too often
-    # if
-    # clean_up_tables_with_prefix(
-    #     connection_string=connection_string, table_prefix=f"{TAXI_DATA_TABLE_NAME}_"
-    # )
+    clean_up_tables_with_prefix(
+        connection_string=connection_string, table_prefix=f"{TAXI_DATA_TABLE_NAME}_"
+    )
 
     loaded_table: LoadedTable = _load_data(connection_string=connection_string)
     test_df: pd.DataFrame = loaded_table.inserted_dataframe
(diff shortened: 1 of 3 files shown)


tests/integration/db/test_sql_data_splitting.py
```diff
@@ -38,6 +38,7 @@ def _get_connection_string_and_dialect() -> Tuple[str, str]:
     if dialect == "snowflake":
         connection_string: str = get_snowflake_connection_url()
     elif dialect == "bigquery":
+        # still needed for loading the data we keep this
         connection_string: str = get_bigquery_connection_url()
```
Shinnnyshinshin: Keeping `connection_string` for BigQuery as it is needed by the `Datasource`

tests/integration/test_script_runner.py
```diff
@@ -549,20 +549,20 @@ class IntegrationTestFixture:
...
-    #     util_script="tests/test_utils.py",
-    #     other_files=(
-    #         (
-    #             "tests/integration/fixtures/split_data/bigquery_connection_string.yml",
-    #             "connection_string.yml",
-    #         ),
-    #     ),
-    #     extra_backend_dependencies=BackendDependencies.BIGQUERY,
-    # ),
+    IntegrationTestFixture(
```
Shinnnyshinshin, line 552: Uncommented!

alexsherstinsky (approved):
LGTM

PR #4962, opened 2022-04-26
[FEATURE] Update `DataAssistantResult.plot()` return value to emit `PlotResult` wrapper dataclass

cdkini:
- Add return value to `plot` (list of Altair charts) -> New `PlotResult` class
  - Note that these are the raw charts (no concatenation)
- Write tests to check for target strings in result objects

### Definition of Done


great_expectations/rule_based_profiler/types/data_assistant_result/volume_data_assistant_result.py
```diff
@@ -59,6 +59,8 @@ def plot(
 
         self.display(charts=charts, theme=theme)
 
+        return [chart.to_dict() for chart in charts]
```
cdkini: Should this be conditional? Should we always return the config?

I have concerns about polluting the notebook with config. Maybe a boolean flag in the signature?

cdkini: Additionally, we can use `chart.to_json()` as an alternative (returns a string). Thoughts here @NathanFarmer ?

NathanFarmer: We could have both `to_dict` and `to_json` available with boolean flag defaulted to `False`.

NathanFarmer (changes requested):
I think we should default to no return value and allow boolean flags directly in `result.plot()` to allow for returning `dict` or json `str` representation. CC: @alexsherstinsky @donaldheppner

great_expectations/rule_based_profiler/types/data_assistant_result/volume_data_assistant_result.py
donaldheppner: I'd rather not. Let's pick .to_dict() if we are going this way; it is no effort to convert a dictionary to json if folks want it that way. What is the use case for either (or both?)

donaldheppner (changes requested):
I'm not sure why we are doing this ... and are we overburdening this method? Is it so we can create aggregate plot method results downstream? Couple of thoughts:
1) I'd rather return a stronger "Chart" type, even if it is a GE specific chart type, so folks know what the dictionary is intended to represent. I assume we went this way with dictionaries to avoid leaking implementation? We can instead solve the leak with encapsulation vs. dumping the raw data.
2) I do not want to manipulate the return type based on the input parameters (ie: returning either json or dict based on input parameters). This is where returning a richer type that produce the representation you want makes more sense. The richer type also allows you to encapsulate what you expect the caller to do with the result, so we're not having to duplicate that code across all callers.
3) for me, having a reference into a jira issue with ACs is helpful in supporting the review

great_expectations/rule_based_profiler/types/data_assistant_result/chart.py
```diff
@@ -0,0 +1,17 @@
+from dataclasses import dataclass
+from typing import List
+
+import altair as alt
+
+
+@dataclass(frozen=True)
```
alexsherstinsky: ❤️

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
```diff
@@ -509,7 +512,7 @@ def plot(
         self,
         prescriptive: bool = False,
         theme: Optional[Dict[str, Any]] = None,
-    ) -> None:
+    ) -> Chart:
```
alexsherstinsky: ❤️

tests/test_utils.py
```diff
@@ -803,3 +803,43 @@ def set_bootstrap_random_seed_variable(
         rule.variables = build_parameter_container_for_variables(
             variables_configs=variables_dict
         )
+
+
+def find_strings_in_nested_obj(obj: Any, target_strings: List[str]) -> bool:
```
alexsherstinsky, line 780: @cdkini What can (or should?) we do about the following two questions:
1) We have `find_code_in_notebook()` and `replace_code_in_notebook()` in our `CLI` rendering utils.  Should these utilities be combined, or should they remain on their own?
2) When I was developing the above, if `find_strings_in_nested_obj()` existed and I had known about it, I would have attempted to leverage it.  What can we do towards making our team members aware of the useful utilities (for practical purposes as well as for avoiding re-implementing the same logic)?  Happy to brainstorm.  Thanks! /cc @donaldheppner

alexsherstinsky (approved):
LGTM (pending review/approval by @NathanFarmer ).  Thanks!

great_expectations/rule_based_profiler/types/data_assistant_result/chart.py
Shinnnyshinshin: :heart:

tests/test_utils.py
```diff
@@ -803,3 +803,43 @@ def set_bootstrap_random_seed_variable(
...
+    """Recursively traverse a nested structure to find all strings in an input string.
```
Shinnnyshinshin, line 781: recursion! :heart:

Shinnnyshinshin (approved):
Thank you for this Chetan. LGTM pending Nathan and Don's comments.

tests/test_utils.py
cdkini: I think there might be a little bit of overlap but I'd be okay with keeping them separate for now. That feel like a very notebook-specific utility while this one is more general purpose.

And it's totally my fault for not being super vocal about this util! I think in large part it comes from just bringing it up during PR review ("hey I think this util would help your tests"). We should parse through our utils to see what's useful and when each one would be used.

cdkini: Maybe a README that documents the above?

great_expectations/rule_based_profiler/types/data_assistant_result/plot_result.py
```diff
@@ -0,0 +1,19 @@
...
+
+    Please note that contained within this object are the raw Altair charts generated
+    by `DataAssistantResult.plot()`. They may have been concatenated or formatted for
+    purposes of display in Jupyter Notebooks.
+
+    Attributes:
+        charts: The list of Altair charts rendered through `DataAssistantResult.plot()`
+    """
+
+    charts: List[alt.Chart]
```
donaldheppner, line 19: let's say we wanted to encapsulate our use of altair. Could we produce a property for "to_json" and "to_dict" and keep the chart itself private? </bad_idea>?

great_expectations/rule_based_profiler/types/data_assistant_result/volume_data_assistant_result.py
```diff
@@ -49,16 +52,23 @@ def plot(
...
-        charts.append(column_domain_chart)
+        concatenated_column_domain_chart: alt.VConcatChart = alt.vconcat(
+            *column_domain_charts
+        )
+        charts.append(concatenated_column_domain_chart)
 
         self.display(charts=charts, theme=theme)
 
+        # Return the non-concatenated chart to the user for downstream processing
+        return_charts: List[alt.Chart] = table_domain_charts + column_domain_charts
```
donaldheppner: 👍

great_expectations/rule_based_profiler/types/data_assistant_result/volume_data_assistant_result.py
```diff
@@ -167,81 +177,85 @@ def _filter(e: ExpectationConfiguration) -> bool:
...
+        )
+        attributed_values_by_metric_name: Dict[str, ParameterNode] = attributed_metrics[
+            domain
+        ]
 
-            if prescriptive:
-                for kwarg_name in expectation_configuration.kwargs:
-                    df[kwarg_name] = expectation_configuration.kwargs[kwarg_name]
+        # Altair does not accept periods.
+        metric_name = list(attributed_values_by_metric_name.keys())[0].replace(".", "_")
```
donaldheppner: what if we replaced docs with spaces?

tests/test_utils.py
donaldheppner: re # 2, that is why I don't like utility methods, they are undiscoverable. :P

great_expectations/rule_based_profiler/types/data_assistant_result/plot_result.py
cdkini: hmm we definitely could but I think users are going to want to have the Altair Python objects since they can be directly updated during runtime. Some may just use the JSON output but I think we need to enable multiple workflows (even if that means exposing the library choice).

I think this is slightly different from our YAML situation in that this is recognized, well-supported library with a lot of recognition in the data space.

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
```diff
@@ -94,6 +97,7 @@ def get_line_chart(
         metric_type: alt.StandardType,
         domain_name: str,
         domain_type: alt.StandardType,
+        subtitle: Optional[str] = None,
```
cdkini, line 100: Necessary modification to make `get_line_chart` generic and work with both domains. I'm wary of bloating function signatures like this but this one simple change let's us reuse all the logic (since the only difference between the domains was a subtitle to reflect the column name).

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
```diff
@@ -149,12 +156,13 @@ def get_line_chart(
         return line + points
 
     @staticmethod
-    def get_expect_table_values_to_be_between_chart(
+    def get_expect_values_to_be_between_chart(
```
cdkini, line 159: Same note about generic methods.

great_expectations/rule_based_profiler/types/data_assistant_result/volume_data_assistant_result.py
```diff
@@ -167,22 +124,63 @@ def _filter(e: ExpectationConfiguration) -> bool:
...
+        )
+
+        return self._chart_values(
+            df=df,
+            metric_name=metric_name,
+            metric_type=metric_type,
+            domain_name=domain_name,
+            domain_type=domain_type,
+            prescriptive=prescriptive,
+            subtitle=None,
```
cdkini, line 175: New helper that encapsulates all the shared logic between `_create_chart_for_table...` and `_create_chart_for_column`. This circumvents the need for code duplication but let me know if it feels bloated.

great_expectations/rule_based_profiler/types/data_assistant_result/volume_data_assistant_result.py
```diff
@@ -193,59 +191,76 @@ def _create_chart_for_column_domain_expectations(
...
+            subtitle=subtitle,
+        )
 
+    def _chart_values(
+        self,
+        df: pd.DataFrame,
+        metric_name: str,
+        metric_type: alt.StandardType,
+        domain_name: str,
+        domain_type: alt.StandardType,
```
cdkini, line 237: Thought for a later PR - could we make these `namedtuples`? I think the pairing makes sense since they're always supplied together and it would cut down on the signature size.

NathanFarmer: Sounds reasonable to me!

NathanFarmer: Sounds reasonable to me!

NathanFarmer (approved):
LGTM!

great_expectations/rule_based_profiler/types/data_assistant_result/volume_data_assistant_result.py
donaldheppner: I think we could take it a step further and make these objects, encapsulating behavior to produce tool tips and titles for metrics and domains

donaldheppner (approved):

PR #4991, opened 2022-04-29
[MAINTENANCE] Move `DataAssistant` registry capabilities into `DataAssistantRegistry` to enable user aliasing

cdkini:
- Minor refactor of the `DataAssistant` registry process to better encapsulate global variables and functions into relevant classes.
- Enable users who define their own `DataAssistants` the ability to also define aliases (not previously possible due to the responsibility of this feature falling within the scope of the dispatcher).

### Definition of Done

Thank you for submitting!


great_expectations/expectations/registry.py
```diff
@@ -9,7 +9,6 @@
 logger = logging.getLogger(__name__)
 
 _registered_expectations = {}
-_registered_data_assistants = {}
```
cdkini, line 12: Purpose of refactor - Let's limit the amount of logic in the global scope. This can be encapsulated quite well in a data assistant-specific class.

great_expectations/expectations/registry.py
```diff
@@ -95,48 +94,6 @@ def register_expectation(expectation: Type["Expectation"]) -> None:  # noqa: F82
     _registered_expectations[expectation_type] = expectation
 
 
-def register_data_assistant(
```
cdkini, line 98: Purpose of refactor - This are functions but they could be methods. They directly relate to activities around registration, which is responsibility of the `DataAssistantDispatcher`. Wherever possible, let's try to associate functions with classes that cover that particular behavior (within reason of course).

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
```diff
@@ -105,6 +109,13 @@ def __init__(
         )
         self._build_profiler()
 
+    @classmethod
+    def __init_subclass__(cls) -> None:
+        if not hasattr(cls, "__alias__"):
+            raise NotImplementedError(
+                f"Class {cls} lacks required `__alias__` class attribute of type Optional[str]"
+            )
```
cdkini: In order for users to be able to define aliases, they need to be able to configure the value through their own custom implementations of the assistant. As such, we should place the burden of this responsibility within the assistant (as opposed to the dispatcher)

cdkini: I went with a dunder attribute here instead of a property because at the time of usage, we don't have class instance (just the types). I could use an abstract class or static method but opted for a slightly different approach.

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
```diff
@@ -1,21 +1,21 @@
...
 class DataAssistantDispatcher:
     """
     DataAssistantDispatcher intercepts requests for "DataAssistant" classes by their registered names and manages their
     associated "DataAssistantRunner" objects, which process invocations of calls to "DataAssistant" "run()" methods.
     """
 
-    DATA_ASSISTANT_TYPE_ALIASES: Dict[str, str] = {
-        "volume": "volume_data_assistant",
-    }
+    _registered_data_assistants: Dict[str, Type[DataAssistant]] = {}
```
cdkini, line 18: This is truly private now - the dispatcher acts as the interface callers need to interact with to use the values contained herein.

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
```diff
@@ -26,20 +26,20 @@ def __init__(self, data_context: "BaseDataContext") -> None:  # noqa: F821
...
+        ] = DataAssistantDispatcher.get_data_assistant_impl(data_assistant_type=name)
+
+        # If no registered data_assistant_type exists, try aliases for invocation.
         if data_assistant_cls is None:
-            data_assistant_cls = get_data_assistant_impl(
-                data_assistant_type=DataAssistantDispatcher.DATA_ASSISTANT_TYPE_ALIASES.get(
-                    name.lower()
-                )
-            )
+            for type_ in DataAssistantDispatcher._registered_data_assistants.values():
```
cdkini: Less efficient than a direct lookup but we have so few assistants that moving from O(1) to O(N) is negligible.

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
```diff
@@ -61,3 +61,54 @@ def __getattr__(self, name) -> DataAssistantRunner:
             ] = data_assistant_runner
 
         return data_assistant_runner
+
+    @staticmethod
```
cdkini: Should these be class methods instead? Always get tripped up between the two.

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
cdkini: As it stands, you NEED to always define an `__alias__` attr (even if you're just setting to `None`). This may not be the cleanest way to do it but I prefer explicitness to convey intent.

donaldheppner: I'm not clear on the semantics of a dunder method and if it applies here. Can you explain your thinking here?

Aliases are not required. The way to think of it is as a second name under which the assistant is registered. If there is no alias, then the assistant is registered under just one name. The program should not explode if Alias is missing

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
donaldheppner: consider flipping the order here; most folks will reference assistants with the name.

... and why not do `x = DataAssistantDispatcher._registered_data_assistants.get(name)`; `if x is None ....`?

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
donaldheppner: I'd use class methods since you invoke other class methods (and can specify `cls._registered_data_assistants` lower down).

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
```diff
@@ -61,3 +62,54 @@ def __getattr__(self, name) -> DataAssistantRunner:
...
+        data_assistant: Type[DataAssistant],  # noqa: F821
+    ) -> None:
+        """
+        This method executes "run()" of effective "RuleBasedProfiler" and fills "DataAssistantResult" object with outputs.
+
+        Args:
+            data_assistant: "DataAssistant" class to be registered
+        """
+        data_assistant_type = data_assistant.data_assistant_type
+        registered_data_assistants: Dict[
```
donaldheppner: do we need this type hint? can it not be very directly inferred?

great_expectations/rule_based_profiler/data_assistant/volume_data_assistant.py
```diff
@@ -27,6 +27,8 @@ class VolumeDataAssistant(DataAssistant):
         - Others in the future.
     """
 
+    __alias__: str = "volume"
```
donaldheppner, line 30: ... I do like the look of this

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
cdkini: Hmm so dunder ("double underscore") attributes are just convention when defining something that has a special purpose. Python's stdlib often uses them to define core language features (`__init__`, `__new__`, etc).

Some libraries outside of the stdlib, like SQLAlchemy, will make use of them:
```python
base = declarative_base()

class MyModel(Base):
    __table__ = "my_model" # Let's our ORM know what to access within the DB
   ...
```

Marshmallow also seems to use it - I'll keep it as is for now but can easily change if we feel like it's misleading or confusing!

cdkini: In regards to your point about alias exploding, I definitely agree. I can circumvent this by adding a `_alias` attribute to the base class and setting it to `None` so that child classes will pass the check.

We can then get rid of this `__init_subclass__` check and always assume we have the attribute available when iterating through assistant types.

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
cdkini: Although I just mentioned that performance was not an issue, I went with this approach so we could leverage the O(1) lookup of the dictionary and only do the iteration through types if needed.

I can reverse this! In regards to `get`, that definitely works. Funny how I missed that after we just talked about that pattern earlier today 😆

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
cdkini: perfect!

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
cdkini: it can for sure but we've been leaning towards being very explicit with everything. `mypy` and other static analysis tools don't require the additional typing but per discussion with Alex, they don't hurt.

I'm on the fence and could go either way.

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
```diff
@@ -83,6 +83,8 @@ class DataAssistant(metaclass=MetaDataAssistant):
         profiler_config: RuleBasedProfilerConfig = result.profiler_config
     """
 
+    __alias__: Optional[str] = None
```
donaldheppner, line 86: 👍

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
```diff
@@ -29,24 +29,27 @@ def __init__(self, data_context: "BaseDataContext") -> None:  # noqa: F821
...
-        # If no registered data_assistant_type exists, try aliases for invocation.
-        if data_assistant_cls is None:
-            type_: Type[DataAssistant]
-            for type_ in DataAssistantDispatcher._registered_data_assistants.values():
-                if type_.__alias__ == name:
-                    data_assistant_cls = type_
-                    break
+        # Attempt to utilize alias for invocation first.
+        type_: Type[DataAssistant]
+        for type_ in DataAssistantDispatcher._registered_data_assistants.values():
```
donaldheppner: I don't get what's happening here. The `_registered_data_assistants` is a dictionary, we should use it as one. Doesn't the key match the `name` parameter?

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
```diff
@@ -74,29 +78,34 @@ def register_data_assistant(
...
-                )
+        cls._register(data_assistant_type, data_assistant)
+
+        alias: Optional[str] = data_assistant.__alias__
+        if alias is not None:
+            cls._register(alias, data_assistant)
+
+    @classmethod
+    def _register(cls, name: str, data_assistant: Type[DataAssistant]) -> None:
+        registered_data_assistants = cls._registered_data_assistants
```
donaldheppner, line 78: We should just blow up here if we register two assistants with the same name. Throw a ValueError

donaldheppner (changes requested):

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
cdkini: oh wow wait this makes a ton of sense. I originally did not register the aliases in the dict so I wasn't aware we had access to them at this point.

We should now be able to just use `get()` and search through aliases and full names at the same time. Thanks for calling this out!

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
cdkini: I agree - I've changed this to reflect this.

donaldheppner (approved):

PR #4999, opened 2022-05-02
[BUGFIX] RBP Profiling Dataset ProgressBar Fix

Shinnnyshinshin:
- ProgressBar now displays the number of Rules being profiled. This was off because we were not instantiating the progress bar with the number `n` (ie. number of items to process). 
- This PR proposes that we instantiate the ProgressBar with number of Rules and increment by 1 as each Rule is processed. 

**Before**
```
Profiling Dataset: 0it [00:00, ?it/s]
Profiling Dataset: 1it [00:00, ?it/s]
Profiling Dataset: 20it [00:00, ?it/s]
```

**After**
```
Profiling Dataset:   0%|          | 0/2 [00:00<?, ?it/s]
Profiling Dataset:  50%|█████     | 1/2 [00:00<?, ?it/s]
Profiling Dataset: 100%|██████████| 2/2 [00:00<?, ?it/s]`
```

### DIFF
--- a/great_expectations/rule_based_profiler/rule_based_profiler.py
+++ b/great_expectations/rule_based_profiler/rule_based_profiler.py
@@ -245,16 +245,6 @@ class BaseRuleBasedProfiler(ConfigPeer):
                 if "rule_based_profiler" in progress_bars:
                     disable = not progress_bars["rule_based_profiler"]
 
-        # noinspection PyProtectedMember,SpellCheckingInspection
-        progress_bar: tqdm = tqdm(
-            desc="Profiling Dataset",
-            disable=disable,
-        )
-        progress_bar.update(0)
-        progress_bar.refresh()
-        sys.stdout.write("\n")
-        sys.stdout.flush()
-
         effective_variables: Optional[
             ParameterContainer
         ] = self.reconcile_profiler_variables(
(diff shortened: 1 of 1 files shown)


great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -290,9 +290,7 @@ def run(
                 reconciliation_directives=reconciliation_directives,
             )
             self.rule_states.append(rule_state)
-
-            domains_count += len(rule_state.domains)
-            progress_bar.update(domains_count)
+            progress_bar.update(1)
```
alexsherstinsky, line 293: @Shinnnyshinshin I see the benefit of the fix, but I was hoping we could update the progress bar on every `Domain`, because that is the lowest quantum of progress.  I am approving this for now so we can merge it, but if my request is possible in reasonable time effort, that would be awesome (it might require some kind of coordination  between `RuleBasedProfiler.run()` and `Rule.run()`.  Thanks!

alexsherstinsky (approved):
LGTM (please also see a comment, containing a request).  Thanks!

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: I did look this, but it does require one more 'layer' of computation to do the sum calculation ( rules --> rule --> domain ), which I was trying to avoid. Would keeping it at the Rule layer still be ok?

PR #5022, opened 2022-05-04
[MAINTENANCE] Refactor `DataAssistant` plotting to leverage utility dataclasses

cdkini:
- We often pair together a name, title, and type in our plotting logic - as such, it makes sense to group these fields together into dataclasses. Since these pairings share behavior, we can use an abstract class to define what behavior each of these dataclasses needs to implement.
  - Share behavior includes tooltip creation, axis creation, etc

### Definition of Done


great_expectations/rule_based_profiler/types/data_assistant_result/plot_util.py
```diff
@@ -0,0 +1,97 @@
+from abc import ABC, abstractmethod, abstractproperty
+from dataclasses import dataclass
+from typing import Optional, Union
+
+import altair as alt
+
+
+class PlotUtil(ABC):
```
cdkini: There's perhaps a better name for this 😄 suggestions are welcomed.

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
```diff
@@ -130,34 +135,44 @@ def get_line_chart(
...
+        batch_id_plot_util: BatchIdPlotUtil = BatchIdPlotUtil(
+            "batch_id", AltairDataTypes.NOMINAL.value
+        )
+        return DataAssistantResult._get_line_chart(
+            df=df,
+            metric_plot_util=metric_plot_util,
+            domain_plot_util=domain_plot_util,
+            batch_id_plot_util=batch_id_plot_util,
+            subtitle=subtitle,
+        )
```
cdkini: Pattern I'm proposing works as follows - public methods take in the raw arguments, formats them into these utility classes, and passes them downstream to a private method to perform the actual logic. This keeps our API consistent and removes the need for the user to know about these dataclasses.

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
```diff
@@ -199,18 +210,35 @@ def get_expect_values_to_be_between_chart(
...
-        title: Union[str, alt.TitleParams] = f"{metric_title} per {domain_title}"
-        if subtitle:
-            title = alt.TitleParams(title, subtitle=[subtitle])
+        title: alt.TitleParams = DataAssistantResult._determine_plot_title(
+            metric_title=metric_plot_util.title,
+            domain_title=domain_plot_util.title,
+            subtitle=subtitle,
+        )
 
-        batch_id: str = "batch_id"
```
cdkini, line 269: If we like the pattern used herein, I'm happy to apply it to `min_value` and `max_value` in this PR.

great_expectations/rule_based_profiler/types/data_assistant_result/plot_util.py
cdkini: Let's also block this PR until we get docstrings - I'm holding off on documentation until we deem this approach appropriate.

Edit: Just kidding the docstring checker won't let me merge even if I wanted to...

great_expectations/rule_based_profiler/types/data_assistant_result/plot_util.py
```diff
@@ -0,0 +1,101 @@
...
+            shorthand = self.domain_name
+        return axis(
+            shorthand,
+            type=self.domain_type,
+            title=self.title,
+        )
+
+
+@dataclass(frozen=True)
+class BatchIdPlotUtil(PlotUtil):
```
cdkini: Pros/cons of dataclasses here? These were originally larger value bags but it may make less sense now?

alexsherstinsky (approved):
LGTM (once Azure checks pass and others' feedback is incorporated).

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
```diff
@@ -219,48 +231,59 @@ def get_expect_values_to_be_between_chart(
...
+            min_value_plot_util=min_value_plot_util,
+            max_value_plot_util=max_value_plot_util,
+            subtitle=subtitle,
+        )
 
-        title: Union[str, alt.TitleParams] = f"{metric_title} per {domain_title}"
-        if subtitle:
-            title = alt.TitleParams(title, subtitle=[subtitle])
+    @staticmethod
+    def _get_expect_values_to_be_between_chart(
```
NathanFarmer: Can we change this to `_get_expect_domain_values_to_be_between_chart` to make it clear that it works for more than one domain?

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
```diff
@@ -22,6 +22,12 @@
 from great_expectations.rule_based_profiler.types.data_assistant_result.plot_result import (
     PlotResult,
 )
+from great_expectations.rule_based_profiler.types.data_assistant_result.plot_util import (
+    BatchIdPlotUtil,
+    DomainPlotUtil,
+    MetricPlotUtil,
+    PlotUtil,
+)
```
NathanFarmer: Maybe `plot_components` is the module name and classes are `BatchPlotComponent` (also removed the `Id`), `DomainPlotComponent`, `MetricPlotComponent`, and `ExpectationKwargPlotComponent`.

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
```diff
@@ -219,48 +231,59 @@ def get_expect_values_to_be_between_chart(
...
             .mark_line(color=line_color)
             .encode(
-                x=alt.X(
-                    domain_name,
-                    type=domain_type,
-                    title=domain_title,
-                ),
-                y=alt.Y(min_value, type=metric_type, title=metric_title),
+                x=domain_plot_util.plot_on_axis(),
+                y=metric_plot_util.plot_on_axis(shorthand=min_value_plot_util.name),
```
NathanFarmer: I believe what you are calling `shorthand` here is actually the required parameter `field`. The `shorthand` usually refers to a string with the field, and/or type, and/or aggregate (which you correctly described in the docstring for `plot_on_axis`). So if an actual shorthand is passed here I'm not sure what the override behavior is in the `alt.Y` object because you already set the type and title in the method using `self`.

Are we intending for the user to be able to override the Altair data type here? I think I understand why you use `metric_plot_util` and then override the name with `min_value_plot_util.name`, but why wouldn't we just use `y=min_value_plot_component.plot_on_axis()` here?

great_expectations/rule_based_profiler/types/data_assistant_result/plot_util.py
NathanFarmer: I'm happy with the dataclasses since it seems to clean up and simplify `DataAssistantResult`. They would probably be fine in the `data_assistant_result.py` module at the bottom though.

NathanFarmer (changes requested):
Great ideas! I know this is WIP, but I have some comments on the state of this remote version.

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
cdkini: I agree! Done ✅

great_expectations/rule_based_profiler/types/data_assistant_result/plot_util.py
cdkini: I'm erring on the side of a separate file just in case the types expand but let me know if you feel strongly about consolidation.

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
cdkini: yes! good call

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
cdkini: ooh okay I think I misread the documentation I appreciate the clarification.

Hmm let me look into this I remember having some issues but I'll look again with fresh eyes.

cdkini: the docs seem to say it's the shorthand? https://altair-viz.github.io/user_guide/generated/channels/altair.X.html

cdkini: also the reason why I passed in that value instead of running `min_value_plot_component.plot_on_axis()` is because we're mixing metric and min_value attrs.

This was the previous code:
```
y=alt.Y(min_value, type=metric_type, title=metric_title),
```
the `type` and `title` correspond to the `metric_plot_component` but `min_value` has to come from `min_value_plot_component`.

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
```diff
@@ -150,34 +156,48 @@ def get_line_chart(
...
+        )
+        batch_id_component: BatchPlotComponent = BatchPlotComponent(
+            "batch_id", AltairDataTypes.NOMINAL.value
+        )
+        return DataAssistantResult._get_line_chart(
+            df=df,
+            metric_component=metric_component,
+            domain_component=domain_component,
+            batch_id_component=batch_id_component,
+            subtitle=subtitle,
```
NathanFarmer: Can `subtitle` become an optional property of `DomainPlotComponent`? Right now, a subtitle is needed for all domain types except `MetricDomainTypes.TABLE`.

NathanFarmer (changes requested):
One requested change and then I think we are good to go!

great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
```diff
@@ -285,26 +306,22 @@ def get_expect_values_to_be_between_chart(
...
-                x=alt.X(
-                    domain_name,
-                    type=domain_type,
-                    title=domain_title,
-                ),
-                y=alt.Y(min_value, title=metric_title, type=metric_type),
-                y2=alt.Y2(max_value, title=metric_title),
+                x=domain_component.plot_on_axis(),
+                y=metric_component.plot_on_axis(shorthand=min_value_component.name),
+                y2=alt.Y2(max_value_component.name, title=metric_component.title),
```
NathanFarmer: Shouldn't this be `field=min_value_component.name`? I'm struggling to follow the logic here. I feel like we should have a sub-class called `ExpectationKwargPlotComponent` that is used here for `min_value` and `max_value`.

NathanFarmer (changes requested):
We missed updating `shorthand` to `field` and I'm still thinking it would be much clearer to have an `ExpectationKwargPlotComponent` instead of re-using pieces of `MetricPlotComponent` for plotting the kwargs.

NathanFarmer (approved):
LGTM!

PR #5024, opened 2022-05-04
[FEATURE] Splitters and limit sample work on AWS Athena

anthonyburdi:
- Add tests for splitter and limit sample functionality for AWS Athena.
- Refactor code generating connection strings into a single method.
- Add sampling support at Asset level (see also https://github.com/great-expectations/great_expectations/pull/5092)

### Definition of Done


alexsherstinsky (approved):
LGTM (provided Azure checks pass).

tests/integration/db/test_sql_data_sampling.py
```diff
@@ -69,35 +54,54 @@ def _load_data(
...
+        table_name: str = "ten_trips_from_each_month"
+        test_df: pd.DataFrame = load_and_concatenate_csvs(
+            csv_paths=[
+                f"./data/ten_trips_from_each_month/yellow_tripdata_sample_10_trips_from_each_month.csv"
+            ],
+            convert_column_names_to_datetime=["pickup_datetime", "dropoff_datetime"],
+            load_full_dataset=True,
+        )
+    else:
+        print("Preemptively cleaning old tables")
```
cdkini, line 87: pros/cons of making this a log statement? same throughout these tests.

tests/test_utils.py
```diff
@@ -49,8 +50,9 @@
     Select = None
     SQLAlchemyError = None
 
-logger = logging.getLogger(__name__)
 
+logger = logging.getLogger(__name__)
+yaml_handler: YAMLHandler = YAMLHandler()
```
cdkini, line 55: ❤️

tests/test_utils.py
```diff
@@ -603,6 +605,23 @@ def load_data_into_test_database(
...
+                df=all_dfs_concatenated,
+                table_name=table_name,
+                connection=connection,
+            )
+            return return_value
+        except SQLAlchemyError as e:
+            logger.error(
+                """Docs integration tests encountered an error while loading test-data into test-database."""
+            )
+            raise
```
cdkini, line 621: Will this ever log the connection string? Should we be concerned about security at all?

tests/test_utils.py
```diff
@@ -777,6 +833,56 @@ def clean_athena_db(connection_string: str, db_name: str, table_to_keep: str) ->
...
+
+    return f"awsathena+rest://@athena.us-east-1.amazonaws.com/{ATHENA_DB_NAME}?s3_staging_dir={ATHENA_STAGING_S3}"
+
+
+def get_connection_string_and_dialect(
+    athena_db_name_env_var: str = "ATHENA_DB_NAME",
+) -> Tuple[str, str]:
+
+    with open("./connection_string.yml") as f:
+        db_config: dict = yaml_handler.load(f)
```
cdkini, line 871: Should this be `file_relative_path`? Will we ever run this from another location?

cdkini (approved):

tests/test_utils.py
anthonyburdi: Hmm good question. This method is usually invoked via a script that is typically run as a test using test_script_runner.py so this file is relative to the environment where this is invoked. I think there is definitely an opportunity to improve this integration test part of the codebase, maybe using containers instead of running each as a module. Would be worth discussing.

tests/test_utils.py
anthonyburdi: This is a really good question. It does. I just checked. For example: `sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) connection to server at "localhost" (::1), port 5432 failed: FATAL:  database "test_ci_invalid" does not exist.` Let's discuss with the team.

tests/integration/db/test_sql_data_sampling.py
anthonyburdi: Hmm good question. I was following some other tests but glad you are pointing this out. Probably logging is better so we can set verbosity. I'm not sure how these messages show up in a) local runs b) in azure, would be good to test before making changes and make sure it meets our needs.

PR #5071, opened 2022-05-09
[MAINTENANCE] Check that a passed string is parseable as an integer (mssql limit param)

anthonyburdi:
- For mssql we perform string substitution for the limit parameter because it is not substituted during sqlalchemy `query.compile()`. This PR adds an additional check that if the limit parameter `n`  is passed as a string, we check to make sure it contains only digits.
- TODO note left in as there may still be a better method to handle this substitution.

### Definition of Done

Thank you for submitting!


alexsherstinsky (approved):
LGTM

PR #5124, opened 2022-05-15
[BUGFIX] For DataAssistants, added try-except for Notebook tests

Shinnnyshinshin:
## What went wrong? 
- There is an optional cell in the `RBP` and `DataAssistant` Notebook that deletes files that are usually generated by the notebook (like ExpectationSuites, and `.ge_cloud_id` files). This cell is normally commented out, but wasn't in the previous PR 😅 
- In the `test_run_profiler_notebook.py` tests, the is a part of the test that tries to delete these same files.. since they were deleted by running the Notebook, the test was giving  a `FileNotFoundError`. 

## What does this PR fix? 
- Re-commented Optional cells in `DataAssistant` Notebook that was causing tests to fail
- Wrapped the file deletion step in a try-except for `tests/integration/profiling/rule_based_profilers/test_run_profiler_notebook.py`
- [x] My code follows the Great Expectations [style guide]

Thank you for submitting!


tests/test_fixtures/rule_based_profiler/example_notebooks/DataAssistants_Instantiation_And_Running.ipynb
```diff
@@ -698,15 +707,15 @@
...
    "outputs": [],
    "source": [
-    "import shutil, os\n",
-    "shutil.rmtree(\"great_expectations/expectations/tmp\")\n",
-    "os.remove(\"great_expectations/expectations/.ge_store_backend_id\")\n",
-    "os.remove(\"great_expectations/expectations/taxi_data_suite.json\")"
+    "#import shutil, os\n",
+    "#shutil.rmtree(\"great_expectations/expectations/tmp\")\n",
+    "#os.remove(\"great_expectations/expectations/.ge_store_backend_id\")\n",
+    "#os.remove(\"great_expectations/expectations/taxi_data_suite.json\")"
```
Shinnnyshinshin, line 718: Turns out this was the cause of the bug 😓

tests/integration/profiling/rule_based_profilers/test_run_profiler_notebook.py
```diff
@@ -46,10 +46,17 @@ def test_run_rbp_notebook(tmp_path):
...
+        shutil.rmtree(os.path.join(base_dir, "great_expectations/expectations/tmp"))
+        os.remove(
+            os.path.join(
+                base_dir, "great_expectations/expectations/.ge_store_backend_id"
+            )
+        )
+    except FileNotFoundError:
+        # This just means the files were already deleted by running the optional last cell in the notebook.
+        # Therefore we allow the test to pass.
+        pass
```
cdkini: I think this file deletion should occur nested under the `finally` block - if anything goes wrong earlier on in the test, these files will not be deleted.

Also, do we need to write the results to a temp notebook? What are your thoughts on devs just running the original notebook to debug?

Finally, could we please log when we hit the `FileNotFoundError` so that comment you wrote is presented in stdout?

tests/integration/profiling/rule_based_profilers/test_run_profiler_notebook.py
```diff
@@ -89,10 +96,19 @@ def test_run_data_assistants_notebook(tmp_path):
...
-        os.path.join(base_dir, "great_expectations/expectations/.ge_store_backend_id")
-    )
-    os.remove(
-        os.path.join(base_dir, "great_expectations/expectations/taxi_data_suite.json")
-    )
+    try:
+        shutil.rmtree(os.path.join(base_dir, "great_expectations/expectations/tmp"))
+        os.remove(
+            os.path.join(
+                base_dir, "great_expectations/expectations/.ge_store_backend_id"
```
cdkini: Same notes here

cdkini (changes requested):
Minor changes requested - if we move the file deletion to the `finally` block, we will ensure that those files are always deleted (regardless of errors or exceptions).

tests/integration/profiling/rule_based_profilers/test_run_profiler_notebook.py
Shinnnyshinshin: Great suggestions.I thought the point on adding the `FileNotFoundError` to the log was fantastic, so my most recent commit has that change. I have some questions about what you had in mind for `finally` block and was wondering if you had time to discuss tomorrow?

cdkini (approved):

tests/integration/profiling/rule_based_profilers/test_run_profiler_notebook.py
Shinnnyshinshin: thank you for helping me see the light ⚡ 🌞

PR #5153, opened 2022-05-18
[MAINTENANCE] Allow for include/exclude column name args in DataAssistant `run`

cdkini:
- Enable include/exclude column behavior from the RBP and the assistant.
  - Allows for additional customization - can now do column filtering at the Result level and the plotting level.

Previous Design Review notes:
- JIRA: https://superconductive.atlassian.net/browse/GREAT-905
- Loom: https://www.loom.com/share/bf7708e293714916ad9d511956ff9046

### Definition of Done

Thank you for submitting!

### DIFF
--- a/great_expectations/rule_based_profiler/rule_based_profiler.py
+++ b/great_expectations/rule_based_profiler/rule_based_profiler.py
@@ -38,6 +38,9 @@ from great_expectations.rule_based_profiler.config.base import (
     expectationConfigurationBuilderConfigSchema,
     parameterBuilderConfigSchema,
 )
+from great_expectations.rule_based_profiler.domain_builder.column_domain_builder import (
+    ColumnDomainBuilder,
+)
 from great_expectations.rule_based_profiler.domain_builder.domain_builder import (
     DomainBuilder,
 )
@@ -218,6 +221,8 @@ class BaseRuleBasedProfiler(ConfigPeer):
         batch_request: Optional[Union[BatchRequestBase, dict]] = None,
         recompute_existing_parameter_values: bool = False,
         reconciliation_directives: ReconciliationDirectives = DEFAULT_RECONCILATION_DIRECTIVES,
+        include_column_names: Optional[List[str]] = None,
+        exclude_column_names: Optional[List[str]] = None,
     ) -> RuleBasedProfilerResult:
         """
         Executes and collects "RuleState" side-effect from all "Rule" objects of this "RuleBasedProfiler".
@@ -233,6 +238,11 @@ class BaseRuleBasedProfiler(ConfigPeer):
         Returns:
             "RuleBasedProfilerResult" dataclass object, containing essential outputs of profiling.
         """
+        if include_column_names is not None and exclude_column_names is not None:
+            raise ValueError(
+                "You may either use `include_column_names` or `exclude_column_names` (but not both)."
+            )
+
         # Check to see if the user has disabled progress bars
         disable = False
         if self._data_context:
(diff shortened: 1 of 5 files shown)


great_expectations/rule_based_profiler/types/data_assistant_result/data_assistant_result.py
```diff
@@ -118,6 +118,42 @@ def get_attributed_metrics_by_domain(
...
+            raise ValueError(
+                "You may either use `include_column_names` or `exclude_column_names` (but not both)."
+            )
+
+        if self.expectation_configurations is None:
+            return
+
+        filtered_expectation_configurations: List[ExpectationConfiguration] = []
+
+        for expectation_configuration in self.expectation_configurations:
```
alexsherstinsky: @cdkini We also need to apply filtering to the `metrics_by_domain` property (whereby the `domain_type` field of the `Domain` objects, used as keys, may have the `column` field defined).  Some thinking may need to be applied to two-column and multi-column domains (and the same for the `ExpectationConfiguration` filtering, already in place.  Thank you!

alexsherstinsky (changes requested):
@cdkini I left some comments about `Domain` columns filtering and two-column/multi-column situations.  Thanks.

NathanFarmer (changes requested):
My reading of [GREAT-905](https://superconductive.atlassian.net/browse/GREAT-905) is that  the columns should be included/excluded from the `DataAssistant` run, not just the results.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -233,6 +238,11 @@ def run(
         Returns:
             "RuleBasedProfilerResult" dataclass object, containing essential outputs of profiling.
         """
+        if include_column_names is not None and exclude_column_names is not None:
+            raise ValueError(
+                "You may either use `include_column_names` or `exclude_column_names` (but not both)."
+            )
```
cdkini, line 244: The check is added in both places because as a user, I should be able to use the functionality from both the RBP and the assistant.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -528,6 +545,20 @@ def _reconcile_profiler_rules_as_dict(
...
+        include_column_names: Optional[List[str]],
+        exclude_column_names: Optional[List[str]],
+    ) -> None:
+        for rule in rules:
+            domain_builder = rule.domain_builder
+            if domain_builder is not None and isinstance(
+                domain_builder, ColumnDomainBuilder
+            ):
+                domain_builder.include_column_names = include_column_names
+                domain_builder.exclude_column_names = exclude_column_names
```
NathanFarmer, line 560: ❤️

NathanFarmer (approved):
LGTM!

great_expectations/rule_based_profiler/rule_based_profiler.py
alexsherstinsky: @cdkini I do not feel good about this approach, unless this is the only way forward.  Is there any chance we could not merge it and put our minds together?  One thing is to "sanitize/filter" the result.  An entirely another thing is to update the RBP run method. /cc @NathanFarmer -- thank you.

alexsherstinsky (changes requested):
@cdkini @NathanFarmer I would like to request that we hold the merging off to give another potential avenue a chance.  We may not succeed, but I feel that it would be worth the wait to try.  Thank you!

PR #5159, opened 2022-05-18
[FEATURE] Helper method to determine whether Expectation is self-initializing 

Shinnnyshinshin:
-  Helper method that, given the name of an Expectation, returns a boolean that represents whether an Expectation can be auto-intialized.

**The following would return `False`**
 and print : `The Expectation expect_column_to_exist in not able to be self-initialized.`

```python
validator.is_expectation_self_initializing(
            name="expect_column_distinct_values_to_be_in_set"
        )
```

**The following would return `True`**
and print `The Expectation expect_column_mean_to_be_between is able to be self-initialized. Please run by using the auto=True parameter.`
```python
validator.is_expectation_self_initializing(
            name="expect_column_mean_to_be_between"
        )
```


great_expectations/validator/validator.py
```diff
@@ -2162,6 +2162,33 @@ def head(
...
+    @staticmethod
+    def is_expectation_self_initializing(name: str) -> bool:
+        """
+        Given the name of an Expectation, returns a boolean that represents whether an Expectation can be auto-intialized.
+
+        Args:
+            name (str): name of Expectation
+
+        Returns:
+            boolean that represents whether an Expectation can be auto-initialized. Information also outputted to logger.
```
cdkini: We're printing instead of logging right? Do we always want to emit this information to the user? If not, let's just a log at an appropriate level.

Also, does this have to go on the validator? Can it maybe be a static method on the Expectation class? Just brainstorming but I know the validator is quite bloated.

cdkini (approved):

great_expectations/validator/validator.py
Shinnnyshinshin: Great points. 

* I'd like to propose printing instead of logging. I think the most common usecase for this would be in a Notebook, and typically `warning` and above is displayed in Jupyter Notebooks by default.. which I dont think would be appropriate in this case. 

* Regarding your point about Expectation, I tried this yesterday and ran into some circular imports .. I can take another stab at it before merging (I think it makes more sense to have it at the `Expectation` level too)

Shinnnyshinshin:
@cdkini I've updated the PR to move the code the `Expectations` class, and more strictly test the `staticmethod`. If you get a chance, please have another look :)

great_expectations/exceptions/exceptions.py
```diff
@@ -169,6 +169,10 @@ class InvalidExpectationConfigurationError(GreatExpectationsError):
     pass
 
 
+class ExpectationNotFoundError(GreatExpectationsError):
+    pass
```
cdkini, line 173: ❤️

tests/validator/test_validator.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
cdkini: Unused?

PR #5169, opened 2022-05-20
[FEATURE] Example Notebook for self-initializing `Expectations`

Shinnnyshinshin:
Changes proposed in this pull request:
This PR contains a notebook that exercises a code path for the self-initializing `Expectation`, `expect_column_mean_to_be_between()`.

The notebook, `SelfInitializingExpectations.ipynb`, shows how how to use the self-initializing `Expectations`, and against a datasource containing 12 batches (corresponding to 1 year) of taxi_data.

The notebook describes the following major steps:

- Setting up Datasource to connect to taxi_data.
- Configuring BatchRequest to retrieve 12 Batches of taxi_data
- Running `expect_column_mean_to_be_between()` with `auto=True`.
- Description of `RuleBasedProfiler` configuration used to make the `Expectation` self-initializing. 
- Adds integration test for notebook
- Closes GREAT-693

### Definition of Done
Please delete options that are not relevant.

- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added [unit tests](https://docs.greatexpectations.io/docs/contributing/contributing_test#writing-unit-and-integration-tests) where applicable and made sure that new and existing tests are passing.
- [x] I have run any local integration tests and made sure that nothing is broken.

PR #5171, opened 2022-05-20
[MAINTENANCE] Helper method for RBP Notebook tests that does clean-up

Shinnnyshinshin:
- Follow up to PR #5169 
- Adds clean-up code to helper method that leverages `os.path.exists()` as per @anthonyburdi 's suggestion
-

### Definition of Done

Thank you for submitting!

### DIFF
--- a/tests/integration/profiling/rule_based_profilers/test_run_profiler_notebook.py
+++ b/tests/integration/profiling/rule_based_profilers/test_run_profiler_notebook.py
@@ -11,19 +11,21 @@ from great_expectations.data_context.util import file_relative_path
 logger = logging.getLogger(__name__)
 
 
-def clean_up_test_files(paths: List[str]) -> None:
+def clean_up_test_files(base_dir: str, paths: List[str]) -> None:
     """
     Helper method to clean-up the files created by tests
     Args:
+        base_dir str: tmp path created by the test
         paths List(str): paths or directories to delete
     """
     for path in paths:
-        if not os.path.exists(path):
+        full_path: str = os.path.join(base_dir, path)
+        if not os.path.exists(full_path):
             continue
-        if os.path.isdir(path):
-            shutil.rmtree(path)
+        if os.path.isdir(full_path):
+            shutil.rmtree(full_path)
         else:
-            os.remove(path)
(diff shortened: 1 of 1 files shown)


tests/integration/profiling/rule_based_profilers/test_run_profiler_notebook.py
```diff
@@ -10,6 +11,30 @@
 logger = logging.getLogger(__name__)
 
 
+def clean_up_test_files(
+    base_dir: str, additional_files_to_clean_up: Optional[List[str]] = None
+) -> None:
```
cdkini: Could we possibly make this a bit more generic? Rather than defining specific files and directories here, we rely on the arguments to determine what needs to be removed?

```python
def clean_up_test_files(paths: List[str]) -> None:
    for path in paths:
        if not os.path.exists(path):
            continue
        if os.path.isdir(path):
            # remove dir
        else:
            # remove file
```

Definitely a personal opinion but I like explicitly seeing which files are going to be cleaned up in that `finally` block. Perhaps a bit of code duplication is introduced but I think it becomes abundantly clear that we're cleaning up X, Y, and Z file and have no concern about side effects.

cdkini (approved):

tests/integration/profiling/rule_based_profilers/test_run_profiler_notebook.py
```diff
@@ -148,21 +155,12 @@ def test_run_self_initializing_expectations_notebook(tmp_path):
...
-                "Files were already deleted by running the optional last cell in the notebook. "
-                "We therefore allow the test to pass"
+
+    clean_up_test_files(
+        base_dir=base_dir,
+        additional_files_to_clean_up=[
+            os.path.join(
+                base_dir, "great_expectations/expectations/new_expectation_suite.json"
             )
+        ],
```
cdkini: If sticking with this current API, can we do this `join` in the actual method? We have access to the `base_dir` through the args right?

Shinnnyshinshin: done!

PR #5190, opened 2022-05-24
[MAINTENANCE] Don't include infinity in rendered string for diagnostics

kenwade4:
Prevent gallery JSON file from having inf as it causes "invalid JSON" parsing error in frontend

- Update _get_rendered_result_as_string method to return empty string if inf is contained


austiezr (approved):
LGTM! 🚀

rdodev (approved):

PR #5195, opened 2022-05-25
[BUGFIX] OnboardingDataAssistant and Underlying Metrics: Add Defensive Programming Into Metric Implementations So As To Avoid Warnings About Incompatible Data

alexsherstinsky:
### Scope
* Fix `ColumnValuesLengthMin` and `ColumnValuesLengthMax` metric providers to filter out `NULL` columns (cannot compute length of a string that turns into a floating point `NaN` value).
* Fix `ColumnMedian` metric provider to exit uniformly when all values in the column are `NULL` (return `0.0` in this case).
* Update `CANDIDATE_REGEX` in `RegexPatternStringParameterBuilder` to use Matching Groups instead of Capture Groups (the latter generate an "advice" warning from `Python` `RegEx` library).

### Remark
This fixes most of the warnings.  Next is to address warnings and errors, associated with `histogram` computation as well as any remaining exceptions in `Chart` graphing.

- JIRA: GREAT-822/GREAT-969
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!

### DIFF
--- a/great_expectations/expectations/metrics/column_aggregate_metrics/column_median.py
+++ b/great_expectations/expectations/metrics/column_aggregate_metrics/column_median.py
@@ -27,6 +27,11 @@ class ColumnMedian(ColumnAggregateMetricProvider):
     @column_aggregate_value(engine=PandasExecutionEngine)
     def _pandas(cls, column, **kwargs):
         """Pandas Median Implementation"""
+        metrics: Dict[str, Any] = kwargs.get("metrics") or {}
+        nonnull_count = metrics.get("column_values.nonnull.count")
+        if not nonnull_count:
+            return 0.0
+
         return column.median()
 
     @metric_value(engine=SqlAlchemyExecutionEngine, metric_fn_type="value")
@@ -48,10 +53,12 @@ class ColumnMedian(ColumnAggregateMetricProvider):
         column_name = accessor_domain_kwargs["column"]
         column = sa.column(column_name)
         sqlalchemy_engine = execution_engine.engine
+
         """SqlAlchemy Median Implementation"""
         nonnull_count = metrics.get("column_values.nonnull.count")
         if not nonnull_count:
             return None
+
         element_values = sqlalchemy_engine.execute(
             sa.select([column])
             .order_by(column)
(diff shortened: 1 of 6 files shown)


great_expectations/expectations/metrics/column_aggregate_metrics/column_median.py
```diff
@@ -27,6 +27,11 @@ class ColumnMedian(ColumnAggregateMetricProvider):
     @column_aggregate_value(engine=PandasExecutionEngine)
     def _pandas(cls, column, **kwargs):
         """Pandas Median Implementation"""
+        metrics: Dict[str, Any] = kwargs.get("metrics") or {}
```
cdkini, line 30: Could we use `get("metrics", {})` instead?

great_expectations/expectations/metrics/column_aggregate_metrics/column_median.py
```diff
@@ -27,6 +27,11 @@ class ColumnMedian(ColumnAggregateMetricProvider):
...
+        nonnull_count = metrics.get("column_values.nonnull.count")
+        if not nonnull_count:
+            return 0.0
```
cdkini, line 33: Nitpick - more personal preference but when not actually using the value, I think `if x not in metrics` is more appropriate. Happy to leave it as it though.

cdkini (approved):

great_expectations/expectations/metrics/column_aggregate_metrics/column_median.py
alexsherstinsky: @cdkini I believe that we could do that.  I will update in the next PR I am working on.  Thank you!

great_expectations/expectations/metrics/column_aggregate_metrics/column_median.py
alexsherstinsky: @cdkini I believe that here, even if it is there but returns `0` (or `None`), the check must return `True` (so this pattern, which we have everywhere for `metrics`, covers all of these needs).  Thanks!

PR #5196, opened 2022-05-25
Comment out kl_divergence for build_gallery

kenwade4:
- Comment out kl_divergence for build_gallery


rdodev:
@kenwade4 do you expect this to be a one-off or do you think this is worth generalizing? For example

```
skip_list = ['this_expectation', 'that_expectation'] # These expectations will not be in the gallery

if expectation in skip_list:
 continue
```

kenwade4:
Ideally we won't need to skip any Expectations, but this one is causing too many issues at the moment.

rdodev (approved):
Approving as one off, consider generalizing if we need to skip more than one.

PR #5206, opened 2022-05-26
[FEATURE] `DatasourceStore`

cdkini:
- New `DatasourceStore` that mirrors existing Store logic
- Modification of `GeCloudStoreBackend` to allow for new resource type.
- Tests to ensure proper serialization.

JIRA: https://superconductive.atlassian.net/browse/GREAT-910

### Definition of Done

Thank you for submitting!


tests/data_context/store/test_datasource_store.py
```diff
@@ -0,0 +1,140 @@
...
+    empty_datasource_store: DatasourceStore,
+    datasource_config: DatasourceConfig,
+    datasource_key: ConfigurationIdentifier,
+):
+    assert len(empty_datasource_store.list_keys()) == 0
+    empty_datasource_store.set(key=datasource_key, value=datasource_config)
+    assert len(empty_datasource_store.list_keys()) == 1
+
+
+def test_datasource_store_integration(
```
anthonyburdi: It might make sense to mark this as an integration test so we can use flags to exclude/include running it in CI pipelines as we organize our tests a bit more.
```suggestion
@pytest.mark.integration
def test_datasource_store_integration(
```

anthonyburdi (approved):
LGTM! One minor non blocking suggestion.

tests/data_context/store/test_datasource_store.py
cdkini: I love that! Great call, I've committed your suggestion.

great_expectations/data_context/store/datasource_store.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class DatasourceStore(ConfigurationStore):
+    """
+    A DatasourceStore manages Datasources for the DataContext.
+    """
+
+    _configuration_class = DatasourceConfig
+
+    def serialization_self_check(self, pretty_print: bool) -> None:
```
donaldheppner: why is this here? This is test code

great_expectations/data_context/types/base.py
```diff
@@ -976,6 +891,108 @@ def make_datasource_config(self, data, **kwargs):
         return DatasourceConfig(**data)
 
 
+class DatasourceConfig(BaseYamlConfig):
```
donaldheppner: Given this type does not inherit from `BaseYamlConfig` today, can we avoid introducing it? Eventually we don't want any of our config objects to inherit from BaseYamlConfig

donaldheppner (changes requested):

great_expectations/data_context/store/datasource_store.py
cdkini: that's a very good question 😄 I just followed the existing pattern but I do agree that this shouldn't be in our source code. I was under the impression that this was a necessary abstract method to implement to fulfill the parent's contract but I think we can forgo this.

I'll move this logic over to a test.

sdebruyn:
Shouldn't this PR include some documentation about this new feature?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -11,4 +11,6 @@ class AbstractDataContext(ABC):
     TODO: eventually the dependency on ConfigPeer will be removed and this will become a pure ABC.
     """
 
-    pass
+    @abstractmethod
+    def _init_datasource_store(self) -> None:
+        raise NotImplementedError
```
cdkini: To be fleshed out in a subsequent PR

scripts/check_type_hint_coverage.py
```diff
@@ -3,7 +3,7 @@
 from typing import Dict, List, Optional
 
 TYPE_HINT_ERROR_THRESHOLD: int = (
-    2548  # This number is to be reduced as we annotate more functions!
+    2547  # This number is to be reduced as we annotate more functions!
```
donaldheppner: 🙌

tests/data_context/store/test_datasource_store.py
```diff
@@ -0,0 +1,158 @@
...
+    empty_datasource_store: DatasourceStore, datasource_config: DatasourceConfig
+) -> None:
+    store: DatasourceStore = empty_datasource_store
+
+    key: StringKey = StringKey(key="my_datasource")
+    store.set(key=key, value=datasource_config)
+    res: DatasourceConfig = store.get(key=key)
+
+    assert isinstance(res, DatasourceConfig)
+    assert DatasourceConfigSchema().dump(res) == datasource_config.to_dict()
```
donaldheppner: asserting the key here should be sufficient?

tests/data_context/store/test_datasource_store.py
```diff
@@ -0,0 +1,158 @@
...
+    key: GeCloudIdentifier = GeCloudIdentifier(
+        resource_type="datasource", ge_cloud_id="foobarbaz"
+    )
+
+    with patch("requests.patch", autospec=True) as mock_patch:
+        type(mock_patch.return_value).status_code = PropertyMock(return_value=200)
+
+        store.set(key=key, value=datasource_config)
+
+        mock_patch.assert_called_with(
```
donaldheppner, line 120: generally love this, can we leverage the datasource_config() fixture to avoid duplicating the payload here?

tests/data_context/store/test_datasource_store.py
cdkini: so unfortunately that doesn't seem to be the case - while both items are indeed `DatasourceConfig`, the fixture has dictionary attributes and the `res` obj has `Config` types.

Basically, `tripdata_monthly_configured` is of type `dict` in our fixture while it is of type `DataConnectoConfig` once parsed by the store.

Here's a Loom to elaborate: https://www.loom.com/share/983f45c656ce4157a8dd905ee085978d

tests/data_context/store/test_datasource_store.py
cdkini: that's a great call! done!

tests/data_context/store/test_datasource_store.py
cdkini: I also don't think that `DatasourceConfig` has implemented `__eq__` so it will compare by memory address instead of actual values.

tests/data_context/store/test_datasource_store.py
```diff
@@ -124,30 +124,8 @@ def test_datasource_store_retrieval_cloud_mode(
...
-                                }
-                            },
-                            "execution_engine": {
-                                "module_name": "great_expectations.execution_engine",
-                                "class_name": "PandasExecutionEngine",
-                            },
-                            "class_name": "Datasource",
-                        },
-                        "organization_id": "bd20fead-2c31-4392-bcd1-f1e87ad5a79c",
+                        "datasource_config": datasource_config.to_dict(),
```
donaldheppner, line 127: :heart::heart::heart::heart::heart::heart::heart:

cdkini:
@sdebruyn apologies for the delayed response! This feature is very much under development and will still take some work before we're ready to publish it and present it to the larger community.

We'll be sure to keep you updated through Slack and our blog posts as things progress 😄

tests/data_context/store/test_datasource_store.py
donaldheppner: sorry for not being clear, I meant: could we assert key values for equality vs. the whole payload and still be testing what we intend to be

donaldheppner (approved):

PR #5216, opened 2022-05-27
[FEATURE] `InlineStoreBackend`

cdkini:
- Introduction of `InlineStoreBackend` to interface between stores and the `great_expectations.yml` config file.
- Tests to cover basic CRUD behavior (ensure both in-memory object and filesystem are impacted)
- NOTE: This is mean to be a store-agnostic implementation. The primary usage of this backend will be with the `DatasourceStore` but that integration is part of a later ticket (GREAT-926)

Jira: https://superconductive.atlassian.net/browse/GREAT-970

### Definition of Done

Thank you for submitting!


great_expectations/data_context/store/__init__.py
```diff
@@ -36,11 +37,13 @@
     ".html_site_store",
     ".profiler_store",
     ".metric_store",
+    ".datasource_store",
```
cdkini: Missed this in my previous store implementation 😬

great_expectations/data_context/types/base.py
```diff
@@ -2151,6 +2151,10 @@ def get_schema_class(cls):  # type: ignore[no-untyped-def]
     def config_version(self):
         return self._config_version
 
+    @config_version.setter
+    def config_version(self, config_version: float) -> None:
+        self._config_version = config_version
```
cdkini, line 2156: Because of the property above, we need an explicit setter here to fully enable read/write access for all attributes on this config object.

great_expectations/data_context/store/__init__.py
anthonyburdi: All good - nice job catching it. Is there a good test to add here so we catch issues like this in the future? Just off the cuff, there might be a better way: maybe a test for inclusion of all modules with classes that inherit from `Store`?

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -0,0 +1,91 @@
...
+        self._data_context.config[config_var] = value
+        self._data_context._save_project_config()
+
+    def _move(
+        self, source_key: Tuple[str, ...], dest_key: Tuple[str, ...], **kwargs: dict
+    ) -> None:
+        raise StoreBackendError(
+            "InlineStoreBackend does not support moving of keys; the DataContext's config variables schema is immutable"
+        )
+
```
anthonyburdi, line 62: 👍

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -0,0 +1,91 @@
+from typing import Any, List, Optional, Tuple
+
+from great_expectations.data_context.store.store_backend import StoreBackend
+from great_expectations.exceptions.exceptions import StoreBackendError
+from great_expectations.util import filter_properties_dict
+
+
+class InlineStoreBackend(StoreBackend):
```
alexsherstinsky, line 8: @cdkini @donaldheppner This actually underscores one of the reasons for my uncertainty about the `InlineStoreBackend` -- it appears to need `DataContext` -- is this desirable?  Should we using `BaseDataContext` instead?  Better still, I feel that not having `InlineStoreBackend` in favor of a standard upgrade procedure might be a cleaner approach.  Thank you for consideration.

anthonyburdi (approved):

great_expectations/data_context/store/inline_store_backend.py
donaldheppner: In this case I think the DataContext reference makes sense since InlineStoreBackend is intended to store values in the DataContext (the DataContext is effectively the store, in the same way S3 is the store for the S3StoreBackend).

I'd rather the InlineStoreBackend reference DataContext and use its methods to read and manipulate values than have InlineStoreBackend bypass DataContext and manipulate the underlying values directly.

PR #5239, opened 2022-06-03
[MAINTENANCE] show more test output and minor fixes

kenwade4:
Show more useful output for failed tests when building the gallery. Also add some fixes so that more Expectations have all/most tests passing for the new datasource view of the gallery

- Show backend and test names in `check_json_test_result`
- For AssertionErrors, actually show the failed comparisons
- Ensure Polygon/Point don't break JSON serializing
- More use of np.allclose for float comparisons

### Definition of Done


great_expectations/expectations/expectation.py
```diff
@@ -1249,10 +1249,14 @@ def _get_test_results(
                 test=exp_test["test"],
                 raise_exception=False,
             )
+            print(f"\n({exp_test['backend']}, {exp_test['test']['title']})")
```
rdodev, line 1252: What if the user doesn't want this output?

kenwade4: This is strictly for Expectation Diagnostics and building the gallery... this output won't be seen in a "normal user workflow"

rdodev (approved):

PR #5244, opened 2022-06-05
[MAINTENANCE] Misc updates to improve security and automation of the weekly release process

cdkini:
- More automation (auto tagging after release PR goes through)
- Stricter tests to enforce proper release schedule schema
- Updated Azure YAML to run full test suite on BOTH prep and release cut.

### Definition of Done

### DIFF
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -1,0 +1,23 @@
+name: Release Cut
+on:
+  - pull_request
+
+jobs:
+  prep:
+    if: contains(github.event.pull_request.title, 'release')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Install ge_releaser
+        run: |
+          git clone https://github.com/superconductive/ge_releaser.git
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          pip install -e ge_releaser
+      - name: Tag release commit
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          ge_releaser tag
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -5,7 +5,7 @@ on:
       - closed
 
 jobs:
-  prep:
+  cut:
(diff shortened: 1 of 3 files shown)


tests/test_release.py
```diff
@@ -38,6 +38,9 @@ def test_release_schedule_adheres_to_schema(
 
     for date, release_version in release_schedule.items():
 
+        # No old releases should exist within the release schedule (deleted during release process)
+        assert date >= today
+
```
cdkini, line 43: The extra security is a nice-to-have. The `ge_releaser` tool has some input validation but bolstering this test can't hurt.

azure-pipelines.yml
```diff
@@ -52,6 +52,7 @@ resources:
 
 # The pipeline is run under two primary conditions: if cutting a release or as scheduled by the above cron job.
 variables:
+  isReleasePrep: $[contains(variables['Build.SourceBranch'], 'release')]
```
cdkini, line 55: As it stands, our release process works as follows:
* Create a prep PR with changelogs and other misc updates
  - This triggers `dgtest`, which doesn't pick up any tests, and allows the prep PR to merge 100% of the time
* Tag that particular prep PR commit to cut the release
  - This triggers `pytest`, which will run the full suite and validates that our build is not broken.

The issue here is what if the build is actually broken when we go to tag? We've already updated our changelogs and our git history now has this release commit that isn't actually the release. The proposed change here will run the full `pytest` suite on both the release PR and the tag (duplicative but safer).

.github/workflows/release-cut.yml
```diff
@@ -0,0 +1,25 @@
+name: Release Cut
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  prep:
+    if: github.event.pull_request.merged == true && contains(github.event.pull_request.title, 'RELEASE')
```
cdkini, line 9: Once the release prep PR goes through, this action will trigger and run the `ge_releaser tag` command; this will kick off a release-enabled Azure build and automatically publish to PyPI.

tests/test_release.py
Shinnnyshinshin: ❤️

Shinnnyshinshin (approved):
Love it, and really loving how far we have come with this work :) even with the additional testing, I feel like we are so close to the world of "set it and forget it"

PR #5270, opened 2022-06-08
[MAINTENANCE] Remove condition from `autoupdate` GitHub action

cdkini:
- This action never actually runs because the condition is never true - since it should only run once `auto-merge` is enabled, I've just deleted the condition and am hoping it now works 😄 

### Definition of Done


Shinnnyshinshin (approved):
LGTM :)

PR #5286, opened 2022-06-10
[MAINTENANCE] Refactor of `_apply_global_config_overrides()` method to return config

Shinnnyshinshin:
### What is this PR proposing?
- Refactor of `_apply_global_config_overrides()` so that config is not modified in-place, but returns the `DataContextConfig` that has the global overrides applied, and is loaded directly into the `project_config` property. 
- GREAT-953
- Adjusted tests so that usage_stats (which are directly affected by `_apply_global_config_overrides()`) are explicitly enabled.  This has resulted in some tests needing to be adjusted, all at the `cli` level. Further investigation is needed as to why this is coming up, but a few tests have had to have their emitted events adjusted to account for the events that occur when the fixture is built (and not just when the test is run). Further investigative work will be done with @cdkini 

### Definition of Done

Thank you for submitting!


great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -348,8 +348,9 @@ def __init__(
             )
         self._ge_cloud_mode = ge_cloud_mode
         self._ge_cloud_config = ge_cloud_config
-        self._project_config = project_config
-        self._apply_global_config_overrides()
+        self._project_config = self._apply_global_config_overrides(
+            config=project_config
+        )
```
Shinnnyshinshin, line 353: This is the main change proposed in this PR. Instead of 

1. loading `project_config` parameter into the `_project_config` property
2. changing the property in-place through `_apply_global_config_overrides()` method

the PR is proposing 
1. taking the `project_config` parameter, and passing it to the`_apply_global_config_overrides()` method.
2. The `_apply_global_config_overrides()` will make a `deepcopy` and perform the overrides.
3. The resulting config is returned by `_apply_global_config_overrides()` and loaded into the `_project_config` object of the DataContext objec.t

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -481,15 +482,21 @@ def _init_datasources(self, config: DataContextConfig) -> None:
...
                 pass
 
-    def _apply_global_config_overrides(self) -> None:
+    def _apply_global_config_overrides(
+        self, config: DataContextConfig
+    ) -> DataContextConfig:
         # check for global usage statistics opt out
         validation_errors = {}
 
+        config_with_global_config_overrides: DataContextConfig = copy.deepcopy(config)
```
Shinnnyshinshin, line 518: `deepcopy()` is here

tests/cli/test_project.py
```diff
@@ -104,9 +154,6 @@ def test_project_check_on_valid_project_says_so(
 def test_project_check_on_project_with_v2_datasources_and_validation_operators(
     mock_emit, caplog, monkeypatch, titanic_data_context
```
Shinnnyshinshin: `titanic_data_context` does not have usage_stats enabled.

tests/cli/test_project.py
```diff
@@ -161,7 +243,7 @@ def test_project_check_on_project_with_missing_config_file_guides_user(
     monkeypatch.chdir(os.path.dirname(context.root_directory))
     result = runner.invoke(
         cli,
-        ["--v3-api", "project", "check-config"],
```
Shinnnyshinshin, line 164: We dont need the `--v3-api` flag anymore

great_expectations/data_context/data_context/base_data_context.py
Shinnnyshinshin, line 353: The critical change. We intended this to have no side-effect, but the tests have revealed that trying to modify the config in place has resulted in some behavior that was hidden. Further investigation is needed as to the root cause

tests/cli/test_datasource_pandas.py
```diff
@@ -120,13 +117,10 @@ def test_cli_datasource_new(
     mock_emit,
     caplog,
     monkeypatch,
-    empty_data_context,
+    empty_data_context_stats_enabled,
```
Shinnnyshinshin, line 120: tests now explicitly enable `usage_statistics` by using the `_stats_enabled` fixture directly

tests/cli/test_datasource_sqlite.py
```diff
@@ -78,6 +78,43 @@ def test_cli_datasource_list(
...
+                "success": True,
+            }
+        ),
+        mock.call(
+            {
+                "event": "cli.datasource.list.end",
+                "event_payload": {"api_version": "v3"},
+                "success": True,
+            }
+        ),
```
Shinnnyshinshin, line 117: These should have been emitted before. Now emitted because we are explicitly enabling `usage_statistics` for the test at the DataContext-instantiation level.

tests/integration/docusaurus/miscellaneous/migration_guide_postgresql_v3_api.py
```diff
@@ -38,10 +38,12 @@
       connection_string: postgresql+psycopg2://postgres:@localhost/test_ci
     data_connectors:
       default_runtime_data_connector_name:
+        module_name: great_expectations.datasource.data_connector
```
Shinnnyshinshin, line 41: this was causing some some integration tests to fail. We don't actually refer to this file in the Docs, so a  line-number change isn't required.

anthonyburdi: This is strange - it looks like we do supply a missing value in the `DataConnectorConfigSchema`: https://github.com/great-expectations/great_expectations/blob/1d01e61221cf739137c404883457088a7a268c91/great_expectations/data_context/types/base.py#L475. Is this config not going through schema validation?

anthonyburdi (commented):
All looks good, but just want to make sure the addition of `module_name` isn't hiding other issues.

tests/integration/docusaurus/miscellaneous/migration_guide_postgresql_v3_api.py
cdkini: hmm yeah I'm a little confused why this would change

Shinnnyshinshin: Thank you so much for pointing this out @anthonyburdi and @cdkini...this was carryover from when we were getting some strange integration test errors earlier in the week. 

Totally agree that this shouldn't be masking a bigger issue 🙇🏼  removing the addition of `module_name` for this PR

anthonyburdi (approved):
LGTM! Thanks for digging deep to figure out what was going on here!

tests/integration/docusaurus/miscellaneous/migration_guide_postgresql_v3_api.py
Shinnnyshinshin: Update (and copied from Slack) 

Went through the code again, and it does appear that the module_name needs to be included.
* We don’t go through the schema which would have added that field automatically, but load the yaml directly
* Other v3 examples in the same folder (pandas and spark) also include the module name

#### where did this change come from?
Having trouble with this one. Gitblame only blames me from 4 months ago :sweat_smile:  so this test should have been failing

anthonyburdi: Very strange. I don't want to hold this PR up but let's keep our eyes out for any related issues!

anthonyburdi (approved):
LGTM!

PR #5321, opened 2022-06-16
[BUGFIX] Fix for failing Expectation test in `cloud_db_integration` pipeline

Shinnnyshinshin:
- `cloud_db_integration` pipeline, which runs Expectations tests and performance tests for BigQuery, was having a failing expectations test, specifically around the `expect_column_values_to_be_unique` expectation. 
- This test adds a specific test case that is only run for `bigquery` that adds the expected return value, and suppresses the general test for `bigquery`, since it is already handled by the specific test. 

### Definition of Done


alexsherstinsky (approved):
LGTM

PR #5285, opened 2022-06-09
[MAINTENANCE] Move `_apply_global_config_overrides()` to AbstractDataContext

Shinnnyshinshin:
Changes proposed in this pull request:
- Follow up to #5286 PR
- Bite-sized piece of the changes proposed in #5227 (ie. the big scary `DataContext` refactor)
- GREAT-953
- Moving logic around `_apply_global_config_overrides()` to respective classes (`FileDataContext`, `CloudDataContext`, `EphemeralDataContext`)
- Allowing for `AbstractDataContext` to look for configuration in both ENV variable and config file
- Docstrings
- Unittests

### Definition of Done
Please delete options that are not relevant.

- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added [unit tests](https://docs.greatexpectations.io/docs/contributing/contributing_test#writing-unit-and-integration-tests) where applicable and made sure that new and existing tests are passing.
- [x] I have run any local integration tests and made sure that nothing is broken.

PR #5339, opened 2022-06-17
[MAINTENANCE] Refactor `usage_stats_opt_out` method in DataContext

Shinnnyshinshin:
- Follow up to #5285 
- Small refactor to `_check_global_usage_statistics_opt_out()` so that we dont have to "flip" the boolean in our minds. Before if `opt_out` = `True`, then `usage_statistics_enabled` = `False`. 

- Now we use `_get_global_usage_statistics_override()` which returns `True` if `usage_statistics_enabled` = `True`, and `False` if `usage_statistics_enabled` = `False`. 
- GREAT-953

## Update 2022-08-26

`anonymous_usage_stats_enabled` can be set in the following 3 places
        - `great_expectations.yml` (YML)
        - `GE_USAGE_STATS` environment variable (env)
        - `conf` file (conf)
 
If it is set as `False` in *any* of the 3 places, the global value is set to `False`
* This is a change from previous behavior which set the following precedence: `environment variable > home folder > /etc > yml`
* This PR adds a 3 tests that check if setting YML, env, or conf to `False` will set the global usage_statistics flag to `False`. 
* It also removes tests that used to check the test percedence. 

### Definition of Done


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -166,12 +165,13 @@ def _check_global_usage_statistics_opt_out(self) -> bool:
             config.BOOLEAN_STATES = states
             config.read(config_path)
             try:
-                if config.getboolean("anonymous_usage_statistics", "enabled") is False:
-                    # If stats are disabled, then opt out is true
+                if config.getboolean("anonymous_usage_statistics", "enabled"):
                     return True
+                else:
+                    return False
```
anthonyburdi: This is tricky here. If ANY of the usage stats overrides are set then we should turn off usage stats - even if there are conflicting settings. It's probably worth adding a test as well since I think this proposed change would use the first setting file encountered and ignore the rest.

anthonyburdi (changes requested):
I like the change but please add a test and make sure that all locations are checked for any opt out.

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: I will look into this. Great suggestion about the test.

Shinnnyshinshin: Hi @anthonyburdi tried having a closer look at the issue this afternoon, and I wasn't able to find where the refactor affected current behavior. 

I found the tests listed in `tests/core/usage_statistics/test_usage_statistics.py` to be particularly helpful, which describe the following precedence (rather than "If ANY of the usage stats overrides are set then we should turn off usage stats - even if there are conflicting settings" that you described above)

`Test precedence: environment variable > home folder > /etc > yml`

The following existing tests having been renamed to in alignment with the current refactor, and seem to test this pretty thoroughly. 

- `test_global_override_env_var_overrides_home_folder`
- `test_global_override_env_var_overrides_etc`
- `test_global_override_env_var_overrides_etc`
- `test_global_override_env_var_overrides_yml_v013`
- `test_global_override_home_folder_overrides_etc`
- `test_global_override_home_folder_overrides_yml`
- `test_global_override_home_folder_overrides_yml_v013`
- `test_global_override_etc_overrides_yml`
- `test_global_override_etc_overrides_yml_v013`

anthonyburdi: Thanks for looking into that :bow: 
That is fair, I agree that the tests are written with that order of precedence and they are passing in this PR so the tested functionality is preserved. 
In our docs https://docs.greatexpectations.io/docs/reference/anonymous_usage_statistics though we do say `or` which would lead me to believe that setting any to false would turn off usage stats. 
Sorry for harping on this issue, it's an important one to get right.
How about we bring this up with the pod e.g. in standup?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2437,6 +2440,21 @@ def _get_metric_configuration_tuples(
 
         return metric_configurations_list
 
+    @classmethod
+    def validate_config(
+        cls, project_config: Union[DataContextConfig, Mapping]
```
Shinnnyshinshin, line 2445: `validate_config()` 's behavior has been extended to always return a `DataContextConfig` or error out.  This helps with `mypy` errors, which were complaining that parameters were being `Union[DataContextConfig, Mapping]` in one place and `DataContextConfig` in another. 

It has also been moved up to `AbstractDataContext` because `BaseDataContext` and `CloudDataContext` both use it. 


This is the previous implementation
```python
    @classmethod
    def validate_config(cls, project_config: Union[DataContextConfig, Mapping]) -> bool:
        if isinstance(project_config, DataContextConfig):
            return True
        try:
            dataContextConfigSchema.load(project_config)
        except ValidationError:
            raise
        return True
```

alexsherstinsky (approved):
LGTM

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: Is this name still appropriate? Is there perhaps a better name to signal what it is doing now that it isn't returning a `bool`?

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -231,10 +219,11 @@ def __init__(
         Returns:
             None
         """
-        if not BaseDataContext.validate_config(project_config):
-            raise ge_exceptions.InvalidConfigError(
-                "Your project_config is not valid. Try using the CLI check-config command."
-            )
+
+        project_data_context_config: DataContextConfig = (
+            BaseDataContext.validate_config(project_config)
```
cdkini: For static/class methods called in instances, do we want to use the class name or `self`? Good either way but let's be consistent. I think I'm the one who added this class call but I might be the outlier.

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -88,8 +88,13 @@ def __init__(
             project_config = self.retrieve_data_context_config_from_ge_cloud(
                 ge_cloud_config=self._ge_cloud_config,
             )
+
+        project_data_context_config: DataContextConfig = (
+            CloudDataContext.validate_config(project_config)
```
cdkini: Same note

great_expectations/util.py
```diff
@@ -1668,7 +1669,7 @@ def convert_ndarray_decimal_to_float_dtype(data: np.ndarray) -> np.ndarray:
 
 
 def get_context(
-    project_config: Optional[Union["DataContextConfig", dict]] = None,
+    project_config: Optional[Union["DataContextConfig", Mapping]] = None,
```
cdkini, line 1672: Could we fix up this `DataContextConfig` type annotation by using `from __future__ import annotations` and `if TYPE_CHECKING`?

tests/integration/usage_statistics/test_usage_stats_common_messages_are_sent_v3api.py
```diff
@@ -94,7 +94,7 @@ def test_common_usage_stats_are_sent_no_mocking(
     context._usage_statistics_handler._close_worker()
 
     # Make sure usage stats are enabled
-    assert not context._check_global_usage_statistics_opt_out()
+    assert context._get_global_usage_statistics_override()
```
cdkini: I know we had this issue before but I don't love directly accessing a private method here - is it possible to get around this at all? Do we need this assertion?

Feels like we might be testing an implementation detail

cdkini (commented):
Left some misc comments but mainly LGTM!

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2530,21 +2545,24 @@ def _load_config_variables(self) -> Dict:
             return config_var_provider.get_values()
         return {}
 
-    def _check_global_usage_statistics_opt_out(self) -> bool:
+    @staticmethod
+    def _get_global_usage_statistics_override() -> bool:
```
anthonyburdi: How about something like `_is_usage_stats_enabled() -> bool:` since this method returns a boolean?

tests/core/usage_statistics/test_usage_statistics.py
```diff
@@ -83,7 +83,7 @@ def test_consistent_name_anonymization(
 
 
 @pytest.mark.base_data_context
-def test_opt_out_environment_variable_base_data_context(
+def test_global_override_environment_variable_base_data_context(
```
anthonyburdi, line 86: :heart:

tests/core/usage_statistics/test_usage_statistics.py
```diff
@@ -206,54 +206,28 @@ def test_opt_out_yml(tmp_path_factory, monkeypatch):
...
+    """
+    What does this test and why?
 
-    enabled_config = configparser.ConfigParser()
-    enabled_config["anonymous_usage_statistics"] = {"enabled": "True"}
+    anonymous_usage_stats_enabled can be set in the following 3 places
+        - great_expectations.yml (YML)
+        - GE_USAGE_STATS environment variable (env)
+        - conf file (conf)
+    If it is set as `False` in *any* of the 3 places, the global value is set to `False`
```
anthonyburdi, line 220: :heart:

anthonyburdi (approved):
Just a non blocking rename suggestion.

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: I really like this suggestion :) changing now

great_expectations/data_context/data_context/base_data_context.py
Shinnnyshinshin: great question @cdkini. I had a look, and I think it's a mix of class name and self 😅 My opinion is that we should use class name to distinguish between these and instance methods

great_expectations/data_context/data_context/cloud_data_context.py
Shinnnyshinshin: replied above

great_expectations/util.py
Shinnnyshinshin: Good call :) doing now, and for a few more parameters below.

tests/integration/usage_statistics/test_usage_stats_common_messages_are_sent_v3api.py
Shinnnyshinshin: I can understand this concern. I think we can get around this by checking the boolean flag on `anonymous_usage_statistics` property, which we do directly below. If that is the case then we can just remove this line

PR #5366, opened 2022-06-22
[MAINTENANCE] Refactor `DataContextVariables` to only persist state to Store using explicit save command

cdkini:
- Instead of persisting changes made to `DataContextVariables` with each SET command, we want to make persistence an explicit, opt-in command. A new `save_config` command will leverage a given store to save any changes made to the runtime object (specifics depend on environment and variables child class).
- Improved capabilities of the `InlineStoreBackend` - it continues to support granular modification of discrete components of the config but is newly updated to allow for entire config read/writes.
- Refactor of `DataContextVariablesStore` to just be `DataContextStore` and leverage the same pattern as other `ConfigurationStore` child classes - this is much simpler and better adheres to the rest of the codebase.
- Simplification of the `DataContextVariables` constructor and support for variable substitution.
- Updated tests to cover new behavior.

### Definition of Done


great_expectations/data_context/types/data_context_variables.py
```diff
@@ -294,6 +289,10 @@ def get_progress_bars(self) -> Optional[ProgressBarsConfig]:
         """
         return self._get(DataContextVariableSchema.PROGRESS_BARS)
 
+    def save_config(self) -> None:
+        key: DataContextVariableKey = self._get_key("data-context")
+        self.store.set(key=key, value=self.config)
```
cdkini: TBD need to figure out which key to use now that we don't need granular updates to the store with each `set` call.

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -4,10 +4,10 @@
 from great_expectations.data_context.data_context.abstract_data_context import (
     AbstractDataContext,
 )
-from great_expectations.data_context.types.base import DataContextConfig, GeCloudConfig
-from great_expectations.data_context.types.data_context_variables import (
+from great_expectations.data_context.data_context_variables import (
```
cdkini, line 7: Moving the variables out of types!

great_expectations/data_context/data_context_variables.py
```diff
@@ -3,17 +3,23 @@
...
-from great_expectations.data_context.types.resource_identifiers import GeCloudIdentifier
+from great_expectations.data_context.types.resource_identifiers import (
+    ConfigurationIdentifier,
+    GeCloudIdentifier,
+)
+from great_expectations.data_context.util import substitute_config_variable
 
 
 class DataContextVariableSchema(str, enum.Enum):
+    DATA_CONTEXT = "data_context"  # If retrieving/setting the entire config at once
```
cdkini: Our enum now includes `DATA_CONTEXT` to refer to the whole config - we need the ability to modify both individual components of the config as well as the entire config itself. 

Logic exists downstream in the stores to work with these changes.

great_expectations/data_context/data_context_variables.py
```diff
@@ -47,50 +53,56 @@ class DataContextVariables(ABC):
     are persisted for future usage (i.e. filesystem I/O or HTTP request to a Cloud endpoint).
 
     Should maintain parity with the `DataContextConfig`.
+
+    Args:
+        config:        A reference to the DataContextConfig to perform CRUD on.
+        substitutions: A dictionary used to perform substitutions of ${VARIABLES}; to be used with GET requests.
+        _store:        An instance of a DataContextStore with the appropriate backend to persist config changes.
```
cdkini, line 62: A simplification of the variables constructor - we can use the same config as the context.

Additionally, we want to be able to perform variable substitution right as we need it (when the user retrieves a given value). As such, we hold on to a reference of the substitutions dict used by the context.

great_expectations/data_context/data_context_variables.py
```diff
@@ -47,50 +53,56 @@ class DataContextVariables(ABC):
...
-    concurrency: Optional[ConcurrencyConfig] = None
-    progress_bars: Optional[ProgressBarsConfig] = None
-    _store: Optional["DataContextVariablesStore"] = None  # noqa: F821
+    config: DataContextConfig
+    substitutions: Optional[dict] = None
+    _store: Optional["DataContextStore"] = None  # noqa: F821
+
+    def __post_init__(self) -> None:
+        if self.substitutions is None:
+            self.substitutions = {}
```
cdkini, line 71: Mutable default values 😢

great_expectations/data_context/data_context_variables.py
```diff
@@ -47,50 +53,56 @@ class DataContextVariables(ABC):
...
+        val: Any = self.config[key]
+        substituted_val: Any = substitute_config_variable(val, self.substitutions)
+        return substituted_val
+
+    def save_config(self) -> None:
+        """
+        Persist any changes made to variables utilizing the configured Store.
+        """
+        key: ConfigurationIdentifier = self.get_key()
+        self.store.set(key=key, value=self.config)
```
cdkini, line 107: We previously persisted values with each `_set` call - this is not desired behavior.

The user should be able to make modifications as they see fit and ONLY when they deem it appropriate, explicitly invoke the `save_config` command to persist changes using the store. This is a subtle change but is more in-line with how GE currently works.

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -55,42 +66,53 @@ def __init__(
...
-            self._data_context._save_project_config()
-        except AttributeError as e:
-            logger.warning(
-                f"DataContext of type {type(self._data_context)} used with {self.__class__.__name__} is not file-system enabled: {e}"
-            )
+        self._persist_changes()
 
     def _move(
-        self, source_key: Tuple[str, str], dest_key: Tuple[str, str], **kwargs: dict
+        self, source_key: Tuple[str, ...], dest_key: Tuple[str, ...], **kwargs: dict
```
cdkini, line 109: Length of the tuple is variable due to the possibility of different keys and store "frontends"

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -119,14 +141,20 @@ def list_keys(self, prefix: Tuple[str, str] = ()) -> List[str]:
...
+        (
+            resource_type,
+            resource_name,
+        ) = InlineStoreBackend._determine_resource_type_and_name_from_key(key)
 
-        if not resource_name:
+        if resource_type is DataContextVariableSchema.DATA_CONTEXT:
+            raise StoreBackendError(
+                "InlineStoreBackend does not support the deletion of the overall DataContext project config"
+            )
```
cdkini: The only real changes here are some checks to determine if the resource type is the overall context - if so, we perform CRUD on the full object as opposed to a granular part of it.

This implementation works with both the variables as well as the `DatasourceStore`.

great_expectations/data_context/data_context_variables.py
anthonyburdi: I agree about performing variable substitution right as we need it. I don't have a strong preference here but substitution feels like the responsibility of the caller and not the DataContextVariables object itself though.

anthonyburdi (commented):
Looks good, just left one comment to discuss which class bears responsibility of substitution.

anthonyburdi (approved):
LGTM! Thank you for discussing and making changes, I think this ended up in a really good place :)

PR #5369, opened 2022-06-23
[FEATURE] Add atomic `rendered_content` to `ExpectationValidationResult` and `ExpectationConfiguration`

NathanFarmer:
- GREAT-911/GREAT-942
- GREAT-911/GREAT-944
- Changes are behind default feature flag `include_rendered_content=False` in `context.get_validator()`.
- Content is rendered to `ExpectationValidationResult.rendered_content` or `ExpectationValidationResult.expectation_config.rendered_content`. An `InlineRenderer` has been added to accomplish this.
- Atomic renderers are used to render content in all cases. To ensure this is the case, a default atomic renderer (`atomic.prescriptive.kwargs`) was added for cases where no atomic renderers are defined at the expectation level and `atomic.prescriptive.summary` fails.
- A minor change was made to `expect_column_kl_divergence_to_be_less_than` to return rendered graphs as dictionaries rather than json strings.
- Additional tests to be added in follow-up PR for specific types of `RenderedAtomicContent`.

### Definition of Done


great_expectations/expectations/core/expect_column_kl_divergence_to_be_less_than.py
```diff
@@ -740,7 +740,7 @@ def _get_kl_divergence_chart(cls, partition_object, header=None):
                     .properties(width=chart_pixel_width, height=400, autosize="fit")
                 )
 
-                chart = bars.to_json()
+                chart = bars.to_dict()
```
NathanFarmer, line 743: This resulted in the Altair charts being serialized as a json string. It is preferable to pass this as a dictionary.

great_expectations/expectations/expectation.py
```diff
@@ -203,6 +203,29 @@ def _validate(
     ) -> Union[ExpectationValidationResult, dict]:
         raise NotImplementedError
 
+    @classmethod
+    @renderer(renderer_type="atomic.prescriptive.kwargs")
+    def _prescriptive_kwargs(
```
NathanFarmer, line 208: This is now the default atomic prescriptive renderer. It will only be used in cases where all other renderers for a given expectation (or custom expectation) are 1. missing atomic renderers and 2. fail when attempting to use the preferred default renderer `atomic.prescriptive.summary`.

great_expectations/expectations/expectation.py
```diff
@@ -861,6 +893,8 @@ def validate(
             configurations=[configuration],
             runtime_configuration=runtime_configuration,
         )[0]
+        if include_rendered_content:
+            evr.render()
```
NathanFarmer, line 914: This invokes the `InlineRenderer` after the instance attributes `rendered_content` have been added and evaluation parameters have been processed.

great_expectations/render/renderer/inline_renderer.py
```diff
@@ -0,0 +1,127 @@
...
+                except Exception:
+                    pass
+
+        if len(rendered_content) == 0 and isinstance(
+            render_object, ExpectationConfiguration
+        ):
+            renderer_rendered_content = (
+                InlineRenderer._get_renderer_atomic_rendered_content(
+                    render_object=render_object,
+                    renderer_name=default_prescriptive_renderer_name,
```
NathanFarmer, line 109: Only use renderer `atomic.prescriptive.kwargs` if we are rendering an `ExpectationConfiguration` and no other atomic renderers existed/rendered successfully.

great_expectations/render/types/__init__.py
```diff
@@ -501,18 +500,20 @@ def to_json_dict(self):
...
         template: Optional[str] = None,
         params: Optional[dict] = None,
-        schema: Optional[dict] = None,
         header: Optional["RenderedAtomicValue"] = None,
         header_row: Optional[List["RenderedAtomicValue"]] = None,
         table: Optional[List[List["RenderedAtomicValue"]]] = None,
         graph: Optional[dict] = None,
+        kwargs: Optional[dict] = None,
     ) -> None:
+        self.schema: Optional[dict] = schema
```
NathanFarmer, line 512: `self.schema` is used for other types so I moved it out from under `StringValueType`.

great_expectations/core/expectation_configuration.py
```diff
@@ -970,6 +972,8 @@ def __init__(
         self.success_on_last_run = success_on_last_run
         self._ge_cloud_id = ge_cloud_id
         self._expectation_context = expectation_context
+        if include_rendered_content:
+            self.rendered_content = {}
```
cdkini: Could we always set this attribute? I'd like to avoid conditional modifications of the shape of the object if possible.

great_expectations/core/expectation_validation_result.py
```diff
@@ -67,6 +71,8 @@ def __init__(
             "exception_traceback": None,
             "exception_message": None,
         }
+        if include_rendered_content:
+            self.rendered_content = {}
```
cdkini: Same note here.

great_expectations/core/expectation_validation_result.py
```diff
@@ -182,6 +188,35 @@ def __str__(self) -> str:
...
+        inline_renderer_config: Dict[str, Union[str, ExpectationValidationResult]] = {
+            "class_name": "InlineRenderer",
+            "render_object": self,
+        }
+        module_name: str = "great_expectations.render.renderer.inline_renderer"
+        inline_renderer = instantiate_class_from_config(
+            config=inline_renderer_config,
+            runtime_environment={},
+            config_defaults={"module_name": module_name},
+        )
```
cdkini, line 213: Could we instantiate the object with the constructor instead using `instantiate_class_from_config`? Since we know the type, we don't necessarily need the magic method right?

great_expectations/expectations/core/expect_column_kl_divergence_to_be_less_than.py
cdkini, line 743: Why this change?

great_expectations/expectations/registry.py
```diff
@@ -65,9 +65,36 @@ def register_renderer(
...
 
 def get_renderer_names(object_name: str) -> List[str]:
+    """Gets renderer names for a given Expectation or Metric.
+
+    Args:
+        object_name: The name of an Expectation or Metric for which to get renderer names.
+
+    Returns:
+        A list of renderer names for the Expectation or Metric.
+    """
```
cdkini, line 75: Could we modify our internal linters (docstrings/type hints)?

great_expectations/render/renderer/inline_renderer.py
```diff
@@ -0,0 +1,127 @@
...
+                    renderer_rendered_content = (
+                        self._get_renderer_atomic_rendered_content(
+                            render_object=render_object,
+                            renderer_name=renderer_name,
+                            expectation_type=expectation_type,
+                        )
+                    )
+                    rendered_content.append(renderer_rendered_content)
+                except Exception:
+                    pass
```
cdkini: Could we log this instead of a raw pass?

great_expectations/render/renderer/inline_renderer.py
```diff
@@ -0,0 +1,127 @@
...
+    def _get_renderer_atomic_rendered_content(
+        render_object: Union[ExpectationConfiguration, ExpectationValidationResult],
+        renderer_name: str,
+        expectation_type: str,
+    ) -> Optional[RenderedAtomicContent]:
+        renderer_rendered_content: Optional[RenderedAtomicContent] = None
+        renderer_tuple: Optional[tuple] = get_renderer_impl(
+            object_name=expectation_type, renderer_type=renderer_name
+        )
+        if renderer_tuple is not None:
```
cdkini, line 127: Is it possible to make this a defined type (like a `namedtuple`)?

great_expectations/validator/validator.py
```diff
@@ -156,12 +157,14 @@ def __init__(
...
-        :param data_context (Optional[DataContext]):
-        :param batches (Optional[List[Batch]]):
+        Args:
+            execution_engine: The ExecutionEngine to be used to perform validation.
+            interactive_evaluation: Whether the Validator should perform evaluation when Expectations are added.
+            expectation_suite: The ExpectationSuite to validate.
+            expectation_suite_name: The name of the ExpectationSuite to validate.
+            data_context: The DataContext associated with this Validator.
+            batches: The batches for which to validate.
+            include_rendered_content: Whether the ExpectationValidationResult should return rendered_content.
```
cdkini: ❤️

cdkini (changes requested):
Minor changes requested (primarily around dynamic setting of attrs)

great_expectations/core/expectation_configuration.py
NathanFarmer: We can, but just note that this will mean the default behavior of `ExpectationValidationResult` will be to contain an empty `rendered_content` attribute. We could just modify the `__repr__` to hide it. Until I pull the feature flag out it might be best to leave it like this...

great_expectations/core/expectation_validation_result.py
NathanFarmer: 1) This follows the existing pattern for renderers, but if it is no longer preferred we can use the class. 2. This will result in a circular import that I will need to figure out how to address in the renderer.

great_expectations/expectations/registry.py
NathanFarmer: I'm not following what the question is here. I documented this to keep it in line with the method I added (`get_renderer_names_from_renderer_prefix`).

great_expectations/render/renderer/inline_renderer.py
NathanFarmer: Yes, good idea!

great_expectations/render/renderer/inline_renderer.py
NathanFarmer: Yes, good idea!

great_expectations/expectations/core/expect_column_kl_divergence_to_be_less_than.py
NathanFarmer: This resulted in the Altair charts being serialized as a json string. It is preferable to pass this as a dictionary so the string doesn't need to be parsed later.

great_expectations/core/expectation_configuration.py
NathanFarmer: It felt wrong to pass `include_rendered_content=False` and then have a new attribute called `rendered_content` anyway.

NathanFarmer: It felt wrong to pass `include_rendered_content=False` and then have a new attribute called `rendered_content` anyway.

great_expectations/expectations/registry.py
cdkini: ah sorry I meant the thresholds in `scripts/`. We have constants in `check_type_hint_coverage.py` and `check_docstring_coverage.py` that need updating 😄

great_expectations/core/expectation_validation_result.py
cdkini: 1. I'd vote for using a new pattern so we can move away from `instantiate_class_from_config` but am happy to leave it as is if we want to be consistent.
2. Would this still be an issue if we imported within the function scope?

great_expectations/core/expectation_configuration.py
cdkini: Could it be a nullable attribute? Make it `None` if `False`?

great_expectations/expectations/registry.py
NathanFarmer: Got it ✅

great_expectations/core/expectation_validation_result.py
NathanFarmer: > Would this still be an issue if we imported within the function scope?

You mean like this?
```
def render(self) -> None:
    from great_expectations.render.renderer.inline_renderer import InlineRenderer

    inline_renderer = InlineRenderer(render_object=self)

    (
        self.expectation_config.rendered_content,
        self.rendered_content,
    ) = inline_renderer.render()
```
That works, but I'm not sure that I've seen it anywhere else in the codebase. The options seem to be:
1. Do it with `instantiate_class_from_config` like it is now.
2. Import `InlineRenderer` at the top of the module and remove `ExpectationConfiguration` and `ExpectationValidationResult` objects from the renderer - inferring them from their attributes. And then maybe update the other renderers to follow this pattern.
3. Import `InlineRenderer` in the `render` method, and then maybe update the other renderers to follow this pattern.

great_expectations/render/renderer/inline_renderer.py
NathanFarmer: ✅

great_expectations/render/renderer/inline_renderer.py
NathanFarmer: Per our offline conversation, not going to do this as it is a breaking change for cloud to return something different from `get_renderer_impl`.

great_expectations/expectations/expectation.py
```diff
@@ -682,11 +701,19 @@ def get_allowed_config_keys(cls):
 
     def metrics_validate(
         self,
-        metrics: Dict,
+        metrics: dict,
         configuration: Optional[ExpectationConfiguration] = None,
         runtime_configuration: dict = None,
         execution_engine: ExecutionEngine = None,
+        **kwargs,
```
cdkini: Please annotate as dict

great_expectations/expectations/expectation.py
```diff
@@ -713,22 +740,42 @@ def metrics_validate(
...
+            include_rendered_content=include_rendered_content,
         )
         return evr
 
     @staticmethod
-    def _build_evr(raw_response, configuration) -> ExpectationValidationResult:
+    def _build_evr(
+        raw_response: Union[ExpectationValidationResult, dict],
+        configuration: ExpectationConfiguration,
+        **kwargs,
```
cdkini: Please annotate as dict

great_expectations/core/expectation_configuration.py
```diff
@@ -1365,13 +1376,22 @@ def metrics_validate(
         self,
         metrics: Dict,
         runtime_configuration: dict = None,
-        execution_engine=None,
+        execution_engine: "ExecutionEngine" = None,  # noqa: F821
+        **kwargs,
```
cdkini: Please annotate as dict

great_expectations/core/expectation_validation_result.py
```diff
@@ -42,12 +48,19 @@ def get_metric_kwargs_id(metric_name, metric_kwargs):
...
-        expectation_config=None,
-        result=None,
-        meta=None,
-        exception_info=None,
+        success: Optional[bool] = None,
+        expectation_config: Optional["ExpectationConfiguration"] = None,  # noqa: F821
+        result: Optional[dict] = None,
+        meta: Optional[dict] = None,
+        exception_info: Optional[dict] = None,
+        **kwargs,
```
cdkini: Please annotate as dict

cdkini (approved):

great_expectations/core/expectation_configuration.py
NathanFarmer: This attribute is no longer conditional, but will stay private `self._rendered_content` until the feature flag is pulled out.

great_expectations/expectations/expectation.py

great_expectations/expectations/expectation.py

great_expectations/core/expectation_configuration.py
NathanFarmer: ✅

great_expectations/core/expectation_validation_result.py
NathanFarmer: ✅

PR #5385, opened 2022-06-25
[MAINTENANCE] Migration of `project_config_with_variables_substituted` to `AbstractDataContext`

Shinnnyshinshin:
- Migration `get_config_with_variables_substituted()` from `BaseDataContext`
  - code that is run by default is moved to `AbstractDataContext`
  - Cloud-specific logic is moved to `CloudDataContext. -  
- Associated properties and methods also moved to `AbstractDataContext`

**Also Changed**
- Call to `super()` (ie. `AbstractDataContext`) by child classes are moved to the last step of the initialization. This ensures that `DataContext`-specific parameters are initialized before running any common logic. 

- GREAT-1009

### Definition of Done

Thank you for submitting!


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -267,6 +290,49 @@ def config_variables(self) -> Dict:
...
+    def root_directory(self) -> Optional[str]:
+        """The root directory for configuration objects in the data context; the location in which
+        ``great_expectations.yml`` is located.
+
+        Why does this exist in AbstractDataContext? CloudDataContext and FileDataContext both use it
+
+        """
+        if hasattr(self, "_context_root_directory"):
+            return self._context_root_directory
+        return
```
cdkini: why `hasattr` here? and nitpick but I don't think we need the return on 307

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -267,6 +290,49 @@ def config_variables(self) -> Dict:
...
+    ) -> Optional[str]:
+        """
+        Why does this exist in AbstractDataContext? CloudDataContext and FileDataContext both use it
+        """
+        if path is None:
+            return
+        if os.path.isabs(path):
+            return path
+        else:
+            return os.path.join(self.root_directory, path)
```
cdkini: Nit - Minor refactor 
```suggestion
        if path is None or os.path.isabs(path):
            return path
        return os.path.join(self.root_directory, path)
```

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -293,3 +359,23 @@ def _determine_substitutions(self) -> dict:
...
+
+    def get_config_with_variables_substituted(
+        self, config: Optional[DataContextConfig] = None
+    ) -> DataContextConfig:
+        """
+        Substitute vars in config of form ${var} or $(var) with values found in the following places,
+        in order of precedence: ge_cloud_config (for Data Contexts in GE Cloud mode), runtime_environment,
+        environment variables, config_variables, or ge_cloud_config_variable_defaults (allows certain variables to
+        be optional in GE Cloud mode).
+        """
```
cdkini, line 374: While it isn't particularly detrimental here, I think the less we reference child classes in `AbstractDataContext`, the better. At this point in the inheritance hierarchy, we have no knowledge of things like `CloudDataContext`.

I think as much as possible, we should treat the methods herein as black boxes and have the child classes fill in the details. The order of precedence is determined by the child class and can be abstracted away by the `_determine_substitutions` method from the child. Perhaps this documentation better fits there?

All that being said, I know this refactor is less than straightforward 😄 just want to try and keep this class as abstract as possible.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -267,6 +290,49 @@ def config_variables(self) -> Dict:
...
+
+    @property
+    def project_config_with_variables_substituted(self) -> DataContextConfig:
+        return self.get_config_with_variables_substituted()
+
+    @property
+    def plugins_directory(self):
+        """The directory in which custom plugin modules should be placed.
+
+        Why does this exist in AbstractDataContext? CloudDataContext and FileDataContext both use it
```
anthonyburdi, line 320: Since `CloudDataContext` currently references the filesystem, how do we feel about inheriting from `FileDataContext` instead of `AbstractDataContext` e.g. `CloudDataContext(FileDataContext)`? That might allow us to remove the file based methods/attributes from `AbstractDataContext`.

anthonyburdi: Alternatively, we could refactor the file interaction code into a separate class that is injected as a dependency where we need it (File and Cloud).

Shinnnyshinshin: @anthonyburdi this is a great idea, and I think it will definitely make things simpler going forward. I can implement the adjustment to the hierarchy between `CloudDataContext` and `FileDataContext` in this PR and will make a note of the possibility of the file interaction code into separate class. 

Thanks for the synchronous conversation as well

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -282,6 +305,46 @@ def config_variables(self) -> Dict:
...
+    def _normalize_absolute_or_relative_path(
+        self, path: Optional[str]
+    ) -> Optional[str]:
+        """
+        Why does this exist in AbstractDataContext? CloudDataContext and FileDataContext both use it
+        """
+        if path is None or os.path.isabs(path):
+            return path
+        return os.path.join(self.root_directory, path)
+
```
Shinnnyshinshin: this might be able to move to FileDataContext

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: update: after a conversation with @donaldheppner this understanding has been updated: 
- The key difference between *`DataContext` is going to be where the initial configuration comes from : 
- `CloudDataContext` loads config from endpoint
- `FileDataContext` loads `great_expectations.yml` from disk
-  `EphemeralDataContext` doesn't have a yml at all, but loads from in-memory dictionary. 

Afterwards a lot of the behavior is going to be shared, which means all DataContexts will have a way to access the file system, config files, Stores on filesystem, etc (which is what GE currently supports). 

This means a LOT of the functionality is going to be living in AbstractDataContext, and `Cloud` and `File` will remain as siblings.

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: This is actually being addressed in #5339 where we are intending to get rid of the hierarchy and have usage_stats be disabled if an override is set in any of the above-mentioned locations . I'd like to propose that we maintain this functionality for this PR to preserve current behavior (#5339 will require modifying behavior and some related tests)

great_expectations/data_context/data_context/abstract_data_context.py
donaldheppner: I'd like to see this resolved before merging the PR

donaldheppner: I'd like to see this resolved before merging the PR

donaldheppner (changes requested):

Shinnnyshinshin:
https://github.com/great-expectations/great_expectations/pull/5385#discussion_r911565979

This was addressed in most recent push:  `root_directory` exists in `BaseDataContext`, `FileDataContext` and `CloudDataContext`

donaldheppner (approved):

PR #5404, opened 2022-06-28
[MAINTENANCE] `DataContext` Refactor. Migration of datasource and store

Shinnnyshinshin:
# What has changed? 
### Additional steps of `BaseDataContext` instantiation now migrated to `AbstractDataContext` and related classes
  - initialization of `stores`
  - `construct_data_context_id()`
  - initialization of `usage_statistics`
  - initialization of `datasources`
  - **Note**: This PR leaves one last step of the initialization to a follow-up PR, namely the instantiaionstep remains, which is the instantiation of `DataAssistant`property.  The follow-up PR will result in `get_validator()` and additional `ExpectationSuite methods` to be migrated to `AbstractDataContext` and related methods.
### AbstractMethod created in `AbstractDataContext`
 - `_save_project_config()`: `FileDataContext` and `CloudDataContext` each have their method of saving project_config. `EphemeralDataContext` just writes a logger message. Currently `BaseDataContext` also has a `_save_project_config()` method, since its parent is `EphemeralDataContext` (which doesn't really do a save)
 - `_save_project_config()` is renamed to `_save_project_config_to_file()`

### Improvements to handling of `ValidationOperators` in anonymization step. 
- `ValidationOperators` remain in `BaseDataContext`, and they are no longer used in the V3-API. The logic surrounding `ValidationOperators` was not migrated, and we also allow the anonymizer to anonymize the init_payload without having `validation_operators` included (was required before) 


ghost:
<details>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=5404&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

    [Try it now!](https://app.codesee.io/r/reviews?pr=5404&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)
</details>

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=5404&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map Legend

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
```diff
@@ -149,12 +149,16 @@ def _anonymize_stores_init_payload(
         return anonymized_values
 
     def _anonymize_validation_operator_init_payload(
-        self, payload: Dict[str, "ValidationOperator"]  # noqa: F821
+        self, payload: Optional[Dict[str, "ValidationOperator"]]  # noqa: F821
```
anthonyburdi: If `Optional` should this default to `None`?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -38,6 +68,52 @@ class AbstractDataContext(ABC):
         "/etc/great_expectations.conf",
     ]
     DOLLAR_SIGN_ESCAPE_STRING = r"\$"
+    TEST_YAML_CONFIG_SUPPORTED_STORE_TYPES = [
```
anthonyburdi, line 78: At some point to clean up the length of the data context classes I would love to see these lists and `test_yaml_config` itself moved to a separate class which is invoked from the data context.

great_expectations/data_context/data_context/file_data_context.py
```diff
@@ -44,6 +47,14 @@ def __init__(
     def _init_variables(self) -> FileDataContextVariables:
         raise NotImplementedError
 
+    def _save_project_config(self) -> None:
```
anthonyburdi: Is it worth renaming this method to `_save_project_config_to_disk` for clarity? This would apply to all data contexts.

great_expectations/data_context/data_context/ephemeral_data_context.py
```diff
@@ -41,3 +41,8 @@ def __init__(
 
     def _init_variables(self) -> EphemeralDataContextVariables:
         raise NotImplementedError
+
+    def _save_project_config(self) -> None:
```
anthonyburdi, line 50: Is this necessary? Maybe better to throw an error if this is attempted.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -58,25 +134,578 @@ def __init__(self, runtime_environment: dict):
...
+
+            raise ge_exceptions.StoreConfigurationError(
+                f'Attempted to access the Profiler store named "{profiler_store_name}", which is not a configured store.'
+            )
+
+    @property
+    def concurrency(self) -> Optional[ConcurrencyConfig]:
+        return self.project_config_with_variables_substituted.concurrency
+
+    def add_datasource(
```
anthonyburdi, line 423: I think the `usage_statistics_enabled_method` decorator should be here in the parent class. It would be worth checking to see if there other instances in this PR of methods where the `usage_statistics_enabled_method` decorator is not in the parent class but only in the child class.

anthonyburdi (changes requested):
Started to take a look, glad to see this coming together! I found a few items to address.

great_expectations/data_context/data_context/file_data_context.py
Shinnnyshinshin: I really think so. thanks for brining this up

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: there are definitely going to be a few of those.. I tried to keep all the decorators in the `BaseDataContext` class for now, and was thinking of tackling the `usage_stats` decorators in a follow-up PR. 

What are your thoughts on the sequencing ? Should it be handled this time around?

great_expectations/data_context/data_context/ephemeral_data_context.py
Shinnnyshinshin: Good point :) Converting this into an error.

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
```diff
@@ -93,8 +93,8 @@ def _get_anonymizer(self, obj: object, **kwargs) -> Optional[BaseAnonymizer]:
         return None
 
     def anonymize_init_payload(self, init_payload: dict) -> dict:
-        anonymized_init_payload = {}
-        anonymizer_funcs = {
+        anonymized_init_payload: dict = {}
```
billdirks: Do we know anything about these `dict`s, like the keys are always strings? That is could we make these `Dict[str, Any]`? That's a little more specific than `dict`.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -58,25 +136,560 @@ def __init__(self, runtime_environment: dict):
...
+
+        self._evaluation_parameter_dependencies_compiled = False
+        self._evaluation_parameter_dependencies = {}
 
     @abstractmethod
     def _init_variables(self) -> DataContextVariables:
         raise NotImplementedError
 
+    @abstractmethod
+    def _save_project_config_to_disk(self) -> None:
```
billdirks: A better name for this method might be `_save_project_config` since the subclasses need to implement this and saving to disk doesn't make sense for all of them. In the docstring it seems like you want the `CloudDataContext` to use this method to save to the cloud endpoint.

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -49,9 +52,46 @@ def __init__(
             runtime_environment=runtime_environment,
         )
 
+    def list_expectation_suite_names(self) -> List[str]:
+        """
+        Lists the available expectation suite names. If in ge_cloud_mode, a list of
+        GE Cloud ids is returned instead.
+        """
+        if self.ge_cloud_mode:
```
billdirks: Will `ge_cloud_mode` every `False` in a `CloudDataContext`? Also, this method will return `None` if it is `False` so the return type will need to be updated.

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -49,9 +52,46 @@ def __init__(
...
+        """
+
+        # if in ge_cloud_mode, use ge_cloud_organization_id
+        return self.ge_cloud_config.organization_id
+
+    def _save_project_config_to_disk(self) -> None:
+        """Save the current project to disk."""
+        if self.ge_cloud_mode:
+            logger.debug(
+                "ge_cloud_mode detected - skipping DataContext._save_project_config"
```
billdirks: The log message should match the method name.

Also, in the `EphemeralDataSource` we `raise` while here we log. Should we be doing the same thing or will these be updated to be functional in the future?

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -90,6 +130,37 @@ def get_config_with_variables_substituted(
...
         )
 
+    def save_expectation_suite(
+        self,
+        expectation_suite: ExpectationSuite,
+        expectation_suite_name: Optional[str] = None,
+        overwrite_existing: bool = True,
+        ge_cloud_id: Optional[str] = None,
+        **kwargs,
+    ):
```
billdirks: You need to annotate the return type to `None`. The default return type is `Any`.

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -90,6 +130,37 @@ def get_config_with_variables_substituted(
...
+        key: GeCloudIdentifier = GeCloudIdentifier(
+            resource_type="expectation_suite",
+            ge_cloud_id=ge_cloud_id
+            if ge_cloud_id is not None
+            else str(expectation_suite.ge_cloud_id),
+        )
+        if self.expectations_store.has_key(key) and not overwrite_existing:
+            raise ge_exceptions.DataContextError(
+                "expectation_suite with GE Cloud ID {} already exists. If you would like to overwrite this "
+                "expectation_suite, set overwrite_existing=True.".format(ge_cloud_id)
```
billdirks: It seems like we have a preference for f-strings over format strings in this file.

great_expectations/data_context/data_context/ephemeral_data_context.py
```diff
@@ -41,3 +46,48 @@ def __init__(
 
     def _init_variables(self) -> EphemeralDataContextVariables:
         raise NotImplementedError
+
+    def _save_project_config_to_disk(self) -> None:
+        """Since EphemeralDataContext does not have config as a file, display logging message instead"""
+        raise ge_exceptions.DataContextError(
```
billdirks: Do we really want to `raise` here? It seems like one may call save in a workflow that is agnostic of the underlying data context type. That is raising here would make the `AbstractDataContext` abstraction leaky.

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, salt: Optional[str] = None) -> None:
             for strategy in self._strategies
         }
 
-    def anonymize(self, obj: Optional[object] = None, **kwargs) -> Any:
+    def anonymize(self, obj: Optional[object] = None, **kwargs: Optional[Any]) -> Any:
```
cdkini: Nit - `**kwargs` is generally annotated as `dict` since it is a key-value pairing of arg name to `Any` value.

cdkini: Same note throughout!

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: Additionally, I'd really love to get away from these "magic" strings. I think this isn't entirely feasible due to circular imports and isn't a result of your refactor here but what are your thoughts on the following:
```python
ExpectationsStore.__name__ # instead of "ExpectationStore" name
```

Perhaps something to look into down the line so we have a better association with our existing types and aid `dgtest`.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -58,25 +136,560 @@ def __init__(self, runtime_environment: dict):
...
+        """
+        Each DataContext will define how its project_config will be saved.
+            - FileDataContext : Filesystem.
+            - CloudDataContext : Cloud endpoint
+            - Ephemeral : not saved, and logging message outputted
+        """
+        raise NotImplementedError
```
cdkini: Any reason why we didn't change the existing `_safe_project_config` method? Not 100% sure but I think "disk" generally refers to the filesystem right? Is this naming applicable for all environments and child classes that may use it?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -58,25 +136,560 @@ def __init__(self, runtime_environment: dict):
...
+    def root_directory(self) -> Optional[str]:
+        """The root directory for configuration objects in the data context; the location in which
+        ``great_expectations.yml`` is located.
+
+        Why does this exist in AbstractDataContext? CloudDataContext and FileDataContext both use it
+
+        """
+        if hasattr(self, "_context_root_directory"):
+            return self._context_root_directory
+        return
```
cdkini: Nitpick - instead of a bare `return`, could we return `None` to be very explicit?

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: I think I'm okay defer that to a subsequent PR as long as we aren't breaking anything and it's top of mind - perhaps this is best since we have the release coming up tomorrow and can make larger scale changes after it goes through.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -185,39 +795,18 @@ def _load_config_variables(self) -> Dict:
...
-        Looks for config value in environment_variable and config file section
-
-        Args:
-            environment_variable (str): name of environment_variable to retrieve
-            conf_file_section (str): section of config
-            conf_file_option (str): key in section
-
-        Returns:
-            Optional string representing config value
+        Why does this exist in AbstractDataContext? CloudDataContext and FileDataContext both use it
```
cdkini: I don't love this being on the abstract since it's a utility that isn't entirely applicable to contexts as a whole. That being said, I'd leave it in here for now and maybe leave a comment to refactor it out?

For now, could we make it a static method since it does not rely on instance state?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -295,50 +918,56 @@ def variables(self) -> DataContextVariables:
...
+
+        datasource_store: DatasourceStore = DatasourceStore(
+            store_name=store_name,
+            store_backend=store_backend,
+            runtime_environment=runtime_environment,
+        )
+        self._datasource_store = datasource_store
+        if hasattr(self, "_context_root_directory"):
+            return self._context_root_directory
+        return
```
cdkini: Should this be here? I think this is an error since our store instantiation never returns a value (perhaps an artifact from the context root dir property).

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -757,10 +560,6 @@ def usage_statistics_handler(self) -> Optional[UsageStatisticsHandler]:
     def anonymous_usage_statistics(self):
         return self.project_config_with_variables_substituted.anonymous_usage_statistics
 
-    @property
-    def concurrency(self) -> Optional[ConcurrencyConfig]:
-        return self.project_config_with_variables_substituted.concurrency
-
     @property
     def progress_bars(self) -> Optional[ProgressBarsConfig]:
         return self.project_config_with_variables_substituted.progress_bars
```
cdkini: Should we move all these properties over to the abstract data context?

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -1757,111 +1442,19 @@ def add_datasource(
...
-    def _instantiate_datasource_from_config(
-        self, name: str, config: Union[dict, DatasourceConfig]
-    ) -> Union[LegacyDatasource, BaseDatasource]:
-        """Instantiate a new datasource to the data context, with configuration provided as kwargs.
         Args:
-            name(str): name of datasource
-            config(dict): dictionary of configuration
+            name ():
+            initialize ():
+            **kwargs ():
```
cdkini: Could we flesh this out or remove if not applicable? Perhaps we just add a reference to the parent implementation?

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -3823,60 +3103,60 @@ def _test_instantiation_of_misc_class_from_yaml_config(
...
-                            )
-                        )
-
-        return metric_configurations_list
+    # @staticmethod
+    # def _get_metric_configuration_tuples(metric_configuration, base_kwargs=None):
+    #     if base_kwargs is None:
+    #         base_kwargs = {}
+    #
+    #     if isinstance(metric_configuration, str):
```
cdkini: Why this commented block? Should this be migrated over to abstract as well?

cdkini: Or can we just delete it?

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -49,9 +52,46 @@ def __init__(
...
 
+    def list_expectation_suite_names(self) -> List[str]:
+        """
+        Lists the available expectation suite names. If in ge_cloud_mode, a list of
+        GE Cloud ids is returned instead.
+        """
+        if self.ge_cloud_mode:
+            return [
+                suite_key.ge_cloud_id for suite_key in self.list_expectation_suites()
+            ]
```
cdkini: If we're adhering to this return signature, could we return an empty list if not in `ge_cloud_mode`? Will we always be in `ge_cloud_mode` here?

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -49,9 +52,46 @@ def __init__(
...
+    def _construct_data_context_id(self) -> str:
+        """
+        Choose the id of the currently-configured expectations store, if available and a persistent store.
+        If not, it should choose the id stored in DataContextConfig.
+        Returns:
+            UUID to use as the data_context_id
+        """
+
+        # if in ge_cloud_mode, use ge_cloud_organization_id
+        return self.ge_cloud_config.organization_id
```
cdkini, line 82: ❤️ I love how simple this is compared to before

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -49,9 +52,46 @@ def __init__(
...
+
+    def _save_project_config_to_disk(self) -> None:
+        """Save the current project to disk."""
+        if self.ge_cloud_mode:
+            logger.debug(
+                "ge_cloud_mode detected - skipping DataContext._save_project_config"
+            )
+            return
```
cdkini: Same general note - when will we be using the `CloudDataContext` without being in `ge_cloud_mode`?

great_expectations/data_context/data_context/ephemeral_data_context.py
```diff
@@ -41,3 +46,48 @@ def __init__(
...
+            "EphemeralDataContext has been asked to save project_config to file,"
+            "which is illegal. Please check your config and try again."
+        )
```
cdkini: Hmm so we raise an exception in this case but log with a `CloudDataContext` - do we happen to know why?

great_expectations/data_context/data_context/ephemeral_data_context.py
```diff
@@ -41,3 +46,48 @@ def __init__(
...
+
+    def save_expectation_suite(
+        self,
+        expectation_suite: ExpectationSuite,
+        expectation_suite_name: Optional[str] = None,
+        overwrite_existing: bool = True,
+        **kwargs,
+    ):
```
cdkini: Same note about `**kwargs` - I think we could probably type annotate this whole signature and knock another linter error off the list 😄

great_expectations/data_context/data_context/file_data_context.py
```diff
@@ -44,6 +52,52 @@ def __init__(
...
+        with open(config_filepath, "w") as outfile:
+            self.config.to_yaml(outfile)
+
+    def save_expectation_suite(
+        self,
+        expectation_suite: ExpectationSuite,
+        expectation_suite_name: Optional[str] = None,
+        overwrite_existing: bool = True,
+        **kwargs,
+    ):
```
cdkini: Same note here.

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -195,7 +203,14 @@ def _validate_key(self, key: Tuple[str, ...]) -> None:
             )
 
     def _save_changes(self) -> None:
-        self._data_context._save_project_config()
+        # TODO: this logic will be cleaned up to not use self._data_context
```
cdkini: How do we envision this happening? I don't love this bidirectional relationship between the context and the store backend (each has a reference to each other) but I think this is the direction we wanted to take.

Is there a proposed alternative? Would love to clean this up as well 😄

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -1028,7 +1027,7 @@ def _get_effective_domain_builder_property_value(
     @staticmethod
     def run_profiler(
         data_context: "BaseDataContext",  # noqa: F821
-        profiler_store: ProfilerStore,
+        profiler_store: "ProfilerStore",
```
cdkini: If removing the direct import, could we add the `# noqa: F821` flake8 directive? Please apply throughout

tests/conftest.py
```diff
@@ -3135,7 +3135,9 @@ def alice_columnar_table_single_batch_context(
 ):
     context: DataContext = empty_data_context_stats_enabled
     # We need our salt to be consistent between runs to ensure idempotent anonymized values
-    context._usage_statistics_handler = UsageStatisticsHandler(
+    # <WILL> 20220630 - this is part of the DataContext Refactor and will be removed
+    # (ie. adjusted to be context._usage_statistics_handler)
+    context._data_context._usage_statistics_handler = UsageStatisticsHandler(
```
cdkini, line 3140: I more than trust you with this work but do we have a running list of what needs to be reverted once the full migration goes through? I'm liable to add additional changes that might add to this list with my variables work and just want to make sure nothing slips through the cracks.

Perhaps just a grep for `<WILL>`?

cdkini (changes requested):
This feels like a great step in the right direction! Left lots of comments but most of them are me trying to better understand the overall approach and desired end result.

Happy to discuss these on a synchronous call as needed.

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: My concern was if someone uses these new classes that the usage stats would not be tracked. If it is a fast follow and before these classes are documented for general public use I'm ok with leaving it to a follow-on PR.

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: The reason I suggested to disk is to more clearly describe what this method is doing. If we have a generic method in the abstract class it can call this more descriptive method in its child implementation. I.e. `AbstractDataContext._save_project_config()` is overridden as 
```
class FileDataContext():
    ...
    def _save_project_config(self): 
        self._save_project_config_to_disk()
```

billdirks: I think I'm agreeing with @cdkini and @anthonyburdi which is this method should be called `_save_project_config` here in the abstract class. In the implementation in `FileDataContext` you can define a private method there called `_save_project_config_to_disk` and call it as @anthonyburdi describes above from there. Here, it should have a generic name that is relevant to all implementers.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -58,25 +136,560 @@ def __init__(self, runtime_environment: dict):
...
+            from great_expectations.data_context.store.checkpoint_store import (
+                CheckpointStore,
+            )
+
+            if CheckpointStore.default_checkpoints_exist(
+                directory_path=self.root_directory
+            ):
+                return DataContextConfigDefaults.DEFAULT_CHECKPOINT_STORE_NAME.value
+            if self.root_directory:
+                error_message: str = f'Attempted to access the "checkpoint_store_name" field with no `checkpoints` directory.\n  Please create the following directory: {os.path.join(self.root_directory, DataContextConfigDefaults.DEFAULT_CHECKPOINT_STORE_BASE_DIRECTORY_RELATIVE_NAME.value)}\n  To use the new "Checkpoint Store" feature, please update your configuration to the new version number {float(CURRENT_GE_CONFIG_VERSION)}.\n  Visit https://docs.greatexpectations.io/docs/guides/miscellaneous/migration_guide#migrating-to-the-batch-request-v3-api to learn more about the upgrade process.'
```
billdirks: I'd prefer to see thing string broken up and meet our line length limit unless we have a policy about keeping strings greppable (or some other policy). Python will concatenate multiple line strings for you. You can prepend all lines or the specific line you want with an `f` if you want to interpolate:
```
    error_message: str = ('Attempted to access the "checkpoint_store_name" field '
                                          f'with no `checkpoints` directory.\n ...'
```

I also don't think the `str` annotation is necessary since construction is happening in this line and the correct type will be inferred. I know we like to be explicit so 🤷 .

great_expectations/data_context/data_context/abstract_data_context.py
billdirks: The annotated return type should be made `Optional` if this can return None.

billdirks (commented):
Here is a first pass of comments. I didn't make it through everything, but I see you now have other reviewers as well.

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
Shinnnyshinshin: Thank you very much for this @cdkini. Updating now

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
Shinnnyshinshin: Great point. Thanks @billdirks

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: Thanks for pointing this out. Done and Done

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: Totally right about this.... No idea how this ended up here 😓

great_expectations/data_context/data_context/base_data_context.py
Shinnnyshinshin: great suggestion. There are a few like `ge_cloud_mode`, `ge_cloud_config` that I'd like to propose we move in a subsequent PR.

great_expectations/data_context/data_context/base_data_context.py
Shinnnyshinshin: 😓 my fault. This has been removed now.

great_expectations/data_context/data_context/base_data_context.py
Shinnnyshinshin: docstring has been updated. I think it's worth fleshing out

great_expectations/data_context/data_context/cloud_data_context.py
Shinnnyshinshin: Thank you for catching this. It's an artifact of copy-pasting the code from `BaseDataContext`... we will always be in `ge_cloud_mode`.

Shinnnyshinshin: `list_expectation_suite_names()` has also been updated

great_expectations/data_context/data_context/cloud_data_context.py
Shinnnyshinshin: updated this one as well 🙇🏼 thanks for the catch

great_expectations/data_context/data_context/ephemeral_data_context.py
Shinnnyshinshin: This should both be a logging message

great_expectations/data_context/data_context/ephemeral_data_context.py
Shinnnyshinshin: done!

great_expectations/rule_based_profiler/rule_based_profiler.py
Shinnnyshinshin: Done

tests/conftest.py
Shinnnyshinshin: Great point @cdkini. This is being kept track in the following Confluence Doc : 
- https://superconductive.atlassian.net/wiki/spaces/SUP/pages/423821558/Data+Context+Refactor+-+Working+Document+init+migration

great_expectations/data_context/data_context/cloud_data_context.py
Shinnnyshinshin: Thank you for the catch @billdirks. `ge_cloud_mode` is always true for `CloudDataContext`. 

This was actually brought up by @cdkini and it's been adjusted so the `if self.ge_cloud_mode` check is now gone.

great_expectations/data_context/data_context/cloud_data_context.py
Shinnnyshinshin: Thank you @billdirks

great_expectations/data_context/data_context/cloud_data_context.py
Shinnnyshinshin: Updated. Thank you

great_expectations/core/usage_statistics/anonymizers/anonymizer.py
Shinnnyshinshin: Yes, thanks for the catch.

great_expectations/data_context/data_context/file_data_context.py
Shinnnyshinshin: update: after discussion, this is being renamed back to `_save_project_config()`.

great_expectations/data_context/data_context/ephemeral_data_context.py
Shinnnyshinshin: @anthonyburdi Keeping this as a logger message would maintain the pattern already set by `CloudDataContext`.  Would it be ok to keep it as a logging message for now?

great_expectations/data_context/data_context/ephemeral_data_context.py
Shinnnyshinshin: You're absolutely right. `EphemeralDataContext` and `CloudDataContext` both output a logging message instead now

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: This one was a bit of a journey, but I was able to format this message along with other ones in `abstract_data_context` that were super long.

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: @cdkini on second thought, I think this one actually should be an instance method since it depends on `self.root_directory`.

tests/conftest.py
Shinnnyshinshin: I thought a little more about this, and I think an additional tag would be helpful for this too. I'll add this to the PR

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: I really like this idea.  I started putting the following note throughout the code. I'll finish up in the morning

  `# NOTE: <DataContextRefactor>`

tests/conftest.py
Shinnnyshinshin: As mentioned above: this tag will be added where we need to come back:

`# NOTE: <DataContextRefactor>`

great_expectations/data_context/data_context/cloud_data_context.py
Shinnnyshinshin: I believe these should both be log messages to prevent a valid usecase (EphemeralDataContext and CloudDataContext) from having to throw an error every time.

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -77,6 +77,8 @@ def _get(self, key: Tuple[str, ...]) -> Any:
         if resource_type is DataContextVariableSchema.ALL_VARIABLES:
             return project_config
 
+        # TODO: this logic will be cleaned up to not use `self._data_context`
+        self._data_context._save_project_config()
```
Shinnnyshinshin: This was added by mistake. Removed in most recent commit

great_expectations/data_context/store/inline_store_backend.py
Shinnnyshinshin: Added the following note saying that we will be handling this responsibility through the DataContext Variables

```
 # NOTE: <DataContextRefactor> This responsibility will be moved into DataContext Variables object
```

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: From discussion : We will need to create new events for each "type" of DataContext. This task is being tracked by GREAT-961

Shinnnyshinshin: In a follow-up PR, add decorators to `add_datasource()` and methods like these to siblings. 

If a method is decorated in `BaseDataContext` or `DataContext` AND the method exists in one of the new DataContext objects, then add the decorator there too.

cdkini (approved):

anthonyburdi (approved):
This is a great step forward, thank you for persevering @Shinnnyshinshin!!

PR #5417, opened 2022-06-30
[FEATURE] Add `DataContext.update_datasource` CRUD method

cdkini:
- New `update_datasource` method to enable updating of existing datasources from notebooks
- Updated tests for `DatasourceStore`

### Definition of Done

Thank you for submitting!


great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -1913,6 +1913,22 @@ def _instantiate_datasource_from_config(
...
+        datasource_name: str,
+        datasource_config: DatasourceConfig,
+    ) -> None:
+        """
+        Updates a DatasourceConfig that already exists in the store.
+
+        Args:
+            datasource_name: The name of the Datasource to update.
+            datasource_config: The config object to persist using the DatasourceStore.
+        """
```
billdirks, line 1967: Do you want to add a `Raises` section here or do you not want to bake that into the API?

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -1913,6 +1913,22 @@ def _instantiate_datasource_from_config(
             )
         return datasource
 
+    def update_datasource(
```
donaldheppner, line 1954: It feels like there is more to do here, and that maybe we don't want to pass a config object here but a datasource one. If someone grabs a context with a datasource, makes changes, how do they save it? Other methods work to maintain the cached datasource collection, is that same effort needed here?

donaldheppner (changes requested):

tests/data_context/store/test_datasource_store.py
```diff
@@ -164,3 +175,83 @@ def test_datasource_store_with_inline_store_backend(
...
+    updated_datasource_config = copy.deepcopy(datasource_config)
+    updated_datasource_config.data_connectors["tripdata_monthly_configured"][
+        "base_directory"
+    ] = updated_base_directory
+
+    datasource_store_with_single_datasource.update_by_name(
+        datasource_name=datasource_name, datasource_config=updated_datasource_config
+    )
+
+    key: DataContextVariableKey = DataContextVariableKey(
```
billdirks: My preference would be to remove the annotations on `key` and `actual_config` since you instantiate/cast it in the same line and `mypy` will infer this.

tests/data_context/store/test_datasource_store.py
```diff
@@ -164,3 +175,83 @@ def test_datasource_store_with_inline_store_backend(
...
+        resource_type=DataContextVariableSchema.DATASOURCES,
+        resource_name=datasource_name,
+    )
+    actual_config: DatasourceConfig = cast(
+        DatasourceConfig, datasource_store_with_single_datasource.get(key=key)
```
billdirks, line 241: Just looked and the backing store for this config is a dict and not typed 😢

billdirks (approved):
LGTM!

tests/data_context/store/test_datasource_store.py
cdkini: Ugh yeah our typing needs some work - lots of bags of data that could use structure. We've been incrementally adding dataclasses to provide some of that type safety but it's a work in progress.

tests/data_context/store/test_datasource_store.py
cdkini: I agree! We've been leaning towards extra verbosity but this is quite obvious when we see the constructor call in the same line. Updating now.

great_expectations/data_context/data_context/base_data_context.py
cdkini: I agree! I've modified the signature to take in an actual datasource.

Additionally, I've made it such that `BaseDataContext` doesn't persist by default and `DataContext` does. The cache has also been updated.

Let me know if this is closer to what you're envisioning.

great_expectations/data_context/data_context/base_data_context.py
cdkini: That's a good question! I'm not sure what is a best practice here. I've leaned towards only adding `Raises` if an exception is raised within the top level scope of the function body.

Any preference?

donaldheppner (approved):

great_expectations/data_context/data_context/base_data_context.py
billdirks: There is a small comment about this made in the Google python style guide:
https://google.github.io/styleguide/pyguide.html#doc-function-raises

My reading of it is If you want to codify that raising this exception is part of this functions API and callers should be handling it then document it, otherwise don't. I'm not sure where/how this is planning on being used.

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -1951,6 +1951,31 @@ def _instantiate_datasource_from_config(
...
+        self,
+        datasource_name: str,
+        datasource: Union[LegacyDatasource, BaseDatasource],
```
billdirks, line 1957: `datasource` has a `name` property so passing `datasource_name` now seems redundant.

great_expectations/data_context/data_context/data_context.py
```diff
@@ -471,6 +471,25 @@ def add_datasource(
 
         return new_datasource
 
+    def update_datasource(
+        self,
+        datasource_name: str,
+        datasource: Union[LegacyDatasource, BaseDatasource],
```
billdirks, line 477: Similar comment here about `datasource_name`.

PR #5441, opened 2022-07-05
[MAINTENANCE] Remove `InlineRenderer` invocation feature flag from `ExpectationValidationResult`

NathanFarmer:
- GREAT-911/GREAT-963
- Make `rendered_content` a required property that is only populated if rendering is invoked and drop key during serialization if it is `None`
- Allow `include_rendered_content` to be passed to `Checkpoint` `validations` in addition to `Validator`
- Drop `None` attributes from `RenderedAtomicContent` during serialization
- Add class `RenderedAtomicValueGraph` for serialization when `graph` attribute is accessed directly
- Add test for serialization of `rendered_content` `GraphType`
- Remove feature flag

### Definition of Done


great_expectations/render/types/__init__.py
```diff
@@ -501,44 +508,91 @@ class RenderedAtomicValue(DictDot):
...
 
+    def __repr__(self) -> str:
+        return json.dumps(self.to_json_dict(), indent=2)
+
+    def __str__(self) -> str:
+        return json.dumps(self.to_json_dict(), indent=2)
+
+    def to_json_dict(self) -> dict:
+        """Returns RenderedAtomicValue as a json dictionary."""
+        d = renderedAtomicValueSchema.dump(self)
```
NathanFarmer, line 546: `None` attributes will be dropped during this step and will only be added to the dictionary below if they are found.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -306,6 +306,9 @@ def _run_validation(
             expectation_suite_ge_cloud_id: str = substituted_validation_dict.get(
                 "expectation_suite_ge_cloud_id"
             )
+            include_rendered_content: bool = substituted_validation_dict.get(
+                "include_rendered_content"
+            )
```
NathanFarmer: This allows `Checkpoint` config object `validations` to also accept the `include_rendered_content` argument.

great_expectations/core/expectation_configuration.py
```diff
@@ -1070,8 +1071,12 @@ def kwargs(self, value: dict) -> None:
...
 
     @property
-    def include_rendered_content(self) -> str:
-        return self._include_rendered_content
+    def rendered_content(self) -> dict:
+        return self._rendered_content
+
+    @rendered_content.setter
+    def rendered_content(self, value: dict) -> None:
+        self._rendered_content = value
```
NathanFarmer: `rendered_content` is now a required property, but the key is dropped from serialization if it is `None`.

great_expectations/render/types/__init__.py
```diff
@@ -501,44 +508,91 @@ class RenderedAtomicValue(DictDot):
...
+        if "table" in d:
+            d["table"] = self.table
+        if "graph" in d:
+            d["graph"] = self.graph.to_json_dict()
+        if "kwargs" in d:
+            d["kwargs"] = self.kwargs
+        return d
+
+
+class RenderedAtomicValueGraph(DictDot):
```
NathanFarmer, line 556: This class was built so that the `graph` attr will be properly serialized if it is accessed directly.

great_expectations/validator/validator.py
```diff
@@ -1978,6 +1964,9 @@ def validate(
                 configurations=expectations_to_evaluate,
                 runtime_configuration=runtime_configuration,
             )
+            if self._include_rendered_content:
+                for validation_result in results:
+                    validation_result.render()
```
NathanFarmer, line 1969: `Validator.validate` is used by `Checkpoint.run` so this enables rendering when only a `Checkpoint` is explicitly defined.

great_expectations/checkpoint/checkpoint.py
cdkini: Nit - This is an `Optional[bool]` right? If we want a strict bool, could we set the 2nd param in `get` to be `False`?

great_expectations/checkpoint/util.py
```diff
@@ -206,6 +206,8 @@ def get_substituted_validation_dict(
             validation_dict.get("runtime_configuration", {}),
             dedup=True,
         ),
+        "include_rendered_content": validation_dict.get("include_rendered_content")
+        or substituted_runtime_config.get("include_rendered_content"),
```
cdkini: Same note here - if possible, can we keep the type a `bool` and not create the possibility of a null value?

great_expectations/core/expectation_configuration.py
```diff
@@ -1070,8 +1071,12 @@ def kwargs(self, value: dict) -> None:
         self._kwargs = value
 
     @property
-    def include_rendered_content(self) -> str:
-        return self._include_rendered_content
+    def rendered_content(self) -> dict:
+        return self._rendered_content
```
cdkini: This return type is not necessarily true though is it? Per the constructor, it looks like an `Optional[List[RenderedAtomicContent]]`?

cdkini: It also has the possibility of being a `dict`? I don't have the full picture so please let me know 😄

great_expectations/core/expectation_configuration.py
```diff
@@ -1418,10 +1419,12 @@ class ExpectationConfigurationSchema(Schema):
...
-    REMOVE_KEYS_IF_NONE = ["ge_cloud_id", "expectation_context"]
+    REMOVE_KEYS_IF_NONE = ["ge_cloud_id", "expectation_context", "rendered_content"]
 
     @post_dump
     def clean_null_attrs(self, data: dict, **kwargs):
+        """Removes the attributes in ExpectationConfigurationSchema.REMOVE_KEYS_IF_NONE during serialization if
+        their values are None."""
         data = copy.deepcopy(data)
         for key in ExpectationConfigurationSchema.REMOVE_KEYS_IF_NONE:
             if key in data and data[key] is None:
```
cdkini: Could we possibly annotate this with `**kwargs: dict` and return `-> dict`?

great_expectations/core/expectation_validation_result.py
```diff
@@ -77,8 +81,7 @@ def __init__(
             "exception_traceback": None,
             "exception_message": None,
         }
-        if self.include_rendered_content:
-            self.rendered_content = None
```
cdkini, line 81: Woohoo!

great_expectations/core/expectation_validation_result.py
```diff
@@ -327,15 +341,17 @@ def convert_result_to_serializable(self, data, **kwargs):
...
-    #     # if not hasattr(data, 'meta'):
-    #     #     pass
-    #     # elif len(data.meta) == 0:
-    #     #     del data.meta
-    #     # return data
-    #     pass
+    REMOVE_KEYS_IF_NONE = ["rendered_content"]
+
+    @post_dump
+    def clean_null_attrs(self, data: dict, **kwargs) -> dict:
```
cdkini: Same note with `**kwargs` here.

great_expectations/render/types/__init__.py
```diff
@@ -501,44 +508,91 @@ class RenderedAtomicValue(DictDot):
     def __init__(
         self,
         schema: Optional[dict] = None,
+        header: Optional["RenderedAtomicValue"] = None,
```
cdkini, line 511: Should we add `# noqa` here?

cdkini: It looks like we're able to annotate properly on L520. Could we do that throughout the file if we don't get cyclic import issues?

great_expectations/render/types/__init__.py
```diff
@@ -501,44 +508,91 @@ class RenderedAtomicValue(DictDot):
...
+        if "params" in d:
+            d["params"] = self.params
+        if "header_row" in d:
+            d["header_row"] = self.header_row
+        if "table" in d:
+            d["table"] = self.table
+        if "graph" in d:
+            d["graph"] = self.graph.to_json_dict()
+        if "kwargs" in d:
+            d["kwargs"] = self.kwargs
```
cdkini: I get the need for this but is there a way to make it a bit more concise/maintainable? Can't think of it since `self.graph` is an edge case but just thinking out loud.

great_expectations/render/types/__init__.py
```diff
@@ -552,6 +606,28 @@ class Meta:
...
+        "table",
+        "params",
+        "header_row",
+        "table",
+        "graph",
+        "kwargs",
+    ]
+
+    @post_dump
+    def clean_null_attrs(self, data: dict, **kwargs):
```
cdkini: Same note about kwargs and return.

tests/expectations/test_run_diagnostics.py
```diff
@@ -544,7 +544,7 @@ def test_run_diagnostics_on_an_expectation_with_errors_in_its_tests():
         "error_diagnostics",
         "validation_result",
     }
-    assert tests[3]["test_passed"] == False
+    assert tests[3]["test_passed"] is False
```
cdkini, line 547: Thank you for this 😍

great_expectations/checkpoint/checkpoint.py
NathanFarmer: It is a strict `bool` since we want this as `False` by default. I have updated the `get` with a `False` default! ✅

great_expectations/checkpoint/util.py
NathanFarmer: I have also update this to use `False` if both `validation_dict` and `subsituted_runtime_config` return `None` for `include_rendered_content`. ✅

great_expectations/core/expectation_configuration.py
NathanFarmer: You were initially correct - this should read `Optional[List[RenderedAtomicContent]]`

great_expectations/core/expectation_configuration.py
NathanFarmer: ✅

great_expectations/core/expectation_validation_result.py
NathanFarmer: ✅

great_expectations/render/types/__init__.py
NathanFarmer: Something about the fact that this class has self-referential type hints means that you need to quote in the signature.

great_expectations/render/types/__init__.py
NathanFarmer: ✅

great_expectations/render/types/__init__.py
NathanFarmer: I was able to refactor this with `getattr` ✅

great_expectations/render/types/__init__.py
NathanFarmer: I get "Unresolved reference" when the quotes are removed.

cdkini: ah okay good call, thank you!

cdkini (approved):

roblim (approved):
LGTM!

PR #5475, opened 2022-07-09
[BUGFIX] Rendering Test failures

Shinnnyshinshin:
- follow up to #5474


NathanFarmer (approved):
LGTM!

great_expectations/render/types/__init__.py
```diff
@@ -299,6 +299,9 @@ def __str__(self):
         )
         return string
 
+    def __eq__(self, other):
+        return str(self) == str(other)
+
```
Shinnnyshinshin, line 304: this was necessary because the tests in `tests/render/test_page_renderer.py` were being done between two `RenderedStringTemplateContent` objects, which were not able to be compared directly in some cases.  This `__eq__` method makes the comparison explicit.

PR #5486, opened 2022-07-11
[MAINTENANCE] Adding fixtures Query Expectations tests 

Shinnnyshinshin:
- Very small change to add reading `Titanic.csv` file as fixture 
- In the process, we are also using `file_relative_path()` instead of hard coding a file path, because it provides a more organized way of specifying a path from the path of the current file (`__file__`)
- Not using `file_relative_path()` was causing errors in some cases

### Definition of Done

Thank you for submitting!

### DIFF
--- a/tests/expectations/core/conftest.py
+++ b/tests/expectations/core/conftest.py
@@ -3,6 +3,7 @@ import os
 import pandas as pd
 import pytest
 
+from great_expectations.core.batch import BatchRequest, RuntimeBatchRequest
 from great_expectations.data_context.util import file_relative_path
 
 
@@ -21,3 +22,25 @@ def titanic_df() -> pd.DataFrame:
     )
     df = pd.read_csv(path)
     return df
+
+
+@pytest.fixture
+def sqlite_runtime_batch_request() -> RuntimeBatchRequest:
+    return RuntimeBatchRequest(
+        datasource_name="my_sqlite_db_datasource",
+        data_connector_name="default_runtime_data_connector_name",
+        data_asset_name="titanic",
+        runtime_parameters={"query": "SELECT * FROM titanic LIMIT 100"},
+        batch_identifiers={"default_identifier_name": "test_identifier"},
+        batch_spec_passthrough={"create_temp_table": False},
+    )
+
+
+@pytest.fixture
+def sqlite_batch_request() -> BatchRequest:
+    return BatchRequest(
+        datasource_name="my_sqlite_db_datasource",
+        data_connector_name="default_inferred_data_connector_name",
+        data_asset_name="titanic",
+        batch_spec_passthrough={"create_temp_table": False},
+    )
--- a/tests/expectations/core/conftest.py
+++ b/tests/expectations/core/conftest.py
@@ -1,0 +1,23 @@
+import os
+
+import pandas as pd
+import pytest
+
+from great_expectations.data_context.util import file_relative_path
+
+
+@pytest.fixture
+def titanic_df() -> pd.DataFrame:
+    path = file_relative_path(
+        __file__,
+        os.path.join(
+            "..",
+            "..",
+            "test_fixtures",
+            "configuration_for_testing_v2_v3_migration",
+            "data",
+            "Titanic.csv",
+        ),
+    )
+    df = pd.read_csv(path)
+    return df
--- a/tests/expectations/core/conftest.py
+++ b/tests/expectations/core/conftest.py
@@ -6,7 +6,7 @@ import pytest
 from great_expectations.data_context.util import file_relative_path
 
 
-@pytest.fixture
+@pytest.fixture(scope="module")
 def titanic_df() -> pd.DataFrame:
     path = file_relative_path(
         __file__,
--- a/tests/expectations/core/conftest.py
+++ b/tests/expectations/core/conftest.py
@@ -3,7 +3,6 @@ import os
 import pandas as pd
 import pytest
 
-from great_expectations.core.batch import BatchRequest, RuntimeBatchRequest
 from great_expectations.data_context.util import file_relative_path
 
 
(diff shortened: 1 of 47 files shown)


cdkini (approved):

Kilo59 (approved):

tests/expectations/core/conftest.py
```diff
@@ -0,0 +1,23 @@
+import os
+
+import pandas as pd
+import pytest
+
+from great_expectations.data_context.util import file_relative_path
+
+
+@pytest.fixture
```
Kilo59:
```suggestion
@pytest.fixture(scope="module")
```
Should we change the scope of this fixture to prevent repeated reads of the CSV file and speed up the tests?
https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

The above assumes this won't lead to any test-leakage which may be incorrect.

Kilo59 (approved):

tests/expectations/core/conftest.py
Shinnnyshinshin: This is a great suggestion @Kilo59 thank you :)

tests/expectations/core/conftest.py
```diff
@@ -0,0 +1,23 @@
+import os
+
+import pandas as pd
+import pytest
+
+from great_expectations.data_context.util import file_relative_path
+
+
+@pytest.fixture(scope="module")
+def titanic_df() -> pd.DataFrame:
```
billdirks, line 10: Do we currently use any type of immutable dataframe? That would enforce that this object won't be modified when shared between tests.

tests/expectations/core/test_expect_queried_column_value_frequency_to_meet_threshold.py
```diff
@@ -178,11 +178,9 @@ def test_expect_queried_column_value_frequency_to_meet_threshold_spark(
     warns,
     spark_session,
     basic_spark_df_execution_engine,
+    titanic_df,
 ):
-    df = pd.read_csv(
-        "tests/test_fixtures/configuration_for_testing_v2_v3_migration/data/Titanic.csv"
-    )
-
+    df = titanic_df
```
alexsherstinsky: @Shinnnyshinshin How about adding type hints -- for future benefits.  Thanks!

tests/expectations/core/test_expect_queried_column_value_frequency_to_meet_threshold.py
```diff
@@ -238,10 +236,9 @@ def test_expect_queried_column_value_frequency_to_meet_threshold_override_query_
     warns,
     spark_session,
     basic_spark_df_execution_engine,
+    titanic_df,
 ):
-    df = pd.read_csv(
-        "tests/test_fixtures/configuration_for_testing_v2_v3_migration/data/Titanic.csv"
-    )
+    df = titanic_df
```
alexsherstinsky: @Shinnnyshinshin Same: requesting type hints.

tests/expectations/core/test_expect_queried_table_row_count_to_be.py
```diff
@@ -126,10 +126,9 @@ def test_expect_queried_column_value_frequency_to_meet_threshold_spark(
     row_condition,
     spark_session,
     basic_spark_df_execution_engine,
+    titanic_df,
 ):
-    df = pd.read_csv(
-        "tests/test_fixtures/configuration_for_testing_v2_v3_migration/data/Titanic.csv"
-    )
+    df = titanic_df
```
alexsherstinsky: @Shinnnyshinshin Same: requesting type hints.  Thank you!

tests/expectations/core/test_expect_queried_table_row_count_to_be.py
```diff
@@ -166,10 +165,9 @@ def test_expect_queried_column_value_frequency_to_meet_threshold_override_query_
...
```
alexsherstinsky: @Shinnnyshinshin and here too -- type hints.  Thanks!

alexsherstinsky (changes requested):
@Shinnnyshinshin Since this build needs to be updated, might as well incorporate my small suggestions.  Thank you!

tests/expectations/core/conftest.py
Shinnnyshinshin: Hi @billdirks I dont think we use any immutable dataframes currently, and I really like the idea. Does pandas have a way of specifying this internally? I did a google search and found something called [static frame](https://static-frame.readthedocs.io/en/latest/articles/upgrade.html), which claims to do what you are describing.

tests/expectations/core/test_expect_queried_column_value_frequency_to_meet_threshold.py
Shinnnyshinshin: thanks for the catch :)

tests/expectations/core/test_expect_queried_table_row_count_to_be.py
Shinnnyshinshin: done!

tests/expectations/core/test_expect_queried_table_row_count_to_be.py
Shinnnyshinshin: done

tests/expectations/core/test_expect_queried_column_value_frequency_to_meet_threshold.py
Shinnnyshinshin: done

tests/expectations/core/conftest.py
Kilo59: I like these ideas.

Might also just be able to do it ourselves, and avoid adding another dependency.
Like subclass `DataFrame` and overwrite the `__setitem__` method (or just monkeypatch it).

Another solution might be to switch the scope back to 'function' and use an LRU cache, I'm not sure how the standard [`@funtools.cache`](https://docs.python.org/3/library/functools.html#functools.cache) decorator works with fixtures, might need to do some investigating there. But the basic idea is that it only actually parses the csv the first time the function is called, and it returns the cached dataframe on subsequent calls, which gets around the test-leakage issue.

But also given that this fixture is only being used by a small number of tests and only in this module I think this is fine for now. But definitely at least worth a `TODO` comment.

alexsherstinsky (approved):
LGTM!  Thank you!

PR #5487, opened 2022-07-12
[MAINTENANCE] Move test_yaml_config to a separate class

anthonyburdi:
- Pure refactor - no logic change intended. 
- Move `BaseDataContext.test_yaml_config()` to a separate class called `YamlConfigValidator` which also contains class attributes.
- The scope of this PR is to move this code out of the data context, and not to improve test_yaml_config itself. It also calls some private methods on the data context which should be addressed in the future but are not considered to keep the scope of this PR contained.

### Definition of Done


great_expectations/core/usage_statistics/usage_statistics.py
```diff
@@ -588,12 +588,32 @@ def send_usage_message(
...
+    """Send a usage statistics message using an already instantiated handler."""
+    # noinspection PyBroadException
+    try:
+        if handler is not None:
+            message: dict = {
+                "event": event,
+                "event_payload": event_payload,
+                "success": success,
+            }
             handler.emit(message)
```
Shinnnyshinshin: :) love it

Shinnnyshinshin (approved):
Thank you so much sir

Shinnnyshinshin (changes requested):
Thank you so much for this @anthonyburdi I think it's a huge step in the right direction, but I think there is some remaining work on getting existing tests to pass (other than the query expectations issues we have been experiencing) . Would you be able to have a look?

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -2634,441 +2632,12 @@ def test_yaml_config(
...
-
-        checkpoint_class_args: dict = filter_properties_dict(
-            properties=checkpoint_config,
-            delete_fields={"class_name", "module_name"},
-            clean_falsy=True,
-        )
-
-        if class_name == "Checkpoint":
-            instantiated_class = Checkpoint(data_context=self, **checkpoint_class_args)
-        elif class_name == "SimpleCheckpoint":
```
Kilo59, line 2728: [non-blocking]
I'm missing a lot of context here so please take this with a grain of salt but...

Do you know why we are doing this string comparison on the class name rather than an `isinstance` or `issubclass` check?

great_expectations/data_context/config_validator/yaml_config_validator.py
```diff
@@ -0,0 +1,614 @@
...
+
+        return instantiated_class, usage_stats_event_payload
+
+    def _test_instantiation_of_misc_class_from_yaml_config(
+        self,
+        name: Optional[str],
+        config: CommentedMap,
+        runtime_environment: dict,
+        usage_stats_event_payload: dict,
+    ) -> Tuple[Any, dict]:
```
Kilo59, line 528: [non-blocking]
Is there an ABC class that could be used here for first item in the tuple?
```python
Tuple[SomeABCclass, dict]
```

great_expectations/data_context/config_validator/yaml_config_validator.py
```diff
@@ -0,0 +1,614 @@
...
+
+        usage_stats_event_payload: dict = anonymizer.anonymize(
+            obj=instantiated_class, name=profiler_name, config=profiler_config
+        )
+
+        return instantiated_class, usage_stats_event_payload
+
+    def _test_instantiation_of_misc_class_from_yaml_config(
+        self,
+        name: Optional[str],
```
Kilo59, line 524: [non-blocking]
Just something to consider.
Taken from the FastAPI docs...

>If you are using a Python version below 3.10, here's a tip from my very subjective point of view:

>🚨 Avoid using Optional[SomeType]
>Instead ✨ use Union[SomeType, None] ✨.
>
>Both are equivalent and underneath they are the same, but I would recommend Union instead of Optional because the word "optional" would seem to imply that the value is optional, and it actually means "it can be None", even if it's not optional and is still required.


Kilo59 (commented):
Left some non-blocking comments/questions.

great_expectations/data_context/data_context/base_data_context.py
anthonyburdi: Good question! At this point we are doing a comparison with the string class name from the yaml config passed in. We don't actually have an instantiated class yet.

great_expectations/data_context/config_validator/yaml_config_validator.py
anthonyburdi: I don't think so because this method is more of a catchall if we don't know the type. Maybe in the future we can update if we have a consistent base generic type.

great_expectations/core/usage_statistics/usage_statistics.py
```diff
@@ -588,12 +588,32 @@ def send_usage_message(
...
+
+
+def send_usage_message_from_handler(
+    event: str,
+    handler: Optional[UsageStatisticsHandler] = None,
+    event_payload: Optional[dict] = None,
+    success: Optional[bool] = None,
+) -> None:
+    """Send a usage statistics message using an already instantiated handler."""
+    # noinspection PyBroadException
```
Kilo59, line 609: Would it be worth logging the error at the debug level here?
I'm prepared to accept that It might not be...

great_expectations/core/usage_statistics/usage_statistics.py
```diff
@@ -588,12 +588,32 @@ def send_usage_message(
...
+    try:
+        if handler is not None:
```
Kilo59:
```suggestion
        if handler:
```
In general, I reserve the explicit `is None`/`is not None` check for cases where we need to distinguish between `truthy/falsy` and `None`. But it doesn't look like that's the case here.

Kilo59 (approved):
2 non-blocking questions/suggestions.

great_expectations/core/usage_statistics/usage_statistics.py
anthonyburdi: Good call, added. This is consistent with other methods in this module.

great_expectations/data_context/config_validator/yaml_config_validator.py
```diff
@@ -0,0 +1,623 @@
...
+        Helper to create datasource instance and update usage stats payload.
+        See `test_yaml_config` for more details.
+        """
+        print(f"\tInstantiating as a Datasource, since class_name is {class_name}")
+        datasource_name: str = name or config.get("name") or "my_temp_datasource"
+        datasource_config = datasourceConfigSchema.load(config)
+        datasource_config.name = datasource_name
+        instantiated_class = cast(
+            Datasource,
+            self._data_context._instantiate_datasource_from_config_and_update_project_config(
```
cdkini, line 388: I have reservations about the access of private methods here but understand why we're doing it (was previously not an issue when this method was within the same class).

great_expectations/data_context/config_validator/yaml_config_validator.py
cdkini: If this is indeed something we want to do, we should take care of it across the board - it would be relatively straightforward to generate a script to do this for us if that's the direction we want to take.

great_expectations/data_context/config_validator/yaml_config_validator.py
```diff
@@ -0,0 +1,623 @@
...
+    # Fallback for python < 3.8
+    from typing_extensions import Literal  # type: ignore[misc]
+
+# TODO: check if this can be refactored to use YAMLHandler class
+yaml = YAML()
+yaml.indent(mapping=2, sequence=4, offset=2)
+yaml.default_flow_style = False
+
+
+class _YamlConfigValidator:
```
cdkini, line 59: Is this a convention we're using moving forward? I like it but what exactly does it mean for our users? That a piece of code is not to be touched/lacks extensibility?

cdkini (approved):

PR #5499, opened 2022-07-13
[MAINTENANCE] `/profiler` Cloud endpoint support

cdkini:
- Misc updates to profiler code to make sure it works with our Cloud product
  - Add `ge_cloud_response_json_to_object_dict` to `ProfilerStore`
  - Add `DataContext.add_profiler` to add a top level method to create and save a profiler config.
  - Continual refactor of magic strings to `GeCloudRESTResource` enum

### Definition of Done


great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -28,6 +28,7 @@ class GeCloudStoreBackend(StoreBackend, metaclass=ABCMeta):
         "expectation_suite": "suite",
         "rendered_data_doc": "rendered_data_doc",
         "data_context_variables": "data_context_variables",
+        "profiler": "profiler_config",
```
wookasz:
```suggestion
        "profiler": "profiler",
```

but happy to change this in cloud

great_expectations/data_context/store/profiler_store.py
```diff
@@ -59,3 +59,14 @@ def serialization_self_check(self, pretty_print: bool) -> None:
...
                 f"\tTest key and value successfully removed from Profiler store: {test_value}\n"
             )
+
+    def ge_cloud_response_json_to_object_dict(self, response_json: dict) -> dict:
+        """
+        This method takes full json response from GE cloud and outputs a dict appropriate for
+        deserialization into a GE object
+        """
+        ge_cloud_profiler_id = response_json["data"]["id"]
+        profiler_config_dict = response_json["data"]["attributes"]["profiler_config"]
```
wookasz:
```suggestion
        profiler_config_dict = response_json["data"]["attributes"]["profiler"]
```

to correspond with previous comment

wookasz (approved):

wookasz (commented):
currently, when i call save_profiler and attempt to save without a ge_cloud_id I get this

great_expectations/data_context/store/ge_cloud_store_backend.py
cdkini: nope happy to modify on my end to make it consistent with Cloud! fixing now

great_expectations/data_context/store/profiler_store.py
cdkini: done!

Kilo59 (approved):

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -2185,9 +2188,14 @@ def save_profiler(
...
-        key: Union[
-            GeCloudIdentifier, ConfigurationIdentifier
-        ] = self.profiler_store.determine_key(name=name, ge_cloud_id=ge_cloud_id)
+        key: Union[GeCloudIdentifier, ConfigurationIdentifier]
+        if self.ge_cloud_mode:
+            key = GeCloudIdentifier(
+                resource_type=GeCloudRESTResource.PROFILER, ge_cloud_id=ge_cloud_id
+            )
+        else:
+            key = ConfigurationIdentifier(configuration_key=name)
```
Kilo59, line 2193: Both of these ideas below might be bad. But even if they are good ideas, I don't think they should stop merging 😄 . Can be a future enhancement.

1. How hard would it be to use some polymorphism to derive the key rather than all these if/else blocks?

Like move the id determining logic to a `RuleBasedProfiler.identifier` property.

```python
    def save_profiler(
        self,
        profiler: RuleBasedProfiler,
        name: Optional[str] = None,
    ) -> None:
        self.profiler_store.set(key=profile.key, value=profiler.config)
```

2. Update `ConfigurationIdentifier.__init__` (or `__new__`) so that it can generate `GeCloudIdentifier` as well. Which prevents us from having so much conditional `if cloud else:` logic.

cdkini: oh I definitely agree I don't love the changes here 😬 

I wanted to do the minimal required changes just to get it working but this overall feature needs some TLC. Making modifications to the actual profiler object may cause some issues so I opted for the less elegant solution here.

I believe we're eventually going to add an id onto the profiler itself which should be able to remedy most of this. Would love to hear what other people think though maybe now is the time for such changes?

great_expectations/data_context/store/profiler_store.py
```diff
@@ -59,3 +63,14 @@ def serialization_self_check(self, pretty_print: bool) -> None:
             print(
                 f"\tTest key and value successfully removed from Profiler store: {test_value}\n"
             )
+
+    def ge_cloud_response_json_to_object_dict(self, response_json: dict) -> dict:
```
anthonyburdi, line 67: Love this

anthonyburdi (approved):
LGTM!

great_expectations/data_context/data_context/base_data_context.py
billdirks: I thought that part of @Shinnnyshinshin's changes was to remove branching on `ge_cloud_mode`. I just want to make sure we aren't backtracking his work here.

cdkini: this will all be modified shortly in this sprint as we add an id to the profiler object!

cdkini: but this will unblock Cloud for the time being

PR #5506, opened 2022-07-14
[MAINTENANCE] use invoke for common contrib/dev tasks

Kilo59:
## Changes proposed in this pull request:

~Combing a few developer/contributor quality of life improvements in this PR, but we may want to break these up into separate PRs before merging.~

I've moved anything not related to the `invoke` task file to other PR will be opening those shortly.

1. ~Creating a [`pyproject.toml` file](https://peps.python.org/pep-0518/) to define the build system ge uses and start centralizing tool config to it (black, isort, pytest, etc.)]~ moved -> https://github.com/great-expectations/great_expectations/pull/5525

2. Begin to use the [`invoke` python command line task runner](https://www.pyinvoke.org/) for running common developer tasks. Define a set of starter tasks like linting, formatting, type checking, etc. These can serve as a guide for future tasks. By providing easy to use and understand generic command line tasks like `lint`, `fmt` etc. we make it easy for new contributors to get up and running quickly without having to know anything about the details of our particular tooling. This also makes it easier to change the underlying tools and/or configuration without breaking anyones workflow. 
But contributors are still free to use each of these tools directly.
Alternatives to `invoke`: Make, [duty](https://github.com/pawamoy/duty), [Pipfile scripts](https://pipenv-fork.readthedocs.io/en/latest/advanced.html#custom-script-shortcuts).

4. ~Building on the [excellent work](https://github.com/great-expectations/great_expectations/blob/develop/scripts/check_type_hint_coverage.py) of  @cdkini begin to fully type-check select ge modules.~ moved -> #5540 

5. ~[Enforce our pre-commit hooks in the CI](https://pre-commit.com/#usage-in-continuous-integration) (WIP) - [use pre-commit ci instead of azure?
](https://pre-commit.ci/)~ moved -> #5526 

image

### TLDR Summary 

- ~add pyproject.toml file and move tool config to it~
  - ~move isort config~
  - ~move pytest config~
- [x]  add `invoke` task runner and define common developer tasks
  - [x]  invoke fmt
  - [x]  invoke sort
  - [x]  invoke lint
  - [x]  invoke hooks (run pre-commit hooks)
  - [x]  invoke types (type-check?)
  - [x] update CI to use new invoke tasks
- ~add check-json pre-commit hook~
- ~[update CI to enforce pre-commit hooks](https://pre-commit.com/#usage-in-continuous-integration)~
- ~begin "real" mypy static type checking of select modules~
  - ~fix all typing errors on select modules~
[...]

PR #5526, opened 2022-07-18
[MAINTENANCE] enforce pre-commit in ci

Kilo59:

Enforce our pre-commit hooks with a CI check.
Either in Azure or using the [pre-commit-ci](https://pre-commit.ci/)

### Definition of Done

- ~My code follows the Great Expectations [style guide]~(https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- ~I have commented my code, particularly in hard-to-understand areas~
- [x] I have made corresponding changes to the documentation - waiting on #5506 for this change
- ~I have added [unit tests](https://docs.greatexpectations.io/docs/contributing/contributing_test#writing-unit-and-integration-tests) where applicable and made sure that new and existing tests are passing.~
- ~I have run any local integration tests and made sure that nothing is broken.~

Thank you for submitting!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,6 +2,7 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.2.0
     hooks:
+      - id: check-json
       - id: check-yaml
         exclude: docs/.*|tests/data_context/fixtures/bad_yml/great_expectations\.yml
       - id: end-of-file-fixer
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,6 +2,7 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.2.0
     hooks:
+      - id: check-ast
       - id: check-json
       - id: check-yaml
         exclude: docs/.*|tests/data_context/fixtures/bad_yml/great_expectations\.yml
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,5 +42,8 @@ ci:
     autoupdate_branch: 'develop'
     autoupdate_commit_msg: '[pre-commit.ci] pre-commit autoupdate'
     autoupdate_schedule: monthly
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -30,3 +31,15 @@ repos:
...
+ci:
+    autofix_commit_msg: |
+        [pre-commit.ci] auto fixes from pre-commit.com hooks
+
+        for more information, see https://pre-commit.ci
+    autofix_prs: false
+    autoupdate_branch: 'develop'
+    autoupdate_commit_msg: '[pre-commit.ci] pre-commit autoupdate'
+    autoupdate_schedule: monthly
+    skip: [flake8]
```
Kilo59: Just wanted to touch on a couple things here.
Want to initially role this out without the autofix_prs.

I'm setting it to skip flake8 for 2 reasons...

We already have a linting CI step in azure.
Unlike most pre-commit hooks, it doesn't actually fix the issue. (although neither do check-yaml/check-json)

cdkini: Could we leave a comment in the YAML noting why we skip? Makes sense to me but it might be good to have this context in the config.

cdkini (approved):

.pre-commit-config.yaml
Kilo59: Added.
```
# skip flake8 enforcement with pre-commit CI as it is enforced elsewhere
# flake8 hook also doesn't have an options to auto-fix
# pre-commit hooks should either have mechanisms for auto-fixing or should simple to do manually
```
The last line might be a bit too much editorializing though 😅 . I'm happy to take that out.

tasks.py
```diff
@@ -90,6 +91,11 @@ def hooks(ctx, all_files=False, diff=False):
 
     ctx.run(" ".join(cmds))
 
+    if sync:
+        print("  Re-installing hooks ...")
+        ctx.run(" ".join(["pre-commit", "uninstall"]), echo=True)
+        ctx.run(" ".join(["pre-commit", "install"]), echo=True)
```
Kilo59, line 97: Thanks @cdkini for the suggestion.

PR #5529, opened 2022-07-18
[MAINTENANCE] Add type hints to `Store`

cdkini:
- Add type hints to `Store` class

### Definition of Done

### DIFF
--- a/great_expectations/data_context/store/store.py
+++ b/great_expectations/data_context/store/store.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict, Optional
+from typing import Any, Dict, List, Optional, Tuple, Type
 
 from great_expectations.core.data_context_key import DataContextKey
 from great_expectations.data_context.store.ge_cloud_store_backend import (
@@ -71,7 +71,7 @@ class Store:
         """
         return response_json
 
-    def _validate_key(self, key):
+    def _validate_key(self, key: DataContextKey) -> None:
         # STORE_BACKEND_ID_KEY always validated
         if key == StoreBackend.STORE_BACKEND_ID_KEY:
             return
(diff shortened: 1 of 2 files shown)


Kilo59 (approved):

great_expectations/data_context/store/store.py
```diff
@@ -102,42 +102,46 @@ def store_backend_id(self) -> str:
...
+    def store_backend_id_warnings_suppressed(self) -> str:
         """
         Report the store_backend_id of the currently-configured StoreBackend, suppressing warnings for invalid configurations.
         Returns:
             store_backend_id which is a UUID(version=4)
         """
         return self._store_backend.store_backend_id_warnings_suppressed
 
+    @property
+    def config(self) -> dict:
```
billdirks, line 120: Curious why we use ABC elsewhere but use this runtime check here?

billdirks (approved):
LGTM!

One question inline.

great_expectations/data_context/store/store.py
cdkini: I'm honestly not sure haha - should this be changed?

billdirks: I don't know if I care too strongly but there is something to be said about being consistent in the code-base. Also, an ABC class will die at instantiation instead of when the unimplemented method is called so we know earlier if there is an error. I thought there was static analysis tools that could catch the instantiation of an abstract class but a quick google doesn't reveal this to me. If we could catch it via static analysis, I would have a stronger preference to make it an ABC.

Kilo59: @billdirks 
Pylint can catch this. 

https://vald-phoenix.github.io/pylint-errors/plerr/errors/basic/E0110

PR #5531, opened 2022-07-18
Analytics/generate_example_events_json

tannerbeam:
this is intended to be a v1. i may want to add additional stuff in later versions, but i think this gives me what i need for now. 

based on my understanding of gabriel's new `invoke` utility tool, i believe my workflow will be something like:

- locally (i.e. not in databricks) clone latest gx `develop`
- add my own dependencies (`dbx`, `databricks-cli`, etc.) 
- execute the python script: `invoke tests/integration/usage_statistics/create_usage_statistics_example_messages.py` (i don't know the syntax here)
- copy the resulting json file to dbfs: `databricks fs cp usage_statistics_example_messages.json dbfs:/schemas/example_messages/`


tests/integration/usage_statistics/create_usage_statistics_example_messages.py
```diff
@@ -0,0 +1,18 @@
...
+    keys = [k for k in msg.keys()]
+    msg_list = []
+    for key in keys:
+        for item in msg[key]:
+            msg_list.append(item)
+
+    msg_json = json.dumps(msg_list)
+
+    with open("usage_statistics_example_messages.json", "w") as outfile:
+        json.dump(msg_list, outfile)
```
Kilo59, line 18: If you wanted this to be runnable from the command line you would put this in `tasks.py` at the project root.

```python
import json

import invoke

from tests.integration.test_usage_statistics_messages import valid_usage_statistics_messages

@invoke.task
def gen_usage_stats(ctx, path = "usage_statistics.json"):
    msg = valid_usage_statistics_messages

    # remove top-level items from the messages
    keys = [k for k in msg.keys()]
    msg_list = []
    for key in keys:
        for item in msg[key]:
            msg_list.append(item)

    msg_json = json.dumps(msg_list)

    with open(path, "w") as outfile:
        json.dump(msg_list, outfile)
```

And then this can be run from the command line with `invoke gen-usage-stats`.
Assuming [`invoke` has been installed](https://www.pyinvoke.org/).
`pip install invoke`

tannerbeam: since gen_usage_stats is a function then should i return the json from the function instead of writing it to the repo?

Kilo59: That's up to you.
But yeah I might move the function itself into `/scripts` or maybe a `utils` package.
And then you could import it and call it from the invoke task runner.

Something like...

```python
# invoke.py
import invoke

from scripts import usage_stats

@invoke.task
def gen_usage_stats(ctx, path = "usage_statistics.json"):
   stats = usage_stats.get_stats()

    with open(path, "w") as outfile:
        json.dump(stats, outfile)
```

The only think `invoke` gives you is the ability to easily run python code from the command line, and/or run cmd-line tasks from python.

So the above method is runnable from the command line with `invoke gen-usage-stats` or `invoke --path gen-usage-stats my_file.json` 

See https://github.com/great-expectations/great_expectations/pull/5506 for more examples (btw this should be merging today 🤞 )

tannerbeam: bombing this pr, this branch, and my poor venv that i could not get working correctly based on the requirements files.

PR #5555, opened 2022-07-21
[BUGFIX] Update logic used in `DataContext.__init__()` to persist `data_context_id`

cdkini:
- This PR ensures that if we have a difference between our`project_config` and `self.config` and that difference stems from `anonymous_usage_statistics`, we save changes to disk in the `DataContext` constructor to ensure persistent values for subsequent uses of GX.
- If the difference stems from a global override (conf file or env var), we don't consider that to be a true diff and therefore don't trigger a save to disk. The reason behind this is the fact that those values are dynamic and should remain so - saving to disk makes them static.

### Definition of Done


great_expectations/data_context/data_context/data_context.py
```diff
@@ -363,13 +363,63 @@ def __init__(
             ge_cloud_config=ge_cloud_config,
         )
 
-        # save project config if data_context_id auto-generated or global config values applied
-        project_config_dict = dataContextConfigSchema.dump(project_config)
+        # Save project config if data_context_id auto-generated
+        if self._check_for_usage_stats_sync(project_config):
+            self._save_project_config()
+
+    def _check_for_usage_stats_sync(self, project_config: DataContextConfig) -> bool:
```
Kilo59, line 370: Do we account for differences in how arrays/list items are ordered?

`["foo", "bar"] != ["bar", "foo"]`

I don't have enough context to know if we need to care about that. Or even if configs have nested lists.

Kilo59 (approved):

anthonyburdi (approved):

great_expectations/data_context/data_context/data_context.py
Kilo59: Also on the subject of diffs.
This is a really great (and light-weight) pytest-plugin for visualizing diffs on `assert ==` failures.
https://github.com/hjwp/pytest-icdiff

Granted there aren't actually any test `==` checks in this PR 😄 .

cdkini: ordering shouldn't matter here! and I loveeee that pytest plugin haha good call!

PR #5556, opened 2022-07-21
[FEATURE] `DataAssistantResult` plot multiple metrics per expectation

NathanFarmer:
- GREAT-976/GREAT-863
- GREAT-976/GREAT-864
- Make all charts more general by accepting visualization of multiple metrics per expectation and list-like expectation kwargs

### Definition of Done

Thank you for submitting!


great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -109,6 +120,8 @@ class DataAssistantResult(SerializableDictDot):
         "column.mean": "expect_column_mean_to_be_between",
         "column.median": "expect_column_median_to_be_between",
         "column.standard_deviation": "expect_column_stdev_to_be_between",
+        "column.quantile_values": "expect_column_quantile_values_to_be_between",
+        ("column.min", "column.max"): "expect_column_values_to_be_between",
```
billdirks: Could you add a type annotation to `METRIC_EXPECTATION_MAP`? 

I know we talked about this a little in person and I know the left had side being either a string or a tuple is to make the interface nicer so a user doesn't have to write a bunch of single item tuples when defining this. It seems like you don't want to interact with `METRIC_EXPECTATION_MAP` directly though but through some helper method like `_get_metric_expectation_map` so maybe that method should be public (ie remove the `_`) and this map should be private (add a `_`). Another option would be to make a new class with takes a dict in the initializer and then abstracts away this asymmetry in the keys. I guess this is a long winded way of saying, it would be good not to highlight this asymmetry in a public class variable and we should encapsulate it either by making it private and adding public methods in the class to interact with it or by moving it into it's own class.

To record one last thing we talked about. It is fine the right hand side are strings for now but if more than 1 expectation is derived from a metric they may become `lists` or `tuples` in the future.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -689,22 +719,31 @@ def _display_chart_from_dict(
     ) -> None:
         display_chart_dict[chart_title].display()
 
+    @staticmethod
+    def _get_chart_layer_title(layer: alt.Chart) -> Optional[str]:
```
billdirks: Could you add a unit test for this? It took me a little thinking to convince myself this scenario works:
A is an alt.Chart with no title.
A.layer returns [B, C]
B has no title
C has a title.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -736,30 +775,57 @@ def _apply_theme(
...
+        )
+        list_column_names: List[str] = list(
+            col_has_list[col_has_list["has_list"]]["column_name"]
+        )
+
+        if (
+            "table_columns" in list_column_names
+            and len(np.unique(df["table_columns"])) == 1
+        ):
+            df = df.iloc[:1]
```
billdirks: You are reusing the name `df` here so `df` no longer refers to the input value. Could you pick different name for the new variable defined here?

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -736,30 +775,57 @@ def _apply_theme(
...
+            # replicate rows and add flattened column of lists
+            columns: List[int] = [
+                idx
+                for idx, col in enumerate(df.columns)
+                if col not in list_column_names
+            ]
+            df = df.iloc[ilocations, columns].reset_index(drop=True)
+            cols_flat_df: pd.DataFrame = pd.DataFrame(cols_flat).T
+            cols_flat_df.columns = list_column_names
+            df = pd.concat([df, cols_flat_df], axis=1)
```
billdirks: Whatever name to pick for the new `df` in line 792 you will have to us it here now.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -736,30 +775,57 @@ def _apply_theme(
...
+            ilocations: List[int] = list(
+                np.repeat(range(df.shape[0]), df[list_column_names[0]].apply(len))
+            )
+            # replicate rows and add flattened column of lists
+            columns: List[int] = [
+                idx
+                for idx, col in enumerate(df.columns)
+                if col not in list_column_names
+            ]
+            df = df.iloc[ilocations, columns].reset_index(drop=True)
```
billdirks: I'd use a new variable name here.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -783,16 +849,16 @@ def _get_column_set_text(column_set: List[str]) -> Tuple[str, int]:
...
-        metric_name: str,
+        sanitized_metric_names: Set[str],
         sequential: bool,
         subtitle: Optional[str] = None,
     ) -> alt.Chart:
         """
         Args:
             df: A pandas dataframe containing the data to be plotted
-            metric_name: The name of the metric as it exists in the pandas dataframe
+            sanitized_metric_names: A set containing the names of the metrics as they exist in the pandas dataframe
```
billdirks, line 875: What makes these sanitized? Is this some requirement for the caller and, if so, can you update the argument description?

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -888,15 +933,15 @@ def _get_nominal_metric_chart(
...
+        sanitized_metric_names: Set[str],
         sequential: bool,
         subtitle: Optional[str] = None,
     ) -> alt.Chart:
         """
         Args:
             expectation_type: The name of the expectation
             df: A pandas dataframe containing the data to be plotted
-            metric_name: The name of the metric as it exists in the pandas dataframe
+            sanitized_metric_names: A set containing the names of the metrics as they exist in the pandas dataframe
```
billdirks, line 958: It looks like this method and the one above are identical except for the last if block. I know it was like this before but it's a lot of repeated code and it may be helpful to make 1 helper method both these call.

great_expectations/rule_based_profiler/data_assistant_result/plot_components.py
```diff
@@ -59,13 +53,21 @@ def plot_on_y_axis(self) -> alt.Y:
...
 
         Returns:
-            Either an alt.X or alt.Y instance based on desired axis.
+            An instance of alt.Tooltip containing relevant information from the PlotComponent class.
         """
-        raise NotImplementedError
+        return alt.Tooltip(
+            field=self.name,
+            type=self.alt_type,
+            title=self.name.replace("_", " ").title(),
```
billdirks, line 68: Can you use `self.title` here?

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
```diff
@@ -528,7 +528,7 @@ def test_onboarding_data_assistant_plot_returns_proper_dict_repr_of_column_domai
     plot_result: PlotResult = bobby_onboarding_data_assistant_result.plot_metrics()
 
     column_domain_charts: List[dict] = [p.to_dict() for p in plot_result.charts[2:]]
-    assert len(column_domain_charts) == 85
+    assert len(column_domain_charts) == 100
```
billdirks, line 543: In this and the other test file could you add the mark `@pytest.mark.unit` or `@pytest.mark.integration` as appropriate.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1176,30 +1213,36 @@ def _get_nonsequential_expect_domain_values_to_match_set_isotype_chart(
...
             An altair line chart
         """
         metric_type: alt.StandardType = AltairDataTypes.QUANTITATIVE.value
-        metric_plot_component: MetricPlotComponent = MetricPlotComponent(
-            name=metric_name, alt_type=metric_type
-        )
+        metric_plot_component: MetricPlotComponent
+        metric_plot_components: List[MetricPlotComponent] = []
+        for sanitized_metric_name in sanitized_metric_names:
+            metric_plot_component: MetricPlotComponent = MetricPlotComponent(
```
billdirks: You can drop the annotation here since you are instantiating this class of the right

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1264,19 +1315,25 @@ def _get_expect_domain_values_to_be_between_chart(
         min_value: str = "min_value"
         strict_max: str = "strict_max"
         strict_min: str = "strict_min"
+        quantiles: str = "quantiles"
```
billdirks, line 1332: I know this is the pattern from above and it's fine but I prefer not to have type annotations on the construction line since it is clear what this object is.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1554,373 +1594,351 @@ def _get_interactive_expect_column_values_to_be_between_chart(
...
                 min_value,
                 max_value,
                 strict_min,
                 strict_max,
+                quantiles,
+                allow_relative_error,
             ]
         )
 
+        df_columns: List[str]
```
billdirks: I would remove this declaration since this variable is scoped to inside the for loop below.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1554,373 +1594,351 @@ def _get_interactive_expect_column_values_to_be_between_chart(
...
+        df: Optional[pd.DataFrame] = None
         for _, column_df in column_dfs:
-            df = pd.concat([df, column_df], axis=0)
+            df_columns = column_df.columns.intersection(possible_df_columns)
```
billdirks, line 1670: I'd annotate `df_columns: List[str] = ...` here.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1554,373 +1594,351 @@ def _get_interactive_expect_column_values_to_be_between_chart(
...
-                & (alt.datum.max_value > alt.datum[metric_plot_component.name])
-            ) | (
-                (alt.datum.min_value < alt.datum[metric_plot_component.name])
-                & (alt.datum.max_value < alt.datum[metric_plot_component.name])
+            if df is None:
+                df = pd.DataFrame(columns=df_columns)
+
+            df = pd.concat([df, column_df[df_columns]], axis=0)
+
+        strict_min_predicate: bool = False
```
billdirks, line 1677: You can remove this line since you immediately reassign it in the if statement.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1554,373 +1594,351 @@ def _get_interactive_expect_column_values_to_be_between_chart(
...
             )
-        elif strict_min:
-            predicate = (
-                (alt.datum.min_value > alt.datum[metric_plot_component.name])
-                & (alt.datum.max_value >= alt.datum[metric_plot_component.name])
-            ) | (
-                (alt.datum.min_value < alt.datum[metric_plot_component.name])
-                & (alt.datum.max_value <= alt.datum[metric_plot_component.name])
+
+        strict_max_predicate: bool = False
```
billdirks, line 1688: You can remove this line since you immediately reassign it in the if statement.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1554,373 +1594,351 @@ def _get_interactive_expect_column_values_to_be_between_chart(
...
-            ) | (
-                (alt.datum.min_value < alt.datum[metric_plot_component.name])
-                & (alt.datum.max_value < alt.datum[metric_plot_component.name])
+            if df is None:
+                df = pd.DataFrame(columns=df_columns)
+
+            df = pd.concat([df, column_df[df_columns]], axis=0)
+
+        strict_min_predicate: bool = False
+        if strict_min in df.columns:
```
billdirks, line 1678: This `if` block and the next can be combined like:
```
      for strict in [strict_min, strict_max]:
          if strict in df.columns:
            strict_predicate = bool(df[strict].all())
            expectation_kwarg_plot_components.append(
                ExpectationKwargPlotComponent(
                    name=strict,
                    alt_type=AltairDataTypes.NOMINAL.value,
                    axis_title=y_axis_title,
                )
            )
```

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1554,373 +1594,351 @@ def _get_interactive_expect_column_values_to_be_between_chart(
...
-                strict_min_plot_component=strict_min_plot_component,
-                strict_max_plot_component=strict_max_plot_component,
-                predicate=predicate,
-            )
-        else:
-            return DataAssistantResult._get_interactive_expect_column_values_to_be_between_bar_chart(
+        # encode point color based on anomalies
+        predicates: List[Union[bool, int]] = []
+        for metric_plot_component in metric_plot_components:
+            if strict_min_predicate and strict_max_predicate:
```
billdirks: In the strict case should the inequalities be "<" and ">" instead of ">="/"<="?

Also, is there a way to shorten this by doing 2 `if` blocks, one for mins and one for maxes, instead of iterating through all the min/max pair combos. 
```
if strict_min_predictate:
   # put is strict min predicates
else:
  # put in nonstrict min predicates
```
And then do something similar with maxes?

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1554,373 +1594,351 @@ def _get_interactive_expect_column_values_to_be_between_chart(
...
         df: pd.DataFrame,
-        metric_plot_component: MetricPlotComponent,
+        metric_plot_components: List[MetricPlotComponent],
         batch_plot_component: BatchPlotComponent,
         domain_plot_component: DomainPlotComponent,
-    ) -> alt.Chart:
+    ) -> alt.LayerChart:
         title: alt.TitleParams = determine_plot_title(
-            metric_plot_component=metric_plot_component,
+            metric_plot_components=metric_plot_components,
```
billdirks, line 1839: A few of these methods all start be defining the title and the tooltip. Maybe this could be factored into a helper method?

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1970,53 +2014,205 @@ def _get_expect_domain_values_to_be_between_line_chart(
...
 
-        line: alt.Chart = DataAssistantResult._get_line_chart(
-            df=df,
-            metric_plot_component=metric_plot_component,
+        if "quantiles" in df.columns:
+            lower_limit = lower_limit.encode(detail="quantiles")
+            upper_limit = upper_limit.encode(detail="quantiles")
+            band = band.encode(detail="quantiles")
+
+        metric_name: str
```
billdirks, line 2013: I don't think if we gain anything annotating these types outside the for loop. That is, I don't think python will allocate this once instead of per iteration. I'd move the annotations inside to the assignment points like you do for `point_color_condition`.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1970,53 +2014,205 @@ def _get_expect_domain_values_to_be_between_line_chart(
...
+            # encode point color based on anomalies
+            metric_name = metric_plot_component.name
+            predicate = (
+                (alt.datum.min_value > alt.datum[metric_name])
+                & (alt.datum.max_value > alt.datum[metric_name])
+            ) | (
+                (alt.datum.min_value < alt.datum[metric_name])
+                & (alt.datum.max_value < alt.datum[metric_name])
+            )
+            point_color_condition: alt.condition = alt.condition(
```
billdirks: You can get rid of the type annotation here since the construction is explicit here.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1970,53 +2014,205 @@ def _get_expect_domain_values_to_be_between_line_chart(
...
+        tooltip.extend(batch_plot_component.generate_tooltip())
+        tooltip.extend(expectation_kwarg_tooltips)
+        tooltip.extend(
+            [
+                metric_plot_component.generate_tooltip(format=",")
+                for metric_plot_component in metric_plot_components
+            ]
+        )
+
+        lower_limit: alt.Chart = (
```
billdirks: There is a lot of repitition in this file. Maybe removing some would improve readability? You could combine this like:
```
lower_limit: alt.Chart
upper_limit: alt.Chart
lower_limit, upper_limit = (
  alt.Chart(data=df)
            .mark_line(
                color=expectation_kwarg_line_color,
                strokeWidth=expectation_kwarg_line_stroke_width,
            )
            .encode(
                x=alt.X(
                    batch_plot_component.name,
                    type=batch_plot_component.alt_type,
                    title=batch_plot_component.title,
                    axis=alt.Axis(labels=False),
                ),
                y=component.plot_on_axis(),
                tooltip=tooltip,
            )
            .properties(title=title)
        )
  for component in (max_value_plot_component, min_value_plot_component))

File: great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py [hunk 22]
```diff
@@ -1970,53 +2014,205 @@ def _get_expect_domain_values_to_be_between_line_chart(
...
+            )
+            .properties(title=title)
+        )
+
+        if "quantiles" in df.columns:
+            lower_limit = lower_limit.encode(detail="quantiles")
+            upper_limit = upper_limit.encode(detail="quantiles")
+            band = band.encode(detail="quantiles")
+
+        metric_name: str
```
billdirks, line 2151: Similar comment here about annotating types outside of the for loop.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -2077,66 +2299,81 @@ def _get_expect_domain_values_to_be_between_bar_chart(
...
-        ) | (
-            (alt.datum.min_value < alt.datum[metric_name])
-            & (alt.datum.max_value < alt.datum[metric_name])
-        )
-        bar_color_condition: alt.condition = alt.condition(
-            predicate=predicate,
-            if_false=alt.value(Colors.GREEN.value),
-            if_true=alt.value(Colors.PINK.value),
-        )
+        metric_name: str
```
billdirks, line 2280: Similiar comment about declarations here.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -3156,104 +3390,136 @@ def _filter(domain: Domain) -> bool:
...
+                expectation_plot_impl = (
+                    self._get_expect_domain_values_to_be_between_chart
+                )
+            elif DataAssistantResult._all_metric_names_in_iterable(
+                metric_names=sanitized_metric_names, iterable=temporal_metrics
+            ):
+                expectation_plot_impl = self._get_expect_domain_values_temporal_chart
 
-            return plot_impl(
+            return expectation_plot_impl(
```
billdirks, line 3405: expectation_plot_impl could be `None` here since it starts that way and all these `elif`s could evaluate to false.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -3156,104 +3390,136 @@ def _filter(domain: Domain) -> bool:
...
+                metric_names=sanitized_metric_names, iterable=quantitative_metrics
+            ):
+                metric_plot_impl = self._get_quantitative_metrics_chart
+            elif DataAssistantResult._all_metric_names_in_iterable(
+                metric_names=sanitized_metric_names, iterable=temporal_metrics
+            ):
+                metric_plot_impl = self._get_temporal_metrics_chart
 
-            return plot_impl(
+            return metric_plot_impl(
```
billdirks, line 3444: Similarly `metric_plot_impl` could be None.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -3286,108 +3552,145 @@ def _create_return_charts_for_column_domain_expectation(
...
+            elif DataAssistantResult._all_metric_names_in_iterable(
+                metric_names=sanitized_metric_names, iterable=temporal_metrics
+            ):
+                expectation_plot_impl = (
+                    self._get_interactive_expect_column_values_temporal_chart
+                )
 
             return [
-                plot_impl(
+                expectation_plot_impl(
```
billdirks, line 3616: `expectation_plot_impl` could be None.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -3396,46 +3699,65 @@ def _chart_column_values(
...
+            elif DataAssistantResult._all_metric_names_in_iterable(
+                metric_names=sanitized_metric_names, iterable=temporal_metrics
+            ):
+                plot_impl = self._get_interactive_temporal_metrics_chart
 
             return [
                 plot_impl(
                     column_dfs=column_dfs,
-                    metric_name=metric_name,
+                    sanitized_metric_names=sanitized_metric_names,
```
billdirks, line 3657: `plot_impl` could be None.

billdirks (changes requested):
Thanks for making this change! I read most of the changes to `great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py` but I'm sure I missed some things. In general that file is pretty long and repetitive and I'm wondering if we could/should spend some time, perhaps after this PR, to figure out if we can make it more concise. I know it is not always possible when configuring UI components where they are very similar but different in some crucial ways.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: > Could you add a type annotation to METRIC_EXPECTATION_MAP?

Yes, I figured out how! We couldn't figure it out before, because this is a `dataclass` and a mutable object, but I found a solution using `default_factory`.

NathanFarmer: > It seems like you don't want to interact with METRIC_EXPECTATION_MAP directly though but through some helper method like _get_metric_expectation_map so maybe that method should be public (ie remove the _) and this map should be private (add a _). Another option would be to make a new class with takes a dict in the initializer and then abstracts away this asymmetry in the keys. I guess this is a long winded way of saying, it would be good not to highlight this asymmetry in a public class variable and we should encapsulate it either by making it private and adding public methods in the class to interact with it or by moving it into it's own class.

The private/public methods were scrutinized pretty heavily by Abe and the team, so I am hesitant to make methods public that are only being used within `DataAssistantResult` objects and have no use in the public API. I'm leaning towards making the left hand side only tuples if we feel like this is adding too much complexity. I'm not following what we would gain by adding a new class because the contributors/users would still define this map the same way no?

NathanFarmer: > To record one last thing we talked about. It is fine the right hand side are strings for now but if more than 1 expectation is derived from a metric they may become lists or tuples in the future.

I see this now, thank you!

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: I added this scenario to a unit test in `test_data_assistant_result.py` as well as corner cases. ✅

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: This is a different code block `if/else` so it wouldn't know about variables defined on line 792.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: ✅

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: ✅

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: ✅

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: We should discuss this as a team considering this seems to be pervasive across the codebase. If this variable could have more than one type we would want to annotate it though right?

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: 1) This type-hint was actually wrong, but the editor didn't pick up on that (actually `pd.Index`).
2) Doesn't this annotation indicate that the result of the set operation will always be type `pd.Index` regardless of what is contained in `column_df.columns`?

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: My understanding is that this would be seen as redefining the variable for each loop (yes I understand the type is ignored). [PEP 526](https://peps.python.org/pep-0526/#global-and-local-variable-annotations) recommends to add the annotation before the loop.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: Oops missed that one, thanks!

NathanFarmer: We use this variable to determine how the `predicates` should be formed below, so if this block is never entered we still need `strict_min_predicate=False`.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: We use this variable to determine how the predicates should be formed below, so if this block is never entered we still need strict_min_predicate=False.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: Thank you for this. Somehow I ended up with negative logic after thinking about it too long. I have looked at this thing 15 different ways, but it is solved now. 🙇🏻

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: I could envision a helper method called `determine_plot_title_and_generate_tooltip`, but all it would do is call these two other helper methods we have defined already (`determine_plot_title` and `generate_tooltip`). I'm not crazy about doing that just for the sake of saving a few lines and upon closer reflection little to no lines would be saved since the `PlotComponent`s for each helper would need to be explicitly passed depending on the chart type.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: This is another case where I'm doing it [this way](https://peps.python.org/pep-0526/#global-and-local-variable-annotations). If that is what we decide to do then the `point_color_condition` should be outside the loop as well.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: I agree with this ✅

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: Removed the annotation altogether ✅

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
billdirks: I didn't realize this was so heavily scrutinized. 

My main concern is this adds a new type to the keys (before it was just strings). I don't know where this dictionary is consumed but I'm worried that doing this puts a burden on the user. Are they going to now have to introduce a conditional to check types or do some other processing to prepare the key before doing a lookup? If so, we could centralize it in some accessor so it only appears once in the code and the user doesn't have to deal with it.

Kilo59 (commented):
While you are touching these files could you address these 5 `flake8` warnings,
and then remove `profile` from the [setup.cfg](https://github.com/great-expectations/great_expectations/blob/develop/setup.cfg#L29).

I'd also be happy to open PR against this branch.



```
flake8 . --statistics
./great_expectations/profile/basic_suite_builder_profiler.py:537:5: C901 'BasicSuiteBuilderProfiler._profile' is too complex (28)
./great_expectations/profile/basic_dataset_profiler.py:126:5: C901 'BasicDatasetProfiler._profile' is too complex (18)
./great_expectations/profile/basic_dataset_profiler.py:267:17: E265 block comment should start with '# '
./great_expectations/profile/basic_dataset_profiler.py:268:17: E265 block comment should start with '# '
./great_expectations/profile/user_configurable_profiler.py:818:5: C901 'UserConfigurableProfiler._build_expectations_numeric' is too complex (17)
3     C901 'BasicDatasetProfiler._profile' is too complex (18)
2     E265 block comment should start with '# '
```

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: I agree that this is the preferred method to be more concise, although I will admit that I usually find comprehensions like this less readable 🤣 ✅

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: 🙇🏻

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
billdirks: Sounds good. @Kilo59 wrote [a doc about typing and static analysis](https://superconductive.atlassian.net/wiki/spaces/SUP/pages/435978345/Static+Analysis+Type-Checking) and maybe we can record our practices around this there. 
I agree if there is ambiguity in the type the right hand side could return, such as in your example of more than 1 type, we should annotate. Even if the right hand side can only return one type but it's not obvious at the call site, then I think a type annotation is fine. When the right hand side is a literal or a constructor, I think the annotation is noise.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: 🙇🏻 ✅

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: 🙇🏻 ✅

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
billdirks: I don't think I understand. Aren't you redefining this variable each time? Also, I don't see that part of 526 which you are referring to. I see a section about doing this for the loop variables (`column_df` here).

great_expectations/rule_based_profiler/data_assistant_result/plot_components.py
NathanFarmer: `self.title` refers to the `PlotComponent` title which may be different. e.g. `Column Values` on the axis vs `Column Min` and `Column Max` in the tooltip.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
billdirks: If you have a value in an untyped container and you retrieve it, I believe you must `cast` it to force its type and annotating is not enough.

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
NathanFarmer: 🙇🏻 ✅

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
billdirks: Cool. I see.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
billdirks: 👍

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: ✅

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: The map will be defined by the user in their custom `DataAssistantResult` child wrapper class. It is needed for metrics and expectations that either do not have implemented plots, or are custom to their deployment. The idea is that they shouldn't need to override anything in the parent class that reads from this map. If they are using something already built by us (`OnboardingDataAssistant`) there is no need for them to even be aware of this variable.

tests/rule_based_profiler/data_assistant/test_volume_data_assistant.py
```diff
@@ -2481,6 +2497,7 @@ def test_volume_data_assistant_metrics_plot_descriptive_non_sequential_notebook_
     )
 
 
+@pytest.mark.integration
```
billdirks, line 2500: Thanks for all these markers!

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -60,6 +71,11 @@
 ColumnDataFrame = namedtuple("ColumnDataFrame", ["column", "df"])
 
 
+def default_field(obj: Dict[Union[str, Tuple[str]], str]) -> field:
```
billdirks: It seems like this function is sprinkled across a few files. I think it would be better in an util file that we import.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -1701,53 +1719,30 @@ def _get_interactive_expect_column_values_to_be_between_chart(
...
-                    )
-                    | (
-                        (alt.datum.min_value < alt.datum[metric_plot_component.name])
-                        & (alt.datum.max_value < alt.datum[metric_plot_component.name])
-                    )
+                max_value_predicate = (
+                    alt.datum.max_value >= alt.datum[metric_plot_component.name]
                 )
 
+                predicates.append(min_value_predicate & max_value_predicate)
```
billdirks: This looks to be indented one tab too many?

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -2762,10 +2715,10 @@ def _get_interactive_expect_column_values_to_be_between_line_chart(
             points.selection = alt.Undefined
             points.transform = alt.Undefined
 
-            point_color_condition: alt.condition = alt.condition(
+            point_color_condition = alt.condition(
                 predicate=predicates[idx],
-                if_false=alt.value(Colors.GREEN.value),
-                if_true=alt.value(Colors.PINK.value),
+                if_false=alt.value(Colors.PINK.value),
```
billdirks, line 2716: good catch!

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
billdirks: Also, this is just a nit and I'm fine either way.

billdirks (commented):
Thank you!

My requests are to put `default_field` in a shared location and to check the indentation of `predicates.append(min_value_predicate & max_value_predicate)`.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: This needed to be updated after changing `predicates` to use positive logic.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: Nice catch, it was in the block with the default case so I didn't see the effect.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
NathanFarmer: ✅

billdirks (approved):
LGTM!

PR #5559, opened 2022-07-21
[MAINTENANCE] Refactor `DataContext.__init__` to initialize datasources utilizing a given `DataContextConfig`

cdkini:
- Regardless of which context you are using, datasource initialization should stem from the input config object. This is how it worked in the past but was refactored as part of the `DatasourceStore` work. The current implementation leverages the `DatasourceStore` to populate the `DataContext`'s datasource cache; this came to our attention when we tried to make a series of GET requests to the Cloud backend to populate our datasource cache - why make additional requests when we already have the information we need from our first `GET` to retrieve the project config?

### Definition of Done

### DIFF
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -1473,9 +1473,14 @@ class AbstractDataContext(ABC):
 
     def _init_datasources(self) -> None:
         """Initialize the datasources in store"""
-        datasource_name: str
-        datasource_config: DatasourceConfig
-        for datasource_name, datasource_config in self.config.datasources.items():
+        config: DataContextConfig = self.get_config_with_variables_substituted(
+            self.config
+        )
+        datasources: Dict[str, DatasourceConfig] = cast(
+            Dict[str, DatasourceConfig], config.datasources
+        )
+
+        for datasource_name, datasource_config in datasources.items():
             try:
                 config = copy.deepcopy(datasource_config)
                 config_dict = dict(datasourceConfigSchema.dump(config))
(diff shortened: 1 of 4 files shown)


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1473,19 +1473,20 @@ def _initialize_usage_statistics(
...
+        for datasource_name, datasource_config in datasources.items():
             try:
-                datasource: Optional[
-                    Union[LegacyDatasource, BaseDatasource]
-                ] = self.get_datasource(datasource_name=datasource_name)
+                config = copy.deepcopy(datasource_config)
+                config_dict = dict(datasourceConfigSchema.dump(config))
+                datasource = self._instantiate_datasource_from_config(
+                    name=datasource_name, config=config_dict
+                )
```
cdkini, line 1489: This is all legacy code - instead of iterating through the store's keys (which makes `GET` requests in a Cloud-context), we simply want to use the config we have. There's no point in making subsequent requests if the first request we made as part of `DataContext.__init__` already has the information we need.

tests/conftest.py
```diff
@@ -2334,12 +2334,7 @@ def empty_ge_cloud_data_context_config(
 
 
 @pytest.fixture
-@mock.patch(
-    "great_expectations.data_context.store.DatasourceStore.list_keys",
-    return_value=[],
-)
 def empty_cloud_data_context(
-    mock_list_keys: mock.MagicMock,  # Avoid making a call to Cloud backend during datasource instantiation
```
cdkini: No longer making external requests with datasource init so we remove the mock.

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: Love it

tests/conftest.py
Shinnnyshinshin: Love it

Shinnnyshinshin (approved):
A really big fan of these changes. Thanks for the discussion around this @cdkini and @anthonyburdi

PR #5566, opened 2022-07-22
[MAINTENANCE] Update `DataContextVariables` `__str__` and `__repr__` to only include config

cdkini:
- As variables are a lightweight wrapper around a `DataContextConfig`, we want to make sure we're only showing that information to the user.
- Sensitive details stored in the `self.substitutions` dict should be omittted from output in Jupyter Notebooks.

### Definition of Done

### DIFF
--- a/great_expectations/data_context/data_context_variables.py
+++ b/great_expectations/data_context/data_context_variables.py
@@ -71,6 +71,12 @@ class DataContextVariables(ABC):
         if self.substitutions is None:
             self.substitutions = {}
 
+    def __str__(self) -> str:
+        return str(self.config)
+
+    def __repr__(self) -> str:
+        return repr(self.config)
+
     @property
     def store(self) -> "DataContextStore":  # noqa: F821
         if self._store is None:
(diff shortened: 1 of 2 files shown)


great_expectations/data_context/data_context_variables.py
```diff
@@ -71,6 +71,12 @@ def __post_init__(self) -> None:
         if self.substitutions is None:
             self.substitutions = {}
 
+    def __str__(self) -> str:
+        return str(self.config)
+
+    def __repr__(self) -> str:
+        return repr(self.config)
```
Kilo59, line 78: If a class has no `__str__` it will use `__repr__`, do we need both?

Kilo59 (approved):

anthonyburdi (approved):
LGTM! TIL `repr=False`

great_expectations/data_context/data_context_variables.py
cdkini: nope! I always get confused by these - let me remove `__str__`

Kilo59: Ah but it's a `dataclass` so if you don't do this it will use the default `__str__` which we don't want.
So ignore me 😄

PR #5540, opened 2022-07-19
[MAINTENANCE] begin static type-checking

Kilo59:
Changes proposed in this pull request:
- add `type-check` invoke task that runs mypy against a subset of ge sub-packages
  - ~data_asset~
  - ~exceptions~
  - ~jupyter_ux~
  - [x] profile
  - ~self_check~
  - ~types~
- [x] Get Azure pipeline `type-check` step working

-----

### Edit
Updated PR so that it's easier to review. Only type checking the `profile` package in this PR.
Will open a separate PR with the additionally completed module typings (`data_asset`, `exceptions`, `jupyter_ux`, `self_check`, `type`).

----

[I could have used `cast` more heavily](https://adamj.eu/tech/2021/07/06/python-type-hints-how-to-use-typing-cast/) but IMO that is better used when we understand the code well (which I do not).

The most important thing this PR does is begin the process of static type-checking, we can gradually add more and more packages/modules to the list checked by `invoke type-check`.

~We may want to wait until https://github.com/great-expectations/great_expectations/pull/5525 is merged.~ Avoiding `pyproject.toml` to get this merged sooner.

### Definition of Done
Please delete options that are not relevant.

- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added [unit tests](https://docs.greatexpectations.io/docs/contributing/contributing_test#writing-unit-and-integration-tests) where applicable and made sure that new and existing tests are passing.
- ~I have run any local integration tests and made sure that nothing is broken.~

PR #5576, opened 2022-07-23
[MAINTENANCE] type-checking round 2

Kilo59:

Followup to #5540 

1. Enable type checking on the following packages.
```
data_asset
exceptions
jupyter_ux
self_check
types
```
2. Add `invoke type-check --daemon` flag to easily run `mypy` in daemon mode for faster typing feedback.
- [ ] add `--clear-cache` flag for removing the existing `mypy` cache.

### Definition of Done

- ~I have added [unit tests](https://docs.greatexpectations.io/docs/contributing/contributing_test#writing-unit-and-integration-tests) where applicable and made sure that new and existing tests are passing~
- ~I have run any local integration tests and made sure that nothing is broken.~

Thank you for submitting!

### DIFF
--- a/docs/contributing/style_guides/code_style.md
+++ b/docs/contributing/style_guides/code_style.md
@@ -72,7 +72,12 @@ Our CI system will perform static type-checking using [mypy](https://mypy.readth
 Type-checked modules.
 
 ```
+data_asset
+exceptions
+jupyter_ux
 profile
+self_check
+types
 ```
 
 To verify your code will pass the CI type-checker, run `invoke type-check --install-types`.
--- a/great_expectations/core/expectation_diagnostics/expectation_test_data_cases.py
+++ b/great_expectations/core/expectation_diagnostics/expectation_test_data_cases.py
@@ -22,7 +22,7 @@ class Backend(Enum):
 @dataclass
 class TestBackend:
     backend: str
-    dialects: Optional[List[str]]
+    dialects: List[str]
 
     __test__ = False  # Tell pytest not to try to collect this class as a test
 
--- a/great_expectations/data_asset/data_asset.py
+++ b/great_expectations/data_asset/data_asset.py
@@ -10,7 +10,7 @@ import warnings
 from collections import Counter, defaultdict, namedtuple
 from collections.abc import Hashable
 from functools import wraps
-from typing import List
+from typing import Dict, List, Optional, Union
 
 from dateutil.parser import parse
 
@@ -343,7 +343,9 @@ class DataAsset:
         return outer_wrapper
 
     def _initialize_expectations(
-        self, expectation_suite=None, expectation_suite_name=None
+        self,
+        expectation_suite: Union[Dict, ExpectationSuite, None] = None,
+        expectation_suite_name: Optional[str] = None,
     ) -> None:
         """Instantiates `_expectation_suite` as empty by default or with a specified expectation `config`.
         In addition, this always sets the `default_expectation_args` to:
@@ -372,7 +374,7 @@ class DataAsset:
                 expectation_suite_dict: dict = expectationSuiteSchema.load(
                     expectation_suite
                 )
-                expectation_suite: ExpectationSuite = ExpectationSuite(
+                expectation_suite = ExpectationSuite(
                     **expectation_suite_dict, data_context=self._data_context
                 )
             else:
--- a/great_expectations/exceptions/exceptions.py
+++ b/great_expectations/exceptions/exceptions.py
@@ -304,7 +304,7 @@ class PluginClassNotFoundError(DataContextError, AttributeError):
 
 class ClassInstantiationError(GreatExpectationsError):
     def __init__(self, module_name, package_name, class_name) -> None:
-        module_spec = importlib.util.find_spec(module_name, package=package_name)
+        module_spec = importlib.util.find_spec(module_name, package=package_name)  # type: ignore[attr-defined]
         if not module_spec:
             if not package_name:
                 package_name = ""
--- a/great_expectations/exceptions/exceptions.py
+++ b/great_expectations/exceptions/exceptions.py
@@ -2,6 +2,7 @@ import importlib
 import itertools
 import json
 from collections.abc import Iterable
+from typing import Any, Dict, List, Union
 
 from great_expectations.marshmallow__shade import ValidationError
 
@@ -15,13 +16,13 @@ class GreatExpectationsError(Exception):
 class GreatExpectationsValidationError(ValidationError, GreatExpectationsError):
     def __init__(self, message, validation_error=None) -> None:
         self.message = message
-        self.messages = None
+        self.messages: Union[List[str], List[Any], Dict] = []
         if validation_error is not None:
             self.messages = validation_error.messages
 
     def __str__(self) -> str:
         if self.message is None:
-            return self.messages
+            return str(self.messages)
         return self.message
 
 
--- a/great_expectations/exceptions/exceptions.py
+++ b/great_expectations/exceptions/exceptions.py
@@ -304,7 +304,7 @@ class PluginClassNotFoundError(DataContextError, AttributeError):
 
 class ClassInstantiationError(GreatExpectationsError):
     def __init__(self, module_name, package_name, class_name) -> None:
-        module_spec = importlib.util.find_spec(module_name, package=package_name)  # type: ignore
+        module_spec = importlib.util.find_spec(module_name, package=package_name)
         if not module_spec:
             if not package_name:
                 package_name = ""
--- a/great_expectations/exceptions/exceptions.py
+++ b/great_expectations/exceptions/exceptions.py
@@ -304,7 +304,7 @@ class PluginClassNotFoundError(DataContextError, AttributeError):
 
 class ClassInstantiationError(GreatExpectationsError):
     def __init__(self, module_name, package_name, class_name) -> None:
-        module_spec = importlib.util.find_spec(module_name, package=package_name)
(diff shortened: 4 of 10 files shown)


great_expectations/exceptions/exceptions.py
```diff
@@ -15,13 +16,13 @@ def __init__(self, message) -> None:
 class GreatExpectationsValidationError(ValidationError, GreatExpectationsError):
     def __init__(self, message, validation_error=None) -> None:
         self.message = message
-        self.messages = None
+        self.messages: Union[List[str], List[Any], Dict] = []
```
Kilo59, line 19: run-time change ☝️ and 👇

great_expectations/jupyter_ux/__init__.py
```diff
@@ -60,13 +60,12 @@ def set_data_source(context, data_source_type=None):
...
 <p>
 Found more than one data source in the great_expectations.yml of your project:
-<b>{1:s}</b>
+<b>{:s}</b>
 </p>
 <p>
 Uncomment the next cell and set data_source_name to one of these names.
 </p>
 """.format(
-                        data_source_type,
```
Kilo59, line 69: run-time "change", but shouldn't actually change anything since `data_source_type` wasn't being used.

great_expectations/jupyter_ux/expectation_explorer.py
```diff
@@ -1003,7 +1004,7 @@ def generate_parse_strings_as_datetimes_widget_dict(
         **expectation_kwargs,
     ):
         data_asset_name = expectation_state["data_asset_name"]
-        data_asset = self.state["data_assets"].get(data_asset_name)["data_asset"]
+        data_asset = self.state["data_assets"][data_asset_name]["data_asset"]
```
Kilo59, line 1007: run-time changes above, shouldn't change anything assuming we aren't counting on `AttributeError: 'NoneType' object has no attribute ...` exceptions to drive control flow 😄 .

great_expectations/self_check/util.py
```diff
@@ -1137,13 +1150,12 @@ def build_sa_validator_with_data(
     if (
         schemas
         and sa_engine_name in schemas
-        and isinstance(engine.dialect, dialect_classes.get(sa_engine_name))
+        and isinstance(engine.dialect, dialect_classes[sa_engine_name])
```
Kilo59, line 1153: - [x] Double check this run-time change is okay.

This `.get(sa_engine_name)` isn't any safer then  `[sa_engine_name]`.

If the `sa_engine_name` is not found in the dictionary a `None` will be returned and `isinstance()` will throw a type error.

```python

isinstance({'foo': 'bar'}, None)
>>> TypeError: isinstance() arg 2 must be a type or tuple of types
```

☝️  The above assumes there is no `TypeError` handling code higher up in the stack.

docs/contributing/style_guides/code_style.md
```diff
@@ -72,7 +72,12 @@ Our CI system will perform static type-checking using [mypy](https://mypy.readth
 Type-checked modules.
 
 ```
+data_asset
```
billdirks, line 75: Hooray!

great_expectations/core/expectation_diagnostics/expectation_test_data_cases.py
```diff
@@ -22,7 +22,7 @@ class Backend(Enum):
 @dataclass
 class TestBackend:
     backend: str
-    dialects: Optional[List[str]]
+    dialects: List[str]
```
billdirks, line 25: I love it when we remove `Optional`.

great_expectations/jupyter_ux/__init__.py
```diff
@@ -60,13 +60,12 @@ def set_data_source(context, data_source_type=None):
                     """
 <p>
 Found more than one data source in the great_expectations.yml of your project:
-<b>{1:s}</b>
+<b>{:s}</b>
```
billdirks, line 63: What does removing the `1` do? Is it indexing into something?

tasks.py
```diff
@@ -140,25 +140,39 @@ def type_coverage(ctx):
         "packages": f"One or more packages to type-check with mypy. (Default: {DEFAULT_PACKAGES_TO_TYPE_CHECK})",
         "show-default-packages": "Print the default packages to type-check and then exit.",
         "install-types": "Automatically install any needed types from `typeshed`.",
+        "daemon": "Run mypy in daemon mode with faster analysis."
+        " The daemon will be started and re-used for subsequent calls."
+        " For detailed usage see `dmypy --help`.",
```
Kilo59, line 145: mypy_times

☝️ Timing `mypy` runs, with and without using the mypy `daemon`.

1. First `mypy` run with a clean cache.
2. Second `mypy` run with a loaded cache. Slight improvement (~1 second).
3. First `dmypy` run without daemon running. About the same time as normal `mypy` check.
4. Second `dmypy` run with daemon already running. Massive improvement. `~7.5 seconds` to less than `0.2 seconds`

great_expectations/data_asset/data_asset.py
```diff
@@ -343,7 +343,9 @@ def wrapper(self, *args, **kwargs):
         return outer_wrapper
 
     def _initialize_expectations(
-        self, expectation_suite=None, expectation_suite_name=None
+        self,
+        expectation_suite: Union[Dict, ExpectationSuite, None] = None,
```
Kilo59, line 347: Could this be a list too?

great_expectations/self_check/util.py
```diff
@@ -159,7 +172,7 @@
         )
         try:
             getattr(sqla_bigquery, "INTEGER")
-            bigquery_types_tuple = {}
+            bigquery_types_tuple = {}  # type: ignore[var-annotated]
```
billdirks, line 175: Are there tradeoffs between adding this `ignore`s vs marking it as `Any`?

tasks.py
billdirks: cool! If the cache get's in a bad state how does one clear it?

billdirks (commented):
Thanks, this looks good! I have 1 comment about the change in `<b>{:s}</b>` that I'd like to understand before giving the approve.

great_expectations/data_asset/data_asset.py
cdkini: I don't believe so - also, should we annotate with `NoneType` instead of `None`?

tasks.py
```diff
@@ -114,19 +114,19 @@ def type_coverage(ctx):
...
-    # "jupyter_ux",  # 4
+    "jupyter_ux",  # 0
     # "marshmallow__shade",  # 14
     "profile",  # 0
     # "render",  # 87
     # "rule_based_profiler",  # 469
-    # "self_check",  # 10
-    # "types",  # 3
+    "self_check",  # 0
+    "types",  # 0
```
cdkini, line 129: woohooo!

tasks.py
Kilo59: You would delete the `.mypy_cache/` directory that sits at the project root.

I can add a convenience `--clear-cache` flag to `invoke type-check` to make this easier.

great_expectations/jupyter_ux/__init__.py
Kilo59: Yes, it's saying to use the positional argument at index `1` (the second argument).
https://docs.python.org/3/library/string.html#format-examples

`mypy` didn't like that `data_source_type` was being unused, so I removed it and took away the index here.

I also could have changed this to `<b>{0:s}</b>`.

Kilo59: Also here's a good page about the different ways of doing string formatting in python.

https://realpython.com/python-string-formatting/#which-string-formatting-method-should-you-use

The site is pretty good but you will have to login with github/google.

great_expectations/self_check/util.py
Kilo59: I generally try to avoid `Any`s if I can.

I tend to think of an `Any` as a more permanent, "we are giving up on narrowing this type".

Whereas this `ignore` is more of a "we just aren't dealing with this now but we might come back to it".

Honestly, the name of this variable and its usage were confusing.

great_expectations/jupyter_ux/__init__.py
billdirks: Oh, I didn't see the `format` part below. Thanks! And thanks for the link!

billdirks (approved):

great_expectations/core/expectation_diagnostics/expectation_test_data_cases.py
Kilo59: Yes, always having to guard against `Optionals` is annoying.

Based on usage and the `mypy` warnings, the annotation appeared to be wrong, it couldn't be `None`. So I didn't have to make any run-time changes.

great_expectations/data_asset/data_asset.py
Kilo59: I'm not familiar with `NoneType`.
My cursory google didn't find anything, do you have a reference to how it should be used?

>`Optional[X]` is equivalent to `X | None` (or `Union[X, None]`)

https://docs.python.org/3/library/typing.html#typing.Optional

Kilo59: On a second attempt, I found some mentions of `NoneType`, which is the same as `type(None)`.
But for type hints, it's common practice just to use the `None` object itself.

https://docs.python.org/3/library/types.html#types.NoneType
https://peps.python.org/pep-0484/#using-none
https://stackoverflow.com/questions/21706609/where-is-the-nonetype-located

If you see something that suggests we should be using `NoneType` instead of `None` I'm all ears.

PR #5593, opened 2022-07-25
[ANALYTICS] add new invoke tasks to `tasks.py` and create new file `usage_stats_utils.py`

tannerbeam:
- create a utility module in `tests/integration/usage_statistics` for generating usage stats example messages
- add two invoke tasks to `tasks.py` for writing and moving usage stats example messages

Note:
Ran `invoke fmt` and `invoke type-check --install-types` on files but did  not run full suite of integration tests.

### Definition of Done


Shinnnyshinshin (changes requested):
HI @Super-Tanner thank you so much for your work on this PR. I tried playing around with the `azure-pipeline`, and think that it might make things easier to move the `usage_stats` tests to a different file / different pipeline. 

The motivation around this request is around the importing of `valid_usage_statistics_messages` which requires the `great-expectations` package to be installed wherever the test is running. 

One candidate might be : 
* `tests/core/usage_statistics/test_usage_stats_schema.py`, where we already do some usage stats tests, mostly centered around `schema` validation. 
* This would allow for the tests to run in an environment where they have access to the GE package, while still ensuring that they run with the rest of the CI/CD pipeline


Would be happy to hop on a call to discuss if you'd like. Let me know

tasks.py
```diff
@@ -8,9 +8,12 @@
 
 To show task help page `invoke <NAME> --help`
 """
+import json
+
 import invoke
 
 from scripts import check_type_hint_coverage
+from tests.integration.usage_statistics import usage_stats_utils
```
Kilo59: IMO this should be in something like `/scripts` not `/tests`, like the `check_type_hint_coverage` example above.

Shouldn't need to install `pytest` in order to run our type-checkings tasks.

tasks.py
```diff
@@ -162,3 +165,31 @@ def type_check(ctx, packages, install_types=False, show_default_packages=False):
...
+    version = usage_stats_utils.get_gx_version()
+
+    outfile = f"v{version}_example_events.json"
+    with open(outfile, "w") as f:
+        json.dump(events, f)
+
+    print(f"File written to '{outfile}'.")
+
+
+@invoke.task(get_usage_stats_json, aliases=["move-stats"])
```
```suggestion
@invoke.task(pre=[get_usage_stats_json], aliases=["move-stats"])
```

What you already have works fine, but this is slightly more expressive.
https://docs.pyinvoke.org/en/0.11.0/concepts/execution.html#pre-and-post-tasks

Shinnnyshinshin:
Added proposed changes to Loom : @Super-Tanner @Kilo59 

https://www.loom.com/share/9c7f57ace43e471fa763a093d26f42ea

* Encapsulated importing of GE into `try-except`
* Tested both code paths using `invoke`

Shinnnyshinshin:
@Super-Tanner It looks good on my end :) @Kilo59 could you give a final stamp if you feel like the PR is ready?

Shinnnyshinshin (approved):

tasks.py
```diff
@@ -194,3 +202,42 @@ def type_check(
...
+@invoke.task(aliases=["get-stats"])
+def get_usage_stats_json(ctx):
+    """
+    Dump usage stats event examples to json file
+    """
+    if not is_ge_installed:
+        print(
+            "This invoke task requires Great Expecations to be installed in the environment. Please try again."
+        )
+        return
```
```suggestion
    if not is_ge_installed:
       raise invoke.Exit(
            message="This invoke task requires Great Expecations to be installed in the environment. Please try again.",
            code=1
        )
```

Great job. The only suggestion I have would be to make this small, change.
This will cause the task to exit early just as with the return but it uses a `non-zero` exit code which indicates to the calling shell that the task did not succeed.
https://docs.pyinvoke.org/en/stable/api/exceptions.html#invoke.exceptions.Exit

You might also be able to remove the similar logic from `mv_usage_stats_json` as well since `get_usage_stats_json` will always be called first.
You may want to double-check that though.

But these are minor improvements. I'm approving even if you decide not to make this change.

Kilo59 (approved):
Looks good. 1 minor suggestion, but I think it's totally fine to merge as is.

tasks.py
Shinnnyshinshin: oo this is MUCH better. thank you @Kilo59

tannerbeam:
let's gooooooo

PR #5629, opened 2022-07-29
[BUGFIX] Patch flaky E2E GX Cloud tests

cdkini:
- Introduce `--cloud` flag for `pytest`
- Update Azure pipelines to utilize new flag
- Remove inconsistencies from tests
  - Env var substitution - as this one fixture is shared across a few tests, let's minimize changes
  - Remove randomization of suite name - this will prevent overpopulating the dev env with test suites

### Definition of Done

### DIFF
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -980,6 +980,7 @@ class AbstractDataContext(ABC):
     def delete_expectation_suite(
         self,
         expectation_suite_name: Optional[str] = None,
+        ge_cloud_id: Optional[str] = None,
     ) -> bool:
         """Delete specified expectation suite from data_context expectation store.
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -107,6 +107,10 @@ def pytest_configure(config):
         "markers",
         "aws_integration: runs aws integration test that may be very slow and requires credentials",
     )
+    config.addinivalue_line(
+        "markers",
+        "cloud: runs GX Cloud tests that may be slow and requires credentials",
+    )
 
 
 def pytest_addoption(parser):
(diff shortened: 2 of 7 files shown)


tests/data_context/test_base_data_context.py
```diff
@@ -175,8 +175,7 @@ def _closure(context: DataContext) -> Tuple[Validator, str]:
         suites = context.list_expectation_suites()
         expectation_suite_ge_cloud_id = suites[0].ge_cloud_id
 
-        # Randomize name so subsequent tests use a clean slate
-        suite_name = f"suite_{''.join(random.choice(string.ascii_letters + string.digits) for _ in range(8))}"
+        suite_name = "oss_e2e_test_suite"
```
anthonyburdi, line 177: Should we first delete if the suite exists before creating again on each test invocation to make sure we start with a fresh suite and that there are no issues with cross contamination between tests? I know we do `overwrite_existing=True` online 182 but just to be safe that we are starting clean.

anthonyburdi (approved):
Just a suggestion to delete the existing suite before creating the new one.

tests/data_context/test_base_data_context.py
cdkini: done!

tests/conftest.py
```diff
@@ -251,15 +258,23 @@ def pytest_collection_modifyitems(config, items):
     if config.getoption("--docs-tests"):
         # --docs-tests given in cli: do not skip documentation integration tests
         return
+    if config.getoption("--cloud"):
```
billdirks, line 261: Should this `if` block be here? It looks like it early exits if any of these flags are present so passing any of them is equivalent to passing all of them.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -980,6 +980,7 @@ def create_expectation_suite(
     def delete_expectation_suite(
         self,
         expectation_suite_name: Optional[str] = None,
+        ge_cloud_id: Optional[str] = None,
```
billdirks, line 983: This is already merged but I had a question about this:
1. It doesn't look like we do anything with `ge_cloud_id`. Is this to prep for future work?
2. We have `expectation_suite_name` as `Optional` but it looks like we immediately `raise` if the `has_key(key)` fails where key is the name. Should we remove the `Optional` annotation on this argument?

PR #5631, opened 2022-07-29
[MAINTENANCE] `RuleBasedProfiler` corner case configuration changes

NathanFarmer:
- GREAT-761/GREAT-1137
- Change default `quantile_statistic_interpolation_method` to `nearest` for bootstrap estimator due to issue with bimodal distributions of `float`s.
- Add option to turn of bias correction and make it disabled by default
- Increase precision for numeric rules back to 15 from 12

### Definition of Done


great_expectations/expectations/core/expect_column_max_to_be_between.py
```diff
@@ -153,8 +154,9 @@ class ExpectColumnMaxToBeBetween(ColumnExpectation):
                     "estimator": "bootstrap",
                     "n_resamples": 9999,
                     "random_seed": None,
-                    "quantile_statistic_interpolation_method": "auto",
-                    "quantile_bias_std_error_ratio_threshold": 0.25,
+                    "quantile_statistic_interpolation_method": "nearest",
```
alexsherstinsky, line 157: @NathanFarmer Let us not add the setting for `quantile_statistic_interpolation_method` to be "nearest" as a constant.  It is not needed.

great_expectations/expectations/core/expect_column_mean_to_be_between.py
```diff
@@ -144,8 +145,9 @@ class ExpectColumnMeanToBeBetween(ColumnExpectation):
                     "estimator": "bootstrap",
                     "n_resamples": 9999,
                     "random_seed": None,
-                    "quantile_statistic_interpolation_method": "auto",
-                    "quantile_bias_std_error_ratio_threshold": 0.25,
+                    "quantile_statistic_interpolation_method": "nearest",
```
alexsherstinsky, line 148: @NathanFarmer Let us leave `quantile_statistic_interpolation_method` as "auto" -- thanks.

great_expectations/expectations/core/expect_column_median_to_be_between.py
```diff
@@ -144,8 +145,9 @@ class ExpectColumnMedianToBeBetween(ColumnExpectation):
                     "estimator": "bootstrap",
                     "n_resamples": 9999,
                     "random_seed": None,
-                    "quantile_statistic_interpolation_method": "auto",
-                    "quantile_bias_std_error_ratio_threshold": 0.25,
+                    "quantile_statistic_interpolation_method": "nearest",
+                    "quantile_bias_correction": False,
```
alexsherstinsky, line 149: @NathanFarmer If we do not yet have a docstring stating that if `quantile_bias_correction` is False, then `quantile_bias_std_error_ratio_threshold` has no effect, then please let us add it associated with those fields where arguments for `NumericMetricRangeMultiBatchParameterBuilder` class are explained.

alexsherstinsky (changes requested):
Some comments inside.

great_expectations/expectations/core/expect_column_median_to_be_between.py
NathanFarmer: `NumericMetricRangeMultiBatchParameterBuilder` docstring was updated to read:
```
quantile_bias_correction: Applicable only for the "bootstrap" sampling method -- if omitted (default), then
                False is used (bias correction is disabled) and quantile_bias_std_error_ratio_threshold will have no
                effect.
```

great_expectations/expectations/core/expect_column_mean_to_be_between.py
NathanFarmer: This will remain `nearest` per our offline conversation.

great_expectations/expectations/core/expect_column_max_to_be_between.py
NathanFarmer: This will remain `nearest` per our offline conversation.

great_expectations/rule_based_profiler/rule/rule.py
```diff
@@ -112,8 +112,8 @@ def run(
         """
         variables = build_parameter_container_for_variables(
             variables_configs=reconcile_rule_variables(
-                variables=variables,
-                variables_config=convert_variables_to_dict(variables=self.variables),
+                variables=self.variables,
+                variables_config=convert_variables_to_dict(variables=variables),
```
NathanFarmer, line 116: This bug resulted in runtime profiler config variables being substituted with default profiler config variables (should be the other way around).

great_expectations/rule_based_profiler/data_assistant/onboarding_data_assistant.py
```diff
@@ -734,8 +736,9 @@ def _build_datetime_columns_rule() -> Rule:
             "estimator": "bootstrap",
             "n_resamples": 9999,
             "random_seed": None,
-            "quantile_statistic_interpolation_method": "auto",
-            "quantile_bias_std_error_ratio_threshold": 0.25,
+            "quantile_statistic_interpolation_method": "nearest",
```
alexsherstinsky, line 739: @NathanFarmer While you are at it, could you please change `round_decimals` from wherever it says `12` to be `15` in all variables dictionaries for both data assistant classes?  Thanks!

great_expectations/rule_based_profiler/rule/rule.py
alexsherstinsky: @NathanFarmer I believe that this may not necessarily be the correct fix.  The logic for variables reconciliation was put together for a reason.  We actually hold the "overrides", supplied in the method arguments, as the "base".  Could we please consider solving the issue of the default being overwritten differently?  Thanks!

alexsherstinsky (changes requested):
That variable reconciliation fix might not work properly -- we need a different solution, perhaps at the level of class variables, which @billdirks discovered.  Thanks.

donaldheppner (dismissed):
I want to pump the breaks here until we're aligned on how we want the profilers to behave

great_expectations/rule_based_profiler/data_assistant/onboarding_data_assistant.py
NathanFarmer: ✅

great_expectations/rule_based_profiler/rule/rule.py
NathanFarmer: @alexsherstinsky Can you help me understand how using the "overrides" and then updating them (reconciliation strategy is "update") with the defaults would work? It seems to produce the incorrect profiler configuration by always using the defaults that have been specified. It seems to me that the test `test_quentin_expect_column_quantile_values_to_be_between_auto_yes_default_profiler_config_yes_custom_profiler_config_yes` previously passed due to chance, but once I updated the default config the unintended behavior was revealed.

NathanFarmer: @alexsherstinsky Also, this bug seems to be completely separate and apart from the state issues we have seen with using a custom config before using the default config or a different custom config.

alexsherstinsky (approved):
LGTM!

alexsherstinsky:
@NathanFarmer #5687 fixes a typo; apparently `dgtest` did not catch it (it is hard for it to look for strings). /cc @cdkini -- thanks!

PR #5632, opened 2022-07-29
[BUGFIX] Pass `--cloud` flag to `dgtest-cloud-overrides` section of Azure YAML

cdkini:
- Refactor `pytest` flag collection and underlying logic to ensure we don't exit early if a flag is passed.
- Only run Cloud-specific tests in `dgtest-cloud-overrides` (currently running the full suite yet again).

### Definition of Done

### DIFF
--- a/azure-pipelines-dependency-graph-testing.yml
+++ b/azure-pipelines-dependency-graph-testing.yml
@@ -360,7 +360,7 @@ stages:
           - script: |
               # These are tests that are specific to Great Expectations Cloud.
               # In order to ensure coverage, we run them during each CI/CD cycle.
-              pytest -m cloud
+              pytest --cloud
 
             env:
               GE_CLOUD_BASE_URL: $(GE_CLOUD_BASE_URL)
--- a/azure-pipelines-dependency-graph-testing.yml
+++ b/azure-pipelines-dependency-graph-testing.yml
@@ -360,7 +360,7 @@ stages:
           - script: |
               # These are tests that are specific to Great Expectations Cloud.
               # In order to ensure coverage, we run them during each CI/CD cycle.
-              pytest -m cloud
+              pytest -m cloud --cloud
 
             env:
               GE_CLOUD_BASE_URL: $(GE_CLOUD_BASE_URL)
--- a/azure-pipelines-dependency-graph-testing.yml
+++ b/azure-pipelines-dependency-graph-testing.yml
@@ -360,7 +360,7 @@ stages:
           - script: |
               # These are tests that are specific to Great Expectations Cloud.
               # In order to ensure coverage, we run them during each CI/CD cycle.
-              pytest --cloud
+              pytest -m cloud
 
             env:
               GE_CLOUD_BASE_URL: $(GE_CLOUD_BASE_URL)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import copy
 import datetime
 import locale
 import logging
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -107,6 +107,10 @@ def pytest_configure(config):
         "markers",
         "aws_integration: runs aws integration test that may be very slow and requires credentials",
     )
+    config.addinivalue_line(
+        "markers",
+        "cloud: runs GX Cloud tests that may be slow and requires credentials",
+    )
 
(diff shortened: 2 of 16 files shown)


azure-pipelines-dependency-graph-testing.yml
```diff
@@ -360,7 +360,7 @@ stages:
           - script: |
               # These are tests that are specific to Great Expectations Cloud.
               # In order to ensure coverage, we run them during each CI/CD cycle.
-              pytest -m cloud
+              pytest -m cloud --cloud
```
cdkini, line 363: Enables Cloud tests but also narrows the collected tests to just those items.

tests/conftest.py
```diff
@@ -252,29 +252,37 @@ def pytest_generate_tests(metafunc):
...
 def pytest_collection_modifyitems(config, items):
-    if config.getoption("--aws-integration"):
-        # --aws-integration given in cli: do not skip aws-integration tests
-        return
-    if config.getoption("--docs-tests"):
-        # --docs-tests given in cli: do not skip documentation integration tests
-        return
-    if config.getoption("--cloud"):
-        # --cloud given in cli: do not skip GX Cloud integration tests
-        return
```
cdkini, line 263: Previously, if any of these were set, the skip process would end early. This logic has been updated (thank you @billdirks!)

azure-pipelines-dependency-graph-testing.yml
Shinnnyshinshin: Just wanting to clarify what is the difference between `-m cloud` and `--cloud`?

cdkini: `-m` will ONLY collect tests that have that marker. `--cloud` will enable them to actually be run.

```sh
pytest --cloud           # Run ALL tests (including Cloud ones)
pytest -m cloud          # Collect Cloud tests (but do not run due to not having --cloud)
pytest --cloud -m cloud  # Collect Cloud tests AND run them
```

If there's a more elegant solution here, please let me know :)

Shinnnyshinshin (approved):
Thank you for this @cdkini :)

azure-pipelines-dependency-graph-testing.yml
Shinnnyshinshin: works for me for now :) Thanks for teh additional clarity

tests/conftest.py
```diff
@@ -252,29 +252,37 @@ def pytest_generate_tests(metafunc):
...
+    for category in categories:
+        # If flag is provided, exit early so we don't add `pytest.mark.skip`
+        if config.getoption(category.flag):
+            continue
+
+        # For each test collected, check if they use a mark that matches our flag name.
+        # If so, add a `pytest.mark.skip` dynamically.
+        for item in items:
+            if category.mark in item.keywords:
+                marker = pytest.mark.skip(reason=category.reason)
```
billdirks, line 285: Not blocking: You could put the `marker` in the the dataclass instead of the `reason` and then you wouldn't have to create one each time.

tests/conftest.py
```diff
@@ -252,29 +252,37 @@ def pytest_generate_tests(metafunc):
...
-    for item in items:
-        if "aws_integration" in item.keywords:
-            item.add_marker(skip_aws_integration)
-        if "docs" in item.keywords:
-            item.add_marker(skip_docs_integration)
-        if "cloud" in item.keywords:
-            item.add_marker(skip_cloud)
+    for category in categories:
+        # If flag is provided, exit early so we don't add `pytest.mark.skip`
+        if config.getoption(category.flag):
```
billdirks, line 278: Not blocking: You traverse the list of tests for each flag that is set. The list of test is >> than the list of categories. Alternatives:

* You could skip this continue and for each test, check if it's in each category. 
* You could do through the categories first, collect the list of set ones and then go through the tests once and check for the set categories.

I don't think we need to worry about this too much, just wanted to point it out since it seems like you are doing some optimization here.

billdirks (approved):
Thanks for fixing this!

Some comments about optimizing but they are not blocking.

azure-pipelines-dependency-graph-testing.yml
billdirks: This was very useful for me too!

PR #5525, opened 2022-07-18
[MAINTENANCE] move tool config to pyproject.toml - fork

Kilo59:
Changes proposed in this pull request:

Splitting from #5506 
* Creating a [pyproject.toml file](https://peps.python.org/pep-0518/) to define the build system ge uses and start centralizing tool config to it (black, isort, pytest, etc.)]
* Removing run time args from `black` and `isort`. Pulling config from `pyproject.toml` will force consistent usage of the tools in `ci`, local dev environments and via `pre-commit` hooks. 

Here's some lists of tools that support a `pyroject.toml` config
https://github.com/carlosperate/awesome-pyproject
https://github.com/flying-sheep/awesome-python-packaging

### TODO
- [x] resolve potential `versioneer` issue
- [x] remove runtime args for `isort` and `black`
  - [x] set `black` & `isort` config in `pyproject.toml`, prevent the need to keep various `ci` & `pre-commit` runtime parameters in sync. 
- [x] update `isort` version

### Notable References
https://setuptools.pypa.io/en/latest/build_meta.html

### Definition of Done
Please delete options that are not relevant.

- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code
- ~I have commented my code, particularly in hard-to-understand areas~
- ~I have made corresponding changes to the documentation~
- ~I have added [unit tests](https://docs.greatexpectations.io/docs/contributing/contributing_test#writing-unit-and-integration-tests) where applicable and made sure that new and existing tests are passing.~
- [x] ~I have run any local integration tests~ Reinstalled `great_expectations` locally and made sure that nothing is broken.

PR #5633, opened 2022-07-29
[MAINTENANCE] DEVREL-2617/black_declarations

AFineDayFor:
- Keep things in line with each-other.

```> git diff origin/develop
-              black --check --exclude docs/ . || EXIT_STATUS=$?
+              black --check --exclude  '(docs/.*|tests/.*.fixture|.*.ge_store_backend_id)' . || EXIT_STATUS=$?
```

```
> grep 'id: black' -A1 .pre-commit-config.yaml
      - id: black
        exclude: docs/.*|tests/.*.fixture|.*.ge_store_backend_id
```

### Definition of Done

- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code in that it runs in bash correctly and as expected.

### DIFF
--- a/azure-pipelines-dependency-graph-testing.yml
+++ b/azure-pipelines-dependency-graph-testing.yml
@@ -100,7 +100,7 @@ stages:
               pip install isort[requirements]==5.4.2 flake8==3.8.3 black==22.3.0 pyupgrade==2.7.2
               EXIT_STATUS=0
               isort . --check-only --skip docs/ || EXIT_STATUS=$?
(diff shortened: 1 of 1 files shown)


Kilo59:
We could control this via `pyproject.toml` config, and not have to worry about any run-time args.
https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-via-a-file

We can do the same for `isort`.

This PR is to create the initial `pyproject.toml` but it doesn't have any black configs...yet.
https://github.com/great-expectations/great_expectations/pull/5525

PR #5638, opened 2022-08-01
[FEATURE] Update `DataContext.add_checkpoint()` to ensure validations within `CheckpointConfig` contain ids

cdkini:
- Ensure that `add_checkpoint` returns a Checkpoint with ids (including nested values within validations).
- Write a series of integration tests to cover new behavior for all Cloud-backed contexts.

### Definition of Done


great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -352,3 +352,73 @@ def _instantiate_datasource_from_config_and_update_project_config(
                 raise e
 
         return datasource
+
+    def add_checkpoint(
+        self,
+        name: str,
+        config_version: Optional[Union[int, float]] = None,
```
cdkini, line 359: Require this level of verbosity to maintain parity with `BaseDataContext` - can we break off from that pattern here or do we need consistency?

great_expectations/data_context/store/store.py
```diff
@@ -188,3 +189,12 @@ def self_check(self, pretty_print: bool) -> None:
         NotImplementedError(
             f"The test method is not implemented for Store class {self.__class__.__name__}."
         )
+
+    def _build_key_from_config(self, config: AbstractConfig) -> DataContextKey:
```
cdkini, line 193: Promoted to base as it is used throughout the hierarchy

great_expectations/data_context/store/store_backend.py
```diff
@@ -263,6 +264,22 @@ def _has_key(self, key):
     def remove_key(self, key) -> None:
         del self._store[key]
 
+    def build_key(self, name: Optional[str], id_: Optional[str]) -> DataContextKey:
```
cdkini: Promoted to base as it is used throughout the hierarchy

tests/data_context/cloud_data_context/conftest.py
```diff
@@ -0,0 +1,20 @@
+from typing import Callable
+
+import pytest
+
+
+class MockResponse:
```
cdkini, line 6: Shared in both files so utilize `conftest` to act as global state

tests/data_context/cloud_data_context/test_datasource_crud.py
```diff
@@ -250,14 +236,15 @@ def json(self):
         ),
     ],
 )
-def test_data_context_in_cloud_mode_add_datasource(
+def test_cloud_data_context_add_datasource(
```
cdkini: Namespace collision with other test within this file

great_expectations/data_context/store/store_backend.py
anthonyburdi: Nice! Let's use this instead of the version I added to #5640

great_expectations/data_context/store/store_backend.py
```diff
@@ -263,6 +264,22 @@ def _has_key(self, key):
...
+
+        Args:
+            name: The name of the object to be stored
+            id_: The id of the object to be stored
+
+        Returns:
+            key: The key generated as a result of the input name and id_
+
+        Raises:
+            NotImplementedError if not defined by child classes in the StoreBackend hierarchy
```
anthonyburdi: I think this should be optional currently for other StoreBackends

cdkini: So are you saying we should remove the `raise`? I was originally going to use `pass` but I think debugging could be difficult if we don't immediately yell at the caller for using the unimplemented method.

The docstring should definitely signal that this is optional though.

tests/data_context/cloud_data_context/test_checkpoint_crud.py
```diff
@@ -0,0 +1,169 @@
...
+    validation_id_1, validation_id_2 = validation_ids
+
+    with mock.patch(
+        "requests.post", autospec=True, side_effect=mocked_post_response
+    ) as mock_post, mock.patch(
+        "requests.get", autospec=True, side_effect=mocked_get_response
+    ) as mock_get:
+        checkpoint = context.add_checkpoint(**checkpoint_config)
+
+        assert mock_post.call_count == 1
```
anthonyburdi: I think here we want to assert the post was called with the right params

tests/data_context/cloud_data_context/test_checkpoint_crud.py
```diff
@@ -0,0 +1,169 @@
...
+            BaseDataContext,
+            id="BaseDataContext",
+        ),
+        pytest.param("empty_data_context_in_cloud_mode", DataContext, id="DataContext"),
+        pytest.param(
+            "empty_cloud_data_context", CloudDataContext, id="CloudDataContext"
+        ),
+    ],
+)
+def test_base_data_context_in_cloud_mode_add_checkpoint(
```
anthonyburdi: Love the fixtures! In that case, maybe make the test name a bit more generic?

tests/data_context/cloud_data_context/test_datasource_crud.py
anthonyburdi: Yes! Thank you - fixed in https://github.com/great-expectations/great_expectations/pull/5640 as well.

anthonyburdi (commented):
Just a few light suggestions, looking good though!

tests/data_context/cloud_data_context/test_datasource_crud.py
Kilo59: Would it be useful to organize the cloud-data-context vs base-data-context in separate test classes?
https://docs.pytest.org/en/7.1.x/getting-started.html?highlight=testclass#group-multiple-tests-in-a-class

Then you can avoid the namespace collision and use shorter, simpler test names.
Also makes it easier to maintain the appropriate pytest markers by simply marking the different test classes, instead of having to mark every test.

There are also some opportunities to declare shared parameters + fixtures in a single place. Although that might not be too useful in this particular case. 😄 
https://stackoverflow.com/a/62176555/6304433

great_expectations/data_context/store/store_backend.py
Kilo59: @anthonyburdi 
As long as it isn't decorated as an `@abstractmethod` *it is* "optional".

I think ` raise NotImplementedError` is appropriate here.
This just ensures that we catch cases where someone attempts to call `build_key()` but it doesn't force implementation of this method if it would never be called.

Kilo59 (approved):

tests/data_context/cloud_data_context/test_datasource_crud.py
cdkini: Ooh might be worth looking into - I'd vote to defer test refactoring to a subsequent PR/ticket just to get this work across the line but I'm happy to discuss!

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -2075,19 +2075,16 @@ def add_checkpoint(
...
             site_names=site_names,
             slack_webhook=slack_webhook,
             notify_on=notify_on,
             notify_with=notify_with,
             ge_cloud_id=ge_cloud_id,
             expectation_suite_ge_cloud_id=expectation_suite_ge_cloud_id,
         )
-
-        self.checkpoint_store.add_checkpoint(checkpoint, name, ge_cloud_id)
+        self._synchronize_self_with_underlying_data_context()
```
anthonyburdi, line 2087: 👍

great_expectations/data_context/store/store_backend.py
anthonyburdi: For sure re: absence of decorator - I think this line just threw me off `NotImplementedError if not defined by child classes in the StoreBackend hierarchy`

great_expectations/data_context/store/store_backend.py
```diff
@@ -263,6 +264,24 @@ def _has_key(self, key):
...
```
anthonyburdi: How about this? Nitpick so feel free to ignore.
```suggestion
            NotImplementedError if used without being overridden by child classes in the StoreBackend hierarchy
```

tests/data_context/cloud_data_context/test_checkpoint_crud.py
```diff
@@ -0,0 +1,204 @@
...
+                        "checkpoint_config": expected_checkpoint_config,
+                        "organization_id": ge_cloud_organization_id,
+                    },
+                },
+            },
+            headers=request_headers,
+        )
+
+        mock_get.assert_called_with(
+            f"{ge_cloud_base_url}/organizations/{ge_cloud_organization_id}/contracts/",
```
anthonyburdi: I think this should have the checkpoint id in the second call rather than just to the `/contracts/` endpoint so we get the specific checkpoint.

tests/data_context/cloud_data_context/test_datasource_crud.py
anthonyburdi: I agree we can do a better job of organizing the CRUD tests, but I'd vote for a separate PR to separate the diffs from moving things around.

anthonyburdi (changes requested):
Getting close! Just a few items to address.

anthonyburdi (approved):
LGTM! Thank you 🙇

PR #5639, opened 2022-08-01
[MAINTENANCE] add `--clear-cache` flag for `invoke type-check`

Kilo59:

- Minor followup to https://github.com/great-expectations/great_expectations/pull/5576#discussion_r934621741 regarding how to clear the `mypy` cache if it get's into a bad state.

- echo the `black` and `isort` calls made by `invoke fmt`

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -8,6 +8,9 @@ To show all available tasks `invoke --list`
 
 To show task help page `invoke <NAME> --help`
 """
+import pathlib
+import shutil
+
 import invoke
 
 from scripts import check_type_hint_coverage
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -54,7 +54,7 @@ def fmt(ctx, path=".", sort_=True, check=False, exclude=None):
         cmds.append("--check")
     if exclude:
         cmds.extend(["--exclude", exclude])
-    ctx.run(" ".join(cmds))
+    ctx.run(" ".join(cmds), echo=True)
```
Kilo59, line 60: Screen Shot 2022-08-01 at 4 36 57 PM

tasks.py
```diff
@@ -155,6 +161,11 @@ def type_check(
         print("\n".join(DEFAULT_PACKAGES_TO_TYPE_CHECK))
         raise invoke.Exit(code=0)
 
+    if clear_cache:
+        print("  Clearing the mypy cache ... ", end="")
+        ctx.run(" ".join(["rm", "-rf", ".mypy_cache"]))
```
Kilo59: Only issue... this isn't windows friendly.
Could update this to use pure python. With one of `shutil`/`os`/`pathlib`.

billdirks: I'm fine with this. If the goal is for contributors to use this then we should use one of your alternatives.

tasks.py
billdirks: Thank you!

billdirks (commented):
This LGTM, thank you!

I didn't hit approve because automerge is on and I didn't know if you wanted to make this os independent. Let me know and I can hit the button.

billdirks (approved):
LGTM, thanks!

PR #5640, opened 2022-08-01
[FEATURE] Enable retrieve by name for datasource with cloud store backend

anthonyburdi:
- Enable using a datasource `name` to retrieve a datasource configuration from the GeCloudStoreBackend
- Uses a query param to hit the tbd `GET /organizations/<org_id>/datasources?name=<name>` endpoint.

### Definition of Done

Thank you for submitting!

### DIFF
--- a/great_expectations/data_context/store/datasource_store.py
+++ b/great_expectations/data_context/store/datasource_store.py
@@ -120,7 +120,9 @@ class DatasourceStore(Store):
         """
         datasource_key: Union[
             DataContextVariableKey, GeCloudIdentifier
-        ] = self.store_backend.build_key(name=datasource_name)
+        ] = self.store_backend.build_key(
+            resource_type=DataContextVariableSchema.DATASOURCES, name=datasource_name
+        )
         if not self.has_key(datasource_key):
             raise ValueError(
                 f"Unable to load datasource `{datasource_name}` -- no configuration found or invalid configuration."
@@ -160,7 +162,11 @@ class DatasourceStore(Store):
             name = datasource_config.name
         else:
             name = None
-        return self.store_backend.build_key(name=name, id_=id_)
+        return self.store_backend.build_key(
+            resource_type=DataContextVariableSchema.DATASOURCES,
+            name=name,
+            id_=id_,
+        )
 
     def set_by_name(
         self, datasource_name: str, datasource_config: DatasourceConfig
--- a/great_expectations/data_context/store/datasource_store.py
+++ b/great_expectations/data_context/store/datasource_store.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import copy
 from typing import List, Optional, Tuple, Union
 
@@ -90,6 +92,19 @@ class DatasourceStore(Store):
         else:
             return self._schema.loads(value)
 
+    def ge_cloud_response_json_to_object_dict(self, response_json: dict) -> dict:
+        """
+        This method takes full json response from GE cloud and outputs a dict appropriate for
+        deserialization into a GE object
+        """
+        datasource_ge_cloud_id: str = response_json["data"]["id_"]
+        datasource_config_dict: dict = response_json["data"]["attributes"][
+            "datasource_config"
+        ]
+        datasource_config_dict["ge_cloud_id"] = datasource_ge_cloud_id
+
+        return datasource_config_dict
+
     def retrieve_by_name(self, datasource_name: str) -> DatasourceConfig:
         """Retrieves a DatasourceConfig persisted in the store by it's given name.
 
@@ -103,9 +118,9 @@ class DatasourceStore(Store):
         Raises:
             ValueError if a DatasourceConfig is not found.
         """
-        datasource_key: DataContextVariableKey = self._determine_datasource_key(
-            datasource_name=datasource_name
-        )
+        datasource_key: Union[
+            DataContextVariableKey, GeCloudIdentifier
+        ] = self.store_backend.build_key(name=datasource_name)
         if not self.has_key(datasource_key):
             raise ValueError(
                 f"Unable to load datasource `{datasource_name}` -- no configuration found or invalid configuration."
--- a/great_expectations/data_context/store/ge_cloud_store_backend.py
+++ b/great_expectations/data_context/store/ge_cloud_store_backend.py
@@ -7,6 +7,9 @@ from urllib.parse import urljoin
 
 import requests
 
+from great_expectations.data_context.data_context_variables import (
+    DataContextVariableSchema,
+)
 from great_expectations.data_context.store.store_backend import StoreBackend
 from great_expectations.data_context.types.refs import GeCloudResourceRef
 from great_expectations.data_context.types.resource_identifiers import GeCloudIdentifier
@@ -387,9 +390,12 @@ class GeCloudStoreBackend(StoreBackend, metaclass=ABCMeta):
         return self._config
 
     def build_key(
-        self, id_: Optional[str] = None, name: Optional[str] = None
+        self,
+        resource_type: DataContextVariableSchema,
+        id_: Optional[str] = None,
+        name: Optional[str] = None,
     ) -> GeCloudIdentifier:
-        """Get the store backend specific implementation of the key."""
+        """Get the store backend specific implementation of the key. ignore resource_type since it is defined when initializing the cloud store backend."""
         return GeCloudIdentifier(
             resource_type=self.ge_cloud_resource_type,
             ge_cloud_id=id_,
--- a/great_expectations/data_context/store/ge_cloud_store_backend.py
+++ b/great_expectations/data_context/store/ge_cloud_store_backend.py
@@ -148,17 +148,16 @@ class GeCloudStoreBackend(StoreBackend, metaclass=ABCMeta):
 
     def _get(self, key: Tuple[str, ...]) -> dict:
         ge_cloud_url = self.get_url_for_key(key=key)
+        params: Optional[dict] = None
         try:
-            params: dict
             # if name is included in the key, add as a param
             if len(key) > 2 and key[2]:
                 params = {"name": key[2]}
                 ge_cloud_url = ge_cloud_url.rstrip("/")
-                response = requests.get(
-                    ge_cloud_url, headers=self.auth_headers, params=params
-                )
-            else:
-                response = requests.get(ge_cloud_url, headers=self.auth_headers)
+
+            response = requests.get(
+                ge_cloud_url, headers=self.auth_headers, params=params
+            )
             return response.json()
         except JSONDecodeError as jsonError:
             logger.debug(
--- a/great_expectations/data_context/store/ge_cloud_store_backend.py
+++ b/great_expectations/data_context/store/ge_cloud_store_backend.py
@@ -7,10 +7,6 @@ from urllib.parse import urljoin
 
 import requests
 
-from great_expectations.core.data_context_key import DataContextVariableKey
-from great_expectations.data_context.data_context_variables import (
-    DataContextVariableSchema,
-)
 from great_expectations.data_context.store.store_backend import StoreBackend
 from great_expectations.data_context.types.refs import GeCloudResourceRef
 from great_expectations.data_context.types.resource_identifiers import GeCloudIdentifier
--- a/great_expectations/data_context/store/ge_cloud_store_backend.py
+++ b/great_expectations/data_context/store/ge_cloud_store_backend.py
@@ -386,14 +386,12 @@ class GeCloudStoreBackend(StoreBackend, metaclass=ABCMeta):
     def config(self) -> dict:
         return self._config
 
-    def build_key(self, id_: Optional[str] = None, **kwargs) -> GeCloudIdentifier:
-        """Get the store backend specific implementation of the key, ignore irrelevant kwargs."""
-        if "name" in kwargs:
-            return GeCloudIdentifier(
-                resource_type=self.ge_cloud_resource_type,
-                ge_cloud_id=id_,
-                resource_name=kwargs["name"],
-            )
+    def build_key(
+        self, id_: Optional[str] = None, name: Optional[str] = None
+    ) -> GeCloudIdentifier:
+        """Get the store backend specific implementation of the key."""
         return GeCloudIdentifier(
-            resource_type=self.ge_cloud_resource_type, ge_cloud_id=id_
+            resource_type=self.ge_cloud_resource_type,
+            ge_cloud_id=id_,
+            resource_name=name,
         )
--- a/great_expectations/data_context/store/ge_cloud_store_backend.py
+++ b/great_expectations/data_context/store/ge_cloud_store_backend.py
@@ -7,6 +7,10 @@ from urllib.parse import urljoin
 
 import requests
 
+from great_expectations.core.data_context_key import DataContextVariableKey
+from great_expectations.data_context.data_context_variables import (
+    DataContextVariableSchema,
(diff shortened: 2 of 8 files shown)


ghost:
<details>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=5640&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

    [Try it now!](https://app.codesee.io/r/reviews?pr=5640&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)
</details>

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=5640&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map Legend

great_expectations/data_context/store/datasource_store.py
```diff
@@ -103,9 +118,9 @@ def retrieve_by_name(self, datasource_name: str) -> DatasourceConfig:
         Raises:
             ValueError if a DatasourceConfig is not found.
         """
-        datasource_key: DataContextVariableKey = self._determine_datasource_key(
-            datasource_name=datasource_name
-        )
+        datasource_key: Union[
+            DataContextVariableKey, GeCloudIdentifier
+        ] = self.store_backend.build_key(name=datasource_name)
```
anthonyburdi: Using the backend specific build_key method

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -149,7 +153,16 @@ def auth_headers(self) -> Dict[str, str]:
     def _get(self, key: Tuple[str, ...]) -> dict:
         ge_cloud_url = self.get_url_for_key(key=key)
         try:
-            response = requests.get(ge_cloud_url, headers=self.auth_headers)
+            params: dict
+            # if name is included in the key, add as a param
+            if len(key) > 2 and key[2]:
```
anthonyburdi, line 157: In the current architecture, the key here has already been converted into a tuple hence the checking for length. It would be better here to use an identifier type but saving that for a future refactor effort since it may not be trivial and touch several store types.

great_expectations/data_context/types/resource_identifiers.py
```diff
@@ -202,12 +202,16 @@ def from_object(cls, validation_result):
 
 class GeCloudIdentifier(DataContextKey):
     def __init__(
-        self, resource_type: "GeCloudRESTResource", ge_cloud_id: Optional[str] = None
+        self,
+        resource_type: "GeCloudRESTResource",
+        ge_cloud_id: Optional[str] = None,
+        resource_name: Optional[str] = None,
```
anthonyburdi, line 208: Added resource_name so that we can use the name as an identifier in addition to the id

tests/data_context/cloud_data_context/test_datasource_crud.py
```diff
@@ -250,7 +250,7 @@ def json(self):
         ),
     ],
 )
-def test_data_context_in_cloud_mode_add_datasource(
+def test_cloud_data_context_add_datasource(
```
anthonyburdi, line 253: Just fixing a mistake in naming this test in a prior PR

tests/data_context/store/test_datasource_store_cloud_backend.py
```diff
@@ -90,6 +98,61 @@ def json(self):
...
+        mock_has_key.return_value = True
+
+        datasource_store_ge_cloud_backend.retrieve_by_name(
+            datasource_name=datasource_name
+        )
+
+        mock_get.assert_called_with(
+            f"{ge_cloud_base_url}/organizations/{ge_cloud_organization_id}/datasources",
+            headers=request_headers,
+            params={"name": datasource_name},
```
anthonyburdi, line 153: This is the main check, that `name` is included as a url param

great_expectations/data_context/store/ge_cloud_store_backend.py
cdkini: ugh yeah I tried this refactor some time ago and it's tough - we use `key.to_tuple()` for all stores as implemented in `store.py`.

I don't fully remember why we do this but it results in us losing strong typing.

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -149,7 +153,16 @@ def auth_headers(self) -> Dict[str, str]:
...
+                params = {"name": key[2]}
+                ge_cloud_url = ge_cloud_url.rstrip("/")
+                response = requests.get(
+                    ge_cloud_url, headers=self.auth_headers, params=params
+                )
+            else:
+                response = requests.get(ge_cloud_url, headers=self.auth_headers)
```
cdkini:
```suggestion
            params: Optiona[dict] = None
            # if name is included in the key, add as a param
            if len(key) > 2 and key[2]:
                params = {"name": key[2]}
                ge_cloud_url = ge_cloud_url.rstrip("/")
                
            response = requests.get(
                ge_cloud_url, headers=self.auth_headers, params=params
            )
```
Not necessary but a small suggestion since `params` is by default a `NoneType`

great_expectations/data_context/store/store_backend.py
```diff
@@ -204,6 +204,9 @@ def is_ignored_key(self, key):
     def config(self) -> dict:
         raise NotImplementedError
 
+    def build_key(self, **kwargs) -> Any:
+        raise NotImplementedError
```
cdkini: Ooh I did this in my PR too - I like your impl better

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -343,7 +356,7 @@ def remove_key(self, key):
         data = {
             "data": {
                 "type": self.ge_cloud_resource_type,
-                "id": ge_cloud_id,
+                "id_": ge_cloud_id,
```
Kilo59, line 357: I thought we were serializing as `"id"`?

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -380,6 +393,12 @@ def config(self) -> dict:
 
     def build_key(self, id_: Optional[str] = None, **kwargs) -> GeCloudIdentifier:
         """Get the store backend specific implementation of the key, ignore irrelevant kwargs."""
+        if "name" in kwargs:
+            return GeCloudIdentifier(
+                resource_type=self.ge_cloud_resource_type,
+                ge_cloud_id=id_,
+                resource_name=kwargs["name"],
+            )
```
cdkini: Ah this is tough - I don't love kwargs but I see the need for them here. Any way we can avoid and use an actual arg?

great_expectations/data_context/types/resource_identifiers.py
```diff
@@ -225,14 +229,23 @@ def ge_cloud_id(self):
     def ge_cloud_id(self, value) -> None:
         self._ge_cloud_id = value
 
+    @property
+    def resource_name(self) -> str:
+        return self._resource_name
+
     def to_tuple(self):
-        return (self.resource_type, self.ge_cloud_id)
+        return (self.resource_type, self.ge_cloud_id, self.resource_name)
```
cdkini, line 237: So this is the only key type that is length 3? How can generalize some of these tuple conversions so we don't have to `if len x ... else ...`?

tests/data_context/store/test_datasource_store_cloud_backend.py
```diff
@@ -90,6 +98,61 @@ def json(self):
...
+
+        return MockResponse(
+            {
+                "data": {
+                    "id_": id_,
+                    "attributes": {"datasource_config": datasource_config},
+                }
+            },
+            200,
+        )
```
cdkini, line 136: I moved all of this to a `conftest` and used a `pytest.fixture`. Let's hash out what the best course of action is here.

tests/data_context/store/test_datasource_store_cloud_backend.py
```diff
@@ -90,6 +98,61 @@ def json(self):
...
+
+    with patch(
+        "requests.get", autospec=True, side_effect=mocked_response
+    ) as mock_get, patch(
+        "great_expectations.data_context.store.DatasourceStore.has_key", autospec=True
+    ) as mock_has_key:
```
cdkini, line 142: Can we assert call counts as well?

great_expectations/data_context/store/ge_cloud_store_backend.py
Kilo59: I was about to ask the same thing. Does this need `kwargs`? We aren't using them.
Can we use a `name` or `resource_name` keyword arg?
```python
def build_key(self, id_: Optional[str] = None, name: Optional[str] = None) -> GeCloudIdentifier:
```

Or even get rid of the conditional altogether 🤷

great_expectations/data_context/types/resource_identifiers.py
```diff
@@ -202,12 +202,16 @@ def from_object(cls, validation_result):
...
+        self,
+        resource_type: "GeCloudRESTResource",
+        ge_cloud_id: Optional[str] = None,
+        resource_name: Optional[str] = None,
     ) -> None:
         super().__init__()
 
         self._resource_type = resource_type
         self._ge_cloud_id = ge_cloud_id if ge_cloud_id is not None else ""
+        self._resource_name = resource_name if resource_name is not None else ""
```
Kilo59:
```suggestion
        self._resource_name = resource_name or ""
```
You can shorten these expressions to just `x or y`

great_expectations/data_context/types/resource_identifiers.py
Kilo59:
```python
return tuple(x for x in (self.resource_type, self.ge_cloud_id, self.resource_name) if x )
```

Could do this ☝️ 


Example 👇 
```python
tuple(x for x in ["a", "b", None, "d"] if x)

>>> ('a', 'b', 'd')
```

Kilo59: Could also make it return `namedtuple` with an optional `resource_name`
https://docs.python.org/3/library/typing.html#typing.NamedTuple

Although that makes me want to ask another question...
Should this (`GeCloudIdentifier`) just be a `dataclass`?

It already has builtin methods like `asdict`, `astuple`
https://docs.python.org/3/library/dataclasses.html#dataclasses.astuple

great_expectations/data_context/store/datasource_store.py
```diff
@@ -90,6 +92,19 @@ def deserialize(self, value: Union[dict, DatasourceConfig]) -> DatasourceConfig:
         else:
             return self._schema.loads(value)
 
+    def ge_cloud_response_json_to_object_dict(self, response_json: dict) -> dict:
+        """
+        This method takes full json response from GE cloud and outputs a dict appropriate for
+        deserialization into a GE object
```
Kilo59, line 98: Why is this an instance method?

It's not actually using anything from the instance or class, but even just making this a `staticmethod` doesn't totally feel right. 

I think maybe this is a symptom of our lack of a proper [Service Layer](https://martinfowler.com/eaaCatalog/serviceLayer.html) 🤔 ??
Sorry, I don't have any helpful advice for the second point ATM.

great_expectations/data_context/store/ge_cloud_store_backend.py
billdirks: s/Optiona/Optional/

tests/data_context/store/test_datasource_store_cloud_backend.py
```diff
@@ -90,6 +98,61 @@ def json(self):
         )
 
 
+@pytest.mark.cloud
+@pytest.mark.unit
```
billdirks, line 103: Thank you!

great_expectations/data_context/store/datasource_store.py
anthonyburdi: I agree - just keeping scope small in this PR and following the existing pattern.

great_expectations/data_context/store/ge_cloud_store_backend.py
anthonyburdi: I ran into some tough to debug issues with `instantiate_class_from_config` last week. I think we can serialize as id, but I think it will save us some heartache if we just use `id_` instead.

great_expectations/data_context/store/ge_cloud_store_backend.py
anthonyburdi: I originally had it this way, have to change some tests to expect `params=None` but good call.

great_expectations/data_context/store/ge_cloud_store_backend.py
anthonyburdi: Good call 👍

great_expectations/data_context/types/resource_identifiers.py
anthonyburdi: Totally agree this should not be represented as a tuple. I'd like to revisit that effort (refactoring our stores to not use tuples as keys) in a separate PR if that works for ya'll.

tests/data_context/store/test_datasource_store_cloud_backend.py
anthonyburdi: Done! Using `assert_called_once_with`

great_expectations/data_context/types/resource_identifiers.py
anthonyburdi: I hesitate to use this pattern since often I want to explicitly compare to None but in this case it works. Thanks! 🙇

great_expectations/data_context/store/datasource_store.py
cdkini: service layer makes a lot of sense here - I don't have a ton of experience with implementing one in a codebase of this size but could definitely use your knowledge here @Kilo59! Perhaps something for the backlog

great_expectations/data_context/types/resource_identifiers.py
cdkini: works for me!

tests/data_context/store/test_datasource_store_cloud_backend.py
```diff
@@ -78,15 +78,79 @@ def __init__(self, json_data: dict, status_code: int) -> None:
...
-        mock_get.assert_called_with(
+        mock_get.assert_called_once_with(
             f"{ge_cloud_base_url}/organizations/{ge_cloud_organization_id}/datasources/{id_}",
             headers=request_headers,
+            params=None,
+        )
+
+
+@pytest.mark.cloud
+@pytest.mark.unit
```
cdkini, line 103: Is this truly a unit test? I know we've mocked away a lot of the dependencies but it feels like there are still a decent bit of moving parts - moreso asking to better my understanding I'm okay if you and @billdirks agree here.

cdkini (approved):

great_expectations/data_context/types/resource_identifiers.py
Kilo59: I have noticed we use lots of explicit `is None` / `is not None` in our codebase. 
Is there a particular reason for that? 
As general rule I tend to use the `if SOMETHING:` / `if not SOMETHING:` as it delegates the tuthy check to `__bool__`. Unless there is a specific reason to treat `None` differently.

anthonyburdi:
Forced merge with failing cloud e2e tests so that this work can continue while those tests are debugged in parallel.

PR #5645, opened 2022-08-02
[MAINTENANCE] Install `dgtest` test runner utilizing Git URL in CI

cdkini:
- Instead of cloning the `dgtest` repo and using `pip install -e`, we can simply just use the Git URL
- The allows us to utilize the tool without having to clone it.

### Definition of Done

### DIFF
--- a/azure-pipelines-dependency-graph-testing.yml
+++ b/azure-pipelines-dependency-graph-testing.yml
@@ -246,9 +246,8 @@ stages:
 
           - script: |
               # Install dependencies
-              pip install pytest pytest-cov pytest-azurepipelines
-              git clone https://github.com/superconductive/dgtest.git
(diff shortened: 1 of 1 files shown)


Kilo59 (approved):

Kilo59:
Should fix the CI failures in #5525

azure-pipelines-dependency-graph-testing.yml
```diff
@@ -602,9 +597,7 @@ stages:
 
           - script: |
               # Install dependencies
-              pip install pytest
```
Kilo59, line 605: I think you accidentally removed `pytest` here.

cdkini: should be good since it is in `dgtest` `install_requires`

PR #5670, opened 2022-08-03
[MAINTENANCE] Only run Cloud E2E tests in primary pipeline

cdkini:
- E2E Cloud tests that make network requests should only be run on a schedule or release cut.
- Other Cloud tests (unit) should still be run during each CI cycle.

### Definition of Done

### DIFF
--- a/azure-pipelines-dependency-graph-testing.yml
+++ b/azure-pipelines-dependency-graph-testing.yml
@@ -358,7 +358,7 @@ stages:
           - script: |
               # These are tests that are specific to Great Expectations Cloud.
               # In order to ensure coverage, we run them during each CI/CD cycle.
-              pytest tests -m cloud --cloud
+              pytest tests --cloud -m "cloud and not integration"
 
             env:
               GE_CLOUD_BASE_URL: $(GE_CLOUD_BASE_URL)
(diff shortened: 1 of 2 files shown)


azure-pipelines-dependency-graph-testing.yml
```diff
@@ -358,7 +358,7 @@ stages:
           - script: |
               # These are tests that are specific to Great Expectations Cloud.
               # In order to ensure coverage, we run them during each CI/CD cycle.
-              pytest tests -m cloud --cloud
+              pytest tests --cloud -m "cloud and not integration"
```
NathanFarmer, line 361: Does it know operands like "and" and "not" or is this just the name of a pipeline?

azure-pipelines-dependency-graph-testing.yml
```diff
@@ -358,7 +358,7 @@ stages:
           - script: |
               # These are tests that are specific to Great Expectations Cloud.
```
NathanFarmer, line 359: Should we be explicit that these are unit tests since all I can see here is that they are not integration tests?

NathanFarmer (approved):
LGTM!

azure-pipelines-dependency-graph-testing.yml
cdkini: yeah `pytest` syntax is weird but per the docs, it allows for boolean operators like "and"/"or"

azure-pipelines-dependency-graph-testing.yml
cdkini: Hmm I wanted to cover the edge case where we label @cloud but not necessarily @unit or @integration - we haven't been super consistent so I opted for a less granular approach.

Happy to change just lmk!

PR #5676, opened 2022-08-04
[MAINTENANCE] expand flake8 coverage

Kilo59:
Formerly our `flake8` linter only checked and enforced linting issues in the package `great_expectations/core`.

This PR changes this behavior to lint everything that hasn't been explicitly excluded in the flake8 `setup.cfg` section.
To begin with, we are still excluding most of `great_expectations` from flake8 but expect followup PR's to whittle away at this.

Applied a series of low-risk fixes and/or warning suppression via `# noqa` to select modules.

Disabled some warning messages to due conflicts with our other tools (`black` & `isort`).

Added a `max-complexity` warning. The will raise warnings for classes. functions, whose [cyclomatic-complexity](https://dx42.github.io/gmetrics/metrics/CyclomaticComplexityMetric.html) reaches a certain threshold.

Updated `flake8` to the most recent version supported by `pre-commit`

## Summary of Changes
1. Change to flake8 linting to `opt-out` instead of only `great_expecations/core`
2. Initially exclude modules with high number of errors
3. Fix modules with low errors and low risk fixes
4. ignore rules that conflict with `black` or `isort`
5. add `max-complexity` rule
6. update flake8 to `3.9.2` - (latest is `5.0.4`)

## References
https://www.flake8rules.com/
https://dx42.github.io/gmetrics/metrics/CyclomaticComplexityMetric.html

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
     hooks:
       - id: black
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.8.3
+    rev: 3.9.2
     hooks:
       - id: flake8
   - repo: https://github.com/asottile/pyupgrade
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,6 @@ repos:
     rev: 3.8.3
     hooks:
       - id: flake8
-        files: ^(great_expectations/core)
   - repo: https://github.com/asottile/pyupgrade
     rev: v2.7.2
     hooks:
--- a/azure-pipelines-dependency-graph-testing.yml
+++ b/azure-pipelines-dependency-graph-testing.yml
@@ -99,9 +99,8 @@ stages:
           - script: |
               pip install isort[requirements]==5.10.1 flake8==3.9.2 black==22.3.0 pyupgrade==2.7.2 invoke==1.7.1
               EXIT_STATUS=0
-              isort . --check-only || EXIT_STATUS=$?
-              black --check . || EXIT_STATUS=$?
-              flake8 . || EXIT_STATUS=$?
+              invoke fmt --check || EXIT_STATUS=$?
+              invoke lint || EXIT_STATUS=$?
               pyupgrade --py3-plus || EXIT_STATUS=$?
               exit $EXIT_STATUS
 
--- a/azure-pipelines-dependency-graph-testing.yml
+++ b/azure-pipelines-dependency-graph-testing.yml
@@ -97,7 +97,7 @@ stages:
             displayName: 'Use Python 3.7'
 
           - script: |
-              pip install isort[requirements]==5.10.1 flake8==3.8.3 black==22.3.0 pyupgrade==2.7.2
+              pip install isort[requirements]==5.10.1 flake8==3.9.2 black==22.3.0 pyupgrade==2.7.2
               EXIT_STATUS=0
               isort . --check-only || EXIT_STATUS=$?
               black --check . || EXIT_STATUS=$?
--- a/azure-pipelines-dependency-graph-testing.yml
+++ b/azure-pipelines-dependency-graph-testing.yml
@@ -101,7 +101,7 @@ stages:
               EXIT_STATUS=0
               isort . --check-only || EXIT_STATUS=$?
               black --check . || EXIT_STATUS=$?
-              flake8 great_expectations/core || EXIT_STATUS=$?
+              flake8 . || EXIT_STATUS=$?
               pyupgrade --py3-plus || EXIT_STATUS=$?
               exit $EXIT_STATUS
 
--- a/azure-pipelines-dependency-graph-testing.yml
+++ b/azure-pipelines-dependency-graph-testing.yml
@@ -97,7 +97,7 @@ stages:
             displayName: 'Use Python 3.7'
 
           - script: |
-              pip install isort[requirements]==5.10.1 flake8==3.9.2 black==22.3.0 pyupgrade==2.7.2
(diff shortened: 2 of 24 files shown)


.pre-commit-config.yaml
```diff
@@ -21,10 +21,9 @@ repos:
     hooks:
       - id: black
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.8.3
+    rev: 3.9.2
     hooks:
       - id: flake8
-        files: ^(great_expectations/core)
```
Kilo59, line 27: Now that we are using flake8 in more of the code base I could forsee this pre-commit hook causing some un-welcome friction with developers.

IMO a little bit of friction to doing the wrong thing can be good, but we don't want them to be so annoyed that they uninstall these hooks.

Many of the other hooks are able to automatically fix the errors they find so this is less of an issue.
But with `flake8` it may just be warning about something that's more of a best practice when the developer is just trying to get something working and committed (and intends to cleanup later).

Options to deal with this might include removing this as a pre-commit hook, or maybe keeping this pre-commit hooks as is (only check `great_expecations/core`)

setup.cfg
```diff
@@ -12,7 +12,32 @@ tag_prefix =
...
+    great_expectations/render,
 per-file-ignores =
     */__init__.py: F401
+# E501 - line length (black)
+# E203 - whitespace before : (conflicts with black)
+# E402 - module level import not at top of file: (isort)
+extend-ignore = E501,E203,E402
+# https://github.com/pycqa/mccabe
+# TODO: reduce this 10
+max-complexity = 15
```
Kilo59: This may warrant some discussion.

Further explanation of this
https://dx42.github.io/gmetrics/metrics/CyclomaticComplexityMetric.html

.pre-commit-config.yaml
billdirks: We get some friction now for less build up of friction later. I suppose for a brand new contributors it could be an obstacle. I looked a little bit and didn't see a way to do this but if we could apply it to only a whitelisted group of users that would be another alternative. That is superconductive employees and regular community contributors should only commit `flake8` passing code.

great_expectations/data_asset/data_asset.py
```diff
@@ -905,7 +905,7 @@ def validate(
 
             # Warn if our version is different from the version in the configuration
             # TODO: Deprecate "great_expectations.__version__"
-            suite_ge_version = expectation_suite.meta.get(
+            suite_ge_version = expectation_suite.meta.get(  # noqa: F841
```
billdirks: I wonder if we can delete this line. It looks like the warn comment above is out of date. You don't need to address these now. This annotation (thanks for adding it!) makes this line suspicious.

setup.cfg
```diff
@@ -12,7 +12,32 @@ tag_prefix =
...
-max-line-length = 200
-exclude = tests/*,docs/*
+exclude = tests/*,
+    docs/*,
+    build/*,
+    assets/*,
+    contrib/*,
+    versioneer*,
+    examples/*,
+    # TODO: remove the items below and fix linting issues
```
billdirks, line 24: Can you file a ticket, tag it with `hygiene` and put it in the `Mario QOL` epic?

setup.cfg
billdirks: Agreed. Could you ask this in our slack channel?

billdirks (commented):
Thank you so much for this cleanup! LGTM. Before giving the official approve, could you followup in slack to make sure there isn't resistance to the complexity warning?

great_expectations/data_asset/data_asset.py
Kilo59: I agree, I don't think there are any run-time risks from deleting this.

My only worry was that it might make the TODO notes here harder to understand.
I'll go ahead and delete the line but leave the notes??

setup.cfg
Kilo59: https://superconductive.atlassian.net/browse/GREAT-1146

billdirks (approved):
LGTM, thanks!

azure-pipelines-dependency-graph-testing.yml
```diff
@@ -97,11 +97,10 @@ stages:
...
 
           - script: |
-              pip install isort[requirements]==5.10.1 flake8==3.8.3 black==22.3.0 pyupgrade==2.7.2
+              pip install isort[requirements]==5.10.1 flake8==3.9.2 black==22.3.0 pyupgrade==2.7.2 invoke==1.7.1
               EXIT_STATUS=0
-              isort . --check-only || EXIT_STATUS=$?
-              black --check . || EXIT_STATUS=$?
-              flake8 great_expectations/core || EXIT_STATUS=$?
+              invoke fmt --check || EXIT_STATUS=$?
+              invoke lint || EXIT_STATUS=$?
```
Kilo59, line 103: I meant to switch these over back when I originally implemented them but I missed these other pipeline files.

setup.cfg
```diff
@@ -12,7 +12,34 @@ tag_prefix =
...
-exclude = tests/*,docs/*
+exclude = .git,
+    build,
+    tests/*,
+    docs/*,
+    build/*,
+    assets/*,
+    contrib/*,
+    versioneer*,
+    examples/*,
```
Kilo59, line 23: @Shinnnyshinshin Is the examples directory something that would be good to lint against?

.pre-commit-config.yaml
Kilo59: To clarify, we also enforce flake8 in the CI, so no matter whether we have a flake8 pre-commit hook or not, no flake8 violating code will make it into `develop`, its just a question of do we also block local commits via hook.

PR #5682, opened 2022-08-04
[MAINTENANCE] Add usage stats and initial database docker tests to CI

billdirks:
### Definition of Done

Thank you for submitting!

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -121,6 +121,8 @@ stages:
     condition: or(eq(variables.isScheduled, true), eq(variables.isReleasePrep, true), eq(variables.isRelease, true), eq(variables.isManual, true))
     pool:
       vmImage: 'ubuntu-latest'
+    variables:
+      default_python_version: '3.8'
     jobs:
       - job: make_suffix
         steps:
@@ -150,8 +152,8 @@ stages:
               # We repull the develop branch. We could build the local branch that is already present but we should
               # prune it's depth since we don't need the git history.
               CMD="docker buildx build -f docker/Dockerfile.tests --tag greatexpectations/test:${DOCKER_TAG} "
-              CMD+="--target azure --build-arg PYTHON_VERSION=$(python.version) --build-arg SOURCE=github "
-              CMD+="--build-arg BRANCH=develop ."
+              CMD+="--target azure --build-arg PYTHON_VERSION=$(python.version) "
+              CMD+=" --build-arg SOURCE=github --build-arg BRANCH=develop ."
               echo ${CMD}
               eval ${CMD}
             displayName: 'Build python $(python.version) image'
@@ -169,7 +171,7 @@ stages:
               DOCKER_USER: $(DOCKER_USER)
               DOCKER_TOKEN: $(DOCKER_TOKEN)
 
-      - job: comprehensive_test_
+      - job: test_comprehensive_
         dependsOn: [make_suffix, build_push_]
         condition: succeeded()
         timeoutInMinutes: 120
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -183,30 +185,29 @@ stages:
...
-              PULL="docker pull greatexpectations/test:${DOCKER_TAG}"
-              echo ${PULL}
-              eval ${PULL}
               mkdir coverage
               PYTEST="docker run --network=host --mount type=bind,source=${PWD}/coverage,target=/coverage "
-              PYTEST+="-e GE_CLOUD_BASE_URL=$(GE_CLOUD_BASE_URL) -e GE_CLOUD_ACCESS_TOKEN=$(GE_CLOUD_ACCESS_TOKEN) "
-              PYTEST+="-e GE_CLOUD_ORGANIZATION_ID=$(GE_CLOUD_ORGANIZATION_ID) greatexpectations/test:${DOCKER_TAG} "
-              PYTEST+="/bin/bash -c \"pytest --postgresql --spark --cloud --napoleon-docstrings --no-docker-discovery "
+              PYTEST+="greatexpectations/test:${DOCKER_TAG} "
+              PYTEST+="/bin/bash -c \"pytest --postgresql --spark --trino "
```
billdirks: I have added trino to the comprehensive tests. This makes these tests take longer to run but gives us better coverage numbers. Some questions/comments:

1. I will willing to break out trino into it's own job if that is preferred. I would prefer to do that in a followup PR but can also do it in this one.
2. I am going to file a followup ticket to see if we can combine coverage numbers from individual jobs because right now the coverage numbers are not complete. This isn't supported by default but there may be a workaround:
https://github.com/microsoft/azure-pipelines-tasks/issues/10353#issuecomment-668819166

azure-pipelines.yml
```diff
@@ -221,6 +222,87 @@ stages:
               summaryFileLocation: '$(System.DefaultWorkingDirectory)/coverage/coverage.xml'
               reportDirectory: '$(System.DefaultWorkingDirectory)/coverage/htmlcov'
 
+      - job: test_cloud_
```
billdirks, line 217: I broke cloud out to mirror what we do elsewhere since it hits an external dependency and can be flaky.

azure-pipelines.yml
```diff
@@ -221,6 +222,87 @@ stages:
...
+              PYTEST+="pytest tests --cloud -m \"cloud and not unit\""
+              echo ${PYTEST}
+              eval ${PYTEST}
+            env:
+              GE_CLOUD_BASE_URL: $(GE_CLOUD_BASE_URL)
+              GE_CLOUD_ACCESS_TOKEN: $(GE_CLOUD_ACCESS_TOKEN)
+              GE_CLOUD_ORGANIZATION_ID: $(GE_CLOUD_ORGANIZATION_ID)
+            displayName: 'test_cloud_integration'
+
+      - job: test_mysql_
```
billdirks, line 247: I broke mysql out of the comprehensive test job since it required more setup. I think we should leave it separate for now but can fold it in if we want. I'd rather break things out into separate jobs and figure out how to combine coverage numbers though.

billdirks (commented):
After this PR is merged I will start deleting the non-docker tests. I will also file a followup ticket for handling the `cli` tests and the `mssql` tests.

azure-pipelines.yml
```diff
@@ -121,6 +121,8 @@ stages:
     condition: or(eq(variables.isScheduled, true), eq(variables.isReleasePrep, true), eq(variables.isRelease, true), eq(variables.isManual, true))
     pool:
       vmImage: 'ubuntu-latest'
+    variables:
+      default_python_version: '3.8'
```
Shinnnyshinshin: I'm happy with how this looks right now since docker tests are **supplementary** to the existing tests, rather than replacing them (yet). Or if it's not too big of a lift, would it be ok to include a `matrix` of python versions, like we have below?

azure-pipelines.yml
Shinnnyshinshin: Thanks for the research on this Bill, be default, I'm a little resistant to changes that (potentially) increase our test runtimes, especially in cases where there is a time-saving alternative which works just as well. It doesn't have to be in this PR, but would love to see `trino` broken out into its own parallizable step.

azure-pipelines.yml
billdirks: Done. This is in the stage definition and there was only 1 `job` that used it so I converted that to using a `matrix`.

azure-pipelines.yml
billdirks: Done. I moved `trino` to it's own `job` and removed it from here.

azure-pipelines.yml
Shinnnyshinshin: 👍🏼

azure-pipelines.yml
Shinnnyshinshin: 👍🏼

azure-pipelines.yml
Shinnnyshinshin: 👍🏼

Shinnnyshinshin (approved):
LGTM :) Thank you for the discussion earlier. It really helped me get a primer on all the great Docker work you've been doing

Kilo59 (approved):

PR #5693, opened 2022-08-05
[FEATURE] Introduce top-level `default_validation_id` in `CheckpointConfig`

cdkini:
- Update `CheckpointConfig` and `CheckpointConfigSchema` to accept `default_validation_id`, which is necessary for GX Cloud to be able to track the "default" validation (combination of `batch_request`, `expectation_suite_name`, and `action_list`).
- Updated API methods to accept this new field.
- Update to `SuiteValidationResult.meta` to include the checkpoint's `id` for easier tracking of resources in GX Cloud.
- Tests to cover new behavior.

### Definition of Done


great_expectations/checkpoint/checkpoint.py
```diff
@@ -179,15 +183,27 @@ def run(
...
+                validation_result: Optional[ExpectationSuiteValidationResult]
+                meta: ExpectationSuiteValidationResultMeta
+                for run_result in run_results.values():
+                    validation_result = run_result.get("validation_result")
+                    if validation_result:
+                        meta = validation_result.meta
+                        id_ = str(self.ge_cloud_id) if self.ge_cloud_id else None
+                        meta["checkpoint_id"] = id_
+
+                checkpoint_run_results.update(run_results)
```
cdkini, line 202: While this isn't strictly necessary, per discussion with the GX Cloud backend team, it would be helpful to have the checkpoint's `id` readily available from the `meta` of the SVR.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -520,6 +536,7 @@ def __init__(
         batches: Optional[List[dict]] = None,
         ge_cloud_id: Optional[UUID] = None,
         expectation_suite_ge_cloud_id: Optional[UUID] = None,
+        validation_id: Optional[str] = None,
```
cdkini: Don't love expand already verbose signatures but this is necessary due to the schema change - if we shift to a config-based API, we can minimize such issues.

great_expectations/core/expectation_validation_result.py
```diff
@@ -359,14 +369,26 @@ def make_expectation_validation_result(self, data, **kwargs):
...
+class ExpectationSuiteValidationResultMeta(TypedDict):
+    active_batch_definition: BatchDefinition
+    batch_markers: BatchMarkers
+    batch_spec: BatchSpec
+    checkpoint_id: Optional[str]
+    checkpoint_name: str
+    expectation_suite_name: str
+    great_expectations_version: str
+    run_id: RunIdentifier
+    validation_time: str
```
cdkini, line 381: Giving `meta` some semblance of structure (without enforcing it)

tests/core/test_serialization.py
```diff
@@ -519,7 +519,107 @@ def test_checkpoint_config_print(
                     },
                 ],
             },
-            id="config_without_id",
+            id="config_without_any_ids",
+        ),
```
cdkini, line 523: Additional tests to cover all combinations of top-level `validation_id` and nested `id` (within `Checkpoint.validations` list)

great_expectations/checkpoint/checkpoint.py
Kilo59: This is more of future refactor idea... but should we consider using a "parameter-object"?
Like a `Options` type, or even more simply, an `Options` `TypeDict`

Here are some relevant links, about how to refactor methods with lots of parameters. A `parameter-object` is one possible way to do it.

https://refactoring.guru/smells/long-parameter-list
https://refactoring.guru/introduce-parameter-object

cdkini: ah yeah the difficulty here is the fact that this is a very legacy bit of code that is at the forefront of the public API - I'm reluctant to make large scale changes here but perhaps we can implement something like you've noted in a non-breaking way.

Great website btw! I should read that more often.

Kilo59 (approved):

great_expectations/checkpoint/checkpoint.py
donaldheppner: I want to understand this change better -- why do we have 3 different IDs for the same config?

donaldheppner (changes requested):

great_expectations/checkpoint/checkpoint.py
cdkini: renamed `validation_id` to `default_validation_id` to better explain the purpose of the arg

donaldheppner (approved):

PR #5694, opened 2022-08-05
[MAINTENANCE]DEVREL-2289/Stale/Triage

AFineDayFor:
added trouble labels to exempt for cleanup.
one commit includes a disable.


.github/workflows/stale.yml
```diff
@@ -5,8 +5,8 @@ name: StaleBot
 
 on:
   workflow_dispatch:  # https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#defining-inputs-for-manually-triggered-workflows
-  schedule:
-    - cron: '30 1 * * *'
+  # schedule:
```
AFineDayFor, line 8: If we can determine we're not skipping any labels we needn't do this disable

kyleaton (approved):

PR #5664, opened 2022-08-03
[MAINTENANCE] Refactor store logic

anthonyburdi:
Changes proposed in this pull request:

>Note: some of these are being broken out into discrete PRs
- [x] - [minor cloud QOL changes - (`TypeDict`, `timeout`, better returns)](https://github.com/great-expectations/great_expectations/pull/5730)
- [x] - [round tripping `datasourceConfigSchema` object to deal with `id_` vs `id` issues](https://github.com/great-expectations/great_expectations/pull/5697)
- [ ] - `create` vs `set` changes
- [ ] `ge_cloud_store_backend` `GeCloudIdentifier` changes

PR #5697, opened 2022-08-05
[MAINTENANCE] Config/Schema round_tripping

Kilo59:
- `_dict_round_trip()` classmethod for Config (AbstractConfig) to help deserialization/serialization issues related to mismatched keys - (`id_` vs `id`).

Helper method for doing this kind of logic.
https://github.com/great-expectations/great_expectations/pull/5664/files#diff-1cf90172245fa16e9942295256f93f6dd99708bdcf0d2190a4678fafcfa07519

## Alternatives/Future enhancements.
1. ~class method on `Config` that acts like a factory and can ingest a dictionary~
2. Just update all our constructors to just take `id` keyword arguments, and mitigate against any `id()` issues.
3. Let's finally deal with our [primitive-obsession](https://refactoring.guru/smells/primitive-obsession)/[data-clumps](https://refactoring.guru/smells/data-clumps) and start passing around some kind of [`Identifier` object](https://refactoring.guru/introduce-parameter-object) that takes `id` & `name` instead of passing around `id` + `name` in countless constructor calls.

### Definition of Done


great_expectations/core/configuration.py
```diff
@@ -1,10 +1,24 @@
...
+    from typing_extensions import Protocol  # type: ignore[misc]
+
 from great_expectations.types import SerializableDictDot
 
 
+class Loadable(Protocol):
+    """TODO: also be dumpable?"""
+
+    def load(self, x) -> Mapping:
+        """Foo."""
```
Kilo59: Might be nice to not force docstring for Protocol methods. But I think tweaking our docstring checker to recognize this is probably more trouble than it's worth.

tasks.py
```diff
@@ -239,3 +240,34 @@ def mv_usage_stats_json(ctx):
     cmd = cmd.format(outfile)
     ctx.run(cmd)
     print(f"'{outfile}' copied to dbfs.")
+
+
+@invoke.task
+def docker(ctx, name="gx38", tag="dev", source=pathlib.Path.cwd(), build=False):
```
Kilo59: Going to strip this out of this PR.

tests/datasource/test_datasource_serialization.py
```diff
@@ -46,3 +46,8 @@ def test_datasource_config_is_serialized(
     loaded_data = datasourceConfigSchema.load(observed_dump)
     observed_load = DatasourceConfig(**loaded_data)
     assert observed_load.to_json_dict() == datasource_config.to_json_dict()
+
+    assert (
+        DatasourceConfig.dict_round_trip(datasourceConfigSchema, observed_dump)
+        == datasource_config.to_json_dict()
+    )
```
Kilo59: Not a proper test just using this to prove the functionality.
Will create dedicated tests if we decide this is the direction we want to go in.

great_expectations/core/configuration.py
```diff
@@ -16,3 +30,13 @@ def __init__(self, id_: Optional[str] = None, name: Optional[str] = None) -> Non
...
+
+    @classmethod
+    def dict_round_trip(cls, schema: Loadable, target: dict) -> dict:
+        """
+        Round trip a dictionary with a schema so that validation and serialization logic is applied.
+        Example: Swapping `id_` with `id`.
+        """
+        _loaded = schema.load(target)
+        _config = cls(**_loaded)
+        return _config.to_json_dict()
```
cdkini: I think the round trip we want is actually `load` and then `dump`? In speaking with Don, I think we want to avoid `to_json_dict` since it bypasses schema validation and the benefits of Marshmallow.

great_expectations/core/configuration.py
cdkini: Yup don't let that be a hindrance if not applicable - you can always just bump up the number of errors.

great_expectations/core/configuration.py
cdkini: Also instead of using this `Loadable` protocol, might it make more sense to have this be `Schema`? I understand the value in interfaces but I believe this value will always be a Marshmallow `Schema` right?

Kilo59: I think they are both valid.
In this case, the only thing that matters is that the object has a `.load()` method that returns a dict-like or `Mapping` object, it doesn't actually need to be a Marshmallow `Schema`.

Making it a protocol makes it more expressive about precisely what it needs to do.
This can especially be helpful if/when we decide to replace Marshmallow with something else. 

Having said all. We don't use protocols in this code base, and I don't think these are super compelling reasons to introduce this new pattern when `Schema` works just as well. 😄 
So I will update this to just use `Schema`.

Kilo59: Addressing the original question.
`.to_json_dict` is on the Config object
`.dump` is on the Schema object

`schema.load` should apply the schema validation.
https://marshmallow.readthedocs.io/en/stable/marshmallow.schema.html#marshmallow.schema.Schema.load

Kilo59:
A `Config.factory()`, or `Config.from_dict()` method might express the intent of this method better.

great_expectations/core/configuration.py
```diff
@@ -1,10 +1,26 @@
...
+class Loadable(Protocol):
+    """Object should have a `.load(x)` method that returns a dict-like `Mapping` object."""
+
+    def load(self, data) -> Mapping:
+        """Load an object and return a `dict` or dict-like `Mapping` object."""
+        ...
+
+
 class AbstractConfig(ABC, SerializableDictDot):
     """Abstract base class for Config objects. Sets the fields that must be included on a Config."""
```
Kilo59, line 26: Note: use `id` instead of `id_` in constructor and then remove shadowed `id` builtin

1. id field??
2. look for `id_`

great_expectations/core/configuration.py
```diff
@@ -15,6 +15,17 @@ def __init__(self, id_: Optional[str] = None, name: Optional[str] = None) -> Non
         self.name = name
         super().__init__()
 
+    @classmethod
+    def _dict_round_trip(cls, schema: Schema, target: dict) -> dict:
```
cdkini, line 19: Could we make this public? Especially since it is being invoked outside the class?

tasks.py
```diff
@@ -207,10 +207,11 @@ def mv_usage_stats_json(ctx):
         "name": "Docker image name.",
         "tag": "Docker image tag.",
         "build": "If True build the image, otherwise run it. Defaults to False.",
+        "detach": "Run container in background and print container ID. Defaults to False.",
```
cdkini: Do these changes and the `pyproject.toml` ones make sense in this PR? Can we separate them out?

tests/datasource/test_datasource_serialization.py
```diff
@@ -64,13 +64,35 @@
         ),
     ],
 )
-def test_datasource_config_is_serialized(
-    datasource_config: DatasourceConfig, expected_serialized_datasource_config: dict
-):
-    """Datasource Config should be serialized appropriately with/without optional params."""
-    observed_dump = datasourceConfigSchema.dump(datasource_config)
-    assert observed_dump == expected_serialized_datasource_config
+class TestDatasourceConfigSerialization:
```
cdkini, line 67: Could we also update the tests for checkpoints and RBP?

great_expectations/core/configuration.py
Kilo59: @anthonyburdi  and I talked about making this private because we hope to make this unnecessary with some follow-up work.

tasks.py
Kilo59: No they really don't make sense in this PR 😬 . Was trying to just get them in here to prevent blocking the develop merge queue with minor PRs.

I can drop them if need be.

anthonyburdi (commented):
Looks good pending resolution of @cdkini 's comments

great_expectations/core/configuration.py
cdkini: works for me

tasks.py
cdkini: I'd lean towards a separate PR if you don't mind 😄 I agree with the notion but CI should get through these invoke changes pretty quickly!

tests/datasource/test_datasource_serialization.py
Kilo59: Yes, admittedly I only added some very minimal tests for this since it isn't being used yet and we hope to excise it based on my follow-up work.
I will refactor these tests so that we can properly parameterize it with various configs.

tests/core/test_serialization.py
```diff
@@ -686,3 +690,82 @@ def test_checkpoint_config_and_nested_objects_are_serialized(
     assert checkpointConfigSchema.dump(observed_load) == checkpointConfigSchema.dump(
         checkpoint_config
     )
+
+
+@pytest.mark.unit
+@pytest.mark.parametrize(
+    "config,schema,expected_serialized_config",
```
Kilo59, line 701: Need to come up with some minimal parameters for `Checkpoint` and RBP.

tests/core/test_serialization.py
```diff
@@ -748,9 +748,35 @@ def test_checkpoint_config_and_nested_objects_are_serialized(
...
+            CheckpointConfig(  # Not an AbstractConfig
+                # id_="d3a14abd-d4cb-4343-806e-55b555b15c28",
+                name="my_checkpoint",
+            ),
+            checkpointConfigSchema,
+            {"id": "d3a14abd-d4cb-4343-806e-55b555b15c28", "name": "my_checkpoint"},
+            id="CheckpointConfig-minimal_with_name_and_id",
         ),
+        # pytest.param(
+        #     Checkpoint(  # needs `data_context`
```
Kilo59: Also not an `AbstractBaseConfig`

tests/core/test_serialization.py
```diff
@@ -686,3 +694,154 @@ def test_checkpoint_config_and_nested_objects_are_serialized(
...
+                                        "note": "Important remarks about estimation algorithm.",
+                                    },
+                                },
+                            },
+                        ],
+                    },
+                },
+                id_="d3a14abd-d4cb-4343-806e-55b555b15c28",
+            ),
+            ruleBasedProfilerConfigSchema,
```
Kilo59: This fails with this error message. During `schema.load()`
```
great_expectations.marshmallow__shade.exceptions.ValidationError: 
  {'config_version': ['Missing data for required field.'], 'rules': ['Missing data for required field.']}
```

Note: I do expect it to fail because the `expected_serialized_config` is wrong but that's an easy fix.

cdkini (approved):

tests/core/test_serialization.py
Kilo59: Deferring these tests until we actually use `_dict_round_trip()` with these configs (if we ever do).

PR #5703, opened 2022-08-06
[MAINTENANCE] invoke docker

Kilo59:
- minimal `invoke docker` task for building and/or running a local docker container

  1. Build an image locally and then run it with a local volume?
  2. Pull the latest, "official image" and run it without a local volume?
  3.  Something else???

---------

### Definition of Done

Thank you for submitting!


tasks.py
```diff
@@ -239,3 +239,34 @@ def mv_usage_stats_json(ctx):
     cmd = cmd.format(outfile)
     ctx.run(cmd)
     print(f"'{outfile}' copied to dbfs.")
+
+
+@invoke.task
+def docker(ctx, name="gx38", tag="dev", source=pathlib.Path.cwd(), build=False):
```
billdirks: Both the docker build and run commands both make the assumption that you are in the root repo directory so I think we can remove `source` as a parameter for now and set it to the root directly. I think it might also be confusing to have it as a parameter currently since we have the docker build-arg source.

Doing that and adding a docstring already adds value to what we have now. 

I could see us wanting to parameterize the command and have it default to `bash`.

Thanks for getting this started! I'm happy to pick this up and make the edits I suggest if you want to step back from this to focus on your thin-saas work. If you want to continue with this, that is also good with me.

Thanks!

tasks.py
```diff
@@ -240,3 +242,49 @@ def mv_usage_stats_json(ctx):
...
+def docker(ctx, name="gx38local", tag="latest", build=False, cmd="bash"):
+    """
+    Build or run gx docker image.
+    """
+    filedir = os.path.realpath(os.path.dirname(os.path.realpath(__file__)))
+    curdir = os.path.realpath(os.getcwd())
+    if filedir != curdir:
+        sys.exit(
+            "The docker task must be invoked from the same directory as the task.py file at the top of the repo."
+        )
```
```suggestion
        raise invoke.Exit(
            "The docker task must be invoked from the same directory as the task.py file at the top of the repo.", code=1
        )
```
https://docs.pyinvoke.org/en/stable/api/exceptions.html#invoke.exceptions.Exit

Kilo59:
✔️  Looks good.

billdirks (approved):

PR #5711, opened 2022-08-08
Add vectorized is_between for common numpy dtypes

jcampbell:
Use vectorized comparisons for native numpy dtypes.


great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
```diff
@@ -73,6 +75,30 @@ def _pandas(
         if min_value is not None and max_value is not None and min_value > max_value:
             raise ValueError("min_value cannot be greater than max_value")
 
+        # Use a vectorized approach for native numpy dtypes
+        if column.dtype in [int, float] and allow_cross_type_comparisons is not True:
```
billdirks: Should `allow_cross_type_comparisons` always be a boolean? I notice the default value is `None`. Would making the default value False and changing this to `allow_cross_type_comparisons is False` possible?

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
```diff
@@ -73,6 +75,30 @@ def _pandas(
...
+            )
+        elif (
+            column.dtype == np.dtype("datetime64[ns]")
+            and allow_cross_type_comparisons is not True
+        ):
+            if min_value is not None and not isinstance(max_value, datetime.datetime):
+                try:
+                    min_value = parse(min_value)
+                except TypeError:
+                    pass
```
billdirks: If `min_value` is not parsable, does that mean the function was called with invalid arguments and we shouldn't be going down this codepath? I would prefer to die here if this is a bug in the caller. We should at least log.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
```diff
@@ -73,6 +75,30 @@ def _pandas(
...
+        # Use a vectorized approach for native numpy dtypes
+        if column.dtype in [int, float] and allow_cross_type_comparisons is not True:
+            return cls._pandas_vectorized(
+                temp_column, min_value, max_value, strict_min, strict_max
+            )
+        elif (
+            column.dtype == np.dtype("datetime64[ns]")
+            and allow_cross_type_comparisons is not True
+        ):
+            if min_value is not None and not isinstance(max_value, datetime.datetime):
```
billdirks: It looks like you should be passing `min_value` instead of `max_value` to the isinstance call?

billdirks: This `if` block seems a little out of place since we would only enter it if the caller passed in (or took the default value) `parse_strings_as_datetimes=False` but we are doing the conversion anyway. If they have set this parameter to `True` this conversion should have happened in the block above. I would prefer we didn't try to convert the date here.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
```diff
@@ -169,6 +195,29 @@ def is_between(val):
 
         return temp_column.map(is_between)
 
+    @classmethod
+    def _pandas_vectorized(cls, column, min_value, max_value, strict_min, strict_max):
```
billdirks: Is it possible to add type annotations here?

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
```diff
@@ -169,6 +195,29 @@ def is_between(val):
...
+        if min_value is None:
```
billdirks, line 220: It appears we need a guard if both `min_value` and `max_value` are `None`.

billdirks (changes requested):
Thanks for this PR! I've left some comments and questions inline.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
jcampbell: That makes sense to me. I believe the reason the default is None is to be able to disambiguate between the case where it was not provided by the user at all (so isn't explicitly stored in the configuration).

That really makes me think `default` here was being used imprecisely to mix those cases. I think going back to fully disambiguate should be a different PR, but we can check for None and set to False here, and add a comment. LMK if you think that's a good solution.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
jcampbell: Definitely a good catch on the `min_value` vs `max_value`.

Re: `parse_strings_as_datetimes`: this is another thorn, but I think there are two things going on here.

*Even if* `parse_strings_as_datetimes` is False (and now it should *always* be False as we remove that functionality), today, if you pass in a datetime as the *parameter* value (min_value), it will be stringified before this block of code to prepare for a json-compatible expectation configuration.

I think we know the *correct* fix is to make the expectation configuration parameters type aware.

Since they're not, what's actually happening here is that we're saying that if the *column* is a datetime, then we want the thing we're comparing it to to be a datetime as well, so we parse the parameter (`min_value` or `max_value`) *back* into a datetime.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
jcampbell: Good point; I'll just remove.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
jcampbell: +1

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
jcampbell: 👀

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
jcampbell: Updated.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
jcampbell: Looks like that guard is at the very beginning, raising a ValueError on the entire metric.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
billdirks: Sounds good to me.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
billdirks: I see. I believe this conversion code is basically identical to the above. Can we put it in it's own function and call it in both?
It might be useful to add a comment that this may have been stringified and needs to be converted back. Maybe it's not necessary if that is common throughout the code but for me, who is relatively unfamiliar with this part of the codebase, it isn't clear to me.  Typing the function definition would help with this but maybe that would increase the scope of the PR too much.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
billdirks: I see. This that is nonlocal to this call site and an assumption it makes, could you add a comment?

jcampbell: Happy to do a comment, but on reflection it seems like this is in fact better just checked again, since we've got similar logic again. I'll just add the guard.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
jcampbell: I propose we just add the comment at this point, since the `parse` logic above would be both slightly different (because of the try/catch) and not expected to survive a future stronger-typed parameter.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
jcampbell: done.

great_expectations/expectations/metrics/column_map_metrics/column_values_between.py
jcampbell: done.

billdirks (approved):
LGTM!

Thank you for the context adding comments!

PR #5715, opened 2022-08-08
[MAINTENANCE] Leverage `DataContextVariables` in `DataContext` hierarchy to automatically determine how to persist changes

cdkini:
- Instead of defining a `_save_project_config` method on each context, we can just leverage the given instance of `DataContextVariables` (which already knows how to persist values based on environment).
- `BaseDataContext` is an edge case as this method always writes to disk (even if working in a Cloud-backend env).

### Definition of Done

### DIFF
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -204,7 +204,7 @@ class AbstractDataContext(ABC):
 
     def _save_project_config(self) -> None:
         """
-        Each DataContext will define how its project_config will be saved.
+        Each DataContext will define how its project_config will be saved through its internal 'variables'.
             - FileDataContext : Filesystem.
             - CloudDataContext : Cloud endpoint
             - Ephemeral : not saved, and logging message outputted
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -202,7 +202,6 @@ class AbstractDataContext(ABC):
     def _init_variables(self) -> DataContextVariables:
         raise NotImplementedError
 
-    @abstractmethod
     def _save_project_config(self) -> None:
         """
         Each DataContext will define how its project_config will be saved.
@@ -210,7 +209,7 @@ class AbstractDataContext(ABC):
             - CloudDataContext : Cloud endpoint
             - Ephemeral : not saved, and logging message outputted
         """
-        raise NotImplementedError
+        self.variables.save_config()
 
     @abstractmethod
     def save_expectation_suite(
--- a/great_expectations/data_context/store/inline_store_backend.py
+++ b/great_expectations/data_context/store/inline_store_backend.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import Any, List, Optional, Tuple
 
 from great_expectations.core.data_context_key import DataContextVariableKey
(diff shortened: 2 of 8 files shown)


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -202,15 +202,14 @@ def __init__(self, runtime_environment: Optional[dict] = None) -> None:
...
     def _save_project_config(self) -> None:
         """
-        Each DataContext will define how its project_config will be saved.
+        Each DataContext will define how its project_config will be saved through its internal 'variables'.
             - FileDataContext : Filesystem.
             - CloudDataContext : Cloud endpoint
             - Ephemeral : not saved, and logging message outputted
         """
-        raise NotImplementedError
+        self.variables.save_config()
```
cdkini, line 212: Leverage `variables` abstraction and remove the need for abstract method.

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -183,8 +184,14 @@ def _has_key(self, key: Tuple[str, ...]) -> bool:
...
-        # NOTE: <DataContextRefactor> This responsibility will be moved into DataContext Variables object
-        self._data_context._save_project_config()
+        context = self._data_context
+        config_filepath = os.path.join(context.root_directory, context.GE_YML)
+
+        try:
+            with open(config_filepath, "w") as outfile:
+                context.config.to_yaml(outfile)
+        except PermissionError as e:
+            logger.warning(f"Could not save project config to disk: {e}")
```
cdkini: Move writing YAML to disk to this class - keeping this as-is would lead to an infinite loop as `_save_project_config` references the `variables.save_changes` method

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -183,8 +184,14 @@ def _has_key(self, key: Tuple[str, ...]) -> bool:
         return resource_type in self._data_context.config
 
     def _save_changes(self) -> None:
-        # NOTE: <DataContextRefactor> This responsibility will be moved into DataContext Variables object
-        self._data_context._save_project_config()
+        context = self._data_context
+        config_filepath = os.path.join(context.root_directory, context.GE_YML)
```
Kilo59: Can we use `pathlib` for this?
```suggestion
        config_filepath = pathlib.Path(context.root_directory) / context.GE_YML
```

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59: Will hiding this `variables.save_config()` call inside this private method on this ABC make this `.save_config` call harder to follow?

Kilo59 (approved):

great_expectations/data_context/store/inline_store_backend.py
billdirks: Is it ok that this save fails? That is, should we reraise? If not, could you add a comment?

Shinnnyshinshin (approved):
Love it. Thanks for getting this together so quickly @cdkini

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: perhaps but until we can get rid of the impl of this method on `BaseDataContext`, we need it to be its own private method.

great_expectations/data_context/store/inline_store_backend.py
cdkini: done!

great_expectations/data_context/store/inline_store_backend.py
cdkini: this was implemented as a bit of a hotfix due to calls accidentally writing to disk in environments where it wasn't allowed. I'll add a comment!

PR #5726, opened 2022-08-09
[FEATURE] Inline `ExpectationSuite` Rendering

NathanFarmer:
- GREAT-911/GREAT-1050
- Allow `include_rendered_content` boolean to be passed to `context.save_expectation_suite` to render and save the prescriptive `rendered_content` for each `ExpectationConfiguration` in the `ExpectationSuite`.
- Add `include_rendered_content` to `DataContext.variables`.
- Add tests for rendering each Cloud-backed context type with `context.save_expectation_suite` and `context.get_validator`.

### Definition of Done


great_expectations/data_context/types/base.py
```diff
@@ -1049,6 +1049,11 @@ def validate_schema(self, data, **kwargs):
     # noinspection PyUnusedLocal
     @post_load
     def make_datasource_config(self, data, **kwargs):
+        # Add names to data connectors
+        for data_connector_name, data_connector_config in data.get(
+            "data_connectors", {}
+        ).items():
+            data_connector_config["name"] = data_connector_name
```
NathanFarmer: Unrelated bugfix that is being handled by: https://github.com/great-expectations/great_expectations/pull/5763

great_expectations/core/expectation_suite.py
```diff
@@ -937,6 +939,33 @@ def _get_expectations_by_domain_using_accessor_method(
...
+        Renders content using the atomic prescriptive renderer for each expectation configuration associated with
+           this ExpectationSuite to ExpectationConfiguration.rendered_content.
+        """
+        for expectation_configuration in self.expectations:
+            inline_renderer_config: Dict[str, Union[str, ExpectationConfiguration]] = {
+                "class_name": "InlineRenderer",
+                "render_object": expectation_configuration,
+            }
+            module_name = "great_expectations.render.renderer.inline_renderer"
+            inline_renderer = instantiate_class_from_config(
```
NathanFarmer, line 957: Typically I would avoid this method, as we prefer to use typed objects, but this solves circular import issues without decoupling a lot of packages.

cdkini: Is there value in using a `TypedDict`?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -217,6 +217,7 @@ def save_expectation_suite(
         expectation_suite: ExpectationSuite,
         expectation_suite_name: Optional[str] = None,
         overwrite_existing: bool = True,
+        include_rendered_content: bool = False,
```
cdkini: Are we okay with expanding the public signature here? Just want to weigh the pros/cons since it introduces another possible breaking point.

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -285,6 +288,8 @@ def save_expectation_suite(
                 f"If you would like to overwrite this expectation_suite, set overwrite_existing=True."
             )
         self._evaluation_parameter_dependencies_compiled = False
+        if include_rendered_content:
+            expectation_suite.render()
```
cdkini, line 293: Is there value in making the method return a new suite with the updates? Or return itself? I'm okay with in-place too but just thinking of what is easier to reason about.

great_expectations/render/renderer/inline_renderer.py
```diff
@@ -18,13 +19,20 @@
...
         self,
-        render_object: ExpectationValidationResult,
+        render_object: Union[ExpectationConfiguration, ExpectationValidationResult],
     ) -> None:
         super().__init__()
 
-        self._render_object = render_object
+        if isinstance(render_object, ExpectationConfiguration) or isinstance(
+            render_object, ExpectationValidationResult
+        ):
```
cdkini: Nit - `isinstance(render_object, (ExpectationConfiguration, ExpectationValidationResult))`

great_expectations/render/renderer/inline_renderer.py
```diff
@@ -18,13 +19,20 @@
...
+            self._render_object = render_object
+        else:
+            raise InvalidRenderedContentError(
+                f"InlineRenderer can only be used with an ExpectationConfiguration or ExpectationValidationResult, but {type(render_object)} was used."
+            )
 
-    def get_atomic_rendered_content_for_object(
```
cdkini, line 27: Is this a breaking change we should be concerned about?

great_expectations/render/renderer/inline_renderer.py
```diff
@@ -134,15 +141,32 @@ def _get_renderer_atomic_rendered_content(
 
         return renderer_rendered_content
 
-    def render(self) -> Tuple[List[RenderedAtomicContent], List[RenderedAtomicContent]]:
-        """Gets RenderedAtomicContent for a given ExpectationConfiguration or ExpectationValidationResult.
+    def get_expectation_validation_result_rendered_content(
+        self,
+    ) -> Tuple[List[RenderedAtomicContent], List[RenderedAtomicContent]]:
```
cdkini: Value in a `namedtuple`?

tests/checkpoint/test_simple_checkpoint.py
```diff
@@ -3462,3 +3466,47 @@ def test_simple_checkpoint_does_not_pass_dataframes_via_validations_into_checkpo
...
     ):
         context.add_checkpoint(**checkpoint_config)
+
+
+@pytest.mark.unit
+def test_simple_checkpoint_result_validations_include_rendered_content(
+    titanic_pandas_data_context_stats_enabled_and_expectation_suite_with_one_expectation,
+    sa,
+):
+    context: DataContext = titanic_pandas_data_context_stats_enabled_and_expectation_suite_with_one_expectation
```
cdkini: Is this really a unit test? The context in question uses a bunch of file I/O as part of setup right?

tests/checkpoint/test_simple_checkpoint.py
```diff
@@ -3462,3 +3466,47 @@ def test_simple_checkpoint_does_not_pass_dataframes_via_validations_into_checkpo
...
+
+    validation_result_identifier: ValidationResultIdentifier = (
+        result.list_validation_result_identifiers()[0]
+    )
+    assert isinstance(
+        result.run_results[validation_result_identifier]["validation_result"]
+        .results[0]
+        .rendered_content[0],
+        RenderedAtomicContent,
+    )
```
cdkini: Nitpick - Could we separate out the `.results[0].rendered_contents[0]` parsing from the `isinstance` check?

tests/conftest.py
```diff
@@ -2505,15 +2505,20 @@ def cloud_data_context_with_datasource_pandas_engine(
...
     )
-    context.add_datasource(
-        "my_datasource",
-        **config,
-    )
+    with mock.patch(
+        "great_expectations.data_context.store.ge_cloud_store_backend.GeCloudStoreBackend.list_keys"
+    ), mock.patch(
+        "great_expectations.data_context.store.ge_cloud_store_backend.GeCloudStoreBackend._set"
+    ):
```
cdkini, line 2558: Any value in `with mock.patch(...) as mock_xyz` and then asserting that `mock_xyz` has been called or used?

tests/conftest.py
```diff
@@ -2530,10 +2535,73 @@ def cloud_data_context_with_datasource_sqlalchemy_engine(
...
         """,
     )
-    context.add_datasource(
-        "my_datasource",
-        **config,
+    with mock.patch(
+        "great_expectations.data_context.store.ge_cloud_store_backend.GeCloudStoreBackend.list_keys"
+    ), mock.patch(
+        "great_expectations.data_context.store.ge_cloud_store_backend.GeCloudStoreBackend._set"
+    ):
```
cdkini, line 2588: Same note here

tests/data_context/cloud_data_context/test_include_rendered_content.py
```diff
@@ -0,0 +1,184 @@
...
+        ),
+        pytest.param("empty_data_context_in_cloud_mode", DataContext, id="DataContext"),
+        pytest.param(
+            "empty_cloud_data_context", CloudDataContext, id="CloudDataContext"
+        ),
+    ],
+)
+def test_cloud_backed_data_context_save_expectation_suite_include_rendered_content(
+    data_context_fixture_name: str,
+    data_context_type: Type[AbstractDataContext],
```
cdkini: Do we need this value at all? I think we could simplify the fixture if we wanted.

tests/data_context/cloud_data_context/test_include_rendered_content.py
```diff
@@ -0,0 +1,184 @@
...
+        "great_expectations.data_context.store.ge_cloud_store_backend.GeCloudStoreBackend.list_keys"
+    ), mock.patch(
+        "great_expectations.data_context.store.ge_cloud_store_backend.GeCloudStoreBackend._update"
+    ) as mock_update:
+        context.save_expectation_suite(
+            expectation_suite,
+        )
+
+        # remove dynamic great_expectations version
+        mock_update.call_args[1]["value"].pop("meta")
```
cdkini, line 68: Why 1 instead of 0?

tests/data_context/cloud_data_context/test_include_rendered_content.py
```diff
@@ -0,0 +1,184 @@
...
+
+        mock_update.assert_called_with(
+            ge_cloud_id="None",
```
cdkini, line 71: Why "None" instead of `None`?

tests/data_context/cloud_data_context/test_include_rendered_content.py
```diff
@@ -0,0 +1,184 @@
...
+        pytest.param(
+            "cloud_data_context_in_cloud_mode_with_datasource_pandas_engine",
+            DataContext,
+            id="DataContext",
+        ),
+        pytest.param(
+            "cloud_data_context_with_datasource_pandas_engine",
+            CloudDataContext,
+            id="CloudDataContext",
+        ),
```
cdkini: Same note about `data_context_type` here

tests/data_context/test_data_context.py
```diff
@@ -279,6 +280,36 @@ def test_save_expectation_suite(data_context_parameterized_expectation_suite):
     assert expectation_suite.expectations == expectation_suite_saved.expectations
 
 
+@pytest.mark.unit
+def test_save_expectation_suite_include_rendered_content(
+    data_context_parameterized_expectation_suite,
+):
```
cdkini: Is this truly a unit test? Not 100% sure what moving pieces are here but just trying to narrow our definitions.

tests/render/test_inline_renderer.py
```diff
@@ -9,11 +9,29 @@
...
 from great_expectations.data_context import DataContext
+from great_expectations.render.exceptions import InvalidRenderedContentError
 from great_expectations.render.renderer.inline_renderer import InlineRenderer
 from great_expectations.render.types import RenderedAtomicContent
 from great_expectations.validator.validator import Validator
 
 
+@pytest.mark.unit
+def test_inline_renderer_error_message(
+    alice_columnar_table_single_batch_context: DataContext,
```
cdkini: This fixture takes quite some time to put together due to some I/O - could we use a simpler mock context to keep this a unit test?

tests/render/test_inline_renderer.py
```diff
@@ -9,11 +9,29 @@
...
+    )
+    with pytest.raises(InvalidRenderedContentError) as e:
+        InlineRenderer(render_object=expectation_suite)  # type: ignore
+    assert (
+        str(e.value)
+        == "InlineRenderer can only be used with an ExpectationConfiguration or ExpectationValidationResult, but <class 'great_expectations.core.expectation_suite.ExpectationSuite'> was used."
+    )
+
+
+@pytest.mark.unit
```
cdkini: Same note about `unit` here - this feels quite complex for a unit test and uses fixtures like `alice_context` which are quite cumbersome and have moving parts.

cdkini (commented):
Left some general notes but overall, it looks really good to me!

great_expectations/core/expectation_suite.py
Kilo59: https://mypy.readthedocs.io/en/stable/runtime_troubles.html?highlight=if%20type_checking#typing-type-checking

Is this something that can be resolved with a `if TYPE_CHECKING:` conditional?
Happy to sync on this.

great_expectations/data_context/data_context/cloud_data_context.py
NathanFarmer: This just populates a `rendered_content` attribute that starts as `None`. I don't see why we should create a copy and it already updates the pointer. Also, I like how pythonic these two lines are.

great_expectations/render/renderer/inline_renderer.py
NathanFarmer: ✅

great_expectations/render/renderer/inline_renderer.py
NathanFarmer: There is no reason for any client to access this, which is why I took it private before we start actually leveraging the renderer.

tests/checkpoint/test_simple_checkpoint.py
NathanFarmer: This was a mistake. ✅

tests/checkpoint/test_simple_checkpoint.py
NathanFarmer: Sure, but why?

tests/conftest.py
NathanFarmer: Not really because the behavior of the datasource store should be tested here instead of fixtures: https://github.com/great-expectations/great_expectations/blob/develop/tests/data_context/cloud_data_context/test_datasource_crud.py

tests/data_context/cloud_data_context/test_include_rendered_content.py
NathanFarmer: Good catch ✅

tests/data_context/cloud_data_context/test_include_rendered_content.py
NathanFarmer: Because the parameters come from method `_update`, not `list_keys`? That's where the values are populated.

tests/data_context/cloud_data_context/test_include_rendered_content.py
NathanFarmer: Because of this line: https://github.com/great-expectations/great_expectations/blob/1cbf5147c87d045a93eab21ebeda460806580395/great_expectations/data_context/data_context/cloud_data_context.py#L277

tests/data_context/cloud_data_context/test_include_rendered_content.py
NathanFarmer: ✅

tests/data_context/cloud_data_context/test_include_rendered_content.py
Kilo59: I can't imagine that's what we really want?
I probably don't understand all the implications here but...
If we absolutely need the cloud_id to be a string then IMO we should use an empty string, which will evaluate as falsy.

```python
  key = GeCloudIdentifier(
            resource_type=GeCloudRESTResource.EXPECTATION_SUITE,
            ge_cloud_id=ge_cloud_id or "")
```
Although I might question why it needs to be a string.

https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/

great_expectations/data_context/data_context/abstract_data_context.py
NathanFarmer: Per internal conversations, we should leave this as-is for now.

great_expectations/core/expectation_suite.py
NathanFarmer: The `TypedDict` would need to be imported from somewhere (`inline_renderer.py`) which will cause circular imports.

The `TYPE_CHECKING` solution only works for annotations right? The problems seem deeper than that.

great_expectations/render/renderer/inline_renderer.py
NathanFarmer: It does improve readability ✅

tests/data_context/cloud_data_context/test_include_rendered_content.py
NathanFarmer: It tries to convert `None` to an empty string and then that fails as well.
```
class GeCloudIdentifier(DataContextKey):
    def __init__(
        self,
        resource_type: "GeCloudRESTResource",
        resource_name: Optional[str] = None,
    ) -> None:
        super().__init__()

        self._resource_type = resource_type
        self._ge_cloud_id = ge_cloud_id or ""
        self._resource_name = resource_name or ""
```

tests/data_context/test_data_context.py
NathanFarmer: I agree, this marker was a mistake.

NathanFarmer: ✅

tests/render/test_inline_renderer.py
NathanFarmer: I agree, this marker was a mistake. ✅

great_expectations/core/expectation_suite.py
Kilo59: @NathanFarmer Yes `if TYPE_CHEKCING:` is for annotations. Which the `TypeDict` can be. You don't need to actually instantiate the TypeDict class.
```python
if TYPE_CHECKING:
  from inline_render import MyTypeDict


...

my_type_dict: 'MyTypeDict' = {"foo": "bar"}

```

Kilo59: Sorry for this additional ask but could we go ahead and remove `core/expectation_suite.py` from the `mypy` exclude list?
https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L45
According to this it only has `10` outstanding typing errors.
I'm also happy to submit a small PR against your branch with these fixes.

tests/render/test_inline_renderer.py
NathanFarmer: I haven't been able to locate a mock context or expectation suite. I think we should table this until we can mock more of our core objects. I have labeled this as integration.

great_expectations/core/expectation_suite.py
NathanFarmer: This is taken care of. Thanks @Kilo59!

pyproject.toml
```diff
@@ -42,7 +42,6 @@ exclude = [
     'core/config_peer\.py', # 1
     'core/evaluation_parameters\.py', # 6
     'core/expectation_configuration\.py', # 21
-    'core/expectation_suite\.py', # 10
```
Kilo59, line 45: 👏

Kilo59: 1 last typing issue. LSP violations strike again 😄 

-----


great_expectations/data_context/data_context/file_data_context.py:84: error: Argument 4 of "save_expectation_suite" is incompatible with supertype "AbstractDataContext"; supertype defines the argument type as "Optional[bool]"  [override]

great_expectations/data_context/data_context/file_data_context.py:84: note: This violates the Liskov substitution principle

great_expectations/data_context/data_context/file_data_context.py:84: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides

NathanFarmer: Hi @Kilo59 this is still WIP, I will update you when it is ready for review.

tests/data_context/cloud_data_context/test_include_rendered_content.py
```diff
@@ -0,0 +1,170 @@
...
+                    }
+                ],
+                "ge_cloud_id": None,
+                "data_asset_type": None,
+                "expectation_suite_name": "test_suite",
+            },
+        )
+
+
+# TODO: ACB - Enable this test after merging fixes in PRs 5778 and 5763
```
NathanFarmer, line 106: This test should be uncommented in either: https://github.com/great-expectations/great_expectations/pull/5778 or https://github.com/great-expectations/great_expectations/pull/5763

NathanFarmer: @anthonyburdi ^

tests/data_context/cloud_data_context/test_include_rendered_content.py
NathanFarmer: I created GREAT-1195 to track this, as it is out of scope for this PR.

great_expectations/core/expectation_suite.py
```diff
@@ -937,6 +943,33 @@ def _get_expectations_by_domain_using_accessor_method(
...
+    def render(self) -> None:
+        """
+        Renders content using the atomic prescriptive renderer for each expectation configuration associated with
+           this ExpectationSuite to ExpectationConfiguration.rendered_content.
+        """
+        for expectation_configuration in self.expectations:
+            inline_renderer_config: "InlineRendererConfig" = {
+                "class_name": "InlineRenderer",
+                "render_object": expectation_configuration,
+            }
```
cdkini, line 955: This is a dictionary right? Should we have this type hint?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -217,8 +218,8 @@ def save_expectation_suite(
         expectation_suite: ExpectationSuite,
         expectation_suite_name: Optional[str] = None,
         overwrite_existing: bool = True,
-        ge_cloud_id: Optional[str] = None,
-        **kwargs: Optional[dict],
+        include_rendered_content: Optional[bool] = None,
```
cdkini, line 221: Is it possible to just make this `bool`? Why do we need the optional?

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -1395,24 +1394,38 @@ def save_expectation_suite(
...
         """
+        if include_rendered_content is None:
+            if (
+                self.include_rendered_content.expectation_suite is True
+                or self.include_rendered_content.globally is True
+            ):
+                include_rendered_content = True
+            else:
+                include_rendered_content = False
+
```
cdkini: Should we make this a private helper? Seems like we use this logic a decent bit

great_expectations/data_context/data_context_variables.py
```diff
@@ -271,6 +273,19 @@ def progress_bars(self, progress_bars: ProgressBarsConfig) -> None:
...
+        return self._get(DataContextVariableSchema.INCLUDE_RENDERED_CONTENT)
+
+    @include_rendered_content.setter
+    def include_rendered_content(
+        self, include_rendered_content: IncludeRenderedContentConfig
+    ) -> None:
+        self._set(
+            DataContextVariableSchema.INCLUDE_RENDERED_CONTENT,
+            include_rendered_content,
+        )
```
cdkini, line 287: 🙇🏽

great_expectations/data_context/store/data_context_store.py
```diff
@@ -23,6 +23,7 @@ class DataContextStore(ConfigurationStore):
         DataContextVariableSchema.DATASOURCES,
         DataContextVariableSchema.EVALUATION_PARAMETER_STORE_NAME,
         DataContextVariableSchema.EXPECTATIONS_STORE_NAME,
+        DataContextVariableSchema.INCLUDE_RENDERED_CONTENT,
```
cdkini: We want to exclude this from the payload sent to Cloud? Just confirming

tests/checkpoint/test_simple_checkpoint.py
cdkini: ease of debugging if we get an index or key error but I'm good either way

tests/data_context/cloud_data_context/test_include_rendered_content.py
cdkini: This does seem like a bug we should resolve as part of Thin SaaS though. I agree it is out of scope here but we may want to prioritize as "None" isn't a good look on our end.

tests/data_context/test_data_context_variables.py
```diff
@@ -498,18 +523,27 @@ def test_file_data_context_variables_e2e(
...
+    # Prepare updated progress_bars to set and serialize to disk
     updated_progress_bars: ProgressBarsConfig = copy.deepcopy(progress_bars)
     updated_progress_bars.globally = False
     updated_progress_bars.profilers = True
 
+    # Prepare updated include_rendered_content to set and serialize to disk
+    updated_include_rendered_content: IncludeRenderedContentConfig = copy.deepcopy(
+        include_rendered_content
+    )
+    updated_include_rendered_content.expectation_validation_result = True
```
cdkini, line 538: This is an E2E test so once we set this once, all subsequent changes are just going to change `True` to `True`. Is there value in that?

great_expectations/core/expectation_suite.py
NathanFarmer: We had a previous chain about how this should be a `TypedDict` and this method is required due to circular imports.

great_expectations/data_context/data_context/abstract_data_context.py
NathanFarmer: Because this is a public method. We pull from `DataContext` variables otherwise.

great_expectations/data_context/data_context/base_data_context.py
NathanFarmer: It is used across about 6 different modules. We can move it into a util file, but we have identified util files as something to avoid as well.

tests/data_context/cloud_data_context/test_include_rendered_content.py
NathanFarmer: Update: remove `xfail` instead of uncomment.

great_expectations/data_context/store/data_context_store.py
NathanFarmer: Wow nice catch! No this was not intended. ✅

great_expectations/data_context/data_context/base_data_context.py
NathanFarmer: I was stuck on util files, but this should be able to be refactored into `AbstractDataContext`.

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59: I don't understand this answer.
Can we make it default to `Fase` instead of `None`? Or is there a reason to retain `None` as a possible value?
`include_rendered_content: bool = False,`

You don't need to explain the reason, just wanted to make sure this isn't confusion over the meaning of `Optional`.

NathanFarmer: If nothing is passed via this method we need to look to `DataContext` variables for the value. We will only know that is the case if `include_rendered_content` is `None`.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -864,6 +866,12 @@ def get_validator(
                     )
                 )
 
+        include_rendered_content: bool = (
```
Kilo59: Why is `include_rendered_content` a parameter for `get_validator()` if it's getting overwritten here?

Kilo59: Never mind I see what's happening.

great_expectations/data_context/data_context/base_data_context.py
Kilo59: I don't want to block progress, this may be appropriate here but in general I think we tend to over-use inheritance for code de-duplication.

Inheritance should primarily be used for polymorphism, not code re-use.
https://www.arhohuttunen.com/misconceptions-about-code-reuse/

tests/data_context/test_data_context_variables.py
NathanFarmer: Per our offline conversation this is not persistent.

great_expectations/data_context/data_context/abstract_data_context.py
NathanFarmer: In agreement per offline converstations.

tests/checkpoint/test_simple_checkpoint.py
NathanFarmer: ✅

cdkini (approved):

great_expectations/data_context/data_context/ephemeral_data_context.py
```diff
@@ -66,21 +66,22 @@ def _init_datasource_store(self) -> None:
         )
         self._datasource_store = datasource_store
 
-    def save_expectation_suite(  # type: ignore[override]
```
Kilo59, line 69: ❤️

Kilo59 (approved):

PR #5738, opened 2022-08-11
[MAINTENANCE] Clean up type hints around class constructors

cdkini:
- Remove all class type hints on lines that invoke a constructor call.

```bash
#!/bin/bash

grep -roh ": [A-Z][A-Za-z]* = [A-Z][A-Za-z]*(" great_expectations/ | sort -u | while read -r line; do
  replacement=$(echo $line | grep -o " = [A-Z][A-Za-z]*(")
  find . -name "*.py" | xargs sed -i '' "s/$line/$replacement/g"
done
```

### Definition of Done


Kilo59 (approved):
Would be nice make this an invoke task, with a `--check` option that exits with error code if the initial `grep` find violations.

If we register it as a dependency of `invoke type-check` It will automatically run as part of CI.
https://docs.pyinvoke.org/en/stable/getting-started.html#declaring-pre-tasks

Can even add a simple system check so that it does a no-op on windows machines.

kenwade4 (approved):

tannerbeam (approved):
looks like all formatting changes. don't see any material changes to schema or test events.

Kilo59:
https://gist.github.com/cdkini/90868011502c385da8ce231c2b282bb0

PR #5752, opened 2022-08-12
[BUGFIX] Patch flaky E2E Cloud test with randomized suite names

cdkini:
- We're currently hitting an issue where 3 parallel Azure workers are hitting the same suite, causing issues.
- To fix this, we randomize the suite name suffix to ensure we don't touch the same suite.

### Definition of Done


tests/data_context/test_base_data_context.py
```diff
@@ -174,7 +174,11 @@ def _closure(context: DataContext) -> Tuple[Validator, str]:
         # Create a suite to be used in Validator instantiation
         suites = context.list_expectation_suites()
         expectation_suite_ge_cloud_id = suites[0].ge_cloud_id
-        suite_name = "oss_e2e_test_suite"
+
+        # To ensure we don't accidentally impact parallel test runs in Azure, we randomly generate a suite name in this E2E test.
+        # To limit the number of generated suites, we limit the randomization to 5 numbers.
+        rand_suffix = random.randint(1, 5)
+        suite_name = f"oss_e2e_test_suite_{rand_suffix}"
```
cdkini: Is this a large enough margin? I don't want to overly populate the backend with E2E test suites.

If we have 3 parallel workers and they all have a 1/X chance of running a particular suite (where X is the range of possible numbers), what is acceptable to prevent or limit collisions?

Shinnnyshinshin: I dont have the full context here, but what is the "cost" of creating an additional ExpectationSuite in this situation? What if we did something like 9 or 15?

cdkini: No real cost but we populate the `dev` environment (both in the backend and in the UI)

cdkini: bumped to 10

Shinnnyshinshin (approved):
LGTM

tests/data_context/test_base_data_context.py
billdirks: 10 still seems low to me. Does someone have to manage it? Can we delete it afterwards? I'd prefer a much big number. With 3 parallel workers we still have a 28% chance of a collision:

First one is random, 0 chance of collision.
Second collides with 1/10.
Second worker doesn't collide but third one does: 9/10 * 2/10

1/10 + 9/10 * 2/10 = 0.28

Kilo59: Late to the party but...

Can we use the system hostname here?

https://stackoverflow.com/a/4271873/6304433

```python
>>> import platform
>>> platform.node()
...  gabriel-MacBook-Pro
```

PR #5762, opened 2022-08-15
use `request_headers` fixture

Kilo59:
Fix CI for #5758


tests/data_context/conftest.py
```diff
@@ -417,11 +417,6 @@ def ge_cloud_runtime_organization_id():
     return "a8a35168-68d5-4366-90ae-00647463d37e"
 
 
-@pytest.fixture
-def ge_cloud_runtime_access_token():
-    return "b17bc2539062410db0a30e28fb0ee930"
```
Kilo59, line 422: We have a `ge_cloud_access_token` fixture & a `ge_cloud_runtime_access_token` fixture.

We don't need both.
If for some reason we need a different token for a particular test and we want to use a fixture, then we should redefine `ge_cloud_access_token` at the local level.

https://docs.pytest.org/en/6.2.x/fixture.html#overriding-fixtures-on-various-levels

PR #5770, opened 2022-08-16
[MAINTENANCE]  invoke docker --detach and more typing

Kilo59:
- add `--detach` flag to `invoke docker` cmd, for running the docker container detached.
- separate `great_expectations/core` excludes to cover more of `core` and make gradual typing easier

https://docs.docker.com/engine/reference/run/#detached-vs-foreground

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -207,10 +207,11 @@ def mv_usage_stats_json(ctx):
         "name": "Docker image name.",
         "tag": "Docker image tag.",
         "build": "If True build the image, otherwise run it. Defaults to False.",
+        "detach": "Run container in background and print container ID. Defaults to False.",
         "cmd": "Command for docker image. Default is bash.",
     }
 )
-def docker(ctx, name="gx38local", tag="latest", build=False, cmd="bash"):
+def docker(ctx, name="gx38local", tag="latest", build=False, detach=False, cmd="bash"):
     """
     Build or run gx docker image.
     """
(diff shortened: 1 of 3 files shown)


tasks.py
```diff
@@ -221,29 +222,40 @@ def docker(ctx, name="gx38local", tag="latest", build=False, cmd="bash"):
...
+        cmds.extend(
+            [
+                "buildx",
+                "build",
+                "-f",
+                "docker/Dockerfile.tests",
+                f"--tag {name}:{tag}",
+                *[
+                    f"--build-arg {arg}"
+                    for arg in ["SOURCE=local", "PYTHON_VERSION=3.8"]
```
cdkini: We've agreed on 3.8 as the standard here? Or should we test with our minimum supported version 3.7?

cdkini (approved):

tasks.py
```diff
@@ -221,29 +222,40 @@ def docker(ctx, name="gx38local", tag="latest", build=False, cmd="bash"):
...
+                "-it",
+                "--rm",
+                "--mount",
+                f"type=bind,source={filedir},target=/great_expectations",
+                "-w",
+                "/great_expectations",
+                f"{name}:{tag}",
+                f"{cmd}",
+            ]
+        )
```
Shinnnyshinshin, line 271: ❤️

Shinnnyshinshin (approved):
LGTM!

billdirks (approved):
Thank you!

tasks.py
Kilo59: I can make this another CLI argument.

tasks.py
```diff
@@ -221,29 +234,40 @@ def docker(ctx, name="gx38local", tag="latest", build=False, cmd="bash"):
...
-            "--mount",
-            f"type=bind,source={filedir},target=/great_expectations",
-            "-w",
-            "/great_expectations",
-            f"{name}:{tag}",
-            f"{cmd}",
-        ]
+        cmds.append("run")
+        if detach:
+            cmds.append("--detach")
```
Kilo59, line 259: I have to insert the `--detach` here, or it doesn't work.
Or at least it doesn't work if I append to the end of the cmds.

PR #5775, opened 2022-08-17
[MAINTENANCE] use `id` instead `id_` 

Kilo59:

- [x] - replace `DataSourceConfig` constructor/`__init__` `id_` keywords with `id`
- [x] - replace all Config class constructor/`__init__` `id_` keywords with `id`
- [x] - find solution to IDE (pycharm) warnings
  1. https://www.jetbrains.com/help/pycharm/disabling-and-enabling-inspections.html#suppress-inspections
  2. ~https://www.jetbrains.com/help/pycharm/creating-custom-inspections.html~ - can't do this for python code
  3. update pycharm shadow builtin names to globally ignore `id`
      * downside is that I can't figure out how to only ignore the `id` shadow in method signatures.

## Optional changes

- [x] - replace `id_` fields with `id`
- [x] - replace all `id_` constructor calls with `id`

## Notes
Initially, we used `id_` and `self.id_` for keyword arguments and class attributes to avoid conflicts with the built-in` id()`.
When serializing these objects we dump these fields as `id` which leads to headaches when we try to de-serialize or initialize objects with `id`.

We believe this effort to avoid shadowing `id()` leads to more problems than it solves. We only do this as part of small locally scoped methods which don't need to use `id()` anyway.

If for some reason we do need to use `id()` within the method scope we can always use `__builtins__.id`.

It's worth noting that `id()` IS NOT used for identity `x is y` checks. [Gist demo of this](https://gist.github.com/Kilo59/f055b2863351675e7f8d89ddd040c9ff).

```python
foo = "bar"
assert id(foo) == __builtins__.id(foo)
```

### Related SO Discussions

https://stackoverflow.com/questions/77552/id-is-a-bad-variable-name-in-python/79198#79198
https://stackoverflow.com/questions/15667189/what-is-the-id-function-used-for/15667328#15667328

## Questions
Should we...
a. universally replace `id_` keyword arguments with `id`? or
b. limit these changes to class `__init__` constructors? ... but keep `id_` in other non `__init__` functions.

## Alternatives
As mentioned in #5697 some alternatives to this change are below.
Some of these steps could become future enhancements, but all of them require substantially more changes than just using `id` keyword arguments.

1. ~Pulling `id` out of `**kwargs`.~
5. Replacing the `id_` + `name` data-clumps with a custom parameter/`Identifier` object.
6. Using factories for config class instantiation and doing the `id` -> `id_` translation in the factory method.
    * The issue comes from doing `**kwarg` unpacking into `__init__` which is expecting `id_` but gets `id`.
      * Example of this is `pydantic` has a series of `parse_<X>` class methods that can be used to instantiate models/classes from objects or bytes/strings or files.
      * https://pydantic-docs.helpmanual.io/usage/models/#helper-functions
      * https://github.com/pydantic/pydantic/blob/main/pydantic/main.py#L509-L561
      * Example implementation - https://github.com/great-expectations/great_expectations/pull/5794
 

<h2 id="ide-warnings">Dealing with IDE warnings </h2>

### PyCharm
Ideally, we would configure PyCharm to selective ignore the `Shadowing built-in names` inspection only when `id` is part of an `__init__` constructor (or any method signature).

Unfortunately, this does not appear to be possible.
We can configure the inspection to only suppress `id`, but this applies globally.

Screen Shot 2022-08-18 at 10 00 14 PM

The global `id` ignore could be mitigated by using something like this flake8-builtins plugin and configuring it to only ignore `id`  in certain files.
https://github.com/gforcada/flake8-builtins

Also, we might want to ensure that we don't use `id()` in our codebase, and if we need to (which is doubtful), we can instead use `__builtins__.id()`.
This should be easy to enforce via a linter plugin.
Thanks, @wookasz for the idea.

### VS Code

...

## Definition of Done

- ~I have made corresponding changes to the documentation~

Thank you for submitting!


great_expectations/data_context/types/base.py
```diff
@@ -960,7 +960,7 @@ class Meta:
         required=False,
         allow_none=True,
     )
-    id_ = fields.String(
+    id = fields.String(
         required=False,
         allow_none=True,
         data_key="id",
```
Kilo59: Should be able to drop the `data_key` setting.

anthonyburdi: Good call, done!

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -688,10 +688,9 @@ class Meta:
         required=True,
         allow_none=False,
     )
-    id_ = fields.String(
+    id = fields.String(
         required=False,
-        allow_none=False,
-        data_key="id",
+        allow_none=True,
```
anthonyburdi, line 693: I believe this was erroneously set to allow_none=False, which is needed to be backward compatible for existing id-less configs.

Shinnnyshinshin (approved):
Love it :) Thank you for your work with this @anthonyburdi and @Kilo59

PR #5778, opened 2022-08-17
[MAINTENANCE] Add data connector names in datasource config

anthonyburdi:
- Add the data connector name to datasource configs on schema load
Note: This PR is intended to split off part of #5763 and #5664
- Changes to serializers (JsonConfigSerializer is now not specific to datasources, schema is required, rename to  NamedDatasourceSerializer and use in `list_datasources()`, consistent return type)

### Definition of Done

Thank you for submitting!


great_expectations/data_context/types/base.py
```diff
@@ -1049,6 +1055,11 @@ def validate_schema(self, data, **kwargs):
     # noinspection PyUnusedLocal
     @post_load
     def make_datasource_config(self, data, **kwargs):
+        # Add names to data connectors
+        for data_connector_name, data_connector_config in data.get(
+            "data_connectors", {}
+        ).items():
+            data_connector_config["name"] = data_connector_name
```
Kilo59, line 1060: If the answer is yes that's fine but...
Do we want to be doing this in-place mutation of `data`?

Shinnnyshinshin (approved):
LGTM :) Thank you @anthonyburdi

great_expectations/data_context/types/base.py
```diff
@@ -1049,7 +1055,13 @@ def validate_schema(self, data, **kwargs):
     # noinspection PyUnusedLocal
     @post_load
     def make_datasource_config(self, data, **kwargs):
-        return DatasourceConfig(**data)
+        # Add names to data connectors
+        data_copy: dict = copy.deepcopy(data)
```
cdkini: Do we truly need a deepcopy here? I don't think so and would try to avoid it due to the performance hit.

anthonyburdi: I could go either way here. I think it adds some safety if `data` is used elsewhere. I don't have a strong preference.

anthonyburdi: I'll remove it - if we need to ensure safety we can always add it back.

cdkini: hmm I'm pretty confident it isn't used after looking at the underlying Marshmallow code but I could go either way too!

Kilo59 (approved):

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -602,29 +605,49 @@ def get_datasource(
...
 
         Returns:
             List(dict): each dictionary includes "name", "class_name", and "module_name" keys
         """
         datasources: List[dict] = []
-        substitutions: dict = self._determine_substitutions()
 
         datasource_name: str
         datasource_config: DatasourceConfig
+        serializer = ListDatasourceSerializer(schema=datasourceConfigSchema)
```
cdkini: Just brainstorming names since this current one is very strongly coupled with the `list_datasource` method - perhaps something about user friendly, human readable, etc? I'm good with this though.

tests/data_context/store/test_datasource_store.py
```diff
@@ -83,11 +87,22 @@ def _assert_serialized_datasource_configs_are_equal(
...
+        serializers = [DictConfigSerializer(schema=datasourceConfigSchema)] * (
+            len(datasource_configs) + 1
+        )
+    else:
+        if len(serializers) <= 1:
+            raise AssertionError("Must provide at least 2 datasource serializers")
+        if not len(datasource_configs) == len(serializers):
+            raise AssertionError(
+                "Must provide the same number of serializers as datasource configs"
+            )
```
cdkini, line 100: This is all just used internally in tests right? Do we need error checking like this here?

tests/data_context/store/test_datasource_store.py
```diff
@@ -68,6 +71,7 @@ def test_datasource_store_with_bad_key_raises_error(
 
 def _assert_serialized_datasource_configs_are_equal(
     datasource_configs: List[DatasourceConfig],
+    serializers: Optional[List[AbstractConfigSerializer]] = None,
```
cdkini, line 74: Perhaps we pair config with serialized in a series of tuples? Good either way.

This could also be a nice place to use `zip()` to iterate through equally sized collections.

tests/datasource/test_datasource_serialization.py
```diff
@@ -19,6 +23,13 @@
                 "class_name": "Datasource",
                 "module_name": "great_expectations.datasource",
             },
+            {
+                "class_name": "Datasource",
+                "module_name": "great_expectations.datasource",
+                # TODO: AJB 20220818 Why is this not filtered out in the round trip?
```
cdkini: Should this note still be here?

cdkini: Same below

tests/datasource/test_datasource_serialization.py
```diff
@@ -60,39 +77,148 @@
...
             round_tripped.get("id_")
             == observed_dump.get("id")
             == expected_serialized_datasource_config.get("id")
+            == expected_roundtrip_config.get("id_")
         )
+
+        if (
+            hasattr(datasource_config, "data_connectors")
+            and datasource_config.data_connectors
+        ):
```
cdkini, line 181: Could we just do `datasource_config["data_connectors"]`? Connectors a required attr right why would we have to check here?

tests/data_context/store/test_datasource_store.py
anthonyburdi: True, just added reflexively. Probably not required but might be helpful when using this method to build new tests. I know I screwed it up when writing the tests.

tests/datasource/test_datasource_serialization.py
anthonyburdi: Nope! Good catch

tests/datasource/test_datasource_serialization.py
anthonyburdi: Some of the configs don't have data_connectors and it isn't required in the config :/  https://github.com/great-expectations/great_expectations/blob/4bb482c8beffd26d5b504cc8ccb6189cb4903d30/great_expectations/data_context/types/base.py#L889-L890

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: How about `NamedDatasourceSerializer`?

great_expectations/datasource/datasource_serializer.py
```diff
@@ -9,12 +9,12 @@
 serializer = YAMLReadyDictDatasourceConfigSerializer()
 serialized_value = serializer.serialize(datasource_config)
 """
+from great_expectations.core.configuration import AbstractConfig
```
Kilo59: If this is just for type checking I would go the `if TYPE_CHECKING:` route.
This prevents extra import cycles, which slows things down and can lead to cyclic imports.

```python
from typing import TYPE_CHECKING:

if TYPE_CHECKING:
   from great_expectations.core.configuration import AbstractConfig


...

class YAMLReadyDictDatasourceConfigSerializer(AbstractConfigSerializer):
    def serialize(self, obj: 'AbstractConfig') -> dict:

```

anthonyburdi: Good call! Done 🙇

PR #5784, opened 2022-08-17
[MAINTENANCE] Remove `GeCloudIdentifier` creation in `Checkpoint.run()`

cdkini:
- Instead of creating an identifier object, we simply want to pass the checkpoint id (if available)

### Definition of Done

### DIFF
--- a/great_expectations/checkpoint/checkpoint.py
+++ b/great_expectations/checkpoint/checkpoint.py
@@ -382,10 +382,7 @@ class BaseCheckpoint(ConfigPeer):
             )
             checkpoint_identifier = None
             if self.data_context.ge_cloud_mode:
-                checkpoint_identifier = GeCloudIdentifier(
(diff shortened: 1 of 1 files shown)


great_expectations/checkpoint/checkpoint.py
```diff
@@ -382,10 +378,7 @@ def _run_validation(
             )
             checkpoint_identifier = None
             if self.data_context.ge_cloud_mode:
-                checkpoint_identifier = GeCloudIdentifier(
-                    resource_type=GeCloudRESTResource.CHECKPOINT,
-                    ge_cloud_id=str(self.ge_cloud_id),
-                )
+                checkpoint_identifier = self.ge_cloud_id
```
wookasz, line 381: nit: we probably want to rename this to `checkpoint_id` here and where it's passed as an argument to avoid confusion with `GeCloudIdentifier` but I don't know that this is pressing

wookasz (approved):

PR #5792, opened 2022-08-18
[MAINTENANCE] Refactor `save_profiler` to remove explicit `name` and `ge_cloud_id` args

cdkini:
- Remove `name` and `ge_cloud_id` from signature since we can find them in the input profiler object
- Return the updated profiler from `save_profiler` for future use
- Add integration tests for the full get/set workflow

### Definition of Done


great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -2250,7 +2255,16 @@ def save_profiler(
         else:
             key = ConfigurationIdentifier(configuration_key=name)
 
-        self.profiler_store.set(key=key, value=profiler.config)
+        response = self.profiler_store.set(key=key, value=profiler.config)
+        if isinstance(response, GeCloudResourceRef):
+            ge_cloud_id = response.ge_cloud_id
```
cdkini, line 2260: I don't like this very much but it's the cleanest way to get things working as intended! The hope is to refactor this out later on as we shift more Cloud-specific logic into relevant classes.

Without this, the second test I wrote doesn't work.

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -2250,7 +2255,16 @@ def save_profiler(
...
+
+        # If an id is present, we want to prioritize that as our key for object retrieval
+        if ge_cloud_id:
+            name = None
```
cdkini, line 2264: Feels a little weird but we do an XOR with id and name in `get_profiler`.

Should we allow both but prioritize the id?

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -139,6 +139,10 @@ def __init__(
 
         self._rule_states = []
 
+    @property
+    def ge_cloud_id(self) -> Optional[str]:
+        return self._id
```
cdkini, line 144: Named this property `ge_cloud_id` just for parity but I'm happy to change it to whatever we deem acceptable.

great_expectations/data_context/data_context/base_data_context.py
Kilo59: Does this still work if you just attempt to pull the `id` (`ge_cloud_id`) without the `isinstance()` check?
```python
ge_cloud_id = getattr(response, "ge_cloud_id", None)
```

And this will be even better once we replace `ge_cloud_id` with a more generic `id`/`id_`/`identifier`

Kilo59 (approved):
This is much cleaner, great work!

2 small, non-blocking suggestions.

great_expectations/data_context/data_context/base_data_context.py
Kilo59: I agree that it feels a bit weird.

Don't know all the ins & outs of this but, this feels like a decision we should delegate to `get_profiler()`

cdkini: So `get_profiler` is used as a helper here but is also its own stand alone method for object retrieval. If a user tries to get their profiler using both name and id, what is the expected behavior? Should we error or have some order of precedence? We're currently using the latter pattern and in order to get these changes to work with that, we need to ensure that we're getting rid of any conflicts here before invoking the `get_profiler()` method.

great_expectations/data_context/data_context/base_data_context.py
cdkini: hmm I think we need to refactor the `set` response type across the board to clean this up. I'm wary of `getattr` but I also see the merits of using that here. Good with either!

PR #5797, opened 2022-08-20
[MAINTENANCE] Surface cloud errors

Kilo59:
- extract and surface errors from the Ge-Cloud response
- consistently (almost) raise HTTPErrors when 4XX & 5XX status codes returned (not including DELETE)
- better error messages from timeouts
- log all 4XX & 5XX responses at the`warning` level

## Testing
- move `MockResponse` up one level `cloud_data_context.conftest.py`
- add `mock_http_unavailable` fixture and update `MockResponse` 
- move tests from `test_datasource_store_cloud_backend.py` to `test_ge_store_backends.py`

### Definition of Done

- ~I have made corresponding changes to the documentation~


great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -21,6 +21,20 @@
 
 logger = logging.getLogger(__name__)
 
+CONTACT_SUPERCONDUCTIVE_MESSAGE = (
+    "Please contact superconductive at foobar@superconductive.com"
```
Kilo59: What email should this be?
This might be something to externalize via config.

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -206,9 +245,13 @@ def _get(self, key: Tuple[str, ...]) -> ResponsePayload:  # type: ignore[overrid
             raise StoreBackendError(
                 f"Unable to get object in GE Cloud Store Backend: {jsonError}"
             )
-        except (requests.HTTPError, requests.Timeout) as http_exc:
+        except requests.HTTPError as http_err:
             raise StoreBackendError(
-                f"Unable to get object in GE Cloud Store Backend: {http_exc}"
+                f"Unable to get object in GE Cloud Store Backend: {_get_user_friendly_error_message(http_err)}"
```
Kilo59, line 250: Alternative to stuffing all these error details into the exception would be to just log them, at `WARNING` or `ERROR` levels.

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -206,9 +245,13 @@ def _get(self, key: Tuple[str, ...]) -> ResponsePayload:  # type: ignore[overrid
...
+            )
+        except requests.Timeout as timeout_exc:
+            raise StoreBackendError(
+                f"Unable to get object in GE Cloud Store Backend: {timeout_exc}"
```
wookasz: Myabe "Unable to get object in GE Cloud Store Backend. This is likely a transient error. Please try again."

great_expectations/data_context/store/ge_cloud_store_backend.py
erik-hencier: support@greatexpectations.io

great_expectations/data_context/store/ge_cloud_store_backend.py
Kilo59: Added.

great_expectations/data_context/store/ge_cloud_store_backend.py
Kilo59: Thanks! Updated.

tests/data_context/store/test_datasource_store_cloud_backend.py
```diff
@@ -182,3 +183,43 @@ def test_datasource_store_delete_by_id(
...
+            "delete",
+            "delete",
+            [],
+            marks=pytest.mark.xfail(
+                reason="We do not raise errors on delete fail", strict=True
+            ),
+        ),
+    ],
+)
+def test_datasource_http_error_handling(
```
Kilo59, line 204: This is the only new test.

Might want to move this to the new `tests/data_context/store/test_ge_store_backends.py` module.

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -32,6 +44,33 @@ class ResponsePayload(TypedDict):
...
+        support_message.append(f"Request-Id: {request_id}")
+
+    try:
+        error_json: ErrorPayload = http_exc.response.json()
+        errors = error_json.get("errors")
+        if errors:
+            support_message.append(json.dumps(errors))
+
+    except json.JSONDecodeError:
+        support_message.append(f"Please contact superconductive at {SUPPORT_EMAIL}")
```
Kilo59, line 70: Note that this is only provided if there is no JSON content in the response for us to include.
For example, if GE Cloud has an outage.

Otherwise, this error message contains the `errors` fields.

tests/data_context/conftest.py
```diff
@@ -639,3 +643,86 @@ def shared_called_with_request_kwargs(request_headers) -> dict:
...
+
+    def mocked_response(*args, **kwargs):
+
+        return MockResponse(
+            {"code": 503, "detail": "API is unavailable"},
+            503,
+        )
+
+    # should have been able to do this by mocking `requests.request` but this didn't work
+    with unittest.mock.patch.multiple(
```
Kilo59, line 714: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.multiple

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -21,6 +21,18 @@
 
 logger = logging.getLogger(__name__)
 
+SUPPORT_EMAIL = "support@greatexpectations.io"
```
anthonyburdi, line 24: I would shy away from including a hardcoded value like this in the library and instead make it a bit more vague like "please contact support." If we pull the error message from the payload and that has an email or link that can work.

Kilo59: https://github.com/great-expectations/great_expectations/pull/5797/files#r952817953

We could externalize this as an environment variable configuration field, but we need something like this in the event that the Cloud Service is down or totally unavailable.

anthonyburdi: Ah ok thanks for clarifying. Maybe instead of the email addr though point to the website where we will likely have a support button / section? Totally up to you and product for whatever the best approach is to keep friction low - just a suggestion.

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -21,6 +21,18 @@
...
+
+
+class ErrorDetail(TypedDict):
+    code: Optional[str]
+    detail: Optional[str]
+    source: Optional[str]
```
Kilo59, line 30: Is `source` a string or another nested dict?
I've seen conflicting things.

great_expectations/data_context/store/ge_cloud_store_backend.py
Kilo59: I think that's a good idea.

@wookasz requested we provide a contact email, but I'm happy to update it to whatever we want, whether in this PR or in follow-up.

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -426,11 +484,19 @@ def remove_key(self, key):
...
+                f"Unable to delete object in GE Cloud Store Backend: {_get_user_friendly_error_message(http_exc)}"
+            )
             return False
+        except requests.Timeout as timeout_exc:
+            logger.exception(timeout_exc)
+            raise StoreBackendError(
+                "Unable to delete object in GE Cloud Store Backend: This is likely a transient error. Please try again."
+            )
         except Exception as e:
             logger.debug(str(e))
```
anthonyburdi, line 501: Just a thought and non blocking - looks like some repeated code, any way to make this more dry? Maybe we can pull this error handling out into a handler that takes a callable and the error messages? Callable for the logic in the `try:`?

Kilo59: Yes, I have a TODO in here that I'd like to extract all the HTTP stuff into a dedicated client/service.
I plan that as a followup.

great_expectations/data_context/store/ge_cloud_store_backend.py
cdkini: Okay with this for now but I'd like these values to come from the Cloud backend if at all possible (perhaps a refactor down the line).

anthonyburdi (approved):
LGTM! Thanks for doing this!

tests/data_context/store/test_ge_store_backends.py
```diff
@@ -0,0 +1,243 @@
...
+import pytest
+
+from great_expectations.data_context.store import GeCloudStoreBackend
+from great_expectations.data_context.store.ge_cloud_store_backend import (
+    GeCloudRESTResource,
+)
+from great_expectations.data_context.types.base import CheckpointConfig
+
+
+def test_GeCloudStoreBackend(
```
cdkini: As this file is only testing the `GeCloudStoreBackend`, can we rename the file to be something like `test_ge_cloud_store_backend`? I want to make sure it's very clear that this is a Cloud-specific file. Additionally, can we mark each test with `@cloud`? Finally, we could probably split up this test here into multiple unit tests (but that's out of scope here).

cdkini (approved):

tests/data_context/store/test_ge_store_backends.py
anthonyburdi: Looked like just a move but agree with @cdkini 's suggestions ☝️

Kilo59: Yeah, this is just a move, nothing has changed here. I had originally thought I was going to put the new test here as well but I didn't end up doing that.
It might be cleaner if I dropped this commit from the PR and do it as part of the follow-up.

PR #5811, opened 2022-08-23
[MAINTENANCE] Update distinct metric dependencies and implementations

NathanFarmer:
- GREAT-761/GREAT-1201
- GREAT-761/GREAT-1221
- count.distinct_values.count.under_threshold → net new metric
- column.distinct_values → update to not rely on column.value_counts
- column.distinct_values.count → update to not rely on column.value_counts
- column.value_counts → update creation method of Pandas Series

### Definition of Done


ghost:
<details>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=5811&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

    [Try it now!](https://app.codesee.io/r/reviews?pr=5811&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)
</details>

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=5811&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map Legend

great_expectations/expectations/metrics/column_aggregate_metrics/column_value_counts.py
```diff
@@ -112,9 +113,10 @@ def _sqlalchemy(
         results = execution_engine.engine.execute(
             query.select_from(selectable)
         ).fetchall()
+        results_array = np.asarray(results)
         series = pd.Series(
-            [row[1] for row in results],
-            index=pd.Index(data=[row[0] for row in results], name="value"),
+            results_array[:, 1],
```
NathanFarmer: Takes about half the time: https://gist.github.com/NathanFarmer/f2cd3ff62d3c1bf08858cdab642cb2a3

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -30,9 +33,18 @@ def _sqlalchemy(
...
+            metric_domain_kwargs, MetricDomainTypes.COLUMN
+        )
+        column = accessor_domain_kwargs["column"]
+
+        query = sa.select(sa.column(column)).distinct()
+        results = execution_engine.engine.execute(
+            query.select_from(selectable)
+        ).fetchall()
+
+        return {row[0] for row in results}
```
NathanFarmer: Vectorized operation is not faster here due to overhead of converting to and from numpy array.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -42,49 +54,67 @@ def _spark(
...
+        return set(distinct_values)
 
 
 class ColumnDistinctValuesCount(ColumnAggregateMetricProvider):
     metric_name = "column.distinct_values.count"
 
     @column_aggregate_value(engine=PandasExecutionEngine)
-    def _pandas(cls, column, **kwargs):
+    def _pandas(cls, column, _metrics, **kwargs):
+        if "column.distinct_values" in _metrics:
```
NathanFarmer: Curious to know how we feel about this conditional. If we already know the distinct values, we don't need to query again. If we only want the distinct values count, we can get that in a single step instead of forcing a dependency resolution (distinct values) and performing in-memory operations (count) on that.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -19,7 +22,7 @@ class ColumnDistinctValues(ColumnAggregateMetricProvider):
     metric_name = "column.distinct_values"
 
     @column_aggregate_value(engine=PandasExecutionEngine)
-    def _pandas(cls, column, **kwargs):
+    def _pandas(cls, column, **kwargs) -> Set[Any]:
```
billdirks: Thanks for typing the output! Can you type the inputs?

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
billdirks: Maybe add this as a comment in the code since someone may look at this and spend cycles doing that.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -30,9 +33,18 @@ def _sqlalchemy(
...
-        observed_value_counts = metrics["column.value_counts"]
-        return set(observed_value_counts.index)
+    ) -> Set[Any]:
+        selectable, _, accessor_domain_kwargs = execution_engine.get_compute_domain(
+            metric_domain_kwargs, MetricDomainTypes.COLUMN
+        )
+        column = accessor_domain_kwargs["column"]
+
+        query = sa.select(sa.column(column)).distinct()
+        results = execution_engine.engine.execute(
```
billdirks: What are some example queries here? Is this returning all the distinct values for a column?

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -42,49 +54,67 @@ def _spark(
...
-                metric_name="column.value_counts",
-                metric_domain_kwargs=metric.metric_domain_kwargs,
-                metric_value_kwargs={
-                    "sort": "value",
-                    "collate": None,
-                },
-            )
-
-        return dependencies
+        return set(distinct_values)
```
billdirks, line 94: Is this used as a dependency for some other metric? I'm asking because I often hear that we want to keep data out of gx. That is, I understand we would want distinct counts but what is the use case for the distinct values?

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -42,49 +54,67 @@ def _spark(
...
 
 
 class ColumnDistinctValuesCount(ColumnAggregateMetricProvider):
     metric_name = "column.distinct_values.count"
 
     @column_aggregate_value(engine=PandasExecutionEngine)
-    def _pandas(cls, column, **kwargs):
+    def _pandas(cls, column, _metrics, **kwargs):
```
billdirks: What does the `_` in `_metrics` signify?
Also, could you add type annotations here and below?

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -42,49 +54,67 @@ def _spark(
...
+        _metrics,
+        **kwargs,
+    ):
+        if "column.distinct_values" in _metrics:
+            return F.lit(len(_metrics["column.distinct_values"]))
+        return F.countDistinct(column)
+
+
+class ColumnDistinctValuesCountUnderThreshold(ColumnAggregateMetricProvider):
+    metric_name = "column.distinct_values.count.under_threshold"
```
alexsherstinsky, line 132: @NathanFarmer Would you be opposed to a slight renaming -- to "column.distinct_values.count_under_threshold" (replacing the last period with the underscore)?  Thanks!

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
alexsherstinsky: @NathanFarmer If this new implementation is not faster, should we keep the old one (because it requires fewer lines, due to reusing the existing machinery)?

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
billdirks: I am good with this especially if `len` does an attribute lookup vs a loop in python. What is the type of the `_metrics["column.distinct_values"]`?

alexsherstinsky: @NathanFarmer Let us discuss -- I think that there is a potential "no-op"/tautology involved there (easier via zoom).  Thanks!

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -93,21 +123,23 @@ def _sqlalchemy(
...
+    ) -> bool:
+        threshold = metric_value_kwargs.get("threshold")
+        column_distinct_values_count = metrics.get("column.distinct_values.count")
+        return column_distinct_values_count < threshold
 
     @metric_value(engine=SparkDFExecutionEngine)
     def _spark(
         cls,
-        execution_engine: SqlAlchemyExecutionEngine,
+        execution_engine: SparkDFExecutionEngine,
```
billdirks: This seems like an odd parameter since this is called `_spark` and it has a spark decorator. We don't use it here but I'm guessing some interface requires it? It seems like it should have a default or be factored out (in future work). Does that seem reasonable/desirable?

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -125,17 +157,10 @@ def _get_evaluation_dependencies(
             execution_engine=execution_engine,
             runtime_configuration=runtime_configuration,
         )
-
-        if isinstance(
-            execution_engine, (SqlAlchemyExecutionEngine, SparkDFExecutionEngine)
-        ):
-            dependencies["column.value_counts"] = MetricConfiguration(
-                metric_name="column.value_counts",
+        if metric.metric_name == "column.distinct_values.count.under_threshold":
```
billdirks, line 177: Does this mean this is the only column distinct values metric that has a dependency?

great_expectations/expectations/metrics/column_aggregate_metrics/column_value_counts.py
```diff
@@ -29,7 +30,7 @@ def _pandas(
         metric_value_kwargs: Dict,
         metrics: Dict[str, Any],
         runtime_configuration: Dict,
-    ):
+    ) -> pd.Series:
```
billdirks, line 40: Thanks!

tests/expectations/metrics/test_core.py
```diff
@@ -3975,25 +3963,121 @@ def test_distinct_metric_sa(sa):
...
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "execution_engine_type",
+    [
+        pytest.param(SqlAlchemyExecutionEngine, id="sa"),
+        pytest.param(SparkDFExecutionEngine, id="spark"),
+    ],
+)
+def test_distinct_metric_sa_and_spark(
```
alexsherstinsky: @NathanFarmer Is there any danger in encountering NULL execution engine if one is not installed for the given test?  On the other hand, if this is completely safe when one or both execution engines are not installed (giving proper error messages, etc.), should we then create a task to update all tests to combine SQLAlchemy & Spark using the parameterization patter like you put forth here?  Thanks!

alexsherstinsky (changes requested):
@NathanFarmer I added a couple of small requests and questions.  Thank you!

great_expectations/expectations/metrics/column_aggregate_metrics/column_value_counts.py
billdirks: I'm wondering if we can add column aliases and use that here? Maybe if forces a dict instead of a list so is slower?

great_expectations/expectations/metrics/column_aggregate_metrics/column_value_counts.py
```diff
@@ -151,9 +153,11 @@ def _spark(
         elif sort == "count":
             value_counts = value_counts.orderBy(F.desc("count"))
         value_counts = value_counts.collect()
+        # Numpy does not always infer the correct DataTypes for Spark df, so we cannot use vectorized approach.
```
billdirks, line 164: Thanks for the comment!

tests/expectations/metrics/test_core.py
```diff
@@ -3922,7 +3922,7 @@ def test_median_metric_spark(spark_session):
     assert results == {desired_metric.id: 2}
 
 
-def test_distinct_metric_spark(spark_session):
+def test_value_counts_metric_spark(spark_session):
```
billdirks, line 3927: Could you mark these tests with `unit`, `integration`, or `e2e`?

tests/expectations/metrics/test_core.py
```diff
@@ -3975,25 +3963,121 @@ def test_distinct_metric_sa(sa):
...
+        pytest.param(SqlAlchemyExecutionEngine, id="sa"),
+        pytest.param(SparkDFExecutionEngine, id="spark"),
+    ],
+)
+def test_distinct_metric_sa_and_spark(
+    execution_engine_type: Union[SqlAlchemyExecutionEngine, SparkDFExecutionEngine],
+    sa,
+    spark_session,
+):
+    if execution_engine_type is SqlAlchemyExecutionEngine:
```
billdirks: You could parameterize on this so you don't need to do this condition in the test. Whatever you think is clearer works for me.

billdirks (commented):
Thanks for doing this! I have mostly questions as this is mostly new to me. I haven't gone through all the tests yet but will continue to read.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
NathanFarmer: This is no longer applicable since I learned that `_metrics` can only be supplied via `MetricConfiguration` `metric_dependencies` and not via `get_evaluation_dependencies` which is what is used during metric resolution.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
NathanFarmer: @billdirks I added the comment

NathanFarmer: @alexsherstinsky `column.value_counts` should not be required for `column.distinct_values`. 1) `column.value_counts` is a slower query that we should try to avoid if possible and 2) If possible we should not be passing around in-memory data with extra columns (counts) that we don't need.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
NathanFarmer: The only issue I have with this is that it doesn't align to `column_values.z_score.under_threshold`. This depends on `column.distinct_values.count` so shouldn't we communicate that information in the name?

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
NathanFarmer: Yes that is correct. `SELECT DISTINCT(column) FROM table`

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
NathanFarmer: In the future `column.distinct_values` will be used in Expectations that currently leverage `column.value_counts` where the counts aren't needed. The reason why we can't return a `column_aggregate_partial` here is that there is no PySpark function for `distinct` that can be used on a PySpark `column` object. Instead we have to build the entire query and get the values in the metric implementation.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
NathanFarmer: I learned that `_metrics` are only able to be passed in via `MetricConfiguration.metric_dependencies`. This is not how metric resolution works, so using this private object doesn't do anything really.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
NathanFarmer: We do use the `execution_engine` everywhere it is available after updates to this PR.

great_expectations/expectations/metrics/column_aggregate_metrics/column_value_counts.py
NathanFarmer: I'm not sure I understand the question. The `name` is added to the pd.Series and its index. Are you referring to this gist? https://gist.github.com/NathanFarmer/f2cd3ff62d3c1bf08858cdab642cb2a3

tests/expectations/metrics/test_core.py
NathanFarmer: Added `integration` mark ✅

tests/expectations/metrics/test_core.py
NathanFarmer: I tried this to no avail. At first I had `build_engine` methods in a fixture and it was miserable trying to figure out how to use parametrized fixtures. I ended up refactoring this into two tests because the logic was diverging for SQL and Spark. I feel like this should be the approach moving forward since we can have a diverging logic for the different engines in any metric.

tests/expectations/metrics/test_core.py
NathanFarmer: This issue no longer applies as we aren't testing multiple engines per test anymore.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
NathanFarmer: ✅

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
NathanFarmer: Yes, that is correct.

tests/expectations/metrics/test_core.py
NathanFarmer: If you know a better way of parametrizing fixtures, please let us know. We will be doing similar work over the next sprint.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -9,8 +12,21 @@
...
+    pyspark_sql_Column,
+    pyspark_sql_DataFrame,
+    pyspark_sql_Row,
+    sa,
+    sa_sql_expression_ColumnClause,
+    sa_sql_expression_Select,
+    sa_sql_expression_Selectable,
+    sqlalchemy_engine_Engine,
+    sqlalchemy_engine_Row,
+)
```
NathanFarmer: This ensures that this module can be imported even though `sqlalchemy` and `pyspark` aren't part of the base `requirements.txt`.

tests/rule_based_profiler/data_assistant/test_volume_data_assistant.py
```diff
@@ -1666,7 +1666,7 @@ def test_volume_data_assistant_result_serialization(
     assert len(bobby_volume_data_assistant_result.profiler_config.rules) == 2
 
 
-@pytest.mark.unit
+@pytest.mark.integration
```
NathanFarmer, line 1669: All of these tests use fixtures that mean they aren't unit tests.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
```diff
@@ -3386,13 +3386,16 @@ def _create_display_chart_for_column_domain_expectation(
...
         )
 
-        return self._chart_column_values(
-            expectation_type=expectation_type,
-            column_dfs=column_dfs,
-            metric_names=metric_names,
-            plot_mode=plot_mode,
-            sequential=sequential,
-        )
+        if len(column_dfs) > 0:
```
NathanFarmer, line 3395: This bug in plotting was cropping up in various situations when all metrics fail to resolve on a particular domain. I saw it happen while this PR was WIP and this just ensures that the plotting doesn't break in that case.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
alexsherstinsky: @NathanFarmer I see -- got it (agreed).  Thanks.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -19,95 +34,129 @@ class ColumnDistinctValues(ColumnAggregateMetricProvider):
...
-        return set(observed_value_counts.index)
+        metric_domain_kwargs: Dict[str, str],
+        **kwargs,
+    ) -> Set[Any]:
+        """
+        Past implementations of column.distinct_values depended on column.value_counts.
+        This was causing performance issues due to the complex query used in column.value_counts and subsequent
+        in-memory operations.
+        """
+        selectable: sa_sql_expression_Selectable
```
billdirks, line 52: Can you run `invoke types -p great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py`?
I'm curious if these type declarations will result in an error since the return types from `get_compute_domain` are less specific. If so, you can cast the variables to these types after the `get_compute_domain` call.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
```diff
@@ -19,95 +34,129 @@ class ColumnDistinctValues(ColumnAggregateMetricProvider):
     metric_name = "column.distinct_values"
 
     @column_aggregate_value(engine=PandasExecutionEngine)
-    def _pandas(cls, column, **kwargs):
+    def _pandas(cls, column: pd.Series, **kwargs) -> Set[Any]:
```
billdirks, line 37: thank you!

great_expectations/expectations/metrics/import_manager.py
```diff
@@ -18,11 +18,27 @@
...
     from sqlalchemy.engine import reflection
 except ImportError:
-    logger.debug("No SqlAlchemy module available.")
+    logger.debug("No SqlAlchemy.engine module available.")
     reflection = None
     sqlalchemy_engine_Engine = None
     sqlalchemy_engine_Row = None
 
+try:
+    import sqlalchemy.func.count as sa_func_count
```
billdirks, line 27: Is it useful to be able to import `sql.alchemy.engine`  without `sqlalchemy.func` or `sqlalchemy.sql`? I'm just wondering if we should wrap all these imports in 1 try/except.

great_expectations/rule_based_profiler/data_assistant_result/data_assistant_result.py
billdirks: Could you file a ticket and add a comment referencing it here?

billdirks (approved):
Thanks, this LGTM! 

I've added a few inline comments/questions.

great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
NathanFarmer:
```
mypy great_expectations/expectations/metrics/column_aggregate_metrics/column_distinct_values.py
Success: no issues found in 1 source file
```

great_expectations/expectations/metrics/import_manager.py
NathanFarmer: I actually got import errors during tests when I had them in one try/except. I think the tests need to be able to import either one or the other.

alexsherstinsky (approved):

PR #5812, opened 2022-08-23
[FEATURE] add failed expectations list to slack message

itaise:
Hi, 
We are using GE with slack and our users pointed out that a list of failed expectation types would be helpful when there are validation failures. It will be also helpful for deployments that don't use the data docs, so no easy way to see which expectation types failed.

- Add parameter show_failed_expectations to SlackNotificationAction and to SlackRenderer - 
This parameter is False by default, and when it is supplied, a list of failed expectations will be shown in the slack message. 

I tested it locally with a Slack App, you can see a screenshot here:
https://greatexpectationstalk.slack.com/archives/CV828B2UX/p1661263142975739?thread_ts=1661181991.262419&cid=CV828B2UX

Thank you!


talagluck (commented):
Thanks for this, @itaise - this is a great idea! A few notes:
- Would it make sense to also add the columns for which the Expectations are failing?
- Could you add a test for either the action or renderer (or both) to confirm this functionality?

itaise:
Thank you @talagluck :)
Some expectation are on column pairs, tables, or other, and dont have a
column property.
Do you think i should add the column, and the value will be only for the
expectations that have this property?
Sure, ill add a test, once the logic is final.

talagluck:
Hi Itai! I was thinking something like adding the column name, where relevant, the names of both columns for column pairs, etc. and then in the case of the table, adding the word "Table". Does that sound reasonable?

itaise:
Hi @talagluck! sure, thanks. Ill use the default parameter names if they are there. 
And for custom / other expectations it will just not show (at least in Phase 1)

itaise:
Hi @talagluck! I added the column/column pair/table reference to the slack message. 
also added unit tests. 
WDYT? thanks a lot!

great_expectations/render/renderer/slack_renderer.py
```diff
@@ -175,3 +194,39 @@ def _get_report_element(self, docs_link):
...
+            if not expectation["success"]:
+                expectation_name = expectation["expectation_config"]["expectation_type"]
+                expectation_kwargs = expectation["expectation_config"]["kwargs"]
+                failed_expectations_str += self.create_failed_expectation_text(
+                    expectation_kwargs, expectation_name
+                )
+        return failed_expectations_str
+
+    def create_failed_expectation_text(self, expectation_kwargs, expectation_name):
+        expectation_entity = self.get_failed_expectation_entity(
```
talagluck: Could we change the name of this to `get_failed_expectation_domain`?

great_expectations/render/renderer/slack_renderer.py
```diff
@@ -175,3 +194,39 @@ def _get_report_element(self, docs_link):
...
+            return "Table"
+
+        column_name, column_a, column_b = (
+            expectation_config_kwargs.get("column"),
+            expectation_config_kwargs.get("column_A"),
+            expectation_config_kwargs.get("column_B"),
+        )
+        if column_name:
+            return column_name
+        elif column_a and column_b:
```
talagluck, line 234: Could we add one more `if` condition for multicolumn Expectations, looking for a `column_list` kwarg as shown here? https://github.com/great-expectations/great_expectations/blob/d12e9af949181d3a5af59e5e7cc88261b850e83b/great_expectations/expectations/core/expect_multicolumn_sum_to_equal.py#L60

talagluck (commented):
Thanks very much for this, @itaise - this is awesome! A couple of minor changes. Thank you for the comprehensive testing as well. I'm going to ask one of our engineers on our testing side to take a look at this part specifically.

great_expectations/render/renderer/slack_renderer.py
itaise: thank you @talagluck! I made the required changes and added a test for the multi-column as well.

tests/render/test_SlackRenderer.py
```diff
@@ -272,3 +292,131 @@ def test_SlackRenderer_get_report_element():
...
+
+def test_SlackRenderer_get_failed_expectation_domain_column():
+    slack_renderer = SlackRenderer()
+    result = slack_renderer.get_failed_expectation_domain(
+        "expect_column_pair_to_be_equal", {"column": "first"}
+    )
+    assert result == "first"
+
+
+def test_SlackRenderer_get_failed_expectation_domain_column():
```
talagluck: Looks like a test name collision here.

great_expectations/render/renderer/slack_renderer.py
```diff
@@ -175,3 +194,42 @@ def _get_report_element(self, docs_link):
...
+        for expectation in validation_results:
+            if not expectation["success"]:
+                expectation_name = expectation["expectation_config"]["expectation_type"]
+                expectation_kwargs = expectation["expectation_config"]["kwargs"]
+                failed_expectations_str += self.create_failed_expectation_text(
+                    expectation_kwargs, expectation_name
+                )
+        return failed_expectations_str
+
+    def create_failed_expectation_text(self, expectation_kwargs, expectation_name):
```
talagluck: Can we add a return annotation here?

talagluck (commented):
Getting there! Just a couple of final changes. Thanks!

great_expectations/render/renderer/slack_renderer.py
itaise: thanks, done

tests/render/test_SlackRenderer.py
itaise: thanks, done.

talagluck (approved):
LGTM - thanks for the contribution!

PR #5818, opened 2022-08-23
[FEATURE] Improve slack error condition

itaise:
Hi,
I tried triggering a slack notification from GE with a bad token, and didn't get any error. 
That is because slack api returned status code 200, with an error description in the response body.

- add to the check on status code, a check if "ok" field in the response is True. 

When running again with this PR, I got the message, as expected :) 
`Request to Slack webhook returned error 200: {"ok":false,"error":"invalid_auth","warning":"missing_charset","response_metadata":{"warnings":["missing_charset"]}}`

Thank you!


talagluck (commented):
Thanks for this, @itaise! To confirm, will a successful response code for a Slack message always have an "ok" message? If that is the case, then once you update the tests for this, this should be set to approve and merge!

itaise:
Thank you @talagluck, that was a good point - apparently the OK response is different when using webhook vs. token - so I updated the condition for both cases and tested it. also changed the test - so that the Mock will have a json method and appropriate text in it.  
WDYT? thanks!

talagluck (approved):
LGTM - thanks for the update and contribution!

PR #5823, opened 2022-08-23
[Maintenance] Clean up build_parameter_container_for_variables

billdirks:
This removes some unneeded string manipulation and makes control flow more direct. I also remove an unneeded call to `convert_dictionary_to_parameter_node`.

Timing:
Previously running on my test set took between 37s on average over 8 runs (range: 33-48s).
Now 35s on average (range: 32-41s)

Details:
Previously `build_parameter_container_for_variables` would append the string `$variables.` to the variable keys which would then call `build_parameter_container` which parses the string and immediately removes the `$variable` prefix before calling `_build_parameter_node_tree_for_one_parameter`. This PR removes the string manipulation and directly calls `_build_parameter_node_tree_for_one_parameter` from `build_parameter_container_for_variables`.

### Definition of Done

Thank you for submitting!


great_expectations/rule_based_profiler/parameter_container.py
```diff
@@ -274,20 +269,24 @@ def build_parameter_container_for_variables(
...
-    parameter_container = ParameterContainer(parameter_nodes=None)
-    build_parameter_container(
-        parameter_container=parameter_container, parameter_values=parameter_values
+    parameter_node = ParameterNode()
+    key: str
+    value: Any
+    for key, value in variables_configs.items():
+        key_parts: List[str] = [PARAMETER_NAME_ROOT_FOR_VARIABLES] + key.split(
+            FULLY_QUALIFIED_PARAMETER_NAME_SEPARATOR_CHARACTER
+        )
```
Kilo59, line 278: I totally forgot you could do this with lists 😄 (add them that is). 

Similar to `.extend()` ... but better (it returns the new list instead of modifying in-place)
```python
my_list = [PARAMETER_NAME_ROOT_FOR_VARIABLES]
my_list.extend(key.split(
            FULLY_QUALIFIED_PARAMETER_NAME_SEPARATOR_CHARACTER
        ))
```

Kilo59 (approved):

great_expectations/rule_based_profiler/parameter_container.py
billdirks: I actually timed both and the `+` was faster. Maybe it can set aside the space upfront and doesn't temporarily make the initial list?

Kilo59: Yeah, I'm going to start using this more.

Shinnnyshinshin (approved):
More winning :) LGTM

great_expectations/rule_based_profiler/parameter_container.py
alexsherstinsky: Interesting... I have been avoiding the "+" on lists because I was under the impression that allocating a new list would take more time.  Good find, @billdirks !  Thanks!

great_expectations/rule_based_profiler/parameter_container.py
```diff
@@ -274,20 +269,24 @@ def build_parameter_container_for_variables(
...
+        _build_parameter_node_tree_for_one_parameter(parameter_node, key_parts, value)
+    # We only want to set the ParameterContainer key PARAMETER_NAME_ROOT_FOR_VARIABLES if
+    # parameter_node is non-empty since there is downstream logic that depends on this.
+    return (
```
alexsherstinsky, line 282: ❤️

PR #5826, opened 2022-08-24
[MAINTENANCE] Update `list_` methods on `DataContext` to emit names along with object ids

cdkini:
- Cloud-enabled contexts print out ids instead of object names; this makes it difficult to know which object to reference in config.
- The changes herein ensure that expectation suites and checkpoints are printed with their name.

Loom: https://www.loom.com/share/245cf3753e954eeaa4606f16fad2ae89

### Definition of Done

### DIFF
--- a/great_expectations/data_context/store/ge_cloud_store_backend.py
+++ b/great_expectations/data_context/store/ge_cloud_store_backend.py
@@ -430,16 +430,28 @@ class GeCloudStoreBackend(StoreBackend, metaclass=ABCMeta):
             f"{self.ge_cloud_credentials['organization_id']}/"
             f"{hyphen(self.ge_cloud_resource_name)}",
         )
+        resource_type = self.ge_cloud_resource_type
+        attributes_key = self.PAYLOAD_ATTRIBUTES_KEYS[resource_type]
+
         try:
             response = requests.get(url, headers=self.headers, timeout=self.TIMEOUT)
             response_json = response.json()
-            keys = [
-                (
-                    self.ge_cloud_resource_type,
-                    resource["id"],
+
+            keys = []
+            for resource in response_json["data"]:
+                id: Optional[str] = resource.get("id")
+                resource_name: Optional[str] = None
+                attr: Optional[dict] = resource.get("attributes", {}).get(
+                    attributes_key
                 )
-                for resource in response_json.get("data")
-            ]
+                if attr:
+                    resource_name = attr.get("name") or attr.get(
+                        "expectation_suite_name"
+                    )
+
+                key = (resource_type, id, resource_name)
+                keys.append(key)
+
             return keys
         except Exception as e:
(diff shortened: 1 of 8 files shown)


great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -423,23 +423,40 @@ def ge_cloud_resource_type(self) -> GeCloudRESTResource:
...
+        attributes_key = self.PAYLOAD_ATTRIBUTES_KEYS[resource_type]
+
         try:
             response = requests.get(url, headers=self.headers, timeout=self.TIMEOUT)
             response_json = response.json()
-            keys = [
-                (
-                    self.ge_cloud_resource_type,
-                    resource["id"],
+
```
Kilo59, line 440: Can we `.raise_for_status()` here?
```python
  response = requests.get(url, headers=self.headers, timeout=self.TIMEOUT)
  response.raise_for_status()
  response_json = response.json()

```

great_expectations/data_context/types/resource_identifiers.py
```diff
@@ -253,7 +253,10 @@ def from_fixed_length_tuple(cls, tuple_):
         return cls.from_tuple(tuple_)
 
     def __repr__(self):
-        return f"{self.__class__.__name__}::{self.resource_type}::{self.ge_cloud_id}"
+        repr = f"{self.__class__.__name__}::{self.resource_type}::{self.ge_cloud_id}"
+        if self.resource_name:
+            repr += f"::{self.resource_name}"
+        return repr
```
Kilo59, line 259: `non-blocking`
If this is already a well-established pattern then ignore this but...

One suggestion I've heard in the past regarding how to define a `__repr__` is to make it possible to re-create the object using the `repr()` output (for example by pasting it into the REPL).

So for example...
```python
class Car:
    def __init__(self, color, mileage):
        self.color = color
        self.mileage = mileage

    def __repr__(self):
       return (f'{self.__class__.__name__}('
               f'{self.color!r}, {self.mileage!r})')

    def __str__(self):
        return f'a {self.color} car'

my_car = Car('blue', 500)
print(repr(my_car))
print(my_car)

... Car('blue', 500)
... a blue car
```


https://dbader.org/blog/python-repr-vs-str

great_expectations/data_context/store/ge_cloud_store_backend.py
cdkini: done!

great_expectations/data_context/types/resource_identifiers.py
cdkini: ooh I like that! I think we'll probably want to keep the reprs as-is for now due to the fact that we don't consistently do this

Kilo59 (approved):

PR #5830, opened 2022-08-25
[DO NOT MERGE][BUGFIX] Bugs from GX Cloud Checkpoints Demo

roblim:
***NB: changes aren't actual code change suggestions, simply what I had to do to win exception whack-a-mole to get things working for a demo.**

Using this PR to capture bugs/unexpected behavior uncovered in preparation for the Nimbus demo of GX Cloud Checkpoints and any commentary. An archive of the notebook used during the demo and supporting assets (suite json and csv to run Checkpoint against) is available - LMK if you'd like it.

This PR's delta comprises all the changes I needed to apply to OSS GX to get the flow working E2E. Some might be related to actual bugs and others might just be related to planned work that hasn't happened yet. Inline commentary to follow. 

cc: @Kilo59 @cdkini @wookasz @joshua-stauffer

### DIFF
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -815,7 +815,8 @@ class StoreValidationResultAction(ValidationAction):
 
         checkpoint_ge_cloud_id = None
         if self.data_context.ge_cloud_mode and checkpoint_identifier:
-            checkpoint_ge_cloud_id = checkpoint_identifier.ge_cloud_id
+            checkpoint_ge_cloud_id = checkpoint_identifier if isinstance(checkpoint_identifier, str) else \
+                checkpoint_identifier.ge_cloud_id
 
         expectation_suite_ge_cloud_id = None
         if self.data_context.ge_cloud_mode and expectation_suite_identifier:
--- a/great_expectations/checkpoint/util.py
+++ b/great_expectations/checkpoint/util.py
@@ -245,9 +245,11 @@ def get_substituted_batch_request(
     for key, value in validation_batch_request.items():
         substituted_value = substituted_runtime_batch_request.get(key)
         if value is not None and substituted_value is not None:
-            raise ge_exceptions.CheckpointError(
-                f'BatchRequest attribute "{key}" was specified in both validation and top-level CheckpointConfig.'
-            )
+            validation_batch_request[key] = substituted_value
+            substituted_runtime_batch_request.pop(key)
+            # raise ge_exceptions.CheckpointError(
+            #     f'BatchRequest attribute "{key}" was specified in both validation and top-level CheckpointConfig.'
+            # )
 
     effective_batch_request: dict = dict(
         **substituted_runtime_batch_request, **validation_batch_request
--- a/great_expectations/data_context/store/checkpoint_store.py
+++ b/great_expectations/data_context/store/checkpoint_store.py
@@ -195,7 +195,7 @@ class CheckpointStore(ConfigurationStore):
         # Make two separate requests to set and get in order to obtain any additional
         # values that may have been added to the config by the StoreBackend (i.e. object ids)
         ref: Optional[GeCloudResourceRef] = self.set(key, checkpoint_config)  # type: ignore[func-returns-value]
-        if ref:
+        if ref and not isinstance(ref, bool):
             assert isinstance(key, GeCloudIdentifier)
             key.ge_cloud_id = ref.ge_cloud_id
 
--- a/great_expectations/data_context/store/store.py
+++ b/great_expectations/data_context/store/store.py
@@ -194,8 +194,8 @@ class Store:
 
     def _build_key_from_config(self, config: AbstractConfig) -> DataContextKey:
         id_: Optional[str] = None
-        if hasattr(config, "id_"):
-            id_ = config.id_
+        if hasattr(config, "ge_cloud_id"):
+            id_ = config.ge_cloud_id
         name: Optional[str] = None
         if hasattr(config, "name"):
             name = config.name
--- a/great_expectations/self_check/util.py
+++ b/great_expectations/self_check/util.py
@@ -346,14 +346,12 @@ try:
         "DECIMAL": redshifttypes.DECIMAL,
         "DOUBLE_PRECISION": redshifttypes.DOUBLE_PRECISION,
         "FOREIGN_KEY_RE": redshifttypes.FOREIGN_KEY_RE,
-        "GEOMETRY": redshifttypes.GEOMETRY,
         "INTEGER": redshifttypes.INTEGER,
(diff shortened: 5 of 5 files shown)


great_expectations/checkpoint/actions.py
```diff
@@ -815,7 +815,8 @@ def _run(
 
         checkpoint_ge_cloud_id = None
         if self.data_context.ge_cloud_mode and checkpoint_identifier:
-            checkpoint_ge_cloud_id = checkpoint_identifier.ge_cloud_id
+            checkpoint_ge_cloud_id = checkpoint_identifier if isinstance(checkpoint_identifier, str) else \
+                checkpoint_identifier.ge_cloud_id
```
roblim, line 819: This patch was to address an exception that was being thrown only when updating an existing Checkpoint, using `context.add_checkpoint` (see update checkpoint notebook cell).

For some reason, `checkpoint_identifier` was coming back as a `str` only when updating. NB - I didn't actually confirm that the `str` `checkpoint_identifier` was actually the `ge_cloud_id`.

great_expectations/checkpoint/util.py
```diff
@@ -245,9 +245,11 @@ def get_substituted_batch_request(
...
         substituted_value = substituted_runtime_batch_request.get(key)
         if value is not None and substituted_value is not None:
-            raise ge_exceptions.CheckpointError(
-                f'BatchRequest attribute "{key}" was specified in both validation and top-level CheckpointConfig.'
-            )
+            validation_batch_request[key] = substituted_value
+            substituted_runtime_batch_request.pop(key)
+            # raise ge_exceptions.CheckpointError(
+            #     f'BatchRequest attribute "{key}" was specified in both validation and top-level CheckpointConfig.'
+            # )
```
roblim, line 252: This was to address a conflict between `mercury`,  how GX does final substitutions to yield a complete validation config at runtime, and the signature of `RuntimeBatchRequest`. Exception got thrown when attempting to run the checkpoint using a runtime datasource: 

```
batch_request = RuntimeBatchRequest(
            data_connector_name="default_runtime_data_connector",
            datasource_name="Residential Data",
            data_asset_name="raw-residential-data",
            runtime_parameters={
                "batch_data": df_success
            },
            batch_identifiers={
                "color": "blue"
            }
        )

        context.run_checkpoint(
            ge_cloud_id=checkpoint_id,
            batch_request=batch_request,
            run_name=f"{i}-run-run-run"
        )
```

Quoted from[ comment in different PR](https://github.com/superconductive/mercury/pull/1083#discussion_r954371365):

" There's a bit of a conflict here - in mercury, if a batch_request appears in validations, it has to have complete datasource_name/data_asset_name/data_connector_name triple. Likewise, datasource_name and data_connector_name are required positional args for RuntimeBatchRequest.

BUT, when you satisfy those two constraints, it breaks the logic surrounding checkpoint config substitution at runtime WRT to batch requests - it can only be an exclusive union (is this the right term?). An attr can't appear in both a batch request passed in at runtime and in batch_request at top-level, or within validations.

My instinct is that we should probably review some of that update precedence logic in OSS GX and also make sure there aren't other potential places where mercury conflicts. cc: @cdkini @Kilo59"

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -195,7 +195,7 @@ def create(self, checkpoint_config: CheckpointConfig) -> Optional[DataContextKey
         # Make two separate requests to set and get in order to obtain any additional
         # values that may have been added to the config by the StoreBackend (i.e. object ids)
         ref: Optional[GeCloudResourceRef] = self.set(key, checkpoint_config)  # type: ignore[func-returns-value]
-        if ref:
+        if ref and not isinstance(ref, bool):
```
roblim, line 198: This patch was to address an exception that was being thrown only when updating an existing Checkpoint, using `context.add_checkpoint` (see update checkpoint notebook cell).

For some reason, `ref` was coming back as a `bool` only when updating.

great_expectations/self_check/util.py
```diff
@@ -346,14 +346,12 @@
...
         "DOUBLE_PRECISION": redshifttypes.DOUBLE_PRECISION,
         "FOREIGN_KEY_RE": redshifttypes.FOREIGN_KEY_RE,
-        "GEOMETRY": redshifttypes.GEOMETRY,
         "INTEGER": redshifttypes.INTEGER,
         "PRIMARY_KEY_RE": redshifttypes.PRIMARY_KEY_RE,
         "REAL": redshifttypes.REAL,
         "SMALLINT": redshifttypes.SMALLINT,
         "TIMESTAMP": redshifttypes.TIMESTAMP,
         "TIMESTAMPTZ": redshifttypes.TIMESTAMPTZ,
-        "TIMETZ": redshifttypes.TIMETZ,
```
roblim, line 356: This might just be my local env and dependency snafus, but couldn't escape `AttributeError`s here from recently added keys.

great_expectations/data_context/store/store.py
```diff
@@ -194,8 +194,8 @@ def self_check(self, pretty_print: bool) -> None:
 
     def _build_key_from_config(self, config: AbstractConfig) -> DataContextKey:
         id_: Optional[str] = None
-        if hasattr(config, "id_"):
-            id_ = config.id_
+        if hasattr(config, "ge_cloud_id"):
+            id_ = config.ge_cloud_id
```
roblim, line 198: I think this might actually be a `mercury` bug from lagging behind the `ge_cloud_id` -> `id_` transition in OSS GX. Not sure if it's easier/desirable to add this bit here temporarily or just catch up `mercury`. In any event, the call that led to this exception was `context.add_checkpoint`, which still takes `ge_cloud_id` - should prob update?

great_expectations/checkpoint/actions.py
wookasz: We should add a `save_checkpoint` instead. I would expect `add_checkpoint` to fail completely if the checkpoint already exists. We should use `save_checkpoint` for existing checkpoints

edited: I asked an unclear question first time around.

great_expectations/checkpoint/util.py
wookasz: I don't think substituting is the right thing to do here. The runtime batch request is really not a batch request, it's just the runtime params + batch identifiers. The workaround is to not use a `RuntimeBatchRequest` object here

```python
batch_request = {
    "runtime_parameters": {
        "batch_data": df_success
    },
    "batch_identifiers": {
        "color": "blue"
    },
}
checkpoint.run(batch_request=batch_request)
```

great_expectations/data_context/store/checkpoint_store.py
wookasz: same as comment above

PR #5836, opened 2022-08-25
[MAINTENANCE] Update `ExpectationSuite` CRUD on `DataContext` to recognize Cloud ids

cdkini:
- Update suite CRUD to ensure that ids propagate downstream from the Cloud backend.
- Write a number of unit tests to validate behavior.

### Definition of Done


great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -183,23 +183,24 @@ def create_expectation_suite(
...
-        ):
-            raise ge_exceptions.DataContextError(
-                f"expectation_suite with GE Cloud ID {ge_cloud_id} already exists. If you would like to overwrite this "
-                "expectation_suite, set overwrite_existing=True."
-            )
-        self.expectations_store.set(key, expectation_suite, **kwargs)
+
+        response = self.expectations_store.set(key, expectation_suite, **kwargs)
+        if isinstance(response, GeCloudResourceRef):
+            expectation_suite.ge_cloud_id = response.ge_cloud_id
```
cdkini: This logic needs to be moved upstream to the stores but that will come as part of a separate effort

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -246,7 +247,7 @@ def get_expectation_suite(
         )
         if not self.expectations_store.has_key(key):  # noqa: W601
             raise ge_exceptions.DataContextError(
-                f"expectation_suite {expectation_suite_name} not found"
+                f"expectation_suite with id {ge_cloud_id} not found"
```
cdkini, line 250: `has_key` checks always use ID while `name in self.list_expectation_suite_names()` should reference name

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -273,21 +274,15 @@ def save_expectation_suite(
...
-            self.expectations_store.has_key(key)  # noqa: W601
-            and not overwrite_existing
-        ):
-            raise ge_exceptions.DataContextError(
-                f"expectation_suite with GE Cloud ID {ge_cloud_id} already exists. "
-                f"If you would like to overwrite this expectation_suite, set overwrite_existing=True."
-            )
+
+        if not overwrite_existing:
+            self._validate_suite_unique_constaints_before_save(key)
```
cdkini, line 284: Since save works with both new suites and existing ones, we want to be able to ensure that we're respecting the unique constraint on both ids and names. If using a new suite, we want to check against the name and if using an existing one, we want to check against the ID.

Let me know if that makes sense!

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -522,11 +522,11 @@ def remove_key(self, key):
...
-        # self.list_keys() generates a list of length 2 tuples
-        if len(key) == 3:
-            key = key[:2]
-        all_keys = self.list_keys()
-        return key in all_keys
+        # Due to list_keys being inconsistently sized (due to the possible of resource names),
+        # we solely rely on the equality of ids to determine the result of this method
+        id = key[1]
+        all_ids = set(map(lambda key: key[1], self.list_keys()))
+        return id in all_ids
```
cdkini: This feels iffy but due to the change I made last week around `list_expectation_suite_names()` and `list_expectation_suites()`, this method actually broke.

Since all we're really doing is looking to see if we have a key with a given ID, I've updated this method. Let me know if you think we should tackle this in a different manner. We're not actually asserting that the exact key is present but rather its contents (if that makes sense).

great_expectations/data_context/data_context/cloud_data_context.py
anthonyburdi, line 250: I think this error message should also show the `expectation_suite_name` since that is an optional parameter as well and may be more meaningful for the user.

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -183,23 +183,24 @@ def create_expectation_suite(
...
-            self.expectations_store.has_key(key)  # noqa: W601
-            and not overwrite_existing
-        ):
-            raise ge_exceptions.DataContextError(
-                f"expectation_suite with GE Cloud ID {ge_cloud_id} already exists. If you would like to overwrite this "
-                "expectation_suite, set overwrite_existing=True."
-            )
-        self.expectations_store.set(key, expectation_suite, **kwargs)
+
+        response = self.expectations_store.set(key, expectation_suite, **kwargs)
```
Kilo59: Can we add an annotation for the possible types?

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -273,21 +274,15 @@ def save_expectation_suite(
         Returns:
             None
         """
-        ge_cloud_id = expectation_suite.ge_cloud_id
         key = GeCloudIdentifier(
             resource_type=GeCloudRESTResource.EXPECTATION_SUITE,
-            ge_cloud_id=ge_cloud_id
-            if ge_cloud_id is not None
-            else str(expectation_suite.ge_cloud_id),
```
Kilo59, line 281: 👍

great_expectations/data_context/store/ge_cloud_store_backend.py
Kilo59: What are the possible tuple sizes (and values) here?

This could be a good case for a `@overload`
https://mypy.readthedocs.io/en/stable/more_types.html?highlight=overload#function-overloading
I'm happy to create the overload.

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -296,7 +291,30 @@ def save_expectation_suite(
         )
         if include_rendered_content:
             expectation_suite.render()
-        self.expectations_store.set(key, expectation_suite, **kwargs)
+
+        response = self.expectations_store.set(key, expectation_suite, **kwargs)  # type: ignore[func-returns-value]
+        if isinstance(response, GeCloudResourceRef):
+            expectation_suite.ge_cloud_id = response.ge_cloud_id
```
anthonyburdi, line 297: is this expectation_suite variable used after the set call?

great_expectations/data_context/store/ge_cloud_store_backend.py
anthonyburdi: it looks like we were comparing both key[0] and key[1] before, what do we lose by only comparing to key[1] now?

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
```diff
@@ -95,6 +98,68 @@ def mock_get_all_suites_json(
...
+
+    return _mocked_post_response
+
+
+@pytest.fixture
+def mocked_get_response(
+    mock_response_factory: Callable,
+    suite_names_and_ids: Tuple[Tuple[str, str], Tuple[str, str]],
+) -> Callable[[], MockResponse]:
+    suite_id = suite_names_and_ids[0][1]
```
anthonyburdi: How about just passing in `suite_id` here?

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -1345,13 +1345,13 @@ def create_expectation_suite(
         """
         See `AbstractDataContext.create_expectation_suite` for more information.
         """
-        res = self._data_context.create_expectation_suite(  # type: ignore[union-attr]
+        suite = self._data_context.create_expectation_suite(  # type: ignore[union-attr]
```
Shinnnyshinshin, line 1348: ❤️

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
```diff
@@ -134,3 +199,264 @@ def test_list_expectation_suites(
...
+    empty_base_data_context_in_cloud_mode: BaseDataContext,
+    suite_names_and_ids: Tuple[Tuple[str, str], Tuple[str, str]],
+) -> None:
+    context = empty_base_data_context_in_cloud_mode
+
+    suite_name, suite_id = suite_names_and_ids[0]
+    suite = ExpectationSuite(suite_name, ge_cloud_id=suite_id)
+
+    with mock.patch(
+        "requests.put", autospec=True, return_value=mock.Mock(status_code=405)
```
anthonyburdi, line 406: Are we expecting a 405 for put?

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -296,7 +291,29 @@ def save_expectation_suite(
...
+
+    def _validate_suite_unique_constaints_before_save(
+        self, key: GeCloudIdentifier
+    ) -> None:
```
Shinnnyshinshin, line 301: ❤️

Shinnnyshinshin (approved):
Love it. Thank you @cdkini

anthonyburdi (commented):
Looks great, love the comprehensive tests. Left a few comments and suggestions.

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -522,10 +522,16 @@ def remove_key(self, key):
...
+        # Due to list_keys being inconsistently sized (due to the possible of resource names),
+        # we remove any resource names and assert against key ids.
+
+        def _shorten_key(key) -> Tuple[str, str]:
+            if len(key) > 2:
+                key = key[:2]
+            return key
+
+        key = _shorten_key(key)
+        all_keys = set(map(lambda key: _shorten_key(key), self.list_keys()))
```
Kilo59: What's behind the lambda? Can you just map `_shorten_key` directly?

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
```diff
@@ -134,3 +199,264 @@ def test_list_expectation_suites(
...
+def test_create_expectation_suite_namespace_collision_raises_error(
+    empty_base_data_context_in_cloud_mode: BaseDataContext,
+) -> None:
+    context = empty_base_data_context_in_cloud_mode
+
+    suite_name = "my_suite"
+    existing_suite_names = [suite_name]
+    with mock.patch(
+        "great_expectations.data_context.data_context.cloud_data_context.CloudDataContext.list_expectation_suite_names",
+        return_value=existing_suite_names,
```
Kilo59, line 272: `nicetohave`
Could you move these repeated mocks into a well-named fixture?

Same for the HTTP patches, maybe some like this `mock_http_unavailable` fixture.
https://github.com/great-expectations/great_expectations/blob/develop/tests/data_context/conftest.py#L702-L727

great_expectations/data_context/data_context/cloud_data_context.py
cdkini: So we never actually use `expectation_suite_name` in this function (it is only in the signature for parity across the inheritance hierarchy). As we don't currently support get by name, do we still want this update?

great_expectations/data_context/data_context/cloud_data_context.py
cdkini: I was going to return it but since that violates the signature and behavior set throughout the inheritance hierarchy, I opted to update the object in-place. The tests ensure that we have coverage over the id updating the input object.

great_expectations/data_context/store/ge_cloud_store_backend.py
cdkini: ah very true haha I originally used a lambda to parse out `key[1]` but you are right!

Kilo59 (approved):
Pending my question about the use of `lambda`.

I'd like to see some updates on how we use mock, but consider that non-blocking. 😄

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
cdkini: update fixtures so we don't have to parse this out each time

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
cdkini: sure! mind if I do that in a follow-up PR?

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
cdkini: yup! We don't have the PUT endpoint established yet so we hit a 405 and then use the legacy PATCH behavior - this will eventually be removed but as long as we have this fork, we need this logic here.

great_expectations/data_context/store/ge_cloud_store_backend.py
cdkini: `key[0]` should be consistent but I don't think it hurts to compare (especially since we have slight deviations with things like contracts vs checkpoints)

great_expectations/data_context/data_context/cloud_data_context.py
anthonyburdi: You're right - I was thinking checkpoints but the expectation suite id's will be populated on checkpoint create alongside the names so I think we are good. Thanks for clarifying.

great_expectations/data_context/data_context/cloud_data_context.py
anthonyburdi: Makes sense. Thanks for clarifying!

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
anthonyburdi: Thanks for clarifying!

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
Kilo59: Typed `NamedTuples` can make good fixture return objects, and then the type and attribute name can be expressive.
`my_fixture.suite_1.id`
or
`my_fixture[0].name`

anthonyburdi (approved):
Thanks for the clarifications and updates! 🙇

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
cdkini: good call! will update in that subsequent PR to clean this all up

PR #5844, opened 2022-08-25
[MAINTENANCE] Add additional unit tests around `Datasource` caching

cdkini:
- More unit tests around `Datasource` caching (this time using `DataContext` to text persistence and store involvement)

### Definition of Done


tests/data_context/test_data_context_datasources.py
```diff
@@ -58,34 +59,146 @@ def test_data_context_instantiates_inline_store_backend_with_filesystem_config(
...
+    data_context_fixture_name: str,
+    request,
+) -> None:
+    """
+    What does this test and why?
+
+    For persistence-enabled contexts, we should leverage the underlying store in the case
+    of a cache miss.
+
+    For persistence-disabled contexts, we only look at the cache.
```
anthonyburdi: Is this true? I think we still look at the store but it could be an in-memory store

tests/data_context/test_data_context_datasources.py
```diff
@@ -58,34 +59,146 @@ def test_data_context_instantiates_inline_store_backend_with_filesystem_config(
...
+    ) as mock_set:
+        context.add_datasource(
+            name,
+            class_name="PandasDatasource",
+            batch_kwargs_generators={
+                "subdir_reader": {
+                    "class_name": "SubdirReaderBatchKwargsGenerator",
+                }
+            },
+        )
```
anthonyburdi: Maybe two tests? Label this one with the v2_api flag and another with a v3 api datasource?

tests/data_context/test_data_context_datasources.py
```diff
@@ -140,51 +276,109 @@ def test_delete_datasource_updates_cache(
...
+    assert name in context.datasources
 
-    assert not mock_get.called
+    # Ensure that our cache value is updated to reflect changes
+    context.update_datasource(datasource)
+
+    assert name in context.datasources
+
+    retrieved_datasource = context.get_datasource(datasource_name=name)
+    assert retrieved_datasource.boto3_options == boto3_options
```
anthonyburdi: I think we should also check in the cache before calling context.get_datasource() e.g. check context.datasources for your datasource with the right boto3 option

tests/data_context/test_data_context_datasources.py
```diff
@@ -140,51 +276,109 @@ def test_delete_datasource_updates_cache(
...
+    What does this test and why?
+
+    For persistence-disabled contexts, we should only update the cache upon
+    using update to create a new datasource.
+    """
     context = in_memory_runtime_context
 
-    name = "my_fake_datasource_name"
+    name = "my_new_datasource"
+    boto3_options = {"my": "option"}
```
anthonyburdi: Nit but you can skip the boto3_options in this test

anthonyburdi (changes requested):
Looks great! Thanks for doing this. Just a few small tweaks.

tests/data_context/test_data_context_datasources.py
cdkini: nope you are correct! Updated this docstring

tests/data_context/test_data_context_datasources.py
cdkini: since we're moreso testing the persistence and not the actual datasource, could I just use a v3 datasource instead? Don't think we necessarily need two tests for v2 vs v3 when the underlying behavior is the same in regards to caching.

tests/data_context/test_data_context_datasources.py
cdkini: fixed!

tests/data_context/test_data_context_datasources.py
cdkini: ended up using the same v3 config for all tests - let me know if you disagree and I can come back 😄

tests/data_context/test_data_context_datasources.py
cdkini: I agree! Done ✅

anthonyburdi (approved):
Thanks for making those changes!

PR #5853, opened 2022-08-26
[FEATURE] Add `include_rendered_content` to `get_expectation_suite` and `get_validation_result`

NathanFarmer:
- Add `include_rendered_content` to `context.get_expectation_suite`
- Add `include_rendered_content` to `context.get_validation_result`
- Enable rendering an `ExpectationConfiguration` not contained in an `ExpectationSuite`
- Enable rendering in `Checkpoint` using `DataContext` variables
- Simplify `InlineRenderer.get_rendered_content` methods now that `ExpectationSuite`s will always be saved or retrieved with `rendered_content`

### Definition of Done


great_expectations/core/expectation_configuration.py
```diff
@@ -1425,6 +1430,30 @@ def get_domain_type(self) -> MetricDomainTypes:
             'Unable to determine "domain_type" of this "ExpectationConfiguration" object from "kwargs" and heuristics.'
         )
 
+    def render(self) -> None:
+        """
+        Renders content using the atomic prescriptive renderer for this Expectation Configuration to
+            ExpectationConfiguration.rendered_content.
+        """
+        inline_renderer_config: "InlineRendererConfig" = {  # type: ignore[assignment]
```
Kilo59, line 1432: Why are we declaring this an `InlineRendererConfig` if it isn't one?
Should this be... 
```python
inline_renderer_config: Union["InlineRendererConfig", dict]
```

Or 

```python
inline_renderer_config: Union["InlineRendererConfig", SomeTypedDict]
```

?

NathanFarmer: Hi @Kilo59! If you recall in my last rendering PR we went through the exercise of creating this `TypedDict` called `InlineRendererConfig` for the purposes of type annotation in these `render()` methods. See: https://github.com/great-expectations/great_expectations/pull/5726#discussion_r947124333

Kilo59: That's what I had thought but the assignment error threw me off.
Does this dictionary satisfy the TypeDict definition? 

```python
class InlineRendererConfig(TypedDict):
    class_name: str
    render_object: Union[ExpectationConfiguration, ExpectationValidationResult]
```

## Edit
So this looks like it's an issue with `mypy`. I had thought that `TypeDicts` were structural types, so an equivalent dict should be compatible, but it's not 😢 .
https://github.com/python/mypy/issues/4976

So I think the solution (if we want to remove the ignore comment) would be to...
1. instantiate from the `InlineRendererConfig(class_name=foo, render_object=bar)`. But this means importing `InlineRendererConfig` normally which we don't want to do due to cyclic imports.
2. use `cast()` - see https://github.com/great-expectations/great_expectations/pull/5853#discussion_r958485483

Not super happy with solution 2 though, since the type checker won't raise errors if the keys/values are wrong.
I think maybe it's best to leave in the ignore comment until we actually import the type without `if TYPE_CHECKING` and then we can go with solution `# 1`.

great_expectations/core/expectation_configuration.py
```diff
@@ -1419,6 +1424,30 @@ def get_domain_type(self) -> MetricDomainTypes:
...
+            "class_name": "InlineRenderer",
+            "render_object": self,
+        }
```
Kilo59, line 1435: Also need to `from typing import cast`

```suggestion
        inline_renderer_config = cast("InlineRendererConfig", {
            "class_name": "InlineRenderer",
            "render_object": self,
        })
```

## Edit:
On second thought, let's not do this, using cast with this totally defeats the purpose of the type in the first place.

great_expectations/core/expectation_configuration.py
```diff
@@ -1419,6 +1424,30 @@ def get_domain_type(self) -> MetricDomainTypes:
...
+        module_name = "great_expectations.render.renderer.inline_renderer"
+        inline_renderer = instantiate_class_from_config(
```
anthonyburdi, line 1437: Could we just instantiate from the constructor e.g. `inline_renderer = InlineRenderer(render_object=self)`? Or is this a point of extensibility where we would accept a config?

Kilo59 (approved):

great_expectations/core/expectation_configuration.py
NathanFarmer: Leaving as-is per our other conversation. Using `cast` will not be type-safe. Ignoring the assignment error will keep this on our radar.

anthonyburdi (commented):
Looks good, just one question

great_expectations/core/expectation_configuration.py
NathanFarmer: This is the issue with circular imports that we spent a lot of time trying to solve about a month ago and landed on this as a solution for now.

anthonyburdi: Would you mind adding a small comment here in the code with that explanation? That way if we fix the circular import issue in the future we will have a hint that we can safely remove the `instantiate_class_from_config()` call

PR #5856, opened 2022-08-28
[MAINTENANCE] Coveralls for unit-tests

Kilo59:
- publish unit-test coverage report to coveralls.io
- https://coveralls.io/github/great-expectations/great_expectations

## TODO

- in Azure
  - [x] - add `COVERALLS_REPO_TOKEN` variable to the Azure pipeline
  - [x] - pull token from the variable group
  - [ ] - don't block merge if coveralls.io is down
- in Coveralls.io
  - [x] - set `COVERAGE THRESHOLD FOR FAILURE` to `40%`
  - [x] - set `COVERAGE DECREASE THRESHOLD FOR FAILURE` to `0.5%`
  - [x] setup PR gating on the above threshold failures

image


azure-pipelines-dependency-graph-testing.yml
```diff
@@ -245,13 +245,23 @@ stages:
...
 
+          - script: |
+              pip install coveralls
+              export CIRCLE_BRANCH=$BUILD_SOURCEBRANCH
+              coveralls
+            displayName: 'coveralls'
+            env:
+              CIRCLECI: 1
+              CIRCLE_BUILD_NUM: $(Build.BuildNumber)
+              COVERALLS_REPO_TOKEN: $(coveralls_repo_token)
```
Kilo59: We are mimicking circleci here
https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#azure-pipelines-support
https://github.com/TheKevJames/coveralls-python/issues/245

>As per [#245](https://github.com/TheKevJames/coveralls-python/issues/245), our users suggest leaving “keep this value secret” unchecked – this may be secure enough as-is, in that a user making a PR cannot access this variable.

Someone with maintainer access to `great_expectations` needs to grab the `repo_token` from the settings page so that we can add it to our azure pipeline.
https://coveralls.io/github/great-expectations/great_expectations

Screen Shot 2022-08-29 at 9 30 34 AM

azure-pipelines-dependency-graph-testing.yml
```diff
@@ -245,13 +245,23 @@ stages:
...
 
+          - script: |
+              pip install coveralls
+              export CIRCLE_BRANCH=$BUILD_SOURCEBRANCH
+              coveralls
+            displayName: 'coveralls'
+            env:
+              CIRCLECI: 1
+              CIRCLE_BUILD_NUM: $(Build.BuildNumber)
+              COVERALLS_REPO_TOKEN: $(common_secrets.coveralls)
```
Kilo59: The `coveralls` packages takes care of pulling the `COVERALLS_REPO_TOKEN` from the environment.
Need to verify that this is the correct way to pull values from variable groups.

Also, it's possible this is correct but that it being secret is causing issues? See the note below.

>As per [#245](https://github.com/TheKevJames/coveralls-python/issues/245), our users suggest leaving “keep this value secret” unchecked – this may be secure enough as-is, in that a user making a PR cannot access this variable.

## Note
The current coveralls api 422  error is the same error we get without passing any token.

azure-pipelines-dependency-graph-testing.yml
```diff
@@ -180,7 +180,8 @@ stages:
       - job: compatibility_matrix
         condition: eq(stageDependencies.scope_check.changes.outputs['CheckChanges.GEChanged'], true)
         variables:
-          GE_pytest_opts: '--no-sqlalchemy'
+          - group: 'common_secrets'
+          # GE_pytest_opts: '--no-sqlalchemy'  # TODO: resolve using this with `- group`
```
Kilo59, line 184: I commented this `GE_pytest_opts` key in order to get the `coveralls` step working (which it now does 🎊 ), but I assume we need to add this back.

One solution could be to set the `common_secrets` available at the stage level, but that's probably sub-optimal.

azure-pipelines-dependency-graph-testing.yml
```diff
@@ -245,13 +246,23 @@ stages:
...
 
+          - script: |
+              pip install coveralls
+              export CIRCLE_BRANCH=$BUILD_SOURCEBRANCH
+              coveralls
+            displayName: 'coveralls'
+            env:
+              CIRCLECI: 1
+              CIRCLE_BUILD_NUM: $(Build.BuildNumber)
+              COVERALLS_REPO_TOKEN: $(coveralls)
```
Kilo59: This may not be an issue on this PR, but for subsequent PRs ... our unit-test coverage numbers appear to have already dropped below the `40%` threshold since there was nothing to enforce it.

https://coveralls.io/builds/52134850

So (assuming coveralls starts blocking on cov < 40%) we will need to increase our unit-testing coverage or decrease the threshold on coveralls.

azure-pipelines-dependency-graph-testing.yml
```diff
@@ -180,7 +180,9 @@ stages:
       - job: compatibility_matrix
         condition: eq(stageDependencies.scope_check.changes.outputs['CheckChanges.GEChanged'], true)
         variables:
-          GE_pytest_opts: '--no-sqlalchemy'
+          - group: 'common_secrets'
+          - name: GE_pytest_opts
+            value: '--no-sqlalchemy'
```
Kilo59: Initially, I commented this `GE_pytest_opts` key in order to get the `coveralls` step working ([which it now does](https://coveralls.io/builds/52134850) 🎊 ), but I assume we need to add this back.

I think this is equivalent but LMK if that's not correct.

azure-pipelines-dev.yml
```diff
@@ -243,13 +245,23 @@ stages:
...
 
+          - script: |
+              pip install coveralls
+              export CIRCLE_BRANCH=$BUILD_SOURCEBRANCH
+              coveralls
+            displayName: 'coveralls for Unit Tests'
+            env:
+              CIRCLECI: 1
+              CIRCLE_BUILD_NUM: $(Build.BuildNumber)
+              COVERALLS_REPO_TOKEN: $(coveralls)
```
Kilo59, line 263:
```suggestion
              COVERALLS_REPO_TOKEN: $(coveralls)
         continueOnError: true 
```

Kilo59, line 263: Is this the correct way to prevent this step from breaking the pipeline?
Like if `coveralls.io` goes down.

```suggestion
            displayName: 'coveralls for Unit Tests'
            continueOnError: true 
            env:
              CIRCLECI: 1
              CIRCLE_BUILD_NUM: $(Build.BuildNumber)
              COVERALLS_REPO_TOKEN: $(coveralls)
```

PR #5857, opened 2022-08-28
[MAINTENANCE] add `pytest-timeout`

Kilo59:
Use `pytest-timeout` with a `2` second timeout as a very rough check that `unit-tests` are not using slow external dependencies. 

- add `pytest-timeout` dependency
- use `--timeout=2` (seconds) with `Unit Test` Azure pipeline step

Note this really should be more like `1` second, but we are allowing some buffer to account for flaky timing.


billdirks (approved):
LGTM, thanks!

PR #5866, opened 2022-08-29
[MAINTENANCE] Remove vendored `marshmallow__shade`

cdkini:
- Remove `marshmallow__shade` directory and add `marshmallow>=3.7.1,<3.9.0` dependency.
- Update `pre_dump` hook on `ExpectationSuite` to properly convert non-serializable objects

### Definition of Done


great_expectations/core/expectation_suite.py
```diff
@@ -1013,10 +1009,11 @@ def clean_empty(self, data):
...
         data = deepcopy(data)
-        if isinstance(data, ExpectationSuite):
-            data.meta = convert_to_json_serializable(data.meta)
-        elif isinstance(data, dict):
-            data["meta"] = convert_to_json_serializable(data.get("meta"))
+        for key in data:
+            if key.startswith("_"):
+                continue
+            data[key] = convert_to_json_serializable(data[key])
+
```
cdkini, line 1016: we only specified "meta" here but non-serializable objects can live in other parts of the suite - this is the only minor change we need to make to completely remove `marshmallow__shade`

requirements.txt
```diff
@@ -9,6 +9,7 @@ jinja2>=2.10
 jsonpatch>=1.22
 jsonschema>=2.5.1
 makefun>=1.7.0,<2
+marshmallow==3.7.1
```
cdkini: This is the version we vendored - looks like Cloud uses `3.8.0`. Should we match them?

kenwade4: I think so

Kilo59: I don't think we should be pinning this unless we have to. It prevents cloud from updating its version and potentially creates dependency conflicts if our users are installing in environments that also use `marshmallow`.


`marshmallow>=3.7.1`

Maybe put an upper bound of `3.9` but I wouldn't do a strict pin.

Kilo59 (approved):

anthonyburdi (approved):
🚀

PR #5806, opened 2022-08-22
Feature/histogram builder

jcampbell:
Purely performance focused; correct implementation forthcoming.

PR #5872, opened 2022-08-30
[FEATURE] Rename PartitionParameterBuilder into HistogramParameterBuilder with parsimonious evaluation of prerequisite metrics

alexsherstinsky:
### Scope
* Rename `PartitionParameterBuilder` into `HistogramParameterBuilder`
* In `HistogramParameterBuilder`, evaluate prerequisite metrics parsimoniously (only as needed as opposed to all the ones that could be potentially needed).  This helps improve computational performance.
* Update unit tests accordingly.
* Add `docstrings` to explain the slightly more intricate behavior compared to the previous version.

### Note
This pull request builds upon PR #5806 (which should be closed once this one is merged).

- JIRA: GREAT-761/GREAT-1212
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/rule_based_profiler/parameter_builder/histogram_parameter_builder.py
```diff
@@ -158,48 +173,43 @@ def _build_parameters(
...
         """
-        # Obtain bucketize_data directive from "rule state" (i.e., variables and parameters); from instance variable otherwise.
-        bucketize_data = get_parameter_value_and_validate_return_type(
-            domain=domain,
-            parameter_reference=self.bucketize_data,
-            expected_return_type=bool,
-            variables=variables,
-            parameters=parameters,
-        )
+        is_categorical: bool = not self._bucketize_data
```
jcampbell, line 176: This property is still declared as `Union[str, bool]`

great_expectations/rule_based_profiler/parameter_builder/histogram_parameter_builder.py
```diff
@@ -217,11 +227,38 @@ def _build_parameters(
         weights: list
 
         if is_categorical:
-            fully_qualified_column_value_counts_metric_single_batch_parameter_builder_name: str = f"{RAW_PARAMETER_KEY}{self._column_value_counts_metric_single_batch_parameter_builder_config.name}"
+            """
+            In this scenario, even though initial intent was to compute histogram, data turned out to be categorical.
```
jcampbell, line 231: If the user declared `bucketize_data` to be False from the get-go, have we already declared this parameter builder during init?

great_expectations/rule_based_profiler/parameter_builder/histogram_parameter_builder.py
```diff
@@ -158,48 +173,43 @@ def _build_parameters(
...
             )
-            bins_ndarray_as_float: MetricValue
-            if ndarray_is_datetime_type:
-                bins_ndarray_as_float = convert_ndarray_datetime_to_float_dtype(
-                    data=bins
+            bins = column_partition_parameter_node[
+                FULLY_QUALIFIED_PARAMETER_NAME_VALUE_KEY
+            ]
+
+            if bins is None:
```
jcampbell, line 194: Can you describe this case with an example? That may help me understand whether the logic could be simplified. 

In particular, I'm trying to understand the case where the user has asked for bucketized data, but we're returning invalid bins. Why isn't that just an error?

jcampbell (changes requested):

great_expectations/rule_based_profiler/parameter_builder/histogram_parameter_builder.py
alexsherstinsky: @jcampbell It reads for me as:
```
-       bucketize_data: Union[str, bool] = True,
+       bucketize_data: bool = True,
```

Please see Line 51 in `https://github.com/great-expectations/great_expectations/blob/feature/GREAT-761/GREAT-1212/alexsherstinsky/rule_based_profiler/parameter_builder/rename_partition_parameter_builder_into_histogram_parameter_builder_with_conditional_metric_evaluation-2022_08_29-223/great_expectations/rule_based_profiler/parameter_builder/histogram_parameter_builder.py`.  Thanks.

great_expectations/rule_based_profiler/parameter_builder/histogram_parameter_builder.py
alexsherstinsky: @jcampbell Yes, it is indeed there in the constructor if  `bucketize_data` was `False` initially.  But it would not be declared if `bucketize_data` were `True` initially.  This is why here, since we determined that the `bins` are flawed and have to make `bucketize_data` `False` (i.e., `is_categorical` became `True`), we need to instantiate and execute this `"column.value_counts"` parameter builder (because it was not created initially).  Thanks.

great_expectations/rule_based_profiler/parameter_builder/histogram_parameter_builder.py
jcampbell: Ah, I meant the property definition itself (line 160).

great_expectations/rule_based_profiler/parameter_builder/histogram_parameter_builder.py
jcampbell: Sounds like this is the same as the below question then wrt what the case is that we're trying to handle.

great_expectations/rule_based_profiler/parameter_builder/histogram_parameter_builder.py
alexsherstinsky: @jcampbell Very sorry -- that is indeed a typo -- I thought I fixed it before!  Not sure how it crept back in; regardless -- thank you for the thorough review.  I have just fixed it (please refresh).  Thanks.  P.S.: I have not updated the branch with the latest `develop` -- I am assuming that you will do it (let me just have one last glance over before you merge).  Thanks!

great_expectations/rule_based_profiler/parameter_builder/histogram_parameter_builder.py
alexsherstinsky: > Sounds like this is the same as the below question then wrt what the case is that we're trying to handle.

@jcampbell Yes, it is -- but because we decided to declare metric computations "parsimoniously", we do not declare them, until we are absolutely sure that we will need for them to be computed.  (In the old implementation, we computed both, irrespective of the final value of `is_categorical`).

great_expectations/rule_based_profiler/parameter_builder/histogram_parameter_builder.py
alexsherstinsky: @jcampbell I answered this comprehensively before, but am not seeing my answer in GitHub now (it had to do with the original spec in JIRA: GREAT-598).  I am looking for it.

alexsherstinsky: @jcampbell Let me try to repeat here what I wrote earlier (and now cannot find in GitHub): Per JIRA: GREAT-598, we have to support a case when the original request is "column.histogram", but upon calculating the bins, we find that the bins value is incorrect, so we opt to request "column.partition" instead.  I also added the check if bins is NULL to make it more robust.  Furthermore, our "ColumnHistogram" class does not support the "datetime.datetime" type, so encountering such a type also switches from requesting "column.histogram" to "column.partition" ("ColumnPartition" does support "datetime.datetime").  Thanks.

jcampbell: @alexsherstinsky I simply don't see that requirement spelled out in the ticket you mention. Critically, in this case, the parameter builder is intended to be invoked only after the domain has been selected in such a way that we would know whether to build bins for the data first.

I don't understand your reference to support for datetime types -- I would think that datetime types could be treated as either discrete or continuous as appropriate for the case at hand (though generally of course that would also correlate with cardinality).

So, I don't see a case for the complexity proposed here.

alexsherstinsky:
Closing in favor of #5975

PR #5881, opened 2022-08-30
[BUGFIX] many small bugfixes

kenwade4:
**DON'T forget to undo the JSON file rename before merging**

- Add pyogrio as a requirement for expect_column_values_to_be_polygon_area_between.py
- Use correct imported dialect in pytest_generate_tests func of test_expectations_cfe.py
- Update pytest_generate_tests func in test_expectation.py to not die if get_asset call fails
- Add expect_column_values_to_be_valid_arn to skip list in build_gallery.py
- Use Python 3.8 in expectation_gallery pipeline
- Before piping to tee with build_gallery.py, redirect error to stdout too
- Add support for match_like Expectations to redshift/snowflake/bigquery
- Add redshift/snowflake to suppress list on column_mean Expectation
- Update check_json_test_result func to make sure result['result']['unexpected_list'] isn't empty
- Add redshift schemas to some values_to_be_in_set Expectation tests
- Update get_dialect_regex_expression func to check for redshift another way
- Supppress mysql/bigquery tests on contrib expect_column_values_to_be_valid_scientific_notation
- Add a step to expectation_gallery pipeline that shows implemented engines
- Update run_diagnostics to log an error message if no _expectation_config is available
- Update _get_expectation_configuration_from_examples to fallback to trying any test input if None are success:True
- Update _get_execution_engine_diagnostics to not say all engines are implemented when metric_diagnostics_list is empty
- Change test title prefix from positive to negative for contrib exp not_contain_special_characters
- Update output in test_expectation_self_check
- Change how the coverage_score is calculated
- Add a prefix of Core/Contrib to exp_type in get_expectation_file_info_dict

### DIFF
--- a/assets/scripts/build_gallery.py
+++ b/assets/scripts/build_gallery.py
@@ -142,9 +142,13 @@ def get_expectation_file_info_dict(
             if match:
                 if not line.strip().startswith("#"):
                     exp_type_set.add(match.group(1))
-        result[name]["exp_type"] = sorted(exp_type_set)[0]
+        if file_path.startswith("great_expectations"):
+            _prefix = "Core "
+        else:
+            _prefix = "Contrib "
+        result[name]["exp_type"] = _prefix + sorted(exp_type_set)[0]
         logger.debug(
-            f"Expectation type {sorted(exp_type_set)[0]} for {name} in {file_path}"
+            f"Expectation type {_prefix}{sorted(exp_type_set)[0]} for {name} in {file_path}"
         )
 
     os.chdir(oldpwd)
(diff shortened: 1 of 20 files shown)


assets/scripts/build_gallery.py
```diff
@@ -142,9 +142,13 @@ def get_expectation_file_info_dict(
...
                     exp_type_set.add(match.group(1))
-        result[name]["exp_type"] = sorted(exp_type_set)[0]
+        if file_path.startswith("great_expectations"):
+            _prefix = "Core "
+        else:
+            _prefix = "Contrib "
+        result[name]["exp_type"] = _prefix + sorted(exp_type_set)[0]
         logger.debug(
-            f"Expectation type {sorted(exp_type_set)[0]} for {name} in {file_path}"
+            f"Expectation type {_prefix}{sorted(exp_type_set)[0]} for {name} in {file_path}"
```
AFineDayFor, line 151: 👍

AFineDayFor (approved):
🙇‍♀️ merci

PR #5886, opened 2022-08-30
[MAINTENANCE] Add test_expectation_suite_send_usage_message

anthonyburdi:
- Add unit test for `ExpectationSuite.send_usage_message()`
- Showing a mock pattern using a mock/spy to ensure the right events are sent.

### Definition of Done

### DIFF
--- a/tests/core/test_expectation_suite.py
+++ b/tests/core/test_expectation_suite.py
@@ -4,12 +4,12 @@ from copy import copy, deepcopy
 from typing import Any, Dict, List
 
 import pytest
-from great_expectations.core.usage_statistics.events import UsageStatsEvents
 from ruamel.yaml import YAML
 
 from great_expectations import DataContext
 from great_expectations.core.expectation_configuration import ExpectationConfiguration
 from great_expectations.core.expectation_suite import ExpectationSuite
+from great_expectations.core.usage_statistics.events import UsageStatsEvents
 from great_expectations.util import filter_properties_dict
 
 
@@ -631,7 +631,7 @@ class DataContextSendUsageMessageSpy:
     [
         pytest.param(True, id="success=True"),
         pytest.param(False, id="success=False"),
-    ]
+    ],
 )
 def test_expectation_suite_send_usage_message(success: bool):
     """Ensure usage stats event is sent on expectation suite."""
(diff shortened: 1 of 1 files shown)


Kilo59 (approved):
🙌

tests/core/test_expectation_suite.py
```diff
@@ -603,3 +604,55 @@ def test_get_expectations_by_domain_type(
...
+        pytest.param(False, id="success=False"),
+    ]
+)
+def test_expectation_suite_send_usage_message(success: bool):
+    """Ensure usage stats event is sent on expectation suite."""
+
+    dc_message_spy = DataContextSendUsageMessageSpy()
+
+    suite = ExpectationSuite(
+        expectation_suite_name="warning",
```
billdirks: nit: maybe change `warning` to `suite` or something more innocuous.

tests/core/test_expectation_suite.py
```diff
@@ -603,3 +604,55 @@ def test_get_expectations_by_domain_type(
...
+        meta={"notes": "This is an expectation suite."},
+        data_context=dc_message_spy,
+    )
+
+    suite.send_usage_event(success=success)
+
+    assert dc_message_spy.messages
+    assert len(dc_message_spy.messages) == 1
+    assert dc_message_spy.messages[0] == {
+        "event": UsageStatsEvents.EXPECTATION_SUITE_ADD_EXPECTATION.value,
```
billdirks, line 651: This isn't blocking, just a question. Do we want other tests/parameterizations that send in event and event_payload also? 
You could parameterize the inputs as a `parameterized_kwargs` dict and then call `suite.send_usage_event(**parameterized_kwargs)`. When doing the assert you'd need to check for presence/absence of a key in `parameterized_kwargs` to know whether to use the default value or the parameterized value. If that's ugly, we could have a separate test that uses this spy where you send it a nondefault values for these args.

billdirks (commented):
This looks great, thank you!

I've added some comments for you to consider but nothing blocking.

billdirks (approved):

tests/core/test_expectation_suite.py
anthonyburdi: Good call - when we test `context.send_usage_message()` I think that makes sense. For this method though `suite.send_usage_event(success: bool)` only the `success` parameter is available, the rest are set within the method call.

PR #5890, opened 2022-08-31
[MAINTENANCE] Mark existing tests as unit or integration

anthonyburdi:
- Mark existing tests in test_expectation_suite.py as unit or integration
- Two tests are marked as integration tests that use the data context incidentally, these are good examples of tests that can benefit from using a data context test double (in an upcoming PR).

### Definition of Done

### DIFF
--- a/tests/core/test_expectation_suite.py
+++ b/tests/core/test_expectation_suite.py
@@ -237,6 +237,7 @@ def profiler_config():
     return yaml.load(yaml_config)
 
 
+@pytest.mark.unit
 def test_expectation_suite_equality(baseline_suite, identical_suite, equivalent_suite):
     """Equality should depend on all defined properties of a configuration object, but not on whether the *instances*
     are the same."""
@@ -250,6 +251,7 @@ def test_expectation_suite_equality(baseline_suite, identical_suite, equivalent_
     assert baseline_suite != equivalent_suite  # ne works properly
 
 
+@pytest.mark.unit
 def test_expectation_suite_equivalence(
     baseline_suite,
     identical_suite,
@@ -267,6 +269,7 @@ def test_expectation_suite_equivalence(
     assert not single_expectation_suite.isEquivalentTo(baseline_suite)
 
 
+@pytest.mark.unit
 def test_expectation_suite_dictionary_equivalence(baseline_suite):
     assert (
         baseline_suite.isEquivalentTo(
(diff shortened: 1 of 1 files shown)


alexsherstinsky (approved):
LGTM!

tests/core/test_expectation_suite.py
```diff
@@ -237,6 +237,7 @@ def profiler_config():
     return yaml.load(yaml_config)
 
 
+@pytest.mark.unit
 def test_expectation_suite_equality(baseline_suite, identical_suite, equivalent_suite):
     """Equality should depend on all defined properties of a configuration object, but not on whether the *instances*
     are the same."""
```
donaldheppner, line 243: `assert baseline_suite is baseline_suite` can be removed

donaldheppner: this comment applies throughout

tests/core/test_expectation_suite.py
```diff
@@ -250,6 +251,7 @@ def test_expectation_suite_equality(baseline_suite, identical_suite, equivalent_
     assert baseline_suite != equivalent_suite  # ne works properly
 
 
+@pytest.mark.unit
 def test_expectation_suite_equivalence(
     baseline_suite,
     identical_suite,
```
donaldheppner, line 257: I find these parameter names confusing. Can we instead describe the expected differences/similarities?

tests/core/test_expectation_suite.py
```diff
@@ -267,6 +269,7 @@ def test_expectation_suite_equivalence(
     assert not single_expectation_suite.isEquivalentTo(baseline_suite)
 
 
+@pytest.mark.unit
 def test_expectation_suite_dictionary_equivalence(baseline_suite):
     assert (
         baseline_suite.isEquivalentTo(
```
donaldheppner, line 275: there is a lot of duplication in this test body, can we reuse the expectation definition? It'll also make the change more obvious when we make it (in the false case)

tests/core/test_expectation_suite.py
donaldheppner: I feel we are abusing test fixtures here, and it's reducing the readability of the test

tests/core/test_expectation_suite.py
donaldheppner: Also, it seems like we are foregoing using the fixtures we have

donaldheppner (changes requested):
I think the tests within this file need to be rewritten to improve readability and many (if not all) the fixtures removed.

The `profiler_config` fixture loads the yaml defining a profiler for testing an expectation suite, is this required? It seems like the wrong place to do this

PR #5893, opened 2022-08-31
[BUGFIX] Resets static variable after use

billdirks:
This fixes https://superconductive.atlassian.net/browse/GREAT-1127. We currently update a default static variable but don't reset it after use leading to bug and flaky tests.

I new test was added to specifically test this case which I took from @NathanFarmer 's comment in the ticket above.
I've also labeled all the tests in this file as integration tests.

When looking at the diff you may want to ignore whitespace since most of the changes are indentation from the context manager's `with` statement.

Additionally, I have run the modified test file in random order 21 times and it passes. Previously it would fail due to this bug.

### DIFF
--- a/great_expectations/rule_based_profiler/rule_based_profiler.py
+++ b/great_expectations/rule_based_profiler/rule_based_profiler.py
@@ -3,7 +3,8 @@ import datetime
 import json
 import logging
 import sys
-from typing import Any, Callable, Dict, List, Optional, Set, Union
+from contextlib import contextmanager
+from typing import Any, Callable, Dict, Iterator, List, Optional, Set, Union
 
 import great_expectations.exceptions as ge_exceptions
 from great_expectations.core.batch import (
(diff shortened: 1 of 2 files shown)


great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -907,8 +909,13 @@ def _apply_runtime_environment(
             domain_type_directives_list=domain_type_directives_list,
         )
 
-        rule: Rule
-        self._profiler_config.rules = {rule.name: rule.to_json_dict() for rule in rules}
+        try:
+            original_config_rules = self._profiler_config.rules
+            rule: Rule
+            self._profiler_config.rules = {rule.name: rule.to_json_dict() for rule in rules}
+            yield
```
NathanFarmer, line 920: So this will return a generator with `None` in it? And this is required because we are using the `@contextmanager` decorator? And the decorator is required because we want to use `with` in our calls to `_apply_runtime_environment`? And we want to use `with` because...you lost me there.

NathanFarmer (commented):
Please help me understand!

great_expectations/rule_based_profiler/rule_based_profiler.py
billdirks: The purpose of a context manager is to wrap a block of code so one can do some stuff before and after that code executes. That is:
```
with my_context_manager:
    my python code
```
becomes 
``` 
<context manager code before yields>
my python code
<context manager code after yields>
```

Here I want to set `self._profiler_config.rules` for the profiler run and I want to make sure we always reset it afterwards. I use the `try/finally` to make sure we reset this even when there is an exception.

In general, context managers also are able to return something. For example when you open a file you do `with open(file) as f:` and `f` is the file object. I don't care about this returning functionality here but because context managers can do this, they are a `generator iterator`. That lets them yield some value and then still do the cleanup part. Since I don't care about returning anything I just do a bare `yield` so the return type of the context manager is `Iterator[None]`. If the context manager does return something this iterator can only have 1 thing in it which gets used in the `with ... as ...` assignment.

Note, all of this is a shortcut way to write a context manager using the `contextlib` package. The more conventional way to is to write class and define `__enter__` and `__exit__` methods:

```
class MyContextManager:
    def __enter__(self):
        python code to run before

    def __exit__(self, exc_type, exc_value, exc_tb):
        python code to run after
        exception handling also happens here which is what the exc_* args are for.
```

NathanFarmer (approved):
Looks good to me!

billdirks:
Thanks for the review! Apparently black didn't run on these files so I pushed few formatting changes.

billdirks:
I've decided not to merge this. I think a better solution is to make the underlying default profiler config immutable so we aren't mutating and managing global state.

PR #5896, opened 2022-08-31
[BUGFIX] Ensure that ParameterBuilder implementations in Rule Based Profiler properly handle SQL DECIMAL type

alexsherstinsky:
### Scope
* Insure that SQL `DECIMAL` (`NUMERIC` in case of PostGreSQL) type is converted to `float` in metric computations.
* Update existing unit tests.
* Tests done in `Jupyter` notebook on local `PostGreSQL` DBMS.

- JIRA: GREAT-761/GREAT-1220
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/rule_based_profiler/helpers/simple_semantic_type_filter.py
```diff
@@ -127,9 +127,23 @@ def _infer_semantic_domain_type_from_table_column_type(
...
+        if (
+            column_type
+            in (
+                {type_name.upper() for type_name in ProfilerTypeMapping.INT_TYPE_NAMES}
+                | {
+                    type_name.upper()
+                    for type_name in ProfilerTypeMapping.FLOAT_TYPE_NAMES
+                }
+            )
+            or any(
```
Shinnnyshinshin: wanting to check my understanding: by using `startswith()` it makes this inferrence more general?

great_expectations/util.py
```diff
@@ -1514,6 +1514,25 @@ def convert_ndarray_float_to_datetime_tuple(
     return tuple(convert_ndarray_float_to_datetime_dtype(data=data).tolist())
 
 
+def is_ndarray_decimal_dtype(data: np.ndarray) -> bool:
```
Shinnnyshinshin, line 1517: :heart:

great_expectations/util.py
```diff
@@ -1514,6 +1514,25 @@ def convert_ndarray_float_to_datetime_tuple(
...
+    """
+    Determine whether or not all elements of 1-D "np.ndarray" argument are "decimal.Decimal" type objects.
+    """
+    value: Any
+    result: bool = all(isinstance(value, decimal.Decimal) for value in data)
+    return result
+
+
+def convert_ndarray_decimal_to_float_dtype(data: np.ndarray) -> np.ndarray:
```
Shinnnyshinshin, line 1526: :heart:

Shinnnyshinshin (approved):
Thank you for the this work. Appreciate the thorough testing in jsonschema profiler as well

great_expectations/core/profiler_types_mapping.py
```diff
@@ -27,10 +27,14 @@ class ProfilerTypeMapping:
         "INT64",
         "TINYINT",
         "BYTEINT",
+        "ByteType()",
```
billdirks: Some elements in this list are repeated. I see `INTEGER` twice. I also see `integer`. Do we always uppercase this when comparing? If so can you remove `integer` or `INTEGER` altogether?

great_expectations/rule_based_profiler/helpers/simple_semantic_type_filter.py
alexsherstinsky: @Shinnnyshinshin This is to insure that `DECIMAL(a, b)` can be matched to `DECIMAL` (we cannot match on all possible combinations of precision).  Thanks.

great_expectations/rule_based_profiler/helpers/simple_semantic_type_filter.py
```diff
@@ -127,9 +127,23 @@ def _infer_semantic_domain_type_from_table_column_type(
         column_type: str = str(column_types_dict_list[0]["type"]).upper()
 
         semantic_column_type: SemanticDomainTypes
-        if column_type in (
-            {type_name.upper() for type_name in ProfilerTypeMapping.INT_TYPE_NAMES}
-            | {type_name.upper() for type_name in ProfilerTypeMapping.FLOAT_TYPE_NAMES}
+        if (
+            column_type
+            in (
+                {type_name.upper() for type_name in ProfilerTypeMapping.INT_TYPE_NAMES}
```
billdirks: This first case is handled by the second 2. This `if` statement could be changed to only 1 case:
```
if any(
    map(
        lambda type_name: column_type.startswith(type_name.upper()),
        ProfilerTypeMapping.INT_TYPE_NAMES + ProfilerTypeMapping.FLOAT_TYPE_NAMES)):
```

Also, if we are ultimately checking `startswith` do we need all the `float*` variants in the type name lists or could we just use `float`? That is, do we need exact matches elsewhere in the codebase?

great_expectations/rule_based_profiler/parameter_builder/numeric_metric_range_multi_batch_parameter_builder.py
```diff
@@ -544,6 +549,14 @@ def _estimate_metric_value_range(
                 shape=estimation_histogram_shape, fill_value=datetime.datetime.min
             )
         else:
+            if self._is_metric_values_ndarray_decimal_dtype(
+                metric_values=metric_values,
+                metric_value_vector_indices=metric_value_vector_indices,
+            ):
+                metric_values = convert_ndarray_decimal_to_float_dtype(
```
billdirks, line 556: nit: add a type annotation here.

great_expectations/rule_based_profiler/parameter_builder/numeric_metric_range_multi_batch_parameter_builder.py
```diff
@@ -644,6 +657,19 @@ def _is_metric_values_ndarray_datetime_dtype(
 
         return True
 
+    @staticmethod
+    def _is_metric_values_ndarray_decimal_dtype(
+        metric_values: np.ndarray,
+        metric_value_vector_indices: List[tuple],
+    ) -> bool:
+        metric_value_vector: np.ndarray
+        for metric_value_idx in metric_value_vector_indices:
```
billdirks, line 666: Not sure if this is worthwhile but can this `for` loop be vectorized?

great_expectations/rule_based_profiler/parameter_builder/parameter_builder.py
```diff
@@ -581,12 +582,12 @@ def _sanitize_metric_computation(
...
                     elif not (
                         isinstance(metric_value, datetime.datetime)
+                        or isinstance(metric_value, decimal.Decimal)
                         or np.issubdtype(metric_value.dtype, np.number)
                     ):
                         raise ge_exceptions.ProfilerExecutionError(
                             message=f"""Applicability of {self.__class__.__name__} is restricted to numeric-valued \
-metrics (value {metric_value} having numpy.dtype "{str(metric_value.dtype)}" and Python type \
-"{str(type(metric_value))}" was computed).
+metrics (value {metric_value} of type "{str(type(metric_value))}" was computed).
```
billdirks: It looks like we allow numeric types or datetimes? This message can be updated to reflect this.

great_expectations/util.py
```diff
@@ -1514,6 +1514,25 @@ def convert_ndarray_float_to_datetime_tuple(
...
+    """
+    Convert all elements of N-D "np.ndarray" argument from "decimal.Decimal" type to "float" type objects.
+    """
+    convert_decimal_to_float_vectorized: Callable = np.vectorize(
```
billdirks: nit: You can make `Callable` => `Callable[np.ndarray, np.ndarray]`

billdirks (changes requested):
Thanks for this change! I have left a few comments.

great_expectations/core/profiler_types_mapping.py
alexsherstinsky: They are both needed; corresponds to different backend systems.

alexsherstinsky:
> Thanks for this change! I have left a few comments.

Responded!  Thank you very much!

great_expectations/core/profiler_types_mapping.py
alexsherstinsky: got rid of duplication though -- nice catch

great_expectations/rule_based_profiler/helpers/simple_semantic_type_filter.py
alexsherstinsky: @billdirks Good points/question.  However, I would like to keep all those types defined precisely (we painstakingly listed them) -- as well as the logic -- let it exit on first match (also we might change the logic later).  Thanks.

great_expectations/rule_based_profiler/parameter_builder/numeric_metric_range_multi_batch_parameter_builder.py
alexsherstinsky: @billdirks it is already there -- on line 483.  Thanks.

great_expectations/rule_based_profiler/parameter_builder/parameter_builder.py
alexsherstinsky: @billdirks Good point -- I updated the message.  Thanks!

great_expectations/util.py
alexsherstinsky: @billdirks Good to know, but why would we want the two in there?  The actual function accepts only one argument.

great_expectations/rule_based_profiler/helpers/simple_semantic_type_filter.py
billdirks: `any` should short circuit and exit early.
If you want separate conditionals could you get rid of the first one (the equality check) since that works is repeated in the next two.

great_expectations/rule_based_profiler/parameter_builder/numeric_metric_range_multi_batch_parameter_builder.py
billdirks: This is a new variable which shadows the old one of the same name and could have a different type. Because this is an assignment, a developer would need to look at the function signature of `convert_ndarray_decimal_to_float_dtype` to determine the type here.

great_expectations/util.py
billdirks: I wrote that too fast. It should be `Callable[[np.ndarray], np.ndarray]`. Could you update this? This annotates the input and output types. `mypy` could infer this but you've stated you want these type hints for helping developers and a developer won't be able to see that `return convert_decimal_to_float_vectorized(data)` will return an `ndarray` without looking at the definition of `convert_decimal_to_float`.

billdirks (changes requested):
I've left a few more comments.

great_expectations/rule_based_profiler/parameter_builder/numeric_metric_range_multi_batch_parameter_builder.py
alexsherstinsky: @billdirks Perhaps -- I would not outright assume that it cannot be, though I would prefer not to do this right now as part of this PR, because this loop is such that every index points to an `ndarray` (and we are vectorizing that) -- so potentially might require more time to get right.  Thanks.

great_expectations/util.py
alexsherstinsky: Done.  I love this one!  Thanks!

great_expectations/rule_based_profiler/helpers/simple_semantic_type_filter.py
alexsherstinsky: @billdirks I just left the `any` (if we change implementation, we can figure out the cleanest code then).  Regarding listing all the types -- we want to keep them listed as our catalog of the types we know and recognize as they are spelled and with casing as they appear in the various backends.  Having said that, this whole inference of semantic type from "physical" type is not very good (this is why the filter is called "simple semantic filter" and users can supply their own).

alexsherstinsky:
> I've left a few more comments.

@billdirks Incorporated/addressed -- please re-review -- thanks!

billdirks (approved):
LGTM, thanks!

PR #5913, opened 2022-09-01
[MAINTENANCE] Mark all tests within `tests/data_context/stores` dir

cdkini:
- Mark all unmarked tests with either `@pytest.mark.unit` or `@pytest.mark.integration`

### Definition of Done


tests/data_context/store/test_datasource_store.py
```diff
@@ -53,6 +53,7 @@ def datasource_store_with_single_datasource(
     return empty_datasource_store
 
 
+@pytest.mark.unit
```
billdirks, line 56: Does using an empty_datasource_store have any side effects like writing to disk or is everything in memory?

tests/data_context/store/test_store_backends.py
```diff
@@ -1342,6 +1356,8 @@ def test_TupleS3StoreBackend_list_over_1000_keys():
     assert len(keys) == num_keys_to_add + 1
 
 
+@pytest.mark.cloud
+@pytest.mark.unit
```
billdirks, line 1360: We don't need a real `ge_cloud_access_token` in this test right? Should we use a plain string there?

billdirks (commented):
Thanks for doing this. 2 questions inline.

tests/data_context/store/test_datasource_store.py
cdkini: everything is in memory

tests/data_context/store/test_store_backends.py
cdkini: nope! I think that string is just for test anyways it's just a random UUID - will update accordingly

cdkini: so that access token is actually used in `assert_called_with` calls - it is referenced in `shared_called_with_request_kwargs` so it's necessary here. I can decouple these deps but maybe in a separate PR?

billdirks: My understanding is we are trying to be stringent when applying the `unittest` marker since these likely won't be revisited. So a followup PR which decouples this or a comment that says something to the effect of "we don't care what this string is but it is used to match the asserts" would be fine. I just want to be clear to the reader that this doesn't need to be a valid cloud token since we are not hitting cloud directly here.

billdirks (approved):
LGTM, thanks!

I comment about a followup or adding a comment.

PR #5915, opened 2022-09-01
[Maintenance] Cleanup to allow docker test target to run tests in random order

billdirks:
This is prep work to allow our tests to run in random order. Previously I had made a special requirements file that only installed `pytest-random-order` in our azure pipeline. This moves it to our normal contrib requirements file.


great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -152,9 +153,6 @@ class GEDependencies:
         "PyHive",
         "thrift",
         "thrift-sasl",
-        # requirements-dev-test-pipeline.txt:
-        "pytest-azurepipelines",
```
billdirks, line 156: We no longer use `pytest-azurepipelines`.  `pytest-random-order` has been moved to the contrib requirements so is now in a different section of this list at line 133.

azure-pipelines.yml
```diff
@@ -150,7 +150,7 @@ stages:
               # We repull the develop branch. We could build the local branch that is already present but we should
               # prune it's depth since we don't need the git history.
               CMD="docker buildx build -f docker/Dockerfile.tests --tag greatexpectations/test:${DOCKER_TAG} "
-              CMD+="--target azure --build-arg PYTHON_VERSION=$(python.version) "
```
billdirks, line 153: I remove the `azure` target since we don't want this environment to be special in any way so we change this to the `test` target.

requirements-dev-test-pipeline.txt
```diff
@@ -1 +0,0 @@
-pytest-random-order
```
cdkini, line 1: Should this file be deleted?

great_expectations/core/usage_statistics/package_dependencies.py
cdkini: if we remove it here, should we remove all instances in our Azure YAML?

billdirks: It only appears in the `mssql` tests which are outside of docker at this time. We run a coverage report in this step using it though I'm not sure the state of that. Since we are moving to a different coverage solution I'd rather we remove it from that step then.

requirements-dev-test-pipeline.txt
billdirks: Yes, we no longer want this requirements file at all. `pytest-random-order` is now in requirements-dev-contrib.txt where some other test requirements live.

cdkini (approved):

PR #5916, opened 2022-09-01
[MAINTENANCE] Parametrize test_cloud_backed_data_context_add_datasource

anthonyburdi:
- Parametrize and add helper methods and fixtures
- This PR is splitting up work into separate PRs from https://github.com/great-expectations/great_expectations/pull/5664 and https://github.com/great-expectations/great_expectations/pull/5763

### Definition of Done


tests/data_context/cloud_data_context/test_datasource_crud.py
```diff
@@ -15,52 +20,201 @@
...
+    elif config_type == ConfigType.NAME_SUPPLIED_SEPARATELY_AND_INCLUDED_IN_CONFIG:
+        name_and_save_changes_params.update({"name": datasource_name})
+        stored_datasource = context.add_datasource(
+            **name_and_save_changes_params,
+            **datasource_config_with_name.to_dict(),
+        )
+    else:
+        raise Exception(
+            f"config_type must be one of {','.join([str(config_type) for config_type in ConfigType])}"
+        )
```
cdkini, line 130: Instead of branching logic like this in this helper, could we make 3 separate tests that cover each of these cases? I'm all for reducing code complexity but I'd rather we have duplication if it makes it easier to reason about the test.

anthonyburdi: That's what we had before this PR (3 separate cases). The cases are super similar, the differences being how the name is included and the save changes setting. I'm thinking reduce code re-use with this PR, and then hopefully turn these into unit tests as much as possible as part of the testing effort we are undertaking now.

anthonyburdi: Happy to close this PR though if you think this is a step backward.

PR #5900, opened 2022-08-31
[FEATURE] Allowing `schema`  to be passed in as `batch_spec_passthrough` in Spark

Shinnnyshinshin:
Changes proposed in this pull request:
- Follow-up to #5897
- Schemas in Spark Dataframes are defined as `StructType`, which is not serializable.
- This PR adds the `schema.jsonValue()` call which translates the object into a `json`, whenever a schema is passed in as a `batch_spec_passthrough` parameter
- `schema` can be passed in at the `Asset`-level and `DataConnector`-level. 
- `schema` can also be passed in as part of  `BatchRequest` that is part of a `Checkpoint`. 
- The PR adds the necessary changes to `prepare_dump` and `convert_to_json_serializable()` and adds integration tests to `test_serialization.py`. 
- Also includes changes from #5917
### Definition of Done
Please delete options that are not relevant.

- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added [unit tests](https://docs.greatexpectations.io/docs/contributing/contributing_test#writing-unit-and-integration-tests) where applicable and made sure that new and existing tests are passing.
- [x] I have run any local integration tests and made sure that nothing is broken.

Thank you for submitting!

PR #5917, opened 2022-09-01
[FEATURE] `SparkDFExecutionEngine` able to use `schema`

Shinnnyshinshin:
# Proposed Changes
* Follow up to PR #5900 and result of UAT from Spark Notebook for DataAssistants (PR #5904)
* Logic for handling Spark `schema` at the `SparkDFExecutionEngine` involves handling it in 3 ways
   - as `StructType()` object : Spark schema can be directly used by `spark.read()` function
   - as `dict` : `StructType()` has been serialized into JSON. Call `fromJson()` to turn it back into `StructType()` before passing to `spark.read()` function.
   - as `str` : Object has been incorrectly serialized... throw an `ExecutionEngineError` asking user to run `.jsonValue()`

## Why so many tests? 
*  The work around the Spark Notebook for DataAssistant **also** revealed that significant sections of our codebase around Spark were simply not being tested. 
* This PR attempts to add tests at following levels
   - `SparkDFExecutionEngine`: These tests ensure `schema` is being properly handled and loaded into `SparkDFExecutionEngine`. 
   - `Datasource-level` : These tests were missing entirely, and we were only testing `pandas` and `sql`.  Renamed the existing test to `*_pandas_tests.py`, and created `*_sparkdf_tests.py`. These issues were surfaced when trying to load a `schema` into the `Datasource`-config (now enabled by #5900)
   - Checkpoint-level : These tests were missing entirely as well, and was surfaced when we were trying to load a `schema` into the Checkpoint configuration (now enabled by #5900).  Tests were added for `Checkpoint` and `SimpleCheckpoint`.
  

### Definition of Done


great_expectations/datasource/data_connector/configured_asset_file_path_data_connector.py
```diff
@@ -224,10 +224,15 @@ def build_batch_spec(self, batch_definition: BatchDefinition) -> PathBatchSpec:
...
-                self.assets[data_asset_name]["batch_spec_passthrough"]
-            )
+            # batch_spec_passthrough from data asset
+            if isinstance(self.assets[data_asset_name], Asset):
+                batch_spec_passthrough = deepcopy(
+                    self.assets[data_asset_name].batch_spec_passthrough
+                )
+            else:
+                batch_spec_passthrough = deepcopy(
+                    self.assets[data_asset_name]["batch_spec_passthrough"]
```
billdirks: On line 224 above you check for `self.assets[data_asset_name].batch_spec_passthrough`. However, here you use `self.assets[data_asset_name]["batch_spec_passthrough"]`. What's the difference here and is the conditional above correct?

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -260,6 +260,7 @@ def get_batch_data_and_markers(
             reader_options: dict = batch_spec.reader_options or {}
             path: str = batch_spec.path
             azure_url = AzureUrl(path)
+            # TODO <WILL> 202209 - Add `schema` definition to Azure like PathBatchSpec below
```
billdirks: If you add a TODO to the code, I think adding a ticket and referencing it is a good strategy or this outstanding work may get lost.

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -289,8 +290,34 @@ def get_batch_data_and_markers(
...
+                Union[pyspark.sql.types.StructType, dict, str]
+            ] = reader_options.get("schema")
+
+            # schema can be a dict if it has been through serialization step,
+            # either as part of the datasource configuration, or checkpoint config
+            if isinstance(schema, dict):
+                schema_converted: pyspark.sql.types.StructType = (
+                    sparktypes.StructType.fromJson(schema)
+                )
+                schema = schema_converted
```
billdirks: Could you remove `schema_converted` and just do:
```
schema: pyspark.sql.types.StructType = (
                    sparktypes.StructType.fromJson(schema)
                )
```

billdirks (commented):
I haven't looked at the test yet but here are some initial comments.

tests/checkpoint/conftest.py
```diff
@@ -5,8 +5,11 @@
 
 from great_expectations import DataContext
 from great_expectations.core import ExpectationConfiguration
+from great_expectations.core.yaml_handler import YAMLHandler
 from great_expectations.data_context.util import file_relative_path
 
+yaml = YAMLHandler()
```
billdirks: Do we need this on the global scope?

tests/checkpoint/conftest.py
```diff
@@ -135,3 +138,84 @@ def titanic_spark_data_context_with_v013_datasource_with_checkpoints_v1_with_emp
...
     return context
+
+
+@pytest.fixture
+def context_with_single_taxi_csv_spark(
+    empty_data_context, tmp_path_factory, spark_session
+):
+    context = empty_data_context
+
+    base_directory = str(tmp_path_factory.mktemp("test_checkpoint_spark"))
```
billdirks, line 149: Do we have an in memory file system we can use?

great_expectations/execution_engine/sparkdf_execution_engine.py
Shinnnyshinshin: Thanks for the suggestion. Now being tracked by GREAT-1224

tests/checkpoint/conftest.py
Shinnnyshinshin: I believe so, but we currently don't use it extensively... I can look into it as an option for this PR

tests/checkpoint/conftest.py
Shinnnyshinshin: Good call, no we only need it for the one fixture. I can move it down

great_expectations/execution_engine/sparkdf_execution_engine.py
Shinnnyshinshin: done!

great_expectations/datasource/data_connector/configured_asset_file_path_data_connector.py
Shinnnyshinshin: This was a "hack" that was in place before #5883 was merged :)  Change has been removed.

tests/checkpoint/conftest.py
Shinnnyshinshin: Looked into it and we are using `pyarrow` to do some in-memory filesystem tests, mostly for S3 and DBFS.  Using it for Filesystem (like in this case) would be a bit of a departure from existing patterns, but one that I think there would be. Would it be ok to create a follow-up task to re-visit the possibility of using an in-memory filesystem for these types of tests? (not just this fixture, but for other Filesystem tests)

billdirks: Yes, a follow up is fine.

great_expectations/execution_engine/sparkdf_execution_engine.py
billdirks: Can you add that ticket number to the comment? That way when someone is looking at the todo in the future they will have the ticket number and will be able to see the progress. For example, I sometimes see TODOs in code which are done but the comment has not been removed.

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -289,8 +290,33 @@ def get_batch_data_and_markers(
...
+                raise ge_exceptions.ExecutionEngineError(
+                    f"""
+                                Spark schema was not properly serialized.
+                                Please run the .jsonValue() method on the schema object before loading into GE.
+                                schema: your_schema.jsonValue()
+                                """
+                )
             try:
-                reader: DataFrameReader = self.spark.read.options(**reader_options)
+                if schema:
```
billdirks, line 314: Is `self.spark.read.options(**reader_options)` equivalent to `self.spark.read.schema(None).options(**reader_options)`? I'm wondering if we can get rid of the conditional altogether.

billdirks: I was looking at [the source via the docs](https://spark.apache.org/docs/3.1.1/api/python/_modules/pyspark/sql/readwriter.html#DataFrameReader.schema) and it looks like it can't be None.

billdirks (approved):
LGTM, thanks!

I have one outstanding comment that would be nice to integrate.

great_expectations/execution_engine/sparkdf_execution_engine.py
Shinnnyshinshin: Another great suggestion. Looking into this now

great_expectations/execution_engine/sparkdf_execution_engine.py
Shinnnyshinshin: Added the ticket number to the comment!

great_expectations/execution_engine/sparkdf_execution_engine.py
Shinnnyshinshin: `TypeError: schema should be StructType or string`

Shinnnyshinshin: Yup, this is what I see too

PR #5925, opened 2022-09-02
[MAINTENANCE] usage-stats, & utils.py typing

Kilo59:
- breakup `pyproject.toml` mypy packages (in `exclude`)
- `core/usage_statistics` typing
- `utils.py` typing

### TypeGuard

This PR is our first use of a custom `TypeGuard` (PEP 647).
Which is a way to give the type-checker additional details about the kind of type-narrowing a custom `is_<SOMETHING>` function was performing.

https://mypy.readthedocs.io/en/stable/type_narrowing.html?highlight=TypeGuard#user-defined-type-guards

My initial PR made several of these `is_<SOMETHING>` functions into TypeGuards, but I reverted them.
https://github.com/great-expectations/great_expectations/pull/5925#discussion_r962095170

[Numpy types](https://numpy.org/devdocs/reference/typing.html)

### Definition of Done


great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -215,8 +215,8 @@ def get_required_dependency_names(self) -> List[str]:
         """Sorted list of required GE dependencies"""
         return self.GE_REQUIRED_DEPENDENCIES
 
-    def get_dev_dependency_names(self) -> List[str]:
-        """Sorted list of dev GE dependencies"""
```
Kilo59, line 223: This wasn't returning a sorted list.
Maybe it should have been??

pyproject.toml
```diff
@@ -45,33 +46,106 @@ exclude = [
...
+    'render/renderer/profiling_results_overview_section_renderer\.py',  # 2
+    'render/renderer/site_builder\.py',  # 3
+    'render/renderer/suite_edit_notebook_renderer\.py',  # 7
+    'render/renderer/suite_scaffold_notebook_renderer\.py',  # 7
+    'render/renderer/v3/suite_edit_notebook_renderer\.py',  # 11
+    'render/renderer/v3/suite_profile_notebook_renderer\.py',  # 4
     'render/view',  # 11
     'render/types',  # 4
-    'expectations/expectation', # 188
+    'expectations/core',  # 257
```
Kilo59: This is likely much higher because the previous type-check did not install needed type-stubs (which expose additional issues)

great_expectations/util.py
```diff
@@ -1298,19 +1304,19 @@ def is_truthy(value: Any) -> bool:
...
+def is_int(value: Any) -> TypeGuard[int]:
     try:
         int(value)
     except (TypeError, ValueError):
         return False
     return True
 
 
-def is_float(value: Any) -> bool:
+def is_float(value: Any) -> TypeGuard[float]:
```
Kilo59: On closer inspection these don't work as `TypeGuard`s, they are checking that that value is convertible to an `int` or `float` but not that they actually are.
For example `"1.5"` would pass the `is_float` test, but it's still a `str`.

great_expectations/util.py
```diff
@@ -132,7 +138,7 @@ def hyphen(txt: str):
     return txt.replace("_", "-")
 
 
-def profile(func: Callable = None) -> Callable:
```
Kilo59, line 136: Does this change make sense?
Would this ever be valid without proving a `Callable`? I would think not, but I could be mistaken.

cdkini: I'm good with this change since it isn't really public API

cdkini (approved):

PR #5936, opened 2022-09-06
[MAINTENANCE] Reset globals modified in tests

billdirks:
Please `hide whitespace` when looking at this diff is I'm using a contextmanager which indents a large block of code confusing the diff.

Some of our tests update the default profiler configs which are a class variable. Since they don't reset it, these remain set for all other tests run in this session. Some of these tests rely on having the default config.

The "real" fix would be to make the default config immutable. However that is a bigger effort and our current goal is to remove dependencies between tests and run them in a random order.

### DIFF
--- a/tests/integration/profiling/rule_based_profiler/test_profiler_user_workflows.py
+++ b/tests/integration/profiling/rule_based_profiler/test_profiler_user_workflows.py
@@ -2,7 +2,7 @@ import contextlib
 import copy
 import datetime
 from numbers import Number
-from typing import Any, Dict, Iterator, List, Optional, Protocol, Tuple, Type, cast
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Type, cast
 from unittest import mock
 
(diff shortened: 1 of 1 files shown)


tests/integration/profiling/rule_based_profiler/test_profiler_user_workflows.py
```diff
@@ -989,234 +1004,270 @@ def test_bobby_expect_column_values_to_be_between_auto_yes_default_profiler_conf
...
+    original_default_profiler_config = copy.deepcopy(
+        expectation.default_profiler_config
+    )
+    try:
+        yield
+    finally:
+        expectation.default_profiler_config = original_default_profiler_config
+        expectation.default_kwarg_values[
+            "profiler_config"
+        ] = original_default_profiler_config
```
NathanFarmer, line 1038: ❤️

NathanFarmer (approved):
Thank you for this!

PR #5941, opened 2022-09-06
[MAINTENANCE] DatasourceStore refactoring

anthonyburdi:
- Use set instead of create in DatasourceStore
- Return a DatasourceConfig from set() via a get() using the same key
- Remove separate code path: `CloudDataContext._instantiate_datasource_from_config_and_update_project_config`
- Test enhancements

### Definition of Done


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1740,9 +1740,7 @@ def _instantiate_datasource_from_config_and_update_project_config(
             DatasourceInitializationError
         """
         if save_changes:
-            self._datasource_store.set_by_name(  # type: ignore[attr-defined]
-                datasource_name=config.name, datasource_config=config
-            )
+            config = self._datasource_store.set(key=None, value=config)  # type: ignore[attr-defined]
```
anthonyburdi, line 1743: Now that `set()` also calls `get()` to retrieve and return the newly created config (which will be populated with IDs in the cloud store case) we use that config in the datasource cache and to instantiate the datasource.

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -338,49 +336,6 @@ def root_directory(self) -> Optional[str]:
         """
         return self._context_root_directory
 
-    def _instantiate_datasource_from_config_and_update_project_config(
```
anthonyburdi, line 341: This method is no longer needed since creating a new datasource config is handled by the `DatasourceStore.set()` method in all cases.

great_expectations/data_context/store/datasource_store.py
```diff
@@ -145,17 +148,9 @@ def delete(self, datasource_config: DatasourceConfig) -> None:
     def _build_key_from_config(  # type: ignore[override]
         self, datasource_config: DatasourceConfig
     ) -> Union[GeCloudIdentifier, DataContextVariableKey]:
-        if hasattr(datasource_config, "id"):
```
anthonyburdi, line 148: These attributes are required, so there is no need for these hasattr checks.

great_expectations/data_context/store/datasource_store.py
```diff
@@ -172,22 +167,31 @@ def set_by_name(
...
+
+        # Make two separate requests to set and get in order to obtain any additional
+        # values that may have been added to the config by the StoreBackend (i.e. object ids)
+        ref: Optional[Union[bool, GeCloudResourceRef]] = super().set(key, value)  # type: ignore[func-returns-value]
+        if ref and isinstance(ref, GeCloudResourceRef):
+            key.ge_cloud_id = ref.ge_cloud_id  # type: ignore[attr-defined]
+
+        return_value: DatasourceConfig = self.get(key)  # type: ignore[assignment]
+        if not return_value.name and isinstance(key, DataContextVariableKey):
+            return_value.name = key.resource_name
```
anthonyburdi: This is needed to handle adding the name param to v2 datasources.

great_expectations/data_context/store/datasource_store.py
```diff
@@ -172,22 +167,31 @@ def set_by_name(
...
-        ] = self._build_key_from_config(datasource_config)
-        return self.set(key, datasource_config)  # type: ignore[func-returns-value]
+        if not key:
+            key = self._build_key_from_config(value)
+
+        # Make two separate requests to set and get in order to obtain any additional
+        # values that may have been added to the config by the StoreBackend (i.e. object ids)
+        ref: Optional[Union[bool, GeCloudResourceRef]] = super().set(key, value)  # type: ignore[func-returns-value]
+        if ref and isinstance(ref, GeCloudResourceRef):
+            key.ge_cloud_id = ref.ge_cloud_id  # type: ignore[attr-defined]
```
anthonyburdi, line 188: We need to use the id to retrieve the datasource config we just set in the store. We can consider other ways to do this, e.g. check if ref is GeCloudResourceRef and then set `key = ref` to use in the subsequent `self.get(key)` call.

tests/conftest.py
```diff
@@ -2665,9 +2668,122 @@ def cloud_data_context_with_datasource_pandas_engine(
     return context
 
 
+JSONData = Union[AnyPayload, Dict[str, Any]]
```
anthonyburdi: Most of these changes are moving fixtures to a more general conftest.py since they are now used in several test files.

tests/conftest.py
```diff
@@ -2686,6 +2802,8 @@ def cloud_data_context_in_cloud_mode_with_datasource_pandas_engine(
         "great_expectations.data_context.store.ge_cloud_store_backend.GeCloudStoreBackend.list_keys"
     ), mock.patch(
         "great_expectations.data_context.store.ge_cloud_store_backend.GeCloudStoreBackend._set"
+    ), mock.patch(
```
anthonyburdi: Now we need to also mock the new get request kicked off when we call `DatasourceStore.set()`

tests/data_context/cloud_data_context/test_datasource_crud.py
```diff
@@ -60,16 +64,15 @@ def test_base_data_context_in_cloud_mode_add_datasource(
...
 
-    def mocked_post_response(*args, **kwargs):
-        return mock_response_factory({"data": {"id": datasource_id}}, 201)
-
     with patch(
-        "requests.post", autospec=True, side_effect=mocked_post_response
-    ) as mock_post:
+        "requests.post", autospec=True, side_effect=mocked_datasource_post_response
+    ) as mock_post, patch(
+        "requests.get", autospec=True, side_effect=mocked_datasource_get_response
```
anthonyburdi, line 74: mock the new get() call inside `DatasourceStore.set()`

tests/data_context/cloud_data_context/test_datasource_crud.py
```diff
@@ -124,16 +125,12 @@ def mocked_post_response(*args, **kwargs):
         if save_changes:
             # Make sure the id was populated correctly into the created datasource object and config
             assert stored_datasource.id == datasource_id
-            assert retrieved_datasource.id == datasource_id
```
anthonyburdi, line 127: Since we are mocking the retrieved datasource this test would just be checking the fixture accuracy. Removing to make it more clear what we are testing.

tests/data_context/conftest.py
```diff
@@ -591,11 +589,6 @@ def data_context_with_incomplete_global_config_in_dot_dir_only(
         yield
 
 
-@pytest.fixture
```
anthonyburdi, line 594: moved fixtures to top level conftest.py

tests/data_context/store/test_datasource_store.py
```diff
@@ -196,23 +202,22 @@ def test_datasource_store_retrieval_cloud_mode(
...
-    key = GeCloudIdentifier(
-        resource_type=GeCloudRESTResource.DATASOURCE, ge_cloud_id="foobarbaz"
-    )
-
-    with patch("requests.put", autospec=True) as mock_put:
-        type(mock_put.return_value).status_code = PropertyMock(return_value=200)
+    with patch(
+        "requests.post", autospec=True, side_effect=mocked_datasource_post_response
+    ) as mock_post, patch(
+        "requests.get", autospec=True, side_effect=mocked_datasource_get_response
```
anthonyburdi, line 212: mock the new get() call inside `DatasourceStore.set()`

tests/data_context/test_data_context_datasources.py
```diff
@@ -185,7 +189,7 @@ def test_get_datasource_cache_miss(
 @pytest.mark.unit
 def test_DataContext_add_datasource_updates_cache_and_store(
     cloud_data_context_in_cloud_mode_with_datasource_pandas_engine: DataContext,
-    pandas_enabled_datasource_config: dict,
+    datasource_config_with_names: DatasourceConfig,
```
anthonyburdi, line 192: The actual datasource config is not important for this test, as long as it is a valid config.

great_expectations/data_context/store/datasource_store.py
```diff
@@ -172,22 +167,31 @@ def set_by_name(
...
-        key: Union[
-            GeCloudIdentifier, DataContextVariableKey
-        ] = self._build_key_from_config(datasource_config)
-        return self.set(key, datasource_config)  # type: ignore[func-returns-value]
+        if not key:
+            key = self._build_key_from_config(value)
+
+        # Make two separate requests to set and get in order to obtain any additional
+        # values that may have been added to the config by the StoreBackend (i.e. object ids)
+        ref: Optional[Union[bool, GeCloudResourceRef]] = super().set(key, value)  # type: ignore[func-returns-value]
```
donaldheppner, line 186: ... does the return type have to be optional?

great_expectations/data_context/store/datasource_store.py
donaldheppner: This is an unfortunate compromise that leaks; I'd rather have the set method responsible for returning a reference to whatever resource was just created/updated _for all store backends_ and remove `bool` as an expected return value. Maybe this one piece can be captured in a JIRA item? I understand that we had to do it this way for now

great_expectations/data_context/store/datasource_store.py
```diff
@@ -218,3 +222,12 @@ def _determine_datasource_key(self, datasource_name: str) -> DataContextVariable
             resource_name=datasource_name,
         )
         return datasource_key
+
+    def _validate_key(self, key: Union[GeCloudIdentifier, Tuple[str, ...]]) -> None:  # type: ignore[override]
```
donaldheppner: This is a great opportunity to generalize Key into an object that is self-validating, as opposed to making every consumer validate the key using what is effectively a utility method

great_expectations/data_context/store/datasource_store.py
cdkini: Could we add a comment that says this here?

great_expectations/data_context/store/datasource_store.py
```diff
@@ -218,3 +222,12 @@ def _determine_datasource_key(self, datasource_name: str) -> DataContextVariable
...
+        # Continue to support tuple as a key for get method until we can refactor
+        if not isinstance(key, GeCloudIdentifier):
+            super()._validate_key(key)  # type: ignore[arg-type]
+        else:
+            assert isinstance(
+                key, GeCloudIdentifier
+            ), f"Only {GeCloudIdentifier.__name__} is supported in {self.__class__.__name__}"
```
cdkini: Nit - don't think we need `if/else` here
```suggestion
    def _validate_key(self, key: Union[GeCloudIdentifier, Tuple[str, ...]]) -> None:  # type: ignore[override]
        # Continue to support tuple as a key for get method until we can refactor
        assert isinstance(
                key, GeCloudIdentifier
            ), f"Only {GeCloudIdentifier.__name__} is supported in {self.__class__.__name__}"
        super()._validate_key(key)  # type: ignore[arg-type]
```

tests/conftest.py
```diff
@@ -2665,9 +2668,122 @@ def cloud_data_context_with_datasource_pandas_engine(
...
+        return MockResponse(
+            json_data=json_data, status_code=status_code, exc_to_raise=exc_to_raise
+        )
+
+    return _make_mock_response
+
+
+@pytest.fixture
+def datasource_name() -> str:
+    return "my_first_datasource"
```
donaldheppner: I'd rather this name be more descriptive, like "test_datasource_name"

tests/data_context/test_data_context_datasources.py
cdkini: Could we add this to the docstring?

tests/conftest.py
```diff
@@ -2665,9 +2668,122 @@ def cloud_data_context_with_datasource_pandas_engine(
...
+def datasource_config_with_names_and_ids(
+    datasource_config_with_names: DatasourceConfig, datasource_id: str
+) -> DatasourceConfig:
+    updated_config = copy.deepcopy(datasource_config_with_names)
+    updated_config["id"] = datasource_id
+    return updated_config
+
+
+@pytest.fixture
+def mocked_datasource_get_response(
```
donaldheppner: it seems this fixture would be specific to datasource store backend tests, and defined there

tests/conftest.py
```diff
@@ -2665,9 +2668,122 @@ def cloud_data_context_with_datasource_pandas_engine(
...
+                },
+            },
+            200,
+        )
+
+    return _mocked_get_response
+
+
 @pytest.fixture
 def cloud_data_context_in_cloud_mode_with_datasource_pandas_engine(
```
donaldheppner: the implication of this fixture is that it supports very deep integration testing. Again, I'd rather have it defined closer to the tests with no expectation that it be broadly shared

donaldheppner (changes requested):

great_expectations/data_context/store/datasource_store.py
anthonyburdi: It does, currently at least the `InMemoryStoreBackend` and `InlineStoreBackend` return None on `_set()` which propagates up to this return type. We can standardize and return a key for all backends but I'd prefer to do that (and remove the bool type) in a separate effort/PR in case it is not trivial.

great_expectations/data_context/store/datasource_store.py
anthonyburdi: Just replied above about this. I agree about capturing in a separate JIRA and about not being excited about this approach. Created JIRA: https://superconductive.atlassian.net/browse/GREAT-1235 - is there a good epic to put this in?

great_expectations/data_context/store/datasource_store.py
anthonyburdi: Totally agree. I think we need to be careful to be backward compatible though since we currently use tuples as keys. Could this be split into a separate JIRA item?

tests/data_context/test_data_context_datasources.py
anthonyburdi: Done!

great_expectations/data_context/store/datasource_store.py
anthonyburdi: Done!

great_expectations/data_context/store/datasource_store.py
anthonyburdi: I think we do, since we don't want to do the assertion if the key is a tuple. Or am I missing something?

tests/conftest.py
anthonyburdi: Good call, moving

anthonyburdi: Moved to tests/data_context/conftest.py since these fixtures are also currently used in integration tests for cloud backed CRUD

tests/conftest.py
anthonyburdi: Done!

tests/conftest.py
anthonyburdi: Moved!

tests/conftest.py
```diff
@@ -2665,35 +2665,6 @@ def cloud_data_context_with_datasource_pandas_engine(
     return context
 
 
-@pytest.fixture
-def cloud_data_context_in_cloud_mode_with_datasource_pandas_engine(
```
anthonyburdi, line 2669: Moved into data_context/conftest.py
Currently many of these fixtures are used in multiple subfolders (not just store) otherwise I would have moved them further down.

great_expectations/data_context/store/datasource_store.py
donaldheppner: Should this functionality live in the superclass?

donaldheppner: .... upon further investigation, this method can probably be deleted since it's handled in the the base store class (via ge_cloud_mode)

donaldheppner (changes requested):

donaldheppner (approved):

great_expectations/data_context/store/datasource_store.py
anthonyburdi: This method is removed, and ticket great-1241 created to generalize and add self-validation.

PR #5947, opened 2022-09-07
[BUGFIX] Don't include abstract Expectation classes in _retrieve_expectations_from_module

kenwade4:
This was causing build_package_gallery.py to show zero Expectations for the CapitalOne package. A piece of code was trying to access the `expectation_type` attribute on an abstract Expectation class and it was causing an AssertionError, which resulted in the entire list of Expectation diagnostics being empty.

- Don't include abstract Expectation classes in _retrieve_expectations_from_module


talagluck (approved):
LGTM!

PR #5966, opened 2022-09-08
[MAINTENANCE] Expectation suite new unit tests for add_citation

anthonyburdi:
- Add unit tests for add_citation
- Nest unit tests related to a specific method under a class in the test file for easy organization & an easy way to see if any methods are lacking test coverage (to complement automated test coverage metrics).

NOTE: Currently this also includes the changes from https://github.com/great-expectations/great_expectations/pull/5957 until the release is completed and 5957 can get merged.

### Definition of Done


tests/core/test_expectation_suite.py
```diff
@@ -26,7 +33,290 @@ def exp1():
...
+            "column": "a",
+            "value_set": [1, 2, 3, 4, 5],
+            "result_format": "BASIC",
+        },
+        "meta": {"notes": "This is an expectation."},
+    }
+
+
+class TestInitMethod:
+    """Tests related to ExpectationSuite.__init__()"""
```
cdkini, line 50: Just out of curiosity, what led you to use a class-based set of tests as opposed to individual functions? Still trying to learn how best to use this feature.

tests/core/test_expectation_suite.py
```diff
@@ -26,7 +33,290 @@ def exp1():
...
+    def test_expectation_suite_init_overrides(
+        self,
+        fake_expectation_suite_name: str,
+        expect_column_values_to_be_in_set_col_a_with_meta: ExpectationConfiguration,
+    ):
+        class DummyDataContext:
+            pass
+
+        class DummyExecutionEngine:
+            pass
```
cdkini, line 77: Should these inherit the classes they're meant to model? Is there value in using `mock.MagicMock` instead?

tests/core/test_expectation_suite.py
```diff
@@ -26,7 +33,290 @@ def exp1():
...
+            def __dict__(self):
+                raise NotImplementedError
+
+        test_meta = {"this_is_not_json_serializable": NotSerializable()}
+
+        with pytest.raises(InvalidExpectationConfigurationError):
+            ExpectationSuite(
+                expectation_suite_name=fake_expectation_suite_name,
+                meta=test_meta,  # type: ignore[arg-type]
+            )
```
cdkini: could we add `as e` and `assert "blah" in str(e.value)` for any `pytest.raises` calls around error we raise?

tests/core/test_expectation_suite.py
```diff
@@ -26,7 +33,290 @@ def exp1():
...
+            "comment",
+            "citation_date",
+            "batch_spec",
+            "batch_markers",
+        }
+        # batch_definition (along with other keys that are not provided) should be filtered out
+        assert "batch_definition" not in citation_keys
+
+    @pytest.mark.unit
+    @pytest.mark.v2_api
```
cdkini, line 220: ❤️

tests/core/test_expectation_suite.py
```diff
@@ -26,7 +33,290 @@ def exp1():
...
+    def test_add_citation_citation_date_datetime_override(
+        self, empty_suite_with_meta: ExpectationSuite
+    ):
+        empty_suite_with_meta.add_citation(
+            "fake_comment", citation_date=datetime.datetime(2022, 9, 8)
+        )
+        assert (
+            empty_suite_with_meta.meta["citations"][0]["citation_date"]
+            == "2022-09-08T00:00:00.000000Z"
+        )
```
cdkini: Might it make sense to parameterize to consolidate the logic of these two?

cdkini (approved):

tests/core/test_expectation_suite.py
anthonyburdi: I was thinking more about collecting tests related to specific methods in one place. It might be a little easier to see if a method is untested or not have to look far for related tests. Especially when new tests are added the hope is they get into the class and are not at the bottom of the file away from other related tests.
Anyway, this is new for me too with pytest but that was my thought process.

tests/core/test_expectation_suite.py
anthonyburdi: For these dummy classes (just meant to fill the param list) we don't want to inherit from the class with all the overhead / need to implement abstract methods. For this test we don't need to call any methods on them, just to check that what we supply in the param is then stored as an attribute on the class instance.
As for `mock.MagicMock`, that can definitely be used instead here but we don't need the extra functionality of setting return values or checking what methods were called / params passed. So a dummy class might be less overhead. Maybe @billdirks has an opinion!

tests/core/test_expectation_suite.py
anthonyburdi: Good call! I only added for the other one in this PR. Adding now.

tests/core/test_expectation_suite.py
anthonyburdi: Good call! Done.

tests/core/test_expectation_suite.py
anthonyburdi: Our future selves will thank us :)

PR #5967, opened 2022-09-08
[MAINTENANCE] Unit tests for `CheckpointStore`

cdkini:
- Unit tests for `CheckpointStore`

### Definition of Done


great_expectations/checkpoint/checkpoint.py
```diff
@@ -73,8 +73,11 @@ def __init__(
         checkpoint_config: CheckpointConfig,
         data_context: "DataContext",  # noqa: F821
     ) -> None:
-        # Note the gross typechecking to avoid a circular import
-        if "DataContext" not in str(type(data_context)):
+        from great_expectations.data_context.data_context.abstract_data_context import (
+            AbstractDataContext,
+        )
+
+        if not isinstance(data_context, AbstractDataContext):
```
cdkini, line 80: Better error check using an inline import - also makes this more testable.

tests/data_context/store/test_checkpoint_store.py
```diff
@@ -22,6 +30,15 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.fixture
+def checkpoint_store_with_mock_backend() -> Tuple[CheckpointStore, mock.MagicMock]:
```
anthonyburdi, line 34: I like it!

tests/data_context/store/test_checkpoint_store.py
```diff
@@ -350,3 +367,164 @@ def test_serialization_self_check(capsys) -> None:
...
+
+
+@pytest.mark.unit
+def test_get_checkpoint_with_invalid_legacy_checkpoint_raises_error(
+    checkpoint_store_with_mock_backend: Tuple[CheckpointStore, mock.MagicMock]
+) -> None:
+    store, mock_backend = checkpoint_store_with_mock_backend
+    mock_backend.get.return_value = (
+        CheckpointConfig().to_json_dict()
+    )  # Defaults to empty LegacyCheckpoint
```
anthonyburdi, line 504: Thanks for the comment 👍

tests/data_context/store/test_checkpoint_store.py
```diff
@@ -350,3 +367,164 @@ def test_serialization_self_check(capsys) -> None:
...
+        store.get_checkpoint(name="my_checkpoint", ge_cloud_id=None)
+
+    assert (
+        "Attempt to instantiate LegacyCheckpoint with insufficient and/or incorrect arguments"
+        in str(e.value)
+    )
+
+
+@pytest.mark.unit
+def test_add_checkpoint(
```
anthonyburdi, line 516: Does it make sense to call this a legacy checkpoint test and create a separate test with non legacy checkpoint?

anthonyburdi (commented):
Just a question to address before approval.

tests/data_context/store/test_checkpoint_store.py
cdkini: so this particular test doesn't actually have any logical around legacy vs current checkpoints (while the one above did). For that reason, I don't think we need to get as granular and separate it out with more tests.

Let me know if you disagree and I'd be happy to add another test.

anthonyburdi (approved):

tests/data_context/store/test_checkpoint_store.py
anthonyburdi: I think that makes sense. You would only be testing that add_checkpoint was called with a different checkpoint config which shouldn't be different than the test as written.

PR #5994, opened 2022-09-12
[FEATURE] add new metric - query template values

itaise:
Hi, 
1. This is a new metric which will allow to replace templated values based on a dict, for query expectations with multiple values to replace in the query. In our use cases we needed to replace 5 or more values in some queries.

The expectation using the metric should provide a "template_dict" dict - where the keys are the parameters in the query and the values are the values to replace. 

2. I added a new expectation (that our users are asking for), that use this metric. it validates that a value has unique values (e.g. 1,2,3 and not 1,2,2) under some condition. We need it to validate that under some condition ("is_current_row), we have uniqueness (e.g. one order_id)

for example:

**query**
``` select {my_parameter}, {my_second_parameter} from {active_batch} where {my_third_parameter} is not null```

**template_dict**
```
{"my_parameter": "amount", "my_second_parameter": "card", "my_third_parameter": "cash"}
```

The code is based on **query_column_pair** by @austiezr, with minor changes to support the dict.


itaise:
@austiezr  WDYT about that metric?

itaise:
@talagluck  can someone please CR this? thanks a lot!

Shinnnyshinshin:
HI @itaise this is Will from the GE Core Team. I had a look at the PR, and the work looks wonderful (well structured, commented, and error-handled). 

One request that I have is for some tests that will exercise the Metric, ensuring that it works as expected. The test can be added either at the `Metric` or `Expectation` level (or both if that's what you prefer).

Some examples can be found in the [QueryExpectations PR here](https://github.com/great-expectations/great_expectations/pull/5223). Would you be able to have a look and let me know if you have any questions?

itaise:
@Shinnnyshinshin  I added some spark tests as we discussed on slack.
can you please review? 
thanks!

Shinnnyshinshin (approved):
Thank you @itaise for your work, and for the discussion on Slack :) The PR looks great, and will allow #6373 to be merged as well. LGTM!

PR #6003, opened 2022-09-13
[MAINTENANCE] Remove `dgtest` from dev pipeline

cdkini:
- EOL `dgtest` dependency graph test runner in favor of `pytest`
- This marks the full removal of `dgtest` as a test strategy - moving forward, we will fully rely on `pytest` and `unit/integration/e2e` marking strategies

### Definition of Done


azure-pipelines-dev.yml
```diff
@@ -326,41 +320,13 @@ stages:
...
-
-            displayName: 'dgtest-general-overrides'
-
-          - script: |
-              # These are tests that are specific to Great Expectations Cloud.
-              # In order to ensure coverage, we run them during each CI/CD cycle.
-              pytest tests --cloud -m "cloud and not e2e"
-
-            displayName: 'dgtest-cloud-overrides'
+            displayName: 'pytest'
```
anthonyburdi, line 330: Should we still run these "cloud and not e2e" tests? It looks like only "cloud AND e2e" tests were moved to a separate pipeline recently by @billdirks.

anthonyburdi: It looks like we are running the unit test cloud tests in a step above, should we also be running tests marked cloud + integration?

anthonyburdi (commented):
One question about --cloud tests but otherwise looks good!

azure-pipelines-dev.yml
cdkini: you are correct! added back this stage

anthonyburdi (approved):
LGTM!

anthonyburdi:
Force merged since it is failing on the run of dgtest which is being removed.

PR #6006, opened 2022-09-13
[MAINTENANCE] Allow `CloudDataContext` to retrieve and initialize its own project config

cdkini:
- Previously, `CloudDataContext` required that a `project_config` be provided in the constructor as part of instantiation.
  - This was used when `DataContext` instantiated it's instance of `self._data_context` as part of our context-composition pattern.
- In order to make the `CloudDataContext` stand on its own, we need it to have the ability to reach out to the Cloud backend and retrieve the project config.
- By making `project_config` optional and adding all the same `ge_cloud_` args to the constructor that are found in `DataContext`, we can conditionally request the backend for information.
- With this implementation, we should be able to continue using `DataContext` as our primary Cloud-backed context but make strides towards getting `CloudDataContext` to reach full parity.

### Definition of Done


great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -300,11 +300,14 @@ def __init__(
         # initialize runtime_environment as empty dict if None
         runtime_environment = runtime_environment or {}
         if self._ge_cloud_mode:
+            assert ge_cloud_config is not None
```
NathanFarmer: How does this assertion square with the `if project_config is None` conditional in the `CloudDataContext` constructor? It seems like we have a way to access a default config or something based on `CloudDataConext.retrieve_data_context_config_from_ge_cloud`. Maybe I'm not understanding something, but why would we want to be strict about raising an error here?

NathanFarmer (commented):
One question from me!

great_expectations/data_context/data_context/base_data_context.py
NathanFarmer: I see now that we have both `project_config` and `ge_cloud_config`. What is the difference?

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -300,11 +300,20 @@ def __init__(
...
+            if ge_cloud_config:
+                ge_cloud_base_url = ge_cloud_config.base_url
+                ge_cloud_access_token = ge_cloud_config.access_token
+                ge_cloud_organization_id = ge_cloud_config.organization_id
             self._data_context = CloudDataContext(
                 project_config=project_config,
                 runtime_environment=runtime_environment,
                 context_root_dir=context_root_dir,  # type: ignore[arg-type]
-                ge_cloud_config=ge_cloud_config,  # type: ignore[assignment,arg-type]
+                ge_cloud_base_url=ge_cloud_base_url,
```
anthonyburdi, line 314: Any reason to pass these separately instead of as a GeCloudConfig object?

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -48,10 +49,12 @@ class CloudDataContext(AbstractDataContext):
 
     def __init__(
         self,
-        project_config: Union[DataContextConfig, Mapping],
-        context_root_dir: str,
-        ge_cloud_config: GeCloudConfig,
+        project_config: Optional[Union[DataContextConfig, Mapping]] = None,
+        context_root_dir: Optional[str] = None,
         runtime_environment: Optional[dict] = None,
+        ge_cloud_base_url: Optional[str] = None,
```
anthonyburdi, line 55: Same question - could we pass an optional GeCloudConfig instead of these 3 params? or is there a reason to separate?

anthonyburdi (commented):
Just some questions

great_expectations/data_context/data_context/base_data_context.py
cdkini: just consistency with the current `DataContext` constructor - happy to alter this if we see fit but was aiming for parity over anything else.

Thankfully, this is not exactly public API (in the traditional sense) so we have a lot of flexibility.

Let me know your thoughts!

great_expectations/data_context/data_context/cloud_data_context.py
cdkini: Same note as above - consistency with `DataContext` constructor

great_expectations/data_context/data_context/base_data_context.py
anthonyburdi: That makes sense! Thanks for clarifying.

Shinnnyshinshin (approved):
Love it. LGTM

PR #6007, opened 2022-09-13
[MAINTENANCE] Refactor external HTTP request logic into a `Session` factory function

cdkini:
- Encapsulate all external HTTP request logic into a `Session` factory function in a new `http` module
- Add HTTP adapters to configure timeout/retry logic
- Add unit tests

### Definition of Done


great_expectations/core/http_handler.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def _init_headers(self, access_token: str) -> dict:
+        headers = {
+            "Content-Type": "application/vnd.api+json",
+            "Authorization": f"Bearer {access_token}",
+            "Gx-Version": __version__,
+        }
+        return headers
+
+    def get(self, url: str, params: Optional[dict] = None) -> requests.Response:
```
cdkini: How do we abstract away the very specific `requests.Response` return? Perhaps a wrapper `HTTPResponse` class that interacts with the underlying `requests.Response` type?

anthonyburdi: I think that is a great idea.

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -188,6 +187,12 @@ def __init__(
         if not self._suppress_store_backend_id:
             _ = self.store_backend_id
 
+        if http_handler is None:
```
donaldheppner: I don't see where we are passing this in as a parameter.

For now, please consider removing http_handler as a parameter from __init__ and if/when we do add it back, it should be a required dependency

donaldheppner (changes requested):

great_expectations/core/http_handler.py
```diff
@@ -0,0 +1,77 @@
...
+        headers = {
+            "Content-Type": "application/vnd.api+json",
+            "Authorization": f"Bearer {access_token}",
+            "Gx-Version": __version__,
+        }
+        return headers
+
+    def _make_request(
+        self,
+        fn: Callable,
```
billdirks: Can you annotate what this callable's inputs and outputs? That is `Callable[[input type list], output type]`.

great_expectations/core/http_handler.py
```diff
@@ -0,0 +1,77 @@
+from typing import Callable, Optional
```
billdirks: I'm not sure what this gives us over the requests library (or alternatives). We could have a helper method that returns a configured requests [Sessions object](https://requests.readthedocs.io/en/latest/user/advanced/#session-objects) with headers and timeout configurable with defaults set to what you have here. It seems like less to maintain and more fully featured.

billdirks (changes requested):
Thanks for looking into this. I agree we can get some benefit my having some type of configured http request management. I think leveraging the requests library more might be more maintainable since it has a lot of features built in we'd need to implement or write proxies for here.

great_expectations/core/http.py
```diff
@@ -0,0 +1,60 @@
...
+
+from great_expectations import __version__
+
+DEFAULT_TIMEOUT = 20
+
+
+class TimeoutHTTPAdapter(HTTPAdapter):
+    # https://stackoverflow.com/a/62044100
+    # Session-wide timeouts are not supported by requests
+    # but are discussed in detail here: https://github.com/psf/requests/issues/3070
```
Shinnnyshinshin, line 12: ❤️

Shinnnyshinshin (approved):
Totally loving this, and appreciating the discussions that went into this between you and @billdirks. 🙇🏼 thank you both.

great_expectations/core/http.py
```diff
@@ -0,0 +1,60 @@
...
+    def __init__(self, *args, **kwargs):
+        if "timeout" in kwargs:
+            self.timeout = kwargs["timeout"]
```
billdirks: These 2 line are equivalent to `self.timeout = kwargs.pop("timeout")`
Can you call this `default_timeout`? Do we always want to set it to something like `DEFAULT_TIMEOUT`?

great_expectations/core/http.py
```diff
@@ -0,0 +1,60 @@
...
+            del kwargs["timeout"]
+        super().__init__(*args, **kwargs)
+
+    def send(self, request, **kwargs):
```
billdirks: Type annotations on request and return type please.

great_expectations/core/http.py
```diff
@@ -0,0 +1,60 @@
...
+        timeout = kwargs.get("timeout")
+        if timeout is None:
+            kwargs["timeout"] = self.timeout
```
billdirks: `self.timeout` may not be defined since it only occurs in an `if` in `__init__`. 

This is equivalent to `kwargs["timeout"] = kwargs.get("timeout", self.timeout)`

great_expectations/core/http.py
```diff
@@ -0,0 +1,60 @@
...
+        return super().send(request, **kwargs)
+
+
+def session_factory(
```
billdirks: nit: I prefer `get_session` or `create_session` since the result isn't a factory but a session object.

tests/core/test_http.py
```diff
@@ -0,0 +1,33 @@
...
+
+from great_expectations.core.http import DEFAULT_TIMEOUT, session_factory
+
+
+@pytest.mark.unit
+def test_session_factory_contains_appropriate_headers(
+    ge_cloud_access_token: str, request_headers: Dict[str, str]
+) -> None:
+    session = session_factory(access_token=ge_cloud_access_token)
+    for key, val in request_headers.items():
```
billdirks, line 13: Do you want to test for equality or are you expecting there might be more headers in the `session.headers`?

great_expectations/core/http.py
```diff
@@ -0,0 +1,60 @@
+import requests
+from requests.adapters import HTTPAdapter, Retry
+
+from great_expectations import __version__
+
+DEFAULT_TIMEOUT = 20
+
+
+class TimeoutHTTPAdapter(HTTPAdapter):
```
billdirks: Do you want to make this private by prepending a `_`?

billdirks (commented):
Thanks for this! Looks good overall. I have some small comments.

great_expectations/core/http.py
cdkini: done!

great_expectations/core/http.py
cdkini: added `pop`. Also added a default.

Didn't want to modify this too much since I straight up pulled it from SO but I think these changes are appropriate

great_expectations/core/http.py
cdkini: done!

tests/core/test_http.py
cdkini: there are more headers within the session by default - can go either way here but opted to keep the defaults in for now

billdirks: sounds good.

billdirks (approved):
LGTM, thanks!

donaldheppner (approved):

tests/data_context/cloud_data_context/test_checkpoint_crud.py
```diff
@@ -184,6 +183,7 @@ def test_cloud_backed_data_context_add_checkpoint(
         )
 
         mock_post.assert_called_with(
+            mock.ANY,  # requests.Session object
```
Kilo59, line 186: Trying to catch up. Why are using `mock.ANY` instead of the old `shared_called_with_request_kwargs` fixture?
Did we decide this was over-testing?

Kilo59 (approved):
Looks good to me.
I appreciate the simplicity, here.
This is something that would be very easy to over-engineer.

PR #6009, opened 2022-09-13
[FEATURE] context.save_datasource

anthonyburdi:
- Add a new `datasource_save()` method to all Data Contexts. This method takes a datasource, and can be used with the return value of context.test_yaml_config() (which is an initialized datasource).
- It stores the datasource config in the configured store, and uses the stored config to re-initialize a datasource which is then populated in the data context cache, config and returned.

### Definition of Done

Thank you for submitting!


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -461,6 +461,42 @@ def set_config(self, project_config: DataContextConfig) -> None:
...
+            key=None, value=datasource_config
+        )
+        # Use the updated datasource config, since the store may populate additional info on update.
+        self.config.datasources[datasource_name] = updated_datasource_config_from_store  # type: ignore[index,assignment]
+
+        # Also use the updated config to initialize a datasource for the cache and overwrite the existing datasource.
+        substituted_config = self._perform_substitutions_on_datasource_config(
+            updated_datasource_config_from_store
+        )
+        updated_datasource: Datasource = self._instantiate_datasource_from_config(
```
billdirks: I thought we explored the idea of creating a new datasource and returning it but then decided that that might be confusing because there might be references to the datasource elsewhere (eg in a validator) and we might need to guard against using the old, "dirty" datasource. I thought we settled on just updating in place. Does the line above update in place and, if so, can we do that instead of creating a new datasource?

anthonyburdi: Good call. I was going to work on that separately, but since this PR is going in after the release it makes sense to do that in the current PR. I'll make that change and add some related tests.

anthonyburdi:
Note: This also needs to change to save_datasource()

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: On further discussion with @donaldheppner we decided to forego that change in the current PR.

donaldheppner: Is the type hint for this line accurate? It is inconsistent with the return type, and I would guess instantiate class from config could return either a base or legacy datasource

tests/data_context/cloud_data_context/test_datasource_crud.py
```diff
@@ -358,3 +369,281 @@ def test_cloud_data_context_add_datasource(
...
+            LegacyDatasource, BaseDatasource
+        ] = context.save_datasource(datasource)
+
+        # Make sure the datasource config got into the context config
+        assert len(context.config.datasources) == 1
+        assert (
+            context.config.datasources[datasource_name] == datasource_config_with_names
+        )
+
+        cached_datasource = context._cached_datasources[datasource_name]
```
donaldheppner: this should occur after asserting the len of _cached_datasources

tests/data_context/cloud_data_context/test_datasource_crud.py
```diff
@@ -358,3 +369,281 @@ def test_cloud_data_context_add_datasource(
...
+
+        # Make sure the datasource got into the cache
+        assert len(context._cached_datasources) == 1
+
+        # Make sure the stored and returned datasource is the same one as the cached datasource
+        assert id(updated_datasource) == id(cached_datasource)
+        assert updated_datasource == cached_datasource
+
+        # Make sure the stored and returned datasource is a different instance
```
donaldheppner: I think these assertions are unnecessary given the ones above

tests/data_context/cloud_data_context/test_datasource_crud.py
```diff
@@ -358,3 +369,281 @@ def test_cloud_data_context_add_datasource(
...
+        orig_datasource_dict = serializer.serialize(
+            datasourceConfigSchema.load(datasource.config)
+        )
+
+        for attribute in ("name", "execution_engine", "data_connectors"):
+            assert updated_datasource_dict[attribute] == orig_datasource_dict[attribute]
+
+
+@pytest.mark.unit
+def test_non_cloud_backed_data_context_save_datasource_overwrite_existing(
```
donaldheppner, line 449: There is a lot of duplicated code between this test and the one above, please factor it out and reference from both tests.

I am not clear from the description what the difference is between this test and the one above

tests/data_context/cloud_data_context/test_datasource_crud.py
```diff
@@ -358,3 +369,281 @@ def test_cloud_data_context_add_datasource(
...
+        pytest.param(
+            "empty_cloud_data_context",
+            CloudDataContext,
+            id="CloudDataContext",
+        ),
+    ],
+)
+@pytest.mark.cloud
+@pytest.mark.unit
+def test_cloud_backed_data_context_save_datasource_empty_store(
```
donaldheppner: Please limit this test to the stated description: specifically, it should cover the code path that exercises cloud mode (and not the others)

donaldheppner (changes requested):

tests/data_context/cloud_data_context/test_datasource_crud.py
anthonyburdi: This tests the overwrite or "update" functionality by first saving a datasource to the store, then saving another datasource with the same name and a slightly different config. It makes sure there is only 1 datasource in the store and that the config was updated (rather than creating a second datasource).

anthonyburdi: And good call on the duplicated code, coming up in a new commit.

anthonyburdi: So far I've moved some duplicated code into _save_datasource_assertions. Looking for other opportunities.

donaldheppner (approved):

PR #6076, opened 2022-09-23
[BUGFIX] numpy typing behind `if TYPE_CHECKING`

Kilo59:
The `numpy.typing` module was added until v `1.20` but our minimum numpy version is `1.18`.
https://numpy.org/devdocs/reference/typing.html#typing-numpy-typing

Move `numpy.typing` import behind an `if TYPE_CHECKING` block to prevent import issues on installs with `numpy` < `1.20`

See #5925


NathanFarmer (approved):

great_expectations/util.py
```diff
@@ -1566,7 +1580,7 @@ def is_ndarray_decimal_dtype(
     return result
 
 
-def convert_ndarray_decimal_to_float_dtype(data: np.ndarray) -> np.ndarray:
+def convert_ndarray_decimal_to_float_dtype(data: "np.ndarray") -> "np.ndarray":
```
Kilo59: This one could be reverted. It's using `numpy` not `numpy.typing`.

cdkini (approved):

anthonyburdi:
Force merged to keep the release process on track (this code will be tested as part of the full suite before release).

PR #6077, opened 2022-09-23
[BUGFIX] Enable top-level parameters in Data Assistants accessed via dispatcher

NathanFarmer:
- GREAT-1276
- Enable top-level parameters in Data Assistants accessed via dispatcher

### Definition of Done


great_expectations/rule_based_profiler/data_assistant/onboarding_data_assistant.py
```diff
@@ -51,7 +51,8 @@ class OnboardingDataAssistant(DataAssistant):
             that is suspected to have quality issues may benefit from "flag_outliers".
         - include_column_names (list): A list containing the column names you wish to include.
         - exclude_column_names (list): A list containing the column names you with to exclude.
-        - include_column_name_suffixes (list): A list containing the column name suffixes you with to include.
+        - include_column_name_suffixes (list): A list containing the column name suffixes you wish to include.
```
alexsherstinsky, line 54: This will be an accurate aspect of documentation only if the method `context.assistants.<your_data_assistant_name_here>.run` will return these arguments as available.

great_expectations/rule_based_profiler/data_assistant/data_assistant_runner.py
```diff
@@ -303,7 +303,7 @@ def _get_method_signature_parameters_for_domain_type_directives(
                         annotation=accessor_method_return_type,
                     )
                     domain_type_attribute_name_to_parameter_map[key] = parameter
-                elif parameter.default != property_value:
+                elif parameter.default != property_value and property_value is not None:
```
alexsherstinsky, line 306: @NathanFarmer Not sure if this is going to work (maybe, just not sure), the reason being is that `None` in the constructor arguments may mean that the actual constructor code adds certain usecase-specific defaults (e.g., in `great_expectations/rule_based_profiler/domain_builder/categorical_column_domain_builder.py`).  Happy to discuss.  Thanks!

alexsherstinsky:
@NathanFarmer It might be easier to implement the customizability of general `ColumnDomainBuilder` arguments at the individual `Rule` level as well as to allow overrides at the global `DataAssistant` level (we would have to be careful in defining and implementing the reconciliation logic).  Happy to discuss.  Thanks!

great_expectations/rule_based_profiler/data_assistant/onboarding_data_assistant.py
NathanFarmer: I agree. Since we are documenting this in `OnboardingDataAssistant` (as opposed to `DataAssistant`), we should be able to guarantee what arguments are available.

great_expectations/rule_based_profiler/data_assistant/data_assistant_runner.py
NathanFarmer: @alexsherstinsky this allows the user to override the defaults globally (all rules), unless the user wants to pass `None` explicitly to the `run` method. This is because we are always overriding `None` in the `categorical_column_domain_builder.py`:
```
        if exclude_column_name_suffixes is None:
            exclude_column_name_suffixes = [
                "_id",
                "_ID",
            ]
```

Is this the appropriate place to be setting these defaults? It would seem more appropriate to set them in the `OnboardingDataAssistant` itself.

alexsherstinsky: @NathanFarmer I feel that the way it is now works, but we can experiment with alternatives.  The situation is that if we put the defaults into the constructor arguments, this could entail conflicts with serialization.  At least logically, the current implementation customizes the parent class' constructor with the current class' requirements.  Requiring such customization at the `OnboardingDataAssistant` might also work, but will not be ideal, because the customization is actually appropriate at the lower level (i.e., categorical rule behaves the same way for all `DataAssistant` implementations).  Happy to discuss.  Thanks!

NathanFarmer:
@cla-bot check

NathanFarmer:
@cla-bot check

great_expectations/rule_based_profiler/data_assistant/data_assistant_runner.py
NathanFarmer: @alexsherstinsky per our offline conversation, we have decided to leave the defaults in the `DomainBuilder` for now, with possible changes in the future. I have added tests that demonstrate that the `and property_value is not None` conditional leads to our intended behavior (also see updated comment lines 307-308).

great_expectations/rule_based_profiler/data_assistant/data_assistant_runner.py
```diff
@@ -303,8 +303,9 @@ def _get_method_signature_parameters_for_domain_type_directives(
                         annotation=accessor_method_return_type,
                     )
                     domain_type_attribute_name_to_parameter_map[key] = parameter
-                elif parameter.default != property_value:
-                    # For now, prevent customization if default values conflict.  In the future, enable at "Rule" level.
+                elif parameter.default != property_value and property_value is not None:
+                    # For now, prevent customization if default values conflict unless the default DomainBuilder value
+                    # is None. In the future, enable at "Rule" level.
```
alexsherstinsky, line 308: ❤️

alexsherstinsky (approved):
LGTM!  Very elegant -- small change and useful new tests!  Thanks!

tests/rule_based_profiler/data_assistant/test_data_assistant_runner.py
```diff
@@ -0,0 +1,102 @@
...
+        assert (False, f"context.assistants.onboarding.run raise an exception '{exc}'")
+
+    with pytest.raises(TypeError) as e:
+        context.assistants.onboarding.run(
+            batch_request=batch_request, non_existent_parameter="break_this"
+        )
+    assert (
+        e.value.args[0]
+        == "run() got an unexpected keyword argument 'non_existent_parameter'"
+    )
```
Kilo59, line 41: I think you can do this in one go by using `pytest.raises()` matching.
https://docs.pytest.org/en/stable/reference/reference.html?highlight=raises#pytest.raises

```python
    with pytest.raises(TypeError, match="run() got an unexpected keyword argument 'non_existent_parameter"):
        context.assistants.onboarding.run(
            batch_request=batch_request, non_existent_parameter="break_this"
        )
```

Then you shouldn't need the second assert.

NathanFarmer: I am adding this to this PR: https://github.com/great-expectations/great_expectations/pull/6091

PR #6087, opened 2022-09-26
[DOCS] Support for named snippets in documentation

cdkini:
- Custom Remark plugin to add dynamic snippets (referenced by name) in our documentation

### Definition of Done


scripts/remark-named-snippets/snippet.js
```diff
@@ -0,0 +1,130 @@
...
+    text = text.trim()
+    if (text.endsWith("#")) {
+        text = text.substring(0, text.length - 1)
+    }
+    lines = text.split("\n")
+    // TODO: This logic here should probably be cleaned up with regex? Let's also remove the resulting whitespace
+    lines = lines.filter(l => !(l.includes("<snippet") || l.includes("snippet>")))
+    text = lines.join("\n")
+    return text.trim()
+}
```
dctalbot:
```suggestion
function sanitizeText(text: string) {
  text = text.trim()
  if (text.endsWith("#")) {
    text = text.substring(0, text.length - 1)
  }
  return (
    text
      .split("\n")
      // TODO: This logic here should probably be cleaned up with regex? Let's also remove the resulting whitespace
      .filter((l) => !(l.includes("<snippet") || l.includes("snippet>")))
      .join("\n")
      .trim()
  )
}
```

Suggested refactor - the main issue with prev implementation is that `lines` was declared as a global var.

Also, I suggest taking care of the TODO now. I don't think a regex is necessary, and removing whitespace should be simple enough if you want to do that

scripts/remark-named-snippets/snippet.js
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Parses an input directory using an HTML parser to collect snippets.
+ *
+ * @param {string} dir - The directory to parse for snippet definitions.
+ * @returns {object[]} A list of snippet objects parsed from the input directory.
+ */
+function parseDirectory(dir) {
+    let files = glob.sync(dir + "/**/*.py")
```
dctalbot:
```suggestion
    const files = glob.sync(dir + "/**/*.py")
```
Lots of cases where you're using `let` instead of `const`. Lmk if you'd like me to run the code through a linter - it can fix this sort of thing automatically

dctalbot (changes requested):

scripts/remark-named-snippets/index.js
```diff
@@ -0,0 +1,72 @@
...
+            const res = /^name=(?<snippetName>.+?)$/.exec(
+                nameMeta
+            );
+
+            let name = res.groups.snippetName
+            if (!name) {
+                throw new Error(`Unable to parse named reference ${nameMeta}`);
+            }
+
+            name = eval(name) // Remove any surrounding quotes
```
dctalbot: we should avoid using `eval` unless absolutely necessary. If we can't get around this, add a comment to suppress the linter e.g.`// eslint-disable-next-line`

dctalbot (changes requested):

dctalbot (approved):

PR #6091, opened 2022-09-26
[MAINTENANCE] Only update existing `rendered_content` if rendering does not fail with new `InlineRenderer` failure message

NathanFarmer:
- GREAT-1282
- Current behavior is for any `ExpectationSuite`s or `ExpectationValidationResult`s in stores to re-render `rendered_content` upon `get` if `include_rendered_content` is `True`. This will ensure that un-rendered objects are rendered and changes to renderers are reflected in historical `ExpectationValidationResult`s.
- This could be an issue if a user has a custom `Expectation` and renderer that they have changed over time and haven’t taken backwards compatibility into account. The historical `ExpectationValidationResult`s could fail to render. In this case we should not overwrite the existing `rendered_content`.

### Definition of Done


great_expectations/expectations/expectation.py
```diff
@@ -209,25 +209,34 @@ def _validate(
...
+                "params": params_with_json_schema,
+                "schema": {"type": "com.superconductive.rendered.string"},
+            }
         )
         rendered = RenderedAtomicContent(
-            name="atomic.prescriptive.kwargs",
+            name="atomic.prescriptive.failed",
             value=value_obj,
-            value_type="UnknownType",
+            value_type="StringValueType",
```
Kilo59, line 240: Can we use real types here?
`str` or `Any` etc.

NathanFarmer: I don't think so. These are internal renderer block types. We have `StringValueType`, `TableType`, and `GraphType` and they get serialized for use by Cloud. An `Enum` could probably be used, but this is contributor facing so I'm not sure how we would feel about that.

cdkini, line 240: Nit - not necessary for this PR but would love an enum here.

great_expectations/expectations/expectation.py
```diff
@@ -660,6 +669,38 @@ def _get_observed_value_from_evr(self, result: ExpectationValidationResult) -> s
         else:
             return "--"
 
+    @classmethod
+    @renderer(renderer_type="atomic.diagnostic.failed")
+    def _atomic_diagnostic_failed(
+        cls,
+        configuration=None,
```
cdkini: Could we add a type annotation for the args and return?

great_expectations/render/renderer/inline_renderer.py
```diff
@@ -111,31 +116,33 @@ def _get_atomic_rendered_content_from_renderer_names(
...
+                        f'Renderer "{failed_renderer_name}" will be used to render prescriptive content for ExpectationConfiguration.'
+                    )
+                else:
+                    failed_renderer_name = FailedAtomicRendererName.DIAGNOSTIC
+                    logger.info(
+                        f'Renderer "{failed_renderer_name}" will be used to render diagnostic content for ExpectationValidationResult.'
+                    )
+
+                renderer_rendered_content = (
+                    InlineRenderer._get_renderer_atomic_rendered_content(
```
cdkini, line 132: I realize this isn't caused by this PR but any way we can avoid accessing this private method?

great_expectations/render/types/__init__.py
```diff
@@ -1,14 +1,47 @@
...
+
+    def __hash__(self: AtomicRendererPrefix):
+        return hash(self.value)
+
+
+class FailedAtomicRendererName(Enum):
+    """Available failed atomic renderer names"""
+
+    PRESCRIPTIVE = ".".join([AtomicRendererPrefix.PRESCRIPTIVE.value, "failed"])
+    DIAGNOSTIC = ".".join([AtomicRendererPrefix.DIAGNOSTIC.value, "failed"])
```
cdkini, line 34: Can we use these enums in the decorators?

great_expectations/render/types/__init__.py
```diff
@@ -1,14 +1,47 @@
...
+    PRESCRIPTIVE = "atomic.prescriptive"
+    DIAGNOSTIC = "atomic.diagnostic"
+
+    def __eq__(self, other: Union[str, AtomicRendererPrefix]):
+        if isinstance(other, str):
+            return self.value.lower() == other.lower()
+        return self.value.lower() == other.value.lower()
+
+    def __hash__(self: AtomicRendererPrefix):
+        return hash(self.value)
```
cdkini: Can we add return annotations here?

Kilo59: @NathanFarmer 
On a related note. Right now `render/types` is not being type-checked, but it only has a couple of typing issues (4).

Could you remove it from the type-check exclude list and fix (or ignore) the 4 or so issues as part of this PR.
https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L97

And any other low error count modules you feel like fixing 😄

great_expectations/render/renderer/inline_renderer.py
NathanFarmer: I'm not following the reasoning for this. This is used in two places and shouldn't be considered public by anyone using the `InlineRenderer`.

great_expectations/render/types/__init__.py
NathanFarmer: I considered it, but I would like to create a follow-up PR that adds all other renderer names to an enum and then I can do this.

great_expectations/render/types/__init__.py
NathanFarmer: 🤦🏻 ✅

great_expectations/expectations/expectation.py
Kilo59: Why is `configuration` an optional keyword arg? It looks like this method would error if not provided.

NathanFarmer: I have updated this annotation, but I will be creating a follow-up PR that will be addressing all of the existing renderer signatures.

great_expectations/expectations/expectation.py
NathanFarmer: I'm going to add this as a task for this follow-up PR: https://github.com/great-expectations/great_expectations/pull/6112

great_expectations/render/types/__init__.py
NathanFarmer: Follow-up PR: https://github.com/great-expectations/great_expectations/pull/6112

great_expectations/core/expectation_validation_result.py
```diff
@@ -229,7 +229,17 @@ def render(self) -> None:
...
-        self.rendered_content = inline_renderer.get_rendered_content()
+        rendered_content: List[
+            RenderedAtomicContent
+        ] = inline_renderer.get_rendered_content()
+
+        self.rendered_content = (
+            inline_renderer.replace_or_keep_existing_rendered_content(
+                existing_rendered_content=self.rendered_content,
+                new_rendered_content=rendered_content,
+                failed_renderer_name="atomic.diagnostic.failed",
```
Kilo59, line 240: ## Followup
Should these (`renderer_types` / names) be enums as well?

great_expectations/render/renderer/inline_renderer.py
Kilo59: If the method isn't meant for public use then I think the private method is appropriate and even preferable.
Is there a reason this particular method should be part of the public API?

great_expectations/render/renderer/inline_renderer.py
```diff
@@ -56,22 +59,21 @@ def _get_atomic_rendered_content_for_object(
...
-        legacy_renderer_prefix: str
+        renderer_prefix: AtomicRendererPrefix
         if isinstance(render_object, ExpectationConfiguration):
             expectation_type = render_object.expectation_type
-            renderer_prefix = "atomic.prescriptive"
+            renderer_prefix = AtomicRendererPrefix.PRESCRIPTIVE
         elif isinstance(render_object, ExpectationValidationResult):
             expectation_type = render_object.expectation_config.expectation_type
-            renderer_prefix = "atomic.diagnostic"
+            renderer_prefix = AtomicRendererPrefix.DIAGNOSTIC
```
Kilo59, line 68:
```python
 if isinstance(render_object, ExpectationConfiguration):
            expectation_type = render_object.expectation_type
            renderer_prefix = AtomicRendererPrefix.PRESCRIPTIVE.value

```

You could assign the value here and then you don't need to make the `render_prefix.value` call.

Kilo59 (approved):
Couple non-blocking suggestions/questions.

great_expectations/core/expectation_validation_result.py
NathanFarmer: Yes, this is WIP on PR #6112

great_expectations/render/renderer/inline_renderer.py
NathanFarmer: This is WIP on PR #6112

great_expectations/render/renderer/inline_renderer.py
NathanFarmer: @Kilo59 I agree that it should stay this way!

PR #6092, opened 2022-09-26
[MAINTENANCE] Add datasources to ConfigurationBundle

anthonyburdi:
- Add datasources to ConfigurationBundle.
- Convert existing ConfigurationBundle tests to unit tests and increase coverage.

### Definition of Done


great_expectations/data_context/cloud_migrator.py
```diff
@@ -101,6 +119,11 @@ class ConfigurationBundleSchema(Schema):
     _data_context_variables = fields.Nested(
         DataContextConfigSchema, allow_none=False, data_key="data_context_variables"
     )
+    _datasources = fields.List(
+        fields.Nested(DatasourceConfigSchema, allow_none=True, required=True),
+        required=True,
+        data_key="datasources",
+    )
```
cdkini: Do we need the prefix `_` for schema fields? I think the access modifier aspect of the bundle can be determined with the runtime object no?

cdkini: I guess I missed this earlier but why does our Marshmallow schema need the `_` prefix?

tests/data_context/test_cloud_migrator_configuration_bundle.py
```diff
@@ -1,246 +1,265 @@
...
+            "data_connectors": {},
+        }
+        return DatasourceConfig(**datasource_config_dict)
+
+
+class DummyDatasource:
+    pass
+
+
+class StubBaseDataContext:
```
cdkini, line 69: How do we ensure that we reuse this stub logic elsewhere? As a dev, I may not know that someone else on the team has written an implementation that fulfills my needs for a given test.

I know this is specific to `ConfigurationBundle` but just seeing how to make this a more maintainable process.

tests/data_context/test_cloud_migrator_configuration_bundle.py
```diff
@@ -1,246 +1,265 @@
...
+        context: BaseDataContext = stub_base_data_context
+
+        config_bundle = ConfigurationBundle(context)
+
+        assert config_bundle.is_usage_stats_enabled()
+        assert config_bundle._data_context_variables is not None
+        assert len(config_bundle._expectation_suites) == 1
+        assert len(config_bundle._checkpoints) == 1
+        assert len(config_bundle._profilers) == 1
+        assert len(config_bundle._validation_results) == 1
```
cdkini: I'm adding properties for all these attrs as they need to be accessed for printing logic in the migrator

tests/data_context/test_cloud_migrator_configuration_bundle.py
```diff
@@ -1,246 +1,265 @@
...
+
+    @pytest.mark.unit
+    def test_configuration_bundle_created_usage_stats_disabled(
```
cdkini, line 186: Since all this logic is Cloud-specific, can we decorate with `@pytest.mark.cloud` throughout?

anthonyburdi: Oh good call!

tests/data_context/test_cloud_migrator_configuration_bundle.py
anthonyburdi: Nice - I didn't have a use case for properties yet but that makes sense. With those we can also remove the _ prefix from the marshmallow schemas.

great_expectations/data_context/cloud_migrator.py
anthonyburdi: Not necessary if we add accessors, sounds like you are doing this already for the printing logic so we can swap these out as well.

anthonyburdi: See other comments (can remove the prefix with accessors).

tests/data_context/test_cloud_migrator_configuration_bundle.py
anthonyburdi: I think maybe the best way is to use them in fixtures, and move those fixtures to the right level based on where they are used (lowest level needed for existing tests). What would you recommend @Kilo59?

cdkini (approved):

great_expectations/data_context/cloud_migrator.py
anthonyburdi: Added! (and modified the schema)

anthonyburdi: Added! (and modified the schema)

tests/data_context/test_cloud_migrator_configuration_bundle.py
anthonyburdi: Added properties! (and modified the schema and this test)

tests/data_context/test_cloud_migrator_configuration_bundle.py
anthonyburdi: Done!

tests/data_context/test_cloud_migrator_configuration_bundle.py
anthonyburdi: (Note I'm already using StubBaseDataContext in a few fixtures in this file)

PR #6099, opened 2022-09-27
[MAINTENANCE] initial Datasource typings

Kilo59:
- start type checking `datasources`
  - with the exception of `datasource/data_connector/configured_asset_sql_data_connector.py` (had 47 typing issues and PR was already pretty big)
- add typing overloads to `utils.deep_filter_properties_iterable()`

This PR is almost purely typing changes.
There are a few very minor runtime changes that I've explicitly called out with comments.

It's possible I've identified a potential bug in `s3_batch_kwargs_generator.py`
https://github.com/great-expectations/great_expectations/pull/6099#pullrequestreview-1122012459

### Definition of Done


great_expectations/datasource/batch_kwargs_generator/s3_batch_kwargs_generator.py
```diff
@@ -158,9 +159,9 @@ def _get_iterator(
 
     def _build_batch_kwargs_path_iter(
         self, path_list, reader_options=None, limit=None
-    ) -> None:
+    ) -> Iterable[S3BatchKwargs]:
         for path in path_list:
-            yield self._build_batch_kwargs(
+            yield self._build_batch_kwargs(  # type: ignore[call-arg] # no parameters `reader_options` or `limit`
                 path, reader_options=reader_options, limit=limit
             )
```
Kilo59, line 166: ## Potential bug

This method call looks like it would fail due to unrecognized keyword arguments.

great_expectations/datasource/data_connector/file_path_data_connector.py
```diff
@@ -98,7 +98,7 @@ def _get_data_reference_list_from_cache_by_data_asset_name(
             )
         )
 
-        if len(self.sorters) > 0:
+        if self.sorters:
```
Kilo59, line 101: Here's a runtime change.
Because `self.sorters` could be `None`.

great_expectations/datasource/data_connector/file_path_data_connector.py
```diff
@@ -169,7 +169,7 @@ def _get_batch_definition_list_from_batch_request(
...
```
Kilo59, line 172: runtime change

great_expectations/datasource/data_connector/runtime_data_connector.py
```diff
@@ -39,7 +39,7 @@ def __init__(
         self,
         name: str,
         datasource_name: str,
-        execution_engine: Optional[ExecutionEngine] = None,
```
Kilo59, line 42: Another runtime change.
`ExecutionEngine` doesn't appear to be optional at all. The supertype has it as a positional argument and raises an error if it's not provided.

great_expectations/util.py
```diff
@@ -1213,15 +1214,80 @@ def filter_properties_dict(
...
+def deep_filter_properties_iterable(
+    properties: Union[dict, list, set, tuple, None] = None,
     keep_fields: Optional[Set[str]] = None,
     delete_fields: Optional[Set[str]] = None,
     clean_nulls: bool = True,
     clean_falsy: bool = False,
     keep_falsy_numerics: bool = True,
     inplace: bool = False,
-) -> Optional[Union[dict, list, set]]:
+) -> Optional[Union[dict, list, set, tuple]]:
```
Kilo59: Someone correct me if this is wrong, but the method appears to also return `tuple`s.

great_expectations/util.py
```diff
@@ -1213,15 +1214,80 @@ def filter_properties_dict(
...
+def deep_filter_properties_iterable(
+    properties: None,
+    keep_fields: Optional[Set[str]] = ...,
+    delete_fields: Optional[Set[str]] = ...,
+    clean_nulls: bool = ...,
+    clean_falsy: bool = ...,
+    keep_falsy_numerics: bool = ...,
+    inplace: bool = ...,
+) -> None:
+    ...
```
Kilo59, line 1279: These overloads are purely for the typing system. They do not alter runtime behavior at all.
https://mypy.readthedocs.io/en/stable/more_types.html?highlight=overload#function-overloading

But TLDR, each of these overloads is specific to a combination of parameters. So `deep_filter_properties_iterable` on a `dict`, returns a `dict`, `deep_filter_properties_iterable` on a `list` always returns a `list` etc.
If any of these signatures look incorrect, let me know.

Happy to walk through this (or anything else) synchronously.

great_expectations/datasource/batch_kwargs_generator/s3_batch_kwargs_generator.py
```diff
@@ -158,9 +159,9 @@ def _get_iterator(
 
     def _build_batch_kwargs_path_iter(
         self, path_list, reader_options=None, limit=None
-    ) -> None:
+    ) -> Iterable[S3BatchKwargs]:
         for path in path_list:
-            yield self._build_batch_kwargs(
+            yield self._build_batch_kwargs(  # type: ignore[call-arg] # no parameters `reader_options` or `limit`
```
cdkini, line 164: Thank you for this 🙇🏽

great_expectations/datasource/data_connector/file_path_data_connector.py
cdkini: good catch!

great_expectations/util.py
cdkini: ah very interesting, thanks for the explanation - would it be possible to use a generic here? Takes in an object of type `T` and returns `T`?

great_expectations/util.py
cdkini: It can take in any iterable and returns that same iterable type

cdkini (approved):

great_expectations/util.py
Kilo59: @cdkini Yes, great point. Should have done that from the beginning. Also Should reduce the number of overloads from 4 -> 2

great_expectations/util.py
```diff
@@ -1214,55 +1215,19 @@ def filter_properties_dict(
...
-    properties: set,
-    keep_fields: Optional[Set[str]] = ...,
-    delete_fields: Optional[Set[str]] = ...,
-    clean_nulls: bool = ...,
-    clean_falsy: bool = ...,
-    keep_falsy_numerics: bool = ...,
-    inplace: bool = ...,
-) -> set:
-    ...
+T = TypeVar("T", bound=Union[dict, list, tuple, set])
```
Kilo59: This change leads to typing errors when assigning to `properties` in the body of the function.
A bit confused as to why. I would have expected the type-narrowing to convince the checker that `properties` is either a `dict`, `tuple`, etc. when the assignment is performed.

I also tried a few variations on this but never got all the type-errors to go away.

great_expectations/util.py
Kilo59: I expected this to work but keep getting `assignment` errors in the body of the function when trying to use generics.
https://github.com/great-expectations/great_expectations/pull/6099#pullrequestreview-1122468188

Going to revert this for now, but I would prefer to do these using generics.

PR #6102, opened 2022-09-27
[FEATURE] Send user validation results to Cloud backend during migration

cdkini:
- Ensure that we iterate through validation results and send them in sequential requests

### Definition of Done

### DIFF
--- a/great_expectations/data_context/migrator/cloud_migrator.py
+++ b/great_expectations/data_context/migrator/cloud_migrator.py
@@ -25,6 +25,7 @@ from great_expectations.data_context.migrator.configuration_bundle import (
 )
 from great_expectations.data_context.store.ge_cloud_store_backend import (
     AnyPayload,
+    ErrorPayload,
     GeCloudRESTResource,
     GeCloudStoreBackend,
     construct_json_payload,
@@ -44,7 +45,7 @@ class SendValidationResultsErrorDetails:
 
 
 class MigrationResponse(NamedTuple):
-    message: Optional[str]
+    message: str
     status_code: int
     success: bool
 
@@ -272,19 +273,18 @@ class CloudMigrator:
         success = response.ok
         status_code = response.status_code
 
-        response_json: AnyPayload
-        try:
-            response_json = response.json()
-        except requests.exceptions.JSONDecodeError:
-            response_json = cast(AnyPayload, {})
-            success = False
-
-        print(response_json)
-
         # TODO: Handle success/failure cases and parse errors from Cloud responses
         message = ""
         if not success:
-            message = "Something went wrong!"
+            try:
+                response_json = cast(ErrorPayload, response.json())
+                errors = response_json.get("errors", [])
+                for error in errors:
+                    detail = error.get("detail")
+                    if detail:
+                        message += f"{detail}\n"
+            except requests.exceptions.JSONDecodeError:
+                message = "Something went wrong!"
 
         return MigrationResponse(
             message=message, status_code=status_code, success=success
--- a/great_expectations/data_context/migrator/cloud_migrator.py
+++ b/great_expectations/data_context/migrator/cloud_migrator.py
@@ -217,6 +217,7 @@ class CloudMigrator:
             attributes_key="bundle",
             attributes_value=serialized_bundle,
         )
+        print(response)
         # TODO: Handle success/failure cases
 
     def _prepare_validation_results(self, serialized_bundle: dict) -> List[dict]:
@@ -243,6 +244,7 @@ class CloudMigrator:
                 attributes_key=attributes_key,
                 attributes_value=validation_result,
             )
+            print(response)
             # TODO: Handle success/failure cases
 
     def _post_to_cloud_backend(
@@ -276,6 +278,8 @@ class CloudMigrator:
         except requests.exceptions.JSONDecodeError:
             response_json = cast(AnyPayload, {})
             success = False
+
+        print(response_json)
 
         # TODO: Handle success/failure cases and parse errors from Cloud responses
         message = ""
--- a/great_expectations/data_context/migrator/cloud_migrator.py
+++ b/great_expectations/data_context/migrator/cloud_migrator.py
@@ -1,7 +1,7 @@
 """TODO: Add docstring"""
 import logging
 from dataclasses import dataclass
-from typing import List, Optional, Tuple, cast
+from typing import List, Optional, cast
 
 import requests
 
@@ -174,12 +174,10 @@ class CloudMigrator:
         serialized_validation_results = serialized_bundle.pop("validation_results")
 
         if not test_migrate:
-            return
-
-        self._send_bundle_to_cloud_backend(serialized_bundle=serialized_bundle)
-        self._send_validation_results_to_cloud_backend(
-            serialized_validation_results=serialized_validation_results,
-        )
+            self._send_bundle_to_cloud_backend(serialized_bundle=serialized_bundle)
+            self._send_validation_results_to_cloud_backend(
+                serialized_validation_results=serialized_validation_results,
+            )
 
     def _send_bundle_to_cloud_backend(self, serialized_bundle: dict) -> None:
         self._post_to_cloud_backend(
--- a/great_expectations/data_context/migrator/cloud_migrator.py
+++ b/great_expectations/data_context/migrator/cloud_migrator.py
@@ -1,7 +1,7 @@
 """TODO: Add docstring"""
 import logging
 from dataclasses import dataclass
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, cast
 
 import requests
 
@@ -26,12 +26,10 @@ from great_expectations.data_context.store.ge_cloud_store_backend import (
     GeCloudStoreBackend,
     construct_json_payload,
     construct_url,
-    get_user_friendly_error_message,
 )
 from great_expectations.data_context.types.resource_identifiers import (
     ValidationResultIdentifier,
 )
-from great_expectations.exceptions.exceptions import GeCloudError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


great_expectations/data_context/migrator/cloud_migrator.py
```diff
@@ -143,81 +149,142 @@ def migrate_validation_result(
...
+
         if not test_migrate:
-            configuration_bundle_serializer = ConfigurationBundleJsonSerializer(
-                schema=ConfigurationBundleSchema()
-            )
-            configuration_bundle_response: AnyPayload = self._send_configuration_bundle(
-                configuration_bundle, configuration_bundle_serializer
+            self._send_configuration_bundle(serialized_bundle=serialized_bundle)
+            self._send_validation_results(
+                serialized_validation_results=serialized_validation_results,
```
cdkini, line 175: General cleanup just to tighten up the flow a bit - I think this is still a bit fluid as I'm concurrently working on the printing ticket

great_expectations/data_context/migrator/cloud_migrator.py
```diff
@@ -143,81 +149,142 @@ def migrate_validation_result(
...
 
-    def _send_configuration_bundle(
+    def _serialize_configuration_bundle(
+        self, configuration_bundle: ConfigurationBundle
+    ) -> dict:
+        serializer = ConfigurationBundleJsonSerializer(
+            schema=ConfigurationBundleSchema()
+        )
+        serialized_bundle = serializer.serialize(configuration_bundle)
+        return serialized_bundle
```
cdkini, line 211: Placed the serializer within the actual method since it never changes but lmk if you disagree!

great_expectations/data_context/migrator/cloud_migrator.py
```diff
@@ -143,81 +149,142 @@ def migrate_validation_result(
...
+
+    def _send_configuration_bundle(self, serialized_bundle: dict) -> None:
+        print("Sending context configuration (step 2/4)")
+        response = self._post_to_cloud_backend(
+            resource_name="migration",
+            resource_type="migration",
+            attributes_key="bundle",
+            attributes_value=serialized_bundle,
+        )
+        # TODO: Handle success/failure cases
```
cdkini, line 222: The idea here is that we take `response` (which is a parsed `MigrationResponse`) and use that to determine what to print to the user - that logic can be a bit more convoluted and it depends on the Cloud payload so I'm holding off for the time being

great_expectations/data_context/migrator/cloud_migrator.py
```diff
@@ -143,81 +149,142 @@ def migrate_validation_result(
...
+        self, serialized_validation_results: List[dict]
+    ) -> None:
+        # 20220928 - Chetan - We want to use the static lookup tables in GeCloudStoreBackend
+        # to ensure the appropriate URL and payload shape. This logic should be moved to
+        # a more central location.
+        resource_type = GeCloudRESTResource.EXPECTATION_VALIDATION_RESULT
+        resource_name = GeCloudStoreBackend.RESOURCE_PLURALITY_LOOKUP_DICT[
+            resource_type
+        ]
+        attributes_key = GeCloudStoreBackend.PAYLOAD_ATTRIBUTES_KEYS[resource_type]
```
cdkini, line 238: feels icky

great_expectations/data_context/migrator/cloud_migrator.py
```diff
@@ -143,81 +149,142 @@ def migrate_validation_result(
...
+        status_code = response.status_code
 
+        response_json: AnyPayload
         try:
-            response = self._session.post(url, json=data)
-            response.raise_for_status()
-            return response.json()
+            response_json = response.json()
+        except requests.exceptions.JSONDecodeError:
+            response_json = cast(AnyPayload, {})
```
Kilo59: If we get a `JSONDecodeError` this will not be an `AnyPayload` object.

tests/data_context/migrator/test_cloud_migrator.py
```diff
@@ -53,6 +50,53 @@ def test__send_configuration_bundle_sends_valid_http_request(
     )
 
 
+@pytest.mark.unit
+@pytest.mark.cloud
+def test__send_validation_results_sends_valid_http_request():
+    # These values aren't actual creds but resemble values used in production
+    ge_cloud_base_url = "https://app.test.greatexpectations.io"
+    ge_cloud_organization_id = "229616e2-1bbc-4849-8161-4be89b79bd36"
+    ge_cloud_access_token = "d7asdh2efads9afah2e0fadf8eh20da8"
```
anthonyburdi, line 61: Is there a reason to not use the existing global fixtures for these values? Just figure keep things DRY if we need to change them in the future.

great_expectations/data_context/migrator/cloud_migrator.py
anthonyburdi: I like it!

great_expectations/data_context/migrator/cloud_migrator.py
anthonyburdi: I think that is fine and simpler, these are all locally contained private methods so we can always move it out to a param if we need to inject a different serializer in the future without a breaking change.

tests/data_context/migrator/test_cloud_migrator.py
```diff
@@ -53,6 +50,53 @@ def test__send_configuration_bundle_sends_valid_http_request(
     )
 
 
+@pytest.mark.unit
+@pytest.mark.cloud
+def test__send_validation_results_sends_valid_http_request():
```
anthonyburdi: I'm assuming tests to make sure logging messages are printed will be in a separate PR? (that is fine just want to make sure they don't get missed)

great_expectations/data_context/migrator/cloud_migrator.py
anthonyburdi: What if there was a `post_to_cloud_backend()` public method on the GeCloudStoreBackend that we used here? Could also be used internally in the GeCloudStoreBackend. We could create a GeCloudStoreBackend in the CloudMigrator init flow and use that internally here to send the bundle/validations.

great_expectations/data_context/migrator/cloud_migrator.py
cdkini: ooh good call, will clean up!

anthonyburdi (approved):
I like the direction here. I have a few comments that are not blocking for this PR so approving for now contingent on resolving @kilo59's comment

tests/data_context/migrator/test_cloud_migrator.py
cdkini: yup!

tests/data_context/migrator/test_cloud_migrator.py
cdkini: I know @billdirks has pushed back on this a bit in the past and I see merits with both approaches - as these values are simple placeholders, do we need to couple additional dependencies here?

Good either way but just explaining the approach here.

great_expectations/data_context/migrator/cloud_migrator.py
cdkini: this could be useful! I'd want to use the same session object for all requests though - I'd have to think about how to go about this

cdkini: think I'm going to table this for the time being but come back to it if that's alright

great_expectations/data_context/migrator/cloud_migrator.py
Kilo59: Can we just define a `.dict()` or `.to_dict()` method for the `ConfigurationBundle` class?

anthonyburdi: @Kilo59 I thought we were trying to move away from objects defining their own serialization, especially since we need to serialize differently to yaml, json or dict e.g. for datasources (for storing in `great_expectations.yml`, sending to the cloud store, and printing to the screen in `context.list_datasources()`)

tests/data_context/migrator/test_cloud_migrator.py
anthonyburdi: Either is fine. I am thinking more if the url changes (doubtful) or we break something in CI and accidentally start sending usage events from tests to prod. If there are consistent ids to ignore that would be easier. But I'm probably being too cautious.

great_expectations/data_context/migrator/cloud_migrator.py
```diff
@@ -143,81 +150,145 @@ def migrate_validation_result(
...
+        print("Sending validation results (step 4/4)")
+        for i, validation_result in enumerate(serialized_validation_results):
+            response = self._post_to_cloud_backend(
+                resource_name=resource_name,
+                resource_type=resource_type,
+                attributes_key=attributes_key,
+                attributes_value=validation_result,
+            )
+            print(response)
+            # TODO: Handle success/failure cases
```
Kilo59, line 249: https://requests.readthedocs.io/en/latest/user/advanced/?highlight=event#event-hooks

great_expectations/data_context/migrator/cloud_migrator.py
```diff
@@ -143,81 +150,145 @@ def migrate_validation_result(
...
         )
+        response = self._session.post(url, json=data)
+        return self._parse_cloud_response(response=response)
 
-        try:
-            response = self._session.post(url, json=data)
-            response.raise_for_status()
-            return response.json()
+    def _parse_cloud_response(self, response: requests.Response) -> MigrationResponse:
+        success = response.ok
```
Kilo59, line 273: You might want to consider any `2XX` response a success.
This will only consider a `200` code a success but count `201`, `202`, `204` etc as `success = False` which seems wrong.

cdkini: https://requests.readthedocs.io/en/latest/api/?highlight=OK#requests.Response.ok shows that it works for all codes less that 400

PR #6104, opened 2022-09-28
[FEATURE] Introduction of updated `gx.get_context()` method that returns correct DataContext-type

Shinnnyshinshin:
# What is proposed by this PR?

This PR expands functionality to of `gx.get_context()` method to take in parameters and return correct `DataContext`-type (either `DataContext`, `BaseDataContext` or `CloudDataContext`)
 
Usage:

```python
my_context = gx.get_context([parameters])
```

## How does `get_context()` behave?

1. If `gx.get_context()` is run in a folder on the filesystem where `great_expectations init` has been run, then it will return a `DataContext`

 2. If `gx.get_context()` is passed in a `context_root_dir` (which contains `great_expectations.yml`) then it will return a `DataContext`.

3. If `gx.get_context()` is passed in an in-memory `project_config` then it will return `BaseDataContext`.  `context_root_dir` can also be passed in, but the configurations from the in-memory config will override the configurations in the `great_expectations.yml` file.

4. If GX is being run in the cloud, and the information needed for `ge_cloud_config` (ie `ge_cloud_base_url`, `ge_cloud_access_token`, `ge_cloud_organization_id`) are passed in as parameters to `get_context()`, configured as environment variables, or in a `.conf` file, then `get_context()` will return a `CloudDataContext`.

### Definition of Done


great_expectations/util.py
```diff
@@ -1656,10 +1663,90 @@ def convert_ndarray_decimal_to_float_dtype(data: np.ndarray) -> np.ndarray:
...
-def get_context():
-    from great_expectations.data_context.data_context import DataContext
+def get_context(
+    project_config: Optional[Union["DataContextConfig", dict]] = None,
+    context_root_dir: Optional[str] = None,
+    runtime_environment: Optional[dict] = None,
+    ge_cloud_base_url: Optional[str] = None,
+    ge_cloud_access_token: Optional[str] = None,
+    ge_cloud_organization_id: Optional[str] = None,
+    ge_cloud_mode: Optional[bool] = False,
```
Kilo59:
```suggestion
    ge_cloud_mode: bool = False,
```
Shouldn't the type here just be `bool`?

tests/data_context/test_get_data_context.py
```diff
@@ -0,0 +1,230 @@
...
+    assert isinstance(gx.get_context(project_config=config), BaseDataContext)
+
+
+@pytest.mark.unit
+def test_empty_call_returns_data_context(tmp_path: pathlib.Path):
+    project_path = tmp_path / "empty_data_context"
+    project_path.mkdir()
+    project_path_str = str(project_path)
+    gx.data_context.DataContext.create(project_path_str)
+    os.chdir(project_path_str)
```
Kilo59: How does the working dir get changed back?

If we don't already have this, I would recommend a context manager utility method that automatically resets the working directory when it exits.

https://stackoverflow.com/a/431747/6304433
```python
import os
from contextlib import contextmanager

@contextmanager
def working_directory(directory):
    owd = os.getcwd()
    try:
        os.chdir(directory)
        yield directory
    finally:
        os.chdir(owd)
```

tests/data_context/test_get_data_context.py
```diff
@@ -0,0 +1,230 @@
...
+@mock.patch("requests.get")
+@mock.patch.dict(
+    os.environ,
+    {
+        "GE_CLOUD_BASE_URL": "http://hello.com",
+        "GE_CLOUD_ORGANIZATION_ID": "1",
+        "GE_CLOUD_ACCESS_TOKEN": "i_am_a_token",
+    },
+    clear=True,
+)
```
Kilo59: Alternative to using `mock`.

https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html#monkeypatching-environment-variables

tests/data_context/test_get_data_context.py
```diff
@@ -0,0 +1,230 @@
...
+    mock_request.return_value.status_code = 200
+    try:
+        assert isinstance(
+            gx.get_context(
+                ge_cloud_organization_id="1", ge_cloud_access_token="i_am_a_token"
+            ),
+            CloudDataContext,
+        )
+    except:  # Not concerned with constructor output (only evaluating interaction with requests during __init__)
+        pass
```
Kilo59: I'm confused about what's actually being tested here.
If you are using a blanket `try: except:` why even have the assert statement at all?

The comment next to `except` makes me think that maybe a `spy` would be a more appropriate test double for this test???
https://pytest-mock.readthedocs.io/en/latest/usage.html#spy

tests/data_context/test_get_data_context.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+@pytest.mark.cloud
+@mock.patch("requests.get")
+def test_all_cloud_config_in_param_returns_cloud_context(
+    mock_request,
+    request_headers,
+):
+    mock_request.return_value.status_code = 200
```
Kilo59: You could refactor this request mock as a fixture and re-use it for all these tests.

```python
@pytest.fixture
def mock_api_200():
   with mock.patch("requests.get") as patch:
         patch.return_value.status_code == 200
         yield patch
```

Kilo59 (approved):
Excited for this change.

Mostly non-blocking comments.
The only thing I might consider blocking is making sure we deal with the working directory change during testing.

tests/data_context/test_get_data_context.py
Shinnnyshinshin: Love it. Added to `test_utils.py` and updated current test

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -90,6 +90,43 @@ def __init__(
             runtime_environment=runtime_environment,
         )
 
+    @classmethod
+    def is_ge_cloud_config_available(
```
anthonyburdi, line 93: ❤️

tests/data_context/test_get_data_context.py
Shinnnyshinshin: I do think the `monkeypatch` approach is cleaner in this case. Pr has been updated

tests/data_context/test_get_data_context.py
Shinnnyshinshin: The PR has been updated to `mock.patch` the `retrieve_data_context_config_from_ge_cloud()` method instead, which negates the need for the `try:except`, which I believe was masking an error. Thank you for calling this out. It helped reveal a hole in the test.

tests/data_context/test_get_data_context.py
Shinnnyshinshin: similar point to above. The PR has been updated to `mock.patch` the `retrieve_data_context_config_from_ge_cloud()` method instead, which negates the need for the `try:except`, which I believe was masking an error.

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -84,12 +84,50 @@ def __init__(
         self._project_config = self._apply_global_config_overrides(
             config=project_config
         )
-
+        print(self._project_config)
+        print("print:")
```
Kilo59: Did you mean to take this out?

tests/data_context/test_get_data_context.py
```diff
@@ -0,0 +1,218 @@
+import os
```
Kilo59: Do you still need `os` here?

Kilo59 (approved):

great_expectations/data_context/data_context/cloud_data_context.py
Shinnnyshinshin: yes :)

PR #6106, opened 2022-09-28
[FEATURE] Add logging/warnings to GX Cloud migration process

cdkini:
- Add logging/warnings to migration workflow
- Add unit tests around desired stdout/stderr

### Definition of Done


great_expectations/core/expectation_suite.py
```diff
@@ -102,6 +102,10 @@ def __init__(
         ensure_json_serializable(meta)
         self.meta = meta
 
+    @property
+    def name(self) -> str:
+        return self.expectation_suite_name
+
```
cdkini, line 108: Added this so we can do `config["name"]` for all object types in the bundle. We should eventually do this rename anyways but lmk if you think we should have an explicit fork for suites.

great_expectations/data_context/migrator/cloud_migrator.py
```diff
@@ -188,18 +180,60 @@ def _emit_warnings(
...
+            "our documentation for more details.\n"
+        )
 
     def _print_configuration_bundle_summary(
         self, configuration_bundle: ConfigurationBundle
     ) -> None:
-        pass
+        # NOTE(cdkini): Datasources should be verbose, not just summary!
+        to_print = (
+            ("Datasource", configuration_bundle.datasources),
```
cdkini, line 229: This needs to be updated to give a verbose output - will update in a subsequent PR

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -89,7 +89,7 @@ def get_user_friendly_error_message(
     support_message = []
     response: requests.Response = http_exc.response
 
-    logger.warning(f"{http_exc.__class__.__name__}:{http_exc} - {response}")
+    logger.info(f"{http_exc.__class__.__name__}:{http_exc} - {response}")
```
cdkini: Feel bad about changing this but if set at `warning`, it results in double output during the migration process.

tests/data_context/migrator/conftest.py
```diff
@@ -1,5 +1,187 @@
...
+    AnonymizedUsageStatisticsConfig,
+    CheckpointConfig,
+    DataContextConfig,
+    DatasourceConfig,
+)
+from great_expectations.datasource import BaseDatasource, LegacyDatasource
+from great_expectations.rule_based_profiler import RuleBasedProfiler
+
+
+class StubUsageStats:
```
cdkini, line 20: Moved all this to `conftest` so it can be used in multiple files

tests/data_context/migrator/conftest.py
```diff
@@ -1,5 +1,187 @@
...
+        anonymous_usage_stats_is_none: bool = False,
+        include_datasources: bool = True,
+    ) -> StubBaseDataContext:
+        return StubBaseDataContext(
+            anonymous_usage_stats_enabled=anonymous_usage_stats_enabled,
+            anonymous_usage_stats_is_none=anonymous_usage_stats_is_none,
+            include_datasources=include_datasources,
+        )
+
+    return _create_stub_base_data_context
```
cdkini: I'm finding this pattern to be quite effective when working with customizable stubs - I've modified the rest of the fixtures below to leverage this

tests/data_context/migrator/test_cloud_migrator.py
```diff
@@ -8,95 +8,43 @@
...
 )
+from tests.data_context.migrator.conftest import StubBaseDataContext
 
 
-@pytest.mark.unit
-@pytest.mark.cloud
-def test__send_configuration_bundle_sends_valid_http_request(
-    serialized_configuration_bundle: dict,
+@pytest.fixture
+def migrator_factory(
```
cdkini, line 19: Same note here - with both these factories, we have the ability to test a variety of configs

tests/data_context/migrator/test_cloud_migrator.py
```diff
@@ -161,6 +109,91 @@ def _build_mock_migrate(
...
+    def _assert(stdout: str, statements: List[str]) -> None:
+        last_position = -1
+        for statement in statements:
+            position = stdout.find(statement)
+            assert position != -1, f"Could not find '{statement}' in stdout"
+            assert (
+                position > last_position
+            ), f"Statement '{statement}' occurred in the wrong order"
+
+    return _assert
```
cdkini: When do we want global funcs and when do we want fixtures? I tend to lean towards the latter here because I want the test to show an explicit dependency on this helper.

great_expectations/data_context/migrator/cloud_migrator.py
```diff
@@ -1,17 +1,14 @@
 """TODO: Add docstring"""
```
anthonyburdi: Mind adding the docstring? Now that the usage pattern is slightly different would be good to document how it is intended to be used (using the return value of migrate to access any failed to migrate validations)

great_expectations/data_context/migrator/cloud_migrator.py
```diff
@@ -169,11 +157,15 @@ def _migrate_to_cloud(self, test_migrate: bool) -> None:
...
 
-        if not test_migrate:
-            self._send_configuration_bundle(serialized_bundle=serialized_bundle)
-            self._send_validation_results(
-                serialized_validation_results=serialized_validation_results,
-            )
+        if not self._send_configuration_bundle(
+            serialized_bundle=serialized_bundle, test_migrate=test_migrate
+        ):
+            return  # Exit early as validation results cannot be sent if the main payload fails
```
anthonyburdi, line 181: :thumbsup:

great_expectations/data_context/migrator/cloud_migrator.py
```diff
@@ -188,18 +180,60 @@ def _emit_warnings(
...
     def _warn_about_bundle_contains_datasources(self) -> None:
-        pass
+        logger.warning(
+            "Since your existing context includes one or more datasources, "
+            "please note that if your credentials are included in the datasource config, "
+            "they will be sent to the GX Cloud backend. We recommend storing your credentials "
+            "locally in config_variables.yml or in environment variables referenced "
+            "from your configuration rather than directly in your configuration. Please see "
+            "our documentation for more details.\n"
+        )
```
anthonyburdi, line 222: @donaldheppner - just want to double check with you on our wording on these messages.

great_expectations/data_context/migrator/cloud_migrator.py
anthonyburdi: Wait - wouldn't we want to raise an exception in `_send_configuration_bundle` after printing the warning? So no need to return a bool and return early.

anthonyburdi: Nevermind - I forgot that in our spec we are printing the error information from the server and not raising an exception.

great_expectations/data_context/store/ge_cloud_store_backend.py
anthonyburdi: if we want it to be warning only for the migration process, what if we build the string here and then inject the logger function into the method (with default of logger.warning?). Like

```
def get_user_friendly_error_message(
    http_exc: requests.exceptions.HTTPError, logger_fn: Callable = logger.warning
) -> str:
    # TODO: define a GeCloud service/client for this & other related behavior
    support_message = []
    response: requests.Response = http_exc.response

    logger_msg = f"{http_exc.__class__.__name__}:{http_exc} - {response}"
    logger_fn(logger_msg)
```
and in the migrator pass in `logger.info` as `logger_fn`?

tests/data_context/migrator/conftest.py
anthonyburdi: Ah crap - I like the pattern but this code all changed in https://github.com/great-expectations/great_expectations/pull/6107 to use dependency injection to create an empty config. Could you merge those changes in?

anthonyburdi (commented):
Left some comments, still need to look at test files. Submitting in case I get pulled away but will keep reviewing.

tests/data_context/migrator/test_cloud_migrator.py
anthonyburdi: Good question. I don't know if there is a hard and fast rule, but I usually use fixtures that generate a function when I am basically using that function to generate what would normally be passed in as a fixture based on params passed into the generated fn during the test. In other words, if the fixture fn returns something I would have passed in as a fixture, I create the fn as a fixture.

I use a global function when I am using it to encapsulate some repeated assertions like you are doing here.

I think the difference in my mind is that a fixture is more a part of the setup of the test and the global fns are part of the verify of the "setup" "evaluate" "verify"

tests/data_context/migrator/test_cloud_migrator.py
```diff
@@ -209,3 +243,183 @@ def test_no_event_sent_for_migrate_test_run_failure(
...
+    assert_stdout_is_accurate_and_properly_ordered: Callable,
+    capsys,
+):
+    migrator = migrator_with_stub_base_data_context
+
+    with mock.patch("requests.Session.post", autospec=True):
+        migrator._migrate_to_cloud(test_migrate=test_migrate)
+
+    stdout, _ = capsys.readouterr()
+    expected_statements = [
```
anthonyburdi, line 313: Love these tests!

tests/data_context/migrator/test_cloud_migrator.py
```diff
@@ -209,3 +243,183 @@ def test_no_event_sent_for_migrate_test_run_failure(
...
+            "Since your existing context includes one or more datasources, please note that if your credentials are included"
+            in aggregated_log_output
+        )
+
+    assert len(actual_logs) == expected_log_count
+
+
+@pytest.mark.unit
+@pytest.mark.cloud
+@pytest.mark.parametrize("test_migrate", [True, False])
```
anthonyburdi, line 301: :muscle:

tests/data_context/migrator/test_cloud_migrator.py
```diff
@@ -209,3 +243,183 @@ def test_no_event_sent_for_migrate_test_run_failure(
...
+        "Then we will send each of your validation results.",
+        "[Step 1/4]: Bundling context configuration",
+        "Bundled 1 Datasource(s):",
+        "Bundled 1 Checkpoint(s):",
+        "Bundled 1 Expectation Suite(s):",
+        "Bundled 1 Profiler(s):",
+        "[Step 2/4]: Preparing validation results",
+        "[Step 3/4]: Sending context configuration",
+        "[Step 4/4]: Sending validation results",
+        "Error sending validation result",
```
anthonyburdi: Could we also test that the validation result key is printed here?E.g. Error sending validation result <validation result key> (73/235)

great_expectations/data_context/store/ge_cloud_store_backend.py
cdkini: instead of passing an fn, I decided to pass in a log level - my concern with using the fn is that we're explicitly referencing the logger defined in that file. The `logger` object will be different depending on where I call the method.

tests/data_context/migrator/test_cloud_migrator.py
cdkini: going to change to make this a global func

great_expectations/data_context/store/ge_cloud_store_backend.py
anthonyburdi: Good call!

great_expectations/data_context/store/ge_cloud_store_backend.py
```diff
@@ -83,13 +83,13 @@ def construct_json_payload(
 
 
 def get_user_friendly_error_message(
-    http_exc: requests.exceptions.HTTPError,
+    http_exc: requests.exceptions.HTTPError, log_level: int = logging.WARNING
```
anthonyburdi, line 86: YES!!

tests/data_context/migrator/conftest.py
cdkini: Prefer your approach - updated across the uses

great_expectations/data_context/migrator/cloud_migrator.py
cdkini: done!

great_expectations/data_context/migrator/cloud_migrator.py
anthonyburdi: (in _warn_about_usage_stats_disabled and _warn_about_bundle_contains_datasources)

tests/data_context/migrator/test_cloud_migrator.py
cdkini: done!

anthonyburdi (approved):
LGTM! Just please make sure to get approval on the messaging for usage stats and datasources.

PR #6112, opened 2022-09-29
[MAINTENANCE] Add Enums for renderer types

NathanFarmer:
- GREAT-1284
- Leverage `Enum`s in place of `renderer_type` strings
- Move `RenderedAtomicContent` out of `render/types` package in anticipation of removing `types` package
- Only update production Expectations for this PR - `contrib` and `docs` to be updated subsequently

### Definition of Done


great_expectations/render/__init__.py
```diff
@@ -183,7 +183,7 @@ def __repr__(self) -> str:
     def __str__(self) -> str:
         return json.dumps(self.to_json_dict(), indent=2)
 
-    def to_json_dict(self) -> dict:
+    def to_json_dict(self) -> Optional[dict]:
```
Kilo59, line 186: I see that it could be `None` because `graph` is an optional keyword arg.
But should `graph` actually be optional? Can we make this a required parameter?

great_expectations/expectations/expectation.py
```diff
@@ -1552,14 +1559,14 @@ def _get_renderer_diagnostics(
...
         registered_renderers: List[str],
-        standard_renderers: List[str] = [
-            "renderer.answer",
-            "renderer.diagnostic.unexpected_statement",
-            "renderer.diagnostic.observed_value",
-            "renderer.diagnostic.status_icon",
-            "renderer.diagnostic.unexpected_table",
-            "renderer.prescriptive",
-            "renderer.question",
+        standard_renderers: List[Enum] = [
```
Kilo59, line 1562: Could we define a `Literal` type for these.
https://mypy.readthedocs.io/en/stable/literal_types.html#literal-types-and-enums

```suggestion
        standard_renderers: List[Union[LegacyRendererType, LegacyDiagnosticRendererType]] = [
```

Can we use a `Union` of the renderers?
If this is something that might be re-used it's probably better to define it in the same place we define the enums themselves.


```python

class LegacyRendererType(Enum):
   ...

class LegacyDiagnosticRendererType(Enum):
   ...

AnyRendererType = Union[LegacyRendererType, LegacyDiagnosticRendererType]

```
Or if only specific renderers are supported we could use a `Literal`, although I suspect this won't be needed.
https://mypy.readthedocs.io/en/stable/literal_types.html#literal-types-and-enums

great_expectations/render/__init__.py
```diff
@@ -1 +1,275 @@
...
+
+class RenderedAtomicValue(DictDot):
+    def __init__(
+        self,
+        schema: Optional[dict] = None,
+        header: Optional["RenderedAtomicValue"] = None,
+        template: Optional[str] = None,
+        params: Optional[dict] = None,
+        header_row: Optional[List["RenderedAtomicValue"]] = None,
+        table: Optional[List[List["RenderedAtomicValue"]]] = None,
```
Kilo59:
```suggestion
        table: Optional[List[List[RenderedAtomicValue]]] = None,
```
Do you still need to use the string version of these if we are using `from __future__ import annotations`?

Kilo59 (approved):
LGTM

A couple of minor suggestions/questions.

The only other thing I would ask is could you check some of the low type error `renderer` modules and see if any of them can be removed from our type-checking exclude list?
https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L80-L99

great_expectations/render/__init__.py
NathanFarmer: ✅

great_expectations/expectations/expectation.py

great_expectations/render/__init__.py
NathanFarmer: No we can not make it required based on the current design. `RenderedAtomicValue` accepts various combinations of optional components to produce all of the "new style" renderings. This class encapsulates many different classes that were defined by the legacy renderers. The simplifies the contributor experience.

PR #6122, opened 2022-09-30
[MAINTENANCE] add the internal `GE_DATA_CONTEXT_ID` env var to the docker file

Kilo59:
- add a known uuid as the `GE_DATA_CONTEXT_ID` so that the analytics team can filter out usage stats from CI builds and dev environments

### DIFF
--- a/docker/Dockerfile.tests
+++ b/docker/Dockerfile.tests
@@ -20,6 +20,8 @@ RUN git clone --filter=tree:0 --branch ${BRANCH} https://github.com/great-expect
 WORKDIR /great_expectations
 
 FROM build_${SOURCE} AS dev
+# used to filter usage stats coming from CI and dev environments
+ENV GE_DATA_CONTEXT_ID="00000000-0000-0000-0000-00000000e009"
 RUN pip install --no-cache-dir --requirement requirements-dev.txt --constraint constraints-dev.txt
 RUN pip install .
 CMD [ "python", "--version"]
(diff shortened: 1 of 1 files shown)


tannerbeam (approved):

docker/Dockerfile.tests
```diff
@@ -29,3 +29,6 @@ RUN pip install --no-cache-dir --requirement requirements-dev-test.txt --constra
 
 FROM dev as dev-tools
 RUN pip install --no-cache-dir --requirement requirements-dev-tools.txt --constraint constraints-dev.txt
+
+# used to filter usage stats coming from CI and dev environments
```
billdirks: Let's move this to line 23, right after `FROM build_${SOURCE} AS dev`. This means the `dev` target and any derived targets (`test` and `dev-tools`) will have this set. The `test` target is what the azure pipelines use.

I've verified this works by building `test` after this change and seeing the env variable being set:
```
# echo ${GE_DATA_CONTEXT_ID}
00000000-0000-0000-0000-00000000e009
```

billdirks (changes requested):
Thanks for making this change! We definitely don't want to be analyzing the sessions for the azure pipelines.

I'd move this env variable higher in the dockerfile.

docker/Dockerfile.tests
Kilo59: Done

billdirks (approved):
LGTM!

PR #6126, opened 2022-09-30
[MAINTENANCE] self_check linting & initial type-checking

Kilo59:
- begin linting `self_check`
- begin type checking `self_check`
- fix unrelated utils typing regressions (drop if possible)

### Definition of Done


great_expectations/self_check/util.py
```diff
@@ -2560,8 +2561,8 @@ def should_we_generate_this_test(
                 f"All sqlalchemy (including {backend}) is suppressed for test: {expectation_test_case.title} | {extra_debug_info}"
             )
             return False
-    if expectation_test_case.only_for != None and expectation_test_case.only_for:
-        if not backend in expectation_test_case.only_for:
+    if expectation_test_case.only_for is not None and expectation_test_case.only_for:
+        if backend not in expectation_test_case.only_for:
```
Kilo59, line 2565: Runtime changes (although shouldn't change behavior).

great_expectations/self_check/util.py
```diff
@@ -99,14 +100,14 @@
     from pyspark.sql import SparkSession
     from pyspark.sql.types import StructType
 except ImportError:
-    SparkDataFrame = type(None)  # type: ignore
-    SparkSession = None  # type: ignore
-    StructType = None  # type: ignore
+    SparkDataFrame = type(None)
```
alexsherstinsky, line 103: @Kilo59 How come some values are `None` and others `type(None)` -- what should our decision process be?  Thanks!

great_expectations/self_check/util.py
```diff
@@ -1318,7 +1319,7 @@ def get_test_validator_with_data(
                     type_ = schema[col]
                     if type_ in ["IntegerType", "LongType"]:
                         # Ints cannot be None...but None can be valid in Spark (as Null)
-                        vals = []  # type: ignore
+                        vals: list = []
```
alexsherstinsky: @Kilo59 Why not `List[<some type>]`?  Or we do not know the type?  In the latter case `List[Any]` not good?

great_expectations/self_check/util.py
```diff
@@ -1388,7 +1389,7 @@ def get_test_validator_with_data(
 def build_pandas_validator_with_data(
     df: pd.DataFrame,
     batch_definition: Optional[BatchDefinition] = None,
-    context: Optional["DataContext"] = None,  # noqa: F821
+    context: Optional["DataContext"] = None,
```
alexsherstinsky, line 1392: @Kilo59 Having `  # noqa: F821` suppresses the PyCharm warning -- are we sure we want to remove it?

great_expectations/self_check/util.py
```diff
@@ -2774,7 +2775,7 @@ def evaluate_json_test_cfe(validator, expectation_type, test, raise_exception=Tr
     return (result, error_message, stack_trace)
 
 
-def check_json_test_result(test, result, data_asset=None) -> None:
+def check_json_test_result(test, result, data_asset=None) -> None:  # noqa: C901 - 49
```
alexsherstinsky, line 2778: @Kilo59 What does `# noqa: C901 - 49` accomplish here?  Thanks.

great_expectations/util.py
```diff
@@ -1510,12 +1510,15 @@ def isclose(
         operand_a = operand_a.total_seconds()  # type: ignore[assignment]
         operand_b = operand_b.total_seconds()  # type: ignore[assignment]
 
-    return np.isclose(
-        a=np.float64(operand_a),  # type: ignore[arg-type]
-        b=np.float64(operand_b),  # type: ignore[arg-type]
-        rtol=rtol,
-        atol=atol,
-        equal_nan=equal_nan,
+    return cast(
```
alexsherstinsky, line 1513: @Kilo59 What is the reason for the `cast` here?  According to documentation, `If both a and b are scalars, returns a single boolean value.` so we should be fine here, right?  Thanks.

great_expectations/self_check/util.py
Kilo59: Good question, I didn't add these so I don't know what the thinking was here.

alexsherstinsky (changes requested):
@Kilo59 I asked a few questions for my edifications -- changes *might* be needed (not sure).  Thanks.

great_expectations/self_check/util.py
Kilo59: I've fixed the error.
`DataContext` is now imported under a `if TYPE_CHECKING` block.

great_expectations/self_check/util.py
Kilo59: https://superconductive.atlassian.net/wiki/spaces/SUP/pages/435978345/Static+Analysis+Type-Checking#Cyclomatic-Complexity

It is suppressing mccabe complexity warnings.

great_expectations/util.py
Kilo59: This method seems to return an alias of `bool` not the `bool` built-in. This confuses the type-checker.
```
great_expectations/util.py:1513: error: Incompatible return value type (got "bool_", expected "bool")  [return-value]
```

great_expectations/self_check/util.py
Kilo59: `list` is the same as `List[Any]`

However I went ahead and updated it to `List[Union[str, int, float, None]]`

great_expectations/self_check/util.py
alexsherstinsky: @Kilo59 The reason I asked is that they show up with the plus and in green here in this file-based review. :)

great_expectations/self_check/util.py
alexsherstinsky: ❤️

great_expectations/self_check/util.py
alexsherstinsky: 🙇

great_expectations/util.py
alexsherstinsky: @Kilo59 Oh yes, you are correct -- the `numpy` types sometimes have the same spelling as the `Python` types, but are in fact different objects (unfortunately, I must add).  Thanks for reminding of that.

alexsherstinsky (approved):
LGTM

great_expectations/self_check/util.py
Kilo59: The only thing I did was remove the `type: ignore` comments

Issue #6146, opened 2022-10-11
Hive dialect name error


hsheth2:
**Describe the bug**

The Hive dialect name is `b'hive'`, which is of type bytes and not str. This causes issues in the dialect comparison here: https://github.com/great-expectations/great_expectations/blob/1af478a2eb5694ad90d723a057be1c8636b2497b/great_expectations/execution_engine/sqlalchemy_dialect.py#L27-L30

This issue applies to GE versions 0.15.23+. 

This appears to be a regression introduced by https://github.com/great-expectations/great_expectations/pull/5980. Note that `b'foo' == 'foo'` is False but `hash(b'foo') == hash('foo')` is True, which is probably why this code worked before.

**To Reproduce**

We're running into this issue from the DataHub project, which depends on GE as a library. Hopefully the stack trace below has enough context, but I can try to provide more if needed.

Stack trace that I see:

```
Traceback (most recent call last):
  File "/home/mayuri/git/acryldata/datahub-fork/metadata-ingestion/src/datahub/ingestion/source/ge_data_profiler.py", line 906, in _generate_single_profile
    batch = self._get_ge_dataset(
  File "/home/mayuri/git/acryldata/datahub-fork/metadata-ingestion/src/datahub/ingestion/source/ge_data_profiler.py", line 970, in _get_ge_dataset
    batch = ge_context.data_context.get_batch(
  File "/home/mayuri/git/acryldata/datahub-fork/metadata-ingestion/venv/lib/python3.8/site-packages/great_expectations/data_context/data_context/base_data_context.py", line 1153, in get_batch
    return self._get_batch_v2(
  File "/home/mayuri/git/acryldata/datahub-fork/metadata-ingestion/venv/lib/python3.8/site-packages/great_expectations/data_context/data_context/base_data_context.py", line 859, in _get_batch_v2
    return validator.get_dataset()
  File "/home/mayuri/git/acryldata/datahub-fork/metadata-ingestion/venv/lib/python3.8/site-packages/great_expectations/validator/validator.py", line 2427, in get_dataset
    return self.expectation_engine(
  File "/home/mayuri/git/acryldata/datahub-fork/metadata-ingestion/venv/lib/python3.8/site-packages/great_expectations/dataset/sqlalchemy_dataset.py", line 568, in __init__
    if self.engine.dialect.name.lower() == GESqlDialect.BIGQUERY:
[...]

PR #6149, opened 2022-10-12
[BUGFIX] Regression due to `GESqlDialect` `Enum` for Hive

NathanFarmer:
- GREAT-1289
- Remove hard-coded bytes used for Hive dialect
- Add regression test and special case for comparisons against byte string

### Definition of Done

### DIFF
--- a/great_expectations/execution_engine/sqlalchemy_dialect.py
+++ b/great_expectations/execution_engine/sqlalchemy_dialect.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from enum import Enum
-from typing import Any, List
+from typing import Any, List, Union
 
 
 class GESqlDialect(str, Enum):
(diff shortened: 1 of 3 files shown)


great_expectations/execution_engine/sqlalchemy_dialect.py
```diff
@@ -27,6 +27,9 @@ class GESqlDialect(enum.Enum):
     def __eq__(self, other: Union[str, GESqlDialect]):
```
NathanFarmer: Removing dunder methods and adding `str` mixin doesn't work because we want to allow for equivalence with byte strings as well (e.g. b"hive").

great_expectations/execution_engine/sqlalchemy_dialect.py
```diff
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
-import enum
+from enum import Enum
 from typing import Any, List, Union
 
 
-class GESqlDialect(enum.Enum):
+class GESqlDialect(Enum):
```
NathanFarmer, line 7: Using `str` mixin doesn't work here because we want to also allow for equivalence of `bytes` (e.g. `b"hive"`) and multiple mixins are not allowed in this case.

NathanFarmer:
Resolves #6146

Kilo59 (approved):

PR #6153, opened 2022-10-13
[FEATURE] Initial zep datasource protocol.

billdirks:
A goal of the zep initiative is to make datasources easier to configure and extend. This adds an initial zep datasource protocol. I've manually tested that I can create a datasource which implements this protocol and can produce expectations (see https://superconductive.atlassian.net/browse/GREAT-1286).

This protocol will evolve before there are real use cases. In particular, we will need to better define the role of `BatchRequest` and perhaps `Batch` and `ExecutionEngine`.


Kilo59 (approved):

great_expectations/zep/interfaces.py
```diff
@@ -0,0 +1,20 @@
+from typing import List, Protocol, Union
```
Kilo59: Missed this initially, but until we get off of `python 3.7` we'll need to do something like this...

https://stackoverflow.com/a/71995876/6304433

Although I think we can just use the `typing_extension` version, I'm not sure we really need the `else Protocol = object` logic (from the linked example). 🤔

Kilo59 (changes requested):
Just need to account for `Python 3.7` differences.

great_expectations/zep/interfaces.py
billdirks: Thanks for catching this!
Yeah, we don't need the `else` logic. We actually use `typing_extensions` here:
https://github.com/great-expectations/great_expectations/blob/213b5609633a38857ef2cf9f89d2fe7a4112c7dd/tests/integration/profiling/rule_based_profiler/test_profiler_user_workflows.py#L16

billdirks: Done.

Kilo59 (approved):

PR #6155, opened 2022-10-13
[MAINTENANCE] Minor cleanup in preparation for Validator refactoring into separate concerns

alexsherstinsky:
### Scope
This work is a minor refactoring and cleanup as a prerequisite step to the next set of items, aimed at unbundling the concerns, presently contained in the monolithic `Validator` class, into separate, purpose-fit, classes.

- JIRA: GREAT-463/GREAT-1297
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/core/batch.py
```diff
@@ -13,10 +13,21 @@
...
+except ImportError:
+    pd = None
+
+    logger.debug(
+        "Unable to load pandas; install optional pandas dependency for support."
+    )
+
 try:
     import pyspark
+    from pyspark.sql import DataFrame as SparkDataFrame
```
NathanFarmer, line 27: Is there a reason why [import_manager.py](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/expectations/metrics/import_manager.py) is nested under the `metrics` directory instead of higher level where we can leverage it in situations like this?

great_expectations/self_check/util.py
```diff
@@ -1389,7 +1389,7 @@ def get_test_validator_with_data(  # noqa: C901 - 31
 def build_pandas_validator_with_data(
     df: pd.DataFrame,
     batch_definition: Optional[BatchDefinition] = None,
-    context: Optional["DataContext"] = None,
+    context: Optional["DataContext"] = None,  # noqa: F821
```
Kilo59, line 1392: You don't need to suppress these errors.
`DataContext` is being imported under an `if TYPE_CHECKING` conditional.

alexsherstinsky: @Kilo59 This was in reaction to `PyCharm` warnings.  Should I simply instruct `PyCharm` to suppress them globally?

great_expectations/core/batch.py
```diff
@@ -494,6 +505,20 @@ def ge_load_time(self):
...
+    def __init__(self, execution_engine) -> None:
+        self._execution_engine = execution_engine
+
+    @property
+    def execution_engine(self):
+        return self._execution_engine
+
+    # noinspection PyMethodMayBeStatic
+    def head(self, *args, **kwargs):
+        # CONFLICT ON PURPOSE. REMOVE.
```
NathanFarmer, line 518: Is this comment relevant to the forthcoming PRs or is this strictly something from the past? What purpose does this method serve if it returns and empty DF?

great_expectations/validator/validator.py
```diff
@@ -1422,7 +1422,7 @@ def batches(self) -> Dict[str, Batch]:
 
     @property
     def loaded_batch_ids(self) -> List[str]:
-        return self.execution_engine.loaded_batch_data_ids
+        return self._execution_engine.loaded_batch_data_ids
```
Kilo59, line 1425: Why the change from the public -> private version of this attribute?

NathanFarmer (approved):
A few non-blocking comments

great_expectations/self_check/util.py
Kilo59: I'm not sure, maybe.
But it looks like you are definitely getting a false positive locally, the linting CI step would have failed if this were a true postive.

great_expectations/validator/validator.py
alexsherstinsky: @Kilo59 This is important -- we strive to rely on private fields inside class as much as possible, thereby reducing the number of public methods only to the minimum that is necessary.  This cleanup step is toward this goal.  Thanks.

great_expectations/core/batch.py
alexsherstinsky: @NathanFarmer Great question, because I tried to reuse the import manager, but ended up with circular imports.  I did not venture to move the import manager to a different place for this work item; having said that, I feel that it would be a great cleanup type of Quality-Of-Life task to do after the present set of pull requests have been merged.  I will add it to our JIRA board.  Once done, some code will be naturally deleted (which is a good thing!).  Thank you.

great_expectations/core/batch.py
alexsherstinsky: @NathanFarmer I preserved this comment from the original version (I did not want to delete it without full context).  Thanks.

great_expectations/validator/validator.py
Kilo59: I agree with using private methods/fields to limit the surface area of the public API.
But in this case, should we remove the public `execution_engine` property as well?
Maybe that's coming later.

Kilo59: I'm on the same page with using private methods to limit the surface area of the public API.
Do we need the public version of this attribute at all?

great_expectations/self_check/util.py
Kilo59: The problem with including these error suppressions is now our CI will not catch the error if it pops up again.
Can you remove them in a follow-up PR?

PR #6156, opened 2022-10-13
[FEATURE] Introduce BatchManager to manage Batch objects used by Validator and BatchData used by ExecutionEngine

alexsherstinsky:
### Scope
The purpose fit `BatchManager` class is introduced and the various aspects of the functionality is is now responsible for is removed from `Validator` and `ExecutionEngine` classes, respectively:
* The new `BatchManager` component focuses on managing `Batch` objects for the various `Validator` purposes.
* The new `BatchManager` component also consistently manages the loaded `BatchData` objects for the various `ExecutionEngine` purposes.
* All existing public APIs have been preserved.
* All existing tests pass under the factored out `BatchManager` module with the corresponding redundant code removed from `Validator` and `Execution` classes.

- JIRA: GREAT-463/GREAT-1297
-
-

Previous Design Review notes:
-
-
-

### Definition of Done


great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,166 @@
...
+from great_expectations.core.id_dict import BatchSpec
+
+logger = logging.getLogger(__name__)
+logging.captureWarnings(True)
+
+
+class BatchManager:
+    def __init__(
+        self,
+        execution_engine: "ExecutionEngine",  # noqa: F821
```
NathanFarmer: I think you can get around this quotation method by doing: 

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from great_expectations.execution_engine import ExecutionEngine
```

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,166 @@
...
+        batch_list: Optional[List[Batch]] = None,
```
NathanFarmer, line 26: Why default `batch_list` to `None` just to immediately reset it to `[]`?

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,166 @@
...
+
+        if batch_list is None:
+            batch_list = []
+
+        self._batch_cache = {}
+        self._batch_data_cache = {}
+
+        self.load_batch_list(batch_list=batch_list)
+        if len(batch_list) > 1:
+            logger.debug(
```
NathanFarmer: Should this be `info` instead of `debug`?

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,166 @@
...
+                f"{len(batch_list)} batches will be added to this Validator.  The batch_identifiers for the active "
+                f"batch are {self.active_batch.batch_definition['batch_identifiers'].items()}"
+            )
+
+    @property
+    def execution_engine(self) -> "ExecutionEngine":  # noqa: F821
```
NathanFarmer: Same comment about return type annotation here.

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,166 @@
...
+        return self._active_batch_id
+
+    @property
+    def active_batch(self) -> Optional[Batch]:
+        """Getter for active Batch"""
+        active_batch_id: Optional[str] = self.active_batch_id
+        batch: Optional[Batch] = (
+            None if active_batch_id is None else self.batch_cache.get(active_batch_id)
+        )
+        return batch
```
NathanFarmer, line 101: Can we chose one way to do this and align the getter methods for `active_batch` and `active_batch_data`? They look very similar, but this one uses comprehension and has more type annotations.

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,166 @@
...
+
+    @property
+    def active_batch_spec(self) -> Optional[BatchSpec]:
+        """Getter for active batch's batch_spec"""
+        if not self.active_batch:
+            return None
+        else:
+            return self.active_batch.batch_spec
```
NathanFarmer: Same comment here. Do we want this to have multiple returns or use comprehension like in `active_batch` getter?

tests/validator/test_validator.py
```diff
@@ -1016,36 +1016,6 @@ def test_adding_expectation_to_validator_not_send_usage_message(
...
-    mar_min_date = "2019-03-01"
-    assert (
-        multi_batch_taxi_validator.active_batch_id == "90bb41c1fbd7c71c05dbc8695320af71"
-    )
-    with pytest.deprecated_call():  # parse_strings_as_datetimes is deprecated in V3
-        assert multi_batch_taxi_validator.expect_column_values_to_be_between(
-            "pickup_datetime", min_value=mar_min_date, parse_strings_as_datetimes=True
-        ).success
-
-    multi_batch_taxi_validator.active_batch_id = "0327cfb13205ec8512e1c28e438ab43b"
```
NathanFarmer, line 1034: Does this no longer work? Is this a breaking change?

NathanFarmer (changes requested):
A few requested changes! Thanks for this PR!

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer Because doing so in constructor arguments generates a security alert from `PyCharm`.

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer Trying... :)

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer Fixed (in next commit) -- thanks!

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer This one is important the way it is -- it includes a bug fix from the previous (merged) pull request, whereby before we did not properly recognize `batch_id` value of `()` (which is completely legitimate).

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer Fixing this one (nice catch -- thanks!).

tests/validator/test_validator.py
alexsherstinsky: @NathanFarmer One of the useful things from this refactor is that obsolete methods are deleted (and hence making the tests for them not needed either).  Thanks.

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -249,7 +253,8 @@ def get_batch_data_and_markers(
                 s3_object = s3_engine.get_object(Bucket=s3_url.bucket, Key=s3_url.key)
             except (ParamValidationError, ClientError) as error:
                 raise ge_exceptions.ExecutionEngineError(
-                    f"""PandasExecutionEngine encountered the following error while trying to read data from S3 Bucket: {error}"""
+                    f"""PandasExecutionEngine encountered the following error while trying to read data from S3 \
```
donaldheppner: I'd like to consider a longer line length to avoid this wrapping

donaldheppner (changes requested):

great_expectations/execution_engine/pandas_execution_engine.py
alexsherstinsky: @donaldheppner But then we get `PyCharm` warnings.  Which way should we go?  Thanks.

great_expectations/core/batch_manager.py
billdirks: @NathanFarmer Unfortunately, default function parameters are evaluated at class definition time. That means, setting a default argument to a mutable object or container type like `[]` or `{}` can lead to some strange behavior:
```
>>> def f(a=[]):
...         a.append(1)
...         print(a)
... 
>>> f()
[1]
>>> f()
[1, 1]
>>> f()
[1, 1, 1]
>>> f(a=[1,2,3])
[1, 2, 3, 1]
>>> f()
[1, 1, 1, 1]
```

great_expectations/core/batch.py
```diff
@@ -566,9 +566,15 @@ def __init__(
         self._batch_kwargs = batch_kwargs or BatchKwargs()
 
     @property
-    def data(self):
+    def data(self) -> Union[BatchData, pd.DataFrame, SparkDataFrame]:
```
billdirks: If you want you can define an alias for this union. That is somewhere above do, `DataType = Union[BatchData, pd.DataFrame, SparkDataFrame]`, and then you can replace this with `DataType`.

great_expectations/core/batch_manager.py
NathanFarmer: @billdirks that makes sense, thanks!

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,169 @@
...
+
+        self.load_batch_list(batch_list=batch_list)
+        if len(batch_list) > 1:
+            logger.info(
+                f"{len(batch_list)} batches will be added to this Validator.  The batch_identifiers for the active "
+                f"batch are {self.active_batch.batch_definition['batch_identifiers'].items()}"
+            )
+
+    @property
+    def execution_engine(self) -> ExecutionEngine:
```
billdirks: Can we delete this property? It seems like the `self._execution_engine` is an implementation detail of the batch manager and other objects don't need to reach inside here to get it?

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,169 @@
...
+    @property
+    def loaded_batch_ids(self) -> List[str]:
+        """IDs of loaded BatchData objects."""
+        return list(self._batch_data_cache.keys())
+
+    @property
+    def active_batch_data_id(self) -> Optional[str]:
+        """The batch id for the default batch data.
+
+        When a specific Batch objec is unavailable, then the data associated with the active_batch_data_id will be used.
```
billdirks: s/objec/object/

Also I don't understand this comment. That batch object is unavailable?

great_expectations/core/batch_manager.py
Kilo59: Also if you don't like string annotations we can do `from __future__ import annotations`
https://mypy.readthedocs.io/en/stable/runtime_troubles.html#future-annotations-import-pep-563

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        if self._active_batch_data_id is not None:
+            return self._active_batch_data_id
+
+        if len(self._batch_data_cache) == 1:
+            return list(self._batch_data_cache.keys())[0]
```
billdirks, line 66: Can we set this batch to the active one and then return its id? It seems like this is getting promoted to the active batch but we are leaving the `_active_batch_data_id` as `None`.

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,169 @@
...
+        """The BatchData object from the currently-active Batch object."""
+        if self.active_batch_data_id is None:
+            return None
+
+        return self._batch_data_cache.get(self.active_batch_data_id)
+
+    @property
+    def batch_cache(self) -> Dict[str, Batch]:
+        """Getter for ordered dictionary (cache) of "Batch" objects in use (with batch_id as key)."""
+        if not isinstance(self._batch_cache, OrderedDict):
```
billdirks: In `__init__` we set `self._batch_cache` to be {}. Can we set to to an `OrderedDict` there to avoid this check here?

Also, we do want the return type of this property to be OrderedDict instead of dict?

As an aside, in python 3.7+, plain dictionaries are specified to preserve order. I'm not opposed to using an OrderedDict since it is a little esoteric but just wanted to mention it.

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,169 @@
...
+    def active_batch_definition(self) -> Optional[BatchDefinition]:
+        """Getter for the active batch's batch definition"""
+        if not self.active_batch:
+            return None
+        else:
+            return self.active_batch.batch_definition
+
+    def reset_batch_cache(self) -> None:
+        """Clears Batch cache"""
+        self._batch_cache = {}
```
billdirks: Can we make this an ordered dict since it is converted upon retrieval?

great_expectations/core/batch_manager.py
NathanFarmer: @alexsherstinsky this comment from @Kilo59 will allow you to remove the quotations

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,169 @@
...
+        self._active_batch_id = None
+
+    def load_batch_list(self, batch_list: List[Batch]) -> None:
+        if batch_list is None:
+            batch_list = []
+
+        batch: Batch
+        for batch in batch_list:
+            try:
+                assert isinstance(
```
billdirks, line 139: Is mypy being run on this file? I'm wondering if we need this or if can be enforced by the type checker.

billdirks (commented):
I've only looked at batch.py and batch_manager.py so far but wanted to send out my comments.

tests/validator/test_validator.py
NathanFarmer: What makes this obsolete?

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,169 @@
...
+    from great_expectations.execution_engine import ExecutionEngine
+
+logger = logging.getLogger(__name__)
+logging.captureWarnings(True)
+
+
+class BatchManager:
+    def __init__(
+        self,
+        execution_engine: "ExecutionEngine",
```
NathanFarmer: if you import `from __future__ import annotations` then you can remove the quotation.

great_expectations/execution_engine/execution_engine.py
```diff
@@ -468,7 +441,8 @@ def add_column_row_condition(
 
         Args:
             domain_kwargs: the domain kwargs to use as the base and to which to add the condition
-            column_name: if provided, use this name to add the condition; otherwise, will use "column" key from table_domain_kwargs
+            column_name: if provided, use this name to add the condition; otherwise, will use "column" key from
+            table_domain_kwargs filter_null: if true, add a filter for null values
```
billdirks: I think a newline got moved here?

great_expectations/execution_engine/execution_engine.py
```diff
@@ -619,7 +593,8 @@ def _split_table_metric_domain_kwargs(
                     map(lambda element: f'"{element}"', unexpected_keys)
                 )
                 logger.warning(
-                    f'Unexpected key(s) {unexpected_keys_str} found in domain_kwargs for domain type "{domain_type.value}".'
+                    f"""Unexpected key(s) {unexpected_keys_str} found in domain_kwargs for domain type \
```
billdirks: Is the `\` a stray?

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,169 @@
...
+        )
+        return batch
+
+    @property
+    def active_batch_spec(self) -> Optional[BatchSpec]:
+        """Getter for active batch's batch_spec"""
+        if not self.active_batch:
+            return None
+
+        return self.active_batch.batch_spec
```
NathanFarmer, line 109: @alexsherstinsky still, across these getters we are using inconsistent logic. Sometimes just `if` with two `return` statements, sometimes `if/else` with two `return` statements, sometimes like line 111 with `if/else` and one `return` statement. They should be able to all look the same I think.

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,169 @@
...
+
+    @property
+    def active_batch_markers(self) -> Optional[BatchMarkers]:
+        """Getter for active batch's batch markers"""
+        if not self.active_batch:
+            return None
+        else:
+            return self.active_batch.batch_markers
```
NathanFarmer: `if/else` with two `returns`

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,169 @@
...
+
+    @property
+    def active_batch_definition(self) -> Optional[BatchDefinition]:
+        """Getter for the active batch's batch definition"""
+        if not self.active_batch:
+            return None
+        else:
+            return self.active_batch.batch_definition
```
NathanFarmer: `if/else` with two `returns`

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -185,15 +185,16 @@ def configure_validator(self, validator) -> None:
...
-    def load_batch_data(self, batch_id: str, batch_data: Any) -> None:
+    def load_batch_data(
+        self, batch_id: str, batch_data: Union[PandasBatchData, pd.DataFrame]
+    ) -> None:
         if isinstance(batch_data, pd.DataFrame):
             batch_data = PandasBatchData(self, batch_data)
-        elif isinstance(batch_data, PandasBatchData):
-            pass
-        else:
+        elif not isinstance(batch_data, PandasBatchData):
```
billdirks, line 193: Not sure if we need this. Depends on the state of type checking in our codebase here.

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -214,9 +215,11 @@ def get_batch_data_and_markers(
             batch_data = batch_spec.batch_data
             if isinstance(batch_data, str):
                 raise ge_exceptions.ExecutionEngineError(
-                    f"""PandasExecutionEngine has been passed a string type batch_data, "{batch_data}", which is illegal.
-Please check your config."""
+                    f"""PandasExecutionEngine has been passed a string type batch_data, "{batch_data}", which is \
```
billdirks: Stray `\` since this is in `"""`? Below also?

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,169 @@
...
+
+        return None
+
+    @property
+    def active_batch_data(self) -> Optional[BatchData]:
+        """The BatchData object from the currently-active Batch object."""
+        if self.active_batch_data_id is None:
+            return None
+
+        return self._batch_data_cache.get(self.active_batch_data_id)
```
NathanFarmer, line 76: One `if` with two `returns`.

NathanFarmer (changes requested):
I still think we should try to make the style consistent for the getters.

great_expectations/core/batch_manager.py
Kilo59: Was about to ask the same question.
The minimum python for gx is 3.7. So if we only care about key order, I would suggest we stick to a plain `dict`... however there might be some differences laid out here that are important 🤷‍♂️ 
https://docs.python.org/3/library/collections.html#ordereddict-objects

>Some differences from [dict](https://docs.python.org/3/library/stdtypes.html#dict) still remain:

>1. The regular [dict](https://docs.python.org/3/library/stdtypes.html#dict) was designed to be very good at mapping operations. Tracking insertion order was secondary.

>2. The [OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict) was designed to be good at reordering operations. Space efficiency, iteration speed, and the performance of update operations were secondary.

>3. The [OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict) algorithm can handle frequent reordering operations better than [dict](https://docs.python.org/3/library/stdtypes.html#dict). As shown in the recipes below, this makes it suitable for implementing various kinds of LRU caches.

>4. The equality operation for [OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict) checks for matching order.

>5. A regular [dict](https://docs.python.org/3/library/stdtypes.html#dict) can emulate the order sensitive equality test with p == q and all(k1 == k2 for k1, k2 in zip(p, q)).

>6. The popitem() method of [OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict) has a different signature. It accepts an optional argument to specify which item is popped.

>7. A regular [dict](https://docs.python.org/3/library/stdtypes.html#dict) can emulate OrderedDict’s od.popitem(last=True) with d.popitem() which is guaranteed to pop the rightmost (last) item.

>8. A regular [dict](https://docs.python.org/3/library/stdtypes.html#dict) can emulate OrderedDict’s od.popitem(last=False) with (k := next(iter(d)), d.pop(k)) which will return and remove the leftmost (first) item if it exists.

>9. [OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict) has a move_to_end() method to efficiently reposition an element to an endpoint.

>10. A regular [dict](https://docs.python.org/3/library/stdtypes.html#dict) can emulate OrderedDict’s od.move_to_end(k, last=True) with d[k] = d.pop(k) which will move the key and its associated value to the rightmost (last) position.

>11. A regular [dict](https://docs.python.org/3/library/stdtypes.html#dict) does not have an efficient equivalent for OrderedDict’s od.move_to_end(k, last=False) which moves the key and its associated value to the leftmost (first) position.

great_expectations/execution_engine/execution_engine.py
```diff
@@ -184,11 +190,12 @@ def __init__(
             if key in self.recognized_batch_spec_defaults
         }
 
-        self._batch_data_dict = {}
+        self._batch_manager = BatchManager(execution_engine=self)
```
billdirks, line 193: A naive question. Thoughts on pulling the `BatchManager` totally out of `ExecutionEngine`? I know a goal of this effort is to "unbundle" things so I'm wondering if letting a caller manage when it needs a batch manager is helpful/realistic.

great_expectations/core/batch_manager.py
Kilo59: This is actually a handy "feature" when doing recursive functions.

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -357,12 +364,12 @@ def dataframe(self):
...
         """
         # Changed to is None because was breaking prior
-        if self.active_batch_data is None:
+        if self.batch_manager.active_batch_data is None:
             raise ValueError(
                 "Batch has not been loaded - please run load_batch_data() to load a batch."
             )
 
-        return self.active_batch_data.dataframe
+        return cast(PandasBatchData, self.batch_manager.active_batch_data).dataframe
```
billdirks, line 370: Can we put a return type of the property here and in other places? I haven't checked but I'm wondering it could also us to remove casts elsewhere.

great_expectations/execution_engine/pandas_execution_engine.py
Kilo59: You can separate python strings in parens and they will automatically be joined.

image

### Edit
Technically the parens have nothing to do with it, string literals next to each other are automatically concatenated, but the parens allow you to span this across multiple lines.

Just be sure you are NOT including commas `,` else it becomes a tuple.

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks This is existing (old code) -- whereby `BatchData` is available, but the `Batch` object is not (or not yet) -- this supports quite a few tests, and I would not want to change it right now.  Thanks.

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks Will try (in the next commit).  Thanks.

great_expectations/core/batch.py
alexsherstinsky: @billdirks Will try (in the next commit).  Thanks.

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks Not quite -- `BatchData` and `Batch` objects are tracked separately (`batch_cache` is reset when new `Validator` is instantiated).  I would not be comfortable with changing this in the present pull request.  Thanks.

great_expectations/core/batch_manager.py
Kilo59: Looks like it is.
`core/batch.py` and some other modules are ignored but this one should be checked.

https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L51-L54

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks Will try in the next commit (previously, regular dictionary did not work).  Thanks.

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks Will try in the next commit -- thanks!

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks This is an old "defensive programming" check, which should probably stay for now.  Thanks.

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer Will try -- in the next commit.  Thanks.

billdirks (commented):
Thanks for making this change and peeling some logic out of the validator!

I haven't looked at the specifics but I notice that we need to `cast` the returned data objects and am wondering if there is a way to reduce this.

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer I will look through it and fix wherever there is an opportunity to do so.  Thanks.

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer Great catch -- will fix (in next commit) -- thanks!

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer Great catch -- will fix (in next commit) -- thanks!

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer This should be our preferred pattern for such cases (since return is "final").  Thanks.

great_expectations/core/batch.py
Kilo59: This is more of a nit but isn't the property here superfluous now that it has a `setter` that allows modification?

It doesn't really protect `self._data` or calculate anything.

I can see the case for keeping it just to stay consistent with other parts of our code though.

great_expectations/core/batch_manager.py
donaldheppner: If we only support `Batch` here, we should raise an error rather than log a warning

great_expectations/execution_engine/execution_engine.py
donaldheppner: I think this could be a good idea, especially since so many of the other properties reference attributes of the BatchManager via the ExecutionEngine

great_expectations/execution_engine/pandas_execution_engine.py
donaldheppner: I think the warning will go away if you adopt our standard line length of 120 characters

donaldheppner (dismissed):

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer This seems to have worked beautifully!  Thanks! :-)

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer I fixed them everywhere in this module -- thanks for spotting them.

great_expectations/execution_engine/execution_engine.py
alexsherstinsky: @billdirks Nice catch -- thank you (fixed).

great_expectations/execution_engine/execution_engine.py
alexsherstinsky: @billdirks I added it on purpose (to avoid `PyCharm` line length warnings).  Thanks.

great_expectations/execution_engine/pandas_execution_engine.py
alexsherstinsky: @billdirks This is needed -- enables tests that supply `BatchData` but not formal `Batch` object (similar to an earlier comment).  Thanks.

great_expectations/execution_engine/pandas_execution_engine.py
alexsherstinsky: @billdirks I added it on purpose (to avoid `PyCharm` line length warnings).  Thanks.

great_expectations/execution_engine/execution_engine.py
alexsherstinsky: @billdirks I am not sure how that would work -- `ExecutionEngine` needs services from `BatchManager` (`ExecutionEngine` used to have all of the relevant logic inside itself, but now it is encapsulated in `BatchManager`).

great_expectations/execution_engine/pandas_execution_engine.py
alexsherstinsky: @billdirks Will do (in next commit).  Thanks.

great_expectations/execution_engine/pandas_execution_engine.py
alexsherstinsky: @donaldheppner My `PyCharm` already says 120 -- I am not sure what is going on here (still trying).  Thanks.

great_expectations/execution_engine/pandas_execution_engine.py
billdirks: mypy should fail unless we have a `PandasBatchData` or a `pd.DataFrame` object so if mypy is passing I'm not sure we can ender this `raise` codeblock. Are tests that pass mypy hitting this exception?

great_expectations/execution_engine/pandas_execution_engine.py
billdirks: I'm surprised pycharm is throwing a line length warning if the line is < our line length limit. 
Most python code I've looked at doesn't use `\` but `"` each line since python will concatenate them together.

great_expectations/execution_engine/execution_engine.py
alexsherstinsky: Fixed.

great_expectations/execution_engine/pandas_execution_engine.py
alexsherstinsky: Fixed.

great_expectations/execution_engine/pandas_execution_engine.py
Kilo59: We don't currently run mypy against the tests

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,164 @@
...
+        batch_list: Optional[List[Batch]] = None,
+    ) -> None:
+        """
+        Args:
+            execution_engine: The ExecutionEngine to be used to access cache of loaded Batch objects.
+            batch_list: List of Batch objects available from external source (default is None).
+        """
+        self._execution_engine = execution_engine
+
+        self._active_batch_id = None
```
billdirks: It would be nice to type the member variables defined here, eg `self._active_batch_id`, `self._active_batch_data_id`, etc. Could you add these?

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,164 @@
...
+        self._active_batch_data_id = None
+
+        if batch_list is None:
+            batch_list = []
+
+        self._batch_cache = OrderedDict()
+        self._batch_data_cache = {}
+
+        self.load_batch_list(batch_list=batch_list)
+        if len(batch_list) > 1:
```
billdirks: nit: I would remove this `len` check since it protects logging and a user can set the logging level to manage this. But 🤷 .

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,164 @@
...
+    @property
+    def loaded_batch_ids(self) -> List[str]:
+        """IDs of loaded BatchData objects."""
+        return list(self._batch_data_cache.keys())
+
+    @property
+    def active_batch_data_id(self) -> Optional[str]:
+        """The batch id for the default batch data.
+
+        When a specific Batch is unavailable, then the data associated with the active_batch_data_id will be used.
```
billdirks, line 60: I still don't understand this comment. Can you add something to clarify it? There is no batch as a parameter so what is unavailable,`self._active_batch_data_id`? If so, that is private so what does that mean to a caller that reads this docstring.

great_expectations/core/batch_manager.py
billdirks: As someone who is not as familar with this code, could you explain this to me what is the relationship between `self._active_batch_data_id` and `list(self._batch_data_cache.keys())[0]`? I guess if I call `active_batch_data_id` I expect to get that but it looks like I'm getting something else? Is it some type of default?

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,164 @@
...
+    @property
+    def batch_cache(self) -> Dict[str, Batch]:
+        """Getter for ordered dictionary (cache) of "Batch" objects in use (with batch_id as key)."""
+        return self._batch_cache
+
+    @property
+    def active_batch_id(self) -> Optional[str]:
+        """Getter for active Batch ID"""
+        active_batch_data_id: str = self._active_batch_data_id
+        if active_batch_data_id != self._active_batch_id:
```
billdirks, line 87: Do we ever expect `self._active_batch_data_id` and `self._active_batch_id` to differ? This makes it seem like they shouldn't except maybe locally in some method (and if so we could use a local for one of these). Basically I'm curious:
1. Should these ever differ?
2. If they differ is this an error and should we die?
3. Do we need 2 variables?

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,164 @@
...
+
+        return self.active_batch.batch_definition
+
+    def reset_batch_cache(self) -> None:
+        """Clears Batch cache"""
+        self._batch_cache = OrderedDict()
+        self._active_batch_id = None
+
+    def load_batch_list(self, batch_list: List[Batch]) -> None:
+        if batch_list is None:
```
billdirks, line 133: nit: mypy should enforce that `batch_list` is never `None`.

great_expectations/execution_engine/execution_engine.py
billdirks: I only see `batch_manager` getting used in this file in the getter and in `load_batch_manager(...)` where is it just being used as a proxy so I thought I'd ask. It is a naive question and I don't know if breaking it out would be beneficial.

great_expectations/execution_engine/pandas_execution_engine.py
billdirks: I see. We probably shouldn't be passing a non `Union[PandasBatchData, pd.DataFrame]` in tests since "real" code would fail mypy so it seems like a strange thing to test.
We should file a cleanup test ticket.

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -352,17 +357,17 @@ def _apply_splitting_and_sampling_methods(self, batch_spec, batch_data):
         return batch_data
 
     @property
-    def dataframe(self):
+    def dataframe(self) -> pd.DataFrame:
```
billdirks, line 360: It looks like we cast the output to `PandasBatchData` on line 370. Can we set the return type to match that?

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -467,15 +472,19 @@ def get_domain_records(
         batch_id = domain_kwargs.get("batch_id")
         if batch_id is None:
             # We allow no batch id specified if there is only one batch
-            if self.active_batch_data_id is not None:
-                data = self.active_batch_data.dataframe
+            if self.batch_manager.active_batch_data_id is not None:
+                data = cast(
```
billdirks, line 476: Can we remove this cast if we fix update the return type on line 360 above?

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -204,24 +204,25 @@ def __init__(
         self._data_sampler = SparkDataSampler()
 
     @property
-    def dataframe(self):
+    def dataframe(self) -> DataFrame:
```
billdirks, line 207: Similar comment here to the one in the Panda's file. Could we update this to `SparkDFBatchData` and then remove the cast on line 450 and 459?

tests/execution_engine/test_sqlalchemy_execution_engine.py
```diff
@@ -291,7 +295,9 @@ def test_get_domain_records_with_column_domain(sa):
     expected_column_df = df.iloc[:3]
     engine = build_sa_engine(expected_column_df, sa)
     expected_data = engine.engine.execute(
-        sa.select(["*"]).select_from(engine.active_batch_data.selectable)
+        sa.select(["*"]).select_from(
+            cast(SqlAlchemyBatchData, engine.batch_manager.active_batch_data).selectable
```
billdirks, line 299: This is an observation with a musing and not a code change request:
We `cast` in this file and the above files and it makes me wonder if there is a way to reduce this since casting means that the static analyzer is missing information to do all the checks that it could. Maybe `Generics`? Maybe something simpler?

billdirks (changes requested):
Thanks for this PR and the fixes so far. I have made some more comments and thought I'd categorize them since some are questions that I'd like to understand the answers to and are not requests for changes:

**Change requests**
1. Request for types in `BatchManager.__init__`
2. Clarify the docstring for `active_batch_data_id`. I'm not sure what this is trying to say. It may be the old docstring but I wouldn't have understood that either.
3. Fix the return types and remove some casting in pandas_execution_engine.py and  sparkdf_execution_engine.py

**Questions**
1. My question in [this thread](https://github.com/great-expectations/great_expectations/pull/6156/files#r997323842). What is the relationship
2. My questions in the `active_batch_id` property code.

**Musings**: These are thoughts that don't need followup (though feel free, I'd like to hear your thoughts)
1. The thread about removing the branch manager for the execution engine.
2. Removing the casting in the sql execution engine or possible all the casting.

**Other**
1. Filing a ticket about removing tests that would violate the type signature of functions.

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks This is an old "safety fuse" -- if the `_active_batch_data_id` is unavailable (did not get set for whatever reason), then if there is only one loaded `BatchData` (one unambiguously), then it will play the role of an `active_batch_data_id` -- some non-NULL value is needed to compute a metric (by the particular `ExecutionEngine`), and we have been using it.  But if there is more than one, we cannot pinpoint the active `BatchData`, in which case we give up.  Thanks.

great_expectations/execution_engine/pandas_execution_engine.py
alexsherstinsky: @billdirks I thought that I fixed this per your feedback (by defining an alias type).  Thanks.

great_expectations/execution_engine/execution_engine.py
alexsherstinsky: @billdirks I also do not see how we would obtain a handle to `BatchManager` without instantiating it in the `ExecutionEngine` parent class, other than creating a singleton; but we already have `DataContext` as the singleton, and we do not want to create another one right now (moreover, `ExecutionEngine` does not use `DataContext`).  So would prefer to leave it like it is for now (it is congruent with the existing code).  Thanks.

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks I could, but I thought that this was against our style.  Do you know for certain?  Thanks.

alexsherstinsky: @billdirks Done (in next commit).  Thanks.

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks Per our brief discussion, this block has been removed (this information is not very interesting).  Thanks.

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks I believe that I had written up an answer in another place in this PR -- happy to discuss.  Thanks.

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks I believe that I had written up an answer in another place in this PR -- in addition, your question is important: yes, `active_batch_data_id` and `active_batch_id` can be different.  The former refers to the most recent loaded `BatchData`, while the latter refers to the most recent requested `Batch`.  There can be situations where no new `BatchData` objects have been loaded; however, a new `Validator` was instantiated and requested `Batch` objects, whose `BatchData` has already been loaded (and cached).  Since `BatchData` IDs are from the same name space as `Batch` IDs, this helps avoid unnecessary loading of data from the backends.  Thanks.

great_expectations/execution_engine/pandas_execution_engine.py
alexsherstinsky: @billdirks I believe we have `.dataframe` at the end -- so `pd.DataFrame` is the correct return value.  Thanks.

great_expectations/execution_engine/pandas_execution_engine.py
alexsherstinsky: @billdirks Per my comment above, I think that the existing code does the right thing, but happy to discuss.  Thanks.

great_expectations/execution_engine/sparkdf_execution_engine.py
alexsherstinsky: @billdirks Please let me know whether or not the prior clarifications make sense (we are actually getting the `dataframe` property out).  Thanks.

tests/execution_engine/test_sqlalchemy_execution_engine.py
alexsherstinsky: @billdirks I am not sure, but the idea is to get the `BatchData` of the appropriate type, and then the final "dot accessor" works without `PyCharm` warnings.

great_expectations/core/batch_manager.py
alexsherstinsky: @billdirks This is a good point (fixed).  Thanks!

great_expectations/execution_engine/pandas_execution_engine.py
alexsherstinsky: @billdirks I do not think that we can -- the cast is for the `BatchData` subclass, while the return type is for the actual `data` object it contains (the `BatchData` itself is an encapsulation, not the data itself).  Thanks.

great_expectations/core/batch_manager.py
billdirks: My concern is for a future developer who goes to modify this function. They aren't going to know what the Batch being unavailable means. I'd like to see the docstring updated with your explanation here: https://github.com/great-expectations/great_expectations/pull/6156/files#r997586853

great_expectations/core/batch_manager.py
billdirks: Why do we log a warning? This sounds like a normal situation. Should we log at all? If we want to log, can we do it on the info level?

great_expectations/execution_engine/execution_engine.py
billdirks: We don't need to do this now. We definitely don't want to introduce a singleton. My observation that the `ExecutionEngine` doesn't use the `BatchManager` (maybe a subclass does?) and the client do use the `BatchManager` directly. But this isn't blocking this change.

great_expectations/execution_engine/pandas_execution_engine.py
billdirks: I think there is a misunderstanding. Let's chat. This comment is about the branch `elif not isinstance(batch_data, PandasBatchData)`. To enter it, `batch_data` is not a `pd.DataFrame` (the if condition) or a  `PandasBatchData` (the elif). However mypy should enforce that is it one of these types.

great_expectations/execution_engine/pandas_execution_engine.py
billdirks: Oh, I see. I misread this.

great_expectations/execution_engine/pandas_execution_engine.py
billdirks: I see. I misread it earlier. Let's chat about how we can remove all this casting and whether we want to do it now or later.

tests/execution_engine/test_sqlalchemy_execution_engine.py
billdirks: Yes, let's chat.

great_expectations/core/batch_manager.py
```diff
@@ -0,0 +1,159 @@
...
+
+    @property
+    def batch_cache(self) -> Dict[str, Batch]:
+        """Getter for ordered dictionary (cache) of "Batch" objects in use (with batch_id as key)."""
+        return self._batch_cache
+
+    @property
+    def active_batch_id(self) -> Optional[str]:
+        """Getter for active Batch ID"""
+        active_batch_data_id: str = self._active_batch_data_id
```
billdirks, line 86: When I run `mypy great_expectations/core/batch_manager.py` I can the following error:
```
error: Incompatible types in assignment (expression has type "Optional[str]", variable has type "str")  [assignment]
```
It looks like this file isn't being covered `invoke types`. Can you fix this error (`active_batch_data_id` is an Optional[str]) and add this file so it is type checked.

great_expectations/core/batch_manager.py
billdirks: @alexsherstinsky will edit docstring in the next pr.

great_expectations/core/batch_manager.py
billdirks: @alexsherstinsky says this only happens in tests. We should file a ticket to clean this up in tests and then we can throw an error here.

great_expectations/execution_engine/pandas_execution_engine.py
billdirks: Talked to @alexsherstinsky. This cast to to clean up some pycharm warnings so it improves the current state. We can file a followup ticket to be done in a separate PR to investigating cleaning up the casts.

billdirks (approved):
@alexsherstinsky is going to make the type and docstring changes in a followup PR.

NathanFarmer (approved):
Looks good to me as well!

PR #6165, opened 2022-10-18
[FEATURE] Introduce MetricsCalculator and Refactor Redundant Code out of Validator

alexsherstinsky:
### Scope
* Place attributes and methods associated with metrics calculation into the dedicated `MetricsCalculator` class and use it inside `Validator` and other modules.
* All existing tests pass.
* Added unit tests for the new `MetricsCalculator` class.

- JIRA: GREAT-463/GREAT-1297
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/core/batch_manager.py
```diff
@@ -55,9 +55,15 @@ def loaded_batch_ids(self) -> List[str]:
 
     @property
     def active_batch_data_id(self) -> Optional[str]:
-        """The batch id for the default batch data.
+        """
+        The Batch ID for the default "BatchData" object.
 
         When a specific Batch is unavailable, then the data associated with the active_batch_data_id will be used.
+
+        This is a "safety valve" provision.  If self.__active_batch_data_id is unavailable (e.g., did not get set for
```
NathanFarmer: Double `_` should be single `_`.

great_expectations/validator/metrics_calculator.py
```diff
@@ -0,0 +1,400 @@
...
+                metric_name="table.head",
+                metric_domain_kwargs=domain_kwargs,
+                metric_value_kwargs={
+                    "n_rows": n_rows,
+                    "fetch_all": fetch_all,
+                },
+            )
+        )
+
+        df: pd.DataFrame
```
NathanFarmer: My understanding is that we have come to an agreement that this type hint is not necessary since the assignment of both `df` variables below explicitly states that a `pd.DataFrame` will be instantiated.

great_expectations/validator/metrics_calculator.py
```diff
@@ -0,0 +1,400 @@
...
+            metrics={metric.metric_name: metric},
+        )[metric.metric_name]
+
+    def get_metrics(
+        self,
+        metrics: Dict[str, MetricConfiguration],
+    ) -> Dict[str, Any]:
+        """
+        Args:
+            metrics: Dictionary of desired metrics to be resolved; metric_name is key and MetricConfiguration is value.
```
NathanFarmer, line 135: Why do we need `Dict[str, MetricConfiguration]` when the `MetricConfiguration` already contains the `metric_name` string? Can't we just use a `List[MetricConfiguration]`?

NathanFarmer: What purpose does `get_metrics` serve in comparison to `compute_metrics`?

great_expectations/validator/metrics_calculator.py
```diff
@@ -0,0 +1,400 @@
...
+        metric_configuration: MetricConfiguration,
+        runtime_configuration: Optional[dict] = None,
+    ) -> None:
+        """Obtain domain and value keys for metrics and proceeds to add these metrics to the validation graph
+        until all metrics have been added."""
+
+        metric_impl = get_metric_provider(
+            metric_configuration.metric_name, execution_engine=self._execution_engine
+        )[0]
+        metric_dependencies = metric_impl.get_evaluation_dependencies(
```
NathanFarmer, line 212: Not actionable for this PR, but in this method, `get_evaluation_dependencies`, and `resolve_validation_graph` is where I think we should look into implementing the logic for conditional evaluation dependencies. We could return multiple evaluation dependency graphs here, and then determine which is preferable during the resolution stage depending on other requested metrics. I will write this up in a document for review.

great_expectations/validator/validator.py
```diff
@@ -146,37 +139,202 @@ def __init__(
...
+        self._expose_dataframe_methods: bool = False
 
-        self.interactive_evaluation = interactive_evaluation
+        self.interactive_evaluation: bool = interactive_evaluation
         self._initialize_expectations(
             expectation_suite=expectation_suite,
             expectation_suite_name=expectation_suite_name,
         )
-        self._default_expectation_args = copy.deepcopy(
+        self._default_expectation_args: dict = copy.deepcopy(
```
NathanFarmer: This looks to be `Dict[str, Union[bool, str]]`

scripts/check_type_hint_coverage.py
```diff
@@ -4,7 +4,7 @@
 from typing import Dict, List, Optional
 
 TYPE_HINT_ERROR_THRESHOLD: int = (
-    1000  # This number is to be reduced as we annotate more functions!
+    1500  # This number is to be reduced as we annotate more functions!
```
NathanFarmer, line 7: Can we update this comment to explain when this number may need to be increased? I don't follow why based on the changes in `pyproject.toml`.

tests/core/test_metrics_calculator.py
```diff
@@ -0,0 +1,128 @@
...
+from great_expectations.util import isclose
+from great_expectations.validator.metric_configuration import MetricConfiguration
+from great_expectations.validator.metrics_calculator import MetricsCalculator
+from great_expectations.validator.validator import Validator
+
+
+@pytest.fixture
+def integer_and_datetime_sample_dataset() -> dict:
+    week_idx: int
+    data: dict = {
```
NathanFarmer: Both type hints are self-evident here.

tests/core/test_metrics_calculator.py
```diff
@@ -0,0 +1,128 @@
...
+    Expected partion boundaries are pre-computed algorithmically and asserted to be "close" to actual metric values.
+    """
+    validator_with_data: Validator = get_test_validator_with_data(
+        execution_engine=backend,
+        data=integer_and_datetime_sample_dataset,
+    )
+
+    metrics_calculator: MetricsCalculator = validator_with_data.metrics_calculator
+
+    seconds_in_week: int = 604800
```
NathanFarmer: Self-evident type hint

tests/core/test_metrics_calculator.py
```diff
@@ -0,0 +1,128 @@
...
+
+    n_bins: int = 10
```
NathanFarmer: Self-evident type hint

tests/core/test_metrics_calculator.py
```diff
@@ -0,0 +1,128 @@
...
+
+    increment: Union[float, datetime.timedelta]
+    idx: int
```
NathanFarmer, line 81: Self-evident type hints - `element` is not since it is what is returned.

NathanFarmer (changes requested):
A few requests/comments.

great_expectations/core/batch_manager.py
alexsherstinsky: @NathanFarmer nice catch (fixing in next commit).  Thanks!

great_expectations/validator/metrics_calculator.py
alexsherstinsky: @NathanFarmer not necessary strictly or optionally?

great_expectations/validator/metrics_calculator.py
alexsherstinsky: @NathanFarmer Old API that we must preserve.

great_expectations/validator/validator.py
alexsherstinsky: @NathanFarmer I am afraid it could go wider.  How about we leave it as this for now (against my immediate wish -- and yours, too).  Thanks.

scripts/check_type_hint_coverage.py
alexsherstinsky: @NathanFarmer This came from somewhere else (I did not modify this, as far as I can recall).  Thanks.

great_expectations/validator/metrics_calculator.py
alexsherstinsky: Fixed.

tests/core/test_metrics_calculator.py
alexsherstinsky: Fixed.

tests/core/test_metrics_calculator.py
alexsherstinsky: Fixed.

tests/core/test_metrics_calculator.py
alexsherstinsky: Leaving these as is for now.

tests/core/test_metrics_calculator.py
alexsherstinsky: Fixed.

great_expectations/validator/metrics_calculator.py
NathanFarmer: @alexsherstinsky `MetricsCalculator` is a new class. Can't we just call `compute_metrics` from the legacy `get_metrics` methods?

great_expectations/validator/validator.py
NathanFarmer: @alexsherstinsky we have `mypy` now to catch these things in case the class variable changes. Right now we know that the exact type is `Dict[str, Union[bool, str]]`.

great_expectations/validator/validator.py
```diff
@@ -146,37 +139,202 @@ def __init__(
...
+
+        (To be deprecated in favor of using methods in "MetricsCalculator" class.)
+
+        Args:
+            metrics: Dictionary of desired metrics to be resolved; metric_name is key and MetricConfiguration is value.
+
+        Returns:
+            Return Dictionary with requested metrics resolved, with metric_name as key and computed metric as value.
+        """
+        return self._metrics_calculator.get_metrics(metrics=metrics)
```
NathanFarmer, line 295: This could call `compute_metrics`

great_expectations/validator/metrics_calculator.py
```diff
@@ -0,0 +1,399 @@
...
+            )
+
+        return df.reset_index(drop=True, inplace=False)
+
+    def get_metric(
+        self,
+        metric: MetricConfiguration,
+    ) -> Any:
+        """return the value of the requested metric."""
+        return self.get_metrics(
```
NathanFarmer, line 125: This could call `compute_metrics`. We could rename method to `compute_metric`.

great_expectations/validator/validator.py
```diff
@@ -146,37 +139,202 @@ def __init__(
...
+    def get_metric(
+        self,
+        metric: MetricConfiguration,
+    ) -> Any:
+        """
+        Convenience method that returns the value of the requested metric.
+
+        (To be deprecated in favor of using methods in "MetricsCalculator" class.)
+        """
+        return self._metrics_calculator.get_metric(metric=metric)
```
NathanFarmer, line 278: This could call `compute_metric`

scripts/check_type_hint_coverage.py
NathanFarmer: @alexsherstinsky is it possible to modify it back to 1000 then?

NathanFarmer (changes requested):
I have a few more questions!

scripts/check_type_hint_coverage.py
Kilo59: I made this change.
`1000` will cause the CI to fail.
The changes to what is being excluded/included in the mypy config, sort of break the way this script used to operate.

great_expectations/validator/validator.py
alexsherstinsky: @NathanFarmer Unfortunately no -- the return types are different.  One uses metric_name as the key, the other uses metric ID as the key.

great_expectations/validator/metrics_calculator.py
alexsherstinsky: @NathanFarmer Not exactly -- the signatures are different, and if we did, we would repeat the transformation already in `get_metrics()`.

great_expectations/validator/validator.py
alexsherstinsky: @NathanFarmer Unfortunately cannot without code duplication.

great_expectations/validator/metrics_calculator.py
NathanFarmer: @alexsherstinsky see my latest comment on line 135. This would mean `get_metric` and `get_metric` go away entirely and would be replaced with `compute_metric` and `compute_metrics`. Anything from the legacy API would call `compute_metric` or `compute_metrics` under the hood.

great_expectations/validator/validator.py
NathanFarmer: @alexsherstinsky I'm not sure what you mean. This would call `self._metrics_calculator.compute_metric(metric=metric)`.

great_expectations/validator/validator.py
NathanFarmer: What I'm saying is there is no need for `MetricsCalculator` to use `metric_name` as a key in a dictionary when `MetricConfiguration` already has the `metric_name` as an attribute.

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
```diff
@@ -275,14 +268,39 @@ def get_column_standard_deviation_metric_multi_batch_parameter_builder(
                 metric_value_kwargs=None,
             )
 
+        @staticmethod
+        def build_metric_multi_batch_parameter_builder(
+            metric_name: str,
+            metric_domain_kwargs: Optional[
+                Union[str, dict]
+            ] = DOMAIN_KWARGS_PARAMETER_FULLY_QUALIFIED_NAME,
```
Kilo59, line 277: `Optional[Union[str, dict]]` is the same as `Union[str, dict, None]`

For `metric_domain_kwargs`, is `None` actually a valid value?

`Optional` doesn't really mean "you don't need to provide this value", it means "`None` is a valid value".

great_expectations/validator/validator.py
alexsherstinsky: @NathanFarmer the types of arguments are not the same.

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
alexsherstinsky: @Kilo59 Indeed, `None` is a common value, and that was precisely the intended use of `Optional` type.  Thanks.

great_expectations/validator/metrics_calculator.py
alexsherstinsky: @NathanFarmer The `compute_metrics` uses a different signature (list of `MetricConfiguration`); moreover, its output is a dictionary whose keys are tuples (not strings).

great_expectations/validator/validator.py
NathanFarmer: @alexsherstinsky I understand that. We spoke offline and agree to disagree for now, but will continue this conversation.

NathanFarmer (approved):
Thank you for this!

PR #6171, opened 2022-10-19
[MAINTENANCE] Re-add missing Expectations to gallery and include package names

kenwade4:
- Move xml_parseable Expectation to semantic_types package
- Update build_gallery script to include package names in JSON file


assets/scripts/build_gallery.py
```diff
@@ -116,6 +116,9 @@ def get_expectation_file_info_dict(
 
     for file_path in sorted(files_found):
         file_path = file_path.replace(f"{repo_path}{os.path.sep}", "")
+        package_name = os.path.basename(os.path.dirname(os.path.dirname(file_path)))
+        if package_name == "expectations":
+            package_name = "core"
```
anthonyburdi, line 121: First time reviewing these files so take this with a grain of salt. I'm having trouble understanding what is happening here - the core expectations have a different nested directory structure than the contrib ones. It looks like we are handling them the same here - what am I missing? Do you think we can break this method up into at least two methods (one for core, one for contrib)?

anthonyburdi (approved):
LGTM - discussed some potential improvements to the script but won't implement at this time.

PR #6177, opened 2022-10-19
Make sure that quick stage check returns 0 if there are no problems

kenwade4:
- Make sure that quick stage check returns 0 if there are no problems


austiezr (approved):
🚀

PR #6179, opened 2022-10-20
[MAINTENANCE] Add unit test for MetricsCalculator.get_metric() Method -- as an example template

alexsherstinsky:
### Scope
A tight unit test for the `MetricsCalculator.get_metric(MetricConfiguration)` method is sought, such that the implementation of this test can be used as an example for developing other tight unit tests.

#### General Approach

Use the “Mock” approach for as many boundary objects/methods surrounding the core functionality of the method under test.

#### Particular Approach

* Since a method on `MetricsCalculator` is under test and `MetricsCalculator` is instantiated with `ExecutionEngine`, use a `mock.MagicMock` for the `ExecutionEngine` constructor argument.
* Since the logic of `MetricsCalculator.get_metric(MetricConfiguration)` consists of calling `MetricsCalculator.get_metrics(Dict[str, MetricConfiguration])` and extracting the key corresponding to `metric_name`, use `mock.patch` for the `MetricsCalculator.get_metrics` return value.
* Since `MetricConfiguration` is not under test, use mock.patch for the `MetricConfiguration.metric_name` and `MetricConfiguration.metric_domain_kwargs` properties, respectively.
* Make sure that `MetricConfiguration` constructor arguments and the mock.patch return values for the corresponding properties and method return values are mutually consistent.

- JIRA: GREAT-1103/GREAT-1316
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


tests/core/test_metrics_calculator.py
```diff
@@ -125,3 +127,31 @@ def test_column_partition_metric(
         )
         for idx, element in enumerate(results[desired_metric.id])
     )
+
+
+@mock.patch("great_expectations.execution_engine.execution_engine.ExecutionEngine")
+def test_get_metric(execution_engine: mock.MagicMock):
+    metrics_calculator = MetricsCalculator(execution_engine=execution_engine)
```
Kilo59: We should (almost) never use `MagicMock` without providing a spec.
https://realpython.com/python-mock-library/#avoiding-common-problems-using-specifications

An Even better alternative would be to define a proper double and just pass in the object to `MetricsCalculator(execution_engine=exec_engine_double)`.  Then we have proper isolation and a clear contract, and then we know for certain that a change to `ExecutionEngine` will never break this test.

```python
class ExecutionEngineStub:
    def some_method_the_test_needs(self) -> str:
           return "foobar"

def test_get_metric_does_something_specific():
    metrics_calculator = MetricsCalculator(execution_engine=ExecutionEngineStub())
    ...
```

tests/core/test_metrics_calculator.py
```diff
@@ -125,3 +127,31 @@ def test_column_partition_metric(
         )
         for idx, element in enumerate(results[desired_metric.id])
     )
+
+
+@mock.patch("great_expectations.execution_engine.execution_engine.ExecutionEngine")
+def test_get_metric(execution_engine: mock.MagicMock):
```
Kilo59: This test needs a better name, I can't tell from this name what components are incidental to the test (and can be doubled/mocked) and what parts are intrinsic to the test (and should not be doubled).
Also because I don't know exactly what the intent of the test is, it makes it hard for me to quickly provide useful feedback or suggestions.

tests/core/test_metrics_calculator.py
```diff
@@ -125,3 +127,31 @@ def test_column_partition_metric(
...
+    ) as mock_get_metrics_method:
+        metric_configuration = MetricConfiguration(
+            metric_name=metric_name,
+            metric_domain_kwargs=metric_domain_kwargs,
+        )
+        metric_value: Any = metrics_calculator.get_metric(metric=metric_configuration)
+        mock_get_metrics_method.assert_called_once_with(
+            metrics={metric_name: metric_configuration}
+        )
+        assert metric_value == mock_get_metrics_method()[metric_name]
```
Kilo59: You are mocking so many things here it makes the value of these assertions somewhat questionable, these assertions look like they are (mostly) testing your mocks themselves.

Here I'll come back to the name of the test, what is this test meant to prove?

tests/core/test_metrics_calculator.py
```diff
@@ -125,3 +127,31 @@ def test_column_partition_metric(
...
+        new_callable=mock.PropertyMock,
+        return_value=metric_name,
+    ), mock.patch(
+        "great_expectations.validator.metric_configuration.MetricConfiguration.metric_domain_kwargs",
+        new_callable=mock.PropertyMock,
+        return_value=IDDict(metric_domain_kwargs),
+    ), mock.patch(
+        "great_expectations.validator.metrics_calculator.MetricsCalculator.get_metrics",
+        return_value={metric_name: "my_metric_value"},
+    ) as mock_get_metrics_method:
```
Kilo59: ~You shouldn't be mocking the method you are trying to test, that defeats the purpose.~

Edit: I missed the `get_metrics` vs `get_metric` difference.

tests/core/test_metrics_calculator.py
```diff
@@ -125,3 +127,31 @@ def test_column_partition_metric(
...
+    metric_name = "my_metric_name"
+    metric_domain_kwargs: dict = {}
+    with mock.patch(
+        "great_expectations.validator.metric_configuration.MetricConfiguration.metric_name",
+        new_callable=mock.PropertyMock,
+        return_value=metric_name,
+    ), mock.patch(
+        "great_expectations.validator.metric_configuration.MetricConfiguration.metric_domain_kwargs",
+        new_callable=mock.PropertyMock,
+        return_value=IDDict(metric_domain_kwargs),
```
Kilo59: It seems like you should be using a `MetricConfiguationStub` here (or the real thing) instead patching in this stuff.
Makes the test much more confusing then it needs to be.

tests/core/test_metrics_calculator.py
```diff
@@ -125,3 +127,31 @@ def test_column_partition_metric(
         )
         for idx, element in enumerate(results[desired_metric.id])
     )
+
+
```
Kilo59, line 130:
```suggestion
@pytest.mark.unit
```

Kilo59 (changes requested):
It's very unclear what the value or intent of this test is.

I would suggest checking the actual code coverage of this test and make sure it's exercising the sections of the code you intend.

~From what I can see, you don't need to use `mock.patch` at all.~

tests/core/test_metrics_calculator.py
alexsherstinsky: @Kilo59 We are not using any methods inside `ExecutionEngine` here -- it is only present as an argument for instantiating `MetricsCalculator`.  Does this change your recommendation? Thanks.

tests/core/test_metrics_calculator.py
alexsherstinsky: @Kilo59 The name of this test reflects the fact that the `get_metric` method of the `MetricsCalculator` class is being tested.  I will modify it to better reflect its purpose.  Thanks.

tests/core/test_metrics_calculator.py
alexsherstinsky: @Kilo59 @donaldheppner I added comments to the test.  Happy to discuss.  Thanks!

tests/core/test_metrics_calculator.py
alexsherstinsky: @Kilo59 Let us discuss to make sure we agree on the best alternative.  Thanks!

tests/core/test_metrics_calculator.py
Kilo59: No, it just means the double can be very simple (AKA a dummy). As in, it doesn't have to do anything.
It might even be `None`.

```python
class ExecutionEngineDummy:
     pass

def test_get_metric_does_something_specific():
    metrics_calculator = MetricsCalculator(execution_engine=ExecutionEngineDummy())
    ...
    metrics_calculator = MetricsCalculator(execution_engine=object())
    ...
    metrics_calculator = MetricsCalculator(execution_engine=None)
 ```

File: tests/core/test_metrics_calculator.py (hunk 3 above)
Comment (Kilo59) [reply tests/core/test_metrics_calculator.py]: With the changes, the checks are easier to reason about. I'm fine with it now.

File: tests/core/test_metrics_calculator.py [hunk 7]
```diff
@@ -125,3 +127,32 @@ def test_column_partition_metric(
...
+        1) MetricsCalculator.get_metric() calls MetricsCalculator.get_metrics() exactly once; and
+        2) MetricsCalculator.get_metric() correctly retrieves result from dictionary, returned by
+           MetricsCalculator.get_metrics() by using the "metric_name" as the key.
+    """
+    metrics_calculator = MetricsCalculator(execution_engine=execution_engine)
+    metric_name = "my_metric_name"
+    metric_value = "my_metric_value"
+    with mock.patch(
+        "great_expectations.validator.metrics_calculator.MetricsCalculator.get_metrics",
+        return_value={metric_name: metric_value},
```
Kilo59:
```suggestion
        return_value={metric_name: metric_value, "wrong_metric": "wrong_value"},
```
@alexsherstinsky would adding this make sense? Maybe it's of marginal value, in which case, don't add it.

tests/core/test_metrics_calculator.py
Kilo59: I'm fine with what you have now, but if you want isolation from the real  `MetricConfiguation` you should be able to just use a `namedtuple` (or `IDDict`/`DotDict`) in place of the real thing.
I still don't see the need for the `mock.patch` here.

tests/core/test_metrics_calculator.py
alexsherstinsky: @Kilo59 What is the reason for name changes?  As far as I understand, we are testing for "actual" (not "wrong") values.  Or are you recommending an additional test?  Happy to discuss.   Thanks.

tests/core/test_metrics_calculator.py
alexsherstinsky: @Kilo59 The `MetricConfiguration` object is the one input to `get_metric()` -- the method under test.  So we need it, but its internals must be maximally mocked leaving only the aspects necessary for the test as un-mocked (to prevent test leakage).  I updated the docstring to emphasize this.  Thanks!

Kilo59 (approved):

tests/validator/test_metrics_calculator.py
```diff
@@ -125,3 +126,47 @@ def test_column_partition_metric(
...
+
+    In the present test case, the role of "ExecutionEngine" is limited to providing the required constructor argument to
+    the "MetricsCalculator" class (one of whose methods is under test); hence, a "DummyExecutionEngine" is employed.
+
+    The "with mock.patch" is used judiciously, trading off the focus on the functionality under test (i.e., avoiding
+    "test leakage") against going as far as mocking all non-essential methods and properties, favoring code readability.
+    """
+
+    # noinspection PyAbstractClass
+    class DummyExecutionEngine(ExecutionEngine):
```
Kilo59:
```suggestion
    class DummyExecutionEngine:
```

You don't need to inherit here, and you shouldn't because it's a dummy not a stub.

alexsherstinsky: @Kilo59 Then we will need cast.

Kilo59: You don't really need to cast either, but If you want to do it go ahead.

alexsherstinsky: @Kilo59 Otherwise, we get PyCharm warnings (and when these warnings are justified, I feel it is better to accommodate them).

PR #6180, opened 2022-10-20
[MAINTENANCE] Remove SQL for expect_column_discrete_entropy_to_be_between

kenwade4:
- Comment out _sqlalchemy method since all tests on all SQL backends for this fail


anthonyburdi (approved):
Thank you!

PR #6184, opened 2022-10-20
[MAINTENANCE] Incorporate ValidatorGraph Populating/Resolution Methods in ValidationGraph Module

alexsherstinsky:
### Scope
* Consolidate all `ValidationGraph` methods (e.g., `build_metric_dependency_graph()`, `resolve_validation_graph()`, etc.) and make appropriate methods in that class, such as `_parse()`, as private methods (no publicly-valuable usefulness).
* Update all references the `ValidationGraph` methods to point to this object (which now contains not only the state of the graph, but encapsulates all the methods, pertinent to this entity).
* Update tests to exercise the more effective object references properly.
* Move tests to modules that is most appropriate for their purpose.

- JIRA: GREAT-1103/GREAT-1318
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


donaldheppner (changes requested):
We're not ready for this change; the progress bar needs to be factored out, and computing metrics are not the concern of the `ValidationGraph` -- the graph should still be about graphing only.

Do not merge.

alexsherstinsky:
> 

@donaldheppner One comment: There is no computing metrics in the `ValidationGraph` right now.  There is, however, a call to `resolve_validation_graph()`, because it is an operation on the graph (that is why that method has the `self` reference -- so that it can parse the graph).  It, in turn, calls `execution_engine.resolve_metrics()` -- which is how the metrics get computed.  So this is both, graph operations, and graph resolution computation.  If you have ideas for a better module for these operations, I am happy to discuss.  Thanks.

tests/validator/test_validation_graph.py
```diff
@@ -185,3 +197,279 @@ def test_ExpectationValidationGraph_get_exception_info(
 
     assert left_exception in exception_info
     assert right_exception in exception_info
+
+
+@pytest.mark.integration
+def test_parse_validation_graph():
+    class FakePandasExecutionEngine:
```
Kilo59:
```suggestion
    class DummyPandasExecutionEngine:
```

I would call it a dummy, not a fake.
"Fakes" are typically full implementations.
https://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs

>Dummy objects are passed around but never actually used. Usually they are just used to fill parameter lists.

>Fake objects actually have working implementations, but usually take some shortcut which makes them not suitable for production (an [in memory database](https://martinfowler.com/bliki/InMemoryTestDatabase.html) is a good example).

alexsherstinsky: @Kilo59 I called this one "Fake" because it does a little bit more than a Dummy in the sense that it sets the type of the object.  Do you still think that it should be "Dummy"?  Thanks!

Kilo59: I see, on second look this, might be more of a "Stub"

>Stubs provide canned answers to calls made during the test, usually not responding at all to anything outside what's programmed in for the test.

`PandasExecutionEngineStub`

alexsherstinsky: @Kilo59 Thank you very much -- updated!

tests/validator/test_validation_graph.py
```diff
@@ -185,3 +197,279 @@ def test_ExpectationValidationGraph_get_exception_info(
 
     assert left_exception in exception_info
     assert right_exception in exception_info
+
+
+@pytest.mark.integration
```
billdirks: Is this an integration test or a unit test?

tests/validator/test_validation_graph.py
```diff
@@ -185,3 +197,279 @@ def test_ExpectationValidationGraph_get_exception_info(
...
+def test_parse_validation_graph():
+    class PandasExecutionEngineStub:
+        pass
+
+    PandasExecutionEngineStub.__name__ = "PandasExecutionEngine"
```
billdirks: Why rename?

tests/validator/test_validation_graph.py
```diff
@@ -185,3 +197,279 @@ def test_ExpectationValidationGraph_get_exception_info(
 
     assert left_exception in exception_info
     assert right_exception in exception_info
+
+
+@pytest.mark.integration
+def test_parse_validation_graph():
+    class PandasExecutionEngineStub:
```
billdirks: Does it matter that it's a `Pandas` here? If so, can you just use the real execution engine? That is 
```
import ge

ge.execution_engine.PandasExecutionEngine() 
```
instantiates a real execution engine. What does the stub get you here?

tests/validator/test_validation_graph.py
```diff
@@ -185,3 +197,279 @@ def test_ExpectationValidationGraph_get_exception_info(
...
+    )
+
+    for metric_configuration in validation_dependencies["metrics"].values():
+        graph.build_metric_dependency_graph(
+            metric_configuration=metric_configuration,
+            runtime_configuration=None,
+        )
+
+    ready_metrics, needed_metrics = graph._parse(metrics=dict())
+    assert len(ready_metrics) == 2 and len(needed_metrics) == 9
```
billdirks: This seems like the part you want to test based on the test name. The part above seems like setup that is repeated in a few tests. Could you move that out into its own helper setup function that each of the tests call?

billdirks (commented):
I only looked at one test and left a few comments.

tests/validator/test_validation_graph.py
alexsherstinsky: @billdirks Great catch -- all fixed.  Thanks!

tests/validator/test_validation_graph.py
alexsherstinsky: @billdirks We need the engine to be Pandas for the logic of the test to work properly -- so this stubbing achieves the purposes without instantiating the actual `PandasExecutionEngine` class.  Thanks.

tests/validator/test_validation_graph.py
alexsherstinsky: @billdirks This is in the name of using as little resource as possible to focus on the matter of the test; so we do this -- the execution engine we need is much less than the full `PandasExecutionEngine` instance, but a little bit more than a "Dummy" one.  Thanks.

tests/validator/test_validation_graph.py
alexsherstinsky: @billdirks Updated -- great idea -- thanks!

great_expectations/expectations/core/expect_column_kl_divergence_to_be_less_than.py
```diff
@@ -33,9 +34,12 @@
     parse_row_condition_string_pandas_engine,
     substitute_none_for_missing,
 )
+from great_expectations.validator.exception_info import ExceptionInfo
 from great_expectations.validator.metric_configuration import MetricConfiguration
 from great_expectations.validator.validation_graph import ValidationGraph
-from great_expectations.validator.validator import Validator
+
+logger = logging.getLogger(__name__)
+logging.captureWarnings(True)
```
donaldheppner, line 44: Why is this here?

great_expectations/validator/validation_graph.py
```diff
@@ -1,9 +1,21 @@
...
+
+import great_expectations.exceptions as ge_exceptions
 from great_expectations.core.expectation_configuration import ExpectationConfiguration
+from great_expectations.execution_engine import ExecutionEngine
+from great_expectations.expectations.registry import get_metric_provider
 from great_expectations.validator.exception_info import ExceptionInfo
 from great_expectations.validator.metric_configuration import MetricConfiguration
 
+logger = logging.getLogger(__name__)
+logging.captureWarnings(True)
```
donaldheppner, line 15: why is this here?

tests/validator/test_validation_graph.py
```diff
@@ -38,14 +50,62 @@ def expect_column_values_to_be_unique_expectation_config() -> ExpectationConfigu
...
     expect_column_values_to_be_unique_expectation_config: ExpectationConfiguration,
 ) -> ExpectationValidationGraph:
     return ExpectationValidationGraph(
         configuration=expect_column_values_to_be_unique_expectation_config
     )
 
 
+@pytest.fixture
+def expect_column_value_z_scores_to_be_less_than_expectation_validation_graph(
+    expect_column_value_z_scores_to_be_less_than_expectation_config,
```
donaldheppner: this parameter is unused

tests/validator/test_validation_graph.py
```diff
@@ -38,14 +50,62 @@ def expect_column_values_to_be_unique_expectation_config() -> ExpectationConfigu
...
+):
+    class PandasExecutionEngineStub:
+        pass
+
+    PandasExecutionEngineStub.__name__ = "PandasExecutionEngine"
+    pandas_execution_engine_stub = cast(ExecutionEngine, PandasExecutionEngineStub())
```
donaldheppner: this code and class is duplicated in a number of tests, consider making it a fixture

tests/validator/test_validation_graph.py
```diff
@@ -133,30 +193,41 @@ def test_ValidationGraph_add(metric_edge: MetricEdge) -> None:
 @pytest.mark.unit
 def test_ExpectationValidationGraph_constructor(
     expect_column_values_to_be_unique_expectation_config: ExpectationConfiguration,
-    expectation_validation_graph: ExpectationValidationGraph,
+    expect_column_values_to_be_unique_expectation_validation_graph: ExpectationValidationGraph,
```
donaldheppner: I don't think this test is necessary; there is no complexity in the constructor and the code is covered in every other test

tests/validator/test_validation_graph.py
```diff
@@ -178,10 +249,226 @@ def test_ExpectationValidationGraph_get_exception_info(
...
 
     assert left_exception in exception_info
     assert right_exception in exception_info
+
+
+@pytest.mark.unit
+def test_parse_validation_graph(
+    expect_column_value_z_scores_to_be_less_than_expectation_validation_graph: ValidationGraph,
+):
+    available_metric: Dict[Tuple[str, str, str], Any]
```
donaldheppner: this declaration can be combined with the assignment

tests/validator/test_validation_graph.py
```diff
@@ -178,10 +249,226 @@ def test_ExpectationValidationGraph_get_exception_info(
...
+
+    graph = ValidationGraph(execution_engine=pandas_execution_engine_stub)
+    try:
+        graph.build_metric_dependency_graph(
+            metric_configuration=MetricConfiguration(
+                "column_values.not_a_metric", IDDict()
+            ),
+        )
+    except ge_exceptions.MetricProviderError as e:
+        graph = e
```
donaldheppner: seems strange to assign graph of an Error, consider another variable here for readability

tests/validator/test_validation_graph.py
```diff
@@ -178,10 +249,226 @@ def test_ExpectationValidationGraph_get_exception_info(
...
+
+    runtime_configuration = {
+        "catch_exceptions": True,
+        "result_format": {"result_format": "BASIC"},
+    }
+
+    execution_engine = PandasExecutionEngine()
+
+    # noinspection PyUnusedLocal
+    validator = Validator(execution_engine=execution_engine, batches=[batch])
```
donaldheppner: I would not expect to be referencing the Validator in a test of the Validation Graph

donaldheppner: this variable is unused; will this test pass without it? If not, why not (that would be kind of scary)?

tests/validator/test_validation_graph.py
```diff
@@ -178,10 +249,226 @@ def test_ExpectationValidationGraph_get_exception_info(
...
+
+    validation_dependencies = ExpectColumnMaxToBeBetween().get_validation_dependencies(
```
donaldheppner: we can include a type-hint here of `dict[str, MetricConfiguration]`

tests/validator/test_validation_graph.py
```diff
@@ -178,10 +249,226 @@ def test_ExpectationValidationGraph_get_exception_info(
...
+                "batch_identifiers": {
+                    "pipeline_stage_name": 0,
+                    "airflow_run_id": 0,
+                    "custom_key_0": 0,
+                },
+            }
+        )
+    )
+
+    expectation_configuration = ExpectationConfiguration(
```
donaldheppner: is this meant to be the aborted metric? consider naming `bad_configuration`

tests/validator/test_validation_graph.py
```diff
@@ -178,10 +249,226 @@ def test_ExpectationValidationGraph_get_exception_info(
...
+    exception_info = next(iter(aborted_metric_info_item["exception_info"]))
+    assert (
+        exception_info["exception_message"]
+        == 'Error: The column "not_in_table" in BatchData does not exist.'
+    )
+
+
+@pytest.mark.unit
+def test_progress_bar_config_enabled():
+    class DummyMetricConfiguration:
```
donaldheppner, line 353: These Dummys were called Stubs in the previous tests; I think Dummy is the more accurate term for both places

tests/validator/test_validation_graph.py
```diff
@@ -178,10 +249,226 @@ def test_ExpectationValidationGraph_get_exception_info(
...
+        pass
+
+    class DummyExecutionEngine:
+        pass
+
+    dummy_metric_configuration = cast(MetricConfiguration, DummyMetricConfiguration)
+    dummy_execution_engine = cast(ExecutionEngine, DummyExecutionEngine)
+
+    # ValidationGraph is a complex object that requires len > 3 to not trigger tqdm
+    with mock.patch(
```
donaldheppner, line 363: this series of mocks is duplicated between this method and the next; consider factoring into its own method or fixture

tests/validator/test_validation_graph.py
```diff
@@ -178,10 +249,226 @@ def test_ExpectationValidationGraph_get_exception_info(
...
+        len(
+            expect_column_value_z_scores_to_be_less_than_expectation_validation_graph.edges
+        )
+        == 33
+    )
+
+
+@pytest.mark.unit
+def test_populate_dependencies_with_incorrect_metric_name():
+    class PandasExecutionEngineStub:
```
donaldheppner, line 229: This is a dummy

tests/validator/test_validation_graph.py
```diff
@@ -38,14 +50,62 @@ def expect_column_values_to_be_unique_expectation_config() -> ExpectationConfigu
...
+
+    for metric_configuration in validation_dependencies["metrics"].values():
+        graph.build_metric_dependency_graph(
+            metric_configuration=metric_configuration,
+            runtime_configuration=None,
+        )
+
+    return graph
+
+
```
donaldheppner, line 100: I find this parameterization confusing. I would rather see this broken into two tests, one for table, one for column, and have a (non-test) method contain the assertions to reduce the duplication.

On deeper inspection, this is testing a simple assignment constructor. This test, `test_MetricEdge_init`, can be removed

donaldheppner (changes requested):

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner @Kilo59 specifically did not want a fixture, until we all had a chance to discuss.  Can we put it up on the agenda soon?  Thank you.

great_expectations/expectations/core/expect_column_kl_divergence_to_be_less_than.py
alexsherstinsky: @donaldheppner Because we are now using  `graph.resolve_validation_graph()` correctly and consistently, which is defined as is currently in this module -- and used with the logging.  This correct usage is in "great_expectations/validator/metrics_calculator.py" starting on line 178.  Thanks.

great_expectations/validator/validation_graph.py
alexsherstinsky: @donaldheppner Same reason as above.  We are now following the consistent usage pattern, as defined in "great_expectations/validator/metrics_calculator.py" starting on line 178. Thanks.

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner Thanks for the good catch -- I deleted it (in next commit).

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner No, it was a typo -- thanks for alluding me to it -- fixed (in next commit).

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner Fixed!

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner This is used for the side-effect of using the `Validator` instantiation -- it gets `BatchData` loaded.  I believe I can come up with a more straightforward way of doing this (given our recent results with refactoring `BatchManagement` concerns out of `Validator` and others).  I will do this first thing on Monday morning (10/24/2022).  Thanks.

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner Added proper type hints (in next commit).  Thanks.

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner This is meant to indicate an `ExpectationConfiguration`, which will depend on a metric that will fail, because it will be asked to access a column that does not exist.  Do we still want to change anything in the variable names?  I feel it is not necessary, but happy to discuss.  Thanks.

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner They are different.  Dummies here are true Dummies -- they are just placeholders.  However, the Stubs here actually do something (one thing, to be precise, they ensure that the object's type will be set as needed).  Thanks.

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner @Kilo59 was explicitly against doing this; we should discuss together.  Thanks.

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner According to our logic, it is a stub.  Thanks. /cc @Kilo59

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner I agree -- @cdkini do you mind if I delete this test, or if you created with a specific purpose in mind, please let us know soon?  Thanks!

tests/validator/test_validation_graph.py
Kilo59: My request not to have it in a fixture was because I think the value of being to clearly see that the Dummy doesn't define any behavior (and is a true "dummy") outweighs the minor duplication. Although I'm not sure if this example is really a Dummy or Stub.

In general do think we should use fixtures and remove duplication when the Double (Mock, Stub, Fake etc.) becomes more complex. Although I don't know where the line is there. I think we should decide this as a team.

tests/validator/test_validation_graph.py
Kilo59: @alexsherstinsky can you explain what the stub actually does?
Why does it need to have it's `__name__` modified like that? Does the test break if you leave it un-modified?

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner Deleted (in next commit) after confirming with @cdkini.  Thanks.

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner I agree -- @cdkini do you mind if I delete this test, or if you created with a specific purpose in mind, please let us know soon? Thanks!

alexsherstinsky: @donaldheppner Deleted (in next commit) after confirming with @cdkini. Thanks.

tests/validator/test_validation_graph.py
alexsherstinsky: @Kilo59 The goal in this and similar situations in this module, where I declared `PandasExecutionEngineStub`, is for the line 2214 in `get_validation_dependencies()` in `great_expectations/expectations/expectation.py` to return `True` -- we need the smallest possible object that can be perceived as having the type of `PandasExecutionEngine` without needing any other behaviors from it.  Can you recommend a better way, or is this a good way to do this?  Thanks!

great_expectations/validator/validation_graph.py
```diff
@@ -28,20 +40,199 @@ def id(self):
...
+        return {edge.id for edge in self._edges}
+
+    @property
+    def execution_engine(self) -> ExecutionEngine:
+        """Returns "ExecutionEngine" object, used for resolving "MetricConfiguration" objects in graph."""
+        return self._execution_engine
+
+    @execution_engine.setter
+    def execution_engine(self, value: ExecutionEngine):
+        self._execution_engine = value
```
Kilo59: ## Nit

Why do we use `@propety` here?
Functionally this is the same as just having a `self.execution_engine` attribute that we access and modify directly.

We aren't actually protecting the underlying `self._execution_engine`. So it seems like it's just more lines of code for no obvious benefit.

If it's about preserving typing details we can just do this...

```python
class ValidationGraph:

    def __init__(self, ..., execution_engine: ExecutionEngine) -> None:
        ...
        self.execution_engine = execution_engine
        
    ...

```

As an aside, the typing appears wrong. If `execution_engine` could be `None` as part of `__init__`, then the properties can also be `None`.

Can we type_check this module?

great_expectations/validator/validation_graph.py
```diff
@@ -28,20 +40,199 @@ def id(self):
 
 
 class ValidationGraph:
-    def __init__(self, edges: Optional[List[MetricEdge]] = None) -> None:
+    def __init__(
+        self,
+        edges: Optional[List[MetricEdge]] = None,
+        execution_engine: Optional[ExecutionEngine] = None,
```
Kilo59, line 46: Can `execution_engine` really be `None`? Should this be required?

tests/validator/test_validation_graph.py
Kilo59: No, that seems reasonable. Thanks for the explanation.

The only suggestion I would have is to define it like this, but this might not work.
https://stackoverflow.com/a/11769062/6304433

```python
class PandasExecutionEngineStub:
   __name__ = "PandasExecutionEngine"
```

alexsherstinsky: @Kilo59 I tried this before -- and tried it again just now -- it does not work, because the underlying code I mentioned recognizes the class as of the `PandasExecutionEngineStub` type, instead of as `PandasExecutionEngine` type.

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner I was able to remove the need to instantiate the `Validator` -- instantiating the `PandasExecutionEngine` with appropriate arguments is necessary and sufficient.  Thanks.

great_expectations/validator/validation_graph.py
alexsherstinsky: @Kilo59 Any idea why this module is not getting type checked automatically?  What should be added and where to force type checking on new modules that we create?  I am looking into your other questions in this comment.  Thank you!

alexsherstinsky: @Kilo59 This property is not used, so I deleted the related code.  Thanks for the great catch!

great_expectations/validator/validation_graph.py
alexsherstinsky: @Kilo59 Yes, the `execution_engine` here is optional (as in "not always needed").  Thanks.

tests/validator/test_validation_graph.py
```diff
@@ -308,42 +237,23 @@ class PandasExecutionEngineStub:
     pandas_execution_engine_stub = cast(ExecutionEngine, PandasExecutionEngineStub())
 
     graph = ValidationGraph(execution_engine=pandas_execution_engine_stub)
+    exc: Optional[ge_exceptions.GreatExpectationsError] = None
```
donaldheppner: Consider using pytest.raises instead: https://docs.pytest.org/en/stable/reference/reference.html?highlight=raises#pytest.raises

donaldheppner (changes requested):

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner Good point (incorporated/committed).  Thanks!

tests/validator/test_validation_graph.py
```diff
@@ -178,10 +178,209 @@ def test_ExpectationValidationGraph_get_exception_info(
...
+        )
+
+    assert (
+        e.value.message
+        == "No provider found for column_values.not_a_metric using PandasExecutionEngine"
+    )
+
+
+@pytest.mark.integration
+def test_resolve_validation_graph_with_bad_config_catch_exceptions_true():
```
donaldheppner, line 252: I think the code paths here can be tested without an integration test; we're not trying to test that the ExecutionEngine raises errors, we are trying to test that the ValidationGraph properly handles those errors.

This would be a good case for a Mock

tests/validator/test_validation_graph.py
donaldheppner: I'm fine with it not being a fixture, but a well named method here is then the way to go. Duplication in tests is still duplication, and there is value in knowing that both tests use the same setup

donaldheppner (changes requested):

tests/validator/test_validation_graph.py
alexsherstinsky: @donaldheppner Investigating....

alexsherstinsky: @donaldheppner This has been done.  Thanks.

great_expectations/validator/validation_graph.py
```diff
@@ -1,9 +1,21 @@
+import logging
```
Kilo59, line 1: @alexsherstinsky before merging this can you enable type-checking for this module?
Previously, this module only had 2 out-standing typing issues.

https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L146
To enable type-checking you need to remove the line I've linked above.

Kilo59 (changes requested):
Type-Checking

https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L146

great_expectations/validator/validation_graph.py
alexsherstinsky: @Kilo59 I have done it and committed -- now trying to reopen this PR, which I accidentally closed.  Thanks.

alexsherstinsky:
Trying to reopen this pull request.

alexsherstinsky:
The content of this (accidentally closed) pull request has been moved to #6228 which is presently open.  Thank you.

PR #6186, opened 2022-10-21
[MAINTENANCE] Add E2E tests for Cloud-backed `Datasource` CRUD

cdkini:
- Update `delete_datasource` to work with Cloud
- Write E2E test

### Definition of Done


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -724,7 +724,8 @@ def delete_datasource(
             raise ValueError(f"Datasource {datasource_name} not found")
 
         if save_changes:
-            self._datasource_store.delete_by_name(datasource_name)  # type: ignore[attr-defined]
+            datasource_config = datasourceConfigSchema.load(datasource.config)
+            self._datasource_store.delete(datasource_config)  # type: ignore[attr-defined]
```
cdkini, line 1018: We need a way to pass down the `id` to the store - using the config is the cleanest way (as opposed to just name)

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -370,6 +370,9 @@ def get_config_with_variables_substituted(
             )
         )
 
+    def delete_datasource(self, datasource_name: str) -> None:
+        super().delete_datasource(datasource_name=datasource_name, save_changes=True)
```
cdkini: `save_changes` rears its ugly head - I think we should refactor this out ASAP

great_expectations/data_context/store/datasource_store.py
```diff
@@ -125,17 +125,6 @@ def retrieve_by_name(self, datasource_name: str) -> DatasourceConfig:
...
-    def delete_by_name(self, datasource_name: str) -> None:
-        """Deletes a DatasourceConfig persisted in the store by it's given name.
-
-        Args:
-            datasource_name: The name of the Datasource to retrieve.
-        """
-        datasource_key: DataContextVariableKey = self._determine_datasource_key(
-            datasource_name=datasource_name
-        )
-        self.remove_key(datasource_key)
```
cdkini, line 137: Didn't really find a need for this anymore and think standardization between stores is important - let me know if you disagree but IIRC, we wanted to remove these helpers

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -724,7 +724,8 @@ def delete_datasource(
             raise ValueError(f"Datasource {datasource_name} not found")
 
         if save_changes:
-            self._datasource_store.delete_by_name(datasource_name)  # type: ignore[attr-defined]
+            datasource_config = datasourceConfigSchema.load(datasource.config)
```
Kilo59, line 1017: Sorry, not part of this PR, but for line `718` should that conditional be tweaked to account for empty strings as well?
```python
        if datasource_name is None:  # this would allow an '' which we probably don't want.
            raise ValueError("Datasource names must be a datasource name")
        ...
        if not datasource_name:
            raise ValueError("Datasource names must be a datasource name")
```

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59, line 1018: Can this `attr-defined` error be easily resolved?

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -370,6 +370,9 @@ def get_config_with_variables_substituted(
             )
         )
 
+    def delete_datasource(self, datasource_name: str) -> None:  # type: ignore[override]
```
Kilo59: Can we update the parent class signature so that we aren't doing an LSP violation here?

## Edit

image

Kilo59: 1. Is `datasource_name` being `Optional` on the Superclass correct? If not changing it there fixes that issue.

2. I'm guessing `save_changes` isn't an option because it's always going to be true? To deal with that we can raise an error if they provide something other than false. And that way we preserve LSP.

```python
def delete_datasource(self, datasource_name: str, save_changes: bool = True) -> None:
    if save_changes is not True:
        raise ValueError(f"`save_changes` for `{self.__class__.__name__}.delete_datasource()` must be `True`")
    ...
```

tests/data_context/store/test_datasource_store.py
```diff
@@ -335,20 +335,6 @@ def test_datasource_store_retrieve_by_name(
     )
 
 
-@pytest.mark.unit
-def test_datasource_store_delete_by_name(
```
anthonyburdi, line 339: Should we add back in a unit test for the delete method? Or do we already have that somewhere else? (Besides the e2e test I mean)

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: an easy fix! good catch, I've changed this in the next commit

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: I don't believe so - the `self._datasource_store` attr is only ever instantiated by child classes. We likely want to change this but it may take some more work

great_expectations/data_context/data_context/cloud_data_context.py
cdkini: I think we want to remove `save_changes` entirely but that will be a more involved effort (probably immediately following this PR)

tests/data_context/store/test_datasource_store.py
cdkini: good call, added a unit test!

great_expectations/data_context/data_context/cloud_data_context.py
Kilo59: The only remaining issue I have here is with point `2`, that LSP violation is likely to actually break existing setups for users if they swap their Datacontext class to a Cloud one.

Example... currently this would break.
```python
my_datacontext.delete_datasource("datasourous_rex", save_changes=True)
```

But really the only reason this should break is if they tried to do this...
```python
my_datacontext.delete_datasource("datasourous_rex", save_changes=False)
```

But we can entirely remove this pain point for moving to cloud with the suggested change above.

Kilo59 (changes requested):
As per below, we should address this LSP violation...
https://mypy.readthedocs.io/en/stable/error_code_list.html#check-validity-of-overrides-override

https://github.com/great-expectations/great_expectations/pull/6186#discussion_r1001914429

`delete_datasource()` shouldn't error if a user tries to set `save_changes = True`.

Maybe don't worry about the `Optional` `datasource_name` as much though.

Kilo59 (approved):

PR #6188, opened 2022-10-21
[MAINTENANCE] Migrate methods from `BaseDataContext` to `AbstractDataContext`

cdkini:
- To ensure that all classes within the context hierarchy have access to the same behaviors, we want to port these methods to the parent class.

### Definition of Done


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -790,6 +810,31 @@ def add_checkpoint(
...
+            checkpoint_config=checkpoint_config,
+            data_context=self,
+            name=name,
+        )
+
+        return checkpoint
+
+    def delete_checkpoint(
+        self,
+        name: Optional[str] = None,
```
Kilo59, line 1103: This seems like it should be required?

Same for the other `get_profiler()`, `delete_profiler()`, how does the Datacontext know which item to select without a name or identifier? Does it just grab the first one?

cdkini: so it's essentially an XOR - one needs to be passed in. I agree that it is confusing but since the nature of this PR is simply porting over the methods to the parent class, I think we should tackle this matter in a subsequent ticket.

Kilo59 (approved):

PR #6193, opened 2022-10-21
[MAINTENANCE] M/great 1313/abstract data context changes

billdirks:
Currently 3 properties are accessed on a `Batch` object when computing expectations: `id`, `data`, and `batch_definition`. The purpose of this PR is to eliminate the need to call `batch_definition` since we want to remove it (or greatly simplify it) from a zep `Batch` object.

The `batch_definition` is only used to to get the `execution_engine` in the call to `get_validator` on a data context. My strategy is to factor `get_validator_using_batch_list` into 2 parts. One part that gets the execution engine (the first part of the new `get_validator_using_batch_list`) and the second part, `_get_validator_using_batch_list`, that does everything else. This preserves the public function `get_validator_using_batch_list` while letting us reuse the functionality of the second part.

We don't have a `Batch` zep class defined yet and we are actively working on that. Because of that, I have the if statement on line `967` with a preceding comment which is a bit hacky. When a zep `Batch` is better defined and we understand the specifics of the workflow better this will be made neater.

### DIFF
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -912,12 +912,10 @@ class AbstractDataContext(ABC):
                     batch_request_list is not None,
                 ]
             )
-            != 1
+            > 1
         ):
             raise ValueError(
-                "Exactly one of batch, batch_list, batch_request, or batch_request_list can be specified: "
-                f"batch = {batch}, batch_list = {batch_list}, batch_request = {batch_request}, "
-                f"batch_request_list = {batch_request_list}"
+                "No more than one of batch, batch_list, batch_request, or batch_request_list can be specified"
             )
 
         if batch_list:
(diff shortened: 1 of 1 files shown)


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -980,23 +1007,32 @@ def get_validator_using_batch_list(
...
+        return self._get_validator_using_batch_list(
+            expectation_suite=expectation_suite,
+            batch_list=batch_list,
+            execution_engine=execution_engine,
+            include_rendered_content=include_rendered_content,
+            **kwargs,
+        )
 
+    # noinspection PyUnusedLocal
+    def _get_validator_using_batch_list(
```
billdirks, line 1023: This is the old `get_validator_using_batch_list` without the part that retrieved the `execution_engine`.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -955,13 +955,40 @@ def get_validator(
...
+        self._check_len_batch_list(batch_list)
+        # For ZEP datasources, we are trying to simplify our datatypes and we don't currently have `batch_definitions`
+        # defined on a `Batch`. Because of this, we grab the execution engine using a batch_request object.
+        # Once a ZEP `Batch` gets defined we will clean this up and with a more appropriate type check.
+        # NOTE: for non-ZEP workflows, both branches are equivalent.
+        # NOTE: `batch_request_list != [None]` is a workaround when no batch, batch_list, batch_request,
+        # batch_request_list is passed in. This should not be allowed but many tests currently do this and need to be
+        # fixed.
+        execution_engine: ExecutionEngine
+        if batch_request_list and batch_request_list != [None] and datasource_name:
```
billdirks, line 967: The expectation calculation pathway will go down this first branch which it what the ZEP workflow will do. This avoids looking at a `batch.batch_definition` to figure out the `execution_engine`. Both pathways generate the same execution engine, they just look for the datasource in different places.

billdirks:
I'm going to try an alternative approach which I believe will be cleaner.

PR #6194, opened 2022-10-24
[MAINTENANCE] ZEP POC - Asset Type Registration

Kilo59:

This PR contains POCs for what declaring supported `AssetType` for a `Datasource` would look like.

- [x] use a logger instead of `print`
- [x] mechanism for registering all the various `Datasource` -> `DataAsset` types
- [x] define custom `BiDict` for storing and retrieving types & a simplified name. Example: `PandasDatasource` -> `'pandas'`
- [x] generate `add_<ASSET_NAME>` methods to the `Datasource` class
   - [x] `add_<ASSET_NAME>` methods attach and return `AssetType`
   - [x] can retrieve assets by name

## Notes for reviewers

In my latest commit I broke up some of these modules to avoid future merge conflicts.
But this makes these changes much harder to review, below is the `metadatasource.py` module before the refactor work.

I'm also happy to drop this last commit and do the refactor in a fast-follow PR.

https://github.com/great-expectations/great_expectations/blob/a6c1e70ef4b6b5abe072c4ffc4709a7ec4e7ede0/great_expectations/zep/metadatasource.py

### Example Datasource + DataAsset defintion

```python
class FileAsset:
    file_path: pathlib.Path
    delimiter: str
    ...


class MyOtherAsset:
    foo: str
    bar: List[int]


class PandasDatasource(metaclass=MetaDatasource):

    name: str
    asset_types = [FileAsset, MyOtherAsset]

    def __init__(self, name: str):
        self.name = name
```


tests/zep/test_metadatasource.py
```diff
@@ -0,0 +1,124 @@
...
+    class RedDatasource(metaclass=MetaDatasource):
+        assets: MutableMapping[str, DataAsset]
+        asset_types = [BlueAsset]
+
+        def __init__(self, name: str) -> None:
+            self.name = name
+            self.assets = {}
+
+        def get_asset(self, asset_name: str) -> DataAsset:
+            return self.assets[asset_name]
```
Kilo59, line 89: It would be nice if these methods could be inherited.

Normally we could actually define a default implementation for the `get_asset()` method, and the implementor could choose to use it (inheriting from the protocol) or use their own.

```python
    class BlueAsset(DataAsset):
        pass

    class RedDatasource(Datasource, metaclass=MetaDatasource):
        assets: MutableMapping[str, DataAsset]
        asset_types = [BlueAsset]

```

But doing so we get this error `TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases`

We could deal with this in a couple of ways...
1. Make `Datasource` an ABC instead of `Protocol`
2. Inject the `get_asset` method as part of `MetaDatasource.__new__` - although I feel we might be over-using the metaclass magic here.
3. Fix the Metaclass conflict???

great_expectations/zep/bi_directional_dict.py
```diff
@@ -0,0 +1,62 @@
+import pprint
+from collections import UserDict
+from typing import Hashable, Iterable, Mapping, Optional, TypeVar
+
+T = TypeVar("T", bound=Hashable)
+
+
+class BiDict(
```
cdkini: we have a similar impl in `util.py`: https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/util.py#L109

I'm good keeping this one if you prefer

Kilo59: Ah... yeah this looks familiar... probably where I got the idea 😆 .
I may keep this one due to the generics support. But I'll be sure and remove one or combine them once we leave the POC phase.

great_expectations/zep/bi_directional_dict.py
```diff
@@ -0,0 +1,62 @@
...
+    UserDict,
+    Mapping[T, T],
```
billdirks: I believe this would be better with `Mapping[Hashable, Hashable]`. You use `Mapping[Hashable, Hashable]` in the declaration of `__dict` in `__init__` so the type checker will likely resolve it to this anyways.

The rest of this comment a digression into generics.

Usually, you use the generic `T` in the class declaration when you need to use that type in the class definition. That is, here you'd use `__dict: Mapping[T, T]` instead of `Mapping[Hashable, Hashable]`.

When the typechecker performs the check you can think of `T` being resolved to a concrete class (the exact class depends on how the class is used). This is useful when you want to enforce types matching like `def combine(a : T, b: T) -> T` which takes 2 objects of the same type and produces a third object of that type. So `Mapping[T, T]` would resolve with the `T`s being identical types, here `Hashable`. When I see this type of constraint on a mapping I think of the constraint that the keys are all identical type and the values are all identical type, perhaps different from the keys. So:
```
T = TypeVar("T", bound=Hashable)
S = TypeVar("S", bound=Hashable)

my_dict: Mapping[T, S]
```
So S and T both resolve to something bound by `Hashable` but may resolve to different things like `str` and `int`.

great_expectations/zep/bi_directional_dict.py
```diff
@@ -0,0 +1,62 @@
...
+    Because of this, all values must be Hashable.
+
+    If a Mapping-like object is passed as the first parameter, its key/values will be
+    unpacked (and combined with kwargs) into the new `BiDict` object.
+    """
+
+    def __init__(
+        self,
+        __dict: Optional[Mapping[Hashable, Hashable]] = None,
+        **kwargs: Hashable,
```
billdirks: What's the purpose of `kwargs` here? We could make this stricter than `UserDict`.

great_expectations/zep/bi_directional_dict.py
```diff
@@ -0,0 +1,62 @@
...
+    ):
+        if isinstance(__dict, Mapping):
```
billdirks: Does rewriting this function like this work:
```
__dict = __dict or {}
super().__init__(__dict, **kwargs)
```
It seems odd you are unpacking `__dict` into `kwargs` on line 26 and then passing both `__dict` and `kwargs` into `super`.

great_expectations/zep/bi_directional_dict.py
```diff
@@ -0,0 +1,62 @@
...
+        value = self.data.pop(key)
+        super().pop(value, None)
+
+    def __setitem__(self, key: Hashable, value: Hashable):
+        if key in self:
+            del self[self[key]]
+        if value in self:
+            del self[value]
+        super().__setitem__(key, value)
+        super().__setitem__(value, key)
```
billdirks: This data structure worries me a little bit since a key with an identical value to something previously inserted will overwrite it.

great_expectations/zep/bi_directional_dict.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __str__(self) -> str:
+        return str(self.data)
+
+    def contains_anyof(self, collection_: Iterable[Hashable]) -> bool:
+        return bool(set(collection_).intersection(self.keys()))
+
+
+# This makes BiDict pretty printable like a regular dict
+if isinstance(getattr(pprint.PrettyPrinter, "_dispatch"), dict):
```
billdirks: `pprint` seems to work for me so I wonder if this is necessary. We can discuss in a video chat tomorrow. In general I'd prefer to add to `asdict` method on this data structure (like dataclasses) than patching `pprint`.

great_expectations/zep/context.py
```diff
@@ -0,0 +1,149 @@
...
+            ds_type, suffix_to_remove="_datasource"
+        )
+
+        method_name = f"add_{simplified_name}"
+        LOGGER.info(
+            f"2a. Registering {ds_type.__name__} as {simplified_name} with {method_name}() factory"
+        )
+
+        pre_existing = cls.__source_factories.get(method_name)
+        if not pre_existing:
```
billdirks: nit: I'd prefer 
```
if pre_existing:
    raise ...
```
so this code can be less indented.

great_expectations/zep/context.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Contains a `.type_lookup` dict-like two way mapping between previously registered `Datasource`
+    or `DataAsset` types and a simplified name for those types.
+    """
+
+    type_lookup: BiDict[Union[str, type]] = BiDict()
+    __source_factories: Dict[str, SourceFactoryFn] = {}
+
+    @classmethod
+    def register_factory(
```
billdirks, line 46: I wonder if we could simplify this a bit. I think what we care about is unique `add` method names. 
I see 2 scenarios:
1.  2 identical class names in different files, say both name `MyFavoriteDatasource`. These will get converted to the same simple name and I want to see an error.
2. 2 non-identical class names, say `MyFavoriteDatasource` and `MyFavorite`, that get mapped to the same simple name. I want to see an error here too.
For assets, I can have 2 identical simplified method names, as long as they mapped to a different datasource, there shouldn't be an error.
So maybe we can just use a `dict` where the keys are datasource simplified names and whose value is a `set` of simplified data asset names. Does that sound right to you?

great_expectations/zep/context.py
```diff
@@ -0,0 +1,149 @@
...
+            return self.__source_factories[name]
+        except KeyError:
+            raise AttributeError(name)
+
+    def __dir__(self) -> List[str]:
+        """Preserves autocompletion for dynamic attributes."""
+        return [*self.factories, *super().__dir__()]
+
+
+class DataContext:
```
billdirks, line 116: Not something we need to do now, but for this initial zep push we aren't going to rewrite the `DataContext`. How do we incorporate this work there? Can we just add the `sources` property there? I am going to need to do something like this to integrate our work together and test computing expectations.

great_expectations/zep/interfaces.py
```diff
@@ -1,14 +1,28 @@
...
 
 
+class DataAsset(abc.ABC):
+    name: str
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+
+DataAssetType = Type[DataAsset]
```
billdirks: We can get rid of this or change `Type[DataAsset]` below to `DataAssetType`.

great_expectations/zep/interfaces.py
```diff
@@ -1,14 +1,28 @@
...
+
+
+@runtime_checkable
 class Datasource(Protocol):
     execution_engine: ExecutionEngine
-    asset_types: List[type]
+    asset_types: List[Type[DataAsset]]
```
billdirks, line 20: I like this!

great_expectations/zep/interfaces.py
```diff
@@ -21,3 +35,8 @@ def get_batch_list_from_batch_request(
         Returns:
             A list of batches. The list may be empty.
         """
+
+    def get_asset(self, asset_name: str) -> DataAsset:
```
billdirks, line 36: Oh that's interesting, I didn't realize we could add implementations to a `Protocol`. 
I agree with your comment below about protocol's and inheritance. If we have to inherit from it then we should just be using `ABC`s which I think is what we should be doing. More on that in your comment below.

tests/zep/test_metadatasource.py
billdirks: This is a really good point. I think we should change Datasource to not be a protocol now that we need implementation. I haven't tried this but reading the docs maybe something like this would work:

Make `MetaDatasource` a subclass of ABC. Make a new `Datasource` class whose metaclass is `MetaDatasource`. Implementers now inherit from `Datasource` and implement the abstract methods there. [A quick google shows one thing to do when subclassing](https://stackoverflow.com/questions/43858592/how-to-subclass-abcmeta) `ABC` is to make the `new` and `__init__` signatures align.
```
class MetaDatasource(metaclass=ABC):
   ...

class Datasource(metaclass=MetaDatasource)
  # This is now an abstract class and also has our MetaDatasource logic

# Datasource implementor does this
class MyDatasource(Datasource):
  ...
```

billdirks (commented):
I learned a lot from reading this PR, thank you!
Let's talk through it over video tomorrow.
I'd like to simplify how we avoid conflicts in registration.
I also agree with what your suggestion about making the `Protocol` an `ABC` and have outlined something we should try.

great_expectations/zep/context.py
Kilo59: >Can we just add the sources property there?

Yeah, that's what I was thinking.

great_expectations/zep/context.py
Kilo59: Agreed 😄

great_expectations/zep/bi_directional_dict.py
Kilo59: It was so that it could be instantiated with keywords, in the way a real `dict` can.
```python
dict(foo="bar", fizz="buzz)
BIDict(foo="bar", fizz="buzz)
```

great_expectations/zep/bi_directional_dict.py
Kilo59: Ah initially I had something simpler but I was getting the k,v pairs `"__dict": None`.
But this should deal with that as well, and is simpler.

great_expectations/zep/bi_directional_dict.py
Kilo59: I initially intended to have this structure be purpose-built for our type-lookout (in which case it would not have allowed this) but then I went and tried to generalize it 😬 .

I was thinking I might then subclass it with our more specific needs in mind, but then never did it.

great_expectations/zep/bi_directional_dict.py
Kilo59: Being able to pretty print this structure is definitely not necessary 😄 , especially for this particular data structure because by the nature of it only accepting hashable values it's unlikely to be deeply nested.

I mostly just wanted to figure out how to do this, but I'm happy to take it out.

But it does need this in order to be pretty-printable, it won't span-lines out of the box. Unless we do something like first converting it to a `dict` (as you suggested).

```python
import string
d = {
    "alpha": {},
    "bravo": {
        "red": "RED",
        "blue": BiDict(foo="bar", letters=tuple(string.ascii_uppercase[:10])),
    },
}
pprint.pprint(d, width=1)
```

```python
# without PP impl
{'alpha': {},
 'bravo': {'blue': BiDict({'foo': 'bar', 'bar': 'foo', 'letters': ('A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J'), ('A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J'): 'letters'}),
           'red': 'RED'}}

# with PP impl
{'alpha': {},
 'bravo': {'blue': {'bar': 'foo',
                    'foo': 'bar',
                    'letters': ('A',
                                'B',
                                'C',
                                'D',
                                'E',
                                'F',
                                'G',
                                'H',
                                'I',
                                'J'),
                    ('A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J'): 'letters'},
           'red': 'RED'}}
   ```

File: great_expectations/zep/bi_directional_dict.py (hunk 3 above)
Comment (Kilo59) [reply great_expectations/zep/bi_directional_dict.py]: Good points.

I ran into some of these accidently when trying to record a Loom video of this, and found the type-checker was saying things about the returned values that weren't true.

File: great_expectations/zep/interfaces.py [hunk 14]
```diff
@@ -1,14 +1,28 @@
...
+DataAssetType = Type[DataAsset]
+
+
+@runtime_checkable
 class Datasource(Protocol):
     execution_engine: ExecutionEngine
-    asset_types: List[type]
+    asset_types: List[Type[DataAsset]]
+    name: str
+    assets: MutableMapping[str, DataAsset]
```
Kilo59:
```suggestion
    assets: Dict[str, DataAsset]
```

great_expectations/zep/metadatasource.py
```diff
@@ -3,155 +3,89 @@
...
+                f"  {asset_type.__name__} method `{method_name}()` already defined"
+            )
+        attr_annotations = asset_type.__dict__.get("__annotations__", "")
+
+        # TODO: update signature with `attr_annotations`
+        def _add_asset(self: Datasource, name: str, *args, **kwargs):
+            LOGGER.info(f"6. Creating `{asset_type.__name__}` '{name}' ...")
+            data_asset = asset_type(name, *args, **kwargs)
+            self.assets[name] = data_asset
+            return data_asset
```
Kilo59: Don't do this.
Datasources must define the asset-adding methods.

Kilo59: Let's use a decorator to enforce the asset_method is added

Kilo59: We are ejecting the `_add_asset` injection stuff for now.
We may revisit this in the future.

For future reference here is the last commit that contained this logic.
https://github.com/great-expectations/great_expectations/blob/ed7953d30ec7ed49296df4c8d1a2f32232d6b5a6/great_expectations/zep/metadatasource.py

great_expectations/zep/bi_directional_dict.py
Kilo59: I took this out.

tests/zep/test_metadatasource.py
Kilo59: Getting down some notes from the discussion yesterday.

This is the method we are going to pursue, but we will need to add a conditional to `MetaDatasource.__new__` to check if the `cls` is a concrete `Datasource` or the `DatasourceABC`.
```python
# psuedocode
if CLASS is DATASOURCE_ABC:
   return super().__new__()
else:
    DO_FANCY_HOOK_STUFF
```

great_expectations/zep/context.py
Kilo59: As discussed offline.
The `TypeLookup` class will now enforce the uniqueness of the simple name -> type mappings. Throwing errors if there is a conflict.

This is something to be aware of in the future though, we don't want factory registration & type registration to become decoupled.

great_expectations/zep/context.py
```diff
@@ -0,0 +1,149 @@
...
+
+class _SourceFactories:
+    """
+    Contains a collection of datasource factory methods in the format `.add_<TYPE_NAME>()`
+
+    Contains a `.type_lookup` dict-like two way mapping between previously registered `Datasource`
+    or `DataAsset` types and a simplified name for those types.
+    """
+
+    type_lookup = TypeLookup()
```
billdirks, line 42: I find `type_lookup` hard to reason about because it does a lot of work. It's a bidirectionally dictionary that stores both information about datasources and data assets. It would be easier for me if we broke this up. I also think we will need to namespace data assets to datasources. How do you feel about making a `data_asset_type_lookup` and a `datasource_type_lookup`?

great_expectations/zep/context.py
```diff
@@ -0,0 +1,149 @@
...
+            raise ValueError(f"{simplified_name} factory already exists")
+
+        # TODO: simplify or extract the following datasource & asset type registration logic
+        asset_types = asset_types or []
+        asset_type_names = [
+            _get_simplified_name_from_type(t, suffix_to_remove="_asset")
+            for t in asset_types
+        ]
+
+        # NOTE: This check is a shortcut. What we need to protect against is different asset types
```
billdirks: I'd make this a TODO. I think we'll need to namespace the assets to datasources since I can see an asset type being used by more than 1 datasource. We can defer that for now but I'd be more explicit here like:
`# TODO: We should namespace the asset type to the datasource so different datasources can reuse asset types.`

great_expectations/zep/context.py
billdirks: I have tried this and it worked! We will have to `import` the datasources in the data context module to make sure the classes are loaded.

great_expectations/zep/metadatasource.py
```diff
@@ -3,155 +3,59 @@
...
-            raise AttributeError(name)
-
-    def __dir__(self) -> List[str]:
-        return ["register_factory", "factories", *self.factories]
-
-
-class MetaDatasouce(type):
-    def __new__(meta_cls, cls_name, bases, cls_dict) -> MetaDatasouce:
-        print(f"1. {meta_cls.__name__}.__new__() for `{cls_name}`")
+        asset_types: List[type] = cls_dict.get("asset_types")
```
billdirks:
```suggestion
        asset_types: List[Type[DataAsset]] = cls_dict.get("asset_types")
```

great_expectations/zep/metadatasource.py
```diff
@@ -3,155 +3,59 @@
...
+        LOGGER.info(f"1b. Extracting Asset details - {asset_types}")
+        if asset_types:
```
billdirks: I think we can remove this check now that we know `asset_types` is a `List[Type[DataAsset]]`?

great_expectations/zep/context.py
```diff
@@ -0,0 +1,149 @@
...
+
+    type_lookup = TypeLookup()
+    __source_factories: Dict[str, SourceFactoryFn] = {}
+
+    @classmethod
+    def register_factory(
+        cls,
+        ds_type: type,
+        fn: SourceFactoryFn,
+        asset_types: Optional[List[type]] = None,
```
billdirks:
```suggestion
        asset_types: Optional[List[Type[DataAsset]]] = None,
```
Should we make `asset_types` required?

great_expectations/zep/type_lookup.py
```diff
@@ -0,0 +1,53 @@
+from collections import UserDict
+from typing import Hashable, Iterable, Mapping, Optional
+
+
+class TypeLookupError(ValueError):
+    pass
+
+
+class TypeLookup(
```
billdirks, line 9: Thank you! This is a lot clearer to me now and I appreciate it!

billdirks (commented):
Thanks for doing this! I'd like to talk a little bit more about `type_lookup` and if we can break that up or if that would make something else hard. Otherwise LGTM! This is exciting!

great_expectations/zep/context.py
Kilo59: I was thinking we might need to do something exactly like that, but I was punting on that until I got to actually integrating the config loading/deserialization stuff.

great_expectations/zep/metadatasource.py
Kilo59: Yeah, we can rely on the type-checker for this.

great_expectations/zep/context.py
Kilo59: We can, it doesn't really make sense to define an `DataSource` without an asset, apart from making it easier to demo defining Datasource classes 😄

great_expectations/zep/context.py
Kilo59: I'll update this.

billdirks (approved):
LGTM, thanks!

PR #6201, opened 2022-10-25
[MAINTENANCE] Port over usage statistics enabled methods from `BaseDataContext` to `AbstractDataContext`

cdkini:
- Move additional methods to parent class to allow for consistency across the `DataContext` hierarchy

### Definition of Done


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -23,6 +25,13 @@
     cast,
 )
 
+try:
+    from typing import Literal  # type: ignore[attr-defined]
+except ImportError:
+    # Fallback for python < 3.8
+    from typing_extensions import Literal  # type: ignore[misc]
```
NathanFarmer, line 32: Is this preferred over checking for version using `version` library?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -37,13 +46,18 @@
...
 from great_expectations.core.serializer import (
     AbstractConfigSerializer,
     DictConfigSerializer,
 )
+from great_expectations.core.usage_statistics.events import UsageStatsEvents
 from great_expectations.core.util import nested_update
 from great_expectations.core.yaml_handler import YAMLHandler
 from great_expectations.data_asset import DataAsset
+from great_expectations.data_context.config_validator.yaml_config_validator import (
+    _YamlConfigValidator,
```
NathanFarmer, line 59: Are we ok with importing methods marked private? Does the underscore just denote that this isn't part of the public API?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2523,3 +3087,60 @@ def _determine_if_expectation_validation_result_include_rendered_content(
...
+    def test_yaml_config(  # noqa: C901 - complexity 17
+        self,
+        yaml_config: str,
+        name: Optional[str] = None,
+        class_name: Optional[str] = None,
+        runtime_environment: Optional[dict] = None,
+        pretty_print: bool = True,
+        return_mode: Union[  # type: ignore[name-defined]
+            Literal["instantiated_class"], Literal["report_object"]
+        ] = "instantiated_class",
```
```suggestion
        return_mode: Literal["instantiated_class", "report_object"] = "instantiated_class",
```

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -474,6 +502,10 @@ def save_datasource(
         self._cached_datasources[datasource_name] = updated_datasource
         return updated_datasource
 
+    @usage_statistics_enabled_method(
+        event_name=UsageStatsEvents.DATA_CONTEXT_ADD_DATASOURCE.value,
```
NathanFarmer, line 502: Can we use the `str` mixin to get rid of these `value` calls?

Kilo59 (approved):
Can you update the usage of `Literal`, other than that, it looks good.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1744,6 +2112,120 @@ def build_batch_kwargs(
...
+        urls_to_open: List[str] = [site["site_url"] for site in data_docs_urls]
+
+        for url in urls_to_open:
+            if url is not None:
+                logger.debug(f"Opening Data Docs found here: {url}")
+                webbrowser.open(url)
+
+    def get_docs_sites_urls(
+        self,
+        resource_identifier=None,
```
NathanFarmer, line 2147: Would be nice to add an annotation to this parameter as it is the only one missing one in this signature.

NathanFarmer (commented):
Mostly questions from me!

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59: This is one of the examples on `mypy`.

```python
import sys
if sys.version_info >= (3, 8):
    from typing import Literal
else:
    from typing_extensions import Literal
```

Although I think I prefer how @cdkini has it, I don't love having to import `sys`. But I don't feel too strongly about it.

https://mypy.readthedocs.io/en/stable/runtime_troubles.html?highlight=typing_extensions#using-new-additions-to-the-typing-module

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: Yup the purpose is to denote that it is not to be used externally - this is how @anthonyburdi implemented it and I think it works here. We should perhaps standardize moving forward though

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: for sure! mind if I do that in a subsequent PR? The aim with this PR is to simply port over methods as-is

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: I agree! mind if I add it in a subsequent PR?

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: I'm good either way!

great_expectations/data_context/data_context/abstract_data_context.py
NathanFarmer: Works for me!

great_expectations/data_context/data_context/abstract_data_context.py
NathanFarmer: Fine with me!

great_expectations/data_context/data_context/abstract_data_context.py
NathanFarmer: I'm fine either way. `try/except` is less specific and will cause an extra import for the users who are <3.8. We also use `from packaging import version` in various places throughout the codebase.

PR #6208, opened 2022-10-26
[BUGFIX] Fix logic in ExpectationDiagnostics._check_renderer_methods method

kenwade4:
Since expect_column_values_to_match_regex has an _answer_renderer defined, the renderer_types set in the _check_renderer_methods method includes `answer`, which was not accounted for. When using set difference to see that the result matched `all_renderer_types`, it was not True and the check failed (even though there are diagnostic and prescriptive renderers defined).

Now we are using set intersection to check.

- Fix logic in ExpectationDiagnostics._check_renderer_methods method


anthonyburdi (approved):
LGTM!

PR #6210, opened 2022-10-26
[MAINTENANCE] Execution Engine linting & partial typing

Kilo59:
- remove `execution_engine` package from the flake8 exclude list
   - fix or ignore related linting issues
- selectively remove `execution_engine` sub-modules from typing ignore, focusing on low typing count modules and pandas related modules


great_expectations/execution_engine/execution_engine.py
```diff
@@ -327,7 +343,7 @@ def resolve_metrics(
             for k, v in metric_to_resolve.metric_dependencies.items():
                 if v.id in metrics:
                     metric_dependencies[k] = metrics[v.id]
-                elif self._caching and v.id in self._metric_cache:
+                elif self._caching and v.id in self._metric_cache:  # type: ignore[operator] # TODO: update NoOpDict
```
Kilo59, line 346: `NoOpDict` needs the `__contains__`/ `in` operator implemented, to deal with this issue.

The question I have here... is whether should it always return `False` or always return `True`? 
I think always returning `True` is a lot more dangerous.

great_expectations/execution_engine/execution_engine.py
```diff
@@ -196,7 +212,7 @@ def __init__(
         self._caching = caching
         # NOTE: 20200918 - this is a naive cache; update.
         if self._caching:
-            self._metric_cache = {}
+            self._metric_cache: Union[Dict, NoOpDict] = {}
```
cdkini, line 215: Do you know why we defined `NoOpDict`? Seems like an odd object

great_expectations/execution_engine/execution_engine.py
```diff
@@ -288,7 +304,7 @@ def get_batch_data(
     def get_batch_data_and_markers(self, batch_spec) -> Tuple[BatchData, BatchMarkers]:
         raise NotImplementedError
 
-    def resolve_metrics(
+    def resolve_metrics(  # noqa: C901 - 16
```
cdkini, line 307: 😢

cdkini (approved):

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -74,6 +87,9 @@
         "Unable to load pyspark; install optional spark dependency for support."
     )
 
+if TYPE_CHECKING:
+    from pyspark.sql import DataFrame
```
NathanFarmer: Why do we need to import this again?

great_expectations/execution_engine/sqlalchemy_dialect.py
```diff
@@ -37,7 +37,7 @@ def __hash__(self: GESqlDialect):
         return hash(self.value)
 
     @classmethod
-    def _missing_(cls, value: Any) -> None:
+    def _missing_(cls, value: Any) -> GESqlDialect:
```
NathanFarmer: Doesn't returning something from super imply that a non-GESqlDialect Enum type could be returned?

NathanFarmer (approved):
A few non-blocking questions from me.

great_expectations/execution_engine/sparkdf_execution_engine.py
Kilo59: It's because otherwise DataFrame could be None because of the try imports... But this brings up another issue which is now we have to install pyspark to do type-checking which might not be desirable.

I think I'll revert this.

great_expectations/execution_engine/execution_engine.py
Kilo59: No, but from context I would guess for cases where we don't want to actually cache/retrieve metrics.

But still don't know WHY we would want that.

great_expectations/execution_engine/sqlalchemy_dialect.py
Kilo59: Yes, `Any` is probably more appropriate here.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -678,8 +678,8 @@ def get_domain_records(  # noqa: C901 - 24
                     .where(
                         sa.not_(
                             sa.and_(
-                                sa.column(column_A_name) is None,
-                                sa.column(column_B_name) is None,
+                                sa.column(column_A_name) == None,  # noqa: E711
+                                sa.column(column_B_name) == None,  # noqa: E711
```
Kilo59, line 682: Ended up having to revert the `x is None` change.
The `column` `==`/`__eq__` operator must have some logic that returns `None`.

Luckily the tests caught it.

Issue #5929, opened 2022-09-04
[BUG] marshmallow package conflict with airflow


itaise:
Hi, 
In great expectations latest version (0.15.21) marshmallow specific packages requirement was [added](https://github.com/great-expectations/great_expectations/pull/5866)
However, this is not compatible with Airflow, for example, the few latest versions: 
https://raw.githubusercontent.com/apache/airflow/constraints-2.3.4/constraints-3.7.txt
https://raw.githubusercontent.com/apache/airflow/constraints-2.3.3/constraints-3.7.txt
https://raw.githubusercontent.com/apache/airflow/constraints-2.3.2/constraints-3.7.txt

require an **earlier** version of marshmallow (3.16 or 3.17), so our builds are failing on pip conflicts.

Can you please check?

Thanks a lot!

itaise:
@cdkini maybe you have some insight on that? Thanks a lot!

rdodev:
Hey @itaise @cdkini is working on a fix for this. He'll provide an update soon.

itaise:
thanks a lot!

itaise:
Hi, @cdkini  @rdodev  any updates on this? thanks!

orenewix:
Any news?

orenewix:
@cdkini @rdodev can you please provide an update, if this remains on the next Airflow upgrade Wix will have to stop using GE package

cdkini:
Apologies for the delayed response @itaise @orenewix.

Our current pin is as follows: `marshmallow>=3.7.1,<4.0.0`

I believe this was immediately updated after [the referenced PR above](https://github.com/great-expectations/great_expectations/pull/5866) was merged. Please let me know if you're still seeing issues with these updated bounds.

austiezr:
Closed as resolved by #6213 ; if this issue persists, feel free to re-open with further context.

PR #6213, opened 2022-10-26
[MAINTENANCE] Small clean-up of `Marshmallow` warnings (`missing` parameter changed to `load_default` as of 3.13)

Shinnnyshinshin:
- Small PR to clean up `Marshmallow` warnings I was getting when running Expectations tests
```
great_expectations/great_expectations/data_context/types/base.py:2551: RemovedInMarshmallow4Warning: The 'missing' argument to fields is deprecated. Use 'load_default' instead.
```
- This PR replaces `missing` parameter with `load_default` (value to use by default when loading (deserializing))
- `requirements.txt` now requires `marshmallow>=3.13`. Was previously `marshmallow>=3.7.1`.

### For more information 
- [StackOverflow thread that describes the change](https://stackoverflow.com/questions/71821237/marshmallow-field-not-loading-default-value)
- [Marshmallow changelog](https://marshmallow.readthedocs.io/en/stable/changelog.html#id8)


Kilo59:
~Do we need to~ I think we need to bump our min version of MM here to `3.13` ~?~
https://github.com/great-expectations/great_expectations/blob/develop/requirements.txt#L12

See also https://github.com/great-expectations/great_expectations/issues/5929
But I think we should be good. The issue above talks about airflow needing `3.16`

kenwade4:
Good catch @Kilo59

great_expectations/data_context/types/base.py
```diff
@@ -238,19 +238,19 @@ class Meta:
     module_name = fields.String(
         required=False,
         allow_none=True,
-        missing="great_expectations.datasource.data_connector.sorter",
+        load_default="great_expectations.datasource.data_connector.sorter",
```
Kilo59, line 241: You could do something like this....
```python
def mm_string_wrapper(load_default = None, **kwargs) -> fields.String:
   if MARSHMALLOW_DEPRECATED:
       return fields.String(missing=load_default, **kwargs)
   return fields.String(load_defaul=load_default, **kwargs)

module_name = mm_string_wrapper(required=False,
        allow_none=True,
        load_default="great_expectations.datasource.data_connector.sorter")
```

Kilo59 (approved):

PR #6223, opened 2022-10-27
[MAINTENANCE] Move all `requirements-dev*.txt` files to separate dir

cdkini:
- Move all requirements to a separate folder to clean up root of repo
- Update any references to point at new location

### Definition of Done


azure-pipelines-gcp-integration.yml
```diff
@@ -26,7 +26,7 @@ stages:
             displayName: 'Update pip'
 
           - script: |
-              pip install -r requirements-dev.txt --constraint constraints-dev.txt
+              pip install --requirement requirements-dev.txt --constraint constraints-dev.txt
```
cdkini, line 29: Let's opt for the more verbose option for better documentation in our scripts

azure-pipelines-expectation-gallery.yml
```diff
@@ -102,7 +102,12 @@ stages:
             displayName: 'Update pip'
 
           - script: |
-              pip install "google-cloud-bigquery-storage" -r requirements-dev-all-contrib-expectations.txt -e "contrib/cli" -e ".[dev]" -c constraints-dev.txt
+              pip install \
+                "google-cloud-bigquery-storage" \
+                --requirement reqs/requirements-dev-all-contrib-expectations.txt \
+                --editable "contrib/cli" \
+                --editable ".[dev]" \
+                --constraint constraints-dev.txt
```
cdkini, line 110: If longer than 1 or 2 commands, I've opted for a multi-line approach. I think we could probably stand to clean up more in these files but keeping the scope relatively contained

setup.py
```diff
@@ -33,9 +33,14 @@ def get_extras_require():
         "dev-tools",
         "all-contrib-expectations",
     )
-    rx_fname_part = re.compile(r"requirements-dev-(.*).txt")
-    for fname in sorted(glob("requirements-dev-*.txt")):
-        key = rx_fname_part.match(fname).group(1)
+    requirements_dir = "reqs"
+    rx_fname_part = re.compile(rf"{requirements_dir}/requirements-dev-(.*).txt")
+    for fname in glob(f"{requirements_dir}/*.txt"):
```
Kilo59, line 39: `pathlib` has `glob` & `match` methods as well.
And results are actually `Path` objects instead of strings.

I think it's fine to leave this as is though, just an FYI if you weren't already aware.

https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.match
https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob

setup.py
```diff
@@ -33,9 +33,14 @@ def get_extras_require():
         "dev-tools",
         "all-contrib-expectations",
     )
-    rx_fname_part = re.compile(r"requirements-dev-(.*).txt")
-    for fname in sorted(glob("requirements-dev-*.txt")):
-        key = rx_fname_part.match(fname).group(1)
+    requirements_dir = "reqs"
```
kenwade4, line 37: On line 33, can you change that from "dev-tools" to "tools"? It was something added incorrectly in https://github.com/great-expectations/great_expectations/pull/5944

azure-pipelines-gcp-integration.yml
Kilo59: In general, I agree. I slightly prefer the shorthand here though.
Not a blocker.

setup.py
kenwade4: FYI @billdirks @jcampbell for future reference

great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -211,11 +209,25 @@ class GEDependencies:
...
-    def __init__(self, requirements_relative_base_dir: str = "../../../") -> None:
-        self._requirements_relative_base_dir = file_relative_path(
-            __file__, requirements_relative_base_dir
-        )
-        self._dev_requirements_prefix: str = "requirements-dev"
+    DEV_REQUIREMENTS_PREFIX = "requirements-dev"
+    PRIMARY_REQUIREMENTS_FILE = "requirements.txt"
+
+    def __init__(self) -> None:
+        self._requirements_paths = self._init_requirements_paths()
```
cdkini, line 216: Since this really isn't a point of extensibility, I decided to refactor and use `pathlib` as much as possible.

setup.py
kenwade4, line 39: And after line 28, can you add "vertica" (which should have been added in https://github.com/great-expectations/great_expectations/pull/6145)

kenwade4 (changes requested):

setup.py
cdkini: done!

setup.py
cdkini: done!

cdkini: I know about `glob` but I didn't know about `match`!

Kilo59 (approved):

azure-pipelines-expectation-gallery.yml
kenwade4: Agreed... I'll make a ticket for myself to try and update all these places we are calling `pip install` multiple times to be more like this one (so all packages needed for a particular test stage are resolved and installed in the single command)

kenwade4 (approved):

PR #6231, opened 2022-10-27
[MAINTENANCE] Isolate Trino/MSSQL/MySQL tests in `dev` CI

cdkini:
- Only run the specific tests that are trigger by `--trino`, `--mssql`, or `--mysql` instead of rerunning tests that are already covered in other parts of the pipeline

### Definition of Done


tests/profile/test_basic_suite_builder_profiler.py
```diff
@@ -276,10 +276,11 @@ def test__create_expectations_for_string_column(non_numeric_high_card_dataset):
...
     "ignore:DataAsset.remove_expectations*:DeprecationWarning:great_expectations.data_asset"
 )
-@pytest.mark.skipif(
-    is_library_loadable(library_name="trino"),
-    reason="datetime doesnt exist in Trino",
-)
+# @pytest.mark.skipif(
+#     is_library_loadable(library_name="trino"),
+#     reason="datetime doesnt exist in Trino",
+# )
```
cdkini: Our "skip if" logic should be done in one canonical way - let's make that marks.

azure-pipelines-dev.yml
```diff
@@ -49,374 +49,11 @@ variables:
 
 
 stages:
-  - stage: scope_check
```
cdkini: Temporary deletion so we can just test Trino's stage in the pipeline

tests/profile/test_basic_suite_builder_profiler.py
```diff
@@ -276,10 +276,7 @@ def test__create_expectations_for_string_column(non_numeric_high_card_dataset):
 @pytest.mark.filterwarnings(
     "ignore:DataAsset.remove_expectations*:DeprecationWarning:great_expectations.data_asset"
 )
-@pytest.mark.skipif(
-    is_library_loadable(library_name="trino"),
-    reason="datetime doesnt exist in Trino",
-)
+@pytest.mark.trino
```
cdkini: I grepped for Trino in our tests and marked any that rely on the backend.

azure-pipelines-dev.yml
```diff
@@ -373,8 +373,16 @@ stages:
               pip install pytest-cov pytest-azurepipelines
 
               # Run pytest
-              pytest tests --ignore 'tests/cli' --ignore 'contrib/cli/tests' --ignore 'tests/integration/usage_statistics' \
-                --mysql --napoleon-docstrings --junitxml=junit/test-results.xml --cov=. --cov-report=xml --cov-report=html
+              pytest tests/test_definitions tests/expectations \
```
cdkini, line 376: Per some research I did, I think these two directories encapsulate all of our backend-specific logic. Instead of retesting various GX components like RBP, Validator, and DataContext, let's narrow the scope of our testing so we only cover what is specific to Trino/MSSQL/MySQL/etc.

Kilo59 (approved):
Makes sense to me!

azure-pipelines.yml
```diff
@@ -308,9 +308,8 @@ stages:
               PYTEST+="--network=host --mount type=bind,source=${PWD}/script,target=/script "
               PYTEST+="greatexpectations/test:${DOCKER_TAG} /bin/bash -c \""
               PYTEST+="mysql --host=localhost --protocol=TCP --port=3306 --user=root --password='' --reconnect < /script/mysql_setup.sql && "
-              PYTEST+="pytest --mysql --random-order --ignore=tests/cli --ignore=tests/integration/usage_statistics "
               # The last \" closes the quote from bash -c \".
-              PYTEST+="--ignore=tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py\""
+              PYTEST+="pytest --mysql --random-order tests/test_definitions tests/expectations\"
```
billdirks:
```suggestion
              PYTEST+="pytest --mysql --random-order tests/test_definitions tests/expectations\""
```

You are missing the "extra" `"` at the end of this line. The first one cloaes the `"` started on line 309, `bash -c \""`

billdirks (changes requested):
Thank you!

One missing `"` but otherwise LGTM.

billdirks (approved):
LGTM!

We also talked about filing a followup ticket to use markers instead of limiting the directories we pass to pytest to guard against add future tests outside of these directories.

PR #6245, opened 2022-10-28
[MAINTENANCE] Split up `compatability` and `comprehensive` stages in `dev` CI to improve performance

cdkini:
- Split up `pytest tests` in half by using two parallel Azure workers

Proposed performance gains:
Screen Shot 2022-10-28 at 3 29 11 PM

### Definition of Done

### DIFF
--- a/azure-pipelines-dev.yml
+++ b/azure-pipelines-dev.yml
@@ -166,7 +166,7 @@ stages:
       vmImage: 'ubuntu-18.04'
 
     jobs:
-      - job: compatibility_matrix
+      - job: minimal
         condition: eq(stageDependencies.scope_check.changes.outputs['CheckChanges.GEChanged'], true)
         strategy:
           matrix:
@@ -210,8 +210,7 @@ stages:
 
           - script: |
               # Run pytest
-              pytest tests \
-                $(pytest_args) \
+              pytest $(pytest_args) \
                 --no-sqlalchemy \
                 --ignore 'tests/cli' \
                 --ignore 'contrib/cli/tests' \
@@ -277,8 +276,7 @@ stages:
               pip install pytest-cov pytest-azurepipelines
 
               # Run pytest
-              pytest tests \
-                $(pytest_args) \
+              pytest $(pytest_args) \
                 --postgresql \
                 --spark \
                 --ignore 'tests/cli' \
--- a/azure-pipelines-dev.yml
+++ b/azure-pipelines-dev.yml
@@ -166,9 +166,14 @@ stages:
       vmImage: 'ubuntu-18.04'
 
     jobs:
+      # Runs pytest without any additional flags
       - job: minimal
         condition: eq(stageDependencies.scope_check.changes.outputs['CheckChanges.GEChanged'], true)
         strategy:
+          # This matrix is intended to split up our sizeable test suite into two distinct components.
+          # By splitting up slow tests from the remainder of the suite, we can parallelize test runs
+          # at the cost of an additional Azure worker.
+          # To work as intendend, "standard" and "slow" should be equally performant
           matrix:
             standard:
               pytest_args: 'tests --ignore "tests/rule_based_profiler"'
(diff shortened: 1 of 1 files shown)


azure-pipelines-dev.yml
```diff
@@ -166,10 +166,14 @@ stages:
...
+      - job: minimal
         condition: eq(stageDependencies.scope_check.changes.outputs['CheckChanges.GEChanged'], true)
-        variables:
-          GE_pytest_opts: '--no-sqlalchemy'
+        strategy:
+          matrix:
+            standard:
+              pytest_args: 'tests --ignore "tests/rule_based_profiler"'
+            slow:
+              pytest_args: 'tests/rule_based_profiler'
```
cdkini, line 182: Rather than `pytest tests` taking a full hour, let's try and determine a strategy where we use 2 workers and have them tackle 30mins/30mins.

Thanks to refactors done on the config, we have the parallel workers to spare and have the opportunity to cut down our CI times.

azure-pipelines-dev.yml
```diff
@@ -329,7 +360,6 @@ stages:
     jobs:
       - job: mysql
         condition: eq(stageDependencies.scope_check.changes.outputs['CheckChanges.GEChanged'], true)
-        timeoutInMinutes: 120
```
cdkini, line 332: No longer necessary since we're cutting down test times

azure-pipelines-dev.yml
```diff
@@ -259,20 +287,23 @@ stages:
...
             env:
               GE_USAGE_STATISTICS_URL: ${{ variables.GE_USAGE_STATISTICS_URL }}
 
-          - script: |
-              # These are tests that are specific to Great Expectations Cloud.
-              # In order to ensure coverage, we run them during each CI/CD cycle.
-              pytest tests --cloud -m "cloud and not e2e"
-
-            displayName: 'cloud-overrides'
-
```
cdkini, line 275: We cover these in our unit test stage since we pass in `--cloud`

azure-pipelines-dev.yml
```diff
@@ -225,16 +245,24 @@ stages:
...
+            standard:
+              pytest_args: 'tests --ignore "tests/rule_based_profiler"'
+            slow:
+              pytest_args: 'tests/rule_based_profiler'
 
         services:
           postgres: postgres
 
-        variables:
-          GE_pytest_opts: '--postgresql --spark'
```
cdkini, line 236: Since these opts are consistent for all elements in the matrix, I've decided to hardcode.

azure-pipelines-dev.yml
Kilo59: This seems like something we should use marks for.

Or better yet let's use one mark and have the standard jut be `-m 'not MY_NEW_MARK'`

azure-pipelines-dev.yml
Kilo59: Isn't this to force stop run-away tests?
If so we should probably just reduce the `timeoutInMinutes` but not remove it.

billdirks (approved):
This is great!

azure-pipelines-dev.yml
Kilo59: I think this can be a follow-up though.

azure-pipelines-dev.yml
billdirks: By default this is 60 I believe. Removing it doesn't actually remove this constraint.

Kilo59 (approved):

azure-pipelines-dev.yml
cdkini: hmm but how can we mark a full directory? I mainly split it up this way because after some testing, I was able to determine that the RBP tests make up about half our runtime

cdkini: I don't want to use `@pytest.mark.slow` since we used that previously

Kilo59: I don't know how to mark a directory.

but to mark a whole module...

```python
import pytest

pytestmark = pytest.mark.MY_MARKER

```

https://docs.pytest.org/en/7.1.x/example/markers.html#marking-whole-classes-or-modules

PR #6251, opened 2022-10-31
Postgres datasource progress report

billdirks:
This is an initial implementation of a ZEP postgres datasource that can compute expectations. This is intended as a DRAFT and to solicit feedback. It will be cleaned up and broken into multiple PRs once we discuss.

### DIFF
--- a/great_expectations/core/batch.py
+++ b/great_expectations/core/batch.py
@@ -771,11 +771,12 @@ def get_batch_request_from_acceptable_arguments(  # noqa: C901 - complexity 21
     """
 
     if batch_request:
-        if not isinstance(batch_request, (BatchRequest, RuntimeBatchRequest)):
-            raise TypeError(
-                f"""batch_request must be an instance of BatchRequest or RuntimeBatchRequest object, not \
-{type(batch_request)}"""
-            )
+        # BDIRKS I removed this check so the ZEP BatchRequest type could pass this check.
+        # We will need to modify this to get ZEP working.
+        #if not isinstance(batch_request, (BatchRequest, RuntimeBatchRequest)):
+        #    raise TypeError(
+        #        f"""batch_request must be an instance of BatchRequest or RuntimeBatchRequest object, not {type(batch_request)}"""
+        #    )
         datasource_name = batch_request.datasource_name
 
     # ensure that the first parameter is datasource_name, which should be a str. This check prevents users
--- a/great_expectations/data_context/data_context/data_context.py
+++ b/great_expectations/data_context/data_context/data_context.py
@@ -33,6 +33,13 @@ from great_expectations.datasource import LegacyDatasource
 from great_expectations.datasource.new_datasource import BaseDatasource
 from great_expectations.rule_based_profiler.rule_based_profiler import RuleBasedProfiler
 
+# BDIRKS: All ZEP logic should be moved to abstract_data_context.py. However, the class
+# currently BaseDataContext doesn't call super().__init__().
+from great_expectations.zep.metadatasource import _SourceFactories, Datasource as ZepDatasource
+import great_expectations.zep.postgres_datasource
+
+
+
 logger = logging.getLogger(__name__)
 yaml = YAML()
 yaml.indent(mapping=2, sequence=4, offset=2)
@@ -234,6 +241,8 @@ class DataContext(BaseDataContext):
         ge_cloud_access_token: Optional[str] = None,
         ge_cloud_organization_id: Optional[str] = None,
     ) -> None:
+        # BDIRKS BDIRKS this should all move to abstract_data_context
+        self._sources: _SourceFactories = _SourceFactories(self)
         self._ge_cloud_mode = ge_cloud_mode
         self._ge_cloud_config = self._init_ge_cloud_config(
             ge_cloud_mode=ge_cloud_mode,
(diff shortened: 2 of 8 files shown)


great_expectations/core/batch.py
```diff
@@ -771,11 +771,12 @@ def get_batch_request_from_acceptable_arguments(  # noqa: C901 - complexity 21
...
-            raise TypeError(
-                f"""batch_request must be an instance of BatchRequest or RuntimeBatchRequest object, not \
-{type(batch_request)}"""
-            )
+        # BDIRKS I removed this check so the ZEP BatchRequest type could pass this check.
+        # We will need to modify this to get ZEP working.
+        #if not isinstance(batch_request, (BatchRequest, RuntimeBatchRequest)):
+        #    raise TypeError(
+        #        f"""batch_request must be an instance of BatchRequest or RuntimeBatchRequest object, not {type(batch_request)}"""
+        #    )
```
Kilo59, line 779: We should be able to deal with this by registering the `ZEP BatchRequest` as a virtual subclass (of one of these ABC classes)

https://docs.python.org/3/library/abc.html#abc.ABCMeta.register

great_expectations/data_context/data_context/data_context.py
```diff
@@ -33,6 +33,13 @@
 from great_expectations.datasource.new_datasource import BaseDatasource
 from great_expectations.rule_based_profiler.rule_based_profiler import RuleBasedProfiler
 
+# BDIRKS: All ZEP logic should be moved to abstract_data_context.py. However, the class
+# currently BaseDataContext doesn't call super().__init__().
+from great_expectations.zep.metadatasource import _SourceFactories, Datasource as ZepDatasource
+import great_expectations.zep.postgres_datasource
```
Kilo59, line 39: As per my understanding from the related slack thread.

Eventually each `DataContext` does call `super().__init__()`. 
`BaseDataContext` is somewhat of a wrapper around a real data context, and each of those instances does call the ABC `__init__`

https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/data_context/data_context/base_data_context.py#L246-L271

great_expectations/zep/interfaces.py
```diff
@@ -1,18 +1,64 @@
...
 class Datasource(Protocol):
     execution_engine: ExecutionEngine
     asset_types: List[type]
+    name: str
 
     def get_batch_list_from_batch_request(
-        self, batch_request: Union[BatchRequest, RuntimeBatchRequest]
-    ) -> List[Batch]:
+            self, batch_request: "BatchRequest"
+    ) -> "List[Batch]":
```
Kilo59, line 61: Should import `from __future__ annotations` here to get rid of these strings.

great_expectations/zep/postgres_datasource.py
```diff
@@ -0,0 +1,127 @@
...
+    method_name: str
+    column_name: str
+
+
+@dataclasses.dataclass(frozen=True)
+class TableAssetOptions(BatchRequestOptions):
+    # Depending on the splitter method passed in the possible values one could pass in vary. For example for the
+    # splitter `split_on_year_and_month` the possible kwarg keys are `year` and `month`. However, for the splitter
+    # method `split_on_year_and_month_and_day` the possible kwarg keys are `year`, `month` and `day`. Because of
+    # this the table_splitter_kwargs is a loosely typed dictionary.
```
Kilo59, line 35: Do we have typed splitter objects currently? Or is this a new feature?

great_expectations/zep/postgres_datasource.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+        self.assets: Dict[str, DataAsset] = {}
+
+    def add_table_asset(self,
+                        name: str,
+                        table_name: str,
+                        splitter: Optional[TableAssetSplitter]=None) -> TableAsset:
+        asset = TableAsset(datasource=self, name=name, table_name=table_name, splitter=splitter)
+        self.assets[name] = asset
+        return asset
```
Kilo59, line 89:
```suggestion
                        asset_name: str,
                        table_name: str,
                        splitter: Optional[TableAssetSplitter]=None) -> TableAsset:
        asset = TableAsset(datasource=self, name=asset_name, table_name=table_name, splitter=splitter)
        self.assets[asset_name] = asset
        return asset
```
Should we be explicit with both `names` here?

great_expectations/zep/postgres_datasource.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def get_asset(self, name) -> DataAsset:
+        return self.assets[name]
+
+    # When we have multiple types of DataAssets on a datasource, the batch_request argument will be a Union type.
+    # We could pattern match on this type to differentiate. Since we need to support python < 3.10, we can use an
+    # if/else condition instead of match and call a helper function fo each specific type.
```
Kilo59, line 96: This might be a case for actually using `singledispatch` 😄   🤷

great_expectations/data_context/data_context/data_context.py
billdirks: Sounds good. It looks like I will need to do a little plumbing in this file to plumb the variable on the abstract object to this one.

great_expectations/zep/interfaces.py
billdirks: Yes, this seems better.

great_expectations/zep/postgres_datasource.py
billdirks: I will look again but iirc the parameters for creating the batch are created by combining a few untyped `dict`s together.

great_expectations/zep/postgres_datasource.py
billdirks: My preference is to leave it as a `name` attribute since that gives a uniform interface over all these config objects. That's how it is now and I like it. Open to discussing though.

great_expectations/zep/postgres_datasource.py
billdirks: Yeah, that is a good thought. I'll update this comment.

PR #6254, opened 2022-10-31
[BUGFIX] Ensure that `Datasource` credentials are not persisted to Cloud/disk

cdkini:
- Previously, the `Datasource`'s `config` attr contained substituted values
    - This resulted in `save_datasource` sending sensitive credentials over the wire
    - New helper and updated workflow here ensures that we always have the unsubstituted config ready for persistence

### Definition of Done


great_expectations/data_context/config_validator/yaml_config_validator.py
```diff
@@ -190,7 +190,14 @@ def test_yaml_config(  # noqa: C901 - complexity 17
...
         usage_stats_event_name: str = "data_context.test_yaml_config"
 
+        # Chetan - 20221031 -   Based on the particular object type we are attempting to instantiate,
+        # we might need the original config or the substituted config. We should standardize this
+        # such that the various private helper methods below are responsible for substitutions
+        # whenever needed.
         config = self._test_yaml_config_prepare_config(
+            yaml_config=yaml_config, usage_stats_event_name=usage_stats_event_name
+        )
+        config_with_substitutions = self._test_yaml_config_prepare_substituted_config(
```
cdkini: I'll try and tackle this in a subsequent PR

great_expectations/datasource/datasource.py
```diff
@@ -196,6 +201,14 @@ def name(self):
     def config(self):
         return copy.deepcopy(self._datasource_config)
 
+    def update_config(self, config: DatasourceConfig) -> None:
+        """
+        See `BaseDatasource.update_config()` - replicated here for legacy parity.
+        """
+        substitution_serializer = DictConfigSerializer(schema=datasourceConfigSchema)
+        raw_config: dict = substitution_serializer.serialize(config)
+        self._datasource_config.update(raw_config)
```
cdkini: What's the best way to do this? Do we perhaps extend the inheritance hierarchy even further so we can place this method in a new base class? `LegacyDatasource` and `BaseDatasource` follow different inheritance hierarchies.

great_expectations/data_context/config_validator/yaml_config_validator.py
```diff
@@ -331,9 +370,12 @@ def _test_yaml_config_prepare_config(
             )
             raise e
 
+    def _load_config_string_as_commented_map(
+        self, config_str: str, usage_stats_event_name: str
+    ) -> CommentedMap:
         try:
-            config: CommentedMap = yaml.load(config_str_with_substituted_variables)  # type: ignore[arg-type]
-            return config
+            substituted_config: CommentedMap = yaml.load(config_str)  # type: ignore[arg-type]
```
Kilo59: `yaml.load` doesn't accept a string??

cdkini: hmm not sure why we have that `type: ignore` - I'll delete and see if it works.

Is there a way to identify outdated `type: ignore` comments?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2858,6 +2858,8 @@ def _instantiate_datasource_from_config_and_update_project_config(
                 datasource = self._instantiate_datasource_from_config(
                     config=substituted_config
                 )
+                # Sanitize instantiated object's config attr to remove any credentials with their ${VARIABLES} counterparts
+                datasource.update_config(config)
```
donaldheppner: it seems that the solution was to call an `update_config` method to restore the original config via serialization? This is strange, and the method name doesn't reflect its behavior.

I'd rather preserve the original configuration and do variable substitution on a property retrieval (ie: substitute on read, and only when the substituted config is required) rather than add extra logic to "undo" something that shouldn't have been done in the first place. Thoughts welcome!

donaldheppner (changes requested):

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: hmm that's a very fair point - let me dig in and see how that could be done

great_expectations/data_context/config_validator/yaml_config_validator.py
Kilo59: Yes, there is. I thought I had enabled it but I was wrong. PR up shortly to add this.
https://mypy.readthedocs.io/en/stable/config_file.html#confval-warn_unused_ignores

#6270

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59: A dedicated `SecretString` type whose `__str__` returns a masked value could help us here.

`pydantic` has something like this for example. But we could implement our own.
https://pydantic-docs.helpmanual.io/usage/types/#secret-types

https://dusted.codes/the-type-system-is-a-programmers-best-friend

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2776,25 +2785,27 @@ def _build_datasource_from_config(self, config: DatasourceConfig) -> Datasource:
         """
         # We convert from the type back to a dictionary for purposes of instantiation
         serializer = DictConfigSerializer(schema=datasourceConfigSchema)
-        config_dict: dict = serializer.serialize(config)
+        raw_config_dict: dict = serializer.serialize(raw_config)
+        substituted_config_dict: dict = serializer.serialize(substituted_config)
+        substituted_config_dict["raw_config"] = raw_config_dict
```
cdkini: Should `DatasourceConfig` and `DatasourceConfigSchema` be updated to reflect these new changes?

great_expectations/datasource/new_datasource.py
```diff
@@ -444,6 +445,7 @@ def __init__(
         self._datasource_config.update(
             {"data_connectors": copy.deepcopy(data_connectors)}
         )
+        self._raw_config = raw_config or self._datasource_config
```
cdkini: If no `raw_config` is provided through the constructor, we leverage the default config.

In all supported workflows through the `DataContext`, we will have a `raw_config`. Only if the user decides to instantiate a standalone `Datasource` will they run into this situation.

donaldheppner (approved):

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: Should `DatasourceConfig` and `DatasourceConfigSchema` be updated to reflect this new config attr? `self.config` is not included in the Marshmallow schema but this is slightly different in that is it a constructor arg?

great_expectations/datasource/new_datasource.py
```diff
@@ -59,14 +61,16 @@ def __init__(
...
-                "name": name,
-            }
         except Exception as e:
             raise ge_exceptions.ExecutionEngineError(message=str(e))
 
+        self._datasource_config: dict = {
+            "execution_engine": execution_engine,
+            "id": id,
+            "name": name,
+        }
```
cdkini, line 69: The `try/except` was for the `ExecutionEngine` creation - no need to put this `self._datasource_config` creation in the same block.

great_expectations/datasource/new_datasource.py
```diff
@@ -59,14 +61,16 @@ def __init__(
...
+        self._raw_config = raw_config or self._datasource_config
```
cdkini: In the case a user instantiates a `Datasource` outside our desired context flow, we need to ensure that their `raw_config` is populated if they want to persist.

Example:
```python
datasource = Datasource(...)
context.save_datasource(datasource)
```

What if I include creds in my naked datasource? Do we allow users to leak values or should be raise warnings?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2730,8 +2737,10 @@ def _init_datasources(self) -> None:
                 config_dict = dict(datasourceConfigSchema.dump(config))
                 datasource_config = datasourceConfigSchema.load(config_dict)
                 datasource_config.name = datasource_name
+                # Our overall config has already undergone variable substituted here so
+                # there is no difference between raw_config and substituted_config arg values
                 datasource = self._instantiate_datasource_from_config(
-                    config=datasource_config
+                    raw_config=datasource_config, substituted_config=datasource_config
```
cdkini: I have a feeling this may need to change - if we look above, we substitute our ENTIRE project config (datasources and all). That means at this point, we don't have access to our raw configs.

This is used to instantiate datasources and populate the cache but might this cause issues down the line? Say I want to update an existing datasource in the cache and persist it; will I possibly leak credentials here?

I think I'll address this in a follow-up PR - the nature of the PR will be ensuring we have both `raw_config` and `substituted_config` here so we avoid problems downstream.

cdkini: Or maybe I do that in this PR idk

great_expectations/datasource/new_datasource.py
anthonyburdi: I thought better here to return an empty config or `None` since there is no raw config in that case. That way `raw_config` is safe from any unwanted substituted values.

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: Good question. I think so. It would have to be optional to not break existing code.

anthonyburdi: Good question. I think so. It would have to be optional to not break existing code.

great_expectations/datasource/new_datasource.py
anthonyburdi: This is a good question. I'm not sure on this one - would be good to raise with the team.

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: just did it here nevermind!

great_expectations/datasource/new_datasource.py
cdkini: X-posting from Slack:

You had noted that if omitted, raw_config should result in self._raw_config = None . This differs from my implementation, which defaults to self._raw_config = self._datasource_config).

Let’s evaluate the following example with both approaches:
```python
datasource = Datasource(...) # Assume the user is passing sensitive creds
context.save_datasource(datasource)
```

By setting self._raw_config = None, the save_datasource method will break (as the method relies on the value of self.raw_config). By setting self._raw_config = self._datasource_config, we allow the user to make the mistake of sending creds over the wire or writing them to disk.

Is there a happy middle ground? I think we need to support the example workflow I’ve noted above since that’s what is used in the Cloud snippet.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2717,21 +2728,31 @@ def _initialize_usage_statistics(
...
+        config: DataContextConfig = self.config
         datasources: Dict[str, DatasourceConfig] = cast(
             Dict[str, DatasourceConfig], config.datasources
         )
 
+        substitutions = self._determine_substitutions()
+
         for datasource_name, datasource_config in datasources.items():
             try:
                 config = copy.deepcopy(datasource_config)  # type: ignore[assignment]
```
Kilo59, line 2739: ## Non-blocking

Why are we ignoring this?

great_expectations/datasource/new_datasource.py
```diff
@@ -29,10 +29,11 @@ class BaseDatasource:
     def __init__(
         self,
         name: str,
-        execution_engine=None,
+        execution_engine: Optional[dict] = None,
```
Kilo59, line 32: ## non-blocking

Does the execution engine have an expected shape?
Can we define a `TypeDict` for it?

Kilo59 (approved):

great_expectations/datasource/new_datasource.py
```diff
@@ -411,11 +423,12 @@ class Datasource(BaseDatasource):
     def __init__(
         self,
         name: str,
-        execution_engine=None,
-        data_connectors=None,
+        execution_engine: Optional[dict] = None,
+        data_connectors: Optional[dict] = None,
```
Kilo59, line 421: ## Non-blocking

Same comment as elsewhere. If there is a predictable structure to these `dict` can we define a `TypeDict` for them.

anthonyburdi (approved):

tests/data_context/abstract_data_context/test_abstract_data_context_datasource_crud.py
```diff
@@ -24,20 +24,21 @@ def __init__(self):
 
 
 class FakeAbstractDataContext(AbstractDataContext):
```
Kilo59, line 26: ## Nit

Not part of the PR but this isn't a "Fake ABC" it's a "fake" concrete class right?
```suggestion
class FakeDataContext(AbstractDataContext):
```

great_expectations/datasource/new_datasource.py
cdkini: as this is pretty high priority, going to defer until later. But good call!

great_expectations/datasource/new_datasource.py
cdkini: Same note!

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: I'm not sure haha let's see if `mypy` complains

anthonyburdi (approved):

PR #6255, opened 2022-10-31
[FEATURE] Add zep datasources to data context.

billdirks:
Updates returned `add_datasource()` registered function to allow the data context to attach the datasource to itself.


great_expectations/zep/context.py
```diff
@@ -2,15 +2,18 @@
...
 
 from great_expectations.util import camel_to_snake
+from great_expectations.zep.interfaces import DataAsset, Datasource
 from great_expectations.zep.type_lookup import TypeLookup
 
 if TYPE_CHECKING:
-    from great_expectations.zep.interfaces import DataAsset, Datasource
+    from great_expectations.data_context.data_context.data_context import (
+        DataContext as GXDataContext,
+    )
```
Kilo59:
```suggestion
    from great_expectations.data_context import DataContext as GXDataContext
```

great_expectations/zep/context.py
```diff
@@ -131,14 +145,18 @@ def get_context(cls) -> DataContext:
         return cls._context
 
     def __init__(self) -> None:
-        self._sources: _SourceFactories = _SourceFactories()
+        self._sources: _SourceFactories = _SourceFactories(self)
+        self._datasources: Dict[str, Datasource] = {}
```
Kilo59, line 148: ### non-blocking
Can we annotate these attributres above `__init__` as we do with the class level attrs?
https://peps.python.org/pep-0526/#class-and-instance-variable-annotations

great_expectations/zep/context.py
```diff
@@ -104,7 +110,15 @@ def factories(self) -> List[str]:
...
-            return self.__source_factories[name]
+            fn = self.__source_factories[name]
+
+            def wrapped(*args, **kwargs):
+                datasource = fn(*args, **kwargs)
+                # TODO (bdirks): add_zep_datasource to the AbstractDataContext class
+                self._data_context.add_zep_datasource(datasource)
+                return datasource
+
+            return wrapped
```
Kilo59: ## non-blocking

At first glance, I misunderstood this.
All these `add_X` methods are confusing me.

Is there another name we could use to clarify that we are attaching the newly created data-source instance to the DataContext??
Ideally not something terrible like... `attach_newly_instantiated_zep_datasource_to_the_context()` 😆 

Also I think it's fine to merge as and figure this out later.

great_expectations/zep/context.py
```diff
@@ -131,14 +145,18 @@ def get_context(cls) -> DataContext:
...
         LOGGER.info(f"4a. Available Factories - {self._sources.factories}")
         LOGGER.info(f"4b. `type_lookup` mapping ->\n{pf(self._sources.type_lookup)}")
 
     @property
     def sources(self) -> _SourceFactories:
         return self._sources
 
+    def add_zep_datasource(self, datasource: Datasource) -> None:
+        self._datasources[datasource.name] = datasource
+
```
Kilo59: Should this be private? Will we call this outside of the `wrapped` attr call?

Kilo59 (approved):

great_expectations/zep/context.py
billdirks: I've changed this to `_attach_datasource_to_context`.

great_expectations/zep/context.py
billdirks: I've made this private. It's defined on a `DataContext` but is getting called on the `_SourceFactories` class.

great_expectations/zep/context.py
billdirks: I'm not sure what you mean here. You mean the class level attribute `_context` (I've annotated that now)? Or something else?

billdirks (commented):
I've address your comments. I didn't completely understand this one:
https://github.com/great-expectations/great_expectations/pull/6255/files#r1009750670

so let me know if I need to do more to address it. Thanks!

great_expectations/zep/context.py
Kilo59: `self._datasources` as well.
```python
class DataContext:
   ...
   _context: DataContext # I don't think this needs a cls level `None` default
   _datasources: Dict[str, Datatsource]

   def __init__(self) -> None:
        self._sources = _SourceFactories(self)
        self._datasources = {}
```
Although I think this style is probably worth discussing as a team.
Coincidentally I think it addressed some of the problems @NathanFarmer discovered in `expectations.py`

billdirks: Done. I did need to initialize `_context` to `None`.

PR #6262, opened 2022-11-01
[MAINTENANCE] `batches.py` typing

Kilo59:
- initial typing of `batches.py` and select dependencies


great_expectations/core/batch.py
```diff
@@ -706,7 +715,7 @@ def get_batch_request_as_dict(
     if batch_request is None:
         return None
 
-    if isinstance(batch_request, (BatchRequest, RuntimeBatchRequest)):
+    if isinstance(batch_request, BatchRequestBase):
```
Kilo59, line 718: This is the only runtime change in the PR.
It seems obvious at first but the docstring for `BatchRequestBase` makes me second guess.

>    This class is for internal inter-object protocol purposes only.
    As such, it contains all attributes of a batch_request, but does not validate them.
    See the BatchRequest class, which extends BatchRequestBase and validates the attributes.

> BatchRequestBase is used for the internal protocol purposes exclusively, not part of API for the developer users.

Is this check an informal safety mechanism to ensure that only "validated" `BatchRequest` or `RuntimeBatchRequest` get used??

Stepping back from this a bit, it seems like maybe `BatchRequestBase` should be renamed and moved to a sibling class of `BatchRequest` and `RuntimeBatchRequest`, and all 3 should inherit from a common ancestor (or follow a `Protocol`).

great_expectations/types/__init__.py
```diff
@@ -244,7 +244,27 @@ def to_json_dict(self) -> dict:
...
+def safe_deep_copy(
+    data: Union[pd.DataFrame, pd.Series], memo=...
+) -> Union[pd.DataFrame, pd.Series]:
+    ...
+
+
+def safe_deep_copy(
+    data: Union[dict, list, tuple, pd.Series, pd.DataFrame],
+    memo: Optional[dict] = None,
+) -> Union[dict, list, pd.Series, pd.DataFrame]:
```
Kilo59, line 267: Generics seem like the obvious choice here, but I think the recursive nature of this function gave me trouble.
However, I didn't spend much on time on it, might want to give this another shot before merging.

great_expectations/core/batch.py
```diff
@@ -223,39 +221,39 @@ def data_asset_name(self, data_asset_name) -> None:
...
+    def runtime_parameters(self) -> Union[dict, None]:
         return self._runtime_parameters
 
     @runtime_parameters.setter
     def runtime_parameters(self, value: dict) -> None:
         self._runtime_parameters = value
 
     @property
-    def batch_identifiers(self) -> dict:
+    def batch_identifiers(self) -> Optional[dict]:
```
Kilo59, line 248: Inconsistent use of `Optional[<TYPE>]` and `Union[<TYPE>, None]` here 😅 
Should update this.

great_expectations/core/batch.py
```diff
@@ -528,18 +526,28 @@ def head(self, *args, **kwargs):
...
 class Batch(SerializableDictDot):
+    _data: BatchDataType
+    _batch_request: Union[BatchRequestBase, dict]
+    _batch_definition: Union[BatchDefinition, IDDict]
+    _batch_spec: BatchSpec
+    _batch_markers: BatchMarkers
+    _data_context: Optional[AbstractDataContext]
+    _datasource_name: Optional[str]
+    _batch_parameters: Any
+    _batch_kwargs = Union[Dict[str, Any], BatchKwargs]
```
Kilo59, line 537: example of attr annotations

Kilo59:
Closing in in favor of
- https://github.com/great-expectations/great_expectations/pull/8024

Kilo59:
Closing in favor of
- https://github.com/great-expectations/great_expectations/pull/8024

PR #6263, opened 2022-11-01
[MAINTENANCE] ZEP - Datasource base class

Kilo59:
- Create a base `Datasource` class which has `MetaDatasource` as the metaclass
- The metaclass is now an implementation detail that is invisible most to users/contributors
- Contributors no longer need to implement simple methods like `get_asset()`
- Move `_SourceFactories` into `sources.py` to prevent circular imports

## Potential Followups
- make Datasource an ABC or Protocol and resolve metaclass conflicts


great_expectations/zep/metadatasource.py
```diff
@@ -35,23 +35,18 @@ def __new__(
         asset_types: List[Type[DataAsset]] = cls_dict.get("asset_types")
         LOGGER.info(f"1b. Extracting Asset details - {asset_types}")
 
-        cls = type(cls_name, bases, cls_dict)
+        cls = super().__new__(meta_cls, cls_name, bases, cls_dict)
         LOGGER.debug(f"  {cls_name} __dict__ ->\n{pf(cls.__dict__, depth=3)}")
```
Kilo59, line 38: check if this is a base `DataSource` and return early or something

great_expectations/zep/interfaces.py
```diff
@@ -14,10 +15,12 @@ def __init__(self, name: str) -> None:
         self.name = name
 
 
-@runtime_checkable
-class Datasource(Protocol):
-    execution_engine: ExecutionEngine
-    asset_types: List[Type[DataAsset]]
+class Datasource(metaclass=MetaDatasource):
+    # class attrs
+    execution_engine: ClassVar[ExecutionEngine]
```
billdirks: Let's keep `execution_engine` an instance variable. Each instance should instantiate its own in general.

great_expectations/zep/sources.py
```diff
@@ -0,0 +1,125 @@
...
+        LOGGER.info(
+            f"2a. Registering {ds_type.__name__} as {simplified_name} with {method_name}() factory"
+        )
+
+        pre_existing = cls.__source_factories.get(method_name)
+        if pre_existing:
+            raise ValueError(f"{simplified_name} factory already exists")
+
+        # TODO: simplify or extract the following datasource & asset type registration logic
+        asset_types = asset_types or []
```
billdirks: `asset_types` is a required list now so we can remove this line.

great_expectations/zep/sources.py
```diff
@@ -0,0 +1,125 @@
...
+
+            def wrapped(*args, **kwargs):
+                datasource = fn(*args, **kwargs)
+                # TODO (bdirks): _attach_datasource_to_context to the AbstractDataContext class
+                self._data_context._attach_datasource_to_context(datasource)
+                return datasource
+
+            return wrapped
+        except KeyError:
+            raise AttributeError(name)
```
billdirks: Maybe print out the allowed keys here. Something like:
`raise AttributeError(f"No factory {name} in {self.factories}")`

billdirks (commented):
Thanks, looks good! A have a few small changes.

billdirks (approved):
LGTM, thanks!

PR #6265, opened 2022-11-01
[MAINTENANCE] Cleanup `expectation.py` attributes

NathanFarmer:
- Add missing attributes
- Loosen constraints on `Tuple` lengths for `domain_keys`, `success_keys`, and `runtime_keys`
- Leverage `Maturity` `Enum` where possible

### Definition of Done


great_expectations/expectations/expectation.py
```diff
@@ -174,12 +175,16 @@ class MetaExpectation(ABCMeta):
     attribute, or, if that is not set, by snake-casing the name of the class.
     """
 
+    default_kwarg_values: Dict[str, Optional[Any]] = {}
```
```suggestion
    default_kwarg_values: Dict[str, Any] = {}
```
Can we narrow down the possible values here?

Maybe `object`?? Although a union type would be even better.
https://mypy.readthedocs.io/en/latest/dynamic_typing.html#any-vs-object

NathanFarmer: I was able to use `object`, thanks for the suggestion!

great_expectations/expectations/expectation.py
```diff
@@ -984,7 +992,7 @@ def get_result_format(
         configuration: ExpectationConfiguration,
         runtime_configuration: dict = None,
     ) -> Union[Dict[str, Union[str, int, bool]], str]:
-        default_result_format: Optional[str] = self.default_kwarg_values.get(
+        default_result_format: Optional[Any] = self.default_kwarg_values.get(
```
NathanFarmer, line 995: Constraining `default_kwarg_values` from `Any` to `object` meant that this change was required. After reading `mypy` documentation, I'm not sure I understand why though.

great_expectations/expectations/expectation.py
NathanFarmer: See my comment on line 995 about the downstream effects.

great_expectations/expectations/expectation.py
Kilo59: I'd change it back. I actually don't have much experience using `object` as a type.

Could also define a `TypeDict` for `default_kwarg_values`, but if this is a point of extensibility then this may not be desirable.

Kilo59 (approved):

great_expectations/expectations/expectation.py
```diff
@@ -227,20 +232,23 @@ class Expectation(metaclass=MetaExpectation):
     """
 
     version = ge_version
-    domain_keys: Union[tuple, Tuple[str]] = ()
-    success_keys: Union[tuple, Tuple[str]] = ()
-    runtime_keys: Union[tuple, Tuple[str]] = (
+    domain_keys: Tuple[str, ...] = ()
+    success_keys: Tuple[str, ...] = ()
+    runtime_keys: Tuple[str, ...] = (
```
cdkini, line 237: ❤️

cdkini (approved):

great_expectations/expectations/expectation.py
```diff
@@ -174,12 +175,16 @@ class MetaExpectation(ABCMeta):
     attribute, or, if that is not set, by snake-casing the name of the class.
     """
 
+    default_kwarg_values: Dict[str, object] = {}
```
Kilo59, line 178:
```suggestion
    default_kwarg_values: Dict[str, Union[str, bool, None]] = {}
```

???

great_expectations/expectations/expectation.py
NathanFarmer: Actually I take that back. This happened due to my removing of the type override in the `Expectation` class (the constraint was not appropriate for all subclasses). I think I will leave this like it is.

great_expectations/expectations/expectation.py
NathanFarmer: Could also be `int`, `float`, or anything really. Think of all of the possibilities for Expectation parameter defaults.

PR #6270, opened 2022-11-02
[MAINTENANCE] mypy `warn_unused_ignores`

Kilo59:
- ~update `mypy` config setting `warn_unused_ignores = true`~
  - https://mypy.readthedocs.io/en/stable/config_file.html#confval-warn_unused_ignores
- pass `--warn-unused-ignores` as CLI arg in CI
- remove unused ignores
- update `scope_change` rules to consider changes to `setup.cfg` and `pyproject.toml`

I've opted to pass `--warn-unused-ignores` as a CLI argument for our type-checking step instead of making the change in the config. This is because devs could get different `unused-ignore` errors based on their version of python and the installed dependencies and lead to a fair amount of churn where they remove ignores they are seeing locally, only to add them back when the CI complains. Open to any objections, there are some trade-offs to consider here.

As follow-up PR I plan to add a `invoke type-check --ci` flag that devs can use if they want to always ensure they are running the type-checker with the exact parameters as the CI step.

### DIFF
--- a/azure-pipelines-dev.yml
+++ b/azure-pipelines-dev.yml
@@ -121,7 +121,7 @@ stages:
       steps:
       - script: |
           pip install mypy invoke
-          invoke type-check --install-types --warn-unused-ignore
+          invoke type-check --install-types --warn-unused-ignore --pretty
         name: StaticTypeCheck
 
     - job: docstring_checker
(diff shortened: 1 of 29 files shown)


azure-pipelines-dev.yml
```diff
@@ -74,6 +74,8 @@ stages:
 
                 [GEChanged]
                 great_expectations/**/*.py
+                pyproject.toml
+                setup.cfg
```
Kilo59, line 78: Changed the scoping rules here and in 2 other pipelines.

cdkini (approved):

PR #6274, opened 2022-11-02
Add ZEP postgres datasource.

billdirks:
This adds the ZEP postgres datasource. I need to add tests and fix one current test that assumes we don't have any datasources, however, I wanted to get comments to make sure people agree with this approach.


great_expectations/zep/postgres_datasource.py
```diff
@@ -0,0 +1,141 @@
...
+            column_name=column_name,
+            namespace=namespace,
+            template_params=["year", "month"],
+        )
+        return self
+
+
+class PostgresDatasource(Datasource):
+    asset_types = [TableAsset]
+
```
billdirks, line 124: I can add annotations for member variables.

great_expectations/zep/postgres_datasource.py
```diff
@@ -0,0 +1,141 @@
...
+@dataclasses.dataclass(frozen=True)
+class ColumnSplitter:
+    method_name: str
+    column_name: str
+    namespace: str
+    template_params: List[str]
+
+
+class TableAsset(DataAsset):
+    def __init__(
```
billdirks, line 39: I can add annotations for member variables.

great_expectations/zep/interfaces.py
```diff
@@ -43,3 +62,53 @@ def get_asset(self, asset_name: str) -> DataAsset:
         """Returns the DataAsset referred to by name"""
         # This default implementation will be used if protocol is inherited
         return self.assets[asset_name]
+
+
+class Batch:
+    def __init__(
```
billdirks, line 87: I can add annotations for member variables.

great_expectations/zep/interfaces.py
```diff
@@ -1,18 +1,37 @@
...
+@dataclasses.dataclass(frozen=True)
+class BatchRequest:
+    datasource_name: str
+    data_asset_name: str
+    options: BatchRequestOptions
+
 
 class DataAsset(abc.ABC):
-    name: str
+    def __init__(self, name: str):
```
billdirks: When I merged I lost the annotation for `name`. I will add it back.

NathanFarmer: Why remove the return type?

great_expectations/zep/interfaces.py
```diff
@@ -25,7 +44,7 @@ class Datasource(metaclass=MetaDatasource):
     assets: Dict[str, DataAsset]
 
     def get_batch_list_from_batch_request(
-        self, batch_request: Union[BatchRequest, RuntimeBatchRequest]
+        self, batch_request: BatchRequest
```
NathanFarmer, line 59: Eventually we will need to support `RuntimeBatchRequest` here though I would think.

great_expectations/zep/interfaces.py
```diff
@@ -43,3 +62,53 @@ def get_asset(self, asset_name: str) -> DataAsset:
...
+
+class Batch:
+    def __init__(
+        self,
+        datasource: Datasource,
+        data_asset: DataAsset,
+        batch_request: BatchRequest,
+        # BatchDataType is the current type of data in a pre-zep Batch. We might need to define an explicit
+        # interface if we think Datasource implementers will use this as a point of extension.
+        data: BatchDataType,
```
NathanFarmer, line 94: What is an example of a "type of data"?

great_expectations/zep/postgres_datasource.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # We look at the splitters on the data asset and verify that the passed in batch request provides the
+        # correct arguments to specify the batch
+        batch_spec_kwargs = {
+            "type": "table",
+            "data_asset_name": data_asset.name,
+            "table_name": data_asset.table_name,
+            "batch_identifiers": {},
+        }
+        if column_splitter := data_asset._column_splitter:
```
NathanFarmer: Is this allowed in python 3.7?

great_expectations/zep/postgres_datasource.py
```diff
@@ -0,0 +1,141 @@
...
+        # NOTE: We only produce 1 batch right now
+        data_asset = self.get_asset(batch_request.data_asset_name)
+
+        # We look at the splitters on the data asset and verify that the passed in batch request provides the
+        # correct arguments to specify the batch
+        batch_spec_kwargs = {
+            "type": "table",
+            "data_asset_name": data_asset.name,
+            "table_name": data_asset.table_name,
+            "batch_identifiers": {},
```
NathanFarmer: Will we eventually use `BatchMetadata` to populate `batch_identifiers`?

NathanFarmer (commented):
A few comments from me!

great_expectations/zep/interfaces.py
```diff
@@ -43,3 +62,53 @@ def get_asset(self, asset_name: str) -> DataAsset:
...
+        """This represents a batch of data.
+
+        This is usually not the data itself but a hook to the data on an external datastore such as
+        a spark or a sql database. An exception exists for pandas or any in-memory datastore.
+        """
+        # These properties are intended to be READ-ONLY
+        self._datasource: Datasource = datasource
+        self._data_asset: DataAsset = data_asset
+        self._batch_request: BatchRequest = batch_request
+        self._data: BatchDataType = data
```
Kilo59, line 105: ~Can we annotate these on the class?~ Just saw that you already mentioned that 😄 

Also, should we use double underscore name mangling for these? Maybe not, I haven't seen this much in our codebase but thought I'd throw it out there.

```python
class Batch:
       """"""
        id: str
        datasource: Datasource
        data_asset: DataAsset
        batch_request: BatchRequest
        data: BatchDataType

        def __init__(self, ...):
                self.__datasource = ...
                self.__data_asset = ...
                self.__batch_request = ...
                self.__data = ...
```

great_expectations/zep/postgres_datasource.py
Kilo59: No, but given that this is going to be a brand new workflow, I don't think it's unreasonable to require python 3.8.
It won't break any existing uses.

great_expectations/zep/postgres_datasource.py
```diff
@@ -0,0 +1,141 @@
...
+                batch_spec_kwargs["batch_identifiers"].update(
+                    {column_splitter.column_name: column_splitter_kwargs}
+                )
+            except KeyError as e:
+                params_dict = {p: "<value>" for p in column_splitter.template_params}
+                if ns:
+                    params_dict = {ns: {params_dict}}
+                raise PostgresDatasourceError(
+                    "\nYou must specify the batch request splitter options as a dictionary that looks like "
+                    f"this: {params_dict}\nYou've specified {batch_request.options}"
```
Kilo59: ## Non-blocking

You might want to play around with this, but we could also use `pprint.pformat` here.
It takes the same arguments as `pprint` except it returns a formated string instead of printing to the console.

```python

...

raise PostgresDatasourceError(
  "\nYou must specify the batch request splitter options as a dictionary that looks like "
  f"this: {pf(params_dict)}\nYou've specified {pf(batch_request.options)}"
) from e
```

great_expectations/zep/postgres_datasource.py
```diff
@@ -0,0 +1,141 @@
...
+    template_params: List[str]
+
+
+class TableAsset(DataAsset):
+    def __init__(
+        self,
+        datasource: Datasource,
+        name: str,
+        table_name: str,
+        # TODO(bdirks): Provide a mechanism to add splitters in __init__
```
Kilo59: Is there more to it than accepting a `column_splitter: Optional[ColumnSplitter] = None` kwarg?

or is the difficulty dealing with multiple column splitters (🤷 ) and the logic in `get_batch_list_from_batch_request()`?

great_expectations/zep/postgres_datasource.py
```diff
@@ -0,0 +1,141 @@
...
+    Datasource,
+)
+
+
+class PostgresDatasourceError(Exception):
+    pass
+
+
+@dataclasses.dataclass(frozen=True)
+class ColumnSplitter:
```
Kilo59, line 25: 🎊 For immutable dataclasses

great_expectations/zep/interfaces.py
billdirks: Thanks, that was an oversight. I've added it back and added return types to all functions in this PR.

great_expectations/zep/interfaces.py
billdirks: Currently this just takes a ZEP `BatchRequest` and not the current GX `BatchRequest` or `RuntimeBatchRequest`. We will revisit if we need a ZEP runtime version later as we add more features.

great_expectations/zep/interfaces.py
billdirks: Currently it is `BatchDataType = Union[BatchData, pd.DataFrame, SparkDataFrame]` and is defined here:
https://github.com/great-expectations/great_expectations/blob/53ac04c3bacd0e860cbdcec6edf4e843e5e50dc6/great_expectations/core/batch.py#L522
The I will update the comment to be more explicit about the interface.

great_expectations/zep/interfaces.py
billdirks: I'd like to avoid name-mangling unless you feel super strongly about this.

great_expectations/zep/postgres_datasource.py
billdirks: I will do this. Thanks!

great_expectations/zep/postgres_datasource.py
billdirks: The instance variables are declared in the superclass so I've decided not to redeclare them here.

great_expectations/zep/postgres_datasource.py
billdirks: While I'd love to move us to more modern versions of python, I want to disentangle this code from that discussion. I've removed the walrus.

great_expectations/zep/postgres_datasource.py
billdirks: I'm not sure the answer to this because I don't have a good understanding of the use cases for `BatchMetaData`. I don't want to introduce an unstructured dict unless we need to. If we do introduce it, I think it will be an attribute or property on a`Batch` object. It might be a property gets composed at read time from more structured backing attributes.

great_expectations/zep/postgres_datasource.py
billdirks: That is better. Done!

great_expectations/zep/interfaces.py
```diff
@@ -22,10 +41,10 @@ class Datasource(metaclass=MetaDatasource):
     # instance attrs
     name: str
     execution_engine: ExecutionEngine
-    assets: Dict[str, DataAsset]
+    assets: Mapping[str, DataAsset]
```
billdirks, line 56: I've changed this to a `Mapping` since I want subclasses to specify specific types of `DataAsset` subclasses. That is, the postgres datasource overwrites this as `assets: Mapping[str, TableAsset]`. `Dicts` are invariant types so don't allow this subclassing of the values.

Kilo59: No typing errors here? 
Originally I had this defined as a Mapping but the Mapping interface doesn't include __setitem__ so it complained when I tried to set the assets.

Hence is my initial PR had a MutableMapping.

But if this works, there's no issue.

Kilo59 (approved):

great_expectations/zep/interfaces.py
NathanFarmer: It maps to `ExecutionEngine`s 1:1 right now

billdirks: Oh, that's good to know.

PR #6305, opened 2022-11-05
[MAINTENANCE] Delete "metric_dependencies" from MetricConfiguration constructor arguments

alexsherstinsky:
### Scope
It is architecturally wrong to include `metric_dependencies` among `MetricConfiguration` constructor arguments, because this property is never actually needed to instantiate `MetricConfiguration`; it is only supplied via a setter property at runtime.  The fact that having `metric_dependencies` as part of `MetricConfiguration` makes developing tests easier is outweighed by the fact that it is never used in realistic scenarios.  Hence, it is removed, and setters are used instead.

It is important to emphasize that the assignment of `metric_dependencies` is used after these dependencies have been computed, which happens in the body of the graph construction code; setting `metric_dependencies` through the constructor of `MetricConfiguration` is NEVER used (cannot possibly be, because dependencies are computed much later).

- JIRA: GREAT-1103/GREAT-1344
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/validator/metric_configuration.py
```diff
@@ -33,32 +38,34 @@ def __str__(self):
...
+    def metric_dependencies(self) -> Dict[str, MetricConfiguration]:
+        return self._metric_dependencies or {}
 
     @metric_dependencies.setter
-    def metric_dependencies(self, metric_dependencies) -> None:
-        self._metric_dependencies = metric_dependencies
+    def metric_dependencies(
+        self, value: Optional[Dict[str, MetricConfiguration]]
+    ) -> None:
+        self._metric_dependencies = value
```
Kilo59: This point is totally orthogonal to the main point of this PR but I thought I'd bring it up since I've noticed this pattern a lot in our codebase.

The pattern is where we use a `@property` with a setter but we don't actually protect or calculate anything.
I think it just adds noise to the code.

Consider the 2 equivalent class definitions for the class `Square` which takes a `length` and `color`.
Both behave exactly the same but one is almost `20` lines while the other is only `4`.

```python
class Square:
    def __init__(self, length: int, color: str):
        self._length = length
        self._color = color

    @property
    def length(self) -> int:
        return self._length

    @length.setter
    def length(self, length: int):
        self._length = length

    @property
    def color(self) -> str:
        return self.color

    @color.setter
    def color(self, color: str):
        self._color = color

# the class below is functionally equivalent 

class Square:
    def __init__(self, length: int, color: str):
        self.length = length
        self.color = color
```

To be clear, this isn't an anti-property take. They are [very useful and powerful](https://pabloariasal.github.io/2018/11/25/python-descriptors/), but I think we are overusing in many situations.

Below are examples where IMO they do make sense.
1. We want to prevent modifying an attribute (`length`)
2. We want to apply some logic during getting or settings
  a. Like computing a value like `area` from length.
  b. Applying validation during setting `color`.

```python
class Square:
    def __init__(self, length: int, color: str):
        self.length = length
        self._color = color

    @property
    def area(self) -> int:
        return self.length * self.length

    @property
    def color(self) -> str:
        return self.color

    @color.setter
    def color(self, color: str):
        if color not in ["red", "blue", "yellow"]:
            raise ValueError
        self._color = color
```

The above assumes we aren't doing any metaprogramming with `properties`. Like specifically checking objects for `property` types and doing something based on that. 😄 🤷

tests/execution_engine/test_execution_engine.py
```diff
@@ -187,60 +187,66 @@ def test_resolve_metrics_with_aggregates_and_column_map():
         metric_name="column.mean",
         metric_domain_kwargs={"column": "a"},
         metric_value_kwargs=None,
-        metric_dependencies={
-            "table.columns": table_columns_metric,
-        },
     )
+    mean.metric_dependencies = {
```
donaldheppner, line 191: If we are constantly having to initialize `metric_dependencies` after the call to `init`, then it really should be part of `init`.

donaldheppner (changes requested):

tests/execution_engine/test_execution_engine.py
alexsherstinsky: @donaldheppner No, because they are NEVER known at time of `init` -- in all uses of `MetricConfiguration` as a constructor, the `metric_dependencies` is either omitted or set to `None` — having it there was left over from the original implementation a couple of years ago, and it is ALWAYS `None` (because there is no knowledge of what the dependencies should be at the init/constructor time).

Kilo59: >No, because they are NEVER known at time of init

@alexsherstinsky but is that just incidental to how we happen to be using it currently? Or will that always be true?

great_expectations/validator/metric_configuration.py
alexsherstinsky: @Kilo59 I agree with you that properties should be used with good cause and judiciously (especially the ones including the `setter` capability).  Thanks.

tests/execution_engine/test_execution_engine.py
alexsherstinsky: @Kilo59 It has always been true and will continue being so on the account of where/how dependencies get computed and become known (not at `__init__()` stage).  Thanks.

great_expectations/validator/metric_configuration.py
alexsherstinsky: @Kilo59 By the way, I have removed the `setter` from `MetricConfiguration.metric_dependencies()` property (per @donaldheppner 's observation).  Thanks.

great_expectations/validator/metric_configuration.py
```diff
@@ -11,20 +13,23 @@ def __init__(
...
+
             metric_value_kwargs = IDDict(metric_value_kwargs)
-        self._metric_value_kwargs = metric_value_kwargs
-        if metric_dependencies is None:
-            metric_dependencies = {}
-        self._metric_dependencies = metric_dependencies
+
+        self._metric_value_kwargs: IDDict = metric_value_kwargs
+
+        self._metric_dependencies: Optional[Dict[str, MetricConfiguration]] = None
```
Kilo59:
```suggestion
        self.metric_dependencies: Dict[str, MetricConfiguration] = {}
```

Can we get rid of the `metric_dependencies` property?

alexsherstinsky: @Kilo59 If we do it the way you are suggesting, then I think yes.  Do you think this is worth it (i.e., setting it to `dict()` instead of `None`)?  Thanks.

donaldheppner (approved):

PR #6318, opened 2022-11-07
Fallback to isnot method if is_not is not available (old sqlalchemy)

kenwade4:
- Fallback to isnot method if is_not is not available (old sqlalchemy)

### DIFF
--- a/great_expectations/dataset/sqlalchemy_dataset.py
+++ b/great_expectations/dataset/sqlalchemy_dataset.py
@@ -2358,7 +2358,7 @@ WHERE
         if len(parsed_value_set) == 0:
             return False
 
-        try:
+        if hasattr(sa.column(column_A), "is_not"):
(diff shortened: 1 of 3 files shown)


great_expectations/dataset/sqlalchemy_dataset.py
```diff
@@ -2358,19 +2358,34 @@ def expect_column_pair_values_to_be_in_set(
         if len(parsed_value_set) == 0:
             return False
 
-        conditions: List[BinaryExpression] = [
-            sa.and_(
+        try:
```
anthonyburdi: Instead of a try except could we use a conditional around a hasattr check? (similar to the other files in this PR)

anthonyburdi (approved):
Looks good, just a non blocking suggestion.

anthonyburdi (approved):
Thank you for making that change!

PR #6320, opened 2022-11-07
[FEATURE] POC ZEP Config Loading

Kilo59:

- Allow for deserialization of ZEP `Datasource`s from a new config format.
  - Makes Datasource & Dataset base classes `pydantic` models.
  - Add custom validation hooks to allow for loading the correct `Datasource` + `DataAsset` subtype according to the registered `type` value.
  - remove all "simplified name" generation logic, instead pull the "simple name" from the model's `.type` field.
- Updates to `TypeLookup` to allow for an atomic type registration "transaction" that only commits changes when no exceptions occur.
- Raise specific error messages when `Datasource`/`DataAsset` classes are improperly defined.
- add `pydantic` dependency - https://pydantic-docs.helpmanual.io/


great_expectations/zep/alt_config.yaml
```diff
@@ -0,0 +1,12 @@
+datasources:
```
Kilo59: remove this file

great_expectations/zep/config.yaml
```diff
@@ -0,0 +1,31 @@
+datasources:
```
Kilo59: remove this file

great_expectations/zep/logger.py
```diff
@@ -0,0 +1,30 @@
+import logging
+import sys
+
+
+def init_logger(level: int = logging.WARNING):
+    print("setting up logger ...")
```
Kilo59: remove

great_expectations/zep/type_lookup.py
```diff
@@ -1,18 +1,43 @@
...
+from typing_extensions import TypeAlias
+
+LOGGER = logging.getLogger(__name__)
 
 
 class TypeLookupError(ValueError):
     pass
 
 
+ValidTypes: TypeAlias = Union[Hashable, Type]
```
Kilo59: Should update this to be less generic.
The value is either a `str` or a `Type`

great_expectations/zep/postgres_datasource.py
```diff
@@ -139,18 +122,16 @@ def add_table_asset(self, name: str, table_name: str) -> TableAsset:
...
         return asset
 
     def get_asset(self, asset_name: str) -> TableAsset:
         """Returns the TableAsset referred to by name"""
-        try:
-            return self.assets[asset_name]
-        except KeyError as e:
-            raise PostgresDatasourceError(
-                f"No table asset named {asset_name}. Available assets are {self.assets.keys()}"
-            ) from e
```
Kilo59, line 266: @billdirks 
I moved this logic to the super() class `.get_asset` but if you think it would be better to raise a specific `PostgresDatasourceError` here I can catch and re-raise as this error here.

great_expectations/zep/alt_config.yaml
billdirks: This and the file below are just here to demo the functionality and aren't meant to be checked in?

great_expectations/zep/config.py
```diff
@@ -0,0 +1,58 @@
...
+LOGGER = logging.getLogger(__name__.lstrip("great_expectations."))
+
+
+class GxConfig(BaseModel):
+    datasources: Dict[str, Datasource]
+
+    @classmethod
+    def parse_yaml(cls, f: Union[pathlib.Path, str], debug_: bool = False) -> GxConfig:
+        loaded = yaml.load(f)
+        LOGGER.debug(f"loaded from yaml ->\n{pf(loaded, depth=3)}\n")
```
billdirks, line 27: It looks like we log to 2 places since we are using `LOGGER.debug` and `debug_`. Can we use 1 mechanism for this?

great_expectations/zep/config.py
```diff
@@ -0,0 +1,58 @@
...
+
+try:
+    from devtools import debug as pp
+except ImportError:
+    from pprint import pprint as pp  # type: ignore[assignment]
+
+yaml = YAML(typ="safe")
+
+
+LOGGER = logging.getLogger(__name__.lstrip("great_expectations."))
```
billdirks: It seems like elsewhere we do `logging.getLogger(__name__)`. This might also be better for people importing us. Do you mind using the namespaced logger name?

great_expectations/zep/config.py
```diff
@@ -0,0 +1,58 @@
...
+        loaded = yaml.load(f)
+        LOGGER.debug(f"loaded from yaml ->\n{pf(loaded, depth=3)}\n")
+        config = cls(**loaded)
+        if debug_:
+            pp(config)
+        return config
+
+    @validator("datasources", pre=True)
+    @classmethod
+    def _load_datasource_subtype(cls, v: Dict[str, dict]):
```
billdirks, line 33: Where does this get called?

great_expectations/zep/context.py
```diff
@@ -1,15 +1,19 @@
...
+from pydantic import DirectoryPath, validate_arguments
+
+from great_expectations.zep.config import GxConfig
 from great_expectations.zep.sources import _SourceFactories
 
 if TYPE_CHECKING:
     from great_expectations.zep.interfaces import Datasource
 
-LOGGER = logging.getLogger(__name__)
+LOGGER = logging.getLogger(__name__.lstrip("great_expectations."))
```
billdirks: Similar comment about logger name here.

great_expectations/zep/context.py
```diff
@@ -21,20 +25,38 @@ class DataContext:
     """
 
     _context: ClassVar[Optional[DataContext]] = None
+
     _datasources: Dict[str, Datasource]
+    root_directory: Union[DirectoryPath, str, None]
 
     @classmethod
-    def get_context(cls) -> DataContext:
+    def get_context(
```
billdirks, line 33: Will this file go away once we integrate with the "real" `DataContext`?

great_expectations/zep/demos.py
```diff
@@ -1,10 +1,27 @@
-import logging
-from typing import Dict, List, Optional
+import pathlib
```
billdirks, line 3: Just wondering about future of this file. Will it go away once we have a repo where we demo this functionality?

great_expectations/zep/postgres_datasource.py
```diff
@@ -34,7 +35,7 @@ class BatchRequestError(Exception):
 _DEFAULT_MONTH_RANGE = range(1, 13)
 
 
-@dataclasses.dataclass(frozen=True)
+@pydantic_dc.dataclass(frozen=True)
```
Kilo59, line 39: This is a drop-in replacement for a `dataclass` but with type enforcement.
Also needed in order to specify these splitters as part of config yaml files.

great_expectations/zep/sources.py
```diff
@@ -3,33 +3,34 @@
...
     from great_expectations.data_context import DataContext as GXDataContext
+    from great_expectations.execution_engine import ExecutionEngine
     from great_expectations.zep.context import DataContext
     from great_expectations.zep.interfaces import DataAsset, Datasource
 
 SourceFactoryFn = Callable[..., "Datasource"]
 
-LOGGER = logging.getLogger(__name__)
-
+LOGGER = logging.getLogger(__name__.lstrip("great_expectations."))
```
Kilo59: TODO: revert this logger name shortening (here and elsewhere). Was looking at logs a lot and this was helpful, but shouldn't be part of the merged code.

great_expectations/zep/logger.py
```diff
@@ -0,0 +1,30 @@
...
+    try:
+        from loguru import logger
```
Kilo59, line 7: @billdirks would you like me to leave this in here or remove it?

I was going to remove it on the PR and just modify it locally, but if you also find this useful I can leave it in.

great_expectations/zep/interfaces.py
```diff
@@ -20,36 +25,106 @@
...
+        }
+        LOGGER.debug(f"{engine_type} - kwargs: {list(engine_kwargs.keys())}")
+        engine = engine_type(**engine_kwargs)
+        values["execution_engine"] = engine
+        LOGGER.info(f"{registered_ds_type_name} - {engine_type.__name__} - {engine}")
+        return values
+
+    @pydantic.validator("assets", pre=True)
+    @classmethod
+    def _load_asset_subtype(cls, v: Dict[str, dict]):
```
billdirks, line 115: I'm interested to see what having separate runtime vs config models looks like. It would be nice if someone developing against zep could just open this file and understand what the interfaces are. While not a huge amount of code, they will have to sift between which methods are used for functionality vs serialization. 
Would the private methods `_load_execution_engine` and `_load_asset_subtype` go away in the 2 model world?

great_expectations/zep/logger.py
billdirks: I haven't play around with it yet though I am interested. It seems like we want to remove demos.py and put is somewhere else. We can leave this for now and migrate it when we move that file.

great_expectations/zep/metadatasource.py
```diff
@@ -27,30 +32,31 @@ def __new__(
...
+            LOGGER.info(f"5. Adding '{name}' {cls_name}")
+            return cls(name=name, **kwargs)
 
         # TODO: generate schemas from `cls` if needed
 
-        _SourceFactories.register_factory(
-            cls, _datasource_factory, asset_types=asset_types
-        )
+        if cls.__module__ == "__main__":
+            LOGGER.warning(
```
billdirks, line 57: Should we actually skip this registration in this case?

great_expectations/zep/my_file.csv
```diff
@@ -0,0 +1,3 @@
+first,last
```
billdirks: Are we moving things like this to a demo repo and out of here in the future?

great_expectations/zep/postgres_datasource.py
```diff
@@ -49,32 +50,11 @@ def param_names(self) -> List[str]:
...
-
-    @property
-    def name(self) -> str:
-        return self._name
-
-    @property
-    def datasource(self) -> Datasource:
-        return self._datasource
+    column_splitter: Optional[ColumnSplitter] = None
+    name: str
```
billdirks, line 58: Does this mean `name` is now mutable? That is ok, just curious if there is a way to keep it immutable.

great_expectations/zep/postgres_datasource.py
billdirks: Thanks!

great_expectations/zep/postgres_datasource.py
```diff
@@ -252,18 +225,16 @@ def add_table_asset(self, name: str, table_name: str) -> TableAsset:
...
 
     def get_asset(self, asset_name: str) -> TableAsset:
         """Returns the TableAsset referred to by name"""
-        try:
-            return self.assets[asset_name]
-        except KeyError as e:
-            raise PostgresDatasourceError(
-                f"No table asset named {asset_name}. Available assets are {self.assets.keys()}"
-            ) from e
+        return super().get_asset(asset_name)  # type: ignore[return-value] # value is subclass
```
billdirks, line 238: It would be great to remove this method altogether but we want the more specific return type. Is the `type: ignore[return-value]` because of your editor or because of `mypy`. I'm curious if we should cast the return value.

great_expectations/zep/sources.py
billdirks: Ah, I left a comment above and now am seeing this one.

great_expectations/zep/sources.py
```diff
@@ -3,33 +3,34 @@
...
-    if suffix_to_remove:
-        return _remove_suffix(result, suffix_to_remove)
-    return result
+    def __init__(
+        self,
+        message: Union[str, Exception],
+        type_: Optional[Type] = None,
+        field_name: Optional[str] = None,
+    ) -> None:
+        self.type = type_
```
billdirks: Do you use `type_` and `field_name` later? I don't see where. Do they appear in an error message?

great_expectations/zep/sources.py
```diff
@@ -40,84 +41,180 @@ class _SourceFactories:
...
-        LOGGER.debug(f"'{simplified_name}' added to `type_lookup`")
-        cls.__source_factories[method_name] = fn
+    @classmethod
+    def _register_assets(cls, ds_type: Type[Datasource], asset_type_lookup: TypeLookup):
+        errored_on: Optional[Type[DataAsset]] = None
+        try:
+            asset_types: List[Type[DataAsset]] = ds_type.asset_types
+
+            if not asset_types:
+                LOGGER.warning(
```
billdirks: Should we log at the error level here? Should we throw an exception?

great_expectations/zep/sources.py
```diff
@@ -40,84 +41,180 @@ class _SourceFactories:
...
+                    asset_type_lookup[t] = asset_type_name
+                except (AttributeError, KeyError, TypeError) as bad_field_exc:
+                    errored_on = t
+                    LOGGER.warning(
+                        f"{bad_field_exc.__class__.__name__}:{bad_field_exc}"
+                    )
+                    raise TypeError(
+                        f"No `type` field found for `{ds_type.__name__}.asset_types` -> `{t.__name__}` unable to register asset type"
+                    ) from bad_field_exc
+        except TypeError as exc:
```
billdirks: I don' t know if we need to nest catching exceptions here. Can we remove this outer catch block and raise this `TypeRegistrationError` in the inner block?

great_expectations/zep/sources.py
```diff
@@ -40,84 +41,180 @@ class _SourceFactories:
...
+                        raise TypeError(
+                            f"{t.__name__} `type` field must be assigned and cannot be `None`"
+                        )
+                    LOGGER.info(
+                        f"2b. Registering `DataAsset` `{t.__name__}` as {asset_type_name}"
+                    )
+                    asset_type_lookup[t] = asset_type_name
+                except (AttributeError, KeyError, TypeError) as bad_field_exc:
+                    errored_on = t
+                    LOGGER.warning(
```
billdirks: This logging message should have more context. Also, maybe we don't need it since we immediately raise afterwards. Can we include this information in the raised error?

great_expectations/zep/type_lookup.py
billdirks: Add this as a TODO in the code so we don't lose it?

great_expectations/zep/type_lookup.py
```diff
@@ -22,24 +47,43 @@ class TypeLookup(
...
+    def __delitem__(self, key: ValidTypes):
         value = self.data.pop(key)
         super().pop(value, None)
+        self._original_keys.remove(key)
 
-    def __setitem__(self, key: Hashable, value: Hashable):
+    def __setitem__(self, key: ValidTypes, value: ValidTypes):
+        if key is None:
+            raise TypeLookupError(
+                f"`{type(key).__name__}` for {value} is not allowed - bad key"
```
billdirks: `{type(key).__name__}` will evaluate to `NoneType` here. Should we use that string directly? We can then make this a normal string vs an f-string.

great_expectations/zep/type_lookup.py
```diff
@@ -22,24 +47,43 @@ class TypeLookup(
...
+            )
+        if value is None:
+            raise TypeLookupError(
+                f"`{type(value).__name__}` for {key} is not allowed - bad value"
```
billdirks: Similar comment to above.

great_expectations/zep/type_lookup.py
```diff
@@ -49,5 +93,85 @@ def __repr__(self) -> str:
...
+        txn_exc: Union[Exception, None] = None
+
+        txn_lookup_copy = self.__class__()
+        txn_lookup_copy.data = copy.copy(self.data)
+        LOGGER.info("Beginning TypeLookup transaction")
+        try:
+            yield txn_lookup_copy
+        except Exception as exc:
+            txn_exc = exc
+            LOGGER.info(
```
billdirks: I don't think we should log and raise since we are hitting the user with the error 2x. We can just raise where without logging. If we want to log it should be done on the error level.

great_expectations/zep/sources.py
```diff
@@ -40,84 +41,180 @@ class _SourceFactories:
...
+        ds_type_name = ds_type.__fields__["type"].default
+        if not ds_type_name:
+            raise TypeRegistrationError(
+                f"`{ds_type.__name__}` is missing a `type` attribute with an assigned string value"
+            )
+
+        # rollback type registrations if exception occurs
+        with cls.type_lookup.transaction() as type_lookup:
+
+            # TODO: We should namespace the asset type to the datasource so different datasources can reuse asset types.
```
billdirks, line 78: namespacing would also let us reuse names for assets and datasources. That is, I don't think we can currently do `MyDataSource` and `MyAsset` since the simplified name for both is `my`?

great_expectations/zep/sources.py
```diff
@@ -40,84 +41,180 @@ class _SourceFactories:
...
-        method_name = f"add_{simplified_name}"
+        # TODO: check that the name is a valid python identifier (and maybe that it is snake_case?)
+        ds_type_name = ds_type.__fields__["type"].default
+        if not ds_type_name:
+            raise TypeRegistrationError(
+                f"`{ds_type.__name__}` is missing a `type` attribute with an assigned string value"
+            )
+
+        # rollback type registrations if exception occurs
+        with cls.type_lookup.transaction() as type_lookup:
```
billdirks, line 76: This transaction is order dependent and this should be called out in the comments. That is, we must have `cls._register_engine` after the other register methods since if it was first and succeed but registering assets or datasources failed, we wouldn't rollback the engine registration.

great_expectations/zep/type_lookup.py
```diff
@@ -49,5 +93,85 @@ def __repr__(self) -> str:
...
+                f"{exc.__class__.__name__}:{exc} - encountered during TypeLookup transaction"
+            )
+            raise
+        finally:
+            net_new_items = {
```
billdirks: Can we replace `self.data` with `txn_lookup_copy.data` instead of computing `net_new_items` and doing an update?  It seems a bit simpler since we've already constructed the data structure we want.

If we want to report the lengths of the items rolledback in the error message we could do `len(txn_lookup_copy.data) - len(initial_keys)` though I'm not sure if we need to log this.

great_expectations/zep/type_lookup.py
```diff
@@ -22,24 +47,43 @@ class TypeLookup(
 
     def __init__(
         self,
-        __dict: Optional[Mapping[Hashable, Hashable]] = None,
+        __dict: Optional[Mapping[ValidTypes, ValidTypes]] = None,
         **kwargs: Hashable,
     ):
         __dict = __dict or {}
+        self._original_keys: Set[ValidTypes] = set()
```
billdirks: I think we should get rid of this notion of `_original_keys`. We end up maintaining it so it always matches `__dict`. I don't know if anyone cares if they originally added the `str` -> `type` mapping or if they added the `type` -> `str` mapping for a particular `str`, `type` pair.  I think it will be more useful to get the `str` keys or the `type` keys rather than the original keys.

If we are going to maintain 2 data structures, having a `__dict` with these hybrid data types and having `_original_keys` seems complicated. I believe it would be more understandable to have 2 dicts in this case: `str` -> `type` and a `type` -> `str`.

In terms of iterating twice. There just isn't too many of these that I think it will matter. That is it's unclear to me if we call iterating through everything is slower than the hash lookup we do in `original_items`.

tests/zep/test_postgres_datasource.py
```diff
@@ -145,9 +120,10 @@ def test_construct_table_asset_directly_with_splitter():
         asset = postgres_datasource.TableAsset(
             name="my_asset",
             table_name="my_table",
-            datasource=source,
             column_splitter=splitter,
         )
+        # TODO: asset custom init
+        asset._datasource = source
```
billdirks, line 125: Is this a consequence of having a combined runtime/config model?

billdirks (commented):
Thanks for doing this. I've only skimmed the test. Let's chat over video chat tomorrow.

tests/zep/test_postgres_datasource.py
Kilo59: Yes, it causes recursion errors when added to the constructor as if it were a normal field.
We can use a custom init method or a conditional in the config load hook methods to deal with it.
But considering we may be moving to a `2 model` design, I didn't think it was worth that extra complication yet.

great_expectations/zep/sources.py
Kilo59: Correct, I still think we should use namespaces. I'm trying to reduce the scope of changes in this PR where possible.

great_expectations/zep/config.py
Kilo59: These are loading/validation hooks that are called by the `pydantic` framework.
We never call them explicitly.

https://pydantic-docs.helpmanual.io/usage/validators/

great_expectations/zep/alt_config.yaml
Kilo59: Correct. I'll add them to our ZEP slack channel so we don't lose them.

great_expectations/zep/demos.py
Kilo59: Yes

great_expectations/zep/interfaces.py
Kilo59: > Would the private methods _load_execution_engine and _load_asset_subtype go away in the 2 model world?

For sure `_load_asset_subtype` would need to stay, or a method that does the same thing.
I think `_load_execution_engine` could go away, but I'll need to think about it more 🤔

great_expectations/zep/metadatasource.py
Kilo59: Sometimes it causes a collision and is a problem, sometimes it isn't. 
I think it's related to what else is being imported and run but I didn't spend the time to isolate it since it's a contrived condition.

great_expectations/zep/my_file.csv
Kilo59: Yes

great_expectations/zep/postgres_datasource.py
Kilo59: Yes, it can be modified.

We could protect it with pydantic `field` settings.
https://pydantic-docs.helpmanual.io/usage/schema/#field-customization

```python
from pydantic import Field

class TableAsset(DataAsset):
    # Instance fields
    table_name: str
    column_splitter: Optional[ColumnSplitter] = None
    name: str = Field(..., allow_mutation=False)
```

great_expectations/zep/sources.py
Kilo59: No to both.
That was the intention but I didn't end up implementing either use. I will remove them.

great_expectations/zep/sources.py
Kilo59: Yes, will do.

great_expectations/zep/sources.py
Kilo59: You're right we don't. I'm not used to extra stacktrace details provided when we use `raise Exception from e`

great_expectations/zep/sources.py
Kilo59: Correct, and done.

great_expectations/zep/sources.py
Kilo59: Probably, this is a bit of laziness on my part because I initially wanted to be able to define minimal `Datasources` without assets just to show the metaclass magic.

great_expectations/zep/postgres_datasource.py
Kilo59: I think it's `mypy`. 
I initially did these as ignores so that issues were more transparent.
`cast` is probably the better solution.

great_expectations/zep/sources.py
billdirks: Yeah, that sounds good. We could expand the comment but I think it is also fine as is.

great_expectations/zep/type_lookup.py
Kilo59: I may not be following everything here.

We don't allow overwrite of existing keys with `__setitem__`
The reason I added `_original_keys` is to prevent doing an overwrite of existing keys during the transaction commit/`.update()`.

I agree we could achieve the same thing by maintaining a mapping of just the `str`'s or just the `types` but that seems more complicated. Wouldn't that require extra conditionals as part of `__setitem__` or some slightly complicated dictionary comprehension to build it on the fly?

great_expectations/zep/type_lookup.py
Kilo59: That is an order of magnitude simpler, I didn't think of that.

However, if the non-transactional data was altered during the transaction then we have a problem. We shouldn't be using it that way... but it seems like it builds in a footgun for our future selves.

There are solutions to that though (we could make that not allowed), but they may be more complicated than the existing implementation 🤔 

```python
t = TypeLookup()
with t.transaction() as txn_t:
  t["my_type"] = list
  txn_t["my_type"] = set
```

great_expectations/zep/context.py
Kilo59: Yes

great_expectations/zep/logger.py
Kilo59: Will move this to the demo repo.

great_expectations/zep/type_lookup.py
Kilo59: TODO: Store copy and replace at end if error.
This also makes the usage simpler.

Thanks @billdirks 
```python
t = TypeLookup():
  t["my_type"] = list
  t["my_type"] = set # will cause an error and changes will be rolled back
```

great_expectations/zep/type_lookup.py
Kilo59: Removed `original_keys` stuff.

great_expectations/zep/type_lookup.py
Kilo59: removed

pyproject.toml
```diff
@@ -14,7 +14,7 @@ extend_skip_glob = ['venv/*', 'docs/*']
 
 [tool.mypy]
 python_version = "3.7"
-plugins = ["pydantic.mypy"]
+# plugins = ["pydantic.mypy"]
```
billdirks: I'm not sure what this does but is it suppose to be commented out?

great_expectations/zep/type_lookup.py
```diff
@@ -22,24 +45,37 @@ class TypeLookup(
...
```
billdirks: It looks like you skipped this one line when removing `_original_keys`.

tests/zep/config.yaml
```diff
@@ -0,0 +1,22 @@
...
+      with_splitters:
+        type: table
+        name: with_splitters
+        table_name: another_table
+        column_splitter:
+              method_name: foobar_it
+              column_name: my_column
+              name: my_splitter
+              template_params:
+                - alpha
```
billdirks: `template_params` is now `param_defaults` and is a dict. This is in flux though and we can update it when other changes go in.

billdirks (approved):
LGTM, thanks!

great_expectations/zep/type_lookup.py
Kilo59: Good catch. Removing.

tests/zep/config.yaml
Kilo59: Just updated.

pyproject.toml
Kilo59: This was actually the source of the typing cache issue for some reason.
`pydantic` wasn't being installed in (or before) the type-checking ci step, and this was somehow causing the plugin to break mypy.
Installing `pydantic` (which I assume contains the plugin) makes things work again.

PR #6321, opened 2022-11-07
[MAINTENANCE] Refactor variable aggregation/substitution logic into `ConfigurationProvider` hierarchy

cdkini:
- Encapsulate all environment variable, config variable, and runtime environment into a simple inheritance hierarchy.
    - Refactors existing code to leverage these new objects.
- As a side effect of these changes, we can now instantiate a `CloudDataContext` without environment variables.

### Definition of Done


great_expectations/core/config_provider.py
```diff
@@ -0,0 +1,169 @@
...
+            GECloudEnvironmentVariable,
+        )
+
+        return cast(
+            Dict[str, str],
+            {
+                GECloudEnvironmentVariable.BASE_URL: self._cloud_config.base_url,
+                GECloudEnvironmentVariable.ACCESS_TOKEN: self._cloud_config.access_token,
+                GECloudEnvironmentVariable.ORGANIZATION_ID: self._cloud_config.organization_id,
+            },
```
Kilo59: Why do we need to cast here?
Can we give `GECloudEnvironmentVariable` the `str` mixin?
Or say that the root method return type is .... `Dict[Union[str, Enum], str]`

cdkini: it's because `org_id` is nullable for some reason - I think I can assert or cast that instead of this whole dict.

Kilo59: Might be able to use a generic on the ABC here.

Kilo59: Actually, could you ignore the type error here with a comment about `GECloudEnvironmentVariable.ORGANIZATION_ID`?

A specific ignore with a comment is more surgical whereas the `cast` can mask other issues.

Kilo59 (approved):

great_expectations/core/config_provider.py
cdkini: done!

PR #6335, opened 2022-11-09
[MAINTENANCE] partial `cli` + `usage_stats` typing

Kilo59:
- lint all of the `cli` package with flake8
   - ignore complexity warnings in order to bring under-linting coverage 
- bring most of the cli package under type checking coverage
- fixed `unexpected keyword argument` bug in `cli.store.store_list`  - https://github.com/great-expectations/great_expectations/pull/6335/files#r1018392354


great_expectations/cli/store.py
```diff
@@ -48,7 +48,7 @@ def store_list(ctx: click.Context):
         )
     except Exception as e:
         toolkit.exit_with_failure_message_and_stats(
-            context=context,
```
Kilo59, line 51: type-checker caught a real bug/misuse here

great_expectations/cli/batch_request.py
```diff
@@ -424,9 +424,7 @@ def _get_batch_spec_passthrough(
     elif isinstance(datasource, SimpleSqlalchemyDatasource):
         # Some backends require named temporary table parameters.  We specifically elicit those and add them
         # where appropriate.
-        execution_engine: SqlAlchemyExecutionEngine = cast(
-            SqlAlchemyExecutionEngine, datasource.execution_engine
-        )
+        datasource.execution_engine
```
Kilo59: This never gets used.
Should we remove this line too? Or is there a reason to ensure that this attribute exists?

great_expectations/cli/project.py
```diff
@@ -66,13 +66,15 @@ def project_upgrade(ctx: click.Context) -> None:
         sys.exit(1)
 
 
-def do_config_check(target_directory: str) -> Tuple[bool, str, Optional[DataContext]]:
+def do_config_check(
+    target_directory: Union[str, None]
```
Kilo59: This and other similar changes were made because the caller of the method had a value that could have been none.
If needed I also made changes to the downstream methods as needed.
Example: `send_usage_message` in usage stats.

great_expectations/cli/batch_request.py
NathanFarmer: This entire method seems like it could be removed in favor of a single error check in the `else`.

great_expectations/cli/project.py
NathanFarmer: Why not `Optional[str]`?

great_expectations/cli/batch_request.py
cdkini: I want to say we keep it as some sort of check to make sure that property works as intended? we can probably get away from the cast though.

Maybe git blame and let's discuss with whoever wrote it?

great_expectations/cli/python_subprocess.py
```diff
@@ -106,7 +106,7 @@ def execute_shell_command_with_progress_polling(command: str) -> int:
                 errors=None,
             ) as proc:
                 poll_status_code: Optional[int] = proc.poll()
-                poll_stdout: str = proc.stdout.readline()
+                poll_stdout: str = proc.stdout.readline()  # type: ignore[union-attr] # stdout could be None
```
cdkini, line 109: what about `proc.stdout.readline() or ""`?

great_expectations/cli/store.py
NathanFarmer: 🙇🏻

great_expectations/cli/toolkit.py
```diff
@@ -68,7 +73,7 @@ def prompt_profile_to_create_a_suite(
 
 
 def get_or_create_expectation_suite(
-    expectation_suite_name: str,
+    expectation_suite_name: Union[str, None],
```
NathanFarmer: `Optional[str]`?

great_expectations/cli/toolkit.py
```diff
@@ -119,7 +127,7 @@ def get_or_create_expectation_suite(
 
 
 def get_default_expectation_suite_name(
-    data_asset_name: str,
+    data_asset_name: Union[str, None],
```
NathanFarmer: `Optional[str]`?

great_expectations/cli/toolkit.py
```diff
@@ -28,6 +30,9 @@
 from great_expectations.datasource import BaseDatasource
 from great_expectations.validator.validator import Validator
 
+if TYPE_CHECKING:
+    from great_expectations.datasource import LegacyDatasource
```
cdkini: Is there a linter or some sort of established way to determine if an import is only being brought in for purposes of type checking? I handrolled something but I'm trying to get away from writing up my own implementations of everything 😆

great_expectations/cli/toolkit.py
```diff
@@ -84,9 +89,12 @@ def get_or_create_expectation_suite(
...
-            expectation_suite_name = click.prompt(
-                "\nName the new Expectation Suite",
-                default=default_expectation_suite_name,
+            expectation_suite_name = cast(
+                str,
+                click.prompt(
+                    "\nName the new Expectation Suite",
+                    default=default_expectation_suite_name,
+                ),
             )
```
cdkini: Same note - should we `or ""`?

great_expectations/cli/toolkit.py
```diff
@@ -119,7 +127,7 @@ def get_or_create_expectation_suite(
 
 
 def get_default_expectation_suite_name(
-    data_asset_name: str,
+    data_asset_name: Union[str, None],
     batch_request: Optional[
         Union[str, Dict[str, Union[str, int, Dict[str, Any]]]]
```
NathanFarmer, line 137: I don't understand why we landed on this possibly being a string. It would raise an error if it was, so maybe we should just proceed assuming it can't be.

great_expectations/cli/toolkit.py
```diff
@@ -137,7 +145,7 @@ def get_default_expectation_suite_name(
 def tell_user_suite_exists(
     data_context: DataContext,
     expectation_suite_name: str,
-    usage_event: str,
+    usage_event: Union[str, None],
```
NathanFarmer: `Optional[str]`?

great_expectations/cli/v012/datasource.py
```diff
@@ -1075,7 +1078,7 @@ def get_batch_kwargs(
     return (datasource_name, batch_kwargs_generator_name, data_asset_name, batch_kwargs)
 
 
-def _get_batch_kwargs_from_generator_or_from_file_path(
+def _get_batch_kwargs_from_generator_or_from_file_path(  # noqa: C901 - 26
```
cdkini, line 1085: yikes

great_expectations/cli/toolkit.py
```diff
@@ -181,13 +189,13 @@ def get_validator(
     return validator
 
 
-def load_expectation_suite(
+def load_expectation_suite(  # type: ignore[return] # sys.exit if no suite
     data_context: DataContext,
     expectation_suite_name: str,
-    usage_event: str,
+    usage_event: Union[str, None],
```
NathanFarmer: `Optional[str]`?

great_expectations/cli/toolkit.py
```diff
@@ -226,7 +234,7 @@ def load_expectation_suite(
 
 def exit_with_failure_message_and_stats(
     data_context: DataContext,
-    usage_event: str,
+    usage_event: Union[str, None],
```
NathanFarmer: `Optional[str]`?

pyproject.toml
```diff
@@ -39,22 +39,15 @@ exclude = [
...
     'cli/upgrade_helpers/upgrade_helper_v11\.py',  # 59
     'cli/upgrade_helpers/upgrade_helper_v13\.py',  # 17
-    'cli/util\.py',  # 1
     'core/batch\.py', # 29
     'core/expectation_configuration\.py', # 21
     'core/expectation_diagnostics', # 20
     'core/expectation_validation_result\.py', # 9
     'core/usage_statistics/anonymizers', # 96
     'core/usage_statistics/usage_statistics\.py', # 19
-    'core/usage_statistics/util\.py', # 2
```
cdkini: You're crushing it 👊🏽

cdkini (approved):

great_expectations/cli/toolkit.py
```diff
@@ -699,8 +706,9 @@ def upgrade_project_zero_versions_increment(
         update_version=False,
         from_cli_upgrade_command=from_cli_upgrade_command,
     )
+    context: Optional[DataContext]  # type: ignore[no-redef] # allow re-def here
     if exception_occurred or increment_version:
-        context = None
+        context = None  # type: ignore[assignment] # see redef above
```
NathanFarmer: Why `return None` above and then assign `None` here?

great_expectations/cli/toolkit.py
```diff
@@ -1065,12 +1073,12 @@ def get_batch_request_using_datasource_name(
     additional_batch_request_args: Optional[
         Dict[str, Union[str, int, Dict[str, Any]]]
     ] = None,
-) -> Optional[Union[str, Dict[str, Union[str, int, Dict[str, Any]]]]]:
+) -> Optional[Dict[str, Union[str, int, Dict[str, Any]]]]:
```
NathanFarmer, line 1082: Does `sys.exit` cause the string to be returned? If so, why not handle it?

great_expectations/cli/python_subprocess.py
Kilo59: It's `stdout` that's the problem, not the `.readline()` result.

If `stdout` is `None` we will get an attribute error.

great_expectations/cli/v012/datasource.py
```diff
@@ -1227,7 +1230,7 @@ def _get_batch_kwargs_from_generator_or_from_file_path(
                             "\nDoes this file contain a header row?", default=True
                         )
                         batch_kwargs["reader_options"] = {"header": header_row}
-                    batch = datasource.get_batch(batch_kwargs=batch_kwargs)
+                    datasource.get_batch(batch_kwargs=batch_kwargs)
```
NathanFarmer: Does this do something in-place?

NathanFarmer (commented):
A few questions/comments from me!

great_expectations/cli/project.py
Kilo59: Because the word `"optional"` can be confusing to developers.

It isn't "optional"... it could be `None`, but it's never "optional".
`Optional` makes sense for keyword arguments that have a default.

great_expectations/cli/batch_request.py
Kilo59: @alexsherstinsky 
Although I'd prefer to save runtime changes to another PR (unless there's actually a bug or misuse).

great_expectations/cli/v012/datasource.py
Kilo59: I don't know.
I assumed it was another implicit error check.

>if we can get a batch then there was no issue.

But it's worth flagging to @alexsherstinsky that the `batch` here are never used or returned, is that intended?

great_expectations/cli/toolkit.py
Kilo59: In both cases the early `return` means I wouldn't have to do a `type: ignore`, but here the return statements would be far apart which I thought hurt readability.

But I could invert the conditions here and put them next to each other, and then I wouldn't have to ignore errors.

great_expectations/cli/toolkit.py
Kilo59: No, `sys.exit` raises a `SystemExit` exception.

great_expectations/cli/toolkit.py
Kilo59: Possibly... I think I do want to get rid of the cast, so I'll either do this or update the types of the called methods that are complaining.

great_expectations/cli/toolkit.py
Kilo59: I was thinking there is an implicit assumption that this is only a `str` if`data_asset_name` is passed in, in which case it doesn't hit the `elif batch_request` conditional (which would have caused an error). 

Ideally we would encode this assumption by declaring `overload` functions that say...
>if data_asset is None, batch_request can be a string
>if data_asset is not None, batch_request must be a dict

But I was having some minor problems getting it to work as intended and I wasn't even sure if my assumptions there were correct.

great_expectations/cli/project.py
NathanFarmer: Ok I missed that we don't have a default

Kilo59: To clarify for anyone else.
`Union[str, None]` and `Optional[str]` ARE the same thing to the type-checker.
This is just a matter of convention/style.

great_expectations/cli/toolkit.py
Kilo59: It wouldn't surprise me if this existed but I don't know one.

great_expectations/cli/toolkit.py
Kilo59: On second look, in this case I think the `cast` is appropriate. This issue is `click.prompt` doesn't know the `default` return type, so it's declared as an `Any`.

tannerbeam (commented):
looks like there's a lot going on here. i'll leave it to the pros to evaluate, but i generally support better type hints, so...nice work!

great_expectations/cli/toolkit.py
Kilo59: I ended up just returning early. Inverting the conditions made for some kinda hard to read `if not exception_occurred and not icrement_version:`  conditional.
But if folks prefer this I can do that instead.

great_expectations/cli/batch_request.py
```diff
@@ -48,7 +48,7 @@
 def get_batch_request(
     datasource: BaseDatasource,
     additional_batch_request_args: Optional[Dict[str, Any]] = None,
-) -> Dict[str, Union[str, Dict[str, Any]]]:
+) -> Dict[str, Union[str, int, Dict[str, Any]]]:
```
alexsherstinsky, line 46: @Kilo59 To help me recall, when do we get an `int` for `additional_batch_request_args`?  Thanks!

great_expectations/cli/batch_request.py
alexsherstinsky: @Kilo59 This is very weird and looks like you found a bug.  It says that I authored that line, but GitBlame has been inconsistent in my recent reviews (so I am not sure how this bug ended up in there).  Regardless, could you please try deleting that line that references `datasource.execution_engine` entirely to see if it will work without it?  Thanks!

great_expectations/cli/store.py
```diff
@@ -48,7 +48,7 @@ def store_list(ctx: click.Context):
...
+            data_context=context,
```
alexsherstinsky, line 51: ❤️

great_expectations/cli/toolkit.py
alexsherstinsky: @Kilo59 Would `expectation_suite_name: Optional[str]` not be an equivalent shorter version of the same?  Thanks.

great_expectations/cli/toolkit.py
```diff
@@ -84,9 +89,12 @@ def get_or_create_expectation_suite(
             batch_request=batch_request,
         )
         while True:
-            expectation_suite_name = click.prompt(
-                "\nName the new Expectation Suite",
-                default=default_expectation_suite_name,
+            expectation_suite_name = cast(
```
alexsherstinsky, line 97: @Kilo59 For what it is worth, in my environment, `PyCharm` is not showing any warnings for the way it currently is.

great_expectations/cli/toolkit.py
alexsherstinsky: @Kilo59 Would `data_asset_name: Optional[str]` not be appropriate shorter version?  Thanks.

great_expectations/cli/toolkit.py
alexsherstinsky: @Kilo59 Again, would the use of `Optional` for `usage_event` be wrong?  Thanks.

great_expectations/cli/toolkit.py
alexsherstinsky: @Kilo59 Same here -- would `Optional` be appropriate?  Thanks.

great_expectations/cli/toolkit.py
```diff
@@ -302,8 +310,7 @@ def validate_checkpoint(
     failure_message: Optional[str] = None,
 ) -> None:
     try:
-        # noinspection PyUnusedLocal
-        checkpoint: Union[Checkpoint, LegacyCheckpoint] = load_checkpoint(
+        load_checkpoint(
```
alexsherstinsky: @Kilo59 This goes against my preference, because it obscures the fact that `load_checkpoint()` returns a value.  Could we at least perhaps use `_ =` (just to show that in some way)?  Thanks.

great_expectations/cli/toolkit.py
```diff
@@ -343,10 +350,10 @@ def load_checkpoint(
 
 def select_datasource(
     context: DataContext, datasource_name: Optional[str] = None
-) -> BaseDatasource:
+) -> Union[BaseDatasource, LegacyDatasource, None]:
```
alexsherstinsky, line 358: @Kilo59 How come we are replacing the use of `Optional` with the longer version of what seems to be the same type declaration?  Thanks.

great_expectations/cli/v012/suite.py
```diff
@@ -111,7 +111,7 @@ def _suite_edit(
                 batch_kwargs = json.loads(batch_kwargs_json)
                 if datasource:
                     batch_kwargs["datasource"] = datasource
-                _batch = toolkit.load_batch(context, suite, batch_kwargs)
+                toolkit.load_batch(context, suite, batch_kwargs)
```
alexsherstinsky: @Kilo59 How about `_ = ...` just to not lose the interpretability that there can be a return value?  Thanks.

great_expectations/cli/v012/datasource.py
```diff
@@ -1243,7 +1246,7 @@ def _get_batch_kwargs_from_generator_or_from_file_path(
                         "\nDoes this file contain a header row?", default=True
                     )
                     batch_kwargs["reader_options"] = {"header": header_row}
-                batch = datasource.get_batch(batch_kwargs=batch_kwargs)
+                datasource.get_batch(batch_kwargs=batch_kwargs)
```
alexsherstinsky: @Kilo59 How about `_ = ...` to show that there is a return value, we are just not using it.  Thanks.

great_expectations/cli/v012/datasource.py
alexsherstinsky: @Kilo59 If this is a `v012` code, I would leave it as is; we are not supposed to be maintaining it.  In fact, it is prime for full wholesale removal.  There have been a series of JIRA items, focused on removing the V2 codebase; perhaps we can either find and resurrect them or create fresh ones, and bring it up at a meeting.  Thanks!

great_expectations/cli/toolkit.py
```diff
@@ -1084,10 +1093,8 @@ def get_batch_request_using_datasource_name(
             )
         sys.exit(1)
 
-    batch_request: Optional[
-        Union[str, Dict[str, Union[str, int, Dict[str, Any]]]]
-    ] = get_batch_request(
-        datasource=datasource,
+    batch_request = get_batch_request(
```
alexsherstinsky: @Kilo59 How come we are deleting the type declaration?  Thanks.

alexsherstinsky (changes requested):
@Kilo59 I reviewed 12 out of 15 files and noted a few questions; once they are answered, I will finish the entire review.  Thanks for doing this!

great_expectations/cli/toolkit.py
Kilo59: See the linked conversion
https://github.com/great-expectations/great_expectations/pull/6335#discussion_r1019048012

In summary, because the word `optional` can be a source of confusion.

great_expectations/cli/toolkit.py
Kilo59: Answered above

great_expectations/cli/v012/suite.py
Kilo59: Will do.

great_expectations/cli/toolkit.py
Kilo59: Because it was wrong. But I can add it back with the correct union.

great_expectations/cli/toolkit.py
Kilo59: See above.

great_expectations/cli/toolkit.py
Kilo59: `mypy` should raise a warning with any redundant casts.
https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L24

great_expectations/cli/toolkit.py
Kilo59: Done

great_expectations/cli/v012/datasource.py
Kilo59: Done

great_expectations/cli/batch_request.py
Kilo59: @alexsherstinsky I really don't want to get into changing run time behavior in this PR.
This was meant to be a quick PR to bring some of this CLI code under type coverage.

great_expectations/cli/batch_request.py
Kilo59: Because of the `limit` keyword to `BatchRequest`
https://github.com/great-expectations/great_expectations/blob/79c6ed9f1fd944bdc96369fcbfdcd969f0f481bc/great_expectations/core/batch.py#L408

great_expectations/cli/toolkit.py
Kilo59: Added back.

great_expectations/cli/batch_request.py
alexsherstinsky: Oh yes.  Thanks.

great_expectations/cli/toolkit.py
alexsherstinsky: @Kilo59 I feel that diverse style is even more confusion.  I think that we either use `Optional` everywhere or use the `Union` with `None` everywhere.  To me, consistency is more important than which alternatives we adopt.  Hence, until we have agreed upon and implemented a consistent approach across the entire codebase, I would feel better about keeping `Optional`.  Thanks.

alexsherstinsky:
@Kilo59 I am provisionally approving that you fix the failing tests (obviously) and consider my follow-up comment (otherwise, good stuff!).  Thanks!

alexsherstinsky (approved):
LGTM (pending fixing tests and considering the follow up suggesting regarding the use of the `Optional` type).  Thank you!

PR #6341, opened 2022-11-10
Zep PostgresDatasource returns a list of batches.

billdirks:
Previously `get_batch_list_from_batch_request` always sent out a list of 1 item since the batch request had to fully specify a batch. The PR allows one to partial specify a batch request and get a list of batches that match that criteria. Note, along any dimension, one can set it exactly or leave it unset (eg, you can see all months in a year). Functionality to filter on a dimension (eg, just the last 3 months of a year) is coming in a later PR.

As part of this work I removed the nascent namespacing support for batch parameters. These were unused and not fully thought out making this work harder than necessary. We can add namespacing back in a better way when we figure out the use case better.

Thank you for submitting!


great_expectations/zep/postgres_datasource.py
```diff
@@ -65,64 +84,158 @@ def get_batch_request(
...
+            )
         return BatchRequest(
             datasource_name=self.datasource.name,
             data_asset_name=self.name,
             options=options or {},
         )
 
-    def batch_request_template(
+    def _valid_batch_request_options(self, options: BatchRequestOptions) -> bool:
+        return set(options.keys()).issubset(
```
Kilo59, line 106: ## Edit
On closer look, I don't think my suggestion below makes sense given how open `batch_request_options_template` is.

Leaving the comment up in-case it inspires someone to try this technique elsewhere.

-----

This may not be that useful. But we might be to do 2 things to improve the type narrowing here.

1. Define a more specific `PostgresBatchOptions` `TypedDict `
2. Make this `_valid_batch_request_options` method a `TypeGuard` for the `PostgresBatchOptions` type.

I'm not totally sure how a `TypedDict` works with extra keys though.

```python
from typing_extensions import TypedDict, TypeGuard

class PostgresBatchOptions(TypedDict):
    foo: str
    bar: int

def _valid_batch_request_options(self, options: BatchRequestOptions) -> TypeGaurd[PostgresBatchOptions]:
        return set(options.keys()).issubset(
            set(self.batch_request_options_template().keys())
        )

```

great_expectations/zep/postgres_datasource.py
```diff
@@ -65,64 +84,158 @@ def get_batch_request(
...
+            batch_requests.append(batch_request)
+        else:
+            # All options are defined by the splitter, so we look at its default values to fill
+            # in the option values.
+            default_option_values = []
+            for option in unspecified_options:
+                default_option_values.append(
+                    self.column_splitter.param_defaults[option]
+                )
+            for option_values in itertools.product(*default_option_values):
```
Kilo59, line 209: I need to use `itertools` more 😄

Kilo59 (approved):
I will go back and review it in more detail. But I don't see any issues based on the config work that I'm doing.

great_expectations/zep/postgres_datasource.py
billdirks: Extra arguments are not supported in `TypedDicts`. Here's the github issue about it with some interesting discussion: https://github.com/python/mypy/issues/4617

Originally I had implemented `BatchOptions` using generics and had it more strongly typed. I abandoned that after getting some UX feedback and have reverted it to this untyped dict which I'm validated at runtime for a particular DataAsset. To help with the pain of "what can I put here" I've made the `batch_request_options_template` method. I would be nice to make better checks in static analysis but I'm not sure how to do that with the current requirements.

I am not familiar with `TypeGuard`s and will look into that.

great_expectations/zep/postgres_datasource.py
billdirks: I was going to implement this but then did some searching. Happy to have found it!

PR #6345, opened 2022-11-11
[MAINTENANCE] Move config variable substitution logic into `ConfigurationProvider`

cdkini:
- Previously, `substitute_all_config_variables` and `substitute_config_variable` were helper functions sitting in `util.py`
  - This PR encapsulates this behavior into a new helper class `ConfigurationSubstitutor`
- Since config variable aggregation (combining env vars, config variables, and runtime environment) is always used to prepare for variable substitution, this PR combines this behavior into a single method call: `get_substituted_config`.
  - We do this by placing our new `ConfigurationSubstitutor` as an instance attr on the `ConfigurationProvider` class.
- Any classes that require the logic now take in an instance of the `ConfigurationProvider` (i.e. `DataContextVariables`).
- Updated tests to reflect these changes - no functional/breaking changes but lots of diffs due to the number of lines impacted.

To simplify, let's look at the pattern before and now.

Before:
```python
        # AbstractDataContext.get_config_with_variables_substituted
        substituted_config_variables: dict = substitute_all_config_variables(
            self.config_variables,
            dict(os.environ),
            self.DOLLAR_SIGN_ESCAPE_STRING,
        )

        substitutions: dict = {
            **substituted_config_variables,
            **dict(os.environ),
            **self.runtime_environment,
        }

        return DataContextConfig(
            **substitute_all_config_variables(
                config, substitutions, self.DOLLAR_SIGN_ESCAPE_STRING
            )
        )
```

After:
```python
        # AbstractDataContext.get_config_with_variables_substituted
        return DataContextConfig(**self.config_provider.substitute_config(config))
```

### Definition of Done


great_expectations/checkpoint/checkpoint.py
```diff
@@ -302,23 +299,7 @@ def _get_substituted_runtime_kwargs(
...
-            **dict(os.environ),
-            **self.data_context.runtime_environment,
-        }
-
-        return substitute_all_config_variables(
-            data=config,
-            replace_variables_dict=substitutions,
-            dollar_sign_escape_string=self.data_context.DOLLAR_SIGN_ESCAPE_STRING,
-        )
+        return self.data_context._config_provider.substitute_config(config)
```
cdkini: Direct access to private attr feels bad - we have a few solutions here:
1. Keep it as is (this violation is done many times in this class)
2. Pass in config provider in the constructor
3. Add a property so we can access properly

great_expectations/core/config_provider.py
```diff
@@ -1,22 +1,364 @@
...
+try:
+    from google.cloud import secretmanager
+except ImportError:
+    secretmanager = None
+
+logger = logging.getLogger(__name__)
 yaml = YAMLHandler()
 
 
+class ConfigurationSubstitutor:
```
cdkini: This class is quite simply taking `substitute_all_config_variables` and `substitute_config_variable` and putting it in a class. All the logic here is ported over from `util.py`.

great_expectations/core/config_provider.py
```diff
@@ -1,22 +1,364 @@
...
+    def __init__(self) -> None:
+        # Using the @lru_cache decorator on method calls can create memory leaks - an attr is preferred here.
+        # Ref: https://stackoverflow.com/a/68550238
+        self._secret_store_cache = lru_cache(maxsize=None)(
+            self._substitute_value_from_secret_store
+        )
```
cdkini: The secret store method below was previously decorated when it was a function - I believe there are issues with caching instance methods so I've opted for a slightly different approach here.

great_expectations/core/config_provider.py
```diff
@@ -1,22 +1,364 @@
...
+
+    def substitute_all_config_variables(
+        self, data, replace_variables_dict, dollar_sign_escape_string: str = r"\$"
+    ):
```
cdkini: I can come back and type hint all of this (or let me know if you'd prefer I do that now)

great_expectations/core/config_substitutor.py
```diff
@@ -0,0 +1,352 @@
...
+
+
+class _ConfigurationSubstitutor:
+    """
+    TODO
+    """
+
+    def __init__(self) -> None:
+        # Using the @lru_cache decorator on method calls can create memory leaks - an attr is preferred here.
+        # Ref: https://stackoverflow.com/a/68550238
```
cdkini, line 45: This was added by me - everything else in this file while directly ported over from `util.py`

cdkini: Happy to clean up the file (refactors and type hints) but perhaps that should be a subsequent effort

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -332,6 +330,7 @@ def _init_variables(self) -> CloudDataContextVariables:
 
         variables = CloudDataContextVariables(
             config=self._project_config,
+            config_provider=self._config_provider,
```
cdkini: `DataContextVariables` now leverage an instance of `ConfigurationProvider` now that substitution logic is a part of this class.

great_expectations/core/config_substitutor.py
```diff
@@ -0,0 +1,356 @@
...
+        self._secret_store_cache = lru_cache(maxsize=None)(
+            self.substitute_value_from_secret_store
+        )
+
+    def substitute_all_config_variables(
```
cdkini, line 50: Only this method and `substitute_config_variable` should be part of this class' public API - I will leave refactoring the store-specific methods to private methods as a separate task due to the sheer number of changes made in this PR.

tests/core/test_config_substitutor.py
```diff
@@ -0,0 +1,261 @@
...
+    return ConfigurationSubstitutor()
+
+
+@contextmanager
+def does_not_raise():
+    yield
+
+
+@pytest.mark.parametrize(
+    "input_value,method_to_patch,return_value",
```
cdkini, line 23: These are all tests that previously lived in `tests/data_context/test_data_context_utils.py`

great_expectations/core/config_provider.py
```diff
@@ -4,26 +4,45 @@
...
+
+        Args:
+            config: The config object to update.
+            config_values: The dictionary of values to use during the substitution process.
+                           If omitted, any values derived from registered providers will be used.
+
+        Returns:
+            The input config object with any $VARIABLES replaced with their corresponding config values.
+        """
+        if config_values is None:
```
billdirks, line 42: I know we do this a lot throughout the codebase but I find the pattern of allowing a passed in value or acting on internal state confusing. I guess in the other examples internal state is mutated but that doesn't seem to be the case here. Is this to allow overriding of default values?

great_expectations/core/config_provider.py
```diff
@@ -4,26 +4,45 @@
...
+
     @abstractmethod
     def get_values(self) -> Dict[str, str]:
         """
         Retrieve any configuration variables relevant to the provider's environment.
         """
         pass
 
+    def substitute_config(
+        self, config: Any, config_values: Optional[Dict[str, str]] = None
```
billdirks, line 28: Do we know anything about the type of `config` here?

great_expectations/data_context/config_validator/yaml_config_validator.py
```diff
@@ -338,26 +333,16 @@ def _prepare_config_string_with_substituted_variables(
...
+            config_values = config_provider.get_values()
 
-            substitutions: dict = {
-                **substituted_config_variables,  # type: ignore[list-item]
-                **dict(os.environ),
-                **runtime_environment,
-            }
+            # While normally we'd just call `self.config_provider.substitute_config()`,
+            # we need to account for `runtime_environment` values that may have been passed.
+            config_values.update(runtime_environment)
```
billdirks, line 337: This permanently changes `config_values` which is state inside the `config_provider`.  Is this expected by the caller? This ultimately gets triggered in the public function `test_yaml_config` which doesn't sound like something that mutates state.

billdirks (commented):
Thanks for doing this! I haven't looked at most of this PR yet. I need to head off to a meeting right now so thought I'd leave the comments I currently have and come back later.

great_expectations/core/config_provider.py
cdkini: So there are a few options per the method signature (which was ported over from a util file) - `BaseYamlConfig`, `dict`, `list`, or a catch-all "other".

I'd like to make this more specific but current behavior is essentially `Any`.

great_expectations/data_context/config_validator/yaml_config_validator.py
cdkini: so everytime I use `config_provider`, the config values are generated JIT - this is due to the fact that a user can modify env vars, config variables, etc during a runtime and we need to always be ready to capture those changes.

This `update()` call modifies the state of the `config_values` dict but that dict is not saved as state on the provider.

great_expectations/core/config_provider.py
cdkini: We set this to `None` as we're not able to use default dictionaries (due to shared state with mutable defaults in Python).

In essence, we generally use the values derived from `self.get_values()` but also want the ability to take in runtime values - an example of this can be found in `yaml_config_validator.py` (where you've commented on the `update()` call).

Note that the default call to `self.get_values()` generates values JIT and therefore we're never working with internal state. Lmk if that makes sense.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1020,14 +1011,10 @@ def _serialize_substitute_and_sanitize_datasource_config(
...
         """
-        substitutions: dict = self._determine_substitutions()
         datasource_dict: dict = serializer.serialize(datasource_config)
 
         substituted_config: dict = cast(
-            dict,
-            substitute_all_config_variables(
-                datasource_dict, substitutions, self.DOLLAR_SIGN_ESCAPE_STRING
-            ),
+            dict, self.config_provider.substitute_config(datasource_dict)
```
billdirks, line 1017: You could removed the `: dict` here since you are explicitly casting it.

tests/core/test_config_substitutor.py
```diff
@@ -0,0 +1,261 @@
...
+            "success",
+        ),
+        (
+            "secret|https://my-vault-name.vault.azure.net/secrets/my_secret",
+            "great_expectations.core.config_substitutor.ConfigurationSubstitutor.substitute_value_from_azure_keyvault",
+            "success",
+        ),
+    ],
+)
+def test_substitute_value_from_secret_store(
```
billdirks, line 45: It looks like these tests were just moved over from `tests/data_context/test_data_context_utils.py`?
If it is not arduous can you mark these tests as `unit`, `integration`, or `e2e` as appropriate?

billdirks (approved):
Thanks, LGTM! A couple small comments.

cdkini:
c

tests/core/test_config_substitutor.py
cdkini: done!

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: done!

PR #6348, opened 2022-11-13
[MAINTENANCE] Refactor out `termcolor` dependency

cdkini:
- Use built-in `click` functionality so we can get rid of `termcolor`
    - This seems to be the more robust of the two libraries and we already need `click` for CLI processing.

### Definition of Done

### DIFF
--- a/great_expectations/cli/pretty_printing.py
+++ b/great_expectations/cli/pretty_printing.py
@@ -3,6 +3,9 @@ import sys
 from typing import List, Optional
 
 import click
+from typing_extensions import Final
+
+SUPPORTED_CLI_COLORS: Final[List[str]] = ["blue", "cyan", "green", "yellow", "red"]
 
 
 def cli_message(string: str) -> None:
@@ -10,10 +13,9 @@ def cli_message(string: str) -> None:
 
 
 def cli_colorize_string(string: str) -> str:
-    colors = ("blue", "cyan", "green", "yellow", "red")
-    for color in colors:
+    for color in SUPPORTED_CLI_COLORS:
         string = re.sub(
-            f"<{color}>(.*?)</{color}>",
+            f"<[ color ]>(.*?)</{color}>",
             click.style(r"\g<1>", fg=color),
             string,
             flags=re.DOTALL,  # the DOTALL flag means that `.` includes newlines for multiline comments inside these tags
--- a/great_expectations/cli/pretty_printing.py
+++ b/great_expectations/cli/pretty_printing.py
@@ -2,7 +2,15 @@ import re
 import sys
 from typing import List, Optional
 
-from termcolor import colored
+import click
+
+try:
+    from typing import Final
+except ImportError:
+    # Fallback for python < 3.8
+    from typing_extensions import Final
+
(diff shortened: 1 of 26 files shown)


great_expectations/cli/v012/util.py
```diff
@@ -2,47 +2,35 @@
...
-try:
-    from termcolor import colored
-except ImportError:
-    colored = None
+SUPPORTED_CLI_COLORS: Final[List[str]] = ["blue", "cyan", "green", "yellow", "red"]
 
 
-def cli_message(string) -> None:
+def cli_message(string: str) -> None:
     print(cli_colorize_string(string))
```
cdkini, line 28: This entire file is duplicating efforts in the primary CLI directory - while this feels bad (and is bad), this code is outdated and will soon be removed from the codebase.

I don't think it is worth the effort to refactor existing code to refactor out this duplication. See a Slack conversation around this for more information: https://greatexpectationslabs.slack.com/archives/C9GDA5SAY/p1668435895377679

tests/cli/test_datasource_pandas.py
```diff
@@ -8,7 +8,7 @@
 
 from great_expectations import DataContext
 from great_expectations.cli import cli
-from tests.cli.utils import assert_no_logging_messages_or_tracebacks
+from tests.cli.utils import assert_no_logging_messages_or_tracebacks, escape_ansi
```
cdkini, line 11: Our testing for CLI output is incredibly granular and is a bit overkill in my opinion. I've added this `escape_ansi` helper to strip expected and actual outputs so we can assert the important textual content. Colors are helpful but I think validating hundreds of ANSI codes doesn't provide any real value.

Let me know if you disagree!

great_expectations/cli/pretty_printing.py
```diff
@@ -2,33 +2,25 @@
 import sys
 from typing import List, Optional
 
-from termcolor import colored
+import click
+from typing_extensions import Final
+
+SUPPORTED_CLI_COLORS: Final[List[str]] = ["blue", "cyan", "green", "yellow", "red"]
```
Kilo59: If this is a constant should we make this a `tuple`?

cdkini: yeah I like that

cdkini: done!

cdkini: done!

Kilo59 (approved):
🚀

tests/cli/utils.py
```diff
@@ -128,3 +129,9 @@ def assert_no_tracebacks(click_result):
     except ValueError as ve:
         # sometimes stderr is not captured separately
         pass
+
+
+def escape_ansi(line: str) -> str:
```
Kilo59, line 134: If this get called repeatedly on the same string we could speed up the tests with a `functool.lru_cache` here.

Although it's very possible the difference is negligible. 
```suggestion
@functools.lru_cache(maxsize=64)
def escape_ansi(line: str) -> str:
```

PR #6361, opened 2022-11-15
[MAINTENANCE] Pin `mypy` to `0.990`

Kilo59:
The following changes should make our typing CI step much more consistent and reliable.

### Pin mypy version to prevent CI breakage - [as recommended by `mypy`](https://mypy.readthedocs.io/en/stable/existing_code.html#run-mypy-consistently-and-prevent-regressions)
  - Fix typing issues related to latest change - https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html
### Update `mypy` config to always run as if it was python 3.7 (our oldest supported version)
  - fix typing issues on python 3.7
  - fixed 1 potential `pprint` related bug - https://docs.python.org/3/library/pprint.html#pprint.PrettyPrinter

### DIFF
--- a/great_expectations/core/expectation_suite.py
+++ b/great_expectations/core/expectation_suite.py
@@ -752,7 +752,6 @@ class ExpectationSuite(SerializableDictDot):
         pprint.pprint(
             object=pprint_objects,
             indent=2,
-            sort_dicts=False,
         )
 
     def get_grouped_and_ordered_expectations_by_domain_type(
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ skip_gitignore = true
 extend_skip_glob = ['venv/*', 'docs/*']
 
 [tool.mypy]
+python_version = "3.7"
 files = [
     "great_expectations",
     # "contrib" # ignore entire `contrib` package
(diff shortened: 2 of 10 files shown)


great_expectations/core/expectation_suite.py
```diff
@@ -752,7 +752,6 @@ def show_expectations_by_expectation_type(
         pprint.pprint(
             object=pprint_objects,
             indent=2,
-            sort_dicts=False,
```
Kilo59, line 755: @NathanFarmer 
This wasn't added until python 3.8 😢 
https://docs.python.org/3/library/pprint.html#pprint.PrettyPrinter

cdkini (approved):

pyproject.toml
```diff
@@ -13,6 +13,7 @@ skip_gitignore = true
 extend_skip_glob = ['venv/*', 'docs/*']
 
 [tool.mypy]
+python_version = "3.7"
```
Kilo59, line 16: This should help a bit with getting consistent errors across dev environments.
If you are using a python `3.10` feature that doesn't exist in `3.7` mypy should warn you.

PR #6368, opened 2022-11-15
[BUGFIX] Use `requirements.txt` file when installing linting/static check dependencies in CI

cdkini:
- Use pinned requirements in CI (instead of always pulling the latest).
    - Our CI should reflect our local environments - if we have a pin in our requirements, let's use that.

### Definition of Done


Kilo59 (approved):
🙌

PR #6375, opened 2022-11-16
[MAINTENANCE] Make `ConfigurationProvider` and `ConfigurationSubstitutor` private

cdkini:
- Make util classes private (using `_Class` syntax)
- Misc type hints

### Definition of Done


great_expectations/core/config_provider.py
```diff
@@ -6,16 +6,16 @@
...
-from great_expectations.core.config_substitutor import ConfigurationSubstitutor
+from great_expectations.core.config_substitutor import _ConfigurationSubstitutor
 from great_expectations.core.yaml_handler import YAMLHandler
 from great_expectations.data_context.types.base import GXCloudConfig
 
 yaml = YAMLHandler()
 
 
-class AbstractConfigurationProvider(ABC):
+class _AbstractConfigurationProvider(ABC):
```
Kilo59, line 16: I will admit now that I'm seeing it.... Having private ABC's seems a bit odd.
But I don't have the words to express why 😆

cdkini: I can undo this if we don't like it! Don't feel strongly either way 😄

tests/core/test_config_substitutor.py
```diff
@@ -242,11 +209,12 @@ class Response:
...
             None,
         ),
+        # If the regex is not matched, we return the value as-is
         (
             "secret|https://my_vault_name.vault.azure.net/secrets/my-secret/A0000000000000000000000000000000|key",
             None,
-            pytest.raises(ValueError),
-            None,
+            does_not_raise(),
+            "secret|https://my_vault_name.vault.azure.net/secrets/my-secret/A0000000000000000000000000000000|key",
```
cdkini, line 217: This was a faulty test case - normally, you'd run `substitute_config_variable`, which would call `substitute_value_from_secret_store`, which would run this string against a regex and conditionally call this Azure-specific method.

As this string doesn't match the regex, there's no way we'd even hit this method.

TLDR: The test case that was here was testing an impossible situation - we would not raise an exception so I'm changing the test case to match what would actually happen in prod.

great_expectations/core/config_substitutor.py
```diff
@@ -40,16 +40,25 @@ class ConfigurationSubstitutor:
     should be defined herein.
     """
 
+    AWS_PATTERN = r"^secret\|arn:aws:secretsmanager:([a-z\-0-9]+):([0-9]{12}):secret:([a-zA-Z0-9\/_\+=\.@\-]+)"
+    GCP_PATTERN = (
+        r"^secret\|projects\/([a-z0-9\_\-]{6,30})\/secrets/([a-zA-Z\_\-]{1,255})"
+    )
+    AZURE_PATTERN = r"^secret\|(https:\/\/[a-zA-Z0-9\-]{3,24}\.vault\.azure\.net)\/secrets\/([0-9a-zA-Z-]+)"
+
```
cdkini, line 48: Removing a bit of duplication through class attrs

Kilo59 (approved):

PR #6380, opened 2022-11-16
[MAINTENANCE] Handle both `ExpectationConfiguration` and `ExpectationValidationResult` in default Atomic renderers and cleanup `include_column_name`

NathanFarmer:
- Support both `ExpectationConfiguration` and `ExpectationValidationResult` in both diagnostic and prescriptive atomic renderers
- Remove duplicative `include_column_name` logic in all existing renderers

### Definition of Done


great_expectations/expectations/core/expect_column_values_to_not_be_in_set.py
```diff
@@ -233,20 +235,23 @@ def _atomic_prescriptive_template(
...
         cls,
-        configuration=None,
-        result=None,
-        language=None,
-        runtime_configuration=None,
+        configuration: Optional[ExpectationConfiguration] = None,
+        result: Optional[ExpectationValidationResult] = None,
+        language: Optional[str] = None,
+        runtime_configuration: Optional[dict] = None,
         **kwargs,
```
cdkini, line 242: Do we know the return annotation here?

cdkini: And the type of kwargs?

great_expectations/expectations/expectation.py
```diff
@@ -331,23 +343,35 @@ def _atomic_prescriptive_template(
...
 
-        template_str = "$expectation_type(**$kwargs)"
+        expectation_type: str
+        expectation_kwargs: dict
+        if configuration:
+            expectation_type = configuration.expectation_type
+            expectation_kwargs = configuration.kwargs
+        elif result and result.expectation_config:
+            expectation_type = result.expectation_config.expectation_type
+            expectation_kwargs = result.expectation_config.kwargs
```
cdkini: Because we don't have an `else` clause, aren't these variables possibly unbounded?

great_expectations/expectations/expectation.py
```diff
@@ -382,12 +407,20 @@ def _prescriptive_summary(
...
         **kwargs: dict,
     ):
+        expectation_type: str
+        expectation_kwargs: dict
+        if configuration:
+            expectation_type = configuration.expectation_type
+            expectation_kwargs = configuration.kwargs
+        elif result and result.expectation_config:
+            expectation_type = result.expectation_config.expectation_type
+            expectation_kwargs = result.expectation_config.expectation_type
```
cdkini: Same note here

great_expectations/expectations/expectation.py
```diff
@@ -437,22 +470,23 @@ def _diagnostic_meta_properties_renderer(
...
+        if meta_properties_to_render:
             for key in sorted(meta_properties_to_render.keys()):
                 meta_property = meta_properties_to_render[key]
-                if meta_property is not None:
+                if meta_property:
                     try:
                         # Allow complex structure with . usage
+                        assert isinstance(
+                            result.expectation_config, ExpectationConfiguration
+                        )
```
cdkini, line 528: Is this an invariant or is there the possibility that a user will face this?

If the latter, I say let's raise a `ValueError`

contrib/experimental/great_expectations_experimental/expectations/expect_column_discrete_entropy_to_be_between.py
```diff
@@ -309,9 +309,8 @@ def _prescriptive_renderer(
         **kwargs,
     ):
         runtime_configuration = runtime_configuration or {}
-        include_column_name = runtime_configuration.get("include_column_name", True)
         include_column_name = (
-            include_column_name if include_column_name is not None else True
```
cdkini, line 314: Thoughts on this suggestion?
```python
include_column_name = runtime_configuration.get("include_column_name", True)
if include_column_name is None:
    include_column_name = True
```

I'm always wary of one liners like this but I think this is readable enough - will defer to you.
Also, would it be valuable to encapsulate this logic in a util method such that we can share the logic across the hierarchy?

cdkini (commented):
Misc comments but mainly focused on the possibly unbounded variable.

great_expectations/expectations/expectation.py
NathanFarmer: Yes, they are. If I change the `elif` to `else` I will get static type checking errors, because `result` is `Optional`. We have a patter whereby if a renderer fails, it will fall back to the "failure string" renderer. After thinking more about it, I could use an `else` and then `assert isinstance(result.expectation_config, ExpectationConfiguration)` that would solve the `mypy` errors and this renderer would still fail if both `ExpectationValidationResult` and `ExpectationConfiguration` are missing.

contrib/experimental/great_expectations_experimental/expectations/expect_column_discrete_entropy_to_be_between.py
NathanFarmer: Maybe a util method is the way to go. I'm just looking for a way for users to understand what this is doing. It's quite odd.

NathanFarmer: Per our offline convo, we will encapsulate this and other duplicative logic in these renderer templates in a class method.

great_expectations/expectations/core/expect_column_values_to_not_be_in_set.py
NathanFarmer: Per our offline conversation, I will address this in a subsequent PR.

Kilo59 (approved):

great_expectations/expectations/expectation.py
Kilo59: If we are actually expecting a particular error path at runtime then I think a `mypy` ignore error is appropriate.

Kilo59: Actually it sounds like this calls for an overload.
You're saying that even tho `result` & `configuration` are both optional. But one of them must be provided correct?

great_expectations/expectations/expectation.py
NathanFarmer: ✅

great_expectations/expectations/expectation.py
Kilo59: @NathanFarmer 

Just a suggestion. I approved anyway.

great_expectations/expectations/expectation.py

great_expectations/expectations/expectation.py
NathanFarmer: ✅

PR #6384, opened 2022-11-17
[MAINTENANCE] Loosen `pydantic` version requirement

Kilo59:
- loosen pydantic version requirement to be between v1 & v2
- add to contrib req file


reqs/requirements-dev-contrib.txt
```diff
@@ -4,7 +4,7 @@ invoke>=1.7.1
 isort==5.10.1
 mypy==0.991
 pre-commit>=2.6.0
-pydantic>=1.10
+pydantic>=1.0<2.0 # needed for mypy plugin support
```
alexsherstinsky, line 7: @Kilo59 Is it appropriate to have the same line in multiple files?  Or should we consider refactoring?  Perhaps in a follow-on pull request?  Thanks.

alexsherstinsky (approved):
LGTM -- as long as the comment I made is reviewed and a JIRA task created to address in the future (if appropriate).  Thank you!

reqs/requirements-dev-contrib.txt
Kilo59: Yes, we need to update our static typing step to install all our required dependencies, and then we can remove it from here.

PR #6392, opened 2022-11-17
[BUGFIX] Metrics return value no longer returns None for `unexpected_index_list` - Sql and Spark

Shinnnyshinshin:
# Bugfix as part of work being done for #6381 
- `MapMetrics` were found to be returning extraneous values that were set to `None`
- `SparkDFExecutionEngine` and `SqlAlchemyExecutionEngine` do not support the `unexpected_index_list` metric or `partial_unexpected_index_list`. 
- Instead of not outputting these metrics, the output would contain `"partial_unexpected_index_list": None` and `"unexpected_index_list": None
- This PR proposes to only include the `unexpected_index_list` and `partial_unexpected_index_list` *if and only if* they are available, and not include if they are `None`. 

### Definition of Done


great_expectations/expectations/expectation.py
```diff
@@ -3005,26 +3005,25 @@ def _format_map_output(
...
+                    }
+                )
             return_obj["result"].update(
-                {
-                    "partial_unexpected_index_list": unexpected_index_list[
-                        : result_format["partial_unexpected_count"]
-                    ]
-                    if unexpected_index_list is not None
-                    else None,
-                    "partial_unexpected_counts": partial_unexpected_counts,
```
Shinnnyshinshin, line 3062: This if statement was making it so that `partial_unexpected_index_list` was always being included in the `return_obj` even if it was invalid (ie not supported by the `ExecutionEngine`)

great_expectations/expectations/expectation.py
```diff
@@ -3005,26 +3005,25 @@ def _format_map_output(
...
 
     if result_format["result_format"] == "SUMMARY":
         return return_obj
 
-    return_obj["result"].update(
-        {
-            "unexpected_list": unexpected_list,
-            "unexpected_index_list": unexpected_index_list,
-        }
-    )
```
Shinnnyshinshin, line 3074: Same here. This statement ensured that  `unexpected_index_list` was always outputted, even in situations where it wasn't supported by the `ExecutionEngine`

tests/expectations/test_expectation_arguments.py
```diff
@@ -507,7 +507,6 @@ def test_result_format_configured_no_set_default_override(
         "unexpected_count": 0,
         "unexpected_percent": 0.0,
         "partial_unexpected_list": [],
-        "partial_unexpected_index_list": None,
```
Shinnnyshinshin, line 510: here is an example of the value being outputted for a `SparkDFExecutionEngine`,  which doesn't support `unexpected_index`

great_expectations/expectations/expectation.py
```diff
@@ -3052,26 +3052,25 @@ def _format_map_output(
...
+    if unexpected_list is not None:
+        return_obj["result"].update({"unexpected_list": unexpected_list})
+    if unexpected_index_list is not None:
+        return_obj["result"].update({"unexpected_index_list": unexpected_index_list})
```
Shinnnyshinshin, line 3073: This is the fix : a more specific check for whether `unexpected_list` or `unexpected_index_list` is not `None`.

tests/expectations/test_expectation_arguments.py
Shinnnyshinshin: @anthonyburdi this is the change that I mentioned in Slack

anthonyburdi (approved):
LGTM!

PR #6395, opened 2022-11-18
[MAINTENANCE] Migrate additional CRUD methods from `BaseDataContext` to `AbstractDataContext`

cdkini:
- Move over more methods down to the parent class

### Definition of Done


great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -713,60 +568,30 @@ def add_checkpoint(
...
-            ge_cloud_id = response.ge_cloud_id
-
-        # If an id is present, we want to prioritize that as our key for object retrieval
-        if ge_cloud_id:
-            name = None  # type: ignore[assignment]
-
-        profiler = self.get_profiler(name=name, ge_cloud_id=ge_cloud_id)
-        return profiler
+    def list_checkpoints(self) -> Union[List[str], List[ConfigurationIdentifier]]:
+        return self._data_context.list_checkpoints()
```
cdkini, line 577: This pattern for checkpoints and RBP is necessary due to how we use `self._data_context`.

This will be cleaned up in a subsequent ticket such that we use `__getattr__` and/or `__getattribute__` to dynamically pass calls to the underlying `self._data_context`. This will remove the need for redefining functions like this.

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -104,7 +104,7 @@ def default_checkpoints_exist(directory_path: str) -> bool:
         return os.path.isdir(checkpoints_directory_path)
 
     def list_checkpoints(
-        self, ge_cloud_mode: bool
+        self, ge_cloud_mode: bool = False
```
cdkini, line 107: We will eventually want to refactor out `ge_cloud_mode` - for now, I've opted to add a default value (as 99% of usecases will not leverage this flag).

tests/data_context/test_data_context_profilers.py
```diff
@@ -3,25 +3,13 @@
...
-    with mock.patch(
-        "great_expectations.data_context.DataContext.profiler_store",
-    ) as mock_profiler_store:
-        mock_profiler_store.__get__ = mock.Mock(return_value=None)
-        with pytest.raises(ge_exceptions.StoreConfigurationError) as e:
-            empty_data_context.list_profilers()
-
-    assert "not a configured store" in str(e.value)
-
-
```
cdkini, line 24: Deleted this test as it was duplicatively testing the `self.profiler_store` property

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -561,101 +508,9 @@ def save_expectation_suite(
...
     def root_directory(self) -> Optional[str]:
         if hasattr(self._data_context, "_context_root_directory"):
-            return self._data_context._context_root_directory  # type: ignore[union-attr]
+            return self._data_context._context_root_directory
         return None
 
-    @usage_statistics_enabled_method(
-        event_name=UsageStatsEvents.DATA_CONTEXT_BUILD_DATA_DOCS,
-    )
-    def build_data_docs(
```
cdkini, line 570: This was migrated last PR but I forgot to delete the `BaseDataContext` impl

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3643,3 +3679,43 @@ def _init_site_builder_for_data_docs_site_creation(
             },
         )
         return site_builder
+
+    def escape_all_config_variables(
```
Shinnnyshinshin, line 3683: oo this is nice :)

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -674,3 +686,17 @@ def _init_site_builder_for_data_docs_site_creation(
...
         return site_builder
+
+    def _determine_key_for_profiler_save(
+        self, name: str, id: Optional[str]
+    ) -> Union[ConfigurationIdentifier, GXCloudIdentifier]:
+        """
+        Note that this explicitly overriding the `AbstractDataContext` helper method called
+        in `self.save_profiler()`.
+
+        The only difference here is the creation of a Cloud-specific `GXCloudIdentifier`
```
Shinnnyshinshin, line 697: :heart:

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -104,7 +104,7 @@ def default_checkpoints_exist(directory_path: str) -> bool:
...
     ) -> Union[List[str], List[ConfigurationIdentifier]]:
```
Shinnnyshinshin, line 108: :heart:

Shinnnyshinshin (approved):
LGTM. We're getting close!

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3643,3 +3679,43 @@ def _init_site_builder_for_data_docs_site_creation(
...
+        self,
+        value: Union[str, dict, list],
+        dollar_sign_escape_string: str = DOLLAR_SIGN_ESCAPE_STRING,
+        skip_if_substitution_variable: bool = True,
+    ) -> Union[str, dict, list]:
```
Kilo59: Should use a `TypeVar` here.
~Let me double check this is the right syntax....~

https://docs.python.org/3/library/typing.html#typing.TypeVar

```python
from typing import TypeVar

T = TypeVar("T", str, dict, list)

 def escape_all_config_variables(
        self,
        value: T,
        dollar_sign_escape_string: str = DOLLAR_SIGN_ESCAPE_STRING,
        skip_if_substitution_variable: bool = True,
    ) -> T:

```

Kilo59 (changes requested):
LGTM 1 small detail.

We should use a `TypeVar` for `escape_all_config_variables`

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -148,6 +155,9 @@
 yaml = YAMLHandler()
 
 
+T = TypeVar("T", bound=Union[dict, list, str])
```
```
great_expectations/data_context/data_context/abstract_data_context.py:3705: error:
Incompatible return value type (got "Dict[Any, Any]", expected "str") 
[return-value]
                return {
```
https://github.com/great-expectations/great_expectations/pull/6395/commits/88d2196ac74b0730eef634c2de8783d617b7df81
@cdkini This type warning was from your previous commit.
I didn't realize this was a recursive method. That complicates things a bit.

The point of the `TypeVar` in this case is to give a narrower return type than the union.
If I call `escape_all_config_variables` with a `dict` it will tell the type system that a `dict` will be returned rather than a `dict` or  `str` or `list`.
Making the `TypeVar` a `Union` defeats the purpose of using a `TypeVar` to begin with.

I would drop/revert the `chore: update typevar usage` commit.

There are 2 solutions here to deal with this.
1. Use the original `TypeVar` (sans `union`) and ignore the internal `return-type` warning.
2. Use a `typing.overload` instead

-------

### Use `TypeVar("T", dict, list, str)`

For `#1` I think the reason we have to do this is because of when the `TypeVar` is bound/re-bound, the type is bound to `X` type but then when the function calls itself again the type is `Y` and this causes issues. Regardless I think the most important thing is the type being returned when the stack terminates, and less so the "correctness" of the function body. So I think a single comment `# type ignore[retun-value] # recursive call expects str here` is the best approach (if the solution is still to use a `TypeVar`).

```python
# change back to T = TypeVar("T", dict, list, str)
# and just ignore the type warning on this line

        if isinstance(value, dict) or isinstance(value, OrderedDict):
            return {   # type ignore[retun-value] # recursive call expects str here
                k: self.escape_all_config_variables(
                    v, dollar_sign_escape_string, skip_if_substitution_variable
                )
```

--------

## Use `@overload`

`#2` is more "correct" in that it still retains the narrowness of the return type and shouldn't lead to any issues in the function body either. But it's a lot more work than adding `1` type ignore comment. Thankfully there's only 3 overloads to define.
We declare additional function signatures for all of the possible values and their expected return types.

Note the example below can be used as is. NOTE: the `...` are part of the syntax, they are not placeholders meant to be filled in.
```python
# assume -> from typing import overload

    @overload
    def escape_all_config_variables(
        self,
        value: Union[dict, OrderedDict],
        dollar_sign_escape_string: str = ...,
        skip_if_substitution_variable: bool = ...,
    ) -> dict:
        ...

    @overload
    def escape_all_config_variables(
        self,
        value: list,
        dollar_sign_escape_string: str = ...,
        skip_if_substitution_variable: bool = ...,
    ) -> list:
        ...

    @overload
    def escape_all_config_variables(
        self,
        value: str,
        dollar_sign_escape_string: str = ...,
        skip_if_substitution_variable: bool = ...,
    ) -> str:
        ...

    def escape_all_config_variables(
        self,
        value: Union[dict, OrderedDict, list, str],
        dollar_sign_escape_string: str = DOLLAR_SIGN_ESCAPE_STRING,
        skip_if_substitution_variable: bool = True,
    ) -> Union[dict, list, str]:
```

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3643,3 +3683,51 @@ def _init_site_builder_for_data_docs_site_creation(
...
+
+        Returns:
+            input value with all `$` characters replaced with the escape string
+        """
+        if isinstance(value, dict) or isinstance(value, OrderedDict):
+            return cast(
+                T,
+                {
+                    k: self.escape_all_config_variables(
+                        value=v,
```
Kilo59: We shouldn't use `TypeVar`'s this way.

Kilo59 (commented):
Using `T = TypeVar("T", bound=Union[dict, list, str])` is more or less the same as `Union[dict, list, str]` without the `TypeVar`. I've proposed 2 different solutions in the comments below. Both are fully implemented and can be copy pasted.

Sorry, I know this may seem like a minor point but I think we should use this opportunity to do better type narrowing. Also excited to see you take this and run. I'm sure you'll find all sorts of places where we can use these techniques.

Kilo59 (approved):
🚀

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3643,3 +3683,44 @@ def _init_site_builder_for_data_docs_site_creation(
...
+        Args:
+            value: config variable value
+            dollar_sign_escape_string: replaces instances of `$`
+            skip_if_substitution_variable: skip if the value is of the form ${MYVAR} or $MYVAR
+
+        Returns:
+            input value with all `$` characters replaced with the escape string
+        """
+        if isinstance(value, dict) or isinstance(value, OrderedDict):
+            return {  # type: ignore[return-value] recursive call expects str
```
```suggestion
            return {  # type: ignore[return-value] # recursive call expects str
```
This fixes the warning

PR #6412, opened 2022-11-21
[MAINTENANCE] Implement `RendererConfiguration`

NathanFarmer:
- GREAT-1373
- Provide a mechanism to refactor out common logic from each implemented renderer

### Definition of Done

Thank you for submitting!


great_expectations/render/renderer_configuration.py
```diff
@@ -0,0 +1,33 @@
...
+        if self.runtime_configuration:
+            include_column_name = (
+                False
+                if self.runtime_configuration.get("include_column_name") is False
+                else True
+            )
+            styling = self.runtime_configuration.get("styling")
+
+        object.__setattr__(self, "include_column_name", include_column_name)
+        object.__setattr__(self, "styling", styling)
```
NathanFarmer, line 44: This pattern is what is used in `dataclass` constructors ([per the docs](https://docs.python.org/3/library/dataclasses.html#frozen-instances)) and is one of many approaches to setting frozen `dataclass` attributes that aren't allowed in the constructor (outlined [here](https://stackoverflow.com/questions/53756788/how-to-set-the-value-of-dataclass-field-in-post-init-when-frozen-true)).

great_expectations/render/renderer_configuration.py
```diff
@@ -0,0 +1,33 @@
...
+)
+
+
+@dataclass(frozen=True)
+class RendererConfiguration:
+    """Configuration object built for each renderer."""
+
+    configuration: Union[ExpectationConfiguration, None]
+    result: Union[ExpectationValidationResult, None]
+    language: str = "en"
```
NathanFarmer, line 16: `en` is based off of [ISO 639-1 language codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes). An `Enum` based on that list could be valuable. This field is not currently being used anywhere in the open source codebase, but may be in the future. Regardless of GX use, the `language` parameter exists in all renderer signatures, so removing support could break custom renderers.

NathanFarmer, line 16: `en` is based off of [ISO 639-1 language codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes). An `Enum` based on that list could be valuable. This field is not currently being used anywhere in the open source codebase, but may be in the future. If we decide this is not needed right now, it is possible (probable?) that removing the `language` parameter from all existing renderer signatures would not constitute a breaking change, as we don't seem to actually be setting this anywhere.

great_expectations/render/renderer_configuration.py
```diff
@@ -0,0 +1,33 @@
...
+    ExpectationValidationResult,
+)
+
+
+@dataclass(frozen=True)
+class RendererConfiguration:
+    """Configuration object built for each renderer."""
+
+    configuration: Union[ExpectationConfiguration, None]
+    result: Union[ExpectationValidationResult, None]
```
NathanFarmer, line 15: I would prefer for these attributes to be more explicitly named (`expectation_configuration` and `expectation_validation_result`), but this follows the pattern for the existing renderer signatures. In the past it seems we ended up using abbreviations for `expectation_validation_result` (such as `evr` or `result`) in a lot of places to reduce verbosity. Changing these attributes to be more explicitly named would seem to represent a move in that direction everywhere, since I am proposing that this class be instantiated in the contributor-facing renderers themselves for now.

great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
```diff
@@ -222,11 +223,16 @@ def _prescriptive_renderer(
...
+            configuration=configuration,
+            result=result,
+            language=language,
+            runtime_configuration=runtime_configuration,
         )
-        styling = runtime_configuration.get("styling")
+        include_column_name: Union[
+            bool, None
+        ] = renderer_configuration.include_column_name
+        styling: Union[str, None] = renderer_configuration.styling
```
NathanFarmer, line 235: While it looks here like instantiating a `RendererConfiguration` and pulling out its attributes into variables isn't a reduction renderer template code, it is important to note that currently unimplemented functionality (specifically, handling both `configuration` and `result` with type checking in all renderers) is now being implemented by the `RendererConfiguration` class.

great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
```diff
@@ -222,11 +223,14 @@ def _prescriptive_renderer(
...
-            False if runtime_configuration.get("include_column_name") is False else True
+        renderer_configuration = RendererConfiguration(
+            configuration=configuration,
+            result=result,
+            language=language,
+            runtime_configuration=runtime_configuration,
         )
-        styling = runtime_configuration.get("styling")
+        include_column_name: bool = renderer_configuration.include_column_name
+        styling: Union[str, None] = renderer_configuration.styling
```
NathanFarmer, line 233: While it looks here like instantiating a `RendererConfiguration` and pulling out its attributes into variables isn't a reduction renderer template code, it is important to note that currently unimplemented functionality (specifically, handling both `configuration` and `result` with type checking in all renderers) is now being implemented by the `RendererConfiguration` class.

great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
```diff
@@ -222,11 +223,14 @@ def _prescriptive_renderer(
...
-            False if runtime_configuration.get("include_column_name") is False else True
+        renderer_configuration = RendererConfiguration(
+            configuration=configuration,
+            result=result,
+            language=language,
+            runtime_configuration=runtime_configuration,
         )
-        styling = runtime_configuration.get("styling")
+        include_column_name: bool = renderer_configuration.include_column_name
+        styling: Union[dict, None] = renderer_configuration.styling
```
NathanFarmer, line 233: While it looks here like instantiating a `RendererConfiguration` and pulling out its attributes into variables isn't a net reduction renderer template code, it is important to note that currently unimplemented functionality (specifically, handling both `configuration` and `result` with type checking in all renderers) is now being implemented by the `RendererConfiguration` class.

great_expectations/render/renderer_configuration.py
cdkini: yup I agree! I don't love `setattr` but if this allows the object to stay frozen downstream, I'm all for it

great_expectations/render/renderer_configuration.py
cdkini: Let's maybe add a ticket or spike to investigate this? Think we can keep it as-is for now. I'm all for additional points of extensibility but if this isn't going to be used, let's do future us a favor and remove it to minimize any complexity.

cdkini (approved):

great_expectations/render/renderer_configuration.py
donaldheppner: This is a great example of what I call "anticipatory development", anticipating that something might be needed and complicating the design: now it needs to be maintained, it provides no functionality, and when we do go to add multi-language support in the app, there is little chance that the language parameter in renderers is going to be how we do it.

I'm willing to take the risk and remove this parameter: the only impact would be if someone else built custom renderers for their custom expectation that are multi-lingual and rely on this value -- that scenario feels sufficiently unlikely to not worry about this breaking change. We will have to take care on deserialization to not fail should language be present

PR #6414, opened 2022-11-21
[MAINTENANCE] ZEP Multiple Model POC Design

Kilo59:
This is a Proof of concept for breaking ZEP Datasource/DataAssets into 2 different models, a **_runtime_** model and a **_config_** model.

I've stripped out a lot of my original messy implementation so that we can better focus on a couple of key ideas.
This PR is not ready to merge, and I believe we should stick with the existing single-model approach for until (and if) this multi-model approach has had time to bake and becomes the obvious right solution.

### Unavoidable Model Explosion

Moving from a single-model design to a multi-model design will lead to an explosion of models as each new Datasource will need at minimum 4 different model/class definitions instead of only 2.
1. The runtime Datasource class.
2. The config schema Datasource class.
3. At least one runtime DataAsset class
4. At least one config schema DataAsset class.

Our original approach to config schema validation was generic and so we didn't require specific schemas for each type of `Datasource` or `DataAsset`. 

We could mitigate some of this work by using our metaclass hooks to automatically generate config models from runtime models or vice versa, but this is beyond the scope of the initial effort.

### RuntimeModel decorates/wraps the ConfigModel

The runtime model is a wrapper around the underlying config model. The wrapped config models retain all the `pydantic` type safety and other goodness.

We can use `@properties` or other methods to proxy access to the underlying config attributes as needed. (NOTE: these don't need to be defined on the concrete classes)

```python
class FoobarDatasource(Datasource):
   # instance attr
   config: FoobarDatasourceConfig

   ...

   @property
   def execution_engine(self) -> FoobarExecutionEngine:
        return self.config.execution_engine
```

### `asset_factory()` single-dispatch
`@functools.singledispatch`
https://docs.python.org/3/library/functools.html#functools.singledispatchmethod

More single-dispatch examples and explanation
https://rednafi.github.io/digressions/python/2020/04/05/python-singledispatch.html

The current implementation is only an `asset_factory()` but this probably makes more sense as a more general `create_runtime_model_from_config_factory()` method that handles `Datasource` and `DataAsset` runtime/config models.

...


great_expectations/experimental/datasources/multi_model.py
```diff
@@ -0,0 +1,505 @@
...
+        if isinstance(config, DatasourceConfig):
+            self.config = config
+        else:
+            self.config = DatasourceConfig(
+                type=type,
+                name=name,
+                assets=assets,
+                execution_engine=execution_engine,
+                **kwargs,
+            )
```
Kilo59, line 205: ### RuntimeModel decorates/wraps the ConfigModel
We've gone back and forth about whether we want to only accept a `DatasourceConfig` object or allow keyword args.

The important thing is that errors will be raised during `__init__` if things are missing or of the incorrect type.

great_expectations/experimental/datasources/multi_model.py
```diff
@@ -0,0 +1,505 @@
...
+        self.assets: Dict[str, DataAsset] = {}
+
+        # convert the DataAsset configs into runtime objects
+        for asset_name, asset_config in self.config.assets.items():
+            self.assets[asset_name] = self.asset_factory(asset_config, self)  # type: ignore[arg-type] # mypy confused about singledispatch instance method?
+
+    @functools.singledispatch
+    def asset_factory(self, asset_config: DataAssetConfig) -> DataAsset:
+        """Creates runtime asset objects from asset config models."""
+        raise NotImplementedError("No single dispatch functions have been registered")
```
Kilo59, line 217: ### asset_factory() single-dispatch

Now we have to create the correct runtime asset type according to the asset config models.
Rather than having to maintain another mapping or lookup table. Each asset type registers a `@single_dispatch` factory method that accepts a particular `ConfigModel` and returns a `RuntimeModel`.

great_expectations/experimental/datasources/multi_model.py
```diff
@@ -0,0 +1,505 @@
...
+
+
+# Assets must register an asset factory
+# Maybe this is handled by a new/updated Metaclass
+@Datasource.asset_factory.register(TableAssetConfig)
+def _table_asset_factory(  # type: ignore[misc] # mypy confused about self & singledispatch?
+    self: PostgresDatasource,
+    asset_config: TableAssetConfig,
+) -> TableAsset:
+    return TableAsset(config=asset_config, datasource=self)
```
Kilo59, line 428: DataAsset implementations must include a single-dispatch factory method registration step.

great_expectations/experimental/datasources/multi_model.py
```diff
@@ -0,0 +1,505 @@
...
+            )
+        self.datasource = datasource
+
+    def get_batch_request(self, options: Optional[BatchRequestOptions]) -> BatchRequest:
+        raise NotImplementedError
+
+
+class DatasourceConfig(
+    ExperimentalBaseModel,
+    # metaclass=SomeDifferentMetaclass,
```
Kilo59, line 87: This has not been implemented yet but this new Metaclass would need to perform a similar role as the existing `Metadataclass`.
At minimum it would need to register all the `Config` models and map them with the correct `type: str`.
Example: `type` ->  `'postgres': PostgresDatasourceConfig`

It could do other actions as well like...
1. Register the single_dispatch factory methods.
2. Generate a runtime model from a config model (or vice versa).
3. ???

Kilo59:
We aren't adopting this right now. Perhaps in the future

PR #6425, opened 2022-11-23
[CONTRIB] Add number of unique values expectation

itaise:
Hi,
We would like to contribute this expectation. (It is already in use by our users in some non-GX solution)
The expectation check that under some condition (e.g. - is_open_row = True), the number of distinct values in a column is some specific number (e.g. - only 1 distinct value of "end date" - 2099-01-01). 

Ca you please review? thanks a lot! 

CC: @Shinnnyshinshin  - this one is also using the new query metric :)


kenwade4 (approved):

kenwade4:

PR #6429, opened 2022-11-24
[CONTRIB] Feature/add two tables expectation

itaise:
Hi, 
I would like to contribute this expectation. we are already using it as a custom expectation and its very useful for us. 
It gets a template_dict with a column name to check in 2 tables, and a second table name, and checks that the values in the first column exist in the second one. 
This enables comparison of 2 tables in a "left join where null" manner. 
The expectation also gets a condition, which is added to the where clause. 
Can you please review? 
Thanks a lot!


kenwade4 (approved):

kenwade4:

PR #6433, opened 2022-11-27
[CONTRIB] add expectation - check gaps in SCD tables

itaise:
Hi, 
I would like to contribute the following expectation, it is already in use by our users. 
The expectation is intended for slowly changing tables of type II - which have "open date" and "close date" columns. 
The expectation checks that there are no gaps between the close_date and the next open_date, for each entity (e.g. a user in a users table). 
  
Can you please review? 
thanks a lot!


kenwade4 (approved):

kenwade4:

PR #6329, opened 2022-11-09
[FEATURE] Support to include ID/PK in validation result for each row - Pandas

Shinnnyshinshin:
Changes proposed in this pull request:

Introduction of `unexpected_index_columns` value that allows users to specify a primary key (PK) column for identifying rows that failed an Expectation (usually returned as part of the `unexpected_index_list`).
 
- Changes were made to the `map_metric_provider` to take in the parameter from `result_format` and output `unexpected_index_list` as key-value pairs of the primary key column. 

* **Note** This is only the `Pandas` implementation, with `Spark` and `sql` to follow. 

## Can you give me an Example? 
Given the following DataFrame: 

```python 
df = pd.DataFrame(
        {
            "pk_1": [0, 1, 2, 3, 4, 5],
            "pk_2": ["zero", "one", "two", "three", "four", "five"],
            "numbers_with_duplicates": [1, 5, 22, 3, 5, 10],
            "animal_names_no_duplicates": [
                "cat",
                "fish",
                "dog",
                "giraffe",
                "lion",
                "zebra",
            ],
        }
    )

```

We could run the `expect_column_values_to_be_in_set` Expectation on the `numbers_with_duplicates` column (which has `[1, 5, 22, 3, 5, 10]` as its values) with  `[1, 5, 22]` as the `value_set`.

```python
    expectationConfiguration = ExpectationConfiguration(
        expectation_type="expect_column_values_to_be_in_set",
        kwargs={
            "column": "numbers_with_duplicates",
            "value_set": [1, 5, 22],
        },
    )
```

After running the ExpectationConfiguration, we would expect the `unexpected_index_list` to be `[3, 5]` which correspond to the indices of `3` and `10`, the values  that are not in the `value_set`. 

This PR enables the following configuration, which sets `unexpected_index_columns` to be `pk_2`. 

```python

    expectationConfiguration = ExpectationConfiguration(
        expectation_type="expect_column_values_to_be_in_set",
        kwargs={
            "column": "numbers_with_duplicates",
            "value_set": [1, 5, 22],
            "result_format": {
                "unexpected_index_columns": ["pk_2"],
            },
        },
    )
```

After running this new ExpectationConfiguration, we expect the `unexpected_index_list` to be `[{"pk_2": "three"}, {"pk_2": "five"}]` which correspond to the values in the `pk_2` column of the indices of `22` and `10` the `numbers_with_duplicates` column. 

**Note**
[...]

Issue #3195, opened 2021-08-09
Support to include ID/PK in validation result for each row that failed an expectations


KentonParton:
**Is your feature request related to a problem? Please describe.**
When an expectation is run the output includes a “partial_unexpected_list” property of values that were unexpected. While this is useful, in most cases, it doesn't allow teams to identify, resolve, or divert data with poor quality.

It would be great if a sample list or a complete list (result_format SUMMARY, COMPLETE) of ID's for each row that failed an expectation was included in the validation result.

**Describe the solution you'd like**
One could include a column name as an argument in the expectation that they would like to be included in the "partial_unexpected_list" (or a new property).

**Describe alternatives you've considered**
One could try infer the PK for a table but this is not possible for all engine types E.g. Spark. 

**Additional context**
Enabling teams to not only bring light to data quality issues but identify all rows allows them to address poor data quality in real-time instead of requiring manual intervention.

OmarSultan85:
This is a very important feature in my opinion, as Kenton mentioned, we will be able to not only identify the data quality and correctness on the data, but take action as well and help isolate erroneous rows rather than failing the entire dataset.

talagluck:
Thanks for submitting this issue, @KentonParton, and the follow-up, @OmarSultan85 ! We will discuss internally, and I will get back to you in the next few days.

talagluck:
Hi @KentonParton and @OmarSultan85 - apologies for delays! A part of the V3 API was making sure that this functionality was available, but @jcampbell and I worked on this draft PR (#3346) to link up all the pieces and get it working. When you have a chance, we'd love your thoughts on the interface of it, concerns, etc.

Thanks!

OmarSultan85:
Hi @talagluck ,

Thanks a lot for the update, I checked the PR but I am not entirely sure if I understood correctly. What I understand is that now the list of unexpected rows will be returned completely as part of the validation result when its of type COMPLETE.

If that is the case then I believe this would be of great use and would allows us to use great expectations in a pipeline that could split the incoming data into two, process correct rows and insert them and for the unexpected ones perform some kind of action  to clean the data, notify business users, etc.

Anyway I can try this new feature before it is released? I can't wait to integrate it into our product and pipelines :)

talagluck:
Thanks so much for the feedback, @OmarSultan85 ! That is indeed the purpose of the feature, and I'm glad to hear that that will work for you!

[...]

PR #6448, opened 2022-11-29
[FEATURE] Support to include ID/PK in validation result for each row - SQL

Shinnnyshinshin:

SQL implementation of `unexpected_index_column`s value that allows users to specify a primary key (PK) column for identifying rows that failed an Expectation (usually returned as part of the `unexpected_index_list`).  The PR also enables `unexpected_index_query` to be returned to the user, which 

Changes were made to the `map_metric_provider` to take in the parameter from `result_format` and output `unexpected_index_list` as key-value pairs of the primary key column.

Note : This is only the SQL implementation, [Pandas has been merged already](https://github.com/great-expectations/great_expectations/pull/6329), and Spark to follow.

* (Almost) closes `#3195`

### What has changed?
* `unexpected_index_list` and `unexpected_index_query` was added as validation dependencies for `SqlAlchemyExecutionEngine`
* `_sqlalchemy_map_condition_index` was added as MapMetric.
  - This Metric returns the indices (truncated at 10 results for SQL) that contain unexpected values, according to the Primary Key values specified using `unexpected_index_column_names`. **Note** Unlike Pandas, the current SQL implementation does not have a default index that will be returned if `unexpected_index_column_names` is not specified (In other words, in order to see `unexpected_index_list`, the user *must* specify `unexpected_index_column_names`)
  
* `_sqlalchemy_map_condition_query` was added as MapMetric. 
  - This Metric returns the query that can be used to retrieve the full unexpected results.
  - This method depends on two helper methods :
    - `sql_post_compile_to_string()` :   Used by the `_sqlalchemy_map_condition_query()` method to compile SQL select statement with post-compile parameters into a string. Logic lifted directly from [the sqlalchemy documentation](https://docs.sqlalchemy.org/en/14/faq/sqlexpressions.html#rendering-postcompile-parameters-as-bound-parameters) documentation.
    -  `get_sqlalchemy_source_table_and_schema_selectable()`: Used by `_sqlalchemy_map_condition_query()` metric function to return the table associated with the current `Batch`, rather than the `temp_table` that is created as part of running the query. 

### Can you give me an Example? (heavily adopted from the [Pandas example](https://github.com/great-expectations/great_expectations/pull/6329))

Given the following table `animal_names`

```python 
sqlite_path = file_relative_path(
       __file__, "../../test_sets/metrics_test.db"
    )
sqlite_engine = sa.create_engine(f"sqlite:///{sqlite_path}")
df = pd.DataFrame(
       {
           "pk_1": [0, 1, 2, 3, 4, 5],
           "pk_2": ["zero", "one", "two", "three", "four", "five"],
            "animals": [
               "cat",
               "fish",
               "dog",
               "giraffe",
                "lion",
                "zebra",
            ],
        }
    )
df.to_sql(
     name="animal_names",
     con=sqlite_engine,
     index=False,
     if_exists="replace",
 )
```

We could run the `expect_column_values_to_be_in_set` Expectation on the `animals` column  with  `["cat", "fish", "dog"]` as the `value_set` (ie domestic animals).

```python
expectationConfiguration = ExpectationConfiguration(
    expectation_type="expect_column_values_to_be_in_set",
    kwargs={
        "column": "animals",
        "value_set": ["cat", "fish", "dog"],
        "result_format": {
            "result_format": "COMPLETE",
        },
    },
)
```

After running the ExpectationConfiguration, we would expect the `unexpected_index_list` to be ` ["giraffe", "lion", "zebra"]` which correspond to the indices of `4`, `5`, and `6`, the values  that are not in the `value_set`. 

This PR enables the following configuration, which sets `unexpected_index_columns` to be `pk_1`. 
```python
expectationConfiguration = ExpectationConfiguration(
    expectation_type="expect_column_values_to_be_in_set",
    kwargs={
        "column": "animals",
        "value_set": ["cat", "fish", "dog"],
        "result_format": {
            "result_format": "COMPLETE",
        },
    },
)
```
After running this new ExpectationConfiguration, we expect the `unexpected_index_list` to be `[{"pk_1": 3}, {"pk_1": 4}, {"pk_1": 5}]` which correspond to the values in the `pk_1` column of the indices of `["giraffe", "lion", "zebra"]`.

#### What if I have a **lot** of rows in my table?

In order to retrieve the *full list* unexpected values from the table, the result also contains a `unexpected_index_query`, which can be copied into a db client to retrieve **all** the unexpected rows.
 
```sql
SELECT animals, pk_1 
FROM animal_names 
WHERE animals IS NOT NULL AND (animals NOT IN "
  "('cat', 'fish', 'dog'))
```

The query (along with `unexpected_index_list`) is returned as part of the result values returned by the `Validator`. 

```python 
...
"unexpected_index_list": [{"pk_1": 3}, {"pk_1": 4}, {"pk_1": 5}],
"unexpected_index_query": "SELECT animals, pk_1 \n"
"FROM animal_names \n"
  "WHERE animals IS NOT NULL AND (animals NOT IN "
  "('cat', 'fish', 'dog'))",
...
```

### Definition of Done

Thank you for submitting!


great_expectations/expectations/metrics/map_metric_provider.py
```diff
@@ -2329,6 +2336,138 @@ def _sqlalchemy_map_condition_rows(
...
+        )
+    )
+    query_result: List[tuple] = execution_engine.engine.execute(final_query).fetchall()
+
+    unexpected_index_list: Optional[List[Dict[str, Any]]] = []
+
+    # since SQL tables can be **very** large, truncate query_result values at 20, or at `partial_unexpected_count`
+    truncate: int = result_format.get(["partial_unexpected_count"], 20)
+
+    truncated_query_result = query_result[:truncate]
```
Shinnnyshinshin: Here is where we filter the `unexpect_index_list` results, so that we aren't passing **huge** lists around.

great_expectations/expectations/metrics/map_metric_provider.py
```diff
@@ -2329,6 +2336,134 @@ def _sqlalchemy_map_condition_rows(
...
+        domain_records_as_selectable: Union[
+            sa.Table, sa.Select
+        ] = get_sqlalchemy_selectable(domain_records_as_selectable)
+
+    # since SQL tables can be **very** large, truncate query_result values at 20, or at `partial_unexpected_count`
+    final_query: sa.select = (
+        unexpected_condition_query_with_selected_columns.select_from(
+            domain_records_as_selectable
+        ).limit(result_format.get("partial_unexpected_count", 20))
+    )
```
Shinnnyshinshin: Here is where we filter the `unexpected_index_list` results to either `partial_unexpected_count or 20`, so that we aren't passing huge lists around.

great_expectations/expectations/metrics/util.py
```diff
@@ -896,3 +898,53 @@ def is_valid_continuous_partition_object(partition_object):
...
+        engine (sqlalchemy.engine.Engine): Sqlalchemy engine used to do the compilation.
+        select_statement (sqlalchemy.sql.Select): Select statement to compile into string.
+    Returns:
+        String representation of select_statement
+    """
+    compiled = select_statement.compile(
+        engine, compile_kwargs={"render_postcompile": True}
+    )
+    params = (repr(compiled.params[name]) for name in compiled.positiontup)
+    query_as_string = re.sub(r"\?", lambda m: next(params), str(compiled))
```
anthonyburdi: I noticed in the sqlalchemy docs https://docs.sqlalchemy.org/en/14/faq/sqlexpressions.html#rendering-postcompile-parameters-as-bound-parameters that this might only work with sqlite because of this statement `The SQLCompiler.params and SQLCompiler.positiontup are also compatible with render_postcompile, so that the previous recipes for rendering inline bound parameters will work here in the same way, such as SQLite’s positional form:` - would be worth looking into other sql backends to make sure if you haven't already.

tests/expectations/metrics/test_map_metric.py
```diff
@@ -1148,3 +1148,694 @@ def test_sqlite_single_column_summary_result_format(
...
+def test_sql_multiple_unexpected_index_column_names_basic_result_format(
+    sa, sqlite_table_for_unexpected_rows_with_index
+):
+    expectationConfiguration = ExpectationConfiguration(
+        expectation_type="expect_column_values_to_be_in_set",
+        kwargs={
+            "column": "animals",
+            "value_set": ["cat", "fish", "dog"],
+            "result_format": {
+                "result_format": "BASIC",  # SUMMARY will include partial_unexpected_list only, which means unexpected_index_column_names will have no effect
```
anthonyburdi, line 1456: This would be worth discussing with @tal, I would have expected a partial_unexpected_index_list to be included if unexpected_index_column_names are specified. Is it not because of the result_format being BASIC?

tests/expectations/metrics/test_map_metric.py
```diff
@@ -1148,3 +1148,694 @@ def test_sqlite_single_column_summary_result_format(
...
+    )
+    assert len(batch_definition_list) == 1
+    batch_spec: SqlAlchemyDatasourceBatchSpec = my_data_connector.build_batch_spec(
+        batch_definition=batch_definition_list[0]
+    )
+    batch_data, batch_markers = execution_engine.get_batch_data_and_markers(
+        batch_spec=batch_spec
+    )
+    batch = Batch(data=batch_data)
+    validator = Validator(execution_engine, batches=[batch])
```
anthonyburdi: Some of this setup code (in this and other tests) could be moved to a fixture to reduce duplication. If there are minor changes to the setup maybe it could be a parametrized fixture.

great_expectations/expectations/metrics/util.py
anthonyburdi: The test for this could just be for this method to make sure we get the expected query, no need to set up a data connector, validator etc. There are some examples in the splitting / sampling tests, e.g. in test_sqlalchemy_execution_engine_sampling.py this fixture: dialect_name_to_sql_statement and this test test_sample_using_limit_builds_correct_query_where_clause_none

tests/expectations/metrics/test_map_metric.py
Shinnnyshinshin: Great suggestion. I'll get on this :)

tests/expectations/metrics/test_map_metric.py
tal: By definition you can't have an expected element in the unexpected list.

Given the nature of this I'd also expect the `result_format` to be `FOURTRAN` instead of `BASIC`.

All in i think this is completely out of bounds and you should be ashamed.

Shinnnyshinshin: Thank you @tal :) Your input is tremendously valuable

great_expectations/expectations/metrics/util.py
Shinnnyshinshin: This test has been added in the most recent commit and can be found at `tests/expectations/metrics/test_metrics_util.py`

great_expectations/expectations/metrics/util.py
```diff
@@ -992,3 +994,70 @@ def is_valid_continuous_partition_object(partition_object):
...
+    else:
+        params = (repr(compiled.params[name]) for name in list(compiled.params.keys()))
+        query_as_string = re.sub(r"%\(.*?\)s", lambda m: next(params), str(compiled))
+
+        # bigquery inserts extra '`' character for compiled statement.
+        # clean up string before returning
+        if dialect_name == "bigquery":
+            query_as_string = re.sub(r"`", "", query_as_string)
+
+    return query_as_string
```
Shinnnyshinshin: Return SQL query according to backend

great_expectations/expectations/metrics/util.py
```diff
@@ -992,3 +994,70 @@ def is_valid_continuous_partition_object(partition_object):
...
+        params = (repr(compiled.params[name]) for name in compiled.positiontup)
+        query_as_string = re.sub(r"\?", lambda m: next(params), str(compiled))
+
+    else:
+        params = (repr(compiled.params[name]) for name in list(compiled.params.keys()))
+        query_as_string = re.sub(r"%\(.*?\)s", lambda m: next(params), str(compiled))
+
+        # bigquery inserts extra '`' character for compiled statement.
+        # clean up string before returning
+        if dialect_name == "bigquery":
```
Shinnnyshinshin: TODO: adjust test to compare the query from bigquery, rather than adjust the query from bigquery to match the test

anthonyburdi (approved):
LGTM! Thank you for the synchronous review and for adding the tests for param substitution for each backend 🙇

PR #6452, opened 2022-11-30
[MAINTENANCE] Update `RendererConfiguration` to `pydantic` model

NathanFarmer:
- GREAT-1299
- Update `RendererConfiguration` to `pydantic` model
- Handles formatting the json schema for the implementer
- Enables dot notation on params instead of using dictionaries

### Definition of Done


great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
```diff
@@ -141,50 +143,51 @@ def _atomic_prescriptive_template(
         configuration: Optional[ExpectationConfiguration] = None,
         result: Optional[ExpectationValidationResult] = None,
         runtime_configuration: Optional[dict] = None,
-        **kwargs,
-    ):
-        runtime_configuration = runtime_configuration or {}
-        include_column_name = (
-            False if runtime_configuration.get("include_column_name") is False else True
+    ) -> Tuple[str, dict, Union[dict, None]]:
+        renderer_configuration = RendererConfiguration(
```
NathanFarmer, line 144: In the near-future state (once we update the superclass `Expectation`) these atomic "template" methods will be passed an already instantiated `RendererConfiguration` instead of the current params and then immediate instantiation of a `RendererConfiguration`. This is a private method and would not be a breaking change. It is also possible this will become a `staticmethod` if the super can also become one.

great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
```diff
@@ -193,23 +196,19 @@ def _atomic_prescriptive_template(
...
-            params=params,
-            params_with_json_schema=params_with_json_schema,
-            param_key_with_list="value_set",
-        )
-
-        return (template_str, params_with_json_schema, styling)
+        return template_str, params_with_json_schema, renderer_configuration.styling
 
     @classmethod
     @renderer(renderer_type=LegacyRendererType.PRESCRIPTIVE)
```
NathanFarmer, line 211: This legacy renderer isn't actually leveraging the latest `RendererConfiguration` functionality, but I went ahead and cleaned it up while I was in this file.

great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
```diff
@@ -141,50 +143,51 @@ def _atomic_prescriptive_template(
...
-                "value": params.get("condition_parser"),
-            },
-        }
 
-        if params["value_set"] is None or len(params["value_set"]) == 0:
+        params: RendererParams
+        params = renderer_configuration.params
 
-            if include_column_name:
+        if not params.value_set or len(params.value_set.value) == 0:
```
NathanFarmer, line 169: `RendererConfiguration.add_param()` will always add a new attribute to `RendererConfiguration.params`. In order to make the truthiness of a given attribute meaningful (not always `True`), we overload the `__bool__` and check the truthiness of `.value`. If `params.value_set.value` is falsy or 0 this conditional would evaluate to `True`. The `.value` could be appended to `params.value_set` and the behavior would remain the same.

great_expectations/expectations/core/expect_column_values_to_be_in_set.py
```diff
@@ -193,93 +196,87 @@ def _atomic_prescriptive_template(
...
+            for i, v in enumerate(params.value_set.value):
+                renderer_configuration.add_param(
+                    name=f"v__{str(i)}", schema_type="string", value=v
+                )
+
+            params = renderer_configuration.params
 
             values_string = " ".join(
-                [f"$v__{str(i)}" for i, v in enumerate(params["value_set"])]
+                [f"$v__{str(i)}" for i, v in enumerate(params.value_set.value)]
```
NathanFarmer, line 242: Whenever we want to do something with the `.value` other than evaluate it's truthiness, we need to use the `.value` syntax. This is because the `.value` could be anything (it's often an Expectation kwarg), and it's not advisable to overload every dunder method imaginable.

great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
NathanFarmer: This behavior was changed in the subsequent PR. The entire `param` will be `None` if the value is not found so that it will not be serialized to `rendered_content`.

billdirks (approved):

great_expectations/render/renderer_configuration.py
```diff
@@ -1,43 +1,164 @@
...
+    class Config:
+        validate_assignment = True
+        arbitrary_types_allowed = True
+
+    def __init__(self, **kwargs):
+        kwargs = RendererConfiguration._set_expectation_type_and_kwargs(kwargs=kwargs)
+        kwargs = RendererConfiguration._set_include_column_name_and_styling(
+            kwargs=kwargs
+        )
+        super().__init__(**kwargs)
```
Kilo59, line 72: @NathanFarmer have you looked at doing this with a custom validator?
https://docs.pydantic.dev/usage/validators/

great_expectations/render/renderer_configuration.py
```diff
@@ -1,43 +1,164 @@
...
+
+    class Config:
+        validate_assignment = True
+        arbitrary_types_allowed = True
+
+    def __init__(self, **kwargs):
+        kwargs = RendererConfiguration._set_expectation_type_and_kwargs(kwargs=kwargs)
+        kwargs = RendererConfiguration._set_include_column_name_and_styling(
+            kwargs=kwargs
+        )
```
Kilo59, line 71: `non-blocking`
@NathanFarmer have you looked at doing this with a custom validator?
https://docs.pydantic.dev/usage/validators/

Kilo59 (approved):

great_expectations/render/renderer_configuration.py
NathanFarmer: @Kilo59 I did look into it, but I struggled to get it to work since these attribute values depend on the other inputs.

PR #6458, opened 2022-11-30
[BUGFIX] Patch issue with call to `ExpectationAnonymizer` to ensure `DataContext` init events are captured

cdkini:
- The `anonymize` method of `ExpectationAnonymizer` explicitly references an `obj` arg (as well as `**kwargs`)
    - While downstream logic relies on `obj`, we currently have a call that passes in the suite as a kwarg
    - This results in the method blowing up, causing the payload to become malformed.
- This patch resolves an issue that has been present since `0.14.13`

### Definition of Done

### DIFF
--- a/great_expectations/core/usage_statistics/anonymizers/expectation_anonymizer.py
+++ b/great_expectations/core/usage_statistics/anonymizers/expectation_anonymizer.py
@@ -16,9 +16,6 @@ class ExpectationSuiteAnonymizer(BaseAnonymizer):
     def anonymize(
         self, obj: Optional["ExpectationSuite"] = None, **kwargs  # noqa: F821
     ) -> dict:
-        if obj is None:
-            raise ValueError("Cannot anonymize a null payload")
-
         expectation_suite = obj
         anonymized_info_dict = {}
         anonymized_expectation_counts = list()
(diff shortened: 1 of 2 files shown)


great_expectations/core/usage_statistics/anonymizers/expectation_anonymizer.py
```diff
@@ -16,6 +16,9 @@ def __init__(
     def anonymize(
         self, obj: Optional["ExpectationSuite"] = None, **kwargs  # noqa: F821
     ) -> dict:
+        if obj is None:
+            raise ValueError("Cannot anonymize a null payload")
+
```
cdkini: This signature is what is required by the base class - I know this feels a bit funky but this should all be cleaned up shortly.

cdkini: Downstream logic depends on this being non-null

NathanFarmer: But shouldn't this just use the logger? Don't we want to avoid usage stats raising any type of error so the payload fails silently?

NathanFarmer: Or just put all logic that depends on `obj` inside this conditional?

cdkini: so we're already hitting an `AttributeError` here but yeah maybe we shouldn't raise any explicit errors

NathanFarmer (approved):

PR #6467, opened 2022-12-01
[BUGFIX] Add additional error checking to `ExpectationAnonymizer`

cdkini:
- Since the `ExpectationAnonymizer`'s `anonymize` method depends on a non-null object, we should add a check and exit early if faced with a null value.

### Definition of Done

### DIFF
--- a/great_expectations/core/usage_statistics/anonymizers/expectation_anonymizer.py
+++ b/great_expectations/core/usage_statistics/anonymizers/expectation_anonymizer.py
@@ -16,14 +16,19 @@ class ExpectationSuiteAnonymizer(BaseAnonymizer):
     def anonymize(
         self, obj: Optional["ExpectationSuite"] = None, **kwargs  # noqa: F821
     ) -> dict:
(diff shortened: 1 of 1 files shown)


jcampbell:
@cdkini the thing I observed was an expectation suite that had no expectations in it (created by the temporary profiler case). Not sure how that happened, but I don't think this would catch that (which was an attribute error on the `expectation_suite.expectations` line)

great_expectations/core/usage_statistics/anonymizers/expectation_anonymizer.py
```diff
@@ -17,6 +17,9 @@ def anonymize(
         self, obj: Optional["ExpectationSuite"] = None, **kwargs  # noqa: F821
     ) -> dict:
         expectation_suite = obj
+        if not expectation_suite:
+            return {}
```
cdkini, line 22: Falsy check for both null and empty cases

NathanFarmer (approved):

PR #6472, opened 2022-12-01
Add pretty representations for zep pydantic models

billdirks:
This allows one to do `datasource.yaml` to get a yaml string, `datasource.json` to get a json string, `print(datasource)` to get the yaml string, and `datasource.dict()` to get a dictionary representation.

Since some attributes are private and made available via properties, I hit this [pydantic issue](https://github.com/pydantic/pydantic/issues/935). I use a workaround found in that ticket.

Thank you for submitting!

### DIFF
--- a/great_expectations/experimental/datasources/experimental_base_model.py
+++ b/great_expectations/experimental/datasources/experimental_base_model.py
@@ -68,7 +68,7 @@ class ExperimentalBaseModel(pydantic.BaseModel):
     def __str__(self):
         return self.yaml()
 
-    ##### Workaround for https://github.com/pydantic/pydantic/issues/935 ####
+    # Workaround for https://github.com/pydantic/pydantic/issues/935
     #
     # Currently there is no way to output properties using pydantics builtin in json and dict
     # methods. Our current fix does let one use the exclude/include arguments on properties,
@@ -78,22 +78,24 @@ class ExperimentalBaseModel(pydantic.BaseModel):
     # https://github.com/pydantic/pydantic/issues/935#issuecomment-554378904
 
     @classmethod
-    def get_properties(cls) -> List[str]:
+    def _get_properties(cls) -> List[str]:
         return [
             prop for prop in cls.__dict__ if isinstance(cls.__dict__[prop], property)
         ]
 
     def json(self, **kwargs) -> str:
+        """The json representation of the Model."""
         self.__dict__.update(
-            {prop: getattr(self, prop) for prop in self.get_properties()}
+            {prop: getattr(self, prop) for prop in self._get_properties()}
         )
         attribs = super().json(**kwargs)
         return attribs
 
-    def dict(self, **kwargs) -> "DictStrAny":
+    def dict(self, **kwargs) -> DictStrAny:
+        """The dict representation of the Model."""
         self.__dict__.update(
-            {prop: getattr(self, prop) for prop in self.get_properties()}
+            {prop: getattr(self, prop) for prop in self._get_properties()}
         )
         return super().dict(**kwargs)
 
-    ##### End Workaround #####
+    # End Workaround
--- a/great_expectations/experimental/datasources/experimental_base_model.py
+++ b/great_expectations/experimental/datasources/experimental_base_model.py
@@ -5,10 +5,13 @@ import logging
 import pathlib
 from io import StringIO
 from pprint import pformat as pf
-from typing import Type, TypeVar, Union, overload
+from typing import TYPE_CHECKING, List, Type, TypeVar, Union, overload
 
 import pydantic
 from ruamel.yaml import YAML
+
+if TYPE_CHECKING:
+    from pydantic.typing import DictStrAny
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -35,8 +35,8 @@ class BatchRequestError(Exception):
 
 # For our year splitter we default the range to the last 2 year.
 _CURRENT_YEAR = datetime.now(dateutil.tz.tzutc()).year
-_DEFAULT_YEAR_RANGE = range(_CURRENT_YEAR - 1, _CURRENT_YEAR + 1)
-_DEFAULT_MONTH_RANGE = range(1, 13)
+_DEFAULT_YEAR_RANGE = list(range(_CURRENT_YEAR - 1, _CURRENT_YEAR + 1))
```
billdirks, line 38: Making these lists means they can be represented in json and that makes them reusable. That is, we hit a bug where the range iterator would "expire" and return empty values after going once through the iterator.

great_expectations/experimental/datasources/experimental_base_model.py
```diff
@@ -61,3 +64,36 @@ def yaml(
...
+        return self.yaml()
+
+    ##### Workaround for https://github.com/pydantic/pydantic/issues/935 ####
+    #
+    # Currently there is no way to output properties using pydantics builtin in json and dict
+    # methods. Our current fix does let one use the exclude/include arguments on properties,
+    # we always include them. We could fix this by changing the json and dict signatures
+    # to `def dict(self, *, include, exclude, **kwargs)` and add the implementation.
+    # See this comment for one way to do this:
+    # https://github.com/pydantic/pydantic/issues/935#issuecomment-554378904
```
cdkini: Do we have an idea as to when this will be fixed? I'm okay with this since it is made pretty clear.

great_expectations/experimental/datasources/experimental_base_model.py
```diff
@@ -61,3 +64,36 @@ def yaml(
...
+        ]
+
+    def json(self, **kwargs) -> str:
+        self.__dict__.update(
+            {prop: getattr(self, prop) for prop in self.get_properties()}
+        )
+        attribs = super().json(**kwargs)
+        return attribs
+
+    def dict(self, **kwargs) -> "DictStrAny":
```
cdkini: Do we need quotes here? Could we do `from __future__ import annotations` and avoid this?

great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -35,8 +35,8 @@ class BatchRequestError(Exception):
...
+_DEFAULT_MONTH_RANGE = list(range(1, 13))
```
cdkini, line 39: Does the actual value matter here? Could we just do `list(range(12))`?

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -467,3 +467,48 @@ def test_sort_batch_list_by_unknown_key(create_source):
...
+        )
+        asset = source.add_table_asset(name="my_asset", table_name="my_table")
+        asset.add_year_and_month_splitter(column_name="my_col").add_sorters(
+            ["year", "month"]
+        )
+        source_str = source.__str__()
+        assert "order_by:" in source_str
+
+
+def test_datasource_dict_has_properties(create_source):
```
cdkini, line 528: Should all these tests be decorated with `@pytest.mark.unit`?

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -467,3 +467,48 @@ def test_sort_batch_list_by_unknown_key(create_source):
...
+            type(TableAsset.order_by) == property,
+            "This test assumes TableAsset.order_by is a property. If it is not we "
+            "should update this test",
+        )
+        asset = source.add_table_asset(name="my_asset", table_name="my_table")
+        asset.add_year_and_month_splitter(column_name="my_col").add_sorters(
+            ["year", "month"]
+        )
+        source_dict = source.dict()
+        assert type(source_dict["assets"]["my_asset"]["order_by"]) == list
```
cdkini, line 540: Should `dict()` return a `TypedDict`? Perhaps a violation of the parent `pydantic` impl but if we know some semblance of structure, it would be good to capture that in a rich type.

cdkini (approved):

tests/experimental/datasources/test_postgres_datasource.py
Kilo59: I would say no, the model instance itself is the rich type.
We shouldn't be passing around the untyped `.dict()` representation.

great_expectations/experimental/datasources/experimental_base_model.py
billdirks: This bug was filed in 2019 and I think it will be a while before a change is accepted.

It's not closed because people seem to want it. A maintainer is worried about it being a backward compatibility breaking change to turn it on be default and some difficulties with how properties vs attributes are "stored" on the module. He does hint at the solution that I took from the comment thread which I used:
https://github.com/pydantic/pydantic/issues/935#issuecomment-546231002

great_expectations/experimental/datasources/experimental_base_model.py
billdirks: Done.

I was using quotes to match the annotation of the function I'm overriding. I don't think I really need them and have removed them.

great_expectations/experimental/datasources/postgres_datasource.py
billdirks: This corresponds to how months are represented in a date string so go from 1 through 12 inclusive.

tests/experimental/datasources/test_postgres_datasource.py
billdirks: yes, thank you! 
Done.

tests/experimental/datasources/test_postgres_datasource.py
billdirks: The structure is configurable depending on what you add to the datasource. If there are no sorters this field won't be present.
This is a method on our base pydantic models so it has no notion of the available fields for a particular model.

billdirks (commented):
Thanks for the feedback, it's been integrated. It also made me think of some typing cleanup related to a splitters being a list vs an iterator.

tests/experimental/datasources/test_postgres_datasource.py
billdirks: Yes to what @Kilo59 said. We shouldn't be using the `.dict()` at all in our logic. It's something that may be useful as a representation to look at when playing around on a REPL.

Issue #6468, opened 2022-12-01
[BUG] expect "compound column to be unique" is not scalable


itaise:
Hi, 
We are using GE over Trino, and one of the main tests we use is ["expect compound column to be unique"](https://greatexpectations.io/expectations/expect_compound_columns_to_be_unique?filterType=Backend%20support&gotoPage=1&showFilters=false&viewType=Summary)

However, we found that this expectation (in its sql alchemy implementation) is not scalable enough for us. 
We are running this expectation on large tables, and it fails on insufficient resources - it seems that [this metric](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/expectations/metrics/multicolumn_map_metrics/compound_columns_unique.py#L117) is creating a join between two huge tables. 

Instead, we would like to have something much lighter like: 
```
select count(1) from (
select a,b,c, count(1)
from table
group by a,b,c
having count(1) > 1
)
```

This way it will work without a join, and be able to calculate on huge dwh tables like ours..
I wonder if we should create a new custom ex., or should this one be altered? 

Thanks a lot!

austiezr:
Hey @itaise ! Thanks for raising this. 

We'd love to see a Custom Expectation with this approach; from there, we can test and review to establish whether this core expectation should be replaced / altered, or the Custom Expectation maintained alongside.

itaise:
Hi @austiezr,  I created an expectation that solves the issue for us, and works well with our large tables on Trino, can you please check? thanks a lot!

talagluck:
Resolved by #6473 - thanks @itaise!

PR #6473, opened 2022-12-01
[CONTRIB] Add uniqueness expectation

itaise:
Hi, 
This expectation checks uniqueness of a compound key using a query, and the new metric - "query template values". 
Unlike the other existing expectation (from the core library), it can work with big data volumes, as it does not use a join, as mentioned in this issue: 
https://github.com/great-expectations/great_expectations/issues/6468

I tested it with a trino connector, on a table we have with 1.5 B rows and 5 key columns, and it finish in less than 2 minutes, which is good for us. 

Can you please review?
thanks a lot!


anthonyburdi (approved):
LGTM! Thank you for this PR.

itaise:
Thanks a lot! :)

PR #6482, opened 2022-12-02
Query the database for datetime column splitter defaults

billdirks:
We introduce a new splitter `SqlYearMonthSplitter` which is able to query the database to get the time range for the underlying data (instead of using a hardcoded range). This allows us to validate the data appears in the time range we expect and gives more sensible default batches when querying historic data.

This probably could be generalized to generic datetime splitters instead of just a year and month but we avoided this for now because we'd like to get a demo-able version out asap. We can file a followup ticket for generalizing the splitters.

### DIFF
--- a/great_expectations/experimental/datasources/postgres_datasource.py
+++ b/great_expectations/experimental/datasources/postgres_datasource.py
@@ -3,11 +3,9 @@ from __future__ import annotations
 import copy
 import dataclasses
 import itertools
-from datetime import datetime
 from pprint import pformat as pf
 from typing import TYPE_CHECKING, Dict, List, Optional, Type, Union, cast
 
-import dateutil.tz
 import pydantic
 from pydantic import dataclasses as pydantic_dc
 from typing_extensions import ClassVar, Literal, TypeAlias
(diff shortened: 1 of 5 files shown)


great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -33,23 +31,44 @@ class BatchRequestError(Exception):
...
 
-    @property
-    def param_names(self) -> List[str]:
-        return list(self.param_defaults.keys())
+    def param_defaults(self, data_asset: DataAsset) -> Dict[str, List]:
+        raise NotImplementedError
+
+
+@pydantic_dc.dataclass(frozen=True)
+class SqlYearMonthSplitter(ColumnSplitter):
```
billdirks, line 45: This is the main change in this PR. We now query the database for the ranges over which to compute batches.

As noted in the PR description, we make this very specific for year and month splitters instead of datetime splitters in general so we can demo this asap. We will generalize this when we add more splitters.

great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -33,23 +31,44 @@ class BatchRequestError(Exception):
...
+        assert isinstance(data_asset, TableAsset), "data_asset must be a TableAsset"
+        assert isinstance(
+            data_asset.datasource.execution_engine, SqlAlchemyExecutionEngine
+        )
+        with data_asset.datasource.execution_engine.engine.connect() as conn:
+            # We opt to use a raw string instead of sqlalchemy.txt because we don't
+            # need any of the features for this query
+            # https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.text
+            col = self.column_name
+            q = f"select min({col}), max({col}) from {data_asset.table_name}"
```
billdirks, line 68: We hardcode the query now instead of using more advanced SqlAlchemy constructs. We can generalize this if we need to to support other databases.

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -10,14 +10,21 @@
...
-from tests.experimental.datasources.conftest import sqlachemy_execution_engine_mock_cls
+from tests.experimental.datasources.conftest import (
+    DEFAULT_MAX_DT,
+    DEFAULT_MIN_DT,
+    sqlachemy_execution_engine_mock_cls,
+)
+
+# We set a default time range that we use for testing. We grab the years from the
+# mocks we use in tests. For months, we use every month.
+_DEFAULT_TEST_YEARS = list(range(DEFAULT_MIN_DT.year, DEFAULT_MAX_DT.year + 1))
```
billdirks, line 26: Most of the changes in the file are renaming `_DEFAULT_YEAR_RANGE` to `_DEFAULT_TEST_YEARS` and `_DEFAULT_MONTH_RANGE` to `_DEFAULT_TEST_MONTHS`. I changed this since it is no longer a range but a list and to make it more descriptive.

I'll call out changes that aren't these name changes.

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -132,15 +139,10 @@ def test_construct_table_asset_directly_with_no_splitter(create_source):
...
-        splitter = ColumnSplitter(
-            method_name="splitter_method",
-            column_name="col",
-            param_defaults={"a": [1, 2, 3], "b": list(range(1, 13))},
-        )
         asset = TableAsset(
             name="my_asset",
             table_name="my_table",
-            column_splitter=splitter,
+            column_splitter=SqlYearMonthSplitter(column_name="col"),
```
billdirks, line 145: This is the only non-name change in this PR. I now use the new `SqlYearMonthSplitter` since `ColumnSplitter` is abstract.

great_expectations/experimental/datasources/postgres_datasource.py
cdkini: Should we document our hardcoding of values here or will this be remedied soon after the demo? If it will sit around for a while, I'd ask that we put some of this context in code comments.

great_expectations/experimental/datasources/postgres_datasource.py
cdkini: Same with the return value? Will we get a structured type or will it remain a dictionary?

tests/experimental/datasources/conftest.py
```diff
@@ -12,20 +13,59 @@
...
+            query: The SQL query to execute.
+        """
+        return [(DEFAULT_MIN_DT, DEFAULT_MAX_DT)]
+
+
+class _MockSaEngine:
+    @contextmanager
+    def connect(self):
+        """A contextmanager that yields a _MockConnection"""
+        yield _MockConnection()
```
cdkini, line 50: Very cool - would love to pick your brain regarding mocks and better unit testing philosophies in general

cdkini (approved):

great_expectations/experimental/datasources/postgres_datasource.py
billdirks: Sorry, I misused the term hardcoding. We now make a database query to return the year and month dynamically based on the values in the database. When I said we hardcoded the query, I mean we are issuing a "raw" query where we use a string instead of using sqlalchemy objects to abstract the query. We may or may not change this in the future.

I am going to file a ticket to generalize this splitter to support other datetime splitter methods our current system supports so there will be changes coming to this.

great_expectations/experimental/datasources/postgres_datasource.py
billdirks: This is likely remain a dictionary since this has to conform to the `ColumnSplitter` api which will vary over all splitters. The keys will need to be `str`s but the values can range over all value types on can query over.

Perhaps once we learn more we can impose some structure.

PR #6488, opened 2022-12-05
[MAINTENANCE] Leverage `RendererConfiguration` in existing prescriptive templates (2 of 3)

NathanFarmer:
- GREAT-1373
- Leverage `RendererConfiguration` for many `_atomic_prescriptive_template`s
- Surfaces bugfix where numeric `value_set` is no longer denoted as "string"
- Surfaces bugfix where `idxth_column` was missing in params for substitution
- Surfaces bugfix where `params` with `value: None` are no longer rendered, since they go unused

### Definition of Done


great_expectations/expectations/core/expect_column_mean_to_be_between.py
```diff
@@ -259,85 +266,60 @@ def validate_configuration(
...
-            },
-        }
+        for name, schema_type in add_param_args:
+            renderer_configuration.add_param(name=name, schema_type=schema_type)
 
-        template_str = ""
-        if (params["min_value"] is None) and (params["max_value"] is None):
+        params: RendererParams = renderer_configuration.params
+
+        if not params.min_value and not params.max_value:
```
billdirks, line 287: If `min_value` or `max_value` is numeric (vs a string) then 0 will evaluate to `False` but is still a valid limit.

NathanFarmer: This checks for if `min_value` is a `RendererParam` or `None`. It gets set to `None` if no `value` is passed or found in the `kwargs`. If `min_value` is not `None` and you want to check the `value` you can use `min_value.value`.

NathanFarmer: Setting this to `None` instead of a `RendererParam` gives us the ability to exclude it entirely from the `rendered_content`.

tests/expectations/snapshots/snap_test_expectation_atomic_renderers.py
```diff
@@ -348,12 +330,10 @@
...
             "column": {"schema": {"type": "string"}, "value": "my_column"},
-            "condition_parser": {"schema": {"type": "string"}, "value": None},
-            "row_condition": {"schema": {"type": "string"}, "value": None},
             "ties_okay": {"schema": {"type": "boolean"}, "value": True},
-            "v__0": {"schema": {"type": "string"}, "value": 1},
-            "v__1": {"schema": {"type": "string"}, "value": 2},
-            "v__2": {"schema": {"type": "string"}, "value": 3},
+            "v__0": {"schema": {"type": "number"}, "value": 1},
+            "v__1": {"schema": {"type": "number"}, "value": 2},
+            "v__2": {"schema": {"type": "number"}, "value": 3},
```
billdirks, line 336: Not blocking, just curious how this worked before with the type set to `string`?

great_expectations/expectations/expectation.py
```diff
@@ -179,25 +191,43 @@ def inner_func(
 def param_method(param_name: str) -> Callable:
     """
     Decorator that wraps helper methods dealing with dynamic attributes on RendererConfiguration.params. Ensures a given
-    param_name exists on RendererConfiguration.params before executing the helper method because params are defined by
-    the renderer.
+    param_name exists and is not None before executing the helper method. Params are unknown as they are defined by the
+    renderer, and if a given param is None, no value was set/found that can be used in the helper method.
+
+    If a helper method is decorated with @param_method(param_name="<param_name>") and the param attribute does not
```
billdirks, line 197: This seems like a mistake by the author. Should we `raise` an error in this case?

billdirks (commented):
Thanks for doing this. 2 questions inline (one is more of a curiosity).

tests/expectations/snapshots/snap_test_expectation_atomic_renderers.py
NathanFarmer: Cloud just rendered the number as a string, which is fine as long as they don't need to do any math or use different styling for numbers.

great_expectations/expectations/expectation.py
NathanFarmer: When the decorated `classmethod` is called, the `RendererConfiguration` dynamic param attribute is not guaranteed to exist according to the static type checker (`params` are added at runtime). Whether passing around the `RendererConfiguration` (most simple/elegant) or the relevant `RendererParam` to the `classmethod`, the type checker will still require you to confirm that the `RendererParam` is not `None` somewhere (either in the renderer or the helper method). This decorator simplifies the code in the renderer (no need to check for `None` there) and also ensures that a useful error message is produced by the `classmethod` if the renderer implementer did not use the `add_param` instance method to add the param. The decorated methods are `Expectation` `classmethod`s instead of `RendererConfiguration` instance methods because the helpers exist only for simplifying rendering of common params that may not exist for many Expectations.

billdirks (approved):

PR #6492, opened 2022-12-05
[MAINTENANCE] Message for how to install Great Expectations in Cloud Composer by pinning packages

Shinnnyshinshin:
- Small update to the GCP Reference Architecture doc to add the packages that need to be pinned. 
- Tested to make sure the packages still work on Cloud Composer 1 and 2 (They do as of 12/7/2022)

### Rendering looks like this
Screen Shot 2022-12-07 at 10 45 28 AM


anthonyburdi (approved):
Thank you for this note, looks good to me!

PR #6502, opened 2022-12-07
Update json schema validation on usage stats to filter based on format.

jcampbell:

-
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/core/usage_statistics/schemas.py
```diff
@@ -14,7 +14,7 @@
     InstallEnvironment,
 )
 
-SCHEMA: str = "http://json-schema.org/draft-04/schema#"
+SCHEMA: str = "https://json-schema.org/draft/2020-12/schema"
```
tannerbeam, line 17: i would prefer if we had some method of versioning our oss schema. i don't really care what the values are -- even just integers incremented by 1 -- but i think every time there is a material change to the schema that we should change it. there's also a property value that's referenced in `anonymized_init_payload_schema`: 

```    
"properties": {
        "version": {"enum": ["1.0.0"]},
```
I opened a Jira ticket for this eons ago but I don't think anything has been done on it: 
https://superconductive.atlassian.net/browse/GREAT-808

great_expectations/core/usage_statistics/usage_statistics.py
```diff
@@ -198,7 +198,10 @@ def build_envelope(self, message: dict) -> dict:
     @staticmethod
     def validate_message(message: dict, schema: dict) -> bool:
         try:
-            jsonschema.validate(message, schema=schema)
+            jsonschema.validators.Draft202012Validator(
+                schema=schema,
+                format_checker=jsonschema.validators.Draft202012Validator.FORMAT_CHECKER,
```
tannerbeam, line 203: `format_checker=jsonschema.validators.Draft202012Validator.FORMAT_CHECKER`

i shamefully don't really understand how the whole `jsonschema` thing works. where is this defined? is it visible in the repo? you import the `jsonschema` module, which is presumably installed with pip, and somehow it has the custom validations you defined?

tannerbeam: what happens to the message if validation fails? it gets logged but the message doesn't get written? so for a UUID value that fails we get an empty string instead of str(UUID)?

tannerbeam (commented):
i left one suggestion and a couple of questions

tests/core/usage_statistics/test_schema_validation.py
```diff
@@ -317,4 +353,48 @@ def test_anonymized_rule_based_profiler_validation():
...
     # Empty payload (valid)
     message.clear()
-    jsonschema.validate(message, anonymized_rule_based_profiler_run_schema)
+    jsonschema.validators.Draft202012Validator(
+        schema=anonymized_rule_based_profiler_run_schema,
+        format_checker=jsonschema.validators.Draft202012Validator.FORMAT_CHECKER,
+    ).validate(message)
+
+
+def test_uuid_format_validation():
```
anthonyburdi: This is great, thanks for adding a test. Just a stylistic nit but I would split this into two tests. I won't block the PR on that though.

tests/profile/test_jsonschema_profiler.py
```diff
@@ -9,7 +9,7 @@
 def simple_schema():
     return {
         "$id": "https://example.com/address.schema.json",
-        "$schema": "http://json-schema.org/draft-07/schema#",
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
```
anthonyburdi: Could we use `schemas.SCHEMA` here instead of the string? Just to avoid changing in many places in the future.

tests/core/usage_statistics/test_schema_validation.py
```diff
@@ -18,32 +18,47 @@
 
 def test_anonymized_name_validation():
     string = "aa41efe0a1b3eeb9bf303e4561ff8392"
-    jsonschema.validate(string, anonymized_string_schema)
+    jsonschema.validators.Draft202012Validator(
+        schema=anonymized_string_schema,
+        format_checker=jsonschema.validators.Draft202012Validator.FORMAT_CHECKER,
+    ).validate(string)
```
anthonyburdi, line 24: For all the tests using a Draft202012Validator, could we use a fixture with schema param?

great_expectations/core/usage_statistics/usage_statistics.py
anthonyburdi: This is all handled in the AWS lambda that accepts the events. The schema that we generate by running `schemas.py` (as a json file) is included in the lambda deployment and used by the lambda to validate events. Take a look at `https://github.com/greatexpectationslabs/usage_statistics`, which defines the lambda and uses the `serverless-reqvalidator-plugin` to actually do the validation: https://github.com/greatexpectationslabs/usage_statistics/blob/main/usage_statistics/serverless.yml#L4 Happy to do a walkthrough / answer questions of what I know.

anthonyburdi: I believe it rejects the event but I would verify that. Here is a link to docs on the plugin: https://www.serverless.com/plugins/serverless-reqvalidator-plugin

anthonyburdi (approved):
Looks good. Just some style suggestions, but non blocking.

tests/profile/test_jsonschema_profiler.py
jcampbell: I actually think we should revert my change here since it's related to the profiler vice stats.

tests/core/usage_statistics/test_schema_validation.py
jcampbell: done.

tests/core/usage_statistics/test_schema_validation.py
jcampbell: Probably -- but I think the real issue is that we're doing this in the code itself...and I think ideally our test would be of our code there. So, I'd recommend we leave this alone until we refactor.

anthonyburdi: Do you mean testing the `UsageStatisticsHandler.validate_message()` method? We do some tests on that in `tests/core/usage_statistics/test_usage_statistics_handler_methods.py`, we can potentially extend those tests to test all of our example events (messages).

great_expectations/core/usage_statistics/schemas.py
anthonyburdi: FYI @tannerbeam I upped the version to 1.0.1 in this PR, see great_expectations/core/usage_statistics/usage_statistics.py

tests/profile/test_jsonschema_profiler.py
anthonyburdi: Ah I see - JsonSchemaProfiler and we wouldn't want to introduce any breaking changes there. Reverting.

tannerbeam (approved):

PR #6515, opened 2022-12-08
[CONTRIB] fix observed value in custom expectation

itaise:
Hi,
In this experimental expectation (that I developed), the "observed value" was not serializable because it is of type "row proxy", there it was failing the expectation when trying to write it to Postgres DB. 
```
TypeError: ('2900-12-30 00:00:00.000 UTC',) is of type RowProxy which cannot be serialized.
```

I fixed it - by changing it's type to a list. 

Can you please review?

Thanks a lot!


anthonyburdi (approved):
Thank you for this fix!

PR #6524, opened 2022-12-09
add `mypy` badge

Kilo59:
...


Kilo59:
Coming back to this once we have better type coverage

PR #6525, opened 2022-12-09
[FEATURE] F/great 1397/zep checkpoints

billdirks:
This makes us able to run checkpoints using the new ZEP datasources. The main challenges were:
* We do a conversion of the batch request to a dictionary and back (ugh). This needed to be supported with the new ZEP batch requests.
* Usage statistics and anonymization required some data structures with some meta information about a batch (eg batch definition) so we constructed these and added them. There is an open ticket about fixing some other usage statistics issues up and these will hopefully be removed when we focus on fixing usage stats. See https://superconductive.atlassian.net/browse/GREAT-1375 


tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -496,8 +496,7 @@ def test_sort_batch_list_by_unknown_key(create_source):
 @pytest.mark.unit
 def test_data_source_json_has_properties(create_source):
     with create_source(lambda _: None) as source:
-        assert (
-            type(TableAsset.order_by) == property,
+        assert type(TableAsset.order_by) == property, (
```
billdirks: The changes in this file aren't directly related to my change. They are testing errors I uncovered while making this change and rolled them in. These test asserts always succeed since `assert (test, message)` always passes, since it is asserting a tuple, which is distinct from `assert test, (message)`.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -407,7 +407,7 @@ def _run_validation(
         ) as e:
             raise ge_exceptions.CheckpointError(
                 f"Exception occurred while running validation[{idx}] of Checkpoint '{self.name}': {e.message}."
-            )
+            ) from e
```
cdkini, line 410: what exactly does this do?

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -157,6 +162,12 @@ class Batch:
     # to a batch so developers may want to namespace any custom metadata they add.
     metadata: Dict[str, Any]
 
+    # TODO: These legacy fields are currently required. They are only used in usage stats so we
+    #       should figure out a better way to anonymize and delete them.
+    _legacy_batch_markers: BatchMarkers
+    _legacy_batch_spec: BatchSpec
+    _legacy_batch_definition: BatchDefinition
+
```
cdkini, line 170: Will these always be empty/null when we send over the wire?

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -206,3 +230,15 @@ def data(self) -> BatchDataType:
...
+    def batch_markers(self):
+        return self._legacy_batch_markers
+
+    @property
+    def batch_spec(self):
+        return self._legacy_batch_spec
+
+    @property
+    def batch_definition(self):
+        return self._legacy_batch_definition
```
cdkini: Can we type annotate these please?

tests/experimental/datasources/test_integration_sql_datasource.py
```diff
@@ -0,0 +1,107 @@
...
+        context,
+        **checkpoint_config,
+    )
+    checkpoint_result = checkpoint.run()
+
+    # Verify checkpoint runs successfully
+    assert checkpoint_result._success, "Running expectation suite failed"
+    assert (
+        len(checkpoint_result.run_results) == 1
+    ), "More than 1 run was done when we only expected 1"
```
cdkini: Could we assert the specific number of runs through an f-string?

tests/experimental/datasources/test_integration_sql_datasource.py
```diff
@@ -0,0 +1,107 @@
+import pytest
+
+from great_expectations import DataContext
+from great_expectations.checkpoint import SimpleCheckpoint
+from great_expectations.data_context.util import file_relative_path
+
+
+@pytest.mark.integration
+def test_run_checkpoint_and_data_doc(empty_data_context):
```
cdkini, line 9: This is quite a comprehensive test - might it make sense to add a docstring to explain exactly what we're testing and what workflow/use-case this covers?

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -512,8 +511,7 @@ def test_data_source_json_has_properties(create_source):
 @pytest.mark.unit
 def test_data_source_str_has_properties(create_source):
     with create_source(lambda _: None) as source:
-        assert (
-            type(TableAsset.order_by) == property,
+        assert type(TableAsset.order_by) == property, (
```
cdkini: Why `type` instead of `isinstance`? `type` uses string comparisons right?

great_expectations/checkpoint/checkpoint.py
billdirks: If you catch an exception and raise a different one the original backtrace of the first exception gets lost. This can sometimes be a problem when root causing the error. the `from previous_exception` notation makes the backtrace look something like:
```
<Original exception backtrace>
when handling that exception another exception occurred
<backtrace thrown here>
```

great_expectations/experimental/datasources/interfaces.py
billdirks: I don't believe this object is sent over the wire. Our anonymization code uses these fields to construct what is sent of the wire. It does that by calling the getter properties defined on this below.

great_expectations/experimental/datasources/interfaces.py
billdirks: Thank you. Done.

tests/experimental/datasources/test_integration_sql_datasource.py
billdirks: Done.

tests/experimental/datasources/test_integration_sql_datasource.py
billdirks: Done.

tests/experimental/datasources/test_postgres_datasource.py
billdirks: These are class comparisons since `type` returns a `type` object and `property` is a builtin class. The difference between `isinstance` and `type` is `type` compares exact type vs allowing for subclasses. This was done arbitrarily here though and I've updated to using `isinstance`.

great_expectations/checkpoint/checkpoint.py
cdkini: nevermind did some of my own research 😄

great_expectations/experimental/datasources/interfaces.py
cdkini: ahhh okay

cdkini (approved):

PR #6530, opened 2022-12-09
[MAINTENANCE] Leverage `RendererConfiguration` in existing prescriptive templates (3 of 3)

NathanFarmer:
- GREAT-1373
- Remove `styling` from `RendererConfiguration` as it is only a legacy renderer attribute
- Do not serialize `header` when it is `None` in atomic rendered content
- Validate that `RendererParam` `value` matches it's corresponding `schema_type`
- Leverage `root_validator` over `__init__` in `pydantic` model
- Add `RendererConfiguration` tests

### Definition of Done


great_expectations/render/renderer_configuration.py
```diff
@@ -95,49 +97,81 @@ class Config:
...
-    @staticmethod
-    def _set_expectation_type_and_kwargs(kwargs: dict) -> dict:
-        if "configuration" in kwargs and kwargs["configuration"]:
-            kwargs["expectation_type"] = kwargs["configuration"].expectation_type
-            kwargs["kwargs"] = kwargs["configuration"].kwargs
-        elif (
-            "result" in kwargs
-            and kwargs["result"]
-            and kwargs["result"].expectation_config
+    @root_validator(pre=True)
```
Kilo59, line 100: Does this need to be a `pre` validator?
Same question for the other validators.

great_expectations/render/renderer_configuration.py
```diff
@@ -95,49 +97,81 @@ class Config:
...
+    def validate_configuration_or_result(cls, values: dict) -> dict:
```
Kilo59: Can you make all the validators "private" and `@classmethod`s?

Kilo59 (approved):
@NathanFarmer 
Looks good to me, couple of minor suggestions I'd like to see implemented before the merge, but I'm going to go ahead and approve in case this falls off my radar tomorrow.

I think all the great error validation error messages would be a good thing to demo 😄

great_expectations/render/renderer_configuration.py
NathanFarmer: Actually this is the only one that needed it! I removed it from the rest.

great_expectations/render/renderer_configuration.py
NathanFarmer: 1) I'm only able to get validation to occur if I add `@classmethod` AFTER `@root_validator()` - just FYI
2) Once I do that PyCharm complains about `@rootvalidator`: `This decorator will not receive a callable it may expect; the built-in decorator returns a special object`.

Why doesn't the Pydantic documentation use the `@classmethod` decorator?

Kilo59: 1. Yes, order often matters with decorators.
2. I've never seen that. But I'm not using PyCharm.

>Why doesn't the Pydantic documentation use the @classmethod decorator?

I don't know, I hadn't noticed that before. They must be making these methods `classmethods` as part of the pydantic metaclass hook. Otherwise, it wouldn't work.

Given the above, I think it's fine to leave off the `@classmethod` decorator.

PR #6531, opened 2022-12-09
[MAINTENANCE] Refactor `BaseDataContext` and `DataContext` into factory functions

cdkini:
- Introduce `SerializableDataContext` to act as common ground between `FileDataContext` and `CloudDataContext`
- Update `BaseDataContext` and `DataContext` to become functions that return a subclass of `AbstractDataContext`
- Update all static/class calls to `BaseDataContext` and `DataContext`
- Update all type hints

### Definition of Done


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -161,7 +162,7 @@
 T = TypeVar("T", dict, list, str)
 
 
-class AbstractDataContext(ABC):
+class AbstractDataContext(ConfigPeer, ABC):
```
cdkini, line 166: Need to move over to the base class as `BaseDataContext` can no longer inherit

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -237,6 +241,18 @@ def __init__(self, runtime_environment: Optional[dict] = None) -> None:
...
+            and self.config.validation_operators
+        ):
+            for (
+                validation_operator_name,
+                validation_operator_config,
+            ) in self.config.validation_operators.items():
+                self.add_validation_operator(
+                    validation_operator_name,
+                    validation_operator_config,
+                )
```
cdkini, line 258: Ported over from `BaseDataContext` - required for proper validation operator creation

great_expectations/data_context/data_context/base_data_context.py
```diff
@@ -22,542 +15,72 @@
...
+def _resolve_cloud_args(
+    cloud_mode: bool = False,
+    cloud_config: Optional[GXCloudConfig] = None,
+    # <GX_RENAME> Deprecated as of 0.15.37
+    ge_cloud_mode: bool = False,
+    ge_cloud_config: Optional[GXCloudConfig] = None,
+) -> Tuple[bool, Optional[GXCloudConfig]]:
+    cloud_mode = True if cloud_mode or ge_cloud_mode else False
+    cloud_config = cloud_config if cloud_config is not None else ge_cloud_config
+    return cloud_mode, cloud_config
```
cdkini, line 189: Moved from a static method to a stand-alone helper

great_expectations/data_context/data_context/data_context.py
```diff
@@ -1,674 +1,139 @@
...
+
+
+def _init_cloud_config(
+    cloud_mode: bool,
+    cloud_base_url: Optional[str],
+    cloud_access_token: Optional[str],
+    cloud_organization_id: Optional[str],
+) -> Optional[GXCloudConfig]:
+    if not cloud_mode:
+        return None
```
cdkini: Took any class methods that were really specific to `DataContext` and made them free-standing helpers - these are usually methods that combine file and Cloud behavior (but don't fit in `SerializableDataContext`, `CloudDataContext`, or `FileDataContext`)

great_expectations/data_context/data_context/explorer_data_context.py
```diff
@@ -1,45 +1,49 @@
...
 
 
-class ExplorerDataContext(DataContext):
-    def __init__(self, context_root_dir=None, expectation_explorer=True) -> None:
-        """
-            expectation_explorer: If True, load the expectation explorer manager, which will modify GX return objects \
-            to include ipython notebook widgets.
-        """
+class ExplorerDataContext:
+    pass
```
cdkini: Will come back to this 😕

great_expectations/data_context/data_context/file_data_context.py
```diff
@@ -19,20 +20,18 @@
...
     class will exist only for backwards-compatibility reasons.
     """
 
-    GX_YML = "great_expectations.yml"
-
     def __init__(
         self,
-        project_config: DataContextConfig,
         context_root_dir: str,
+        project_config: Optional[DataContextConfig] = None,
```
cdkini, line 42: If not present, we should have legacy `DataContext` behavior where we parse it from the filesystem

azure-pipelines-dev.yml
```diff
@@ -120,8 +120,9 @@ stages:
     - job: static_type_check
       steps:
       - script: |
-          pip install --requirement reqs/requirements-dev-contrib.txt
-          invoke type-check --install-types --warn-unused-ignores --pretty
+          echo 'This is disabled for now!'
+          # pip install --requirement reqs/requirements-dev-contrib.txt
+          # invoke type-check --install-types --warn-unused-ignores --pretty
```
cdkini: Will resolve all of these after the first round of review - should mainly be just `grep` and `sed` to convert old types to `AbstractDataContext` and its subclasses

great_expectations/checkpoint/actions.py
```diff
@@ -48,6 +48,10 @@ class ValidationAction:
     def __init__(self, data_context) -> None:
         self.data_context = data_context
 
+    @property
+    def _cloud_mode(self) -> bool:
+        return hasattr(self.data_context, "cloud_mode") and self.data_context.cloud_mode
```
cdkini: We'll want to clean this up eventually but Cloud leaks into everything so this is one way to encapsulate it a bit

great_expectations/cli/cli.py
```diff
@@ -51,7 +54,9 @@ def data_context(self) -> Optional[DataContext]:
 
     @data_context.setter
     def data_context(self, data_context: DataContext) -> None:
-        assert isinstance(data_context, DataContext)
```
cdkini, line 54: Type hints will be updated so please ignore those for now

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -548,6 +567,19 @@ def concurrency(self) -> Optional[ConcurrencyConfig]:
...
+
+    def _attach_datasource_to_context(self, datasource: XDatasource):
+        # We currently don't allow one to overwrite a datasource with this internal method
+        if datasource.name in self.datasources:
+            raise ge_exceptions.DataContextError(
+                f"Can not write the experimental datasource {datasource.name} because a datasource of that "
+                "name already exists in the data context."
+            )
+        self.datasources[datasource.name] = datasource
+
```
cdkini, line 582: Ported over from `DataContext`

great_expectations/data_context/data_context/explorer_data_context.py
cdkini: Think we can just delete this but maybe I'm being too rash

great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -0,0 +1,443 @@
...
+        self, context_root_dir: str, runtime_environment: Optional[dict] = None
+    ) -> None:
+        self._context_root_directory = context_root_dir
+        super().__init__(runtime_environment=runtime_environment)
+
+    def _init_datasource_store(self):
+        pass
+
+    def _init_variables(self):
+        pass
```
cdkini: Required to fulfill ABC but we never actually instantiate a `SerializableDataContext` by itself - thoughts?

great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -0,0 +1,443 @@
...
+)
+from great_expectations.data_context.util import file_relative_path
+
+logger = logging.getLogger(__name__)
+yaml = YAML()
+yaml.indent(mapping=2, sequence=4, offset=2)
+yaml.default_flow_style = False
+
+
+class SerializableDataContext(AbstractDataContext):
```
cdkini, line 35: TODO: docstring (and bring back any old docstrings I may have deleted)

great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -0,0 +1,443 @@
...
+    GX_UNCOMMITTED_DIR = "uncommitted"
+    BASE_DIRECTORIES = [
+        DataContextConfigDefaults.CHECKPOINTS_BASE_DIRECTORY.value,
+        DataContextConfigDefaults.EXPECTATIONS_BASE_DIRECTORY.value,
+        DataContextConfigDefaults.PLUGINS_BASE_DIRECTORY.value,
+        DataContextConfigDefaults.PROFILERS_BASE_DIRECTORY.value,
+        GX_UNCOMMITTED_DIR,
+    ]
+    GX_DIR = "great_expectations"
+    GX_YML = "great_expectations.yml"  # TODO: migrate this to FileDataContext. Still needed by DataContext
```
cdkini: Needs deletion

great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -0,0 +1,443 @@
...
+
+        config_filepath = os.path.join(self.root_directory, self.GX_YML)  # type: ignore[arg-type]
+
+        try:
+            with open(config_filepath, "w") as outfile:
+                self.config.to_yaml(outfile)
+        except PermissionError as e:
+            logger.warning(f"Could not save project config to disk: {e}")
+
+    def _check_for_usage_stats_sync(self, project_config: DataContextConfig) -> bool:
```
cdkini, line 84: All these methods are coming over from `DataContext`

great_expectations/checkpoint/actions.py
Kilo59: Why not just keep the existing `self.data_context.cloud_mode` until we can improve the abstraction?
This change seems lateral.

great_expectations/data_context/data_context/file_data_context.py
```diff
@@ -95,3 +109,43 @@ def _init_variables(self) -> FileDataContextVariables:
...
+        store = super().add_store(store_name=store_name, store_config=store_config)
+        self._save_project_config()
+        return store
+
+    @classmethod
+    def _load_file_backed_project_config(
+        cls,
+        context_root_directory: Optional[str],
+    ):
+        path_to_yml = os.path.join(context_root_directory, cls.GX_YML)
```
Kilo59:
```suggestion
        path_to_yml = pathlib.Path(context_root_directory, cls.GX_YML)
```
Let's use `pathlib` here, and remove the `os` import.

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -1138,7 +1138,7 @@ def add_profiler(
         )
 
         key: Union[GXCloudIdentifier, ConfigurationIdentifier]
-        if data_context.cloud_mode:
+        if hasattr(data_context, "cloud_mode") and data_context.cloud_mode:
```
Kilo59: Aren't these the same check?

great_expectations/validation_operators/validation_operators.py
```diff
@@ -370,7 +374,7 @@ def run(
 
             run_results = {}
             for batch, async_batch_validation_result in batch_and_async_result_tuples:
-                if self.data_context.cloud_mode:
+                if self._cloud_mode:
```
Kilo59: Same comment as above. Why are we making these `self.data_context.cloud_mode -> if self._cloud_mode` changes?

tests/data_context/test_data_context.py
```diff
@@ -861,10 +864,10 @@ def test_add_store(empty_data_context):
...
 
 # noinspection PyPep8Naming
-def test_ExplorerDataContext(titanic_data_context):
-    context_root_directory = titanic_data_context.root_directory
-    explorer_data_context = ExplorerDataContext(context_root_directory)
-    assert explorer_data_context._expectation_explorer_manager
+# def test_ExplorerDataContext(titanic_data_context):
+#     context_root_directory = titanic_data_context.root_directory
+#     explorer_data_context = ExplorerDataContext(context_root_directory)
+#     assert explorer_data_context._expectation_explorer_manager
```
Kilo59: Probably already on your radar but, don't forget about this.

tests/data_context/test_data_context.py
```diff
@@ -1032,24 +1035,24 @@ def test_data_context_create_does_not_raise_error_or_warning_if_ge_dir_exists(
...
     project_path = str(tmp_path_factory.mktemp("data_context"))
-    DataContext.create(project_path)
+    FileDataContext.create(project_path)
     ge_dir = os.path.join(project_path, "great_expectations")
     assert os.path.isdir(ge_dir)
-    assert os.path.isfile(os.path.join(ge_dir, DataContext.GX_YML))
+    assert os.path.isfile(os.path.join(ge_dir, FileDataContext.GX_YML))
     context = DataContext(ge_dir)
-    assert isinstance(context, DataContext)
+    assert isinstance(context, FileDataContext)
```
Kilo59: Can you annotate the type of this fixture?
Is it `pathlib.Path` or a `str`?

great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -0,0 +1,447 @@
...
+            project_root_dir: path to the root directory in which to create a new great_expectations directory
+            usage_statistics_enabled: boolean directive specifying whether or not to gather usage statistics
+            runtime_environment: a dictionary of config variables that override both those set in
+                config_variables.yml and the environment
+
+        Returns:
+            DataContext
+        """
+
+        if not os.path.isdir(project_root_dir):  # type: ignore[arg-type]
```
Kilo59, line 173: **non-blocking/followup**
Since this is a net new module can we use `pathlib` and remove `os`?
If compatibility is a concern we can just call `str(my_path_obj)` on the path object when passing it.

Here's a handy table that maps `os.path` methods to `pathlib` methods.
https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -0,0 +1,447 @@
...
+            AnonymizedUsageStatisticsConfig
+        ] = self.config.anonymous_usage_statistics
+
+        if (
+            project_config_usage_stats.enabled is False  # type: ignore[union-attr]
+            or context_config_usage_stats.enabled is False  # type: ignore[union-attr]
+        ):
+            return False
+
+        if project_config_usage_stats.explicit_id is False:  # type: ignore[union-attr]
```
Kilo59, line 115: Can we fix these issues on `AnonymizedUsageStatisticsConfig`?
Assuming it's a small change.

great_expectations/checkpoint/actions.py
cdkini: not all contexts have cloud_mode

tests/data_context/test_data_context.py
cdkini: yup going to delete!

tests/data_context/test_data_context.py
cdkini: it's a `pytest` built-in: `_pytest.fixture.TemporaryPathFactory` I believe

great_expectations/data_context/data_context/file_data_context.py
cdkini: done!

great_expectations/validation_operators/validation_operators.py
cdkini: same note as elsewhere - `cloud_mode` is no longer a property so we're doing an `isinstance` check

great_expectations/data_context/data_context/serializable_data_context.py
cdkini: I'd be happy to but would you mind if I did this `os` replacement work in a subsequent PR? Can add to the context epic I'm working in but I'd like to focus on getting this change in before it gets stale.

Will defer to you here.

great_expectations/data_context/data_context/serializable_data_context.py
cdkini: We've got a ticket in the context epic to resolve `mypy` issues - mind if we tackle that there?

great_expectations/data_context/data_context/serializable_data_context.py
Kilo59: Sounds good to me.

great_expectations/checkpoint/actions.py
```diff
@@ -48,6 +48,14 @@ class ValidationAction:
     def __init__(self, data_context) -> None:
         self.data_context = data_context
 
+    @property
+    def _cloud_mode(self) -> bool:
+        from great_expectations.data_context.data_context.cloud_data_context import (
+            CloudDataContext,
+        )
+
+        return isinstance(self.data_context, CloudDataContext)
```
Kilo59: It's still not clear to me why it's not better to have the base/abstract `DataContext` class have a `data_context.cloud_mode` attribute that defaults to false.
I don't think the extra code churn here actually makes the abstraction any better. But maybe I just don't understand it.

It should go without saying that we should work to abstract away these conditional `if cloud_mode` checks in a subsequent effort.

great_expectations/data_context/data_context/serializable_data_context.py
Kilo59: That works for me.

great_expectations/checkpoint/actions.py
billdirks: One goal of this effort is to remove the notion of `cloud` from the data contexts except for the `CloudDataContext`. I think that is what is motivating this. Are we removing this too early, since now we must do checks like this in 3 files? maybe. We should remove these checks in general and not need to know if we are working with a cloud context at all.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -497,9 +495,17 @@ def ge_cloud_id(self) -> Optional[UUID]:
             return None
 
     @property
-    def data_context(self) -> DataContext:
+    def data_context(self) -> AbstractDataContext:
```
billdirks, line 498: I'm guessing we never have a true `AbstractDataContext` here but a subclass. What you are doing is fine. You could use a Generic here. That is at the top of the file do: ` T = TypeVar('T', bound=AbstractDataContext)`. Then in the class declaration you'd do `class BaseCheckpoint(ConfigPeer, Generic[T]):` and replace all instances of `AbstractDataContext` by `T`. This means if the class was initialized with a `FileDataContext`, the type checker would then know that the getter will return a `FileDataContext` and not an abstract one.

This comment doesn't block this PR. I'm trying to highlight a reason you might want to use generics here.

great_expectations/data_context/data_context/abstract_data_context.py
billdirks: `ConfigPeer` is an `ABC` so we do need that here? Previously did all previous `*DataContext`s inherit from `ConfigPeer`? I'm just wondering the ramifications of adding `ConfigPeer` as a parent class here.

billdirks (commented):
Thanks for doing this. I've only looked at a quarter of the files and will do more tomorrow.

great_expectations/checkpoint/actions.py
cdkini: These checks definitely need to go but as it stands, there is quite a bit of Cloud leakage into other parts of the codebase. Since this property is simply encapsulating that leakage here, I'd vote to keep this for now and then work to decouple Cloud in a subsequent effort.

Just to clarify, we're not adding a new behavior here - instead of checking for `cloud_mode`, we're evaluating whether or not we're in a Cloud backed environment through the context type (different ways to achieve the same goal).

Happy to brainstorm a better way to do this.

great_expectations/checkpoint/checkpoint.py
cdkini: nope this is a great point and something I think should go in as part of the implementation - I've paired with @Kilo59 just to clarify some points and have made the necessary changes.

Let me know if you agree!

great_expectations/checkpoint/actions.py
Kilo59: @cdkini I understand that we aren't adding new behavior here. I also accept that this sort of check is still needed.

What I'm saying is that I don't think only having the `cloud_mode` attribute on the `CloudDataContext` really improves things, since you still have to check for it on every `DataContext` anyways. Except now it's harder because we are wary of `AttributeErrors`.

cdkini: oh no we removed `cloud_mode` as an attr now - hence the change to `isinstance` as a check

Kilo59: >oh no we removed cloud_mode as an attr now - hence the change to isinstance as a check

I understand that `cloud_mode` is only an attr on  `CloudDataContext`.

great_expectations/validation_operators/validation_operators.py
```diff
@@ -235,6 +235,14 @@ def __init__(
                 )
             self.actions[action_config["name"]] = new_action
 
+    @property
+    def _cloud_mode(self) -> bool:
```
cdkini: Change name to `is_cloud_context` or something a bit different to better convey what is going on with this temp property?

great_expectations/data_context/data_context/serializable_data_context.py
billdirks: Are you getting a linting check since it expects this class to be concrete? I wonder if you can tell whichever check is raising the issue that this class is still abstract?

Otherwise, you could do `raise NotImplementedError` instead of passing so no one will calls these on accident.

great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -0,0 +1,447 @@
...
+    def _init_variables(self):
+        pass
+
+    def _save_project_config(self) -> None:
+        """
+        See parent 'AbstractDataContext._save_project_config()` for more information.
+
+        Explicitly override base class implementation to retain legacy behavior.
+        """
+        logger.debug("Starting DataContext._save_project_config")
```
billdirks: Is there any helpful info you can provide about the DataContext here? I'm fine with keeping it, but I'm wondering how useful this debug message is.

billdirks (commented):
Thanks for doing this refactor @cdkini! I have one small comment about understanding the linting error that can be resolved by removing the `pass` implementations by telling the linter it's ok that the methods aren't there or by implementing them with a `raise NotImplementedError`.

Otherwise I'm fine with this change and you can @Kilo59 can work through the discussion you are having.

Thanks!

great_expectations/data_context/data_context/serializable_data_context.py
cdkini: done!

Kilo59 (approved):

PR #6554, opened 2022-12-11
[MAINTENANCE] Improve DataAssistant Parameter Builder Naming/Sanitization Mechanism and Enhance TableDomainBuilder

alexsherstinsky:
### Scope
* Currently, Rule-Based Profiler parameter names in `DataAssistant` implementations use `metric_name` (with sanitization to replace periods with underscores).  However, since we generate two parameter name versions, `raw` and `JSON-serialized`, for each `metric_name`, this mechanism causes one of these names to be missed.  Present work repairs this by adding `metric_value_kwargs` ID of corresponding `MetricConfiguration` object to the name, which insures uniqueness.
* For table-pair expectations, the `TableDomainBuilder` needs to recognize the other table.  This capability is added.
* Unit tests for `COLUMN_PAIR` `Domain` objects are added.

- JIRA: GREAT-1347/GREAT-181/GREAT-1437
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/rule_based_profiler/data_assistant/data_assistant.py
```diff
@@ -272,6 +274,7 @@ def get_column_standard_deviation_metric_multi_batch_parameter_builder(
         @staticmethod
         def build_metric_multi_batch_parameter_builder(
             metric_name: str,
+            suffix: Optional[str] = None,
```
NathanFarmer, line 277: If we always plan to pass `str` or `None` to this method do we need this default?

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
```diff
@@ -346,6 +351,7 @@ def build_numeric_metric_multi_batch_parameter_builder(
         @staticmethod
         def build_numeric_metric_single_batch_parameter_builder(
             metric_name: str,
+            suffix: Optional[str] = None,
```
NathanFarmer, line 354: Same comment here

great_expectations/rule_based_profiler/helpers/util.py
```diff
@@ -1050,8 +1051,15 @@ def get_or_create_expectation_suite(
...
-def sanitize_parameter_name(name: str) -> str:
+def sanitize_parameter_name(
+    name: str,
+    suffix: Optional[str] = None,
+) -> str:
     """
-    This method provides display-friendly version of "name" argument.
+    This method provides display-friendly version of "name" argument (with optional "suffix" argument).
     """
+    if suffix:
```
NathanFarmer, line 1074: I am unable to follow what purpose this `suffix` serves.

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
```diff
@@ -276,10 +276,10 @@ def test_onboarding_data_assistant_metrics_count(
         domain,
         parameter_values_for_fully_qualified_parameter_names,
     ) in bobby_onboarding_data_assistant_result.metrics_by_domain.items():
-        if domain.is_superset(domain_key):
+        if domain.is_superset(other=domain_key):
             num_metrics += len(parameter_values_for_fully_qualified_parameter_names)
 
-    assert num_metrics == 2
+    assert num_metrics == 4
```
NathanFarmer, line 282: Why does this double?

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
alexsherstinsky: @NathanFarmer This is just my style -- the default is necessary (as you have seen in auto-initializing expectations and other method calls, I prefer to explicitly specify all parameters for the reader to know what parameters are possible).  At some point, we may wish to reconsider this style, as it may be more Pythonic to only specify what is needed.  I am not sure at this point.  Thanks.

great_expectations/rule_based_profiler/helpers/util.py
alexsherstinsky: @NathanFarmer The suffix enables us to make parameter names unique.  Without it, what can happen is that same metric name, same domain kwargs, but different value kwargs can cause two metric configurations to map to the same parameter name.  By adding a suffix, we can disambiguate them.  Thanks.

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
alexsherstinsky: @NathanFarmer Because with the suffix now, we have uncovered all the parameter builder names with `raw` in them, which got squashed before -- the suffix fixes this bug now.  Thanks.

NathanFarmer: @alexsherstinsky but everywhere I look I see `suffix=None`, so where is this getting set to affect this test?

great_expectations/rule_based_profiler/helpers/util.py
NathanFarmer: @alexsherstinsky Why do we pass a `suffix` string if we are just going to then hash it? Also, how does `suffix` relate to value kwargs in the code itself? How is the user supposed to know what `suffix` means?

alexsherstinsky: @NathanFarmer Please see improved `docstring` and additional unit tests (plus some more "beautification").  Thanks!

great_expectations/rule_based_profiler/helpers/util.py
```diff
@@ -1050,8 +1051,28 @@ def get_or_create_expectation_suite(
...
+    name: str,
+    suffix: Optional[str] = None,
+) -> str:
     """
-    This method provides display-friendly version of "name" argument.
+    This method provides display-friendly version of "name" argument (with optional "suffix" argument).
+
+    In most situations, "suffix" is not needed.  However, in certain use cases, "name" argument is same among different
+    calling structures, whereby "name" can be disambiguated by addition of distinguishing "suffix" argument.  Using
+    list of "MetricConfiguration" objects as example, "metric_name" and "metric_domain_kwargs" are commoonly same among
```
NathanFarmer: "commoonly" -> "commonly"

great_expectations/rule_based_profiler/helpers/util.py
```diff
@@ -1050,8 +1051,28 @@ def get_or_create_expectation_suite(
...
+    them, while "metric_value_kwargs" are different (i.e., calculating values of same metric with different parameters).
+    In this case, supplying "MetricConfiguration.metric_domain_kwargs_id" as "suffix" makes sanitized names unique.
```
NathanFarmer, line 1065: In what case would we not want this behavior? Are we not doing this by default just because the hash is "ugly"? Are there other use cases for what you would want to pass into `suffix`? If not, why not just make `suffix` boolean?

great_expectations/rule_based_profiler/helpers/util.py
alexsherstinsky: @NathanFarmer Fixed -- nice catch -- thank you!

great_expectations/rule_based_profiler/helpers/util.py
alexsherstinsky: @NathanFarmer Yes -- the former, it can be ugly; in many cases we may not care.  If we want to make it permanent, then we face two alternatives: either the caller computes `name + suffix`, or we pass `MetricConfiguration`.  I felt that either was non-ideal.  The former places too much burden on caller, while the latter requires an object that may not be available.  Hence, the compromise.  Thanks.

NathanFarmer: I see, thank you!

NathanFarmer (approved):

PR #6580, opened 2022-12-14
[FEATURE] ZEP - PG `BatchSorter` loading + dumping

Kilo59:
- Remove the `order_by` property and custom `__init__` on `TableAsset`. `order_by` is now just a regular field with a [custom validator](https://docs.pydantic.dev/usage/validators/).
- Make the Postgres `TableAsset.order_by` field de-serializable in either the succinct list of strings format or as a list of dicts (json/yaml objects)
- For now `TableAsset.order_by` will always serialize as a list of objects. We can revisit this in a followup if needed.

## Problems with the serialization of nested Pydantic objects

I expected specifying a custom json encoder format for `BatchSorter` to be a trivial problem, thanks to being able to provide custom dump/load methods for types.
https://docs.pydantic.dev/usage/exporting_models/#json_encoders

However, after a fair amount of painful debugging, I stumbled upon the fact that nested encoders are not working as expected.
https://github.com/pydantic/pydantic/issues/2277

Unfortunately, we can't simply create a custom `.json()` method because for nested models, `.json()` is not called.

So if we want to serialize the `BatchSorter` in the more succinct list of strings format `['-year', '+month']` we will need to use a different approach. Such as a custom `TableAsset.dict()` or using a custom `root_type`.
https://docs.pydantic.dev/usage/models/#custom-root-types
https://stackoverflow.com/questions/68453514/modifying-pydantic-json-encoders-for-a-class-not-working
Should note that this isn't a big lift, but will defer this work until higher priority work is finished.

### Definition of Done


tests/experimental/datasources/config.yaml
```diff
@@ -19,3 +19,18 @@ datasources:
...
+          - metadata_key: year
+          - metadata_key: month
+            reverse: true
+      with_dsl_ish_sorters:
+        type: table
+        name: with_dsl_ish_sorters
+        table_name: another_table
+        order_by:
+          - year
+          - -month
```
Kilo59, line 36: ## Note
After this PR we can parse the succinct `+/-<KEY>` strings from config without issue, but it will always dump/serialize in the more explicit list of objects format.

great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -116,19 +117,21 @@ class TableAsset(DataAsset):
...
+    def parse_orderby_sorter(
+        cls, v: Union[str, BatchSorter]
+    ) -> Union[BatchSorter, dict]:
+        print(f"validator - {type(v)} {v}")
+        if isinstance(v, str):
+            try:
+                return _batch_sorter_from_str(v)
+            except KeyError as err:
+                # TODO: before merge - raise a helpful validation error here
+                raise err
```
Kilo59: TODO

- [x] remove print
- [x] better error

Kilo59: image

great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -403,7 +405,11 @@ def add_table_asset(
         Returns:
             The TableAsset that is added to the datasource.
         """
-        asset = TableAsset(name=name, table_name=table_name, order_by=order_by)
+        asset = TableAsset(
+            name=name,
+            table_name=table_name,
+            order_by=order_by or [],  # type: ignore[arg-type]  # empty list is allowed
```
NathanFarmer: Why does this result in an error?

tests/experimental/datasources/config.yaml
NathanFarmer, line 36: What is the purpose of the `+/-`?

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -493,28 +495,45 @@ def test_sort_batch_list_by_unknown_key(create_source):
             source.get_batch_list_from_batch_request(batch_request)
 
 
+@pytest.mark.unit
+@pytest.mark.parametrize(
+    "order_by",
+    [
+        ["+year", "-month"],
```
NathanFarmer, line 502: Ascending/descending?

Kilo59: Yes

great_expectations/experimental/datasources/postgres_datasource.py
Kilo59: I'm not sure. Possibly I'm using the wrong version of `mypy`, I was surprised when I saw this warning.
I'll remove the type ignore and see what happens in CI.

Kilo59:
```
Argument "order_by" to "TableAsset" has incompatible type
"Union[List[BatchSorter], List[str], List[<nothing>]]";
expected "List[BatchSorter]"  [arg-type]
```
Nevermind, the empty list wasn't the issue.
The type for `order_by` is `List[BatchSorter]` but there's a custom validator on the model that can coerce a `List[str]` into a `List[BatchSorter]`.

So I'm keeping the type ignore but I'll update the note.

tests/experimental/datasources/config.yaml
Kilo59: As you guessed, whether to sort by ascending/descending order. Or default/reverse order.

great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -116,19 +117,18 @@ class TableAsset(DataAsset):
...
-        # passes in `order_by=None`, I want this variable to be `[]` and not `None`.
-        # `pop("order_by", [])` will return None since the order_by key exists in this case.
-        order_by = kwargs.pop("order_by", None) or []
-        self._order_by = _batch_sorter_from_list(order_by)
-        super().__init__(**kwargs)
-
-    @property
-    def order_by(self) -> List[BatchSorter]:
-        return self._order_by
+    order_by: List[BatchSorter] = Field(default_factory=list)
```
Kilo59, line 120: Mutable default values for pydantic fields are not a problem.
https://stackoverflow.com/a/63808835/6304433

We could do this as below...
```python
order_by: List[BatchSorter] = []
```
but this will upset most pythonistas unless they happen to know `pydantic` well... so using the `Field` `default_factory` is more explicit.

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -493,28 +495,45 @@ def test_sort_batch_list_by_unknown_key(create_source):
...
-            "This test assumes TableAsset.order_by is a property. If it is not we "
-            "should update this test",
-        )
         asset = source.add_table_asset(name="my_asset", table_name="my_table")
         asset.add_year_and_month_splitter(column_name="my_col").add_sorters(
             ["year", "month"]
         )
-        source_json = source.json()
+        source_json = source.json(indent=4, sort_keys=True)
+        print(source_json)
```
cdkini, line 530: Should this be `pprint`?

tests/experimental/datasources/test_config.py
```diff
@@ -166,10 +181,26 @@ def test_yaml_file_config_round_trip(
...
     inject_engine_lookup_double, from_yaml_gx_config: GxConfig
 ):
     dumped: str = from_yaml_gx_config.yaml()
 
     assert PG_CONFIG_YAML_STR == dumped
+
+
+@pytest.mark.xfail(
+    reason="Custom BatchSorter serialization logic needs to be implemented"
+)
```
cdkini, line 198: Should we add any of the other named args?

great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -198,7 +198,9 @@ def batch_request_options_template(
         return {p: None for p in self.column_splitter.param_names}
 
     def add_sorters(self, sorters: BatchSortersDefinition) -> TableAsset:
-        self._order_by = _batch_sorter_from_list(sorters)
+        # NOTE: (kilo59) we could use pydantic `validate_assignment` for this
+        # https://docs.pydantic.dev/usage/model_config/#options
+        self.order_by = _batch_sorter_from_list(sorters)
```
cdkini, line 203: Is this captured in a ticket anywhere?

great_expectations/experimental/datasources/postgres_datasource.py
cdkini: Appreciate the context here 🙇🏽 

I say let's stay away from mutable defaults just because they're such a footgun with vanilla Python

tests/experimental/datasources/test_postgres_datasource.py
Kilo59: No because it's just a string. The `indent` keyword makes it pretty.

great_expectations/experimental/datasources/postgres_datasource.py
Kilo59: I don't think this is ticket worthy. We may not even want to do it.

great_expectations/experimental/datasources/postgres_datasource.py
Kilo59: Just to clarify for other readers.
mutable defaults ARE NOT a footgun when used with pydantic.

Both of the examples below are immune from the usual problems associated with [mutable default values](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) in python.
We are only choosing the second example because it's more expressive/explicit.
```python
order_by_a: List[BatchSorter] = []
order_by_b: List[BatchSorter] = Field(default_factory=list)
```

tests/experimental/datasources/test_config.py
Kilo59: Just updated the test to be better and more explicit.

https://github.com/great-expectations/great_expectations/pull/6580/commits/c2770d5e3144c02c4e2dd2dcec58db57b34d93f8

cdkini (approved):

billdirks (approved):
LGTM, thank you!

PR #6586, opened 2022-12-14
[FEATURE] ID/PK - `unexpected_index_list` updated to include actual unexpected value, and EVR to include `unexpected_index_column_names`

Shinnnyshinshin:
#### Changes proposed in this pull request:
- `EVR` now contain `unexpected_index_column_names` which were only in ExpectatinConfiguration before (changes to `expectation.py` to enable)
-  `unexpected_index_list` in `EVR` will now contain actual unexpected value. Change has been made to `pandas` and `sql` implementation
- Update `Checkpoint` tests and `Metrics` tests correspondingly
### Definition of Done


great_expectations/expectations/expectation.py
```diff
@@ -3238,6 +3246,10 @@ def _format_map_output(
         return_obj["result"].update({"unexpected_index_list": unexpected_index_list})
     if unexpected_index_query is not None:
         return_obj["result"].update({"unexpected_index_query": unexpected_index_query})
+    if unexpected_index_column_names is not None:
+        return_obj["result"].update(
+            {"unexpected_index_column_names": unexpected_index_column_names}
+        )
```
Shinnnyshinshin: Ensuring `unexpected_index_column_names` is part of the Validation result object.

great_expectations/expectations/expectation.py
```diff
@@ -3164,6 +3172,7 @@ def _format_map_output(
     unexpected_list: Optional[List[Any]] = None,
     unexpected_index_list: Optional[List[int]] = None,
     unexpected_index_query: Optional[str] = None,
+    unexpected_index_column_names: Optional[Union[int, str, List[str]]] = None,
```
Shinnnyshinshin, line 3176: This is not entirely accurate, but is necessary to keep the typechecker happy (it's actually `Optional[List[str]]`)

great_expectations/expectations/metrics/map_metric_provider.py
```diff
@@ -1786,16 +1786,23 @@ def _pandas_map_condition_index(
...
+                index, expectation_domain_column_name
+            ]
             for column_name in unexpected_index_column_names:
                 column_name = get_dbms_compatible_column_names(
                     column_names=column_name,
                     batch_columns_list=metrics["table.columns"],
                     execution_engine=execution_engine,
                     error_message_template='Error: The unexpected_index_column "{column_name:s}" does not exist in Dataframe. Please check your configuration and try again.',
                 )
+                # add is the actual unexpected value
```
Shinnnyshinshin:
```suggestion
                # add the actual unexpected value
```

great_expectations/expectations/metrics/map_metric_provider.py
```diff
@@ -2500,6 +2513,8 @@ def _sqlalchemy_map_condition_index(
 
     for row in query_result:
         primary_key_dict: Dict[str, Any] = {}
+        # add is the actual unexpected value
```
Shinnnyshinshin:
```suggestion
        # add the actual unexpected value
```

great_expectations/expectations/expectation.py
anthonyburdi: I'm assuming yes, but have you tried removing the `int` and `str` from the union in this definition and all others? I'm surprised that doesn't work. If it doesn't maybe a good idea is to put this note in a code comment.

great_expectations/expectations/metrics/map_metric_provider.py
```diff
@@ -2465,9 +2472,11 @@ def _sqlalchemy_map_condition_index(
 
     column_selector: List[sa.Column] = []
     all_table_columns: List[str] = metrics.get("table.columns")
-    unexpected_index_column_names: List[str] = result_format.get(
+
+    unexpected_index_column_names: List[Union[str, None]] = result_format.get(
```
anthonyburdi: I'm thinking this type hint might instead be `Union[List[str], None]` i.e. either a list of strings or None (not a list that can contain None values), what do you think?

tests/checkpoint/test_checkpoint_result_format.py
```diff
@@ -193,12 +193,27 @@ def test_sql_result_format_in_checkpoint_pk_defined_one_expectation_complete_out
...
+    ]
+    first_result_partial_list: List[Dict[str, Any]] = evrs[0]["results"][0]["result"][
         "partial_unexpected_index_list"
     ]
-    assert first_result_partial_list == [{"pk_1": 3}, {"pk_1": 4}, {"pk_1": 5}]
+    assert first_result_partial_list == [
+        {"pk_1": 3, "animals": "giraffe"},
+        {"pk_1": 4, "animals": "lion"},
+        {"pk_1": 5, "animals": "zebra"},
+    ]
```
anthonyburdi: Not a blocking suggestion, but you might be able to refactor out a lot of this repeated assertion logic into a private module level helper method (like you did with `_add_expectations_and_checkpoint()`).

anthonyburdi: Might open the door to parametrizing all these tests. But your call - only if you think that helps readability.

anthonyburdi (commented):
Looking good! Just a few questions I'd like to address before approving.

great_expectations/expectations/metrics/map_metric_provider.py
Shinnnyshinshin: Ha you're absolutely right. My mistake

tests/checkpoint/test_checkpoint_result_format.py
Shinnnyshinshin: Great suggestion :) If you're ok with it, I'll move the expected values into a separate fixture in this case.

anthonyburdi (approved):
Thank you for making those changes!

PR #6589, opened 2022-12-15
[MAINTENANCE] `mypy` config update

Kilo59:
- update the `pyproject.toml` `mypy` exclude section to clarify its purpose
- disable the `annotation-unchecked` error code. https://github.com/python/mypy/pull/13851. Note this is not considered to be an `error` by mypy and so wouldn't actually catch things in CI without additional configuration. Removing because it creates a lot of noise in the CI.
- enable the [`--ignore-without-code` check](https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-type-ignore-include-an-error-code-ignore-without-code) which enforces that devs provide an error code along with an `# type: ignore` comments.

## Before

image

## After

image


pyproject.toml
```diff
@@ -155,7 +168,6 @@ exclude = [
     'rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder\.py',  # 2
     'rule_based_profiler/parameter_container\.py',  # 7
     'rule_based_profiler/rule_based_profiler\.py',  # 40
-    'tasks\.py',  # Excluded from type checking the rest of the library, may be added in a separate step.
```
Kilo59, line 158: @anthonyburdi why were we ignoring this file?
I'm not getting any errors in the CI.

We should opt to suppress the specific errors if they are unresolvable rather than ignore the whole file.

pyproject.toml
```diff
@@ -70,7 +84,6 @@ exclude = [
     'core/usage_statistics/usage_statistics\.py', # 19
     'core/usage_statistics/util\.py', # 2
     'core/util\.py', # 18
-    'docs/',  # Docs should not be type checked with the rest of the library.
```
Kilo59, line 73: @anthonyburdi 
I moved this to the ALWAYS exclude section and tried to make be visible.

pyproject.toml
```diff
@@ -25,10 +25,24 @@ ignore_missing_imports = true
...
+enable_error_code = [
+    'ignore-without-code'
+    ]
+# The following list of codes are globally ignored, do not add to this list
+disable_error_code = [
+    # annotation-unchecked are 'warning notes', not errors and won't cause mypy to fail
+    # but it does create a lot of noise in the CI mypy step.
+    # https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html
+    'annotation-unchecked'
+    ]
```
Kilo59, line 37: Just want to call out that I am globally disabling this "error" code, which we should generally never do.

But in this case, this "error" code doesn't actually result in an error anyways, so it wouldn't actually prevent any bugs without an additional configuration setting like [`check-untyped-defs`](https://mypy.readthedocs.io/en/stable/config_file.html?highlight=check-untyped-defs#confval-check_untyped_defs)
See the recent [mypy release post](https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html) section titled **Warn about Variable Annotations in Unchecked Functions** for these details.

If anyone has an issue with this let's discuss it.

Also, I initially turned on `check-untyped-defs` with the idea that if the number of errors was low, I would just fix them but... I ended up getting about `200` new errors 😬 .
I do think we should revisit this once we have better coverage of our existing "typed defs".

billdirks (approved):
LGTM!

PR #6596, opened 2022-12-16
[MAINTENANCE] Utilize a `StrEnum` for `ConfigPeer` modes

cdkini:
- Utilize a string enum for `ConfigPeer` logic

### Definition of Done

### DIFF
--- a/great_expectations/core/config_peer.py
+++ b/great_expectations/core/config_peer.py
@@ -49,18 +49,18 @@ class ConfigPeer(ABC):
 
         config: BaseYamlConfig = self.config
 
-        if mode is ConfigOutputModes.TYPED:
+        if mode == ConfigOutputModes.TYPED:
             return config
 
-        if mode is ConfigOutputModes.COMMENTED_MAP:
+        if mode == ConfigOutputModes.COMMENTED_MAP:
             return config.commented_map
 
-        if mode is ConfigOutputModes.YAML:
+        if mode == ConfigOutputModes.YAML:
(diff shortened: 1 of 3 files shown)


great_expectations/core/config_peer.py
```diff
@@ -44,26 +41,26 @@ def config(self) -> BaseYamlConfig:
...
+        mode: ConfigOutputModes = ConfigOutputModes.TYPED,
         **kwargs,
     ) -> Union[BaseYamlConfig, dict, str]:
         if isinstance(mode, str):
             mode = ConfigOutputModes(mode.lower())
 
         config: BaseYamlConfig = self.config
 
-        if mode == ConfigOutputModes.TYPED:
+        if mode is ConfigOutputModes.TYPED:
```
Kilo59: Do we want this to evaluate as True for a str equivalent?
I know that it will, I'm just trying to understand the intent here.

cdkini: ah wait yes maybe `==` is proper? I don't think this is external facing so I'd honestly prefer a plain old enum to be honest

Kilo59: I believe `is` will actually work between 2 equivalent strings (of a certain length), but that's just a side effect of some `CPython`optimizations, but it's not something we should rely on. It's the same when comparing some `int`s, but stops working after like `256`.

I think the `str` enum mixin is fine. But I would keep this an equality rather than an identity check.

Kilo59 (approved):

PR #6599, opened 2022-12-16
[DOCS] fixing wrong line reference on docs

wagneralbjr:
- I`m just fixing line reference. I think that the line reference is wrong. I think that should start at line 93 and not  line 94.


wagneralbjr:
Hello, i have signed the cla-signed. Can i fix that?

kenwade4 (approved):

Issue #6130, opened 2022-10-03
great_expectations\data_context\types\base.py -> 'Missing' Field Deprecated Warnings


PSUlion16:
As a developer running tests - warnings are given for the 'Missing' Field within the marshmallow package used by GE. Marshmallow package also throws similar errors, but is outside of scope of GE.

Impacted file, lines and associated errors are shown below.

GE Version: 0.15.26
Marshmallow Version: 3.18.0

**Impacted File: great_expectations\data_context\types\base.py**

**Lines 236, 241, 264, 360, 598, 922, 1096, 1011 Warning:** 

_RemovedInMarshmallow4Warning: The 'missing' argument to fields is deprecated. Use 'load_default' instead.
    module_name = fields.String(_

**Line 365 Warning:**

_RemovedInMarshmallow4Warning: Passing field metadata as keyword arguments is deprecated. Use the explicit `metadata=...` argument instead. Additional metadata: {'all_none': True}
    module_name = fields.String(_

AFineDayFor:
Howdy @PSUlion16 :wave: thanks for raising this with us and being a part off this stellar community :bow:

We'll raise it with the team; however, if you'd be keen on submitting a contribution that address this, we'll be excited to review it.

PSUlion16:
@AFineDayFor - I can take a crack at it - I will probably not get to this until later this week or weekend. Will let you know if I have issues.

Kilo59:
Should be resolved as part of #6602 will be part of the next release

PR #6602, opened 2022-12-19
[MAINTENANCE] filter `RemovedInMarshmallow4` warnings

Kilo59:
Resolves https://github.com/great-expectations/great_expectations/pull/6440, https://github.com/great-expectations/great_expectations/issues/6130

The deprecations discussed in the above issues are for a `RemovedInMarshmallow4` warning.
We are unlikely to upgrade to `marshmallow` `v4` for multiple reasons.
1. Our new code is using `pydantic` for schema definitions and validation.
2. The dependencies and other applications in our ecosystem also rely on `marshmallow v3` so upgrading to `v4` would also break compatibility with them.

If a package in our ecosystem does need to upgrade to `marshmallow v4` we can opt to re-vendorize the `marshmallow` which would remove the v3 constraint for any dependency resolver.

- [Filter out](https://docs.python.org/3/library/warnings.html#the-warnings-filter) the `RemovedInMarshmallow4` warnings from our codebase.
- Update our pytest configuration to error if the above warning is encountered. This ensures we properly deal with these warnings.

## Alternatives considered

1. Branching logic based on the version of marshmallow.
   a. https://github.com/great-expectations/great_expectations/pull/6440#issuecomment-1332585012
   b. Initial [`deprecation_wrappers.py` POC](https://github.com/great-expectations/great_expectations/blob/95ff3b686ab49473f73d01826d072f9b555ba95a/great_expectations/deprecation_wrappers.py)
2. Always use the non-deprecated arguments and bump the minimum `marshmallow` version.
  a. https://github.com/great-expectations/great_expectations/pull/6213 
  b. https://github.com/great-expectations/great_expectations/pull/6271


great_expectations/deprecation_wrappers.py
```diff
@@ -1,63 +0,0 @@
...
-    Wrapper around `marshmallow.fields.String` that preserves backwards compatibility on older versions.
-    """
-    _raise_for_deprecated_kwargs(DEPRECATED_FIELD_ARGS, kwargs)
-
-    if MARSHMALLOW_VERSION < MM_LOAD_DEFAULT_DEPRECATED_VERSION:
-        kwargs["missing"] = load_default
-    else:
-        kwargs["load_default"] = load_default
-    print(kwargs)
-    return field_type(**kwargs)
```
Kilo59: Decided that it's much simpler (and easier to maintain) to simply filter out the marshmallow4 deprecation warnings. We aren't moving to `marshmallow v4` anytime soon (if ever).

great_expectations/data_context/types/base.py
```diff
@@ -976,7 +982,7 @@ class Meta:
     module_name = fields.String(
         required=False,
         allow_none=True,
-        load_default="great_expectations.execution_engine",
+        missing="great_expectations.execution_engine",
```
Kilo59, line 985: Looks like we forgot to revert this use of `missing` as part of https://github.com/great-expectations/great_expectations/pull/6271

pyproject.toml
```diff
@@ -189,6 +189,8 @@ filterwarnings = [
     "ignore:Using or importing the ABCs:DeprecationWarning:moto.cloudformation.parsing",
     # This deprecation warning comes from getsentry/responses, a mocking utility for requests. It is a dependency in moto.
     "ignore:stream argument is deprecated. Use stream parameter in request directly:DeprecationWarning",
+    # We likely won't be updating to `marhsmallow` 4, these errors should be filtered out
+    "error::marshmallow.warnings.RemovedInMarshmallow4Warning"
```
Kilo59, line 193: This line will cause the tests to error if this warning is encountered.

anthonyburdi (approved):
Thank you!

Shinnnyshinshin (approved):
Thank you for this @Kilo59 🙇🏼  I think the change + note is a elegant solution

PR #6607, opened 2022-12-19
[MAINTENANCE] flake8 coverage - `render`, `juptyer_ux`, `checkpoint` sub-packages

Kilo59:
- fix `flake8` linting issues for the following packages
  - `render`
  - `jupyter_ux`
  - `checkpoint`


great_expectations/checkpoint/actions.py
```diff
@@ -1118,7 +1122,7 @@ class CloudNotificationAction(ValidationAction):
 
     def __init__(
         self,
-        data_context: "DataContext",
+        data_context: DataContext,
```
Kilo59, line 1125: This should be `AbstractDataContext` but this module isn't being type-checked yet.

I'm anticipating we may be renaming `AbstractDataContext` -> `DataContext` so I've left it as `DataContext` to prevent extra churn in this module.

If we decide against this, this will need to change.

setup.cfg
```diff
@@ -22,13 +23,9 @@ exclude = .git,
...
-    great_expectations/checkpoint,
-    great_expectations/jupyter_ux,
-    great_expectations/cli,
-    great_expectations/expectations,
-    great_expectations/dataset,
-    great_expectations/render,
-    scripts/*,
+    great_expectations/cli, # 11 warnings
+    great_expectations/expectations, # 122 warnings
+    great_expectations/dataset, # 52 warnings
```
Kilo59, line 28: Only 3 modules left.

Will have a followup PR for CLI shortly. Doing that one separately because I'll be doing type coverage as well.

great_expectations/checkpoint/actions.py
cdkini: yeah we'll have to go through and change all those instances - I focused on whatever got us across the line for the context refactor PR but there are some artifacts we need to clean up

great_expectations/render/renderer/checkpoint_new_notebook_renderer.py
```diff
@@ -132,7 +132,7 @@ def _add_sample_checkpoint_config(self) -> None:
                 sample_yaml_str,
                 lint=True,
             )
-        except:
+        except Exception:
```
cdkini, line 135: Aren't these the same thing?

cdkini (approved):

great_expectations/render/renderer/checkpoint_new_notebook_renderer.py
Kilo59: The linter doesn't like it.
I think the bare except may also catch `KeyBoardInterrupt` exceptions which doesn't inherit from `Exception`
https://docs.python.org/3/library/exceptions.html#exception-hierarchy

EDIT: yep...
https://www.flake8rules.com/rules/E722.html

>A bare except: clause will catch `SystemExit` and `KeyboardInterrupt` exceptions, making it harder to interrupt a program with Control-C, and can disguise other problems. If you want to catch all exceptions that signal program errors, use except Exception: (bare except is equivalent to except BaseException:).

PR #6610, opened 2022-12-19
[BUGFIX] Simplify metric results processing and improve detection of Decimal types in columns

alexsherstinsky:
### Scope
* Detection of which metric results can and cannot be converted to Numpy array has been simplified, thereby correcting a bug when the conversion should have been done, but was not.
* Refine detection of Decimal column types to allow for mixed float and Decimal type column values.
* Add two tests for the above fixes.

- JIRA: GREAT-1358/GREAT-1465
- JIRA: GREAT-1358/GREAT-1466
- JIRA: GREAT-1358/GREAT-1469
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/util.py
```diff
@@ -1709,14 +1711,12 @@ def convert_ndarray_float_to_datetime_tuple(
     return tuple(convert_ndarray_float_to_datetime_dtype(data=data).tolist())
 
 
-def is_ndarray_decimal_dtype(
-    data: "npt.NDArray",
-) -> TypeGuard["npt.NDArray"]:
+def is_ndarray_decimal_dtype(data: np.ndarray) -> bool:
```
Kilo59: The use of TypeGuard was correct. Can you please explain the issue you are having with it? 

We might be able to use `from __future__ import annotations` here.
And possibly a `npt.NDArray` generic, but that depends on our min version of `numpy`.

Kilo59 (dismissed):

great_expectations/rule_based_profiler/attributed_resolved_metrics.py
```diff
@@ -56,49 +52,35 @@ def _detect_illegal_array_type_or_shape(values: MetricValues) -> bool:
...
-                ):
-                    return True
-
+                values_iterator: Iterator
                 # Components of different lengths cannot be packaged into "numpy.ndarray" type (due to undefined shape).
                 if all(isinstance(value, (list, tuple)) for value in values):
-                    values_iterator: Iterator = iter(values)
-                    first_value_length: int = len(next(values_iterator))
-                    current_value: Sized[Any]
+                    values = deep_filter_properties_iterable(properties=values)
```
billdirks: Can you annotate `values`? I think it can be `None` based on the annotated return type of `deep_filter_properties_iterable` in which case `len` will die.

great_expectations/util.py
alexsherstinsky: @Kilo59 Is it still correct/appropriate/useful, given that I had to change the logic from `all()` to `any()`?  Thanks.

Kilo59: Don't Np Arrays have to be all of the same type, so if any element is a decimal they all are?

> Is it still correct/appropriate/useful, given that I had to change the logic from all() to any()? 

The short answer is yes.

The longer answer is still yes, but it could be better, we could let typing system know about the element type as well.

We will only be able to make use of the generic arrays if the version of `numpy` is > `1.21` 
Then we can declare types like as follows...
`npt.NDArray[np.float64]` 
https://numpy.org/doc/stable/reference/typing.html#numpy.typing.NDArray

great_expectations/util.py
```diff
@@ -1709,14 +1711,12 @@ def convert_ndarray_float_to_datetime_tuple(
...
     """
     Determine whether or not all elements of 1-D "np.ndarray" argument are "decimal.Decimal" type objects.
     """
     value: Any
-    result: bool = all(isinstance(value, decimal.Decimal) for value in data)
+    result: bool = any(isinstance(value, decimal.Decimal) for value in data)
```
billdirks, line 1724: Why the change from `all` to `any`? The function name `is_ndarray_decimal_dtype` seems to imply that `all` is appropriate since, if we pass this check, we expect all the values to be of the decimal dtype?

great_expectations/util.py
alexsherstinsky: @Kilo59 There are exceptions, in which case, the type is `dtype.object` -- which can happen.  This is exactly what occurs in our case:
```
[Decimal('0.0') 2.5] ; TYPE: <class 'numpy.dtype[object_]'>
```
because we have in the database, the occurrence of `Decimal` and `Float` in the same column.  Hence, this subtlety had to be handled.  (Fortunately, both are `SupportsFloat` type).  Thanks.

tests/experimental/datasources/test_integration_sql_datasource.py
```diff
@@ -22,7 +24,11 @@ def test_run_checkpoint_and_data_doc(empty_data_context, include_rendered_conten
     """
     db_file = file_relative_path(
         __file__,
-        "../../test_sets/taxi_yellow_tripdata_samples/sqlite/yellow_tripdata.db",
+        os.path.join(
```
billdirks, line 27: If you take the `os` dependency to make this os agnostic, I think you should break up the whole path since the string `test_sets/taxi_yellow_tripdata_samples/sqlite/yellow_tripdata.db` has `/`s in it.

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant_happy_paths.py
```diff
@@ -330,6 +336,144 @@ def test_sql_happy_path_onboarding_data_assistant(
...
+            "..",
+            "..",
+            "test_sets",
+            "taxi_yellow_tripdata_samples",
+            "sqlite",
+            "yellow_tripdata.db",
+        ),
+    )
+
+    # This actually works with any SQLAlchemy-compatible backend; the naming will reflect this as part of near-term fix.
```
billdirks, line 361: Could you link this ticket? https://greatexpectations.atlassian.net/browse/GREAT-1400

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant_happy_paths.py
```diff
@@ -330,6 +336,144 @@ def test_sql_happy_path_onboarding_data_assistant(
...
+            "estimator": "exact",
+            "random_seed": 2022080401,
+        },
+    )
+    assert len(result.metrics_by_domain) == 36
+    assert len(result.expectation_configurations) == 122
+
+
+@pytest.mark.integration
+@pytest.mark.slow  # 25.51 seconds
```
billdirks, line 392: Is most of the time spent in the call to `ontext.assistants.onboarding.run`?

billdirks (commented):
Thanks for fixing these bugs. I left a few comments inline.

great_expectations/util.py
billdirks: This answers my question below. Instead of testing for `any` would it be better to test that they are all one of these types?

billdirks: That is you can use `all` and pass a list of types to the `isinstance` check?

Kilo59: Were you getting any kind of mypy errors before you changed this?

If so then I'm fine with removing it and I'll revisit this later.

`TypeGuard` doesn't affect runtime behavior at all, but it's possible it was making the typing system think that if this returned false that the type wasn't an `NDArray` at all 🤔 

https://mypy.readthedocs.io/en/stable/type_narrowing.html#user-defined-type-guards

great_expectations/util.py
alexsherstinsky: @billdirks Your example results in having both.  Please see above reply to Gabriel.

tests/experimental/datasources/test_integration_sql_datasource.py
alexsherstinsky: Done -- committed.  Nice catch as I did it in the new test!  Thanks!

Kilo59: Or use `pathlib`
https://treyhunner.com/2019/01/no-really-pathlib-is-great/#Normalizing_file_paths_shouldn%E2%80%99t_be_your_concern

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant_happy_paths.py
alexsherstinsky: @billdirks We are not supposed to put JIRA links into the OSS codebase.

tests/experimental/datasources/test_integration_sql_datasource.py
alexsherstinsky: @Kilo59 I used `os` for this one.

great_expectations/rule_based_profiler/attributed_resolved_metrics.py
alexsherstinsky: @billdirks `values` is already attributed (method argument).  Thanks.

great_expectations/util.py
billdirks: I don't know if you saw my comment in the thread there. I'd prefer you change this back to `all` and update `isinstance` with a list of allowed values. It seems like 1 `Decimal` doesn't guarantee that the rest are `Decimals` or `floats`.

billdirks (changes requested):

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant_happy_paths.py
billdirks: Oh, I didn't we weren't suppose to do this. Thanks.

great_expectations/util.py
alexsherstinsky: @Kilo59 There were no errors -- I just saw the case of an increased complexity without some kind of an example (or a test) to justify it, so the benefit was unclear to me.  But after our discussion, it seems that the real benefit will come in the future, so I brought it back. 😄

great_expectations/rule_based_profiler/attributed_resolved_metrics.py
alexsherstinsky: @billdirks I think we can leave metric_value as same type, because the MetricValues in the argument covers the return type of the filter applied.  Thanks.

billdirks:
```MetricValues = Union[
    MetricValue, List[MetricValue], Set[MetricValue], Tuple[MetricValue, ...]
]```
but
```def deep_filter_properties_iterable(...) -> Union[dict, list, set, tuple, None]```
So it’s changing from something more specific a List[MetricValue] to something less specific.

Comment (billdirks) [reply great_expectations/rule_based_profiler/attributed_resolved_metrics.py]: Also notice the return value can be `None`. This means the next line can crash.

File: great_expectations/rule_based_profiler/attributed_resolved_metrics.py [hunk 7]
```diff
@@ -56,49 +52,35 @@ def _detect_illegal_array_type_or_shape(values: MetricValues) -> bool:
...
-                    return True
+                values = deep_filter_properties_iterable(properties=values)
 
+                values_iterator: Iterator
                 # Components of different lengths cannot be packaged into "numpy.ndarray" type (due to undefined shape).
                 if all(isinstance(value, (list, tuple)) for value in values):
-                    values_iterator: Iterator = iter(values)
-                    first_value_length: int = len(next(values_iterator))
-                    current_value: Sized[Any]
+                    if values is not None and len(values) > 0:
```
billdirks: You need a `None` check above line 62 since `for value in values` will crash if `values` is `None` which is one of the return types of `deep_filter_properties_iterable`

great_expectations/rule_based_profiler/attributed_resolved_metrics.py
```diff
@@ -56,49 +52,35 @@ def _detect_illegal_array_type_or_shape(values: MetricValues) -> bool:
...
+                        first_value_length: int = len(next(values_iterator))
+                        current_value: Sized[Any]
+                        if not all(
+                            len(current_value) == first_value_length
+                            for current_value in values_iterator
+                        ):
+                            return True
+
+                # Components of different types cannot be packaged into "numpy.ndarray" type (due to type mismatch).
+                if len(values) > 0:
```
billdirks: Similarly, you can't call `len` here if `values` is `None`.

great_expectations/rule_based_profiler/attributed_resolved_metrics.py
```diff
@@ -107,17 +89,9 @@ def _detect_illegal_array_type_or_shape(values: MetricValues) -> bool:
...
-            and isinstance(metric_values, (list, tuple))
-            and all(value is None for value in metric_values)
-        ):
-            return np.asarray(metric_values)
-
         return metric_values
-    else:
-        return np.asarray(metric_values)
+
+    return np.asarray(metric_values)
```
billdirks, line 92: `np.asarray(metric_values)` is not a `MetricValues` which is:
`MetricValues =
Union[
    MetricValue, List[MetricValue], Set[MetricValue], Tuple[MetricValue, ...]
]`
The function return types should be updated.

billdirks (changes requested):

great_expectations/rule_based_profiler/attributed_resolved_metrics.py
alexsherstinsky: @billdirks Looking.

alexsherstinsky: @billdirks Per our conversation -- it is included.  Thanks.

billdirks: I see, `MetricValue` has collections in it which I did not expect (and `Any`!).

billdirks (approved):
LGTM.

great_expectations/rule_based_profiler/attributed_resolved_metrics.py
```diff
@@ -37,10 +37,6 @@
```
Kilo59, line 37: This module is currently not being type-checked and has/had `4` outstanding typing warnings
Can you deal with these issues in this PR?
https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L125

Kilo59 (commented):
Can you remove the `attributed_resolved_metrics.py` module from the exclude list and deal with the 4 typing issues.

Other than that I see no issues.

https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L125

PR #6612, opened 2022-12-20
[FEATURE] Add zep datasource data assistant e2e tests.

billdirks:

Thank you for submitting!


tests/experimental/datasources/test_integration_sql_datasource.py
```diff
@@ -163,3 +165,110 @@ def test_run_checkpoint_and_data_doc(empty_data_context, include_rendered_conten
...
+        .add_sorters(["year", "month"])
+    )
+
+    # Setup an expectation suite using the onboarding data assistant
+    expectation_suite_name = "my_onboarding_assistant_suite"
+    context.create_expectation_suite(
+        expectation_suite_name=expectation_suite_name, overwrite_existing=True
+    )
+    batch_request = asset.get_batch_request(batch_request_options)
+    data_assistant_result = context.assistants.onboarding.run(
```
billdirks, line 298: @NathanFarmer How can I determine the number of batches that are part of the `data_assistant_result`? I know the results plotted by calling `data_assistant_result.pot_metrics()` has batches along the x axis so it seems like I can get this number.

great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -3,6 +3,8 @@
 import copy
 import dataclasses
 import itertools
+from collections import namedtuple
```
Kilo59: Can you use the `typing.NameTuple`?
https://docs.python.org/3/library/typing.html#typing.NamedTuple

tests/experimental/datasources/test_integration_sql_datasource.py
```diff
@@ -163,3 +165,110 @@ def test_run_checkpoint_and_data_doc(empty_data_context, include_rendered_conten
...
+        batch_request=batch_request,
+        numeric_columns_rule={
+            "estimator": "exact",
+            "random_seed": 2022080401,
+        },
+        # We exclude congestion_surcharge due to this bug:
+        # https://greatexpectations.atlassian.net/browse/GREAT-1465
+        exclude_column_names=["congestion_surcharge", ]
```
NathanFarmer: Why the trailing comma with a `list`?

great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -83,21 +89,36 @@ def param_defaults(self, data_asset: DataAsset) -> Dict[str, List]:
...
+    q = f'select min({col_name}), max({col_name}) from {table_name}'
+    min_max_dt = list(conn.execute(q))[0]
+    return DatetimeRange(min=min_max_dt[0], max=min_max_dt[1])
+
+
+def _get_sqlite_datetime_range(conn: sqlalchemy.engine.base.Connection, table_name: str,
+                                   col_name: str) -> DatetimeRange:
+    q = f"select STRFTIME('%Y%m%d', min({col_name})), STRFTIME('%Y%m%d', max({col_name})) from {table_name}"
+    min_max_dt = [datetime.strptime(dt, "%Y%m%d") for dt in list(conn.execute(q))[0]]
+    return DatetimeRange(min=min_max_dt[0], max=min_max_dt[1])
```
NathanFarmer: I hate that we are already needing to special case sqlite here. It is possible you could unify these methods by using the [sqlalchemy.func.date](https://stackoverflow.com/questions/16228195/comparing-dates-with-sqlalchemy-sqlite) built-in, as also mentioned [here](https://docs.sqlalchemy.org/en/14/dialects/sqlite.html#date-and-time-types). This would require using the ORM instead of the raw string and would negate your comment on line 106.

tests/experimental/datasources/test_integration_sql_datasource.py
```diff
@@ -163,3 +165,110 @@ def test_run_checkpoint_and_data_doc(empty_data_context, include_rendered_conten
...
+        context: DataContext,
+        sqlite_db_file_relative_path: str,
+        table_name: str,
+        date_splitter_column: str,
+        batch_request_options: Dict[str, Any]
+):
+    db_file = file_relative_path(__file__, sqlite_db_file_relative_path)
+    # Add sqlalchemy datasource.
+    # The current method is called `add_postgres` it really should be `add_sql`.
+    # See https://superconductive.atlassian.net/browse/GREAT-1400
```
NathanFarmer, line 274: I thought the plan was to separate all SQL Alchemy dialects into separate datasource classes (thus decoupling these methods from the 3 `ExecutionEngine`s). The ticket says nothing is postgres-specific, but there is something that is sqlite-specific (where I commented above). If we elevate this to `add_sql` we probably should subclass the sqlite logic. If we intend to demo postgres I would think we would want to alias `add_postgres` to `add_sql` until something postgres-specific emerges. Validation of the connection string would be one example off the top of my head of something that will be postgres-specific.

tests/experimental/datasources/test_integration_sql_datasource.py
NathanFarmer: TLDR: Alex's offline comment is also what I would use. Something like: `len(DataAssistantResult._batch_id_to_batch_identifier_display_name_map)`.

The plots are generated one at a time at the Metric/Expectation-level by iterating through `DataAssistantResult.metrics_by_domain`. I don't think we implemented a clean method that does what you are wanting for the Data Assistant. Each domain should (in theory) have the same number of batch-metric values computed, but if a metric computation fails they could be different.

Kilo59: @NathanFarmer @alexsherstinsky 
Probably not something to fix in this PR, but why doesn't `DataAssistantResult` implement `__len__`?

Kilo59 (approved):

Kilo59:
Where appropriate can we push improvements to follow-up tasks? I would like to build on the work done in this PR.

great_expectations/experimental/datasources/postgres_datasource.py
billdirks: Done.

great_expectations/experimental/datasources/postgres_datasource.py
billdirks: I am going to break up this class into a SQL datasource and Postgres datasource as my next bit of work. I could also add a sqlite one so we don't branch here.

This is my concern about using the ORM. I haven't used sqlalchemy very much so if this isn't an issue let me know. I don't know the type of `col_name`. When I specify the ORM model can I use a sqlalchemy`date` type even if its a `datetime` or other type at the database level? I'm also assuming my model and skip columns that are present in the original table? I hope to sidestep these questions by making specific SQL datasources as described above but if you have experience with this let me know.

tests/experimental/datasources/test_integration_sql_datasource.py
billdirks: Yes, there will be a generic `add_sql` method for generic sql engines and then specific engines for postgres and sqlite. I will do this after this PR is in.

I only added the sqlite branch in this pr because I was using sqlite to test. It maybe would have been better to setup postgres to test but this was easier and worked in previous PRs.

tests/experimental/datasources/test_integration_sql_datasource.py
billdirks: Removed. I though I was going to add more columns but then I didn't need to and I hadn't removed the comma.

tests/experimental/datasources/test_integration_sql_datasource.py
NathanFarmer: @Kilo59 It isn't clear what the `__len__` would represent to users. Number of metrics computed? Number of Expectations suggested? Number of domains? Number of batches per domain?

PR #6613, opened 2022-12-20
[FEATURE] Accept a `pathlib.Path` `context_root_dir`

Kilo59:
- `get_context()` and other DataContexts can accept a `pathlib.Path` objects in addition to `str` paths.

This brings `great_expecations` in line with the rest of the Python standard library, which since python `3.6` accepts a `pathlib.Path` or [`Pathlike` object](https://docs.python.org/3/library/os.html#os.PathLike) in place of any `str` filepath representation.

I went with a minimally invasive approach for this and simply convert any `pathlib.Path` objects to `str` before assigning to an attribute.

The better, longer-term approach would be to convert all `str` path representations to a `pathlib.Path` and pass the richer `Path` object around.

https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/
https://treyhunner.com/2019/01/no-really-pathlib-is-great/
https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

### Definition of Done

- [x] I have ~added~ updated [unit tests](https://docs.greatexpectations.io/docs/contributing/contributing_test#writing-unit-and-integration-tests) where applicable and made sure that new and existing tests are passing.


billdirks (commented):
Thanks for doing this. It seems if we always want `Union[str, pathlib.Path]` for these arguments we should make a `TypeAlias` to make this a little more concise and so we have a concrete thing to point to if new path arguments get added to the codebase. Could you make this `TypeAlias` (maybe in `great_expectations/types/`?) and then use that throughout this PR?

Kilo59:
@billdirks Good idea will do.

great_expectations/alias_types.py
```diff
@@ -0,0 +1,9 @@
+"""This module contains shared TypeAliases"""
```
Kilo59, line 1: used `alias_types.py` instead of `types.py` to avoid weirdness with standard library module names, and to prevent this from being polluted with random types + class definitions.

billdirks (approved):
LGTM, thanks!

PR #6614, opened 2022-12-20
[BUGFIX] Stop overwriting query with static string in RuntimeBatchRequests for SQL

### LINKED ISSUE #4970: Display content of SQL query in validation result doc
**Is your feature request related to a problem? Please describe.**
When I use a RuntimeBatchRequest to define the BigQuery data I want to read with a SQL query, there's no way for me to see what query I ran (and therefore what data was used) in the validation results doc. 

**Describe the solution you'd like**
Instead of `SQLQuery` displayed in the `query` field in the validation doc, I'd like to see the actual text of the query run, e.g. `SELECT * FROM <mytable> ...`. 
Screen Shot 2022-04-08 at 17 02 05

**Describe alternatives you've considered**
The status quo (displaying `SQLQuery` and not the query run) is an alternative. This works okay for adhoc use, e.g. if a user has just run a validation manually and immediately reads the validation docs, but in any other situation it would be really helpful to see the query used.

**Additional context**
Using RuntimeBatchRequest with SQL queries seems like a really powerful way to use BQ data, but not being able to see which query was used for a validation would make it hard for us to use in production.

talagluck:

- Previously, we were overwriting the original query in the BatchSpec with a hard-coded string in the case of a RuntimeBatchRequest. This meant that there was no record of the original query in the validation results nor in Data Docs. 
- This removes that overwrite
- Closes #4970

### Definition of Done

Thank you for submitting!

### DIFF
--- a/great_expectations/execution_engine/sqlalchemy_execution_engine.py
+++ b/great_expectations/execution_engine/sqlalchemy_execution_engine.py
@@ -1225,7 +1225,6 @@ class SqlAlchemyExecutionEngine(ExecutionEngine):
             # query != None is already checked when RuntimeQueryBatchSpec is instantiated
             query: str = batch_spec.query
 
-            batch_spec.query = "SQLQuery"
             batch_data = SqlAlchemyBatchData(
                 execution_engine=self,
                 query=query,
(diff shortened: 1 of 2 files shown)


NathanFarmer (approved):
LGTM!

PR #6609, opened 2022-12-19
[MAINTENANCE] update pip installs in pipelines

kenwade4:
Changes proposed in this pull request:
- Remove api-docs-test from extras_require dict and add reqs to 'test' key
- Update pipelines to use versions of pytest-xxx packages from the reqs files
- Consolidate 'pip install' statements throughout the pipelines and use contrib tools from reqs

PR #6621, opened 2022-12-21
[MAINTENANCE] create cache as part of `--ci` type-checking step

Kilo59:
- Fix static type checking step
  - always create the `mypy` cache directory if it does not exist when running in CI
- add a `--ci` flag to `invoke type-check` and use it in pipelines

Not sure how, but this seems to have started after https://github.com/great-expectations/great_expectations/pull/6609

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -143,10 +143,27 @@ def type_check(
     daemon=False,
     clear_cache=False,
     report=False,
+    ci=False,
 ):
     """Run mypy static type-checking on select packages."""
+    mypy_cache = pathlib.Path(".mypy_cache")
+
+    if ci:
+        mypy_cache.mkdir(exist_ok=True)
+
+        type_check(
+            ctx,
+            packages,
+            install_types=True,
+            pretty=pretty,
+            warn_unused_ignores=True,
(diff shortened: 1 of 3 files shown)


tasks.py
```diff
@@ -143,10 +143,28 @@ def type_check(
     daemon=False,
     clear_cache=False,
     report=False,
+    ci=False,
 ):
     """Run mypy static type-checking on select packages."""
+    mypy_cache = pathlib.Path(".mypy_cache")
+
+    if ci:
+        mypy_cache.mkdir(exist_ok=True)
```
NathanFarmer, line 152: ❤️

tasks.py
```diff
@@ -143,10 +143,28 @@ def type_check(
     daemon=False,
     clear_cache=False,
     report=False,
+    ci=False,
 ):
     """Run mypy static type-checking on select packages."""
+    mypy_cache = pathlib.Path(".mypy_cache")
```
NathanFarmer, line 149: Love the `pathlib` usage.

NathanFarmer (approved):

PR #6623, opened 2022-12-21
[FEATURE] F/great 1400/sql datasource

billdirks:
This renames the current postgres datasource to a sql one and subclasses it to make a postgres specific one. I ran into one issue where, since the `TableAsset` is shared between the super and subclass we attempted to register it twice so I short circuit identical type registrations.

I plan to differentiate the sqlite behavior into a subclass in a subsequent PR by introducing a sqlite datasource class. I also plan on updating tests to reflect the super/subclass. I can do that in this PR or a followup but wanted to get this out because this rename will likely impact other work.


great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -0,0 +1,507 @@
+from __future__ import annotations
```
billdirks, line 1: This is a rename of the old `postgres_datasource.py` to `sql_datasource.py`. I also renamed `PostgresDatasource` to `SQLDatasource`, `PostgresDatasourceError` to `SqlDatasourceError`,  and the type literal from `postgres` to the `str` `sql`, ie `type: str = "sql"`. I couldn't make it the literal sql because it can be overridden by a subtype such as the postgres one. If you know of a better way to do this, let me know.

great_expectations/experimental/datasources/type_lookup.py
```diff
@@ -67,6 +67,9 @@ def __delitem__(self, key: ValidTypes):
         super().pop(value, None)
 
     def __setitem__(self, key: ValidTypes, value: ValidTypes):
+        if key in self and value in self and self[key] == value and self[value] == key:
```
billdirks, line 70: Subclasses try to register the associated asset types in the superclass. I allow this and return early when this is done.

great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -1,507 +1,7 @@
...
-    """
-
-    # class var definitions
-    asset_types: ClassVar[List[Type[DataAsset]]] = [TableAsset]
-
-    # right side of the operator determines the type name
-    # left side enforces the names on instance creation
+class PostgresDatasource(SQLDatasource):
     type: Literal["postgres"] = "postgres"
-    connection_string: str
```
Kilo59, line 452: Now that we have a general SQL datasource type.
Can we use the `PostgresDsn` type for `connection_string`?

https://docs.pydantic.dev/usage/types/#urls

```python
from pydantic import PostgresDsn

class PostgresDatasource(SQLDatasource):
    type: Literal["postgres"] = "postgres"
    connection_string: PostgresDsn
```

Kilo59 (approved):
The `PostgresDsn` might require an extra step to convert back to string before passing it to the execution engine.

We can discuss more and do this as a followup.

great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -1,507 +1,7 @@
...
-            are TableAsset objects.
-    """
-
-    # class var definitions
-    asset_types: ClassVar[List[Type[DataAsset]]] = [TableAsset]
-
-    # right side of the operator determines the type name
-    # left side enforces the names on instance creation
+class PostgresDatasource(SQLDatasource):
     type: Literal["postgres"] = "postgres"
```
Kilo59:
```
great_expectations/experimental/datasources/postgres_datasource.py:7: error:
Incompatible types in assignment (expression has type "Literal['postgres']",
base class "SQLDatasource" defined the type as "Literal['sql']")  [assignment]
        type: Literal["postgres"] = "postgres"

```
I think we just ignore this.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -448,7 +448,7 @@ class SQLDatasource(Datasource):
 
     # right side of the operator determines the type name
     # left side enforces the names on instance creation
-    type: Literal["sql"] = "sql"
```
Kilo59: We need the `Literal` type

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -0,0 +1,507 @@
...
+        assets: An optional dictionary whose keys are table asset names and whose values
+            are TableAsset objects.
+    """
+
+    # class var definitions
+    asset_types: ClassVar[List[Type[DataAsset]]] = [TableAsset]
+
+    # right side of the operator determines the type name
+    # left side enforces the names on instance creation
+    type: str = "sql"
```
Kilo59:
```suggestion
    type: Literal["sql"] = "sql"
```
Need to pair this with a `# type: ignore[assignment]` for the mypy error

PR #6629, opened 2022-12-22
[MAINTENANCE] Enrich `RendererConfiguration` primitive types

NathanFarmer:
- GREAT-1475
- Rename `Param` classes to `Value` as they generalize to other `RendererConfiguration` attributes such as `header_row` and `table`
- Create `RendererSchema` `TypedDict` and use throughout the codebase
- Create `RendererTableValue` `pydantic` model for use in `header_row` and `table`
- Create `RendererConfiguration._RendererParamArgs` `TypedDict` for better annotations of internal dictionary unpacking

### Definition of Done


great_expectations/render/renderer_configuration.py
```diff
@@ -141,20 +152,18 @@ class _RendererParamBase(BaseModel):
...
             arbitrary_types_allowed = True
 
         @root_validator(pre=True)
-        def _validate_schema_matches_value(cls, values: dict) -> dict:
-            schema_type: RendererSchemaType = values["schema"]["type"]
+        def _validate_param_type_matches_value(cls, values: dict) -> dict:
+            param_type: RendererValueType = values["schema"]["type"]
             value: Any = values["value"]
-            if schema_type is RendererSchemaType.STRING:
+            if param_type is RendererValueType.STRING:
```
Kilo59:
```suggestion
            if param_type == RendererValueType.STRING:
```

You should use an equality check here unless you intended this to be false for an equivalent string.
Which I don't think is a good idea.

Often using `is` will work, but it's dangerous to rely on it.

https://stackoverflow.com/a/1504848/6304433

Same comment for the other string `is` checks below.

NathanFarmer: @Kilo59 I went ahead and changed it for added safety, but I think pydantic should already be enforcing this.

Kilo59 (approved):

great_expectations/render/components.py
```diff
@@ -627,8 +630,8 @@ def __init__(
         header: Optional[RenderedAtomicValue] = None,
         template: Optional[str] = None,
         params: Optional[dict] = None,
-        header_row: Optional[List[RenderedAtomicValue]] = None,
-        table: Optional[List[List[RenderedAtomicValue]]] = None,
+        header_row: Optional[List[RendererTableValue]] = None,
+        table: Optional[List[List[RendererTableValue]]] = None,
```
Kilo59, line 634: Would it make things easier if these could never be `None` but could be empty lists?

great_expectations/render/components.py
```diff
@@ -640,8 +643,8 @@ def __init__(
         self.params: Optional[dict] = params
 
         # TableType
-        self.header_row: Optional[List[RenderedAtomicValue]] = header_row
-        self.table: Optional[List[List[RenderedAtomicValue]]] = table
+        self.header_row: Optional[List[RendererTableValue]] = header_row
+        self.table: Optional[List[List[RendererTableValue]]] = table
```
Kilo59, line 647: Would it make things easier if these defaulted to empty lists?
```suggestion
        self.header_row: List[RendererTableValue] = header_row or []
        self.table: List[List[RendererTableValue]] = table or []
```

Kilo59:
@NathanFarmer 
You might also consider adding type-checking to just one of these expectation modules, just as a sanity check for all the render components typing work.

great_expectations/render/components.py
NathanFarmer: I tried this but the `None` indicates that the key should not be serialized. Otherwise we serialize the empty list.

PR #6633, opened 2022-12-22
[MAINTENANCE] Improve type checking for Expectations with atomic renderers leveraging `RendererConfiguration`

NathanFarmer:
- GREAT-1478
- Updated existing renderers that use `RendererConfiguration` due to static type check errors
- Fixed additional Expectation-focused static type check errors

### Definition of Done

[GREAT-1478]: https://greatexpectations.atlassian.net/browse/GREAT-1478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pyproject.toml
```diff
@@ -86,7 +86,7 @@ exclude = [
     'execution_engine/split_and_sample/sparkdf_data_splitter\.py',  # 4
     'execution_engine/split_and_sample/sqlalchemy_data_sampler\.py',  # 10
     'execution_engine/split_and_sample/sqlalchemy_data_splitter\.py',  # 22
-    'expectations/core/expect_column_', # 214
+    'expectations/core/expect_column_', # 147
     'expectations/core/expect_compound_columns_to_be_unique\.py', # 3
```
Kilo59, line 89: If it makes things easier, you could replace this with a more specific regex, rather than trying to fix all `214` warnings at once.

NathanFarmer: If it works for you, I would prefer to do this in the next PR, since I expect this inventory to change drastically from this PR to the next.

contrib/great_expectations_semantic_types_expectations/great_expectations_semantic_types_expectations/expectations/expect_column_values_to_be_valid_price.py
```diff
@@ -92,7 +92,7 @@ class ExpectColumnValuesToBeValidPrice(ColumnMapExpectation):
     default_kwarg_values = {}
 
     def validate_configuration(
-        self, configuration: Optional[ExpectationConfiguration]
+        self, configuration: Optional[ExpectationConfiguration] = None
```
cdkini, line 95: Why this change? Is there ever a situation where this is called without an input config?

tests/integration/docusaurus/expectations/creating_custom_expectations/expect_multicolumn_values_to_be_multiples_of_three.py
```diff
@@ -121,15 +121,12 @@ def validate_configuration(
...
-                    "column_list" in configuration.kwargs
-                ), "column_list must be provided"
-            except AssertionError as e:
-                raise InvalidExpectationConfigurationError(str(e))
+        configuration = configuration or self.configuration
+
+        try:
+            assert "column_list" in configuration.kwargs, "column_list must be provided"
+        except AssertionError as e:
+            raise InvalidExpectationConfigurationError(str(e))
```
cdkini, line 129: Instead of an assert`, can't we just do:
```
if "column_list" not in configuration.kwargs:
    raise InvalidExpectationConfigurationError("column_list_must_be_provided")
```

great_expectations/expectations/expectation.py
```diff
@@ -2413,10 +2413,10 @@ def validate_configuration(
 
 
 class ColumnMapExpectation(TableExpectation, ABC):
-    map_metric = None
+    map_metric = ""
```
cdkini: This change won't be an issue right? Seems to be a base for a lot of classes

great_expectations/expectations/expectation.py
```diff
@@ -2285,7 +2285,7 @@ def _validate_metric_value_between(
         else:
             below_max = True
 
-        success = above_min and below_max
+        success = bool(above_min and below_max)
```
cdkini, line 2288: Is this just for readability or was it required by a linter? I think its assumed to be a bool based on the conditional no?

contrib/great_expectations_semantic_types_expectations/great_expectations_semantic_types_expectations/expectations/expect_column_values_to_be_valid_price.py
NathanFarmer, line 95: In the entire class hierarchy this is what the signature of this method looks like, except for sometimes, at the child Expectation-level. The super always gets called at the beginning of this method and at every level it checks if `configuration` is `None` and if so, use `self.configuration`.

great_expectations/expectations/core/expect_column_values_to_not_be_in_set.py
```diff
@@ -143,16 +142,16 @@ def _prescriptive_template(
...
         renderer_configuration: RendererConfiguration,
     ) -> RendererConfiguration:
-        add_param_args = (
+        add_param_args: List[AddParamArgs] = [
             ("column", RendererValueType.STRING),
             ("value_set", RendererValueType.ARRAY),
             ("mostly", RendererValueType.NUMBER),
             ("parse_strings_as_datetimes", RendererValueType.BOOLEAN),
-        )
+        ]
```
Kilo59, line 150: All things being equal, I think. a tuple would be better since it's immutable.

contrib/great_expectations_semantic_types_expectations/great_expectations_semantic_types_expectations/expectations/expect_column_values_to_be_valid_price.py
NathanFarmer: I'm not aware of the workflow where this could be none, but I think the pattern should remain the same, especially given the `Optional` annotation.

great_expectations/expectations/core/expect_column_value_lengths_to_be_between.py
```diff
@@ -278,30 +276,30 @@ def _prescriptive_template(
...
-        add_param_args = (
+        add_param_args: List[AddParamArgs] = [
             ("column", RendererValueType.STRING),
             ("min_value", [RendererValueType.NUMBER, RendererValueType.DATETIME]),
             ("max_value", [RendererValueType.NUMBER, RendererValueType.DATETIME]),
             ("mostly", RendererValueType.NUMBER),
             ("strict_min", RendererValueType.BOOLEAN),
             ("strict_max", RendererValueType.BOOLEAN),
-        )
+        ]
```
Kilo59, line 286: Similar question as below. Is there a reason we need a `list`? Tuples are immutable and use less memory.

great_expectations/expectations/core/expect_column_unique_value_count_to_be_between.py
```diff
@@ -213,30 +213,30 @@ def _prescriptive_template(
         cls,
         renderer_configuration: RendererConfiguration,
     ) -> RendererConfiguration:
-        add_param_args = (
+        add_param_args: List[AddParamArgs] = [
```
Kilo59, line 216: tuple vs list?

great_expectations/expectations/core/expect_column_stdev_to_be_between.py
```diff
@@ -214,30 +214,30 @@ def _prescriptive_template(
         cls,
         renderer_configuration: RendererConfiguration,
     ) -> RendererConfiguration:
-        add_param_args = (
+        add_param_args: List[AddParamArgs] = [
```
Kilo59, line 217: tuple vs list?

great_expectations/expectations/core/expect_column_values_to_be_unique.py
```diff
@@ -104,64 +112,35 @@ def validate_configuration(
...
-            },
-            "condition_parser": {
-                "schema": {"type": "string"},
-                "value": params.get("condition_parser"),
-            },
-        }
+        renderer_configuration: RendererConfiguration,
+    ) -> RendererConfiguration:
+        add_param_args: List[AddParamArgs] = [
+            ("column", RendererValueType.STRING),
```
cdkini, line 120: Could the first value of this tuple become an enum as well? Not necessary in this PR but just asking!

docs/guides/expectations/creating_custom_expectations/how_to_create_custom_column_aggregate_expectations.md
```diff
@@ -322,7 +322,7 @@ This guide will leave you with a Custom Expectation sufficient for [contribution
 
 If you plan to contribute your Expectation to the public open source project, you should update the `library_metadata` object before submitting your [Pull Request](https://github.com/great-expectations/great_expectations/pulls). For example:
 
-```python file=../../../../examples/expectations/column_aggregate_expectation_template.py#L105-L110
+```python file=../../../../examples/expectations/column_aggregate_expectation_template.py#L104-L109
```
cdkini, line 325: Appreciate all the changes around here - I know these line number refs are a bit annoying

Kilo59 (approved):
Looks good to me.
I would consider reverting the `tuple` -> `list` changes unless there's an obvious benefit that outweighs the memory and immutable benefit of `tuples.

great_expectations/expectations/expectation.py
```suggestion
    map_metric: str = ""
```

Should we say this must be a string?

NathanFarmer: I ended up changing this to `map_metric: Optional[str] = None` to make the type checker happy, because you were right that this is `None` vs `str` specific logic in this class.

great_expectations/expectations/expectation.py
NathanFarmer: Nice catch! This was an unrelated bugfix for @alexsherstinsky due to issues when the variables are `np.bool`.

great_expectations/expectations/expectation.py
NathanFarmer: I ended up needing to revert this due to some `isinstance` checks on `str` vs `None` with error messages in this class.

great_expectations/expectations/core/expect_column_values_to_be_unique.py
NathanFarmer: I don't think it would work like that. These strings usually map the Expectation kwargs and are one of the only Expectation-specific things in the renderer methods.

great_expectations/expectations/core/expect_column_values_to_not_be_in_set.py
NathanFarmer: I think I made this change because I forgot you can annotate a tuple of unknown length `Tuple[AddParamArgs, ...]`. I will make this change in the next PR since I need to change some of it as well.

tests/integration/docusaurus/expectations/creating_custom_expectations/expect_multicolumn_values_to_be_multiples_of_three.py
NathanFarmer: I would prefer to punt on this since this not really in the scope of this PR and is repeated many places. Also, once you change it in those places, you have to update all of the line numbers in the docs again!

great_expectations/expectations/expectation.py
Kilo59: That sounds like incorrect logic.
An empty string shouldn't be a valid `map_metric`??

Kilo59: Seems like that `str` vs `None` logic might be wrong.
I would expect an empty string would be invalid.

PR #6639, opened 2022-12-23
[MAINTENANCE] CI - install from `requirements-types.txt`

Kilo59:
- add new `requirements-types.txt`
  - always install from the above req file for all type-checking steps
  - always install all core `great_expectations` dependencies 
- update requirements files
- fix/ignore regressions

This should fix typing error related to `black` on develop, and ensure that the type-check step on `develop` and `PR`'s is consistent.


reqs/requirements-dev-contrib.txt
```diff
@@ -4,8 +4,4 @@ invoke>=1.7.1
 isort==5.10.1
 mypy==0.991
 pre-commit>=2.6.0
-pydantic>=1.0,<2.0
-pytest-cov>=2.8.1
-pytest-order>=0.9.5
-pytest-random-order>=1.0.4
```
Kilo59, line 10: I don't know why these pytest plugins were even in this file. Seems like they don't belong here.

I don't know why they really belong in `reqs/requirements-dev-lite.txt` either but at least that file has other testing dependencies.
🤷

requirements-types.txt
```diff
@@ -0,0 +1,2 @@
+--requirement requirements.txt
+--requirement reqs/requirements-dev-contrib.txt
```
Kilo59, line 2: I would put this in `/reqs` but I need to be able to install the core `gx` requirements.

I tried just putting `.` which should have just installed the python package in the CWD but I got a `pip` error. (I thought this used to allowed 🤔 ).

I will see if I can do something like...
```
--requirements ../requirements.txt
```

Kilo59: Ok that actually worked 😄

Kilo59: It "worked" but reverted because it failed on `import ge` step. 
`setup.py` `get_extras_require()` could not parse the `../requirements.txt` line.
Didn't seem worth fixing. And is probably undesired behavior as this would create an additional `extras` set.

anthonyburdi: In the pipeline you could just do two steps, `pip install -r requirements.txt` and then `pip install reqs/requirements-types.txt`. That would be more consistent I think.

reqs/requirements-dev-contrib.txt
anthonyburdi: Is `reqs/requirements-dev-contrib.txt` used in other pipelines? Looks like `azure-pipelines-contrib.yml` in the linting stage - are any of these dependencies required for that? (I don't think so but haven't looked deeply).

great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -90,7 +90,6 @@ class GXDependencies:
             "openpyxl",
             "pre-commit",
             "psycopg2-binary",
-            "pydantic",  # dev dependency because of mypy plugin
```
anthonyburdi, line 93: Is this removed because pydantic is a required dependency?

anthonyburdi (commented):
Just some questions before I approve but overall looks good.

great_expectations/core/usage_statistics/package_dependencies.py
Kilo59: Correct, before our required dependencies weren't being installed as part of the type-checking step, so it was also added here so that we would have the pydantic type info.

Now we do install all the required dependencies in the type-checking step, so this is no longer needed.

requirements-types.txt
Kilo59: We could but I would prefer this as requiring two different install commands makes it easier for the different type-checking checks to get out of sync again.

reqs/requirements-dev-contrib.txt
Kilo59: Yes, but the `linting` step doesn't need these `pytest` plugins. I think having them in this file in the first place may have been a mistake.

anthonyburdi (approved):
Thanks for following up, LGTM!

reqs/requirements-dev-contrib.txt
kenwade4: I think that the dev-lite was meant to be the least amount of packages needed to run tests locally. Those plugins aren't needed for a local run, so they were added to the contrib file

requirements-types.txt
kenwade4: If this requirements-types.txt file is just going to be the base package requirements and those from contrib, I think it makes sense to allow contrib to be an extras_requires key (instead of popping it off) and then do `pip install ".[contrib]"

Just curious, why is the full requirements.txt needed in pipeline stages where doing `pip install --requirement requirements-types.txt`?

Kilo59: >Just curious, why is the full requirements.txt needed in pipeline stages where doing pip install --requirement requirements-types.txt?

The typing checker needs those dependencies to do its type-checking.
For example in order to know whether `sqlalchemy` types and methods are being used correctly, it needs `sqlalchemy` installed.

reqs/requirements-dev-contrib.txt
Kilo59: @kenwade4 Thanks for the explanation. That makes sense.

requirements-types.txt
Kilo59: >If this requirements-types.txt file is just going to be the base package requirements and those from contrib, I think it makes sense to allow contrib to be an extras_requires key (instead of popping it off) and then do `pip install ".[contrib]"

Ah okay, so we could have the static type-checking step be...
```
pip install .[contrib]. # or .[types] or something
invoke --ci
```
And that would always work unless something was broken in the package install.
Although it would be slower, and if the package install does break, the type-checker could actually be helpful there 🤔 

I think I would like a requirement file and/or extra package that is explicit about it being for `types`/`type-checking` and not ONLY as part of a general `contrib` collection.

PR #6617, opened 2022-12-20
[MAINTENANCE] `mypy` typing for `core/util.py`

Shinnnyshinshin:
- Typing for `core/util.py`
- `pyproject.toml` has been updated

PR #6640, opened 2022-12-23
[MAINTENANCE] fix remaining core/utils.py typing issues

Kilo59:
This should fix the union type errors for code that calls into the json converter function.
We now use `overloads` to indicate the specific JSON-compatible primitive that will be returned.

Merging into https://github.com/great-expectations/great_expectations/pull/6617


Kilo59:
The `IPython` type error is fixed on develop

Shinnnyshinshin (approved):
Thank you so much for this @Kilo59

PR #6644, opened 2022-12-23
[MAINTENANCE] M/great 1225/async builds branch

billdirks:
Currently our async pipeline always builds develop. However, sometime we need to debug a branch. Currently, we can only do that by making temporary changes to the async pipeline. If you forget to do that you waste time since it build develop. You also must remember to revert the change. This PR makes the async pipeline run the current branch so we no longer need to edit the async pipeline to make it run on the branch we want.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -132,11 +132,22 @@ stages:
       - job: make_suffix
         steps:
           - bash: |
+              # Create a docker allowed image name from the branch name
+              eval `python - <<EOF
+              import os
+              import re
+              branch_name = os.environ["BRANCH_NAME"]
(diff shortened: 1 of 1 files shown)


alexsherstinsky (approved):
@billdirks approving -- just wanted to say that we can name the branch with the prefix `pre_pr` -- and it will get built (I added a trigger for it a while ago, not sure if it is still there).  Your mechanism may be better.  Could you please provide instructions for how to use it?  Thanks!

billdirks:
@alexsherstinsky Thanks for the review! I am looking now and see the `pre_pr` trigger on the PR pipeline. I don't see it on the async pipeline. It may have been there and went away when these tests were dockerized. Before this change the async pipeline would always build and run the docker tests on `develop` regardless of the branch that pushed the build. I'll write something up about how to run the tests.

PR #6656, opened 2022-12-28
[FEATURE] Ensure Pandas ID/PK can use named indices

Shinnnyshinshin:
- Allows `named_indices` to be used for Pandas ID/PK.
- Refactored `_pandas_map_condition_index()` method to allow for
    1. single named index in Pandas DF
    2. multiple named indices Pandas DF.
    4. no named indices but `unexpected_index_column_names` are defined. 
    5. no named indices and no `unexpected_index_column_names` (just return default indices as before)
- Note: In cases where both named indices exist and `unexpected_index_column_names` are defined, we will default to outputting columns that are defined in `unexpected_index_column_names`. If there is a conflict (ie. we define a column that doesn't exist as a named index), then we will raise a `MetricsResolution` error. 
- Tests have been added at the `tests/checkpoint/test_checkpoint_result_format.py`-level. 
- Tests have been added at the `test_metrics_util.py`-level for the helper methods. 

#### What are named indices?
* [Here is a great resource. The method `set_index()` was used in tests that exercise named indices behavior](https://www.w3resource.com/pandas/dataframe/dataframe-set_index.php)
#### DataFrame with default index
Screen Shot 2022-12-28 at 9 40 23 PM

* If `unexpected_index_column_names` are passed in 
```python
{
"result_format": {
   "result_format": "COMPLETE",
    "unexpected_index_column_names": ["pk_2"],
   }
}
# we will return 
[
   {"animals": "giraffe", "pk_2": "three"},
   {"animals": "lion", "pk_2": "four"},
   {"animals": "zebra", "pk_2": "five"},
]
```
* If `unexpected_index_column_names` are **not** passed in 
```python
{
"result_format": {
   "result_format": "COMPLETE",
   }
}
# we will return the default indices 
[ 3, 4, 5] 
```

####  DataFrame with named index `pk_1`
Screen Shot 2022-12-28 at 9 40 55 PM
* If `unexpected_index_column_names` are passed in 

```python 
{
"result_format": {
   "result_format": "COMPLETE",
   }
}
# we will return 
[
   {"animals": "giraffe", "pk_1": 3},
   {"animals": "lion", "pk_1": 4},
   {"animals": "zebra", "pk_1": 5},
]
```
* If non existent `unexpected_index_column_names` is passed in 
```python
{
"result_format": {
   "result_format": "COMPLETE",
    "unexpected_index_column_names": ["i_dont_exist"],
   }
}

# this will give an ERROR
```

#### DataFrame with named index `pk_1` and `pk_2`
Screen Shot 2022-12-28 at 9 41 05 PM

* If 1 `unexpected_index_column_names` is passed in 

```python 
{
"result_format": {
   "result_format": "COMPLETE",
   }
}
# we will return 
[
   {"animals": "giraffe", "pk_1": 3},
   {"animals": "lion", "pk_1": 4},
   {"animals": "zebra", "pk_1": 5},
]
```
* If 2 `unexpected_index_column_names` are passed in 
```python 
{
"result_format": {
   "result_format": "COMPLETE",
    "unexpected_index_column_names": ["pk_1", "pk_2"],
   }
}
# we will return 
[
   {"animals": "giraffe", "pk_1": 3, "pk_2": "three"},
   {"animals": "lion", "pk_1": 4, "pk_2": "four"},
   {"animals": "zebra", "pk_1": 5, "pk_2": "five"},
]
```
* If non existent `unexpected_index_column_names` is passed in 
```python
{
"result_format": {
   "result_format": "COMPLETE",
    "unexpected_index_column_names": ["i_dont_exist"],
   }
}

# this will give an ERROR
```

### Definition of Done

### DIFF
--- a/great_expectations/expectations/metrics/map_metric_provider.py
+++ b/great_expectations/expectations/metrics/map_metric_provider.py
@@ -1783,79 +1783,58 @@ def _pandas_map_condition_index(
 
     column_name: Union[str, quoted_name]
 
-    unexpected_index_list: Optional[List[Dict[str, Any]]] = []
     expectation_domain_column_name: str = domain_kwargs["column"]
 
-    # if we have defined unexpected_index_column_names
-    if "unexpected_index_column_names" in result_format:
-        unexpected_index_column_names: List[str] = result_format[
+    # one named index
+    if df.index.name is not None:
+        unexpected_index_column_names: List[str] = result_format.get(
+            "unexpected_index_column_names", [df.index.name]
+        )
+        unexpected_index_list: Optional[
+            List[Dict[str, Any]]
+        ] = get_unexpected_indices_for_single_pandas_named_index(
+            df,
+            expectation_domain_column_name,
+            unexpected_index_column_names,
+        )
+    # multiple named indices
+    elif df.index.names[0] is not None:
+        unexpected_index_column_names: List[str] = result_format.get(
+            "unexpected_index_column_names", list(df.index.names)
+        )
+        unexpected_index_list: Optional[
+            List[Dict[str, Any]]
+        ] = get_unexpected_indices_for_multiple_pandas_named_indices(
+            df,
+            expectation_domain_column_name,
+            unexpected_index_column_names,
+        )
+    # named columns
+    elif result_format.get("unexpected_index_column_names"):
+        unexpected_index_column_names: List[str] = result_format.get(
             "unexpected_index_column_names"
-        ]
-        if df.index.name is not None:
-            unexpected_index_list = (
-                get_unexpected_indices_for_single_pandas_named_index(
-                    df,
-                    unexpected_index_list,
-                    expectation_domain_column_name,
-                    unexpected_index_column_names,
+        )
+        unexpected_index_list: Optional[List[Dict[str, Any]]] = []
+        unexpected_indices: List[Union[int, str]] = list(df.index)
+        for index in unexpected_indices:
+            primary_key_dict: Dict[str, Any] = dict()
+            primary_key_dict[expectation_domain_column_name] = df.at[
+                index, expectation_domain_column_name
+            ]
+            for column_name in unexpected_index_column_names:
+                column_name = get_dbms_compatible_column_names(
+                    column_names=column_name,
+                    batch_columns_list=metrics["table.columns"],
+                    execution_engine=execution_engine,
+                    error_message_template='Error: The unexpected_index_column "{column_name:s}" does not exist in Dataframe. Please check your configuration and try again.',
                 )
-            )
-        elif df.index.names[0] is not None:
-            unexpected_index_list = (
-                get_unexpected_indices_for_multiple_pandas_named_indices(
-                    df,
-                    unexpected_index_list,
-                    expectation_domain_column_name,
-                    unexpected_index_column_names,
-                )
-            )
-        else:
-            unexpected_indices: List[Union[int, str]] = list(df.index)
-            for index in unexpected_indices:
-                primary_key_dict: Dict[str, Any] = dict()
-                primary_key_dict[expectation_domain_column_name] = df.at[
-                    index, expectation_domain_column_name
-                ]
-                for column_name in unexpected_index_column_names:
-                    column_name = get_dbms_compatible_column_names(
-                        column_names=column_name,
-                        batch_columns_list=metrics["table.columns"],
-                        execution_engine=execution_engine,
-                        error_message_template='Error: The unexpected_index_column "{column_name:s}" does not exist in Dataframe. Please check your configuration and try again.',
-                    )
-                    # add the actual unexpected value
-                    primary_key_dict[column_name] = df.at[index, column_name]
-                unexpected_index_list.append(primary_key_dict)
+                primary_key_dict[column_name] = df.at[index, column_name]
+            unexpected_index_list.append(primary_key_dict)
+    # or just the default indices
     else:
-        if df.index.name:
-            unexpected_index_column_names: List[str] = [df.index.name]
-            unexpected_index_list = (
-                get_unexpected_indices_for_single_pandas_named_index(
-                    df,
-                    unexpected_index_list,
-                    expectation_domain_column_name,
-                    unexpected_index_column_names,
-                )
-            )
-        elif hasattr(df.index, "names"):
-            unexpected_index_column_names: List[str] = list(df.index.names)
-            unexpected_index_list = (
-                get_unexpected_indices_for_multiple_pandas_named_indices(
-                    df,
-                    unexpected_index_list,
-                    expectation_domain_column_name,
-                    unexpected_index_column_names,
-                )
-            )
+        unexpected_index_list = list(df.index)
 
-        else:
-            unexpected_index_list = list(df.index)
-
-    # do the formatting
     if result_format["result_format"] == "COMPLETE":
-        # how does this affect named indices when it comes to printing them out?
-        # i could technically build it out here
-        # TODO: see if i can take care of the rendering here
         return unexpected_index_list
     return unexpected_index_list[: result_format["partial_unexpected_count"]]
 
--- a/great_expectations/expectations/metrics/map_metric_provider.py
+++ b/great_expectations/expectations/metrics/map_metric_provider.py
@@ -4,6 +4,7 @@ from functools import wraps
 from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
 
 import numpy as np
+import pandas as pd
 
 import great_expectations.exceptions as ge_exceptions
 from great_expectations.core import ExpectationConfiguration
@@ -34,10 +35,9 @@ from great_expectations.expectations.metrics.util import (
     Insert,
     Label,
     Select,
+    compute_unexpected_pandas_indices,
     get_dbms_compatible_column_names,
     get_sqlalchemy_source_table_and_schema,
-    get_unexpected_indices_for_multiple_pandas_named_indices,
-    get_unexpected_indices_for_single_pandas_named_index,
     sql_statement_with_post_compile_to_string,
     verify_column_names_exist,
 )
@@ -1747,7 +1747,9 @@ def _pandas_map_condition_index(
     with all of the available "domain_kwargs" keys.
     """
     domain_kwargs = dict(**compute_domain_kwargs, **accessor_domain_kwargs)
-    df = execution_engine.get_domain_records(domain_kwargs=domain_kwargs)
+    domain_records_df: pd.DataFrame = execution_engine.get_domain_records(
+        domain_kwargs=domain_kwargs
+    )
 
     if "column" in accessor_domain_kwargs:
         column_name: Union[str, quoted_name] = accessor_domain_kwargs["column"]
@@ -1767,7 +1769,9 @@ def _pandas_map_condition_index(
             "filter_column_isnull", getattr(cls, "filter_column_isnull", False)
         )
         if filter_column_isnull:
-            df = df[df[column_name].notnull()]
+            domain_records_df = domain_records_df[
+                domain_records_df[column_name].notnull()
+            ]
 
     elif "column_list" in accessor_domain_kwargs:
         column_list: List[Union[str, quoted_name]] = accessor_domain_kwargs[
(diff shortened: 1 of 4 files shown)


tests/checkpoint/test_checkpoint_result_format.py
```diff
@@ -785,3 +785,139 @@ def test_pandas_result_format_in_checkpoint_pk_defined_one_expectation_summary_o
...
+    dict_to_update_checkpoint: dict = {
+        "result_format": {
+            "result_format": "COMPLETE",
+            "unexpected_index_column_names": ["pk_1"],
+        }
+    }
+    # build our own BatchRequest
+    dataframe: pd.DataFrame = pandas_animals_dataframe_for_unexpected_rows_and_index
+    # setting named index
+    dataframe.set_index("pk_1")
```
Shinnnyshinshin: Setting named index for in-memory pandas `DataFrame`

great_expectations/expectations/metrics/map_metric_provider.py
```diff
@@ -1781,17 +1783,41 @@ def _pandas_map_condition_index(
...
-        unexpected_index_column_names: List[str] = result_format[
-            "unexpected_index_column_names"
-        ]
-        # column the expectation is being run on
-        expectation_domain_column_name: str = domain_kwargs["column"]
+    expectation_domain_column_name: str = domain_kwargs.get("column")
 
-        unexpected_indices: List[int] = list(df.index)
+    # one named index
+    if df.index.name is not None:
```
billdirks: This function is rather long. Could you factor out this if/else block into it's own function? It looks like the purpose of this bock is to compute the `unexpected_index_list`,  so something like:
```
unexpected_index_list = computed_unexpected_panda_indexes(df, result_format, ...)
```

great_expectations/expectations/metrics/util.py
```diff
@@ -1058,3 +1059,83 @@ def get_sqlalchemy_source_table_and_schema(
...
+    unexpected_index_values_by_named_index: List[Union[int, str]] = list(df.index)
+    unexpected_index_list: List[Dict[str, Any]] = list()
+
+    for index in unexpected_index_values_by_named_index:
+        primary_key_dict: Dict[str, Any] = dict()
+        # domain column first
+        primary_key_dict[expectation_domain_column_name] = df.at[
+            index, expectation_domain_column_name
+        ]
+        for column_name in unexpected_index_column_names:
```
billdirks: Is `unexpected_index_column_names` always of length 1 in this function? If so, I'd assert that at the beginning of this function. Then you can remove this inner `for` loop.

great_expectations/expectations/metrics/util.py
```diff
@@ -1058,3 +1059,83 @@ def get_sqlalchemy_source_table_and_schema(
...
+        df: reference to Pandas dataframe
+        expectation_domain_column_name: column that Expectation is being run for
+        unexpected_index_column_names: column_names for indices, either named index or unexpected_index_columns
+
+    Returns:
+        List of Dicts that contain ID/PK values
+    """
+    if not expectation_domain_column_name:
+        raise ge_exceptions.MetricResolutionError(
+            message=f"Error: The domain column is currently set to None. Please check your configuration.",
```
billdirks, line 1083: We could get this error if `expectation_domain_column_name` is the empty string `""` so we might want to determine if it's `None` or `""` and update the error message. If this is only called by our code (vs a user facing api) and we are mypy checking this should never be `None`. However, I know we do not have everything covered by mypy.

great_expectations/expectations/metrics/util.py
```diff
@@ -1058,3 +1059,83 @@ def get_sqlalchemy_source_table_and_schema(
...
+        primary_key_dict[expectation_domain_column_name] = df.at[
+            index, expectation_domain_column_name
+        ]
+        for column_name in unexpected_index_column_names:
+            if column_name not in index_names:
+                raise ge_exceptions.MetricResolutionError(
+                    message=f"Error: The column {column_name} does not exist in the named indices. Please check your configuration",
+                    failed_metrics=["unexpected_index_list"],
+                )
+            tuple_index = index_names.index(column_name, 0)
```
billdirks: This logic is identical for each loop through `unexpected_indices` so I'd do this validation once at the beginning of this function outside the `unexpected_indices` loop since that loop may be large.

That is, at the top level do:
```
column_name_to_tuple_index = {}  # add type here
for tuple_index, column_name in enumerate(unexpected_index_column_names):
    if column_name not in index_names:
        raise ge_exceptions.MetricResolutionError(
            message=f"Error: The column {column_name} does not exist in the named indices. Please check your configuration",
             failed_metrics=["unexpected_index_list"],
        )
    column_name_to_tuple_index[column_name] = tuple_index
```
This does the validation early and  removes the call to `index` which is a hidden inner loop.

Then the loop here becomes:
```
for column_name in unexpected_index_column_names:
    primary_key_dict[column_name] = column_name_to_tuple_index[column_name]
```

great_expectations/expectations/metrics/util.py
```diff
@@ -1058,3 +1059,83 @@ def get_sqlalchemy_source_table_and_schema(
         )
     else:
         return engine.batch_manager.active_batch_data.selectable
+
+
+def get_unexpected_indices_for_multiple_pandas_named_indices(
+    df: pd.DataFrame,
```
billdirks: This is a dataframe of only unexpected results? It looks like it since you do `unexpected_indices: List[Union[int, str]] = list(df.index)`. I would name this something that conveys this instead of `df`. Maybe `unexpected_df`?

great_expectations/expectations/metrics/util.py
```diff
@@ -1058,3 +1059,83 @@ def get_sqlalchemy_source_table_and_schema(
...
+                    failed_metrics=["unexpected_index_list"],
+                )
+            tuple_index = index_names.index(column_name, 0)
+            primary_key_dict[column_name] = index[tuple_index]
+        unexpected_index_list.append(primary_key_dict)
+    return unexpected_index_list
+
+
+def get_unexpected_indices_for_single_pandas_named_index(
+    df: pd.DataFrame,
```
billdirks: Similar comment about `df` name here.

billdirks (commented):
Thanks for making this change! I haven't looked at tests yet but have added some comments inline.

tests/expectations/metrics/test_metrics_util.py
```diff
@@ -127,3 +155,133 @@ def test_sql_statement_conversion_to_string_bigquery(test_backends):
...
+    )
+
+
+@pytest.mark.unit
+def test_get_unexpected_indices_for_multiple_pandas_named_indices_named_unexpected_index_wrong_domain(
+    pandas_animals_dataframe_for_unexpected_rows_and_index,
+):
+    dataframe: pd.DataFrame = pandas_animals_dataframe_for_unexpected_rows_and_index
+    updated_dataframe: pd.DataFrame = dataframe.set_index(["pk_1", "pk_2"])
+    expectation_domain_column_name = None
```
billdirks, line 276: You could parameterize the `expectation_domain_column_name` so it is `None` and `""`.

tests/expectations/metrics/test_metrics_util.py
```diff
@@ -127,3 +155,133 @@ def test_sql_statement_conversion_to_string_bigquery(test_backends):
...
+            unexpected_index_column_names=unexpected_index_column_names,
+        )
+    assert e.value.message == (
+        "Error: The domain column is currently set to None. Please check your "
+        "configuration."
+    )
+
+
+@pytest.mark.unit
+def test_get_unexpected_indices_for_multiple_pandas_named_indices(
```
billdirks: nit: Could you move this tests so it is next to `test_get_unexpected_indices_for_multiple_pandas_named_indices_named_unexpected_index_columns` since it's basically the same except for how the index column names are defined.

tests/expectations/metrics/test_metrics_util.py
```diff
@@ -127,3 +155,133 @@ def test_sql_statement_conversion_to_string_bigquery(test_backends):
         )
     else:
         pytest.skip(f"skipping sql statement conversion test for : bigquery")
+
+
+@pytest.mark.unit
```
billdirks, line 160: Thanks for adding unit tests!

tests/checkpoint/test_checkpoint_result_format.py
```diff
@@ -785,3 +785,509 @@ def test_pandas_result_format_in_checkpoint_pk_defined_one_expectation_summary_o
...
+        "unexpected_index_list"
+    )
+    assert first_result_full_list == [
+        {"animals": "giraffe", "pk_1": 3, "pk_2": "three"},
+        {"animals": "lion", "pk_1": 4, "pk_2": "four"},
+        {"animals": "zebra", "pk_1": 5, "pk_2": "five"},
+    ]
+    first_result_partial_list: List[Dict[str, Any]] = evrs[0]["results"][0][
+        "result"
+    ].get("partial_unexpected_index_list")
```
billdirks, line 1173: What's the difference between full and partial here? Does this dataset make them equivalent for this run?

great_expectations/expectations/metrics/map_metric_provider.py
Shinnnyshinshin: Great idea. I'll add the additional layer of abstraction

great_expectations/expectations/metrics/util.py
Shinnnyshinshin: good point

great_expectations/expectations/metrics/util.py
Shinnnyshinshin: `if expectation_domain_column_name is None` and make `Optional` above or `Union[str, None]`

great_expectations/expectations/metrics/util.py
Shinnnyshinshin: omg yes, thanks for the suggestion

great_expectations/expectations/metrics/util.py
Shinnnyshinshin: Yes, and we would want to assert this. Great idea. thank you

tests/checkpoint/test_checkpoint_result_format.py
Shinnnyshinshin: Yes `partial` is set to 20 by default. As discussed, I'll add a test where partial is set to some small number so that the asserts are different.

tests/expectations/metrics/test_metrics_util.py
Shinnnyshinshin: Moving to other location. Thank you for the suggestion :)

great_expectations/expectations/metrics/util.py
Shinnnyshinshin: Added assertion

great_expectations/expectations/metrics/util.py
Shinnnyshinshin: This is now done as well

great_expectations/expectations/metrics/util.py
Shinnnyshinshin: This didn't go as planned. The issue is that the necessary information is in two places:

`domain_records_df_index_names` contains the names of the named indices `[`pk1`, `pk_2`]` etc. 
`unexpected_indices` is the Tuple that contains the index values, but doesn't have the "actual" names of the columns.  So the `enumerate()` wasn't able to help us in this case

However, I moved the check that raises a `MetricResolutionError` outside of the big loop (`unexpected_index_column_names` is typically going to be pretty small), so we can error out earlier.

great_expectations/expectations/metrics/util.py
Shinnnyshinshin: Done :)

great_expectations/expectations/metrics/util.py
Shinnnyshinshin: Done :)

tests/expectations/metrics/test_metrics_util.py
Shinnnyshinshin: Done :)

tests/checkpoint/test_checkpoint_result_format.py
Shinnnyshinshin: Added a test where partial_list is set to 1

great_expectations/expectations/metrics/util.py
```diff
@@ -1058,3 +1060,165 @@ def get_sqlalchemy_source_table_and_schema(
...
+    unexpected_index_list: List[Dict[str, Any]] = list()
+
+    for index in unexpected_indices:
+        primary_key_dict: Dict[str, Any] = dict()
+        # domain column first
+        primary_key_dict[expectation_domain_column_name] = domain_records_df.at[
+            index, expectation_domain_column_name
+        ]
+        for column_name in unexpected_index_column_names:
+            tuple_index = domain_records_df_index_names.index(column_name, 0)
```
billdirks: You may want to define `tuple_index` outside this `for` loop and incorporate it into the `for column_name in unexpected_index_column_names` loop above since the `index` call will loop through the list `domain_records_df_index_names` looking for a match time it is invoked. However, depending on the size of the `domain_records_df_index_names` and the speed of hashing I'm not sure what will be faster so maybe it doesn't matter.

great_expectations/expectations/metrics/util.py
```diff
@@ -1058,3 +1060,165 @@ def get_sqlalchemy_source_table_and_schema(
...
+        domain_records_df.index
+    )
+    unexpected_index_list: List[Dict[str, Any]] = list()
+
+    if not (
+        len(unexpected_index_column_names) == 1
+        and unexpected_index_column_names[0] == domain_records_df.index.name
+    ):
+        raise ge_exceptions.MetricResolutionError(
+            message=f"Error: The column {unexpected_index_column_names[0]} does not exist in the named indices. Please check your configuration",
```
billdirks: Calling `unexpected_index_column_names[0]` in the error message might die if `unexpected_index_column_names[0]` is empty. Maybe have another `if` block above this one that checks `if not unexpected_index_column_names: raise ge_exceptions.MetricResolutionError("unexpected_index_column_names must not be empty")`

billdirks (commented):
I'm not quite through the PR but I'm off to a meeting so wanted to send out the current state of the review.

great_expectations/expectations/metrics/util.py
billdirks: Another simpler solution would be to change `{unexpected_index_column_names[0]}` to `{unexpected_index_column_names[0] if unexpected_index_column_names else "<no column specified>"}` or something similar.

great_expectations/expectations/metrics/util.py
Shinnnyshinshin: I had to think a little bit more about this, but I think it's a great suggestion, even with the uncertainty around the speed of hashing. I'd adjusted the loop above to build a dictionary of the column names and tuple indices. That away this line becomes a much simple dictionary value retrieval operation 🙇🏼 thanks for this

billdirks (approved):
Please do one of the changes suggested here:
https://github.com/great-expectations/great_expectations/pull/6656#discussion_r1059487151

Otherwise LGTM!

PR #6657, opened 2022-12-28
[FEATURE] Add sqlite datasource

billdirks:
There are 2 substantive changes in this PR"

1. We make the `Datasource` class generic over the data asset. This lets us use properties of the particular data asset in the datasource without casting/ignoring errors. 
2. Adds the new sqlite datasource.

### DIFF
--- a/great_expectations/experimental/datasources/interfaces.py
+++ b/great_expectations/experimental/datasources/interfaces.py
@@ -3,7 +3,18 @@ from __future__ import annotations
 import dataclasses
 import logging
 from pprint import pformat as pf
-from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional, Set, Type
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Generic,
+    List,
+    Mapping,
+    Optional,
+    Set,
+    Type,
+    TypeVar,
+)
 
 import pydantic
 from typing_extensions import ClassVar, TypeAlias
@@ -63,7 +74,12 @@ class DataAsset(ExperimentalBaseModel):
         raise NotImplementedError
 
 
-class Datasource(ExperimentalBaseModel, metaclass=MetaDatasource):
+DataAssetType = TypeVar("DataAssetType", bound=DataAsset)
+
+
+class Datasource(
+    ExperimentalBaseModel, Generic[DataAssetType], metaclass=MetaDatasource
+):
     # class attrs
     asset_types: ClassVar[List[Type[DataAsset]]] = []
     # Datasource instance attrs but these will be fed into the `execution_engine` constructor
@@ -80,7 +96,7 @@ class Datasource(ExperimentalBaseModel, metaclass=MetaDatasource):
     # instance attrs
     type: str
     name: str
-    assets: Mapping[str, DataAsset] = {}
+    assets: Mapping[str, DataAssetType] = {}
     _execution_engine: ExecutionEngine = pydantic.PrivateAttr()
 
     def __init__(self, **kwargs):
@@ -103,12 +119,14 @@ class Datasource(ExperimentalBaseModel, metaclass=MetaDatasource):
     @classmethod
     def _load_asset_subtype(cls, v: Dict[str, dict]):
         LOGGER.info(f"Loading 'assets' ->\n{pf(v, depth=3)}")
-        loaded_assets: Dict[str, DataAsset] = {}
+        loaded_assets: Dict[str, DataAssetType] = {}
 
         # TODO (kilo59): catch key errors
         for asset_name, config in v.items():
             asset_type_name: str = config["type"]
-            asset_type: Type[DataAsset] = _SourceFactories.type_lookup[asset_type_name]
+            asset_type: Type[DataAssetType] = _SourceFactories.type_lookup[
+                asset_type_name
+            ]
             LOGGER.debug(f"Instantiating '{asset_type_name}' as {asset_type}")
             loaded_assets[asset_name] = asset_type(**config)
 
@@ -140,7 +158,7 @@ class Datasource(ExperimentalBaseModel, metaclass=MetaDatasource):
             f"{self.__class__.__name__} must implement `.get_batch_list_from_batch_request()`"
         )
 
-    def get_asset(self, asset_name: str) -> DataAsset:
+    def get_asset(self, asset_name: str) -> DataAssetType:
         """Returns the DataAsset referred to by name"""
         # This default implementation will be used if protocol is inherited
         try:
--- a/great_expectations/experimental/datasources/sql_datasource.py
+++ b/great_expectations/experimental/datasources/sql_datasource.py
@@ -14,7 +14,6 @@ from typing import (
     Optional,
     Sequence,
     Type,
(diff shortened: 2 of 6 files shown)


great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -86,50 +88,47 @@ class SqlYearMonthSplitter(ColumnSplitter):
...
-        from great_expectations.execution_engine import SqlAlchemyExecutionEngine
+        return _query_for_year_and_month(
+            table_asset, self.column_name, _get_sql_datetime_range
+        )
+
 
-        assert isinstance(data_asset, TableAsset), "data_asset must be a TableAsset"
-        assert isinstance(
-            data_asset.datasource.execution_engine, SqlAlchemyExecutionEngine
+def _query_for_year_and_month(
```
billdirks, line 101: The strategy here is to take the code that use to be in `param_defaults` and put it here. We parameterize the `query_datetime_range` `Callable` which let's us break out the sqlite specific code.

great_expectations/experimental/datasources/sqlite_datasource.py
```diff
@@ -0,0 +1,121 @@
...
+)
+
+
+class SqliteTableAsset(TableAsset):
+    # Subclass overrides
+    type: Literal["sqlite_table"] = "sqlite_table"  # type: ignore[assignment]
+    column_splitter: Optional[SqliteYearMonthSplitter] = None  # type: ignore[assignment]
+
+    # This asset type will support a variety of splitters
+    def add_year_and_month_splitter(
```
billdirks, line 43: I made the new `SqliteTableAsset` so I could have this helper function add correct splitter.

great_expectations/experimental/datasources/sqlite_datasource.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@pydantic_dc.dataclass(frozen=True)
+class SqliteYearMonthSplitter(ColumnSplitter):
+    method_name: Literal["split_on_year_and_month"] = "split_on_year_and_month"
+    param_names: List[Literal["year", "month"]] = pydantic.Field(
+        default_factory=lambda: ["year", "month"]
+    )
+
+    def param_defaults(self, table_asset: TableAsset) -> Dict[str, List]:
```
billdirks, line 68: I tried making the `table_asset` input parameter more specific (ie `SqliteTableAsset`) using `Generics` but ran into some problems using pydantic and `Generics` together. I decided not to resolve the issue right now.

great_expectations/experimental/datasources/sqlite_datasource.py
```diff
@@ -0,0 +1,121 @@
...
+        Returns:
+            The SqliteTableAsset that is added to the datasource.
+        """
+        asset = SqliteTableAsset(
+            name=name,
+            table_name=table_name,
+            order_by=order_by or [],  # type: ignore[arg-type]  # coerce list[str]
+            # see TableAsset._parse_order_by_sorter()
+        )
+        # TODO (kilo59): custom init for `DataAsset` to accept datasource in constructor?
```
billdirks: This was basically lifted from the sql datasource. I'm not sure if I should keep or delete this comment. 

We may want to use the setter for the `datasource` instead of accessing `_datasource` directly. I didn't do that now because I then wanted to typecheck on assignment which led to defining more methods on each class. I can do this in a followup PR.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -86,50 +88,47 @@ class SqlYearMonthSplitter(ColumnSplitter):
...
+    query_datetime_range: Callable[
+        [sqlalchemy.engine.base.Connection, str, str], DatetimeRange
+    ],
+) -> Dict[str, List]:
+    # We should make an assertion about the execution_engine earlier. Right now it is assigned to
+    # after construction. We may be able to use a hook in a property setter.
+    from great_expectations.execution_engine import SqlAlchemyExecutionEngine
+
+    assert isinstance(
+        table_asset.datasource.execution_engine, SqlAlchemyExecutionEngine
```
Kilo59, line 113: Can this be a custom `pydantic` validator on the `SQLDatasource`?

great_expectations/experimental/datasources/sqlite_datasource.py
```diff
@@ -0,0 +1,121 @@
...
+
+class SqliteDatasource(SQLDatasource):
+    # class var definitions
+    asset_types: ClassVar[List[Type[DataAsset]]] = [SqliteTableAsset]
+
+    # Subclass instance var overrides
+    # right side of the operator determines the type name
+    # left side enforces the names on instance creation
+    type: Literal["sqlite"] = "sqlite"  # type: ignore[assignment]
+    connection_string: str
```
Kilo59: I think we should consider adding some constraints to this `connection_string` field.

It could as simple a custom validator that checks that the `connection_string` `.startswith('sqlite')`

Or is could be with a `pydantic` constrained string type (with a regex).
A benefit of using the constrained type is the regex will show up in the JSON schema for the model.
https://docs.pydantic.dev/usage/types/#constrained-types
https://docs.pydantic.dev/usage/types/#arguments-to-constr

https://docs.sqlalchemy.org/en/14/dialects/sqlite.html#

great_expectations/experimental/datasources/sqlite_datasource.py
Kilo59: If the comments are still on the parent method then I think we could delete them here.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -63,7 +74,12 @@ def get_batch_request(self, options: Optional[BatchRequestOptions]) -> BatchRequ
         raise NotImplementedError
 
 
-class Datasource(ExperimentalBaseModel, metaclass=MetaDatasource):
+DataAssetType = TypeVar("DataAssetType", bound=DataAsset)
+
+
+class Datasource(
+    ExperimentalBaseModel, Generic[DataAssetType], metaclass=MetaDatasource
+):
```
Kilo59, line 82: So now the type-system knows that a `PostgresDatasource` contains `TableAsset`s and not a general/non-specific `DataAsset` type?
🙌

Kilo59 (approved):

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Because this happens in assignment I think the way to use pydantic validators would be to turn on `validate_assignment` on a model config which would turn it on for all fields. I'm not sure we want to do that. I do think it would be better to do this check on the property setter and when we use it, like here. Right now we set it using the private variable `_datasource` so some cleanup is needed. I'll do this in a followup PR.

great_expectations/experimental/datasources/sqlite_datasource.py
billdirks: Added `constr` and some tests.

great_expectations/experimental/datasources/sqlite_datasource.py
billdirks: Done.

PR #6661, opened 2022-12-29
[MAINTENANCE] Make only column y data_alt different in col_pair_equal tests

kenwade4:
Both Trino and Snowflake don't allow strings in numeric columns (the "abc" in final row of column y) and both raise errors when trying to insert original data to a table.

Note that column y is only used in pandas tests anyway.


Shinnnyshinshin (approved):
LGTM. Thanks for finding this fix @kenwade4

Issue #6226, opened 2022-10-27
kwargs are not passed to SqlAlchemy Engine


hoavho:
## Describe the bug
I'd like to run expectations against data in Apache Druid. I tested and it works using SqlAlchemyExecutionEngine. However, when I connect to Druid with https/SSL (using same datasource/SqlAlchemyExecutionEngine config), I get `SSL: CERTIFICATE_VERIFY_FAILED` because the Druid server is configured with self signed cert. SqlAlchemy allows us to configure `connect_args` to skip verification step, or specify location of the cert. I tried to configure GE datasource to pass additional kwargs `connect_args` to SqlAlchemy engine - but it does NOT look like that the kwargs are passed to sqlalchemy engine properly. (I expect I should be able to configure that - per this https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/execution_engine/sqlalchemy_execution_engine.py#L249)

## To Reproduce
Prerequisite: Druid local setup with TLS/SSL enabled

**Steps to reproduce the behavior**:
1. Create new datasource using great_expectation CLI (follow CLI guided instructions to setup SQL datasource with connection string)
2. Update Jupyter notebook as below
```
connection_string = "druid+https://my-druid-host.local:443/druid/v2/sql"

# If schema_name is not relevant to your SQL backend (i.e. SQLite),
# please remove from the following line and the configuration below
schema_name = "druid"

example_yaml = f"""
name: {datasource_name}
class_name: Datasource
execution_engine:
    class_name: SqlAlchemyExecutionEngine
    connection_string: {connection_string}
    connect_args: 
        ssl_verify_cert: False
data_connectors:
    default_runtime_data_connector_name:
        class_name: RuntimeDataConnector
        batch_identifiers:
            - default_identifier_name
        batch_spec_passthrough: 
            create_temp_table: False
    default_inferred_data_connector_name:
        class_name: InferredAssetSqlDataConnector
        include_schema_name: True
        introspection_directives:
            schema_name: {schema_name}
        batch_spec_passthrough: 
            create_temp_table: False
"""
print(example_yaml)
```

3. Run Jupyter notebook

Error

```
[...]

PR #6663, opened 2022-12-29
[BUGFIX] Add connect args to execution engine schema

itaise:
This PR is to solve [this issue](https://github.com/great-expectations/great_expectations/issues/6226).

There is an issue that when creating SQLAlchemy execution engine, if the user send "connect_args" in the execution engine config, it is not passed to the "create_engine" method. 
This is because this attribute is dropped when creating the ExecutionEngineConfigSchema.

I checked it locally and this PR solve the issue

- Add connect_args to the ExecutionEngineConfigSchema

### DIFF
--- a/great_expectations/data_context/types/base.py
+++ b/great_expectations/data_context/types/base.py
@@ -1004,6 +1004,9 @@ class ExecutionEngineConfigSchema(Schema):
     boto3_options = fields.Dict(
         keys=fields.Str(), values=fields.Str(), required=False, allow_none=True
     )
+    connect_args = fields.Dict(
+        keys=fields.Str(), values=fields.Dict(), required=False, allow_none=True
+    )
     azure_options = fields.Dict(
         keys=fields.Str(), values=fields.Str(), required=False, allow_none=True
     )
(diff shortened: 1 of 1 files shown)


amitbend (approved):

great_expectations/data_context/types/base.py
```diff
@@ -1004,6 +1004,9 @@ class Meta:
     boto3_options = fields.Dict(
         keys=fields.Str(), values=fields.Str(), required=False, allow_none=True
     )
+    connect_args = fields.Dict(
```
Shinnnyshinshin, line 1010: Would we be able to add `connect_args` to `ExecutionEngineConfig` class above as well?

Shinnnyshinshin: Otherwise the PR looks great. thanks for catching this @itaise 🙇🏼

itaise: @Shinnnyshinshin  thank you! done :)

Shinnnyshinshin (approved):
🚀

PR #6666, opened 2022-12-29
[DOCS] doc-356 importable prerequisite box with correct theme applied and default entries

Rachel-Reverie:
- Adds an importable Prerequisite admonition box to help standardize that element in our docs.
- the box applies correct theme styling, imported from Docusaurus (currently set to a 'caution' admonition box)
- the box provides default entries for common requirements which can be toggled by passing properties to the element in markdown docs.
- the list of requirements can be extended by providing a markdown list of additional requirements as the content of the box in markdown docs.

Please note: this PR just makes the component available, it doesn't actually apply it to any documents yet.  That will be done in future PRs.

### Definition of Done

### DIFF
--- a/docs/components/_prerequisites.jsx
+++ b/docs/components/_prerequisites.jsx
@@ -1,6 +1,5 @@
 import React from 'react'
 import Admonition from '@theme/Admonition'
-import data from '../term_tags/terms.json'
 
 /**
  * A flexible Prerequisites admonition block.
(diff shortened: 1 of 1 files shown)


rdodev (approved):
Wish we had a better way to inject URLs so they aren't hardcoded. LGTM otherwise.

Rachel-Reverie:
> Wish we had a better way to inject URLs so they aren't hardcoded. LGTM otherwise.

Likewise.  I think I can adjust it to pull them from a .json file, instead -- I'm planning on using one to build out some other "importable" statements so that things like our currently supported Python versions only need to be updated in one spot.  If that works the way I'm planning, I'll do a future update where the predefined "default" requirement entries are also pulled from a `components.json` file, instead.

PR #6670, opened 2022-12-29
[MAINTENANCE] Make only column y data_alt different in col_pair_in_set tests

kenwade4:
Both Trino and Snowflake don't allow strings in numeric columns (the "abc" in final row of column y) and both raise errors when trying to insert original data to a table.


Shinnnyshinshin (approved):
Thank you for catching this @kenwade4

PR #6674, opened 2022-12-30
[FEATURE] F/great 1400/postgres constr

billdirks:
Update postgres datasource to validate the connection string starts with `postgresql`.

### DIFF
--- a/great_expectations/experimental/datasources/postgres_datasource.py
+++ b/great_expectations/experimental/datasources/postgres_datasource.py
@@ -1,16 +1,9 @@
-from typing import TYPE_CHECKING
-
-from pydantic import constr
+from pydantic import PostgresDsn
 from typing_extensions import Literal
 
 from great_expectations.experimental.datasources.sql_datasource import SQLDatasource
 
-if TYPE_CHECKING:
-    PostgresConnectionString = str
-else:
-    PostgresConnectionString = constr(regex=r"^postgresql")
-
 
 class PostgresDatasource(SQLDatasource):
     type: Literal["postgres"] = "postgres"  # type: ignore[assignment]
-    connection_string: PostgresConnectionString
+    connection_string: PostgresDsn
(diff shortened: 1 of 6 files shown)


great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -1,7 +1,16 @@
...
 
+if TYPE_CHECKING:
+    PostgresConnectionString = str
+else:
+    PostgresConnectionString = constr(regex=r"^postgresql")
+
 
 class PostgresDatasource(SQLDatasource):
     type: Literal["postgres"] = "postgres"  # type: ignore[assignment]
+    connection_string: PostgresConnectionString
```
Kilo59: @billdirks did you try using the `PostgresDsn` type for this?
```python
from pydantic import PostgresDsn
```

```suggestion
    connection_string: PostgresDsn
```
https://docs.pydantic.dev/usage/types/#urls

Kilo59 (approved):
I think we should try to use the dedicated `PostgresDsn` type... although it's possible this could fail when we pass it to the underlying `ExecutionEngine` (if it's expecting a string).

If that's the case I'm fine with the `constr` and we could revisit this later.
I do think the errors about specific parts of the URI that are "wrong" make it valuable.

great_expectations/experimental/datasources/postgres_datasource.py
billdirks: I hadn't tried this. Thanks for pointing it out! Done and I added tests.

tests/experimental/datasources/test_sqlite_datasource.py
```diff
@@ -26,6 +27,7 @@ def test_connection_string_starts_with_sqlite():
     assert datasource.connection_string == connection_string
 
 
+@pytest.mark.unit
```
billdirks, line 30: I forgot to add these markers to the equivalent sqlite tests so added them now.

PR #6660, opened 2022-12-29
[FEATURE] ID/PK flag `return_unexpected_index_query` that allows `QUERY` output to be suppressed

Shinnnyshinshin:
- Adds `return_unexpected_index_query` flag to `result_format`. 
- By default, `unexpected_index_query` is returned by `ExecutionEngines` that support it (currently on `SqlAlchemyExecutionEngine`), only when `result_format` is set to `COMPLETE`. 
- setting `return_unexpected_index_query` to `False` will suppress the `unexpected_index_query` even when `result_format` is set to `COMPLETE`
-  Added tests at `tests/checkpoint/test_checkpoint_result_format.py`-level

### Can you give some examples? 
```python 
        "result_format": {
            "result_format": "COMPLETE",
            "unexpected_index_column_names": ["pk_1"],
            "return_unexpected_index_query": True,
        }
```
* **Will** have `unexpected_index_query`, since `return_unexpected_index_query` is explicitly set to `True`

```python 
        "result_format": {
            "result_format": "COMPLETE",
            "unexpected_index_column_names": ["pk_1"],
            "return_unexpected_index_query": False,
        }
```
* **Will not** have `unexpected_index_query`, since `return_unexpected_index_query` is explicitly set to `False`

```python 
        "result_format": {
            "result_format": "COMPLETE",
            "unexpected_index_column_names": ["pk_1"],
        }
```
* **Will** have `unexpected_index_query`, since it is returned by default if result format is set to `COMPLETE`

```python 
        "result_format": {
            "result_format": "SUMMARY",
            "unexpected_index_column_names": ["pk_1"],
        }
```
* **Will not** have `unexpected_index_query`, since result format is set to `SUMMARY`

### Definition of Done
- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added [unit tests](https://docs.greatexpectations.io/docs/contributing/contributing_test#writing-unit-and-integration-tests) where applicable and made sure that new and existing tests are passing.
- [x] I have run any local integration tests and made sure that nothing is broken.

Thank you for submitting!

PR #6562, opened 2022-12-12
[FEATURE] Ensure `result_format` accessed is through Checkpoint, and warns users if `Expectation` or `Validator`-level

Shinnnyshinshin:
# What has changed? 
* We prefer `Checkpoints` to configure `result_format` including `unexpected_index_column_names`.
* Configuring `unexpected_index_column_names` At the `Validator` or `ExpectationConfiguration` level will give a warning. 
* The check is facilitated by `_warn_if_result_format_config_in_expectation_configuration()`, and `_warn_if_result_format_config_in_runtime_configuration()` which are helper methods that have been added to Expectation's `validate()` method.

### Tests have been updated and refactored to reflect this change. 
* Tests that were previously in `test_map_metric.py` have been migrated to `test_checkpoint_result_format.py` and now use a `Checkpoint` configuration. 
* Test has been added to `test_map_metric.py` to check that appropriate warning is given when configuring `result_format` at the `ExpectationConfiguration`-level
* Tests have been added to `test_validator.py` to check that appropriate warning is given when configuring `result_format` at the `Validator`-level
* Metrics that are needed for ID/PK, namely `unexpected_index_list` and `unexpected_index_query` are now tested as unittests in `tests/expectations/metrics/column_map_metrics/test_unexpected_indices_and_query.py` 

### Appendix
* Some example configurations to demonstrate what will and not work as the result of this PR. 
#### The following configurations at the `Checkpoint`-level will work
1. in `CheckpointConfiguration`
   
```python
# if Checkpoint configuration contains the following `runtime_configuration`
checkpoint_dict["runtime_configuration"] = {
        "result_format": {
            "result_format": "COMPLETE",
            "unexpected_index_column_names": ["pk_1"],
        }
    }
```

2. In `context.run_checkpoint()`

```python 
result_format: dict = {
    "result_format": "COMPLETE",
    "unexpected_index_column_names": ["pk_1"],
}
result: CheckpointResult = context.run_checkpoint(
    checkpoint_name="my_checkpoint", result_format=result_format
)
```

#### The following configurations at the `Validator`-level and `Expectation`-level will issue warnings.
1. at `validator.expect()`
```python 
# will raise UserWarning
result_format_config: dict = {
        "result_format": "COMPLETE",
        "unexpected_index_column_names": ["pk_1"],
}

result: ExpectationValidationResult = validator.expect_column_values_to_be_in_set(
    column="animals",
    value_set=["cat", "fish", "dog"],
    result_format=result_format_config,
)
```

[...]

PR #6676, opened 2022-12-30
[FEATURE] Support to include ID/PK in validation result for each row - Spark

### LINKED ISSUE #3195: Support to include ID/PK in validation result for each row that failed an expectations

Shinnnyshinshin:

PySpark implementation of `unexpected_index_column`s value that allows users to specify a primary key (PK) column for identifying rows that failed an Expectation (usually returned as part of the `unexpected_index_list`).  The PR also enables `unexpected_index_query` to be returned to the user, which provides a way for the user to retrieve the full list of unexpected results from the PySpark DataFrame.

Changes were made to the `map_metric_provider` to take in the parameter from `result_format` and output `unexpected_index_list` as key-value pairs of the primary key column.

Note : This is only the Spark implementation, [Pandas has been merged already](https://github.com/great-expectations/great_expectations/pull/6329), and [Spark has been merged as well](https://github.com/great-expectations/great_expectations/pull/6448).

* closes #3195

### What has changed?
* `unexpected_index_list` and `unexpected_index_query` were added as validation dependencies for `SparkDExecutionEngine`
* `_spark_map_condition_index` was added as MapMetric.
   - This Metric returns the indices that contain unexpected values, according to the Primary Key values specified using `unexpected_index_column_names`.
* `_spark_map_condition_index` was added as MapMetric.
   - This Metric returns the indices that contain unexpected values, according to the Primary Key values specified using `unexpected_index_column_names`.
   - Like PR #6660 the MapMetric also supports the `return_unexpected_index_query` flag to suppress output if desired.

### Can you give me an Example? (heavily adopted from the [Pandas](https://github.com/great-expectations/great_expectations/pull/6329) and [SQL](https://github.com/great-expectations/great_expectations/pull/6448) Examples)

Given the following PySpark DataFrame
```python 
 df: pd.DataFrame = pd.DataFrame(
        {
            "pk_1": [0, 1, 2, 3, 4, 5],
            "pk_2": ["zero", "one", "two", "three", "four", "five"],
            "animals": [
                "cat",
                "fish",
                "dog",
                "giraffe",
                "lion",
                "zebra",
            ],
        }
    )
    test_df: "pyspark.sql.dataframe.DataFrame" = spark_session.createDataFrame(
        data=df,
    )  # noqa: F821
```

We could run the `expect_column_values_to_be_in_set` Expectation on the `animals` column  with  `["cat", "fish", "dog"]` as the `value_set` (ie domestic animals).

```python
expectationConfiguration = ExpectationConfiguration(
    expectation_type="expect_column_values_to_be_in_set",
    kwargs={
        "column": "animals",
        "value_set": ["cat", "fish", "dog"],
    },
)
```
We could run a `Checkpoint` that contains the `ExpectationConfiguration` with the following `result_format` configuration, which sets `unexpected_index_columns` to be `pk_1`: 

```python
result_format: dict = {
        "result_format": "COMPLETE",
    }
```

After running the `Checkpoint`, we expect the validation result to contain the following `unexpected_index_list`: 
```python
[{"animals": "giraffe", "pk_1": 3},
{"animals": "lion", "pk_1": 4},
{"animals": "zebra", "pk_1": 5}]
```

#### What if I have a **lot** of rows in my table?

In order to retrieve the *full list* unexpected values from the table, the result also contains a `unexpected_index_query`, which can be used against the original date (represented by `df`), to retrieve the full set of unexpected results. 
 
```python
df.filter(F.expr((animals IS NOT NULL) AND (NOT (animals IN (cat, fish, dog))))
```
The query (along with `unexpected_index_list`) is returned as part of the result values returned by the `Validator`. 

```python 
...
"unexpected_index_list": [{"pk_1": 3}, {"pk_1": 4}, {"pk_1": 5}],
"unexpected_index_query": "df.filter(F.expr((animals IS NOT NULL) AND (NOT (animals IN (cat, fish, dog))))"
...
```

### Definition of Done


tests/expectations/metrics/test_map_metric.py
```diff
@@ -825,112 +800,6 @@ def test_spark_single_column_summary_result_format(
     }
 
 
-# Spark
-def test_spark_single_column_complete_result_format(
```
Shinnnyshinshin, line 829: these were duplicate tests 😅

great_expectations/expectations/metrics/map_metric_provider.py
```diff
@@ -2634,23 +2633,23 @@ def _spark_column_map_condition_value_counts(
...
-
     column_name = get_dbms_compatible_column_names(
         column_names=column_name,
         batch_columns_list=metrics["table.columns"],
         execution_engine=execution_engine,
     )
 
+    df: "pyspark.sql.dataframe.DataFrame" = execution_engine.get_domain_records(
+        domain_kwargs=compute_domain_kwargs
+    )
```
Shinnnyshinshin, line 2633: small refactor: type hinting added and df instantiated closer to where it is first being used.

great_expectations/expectations/expectation.py
```diff
@@ -2608,6 +2609,34 @@ def get_validation_dependencies(
                         metric_value_kwargs=metric_kwargs["metric_value_kwargs"],
                     ),
                 )
+        if isinstance(execution_engine, SparkDFExecutionEngine):
```
anthonyburdi: Are these sections the same? If so maybe you could change the if to check either SQL or Spark e.g. `if isinstance(execution_engine, (SqlAlchemyExecutionEngine, SparkDFExecutionEngine)):`

great_expectations/expectations/metrics/map_metric_provider.py
anthonyburdi: I also like that we now do error checking first.

great_expectations/expectations/metrics/map_metric_provider.py
```diff
@@ -2677,6 +2676,120 @@ def _spark_map_condition_rows(
...
+        domain_kwargs=domain_kwargs
+    )
+
+    result_format = metric_value_kwargs["result_format"]
+    if not result_format.get("unexpected_index_column_names"):
+        raise ge_exceptions.MetricResolutionError(
+            "unexpected_indices cannot be returned without 'unexpected_index_column_names'. Please check your configuration."
+        )
+
+    # withColumn is required to transform window functions returned by some metrics to boolean mask
```
anthonyburdi, line 2722: :heart:

great_expectations/expectations/metrics/map_metric_provider.py
```diff
@@ -2677,6 +2676,120 @@ def _spark_map_condition_rows(
...
+    for row in filtered.collect():
+        dict_to_add: dict = {}
+        for col_name in columns_to_keep:
+            dict_to_add[col_name] = row[col_name]
+        unexpected_index_list.append(dict_to_add)
+
+    if result_format["result_format"] == "COMPLETE":
+        return unexpected_index_list
+    return unexpected_index_list[: result_format["partial_unexpected_count"]]
+
```
anthonyburdi: Do you think we can do the filtering before the collect statement for the case where we are not returning the COMPLETE results? Otherwise it looks like here we are retrieving all of the unexpected values to the machine running GX and then filtering them if COMPLETE is not specified. I'm just concerned that this may cause undesired extra computation / transfer (especially for large datasets).

great_expectations/expectations/metrics/map_metric_provider.py
```diff
@@ -2677,6 +2676,120 @@ def _spark_map_condition_rows(
...
+    if return_unexpected_index_query is False:
+        return None
+
+    (
+        unexpected_condition,
+        _,
+        _,
+    ) = metrics.get("unexpected_condition", (None, None, None))
+
+    # unexpected_condition is an F.column object, meaning the str representation is wrapped in Column<> syntax.
```
anthonyburdi, line 2785: :heart:

anthonyburdi (changes requested):
Looking great! Just a few items to address. Happy to discuss!

great_expectations/expectations/metrics/map_metric_provider.py
anthonyburdi: It would be good to double check for this in the SQL case as well.

great_expectations/expectations/expectation.py
Shinnnyshinshin: great catch. Thank you :)

great_expectations/expectations/metrics/map_metric_provider.py
Shinnnyshinshin: great idea. I'll look into making the change both here and at the SQL stage

Shinnnyshinshin: This is already done for SQL

Shinnnyshinshin (commented):
Thank you for the review @anthonyburdi. Added the requested changes

anthonyburdi (approved):
Thanks for making those changes!

mattslack-db:
As far as I can tell this approach only works when you pass in the `result_format` into the call to `run_checkpoint`. For example, it doesn't get passed through when you call `validate` on a `SparkDFDataset`. Is this by design? 

For example:

```
gdf = SparkDFDataset(some_df)

gdf.validate(expectation_suite, result_format={"result_format": "COMPLETE", "unexpected_index_column_names": ["key"]})
```

Does not populate the `unexpected_index_list`.

Thanks

Shinnnyshinshin:
Hi @mattslack-db thanks for reaching out :) yes this is a design decision due to the fact that we aren't able to persist Non-Checkpoint-level configurations for `result_format`.  A bit more background and context behind the change can be found here: https://github.com/great-expectations/great_expectations/pull/6562

Also I believe `SparkDFDataset` is a deprecated way of interacting with GX at this time, so there might be some additional issues that you run into because of that

PR #6678, opened 2022-12-30
[MAINTENANCE] Pull functionality up from sql_datasource.py to interfaces.py

billdirks:
This PR is refactor of sql_datasource.py. We pull up splitters and sorters into interfaces.py since they are more general than sql datasources. Also, a lot of the functionality in `TableAsset` should be applicable to `DataAsset`s in general so I pulled that up. I will highlight any changes that weren't moving things around.

### DIFF
--- a/great_expectations/experimental/datasources/interfaces.py
+++ b/great_expectations/experimental/datasources/interfaces.py
@@ -156,7 +156,7 @@ class DataAsset(ExperimentalBaseModel):
         self._sort_batches(batch_list)
         return batch_list
 
-    def _batch_from_fully_specified_batch_request(self, request: BatchRequest):
+    def _batch_from_fully_specified_batch_request(self, request: BatchRequest) -> Batch:
         """Returns a Batch object from a fully specified batch request
 
         Args:
--- a/great_expectations/experimental/datasources/interfaces.py
+++ b/great_expectations/experimental/datasources/interfaces.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
 import dataclasses
+import itertools
 import logging
+from datetime import datetime
 from pprint import pformat as pf
 from typing import (
     TYPE_CHECKING,
@@ -10,13 +12,19 @@ from typing import (
     Generic,
     List,
     Mapping,
+    NamedTuple,
     Optional,
+    Sequence,
     Set,
     Type,
     TypeVar,
+    Union,
+    cast,
 )
 
 import pydantic
+from pydantic import Field
+from pydantic import dataclasses as pydantic_dc
 from typing_extensions import ClassVar, TypeAlias
 
 from great_expectations.experimental.datasources.experimental_base_model import (
@@ -36,6 +44,8 @@ if TYPE_CHECKING:
     )
     from great_expectations.execution_engine import ExecutionEngine
 
+
+# Data Asset Interface
 # BatchRequestOptions is a dict that is composed into a BatchRequest that specifies the
 # Batches one wants returned. The keys represent dimensions one can slice the data along
 # and the values are the realized. If a value is None or unspecified, the batch_request
@@ -43,6 +53,10 @@ if TYPE_CHECKING:
 # splitter and we want to query all months in the year 2020, the batch request options
 # would look like:
 #   options = { "year": 2020 }
+class BatchRequestError(Exception):
+    pass
+
+
 BatchRequestOptions: TypeAlias = Dict[str, Any]
 
 
@@ -53,13 +67,19 @@ class BatchRequest:
     options: BatchRequestOptions
 
 
+DataAssetSelf = TypeVar("DataAssetSelf", bound="DataAsset")
+
+
 class DataAsset(ExperimentalBaseModel):
     name: str
     type: str
+    column_splitter: Optional[ColumnSplitter] = None
+    order_by: List[BatchSorter] = Field(default_factory=list)
 
     # non-field private attrs
     _datasource: Datasource = pydantic.PrivateAttr()
 
+    # Properties
     @property
     def datasource(self) -> Datasource:
         return self._datasource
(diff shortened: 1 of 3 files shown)


great_expectations/experimental/datasources/interfaces.py
```diff
@@ -70,13 +90,203 @@ def datasource(self, ds: Datasource):
...
+
+        Returns:
+            A list of batches that match the options specified in the batch request.
+        """
+        # We translate the batch_request into a BatchSpec to hook into GX core.
+        self._validate_batch_request(batch_request)
+
+        batch_list: List[Batch] = []
+        for request in self._fully_specified_batch_requests(batch_request):
+            batch_list.append(self._batch_from_fully_specified_batch_request(request))
```
billdirks, line 155: This `for` loop use to have a bunch of logic inline which got moved to `_batch_from_fully_specified_batch_request` which is now an abstract method on this class.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
 import dataclasses
+import itertools
```
billdirks, line 4: The changes in this file are mostly pulling functionality up from `sql_datasources.py`. I will highlight any changes that happened beyond moving blocks of code.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -70,13 +90,203 @@ def datasource(self, ds: Datasource):
...
+        # We translate the batch_request into a BatchSpec to hook into GX core.
+        self._validate_batch_request(batch_request)
+
+        batch_list: List[Batch] = []
+        for request in self._fully_specified_batch_requests(batch_request):
+            batch_list.append(self._batch_from_fully_specified_batch_request(request))
+        self._sort_batches(batch_list)
+        return batch_list
+
+    def _batch_from_fully_specified_batch_request(self, request: BatchRequest):
```
billdirks: This is a new abstract method which encapsulates how a particular `DataAsset` creates a `Batch`. It's possible this could be generalized further and we may once we start implementing the pandas datasource.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -280,146 +125,63 @@ def add_year_and_month_splitter(
...
 
         Returns:
-            A list of batches that match the options specified in the batch request.
+            A single Batch object specified by the batch request
         """
-        # We translate the batch_request into a BatchSpec to hook into GX core.
-        self._validate_batch_request(batch_request)
-
-        batch_list: List[Batch] = []
+        batch_metadata = copy.deepcopy(request.options)
```
billdirks, line 125: This was the functionality that previously lived in the `for` loop of `get_batch_list_from_batch_request` and is unchanged.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -1,35 +1,25 @@
...
 import pydantic
-from pydantic import Field
 from pydantic import dataclasses as pydantic_dc
-from typing_extensions import ClassVar, Literal, TypeAlias
+from typing_extensions import ClassVar, Literal
 
 from great_expectations.core.batch_spec import SqlAlchemyDatasourceBatchSpec
-from great_expectations.experimental.datasources.interfaces import (
+
+# The noqa: F401 is to import BatchSorter. This is necessary for pydantic because
```
billdirks, line 12: There may be a better way to fix this. Pydantic is implicitly using `BatchSorter`. If I don't import `BatchSorter` I get pydantic errors when running tests (`order_by`, which is of type `List[BatchSorter]` is still a ForwardRef`).  Calling `TableAsset.update_forward_refs()` after defining `TableAsset` fails because `'BatchSorter' is not defined` so this requires the import even if it is not explicitly used. I tried because the error message suggested it, I haven't looked into what is going on. If I import, I do not need to manually call `update_forward_refs`. I haven't dug in anymore and am heading out for the day.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -260,3 +471,74 @@ def batch_spec(self) -> BatchSpec:
...
+
+
+BatchSortersDefinition: TypeAlias = Union[List[BatchSorter], List[str]]
+
+
+def _batch_sorter_from_list(sorters: BatchSortersDefinition) -> List[BatchSorter]:
+    if len(sorters) == 0 or isinstance(sorters[0], BatchSorter):
+        # mypy gets confused here. Since BatchSortersDefinition has all elements of the
+        # same type in the list so if the first on is BatchSorter so are the others.
+        return cast(List[BatchSorter], sorters)
```
Kilo59, line 526: We might want to use a `TypeGuard` for this.

great_expectations/experimental/datasources/sql_datasource.py
Kilo59: Does the `update_forward_refs()` still fail if we do an `if TYPE_CHECKING` import of `BatchSorter`?

I suspect this won't work, but might be worth trying.
https://docs.pydantic.dev/usage/postponed_annotations/

----

On a somewhat related note...
The above section points out that locally scoped module imports can cause problems for `pydantic` when it tries to resolve annotations into their "real" types.

I don't think this is necessary in this case, but if we ever find that global imports are leading to circular dependencies we could potentially deal with this by directly modifying the `module.__dict__` and updating it with whatever the needed `types` are.

> ### Warning
>To resolve strings (type names) into annotations (types), pydantic needs a namespace dict in which to perform the lookup. For this it uses module.__dict__, just like get_type_hints. This means pydantic may not play well with types not defined in the global scope of a module.

billdirks:
I'm going to hold off on merging this PR. I believe the column splitter might be sql specific. We often want to split by file names for pandas. I'm not sure if we want to split on both and need to look into this more.

PR #6681, opened 2023-01-02
[MAINTENANCE] initial type checking for `rules_based_profiler`

Kilo59:
- Begin type-checking select modules in `rules_based_profiler`

NOTE: apart from the items listed below there are no runtime changes as part of this PR
 1. Replace multiple `all()` checks with a new `TypeGuard` function https://github.com/great-expectations/great_expectations/pull/6681#discussion_r1060163253 - [`semantic_types_filter.py`](https://github.com/great-expectations/great_expectations/pull/6681/files/49b2c3524bdd31d870f9383281e844352bab7cd9#diff-ccbfe4894f957ac21827914a3a5c70fdb0ef6de7d35695795270241f6a8f43a6)


great_expectations/rule_based_profiler/builder.py
```diff
@@ -39,9 +39,9 @@ def __init__(
         Args:
             data_context: AbstractDataContext associated with this Builder
         """
-        self._batch_list = None
-        self._batch_request = None
-        self._data_context = data_context
+        self._batch_list: List[Batch] = []
```
Kilo59: This is one of the only runtime changes in this PR.

Kilo59: I'm going to have to revert this change.
There's at least one `if batch_list is None:` style check that makes this not work.

To make this work we would need to change these to `if not batch_list:`.

great_expectations/rule_based_profiler/parameter_builder/mean_table_columns_set_match_multi_batch_parameter_builder.py
```diff
@@ -122,8 +122,8 @@ def _build_parameters(
 
         one_batch_table_columns_names_value: MetricValue
         multi_batch_table_columns_names_sets_as_list: List[Set[str]] = [
-            set(one_batch_table_columns_names_value)
-            for one_batch_table_columns_names_value in table_columns_names_multi_batch_value
+            set({one_batch_table_columns_names_value})  # type: ignore[arg-type] # could be dict
```
Kilo59, line 125:
```suggestion
            set(one_batch_table_columns_names_value)  # type: ignore[arg-type] # could be dict
```

great_expectations/rule_based_profiler/helpers/simple_semantic_type_filter.py
```diff
@@ -60,7 +78,7 @@ def parse_semantic_domain_type_argument(
             return [semantic_type for semantic_type in [semantic_types]]
 
         if isinstance(semantic_types, list):
-            if all(isinstance(semantic_type, str) for semantic_type in semantic_types):
+            if _is_sequence_of(semantic_types, str):
```
Kilo59, line 81: Using a `TypeGuard` to help with type-narrowing

https://mypy.readthedocs.io/en/stable/type_narrowing.html#typeguards-with-parameters

```python
T = TypeVar("T")

def _is_sequence_of(sequence: Sequence, type_: Type[T]) -> TypeGuard[Sequence[T]]:
    return all(isinstance(x, type_) for x in sequence)
```

great_expectations/rule_based_profiler/domain_builder/column_domain_builder.py
```diff
@@ -403,7 +410,7 @@ def _resolve_list_type_property(
         property_value_type: Union[type, Tuple[type, ...]],
         variables: Optional[ParameterContainer] = None,
     ) -> List[type]:
-        property_value: Optional[property_value_type] = getattr(self, property_name, [])
```
Kilo59, line 406: This was not a valid annotation

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -506,16 +506,16 @@ def __init__(
...
     @classmethod
-    def from_commented_map(cls, commented_map: CommentedMap):  # type: ignore[no-untyped-def]
+    def from_commented_map(cls, commented_map: CommentedMap):  # type: ignore[override] # super type accepts Dict
         """Override parent implementation to pop unnecessary attrs from config.
 
         Please see parent BaseYamlConfig for more details.
         """
         try:
-            schema_instance: Any = cls._get_schema_instance()
+            schema_instance = cls._get_schema_instance()
```
alexsherstinsky: @Kilo59 Anything better we can do here?  Given that there are only so many Marshmallow Schema types involved in RBP, is that something we can use to not leave this variable without a type hint?  If not, no huge deal.  Thanks.

Kilo59: Ah, yes if these have to be a marshmallow schema I should be able to use that here.

I will make the change.

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -506,16 +506,16 @@ def __init__(
...
             config: Union[dict, BaseYamlConfig] = schema_instance.load(commented_map)
-            config.pop("class_name", None)
-            config.pop("module_name", None)
+            config.pop("class_name", None)  # type: ignore[union-attr] # BaseYamlConfig has no `.pop()`
```
alexsherstinsky, line 517: @Kilo59 Are we certain about that?  Given that `BaseYamlConfig` inherits dictionary behavior and `__delitem__()` is implemented, should `pop` not be automatically "understood"?  If not, then it is probably because it is not explicitly implemented, which might be confusing the type checker.  Thanks.

great_expectations/rule_based_profiler/config/base.py
```diff
@@ -618,13 +618,13 @@ def resolve_config_using_acceptable_arguments(
...
                 rule=rule,
-                variables_container=effective_variables,
+                variables_container=effective_variables,  # type: ignore[arg-type] # could be None
             )
             for name, rule in effective_rules_dict.items()
         }
 
         return cls(
-            name=profiler.config.name,
+            name=profiler.config.name,  # type: ignore[arg-type] # name could be None
```
alexsherstinsky, line 627: @Kilo59 From what I am seeing, the `name` of `RuleBasedProfiler` is not allowed to be `None`.  Moreover, in terms of Marshmallow Schema, we have:
```
    name = fields.String(
        required=True,
        allow_none=False,
    )
```

Please let me know what you think.  Thank you.

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, data_context: AbstractDataContext) -> None:
         """
         self._data_context = data_context
 
-        self._data_assistant_runner_cache = {}
+        self._data_assistant_runner_cache: dict = {}
```
alexsherstinsky, line 34: ❤️

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
```diff
@@ -109,6 +109,8 @@ class DataAssistant(metaclass=MetaDataAssistant):
         ...
     """
 
+    data_assistant_type: str  # Added by `MetaDataAssistant`
```
alexsherstinsky, line 112: ❤️ -- this is great!  What a simple way to make something that takes effort to see simple.  We have a much more egregious case of this in the `Metrics` system, whereby numerous properties are set via the Meta Class / Classes.  Any interest in making a similar improvement there?  I would be happy to provide some pointers.  Thank you!

great_expectations/rule_based_profiler/config/base.py
Kilo59: I haven't been able to track down why exactly this is the case. But `mypy` isn't able to verify that `name` can never be `None`.

I'll take another look but this might be something to tackle in a followup.

great_expectations/rule_based_profiler/data_assistant_result/onboarding_data_assistant_result.py
```diff
@@ -8,7 +8,7 @@
 
 class OnboardingDataAssistantResult(DataAssistantResult):
     @property
-    def metric_expectation_map(self) -> Dict[Union[str, Tuple[str]], str]:
+    def metric_expectation_map(self) -> Dict[Union[str, Tuple[str, ...]], str]:
```
alexsherstinsky, line 11: @Kilo59 There is also apparently the formal type `Ellipsis` that translates to the three dots -- might be worth trying to see if it works as described.  Thanks.

great_expectations/rule_based_profiler/domain_builder/column_domain_builder.py
```diff
@@ -315,51 +326,47 @@ def get_effective_column_names(
...
+                    lambda candidate_column_name: self.semantic_type_filter.table_column_name_to_inferred_semantic_domain_type_map[  # type: ignore[union-attr,arg-type]
                         candidate_column_name
                     ]
                     in include_semantic_types,
                     effective_column_names,
                 )
             )
 
-        exclude_semantic_types: Union[List[Union[str, SemanticDomainTypes]]] = cast(
+        exclude_semantic_types: List[Union[str, SemanticDomainTypes]] = cast(
```
alexsherstinsky, line 354: @Kilo59 Great catch of the typo.  Thanks!

great_expectations/rule_based_profiler/domain_builder/column_domain_builder.py
```diff
@@ -315,51 +326,47 @@ def get_effective_column_names(
             },
             config_defaults={},
         )
-        self._semantic_type_filter = semantic_type_filter
+        self._semantic_type_filter = semantic_type_filter  # type: ignore[assignment] # could be None
 
-        include_semantic_types: Union[List[Union[str, SemanticDomainTypes]]] = cast(
+        include_semantic_types: List[Union[str, SemanticDomainTypes]] = cast(
```
alexsherstinsky, line 331: @Kilo59 Great catch of the typo.  Thanks!

great_expectations/rule_based_profiler/domain_builder/column_domain_builder.py
alexsherstinsky: @Kilo59 Is it because dynamic annotations are disallowed?  Anything we can do other than "nothing"?  Thanks.

great_expectations/rule_based_profiler/domain_builder/column_pair_domain_builder.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Dict, List, Optional, Set, Union
+from typing import TYPE_CHECKING, ClassVar, Dict, List, Optional, Set, Union
```
alexsherstinsky, line 3: ❤️

great_expectations/rule_based_profiler/estimators/exact_numeric_range_estimator.py
```diff
@@ -56,8 +60,8 @@ def _get_numeric_range_estimate(
             parse_strings_as_datetimes=True,
             fuzzy=False,
         )
-        min_value: MetricValue = np.amin(a=metric_values_converted)
-        max_value: MetricValue = np.amax(a=metric_values_converted)
+        min_value = np.amin(a=metric_values_converted)
```
alexsherstinsky: @Kilo59 How come removing the `MetricValue` type hint is proposed here?  Thanks.

great_expectations/rule_based_profiler/estimators/numeric_range_estimator.py
```diff
@@ -94,7 +94,9 @@ def to_dict(self) -> dict:
         """
         Returns dictionary equivalent of this object.
         """
-        dict_obj: dict = self._configuration.to_dict()
+        dict_obj: dict = (
+            self._configuration.to_dict()  # type: ignore[union-attr] # could be None
```
alexsherstinsky, line 98: @Kilo59 Are we sure that a `None` configuration can happen?

great_expectations/rule_based_profiler/estimators/quantiles_numeric_range_estimator.py
```diff
@@ -22,10 +25,13 @@
...
 from great_expectations.util import convert_ndarray_to_datetime_dtype_best_effort
 
+if TYPE_CHECKING:
+    import numpy.typing as npt
+
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
-DEFAULT_QUANTILES_QUANTILE_STATISTIC_INTERPOLATION_METHOD = "nearest"
+DEFAULT_QUANTILES_QUANTILE_STATISTIC_INTERPOLATION_METHOD: Final[str] = "nearest"
```
alexsherstinsky, line 34: @Kilo59 Is `Final` in `Python`?  Sorry, I did not see it imported or defined.  Thank you.

great_expectations/rule_based_profiler/estimators/quantiles_numeric_range_estimator.py
```diff
@@ -71,7 +77,7 @@ def _get_numeric_range_estimate(
             )
 
         datetime_detected: bool = datetime_semantic_domain_type(domain=domain)
-        metric_values_converted: np.ndaarray
+        metric_values_converted: npt.NDArray
```
alexsherstinsky, line 80: ❤️ `npt.NDArray` looks like a very powerful and useful type.  Thanks!

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
```diff
@@ -136,8 +146,8 @@ def _build_parameters(
         )
         details: dict = parameter_node[FULLY_QUALIFIED_PARAMETER_NAME_METADATA_KEY]
 
-        unique_values: Set[Any] = _get_unique_values_from_nested_collection_of_sets(
-            collection=metric_values
+        unique_values: Set = _get_unique_values_from_nested_collection_of_sets(
```
alexsherstinsky: @Kilo59 I believe that in this case, we should just use `set` (instead of `typing.Set`) or `Set[V]` like you did below.  Thank you.

great_expectations/rule_based_profiler/parameter_container.py
```diff
@@ -13,41 +13,48 @@
...
-RAW_PARAMETER_KEY: str = (
-    f"{PARAMETER_KEY}{RAW_SUFFIX}{FULLY_QUALIFIED_PARAMETER_NAME_SEPARATOR_CHARACTER}"
-)
-
-FULLY_QUALIFIED_PARAMETER_NAME_ATTRIBUTED_VALUE_KEY: str = "attributed_value"
-FULLY_QUALIFIED_PARAMETER_NAME_METADATA_KEY: str = "details"
-FULLY_QUALIFIED_PARAMETER_NAME_VALUE_KEY: str = "value"
-
-RESERVED_TERMINAL_LITERALS: Set[str] = {
+FULLY_QUALIFIED_PARAMETER_NAME_DELIMITER_CHARACTER: Final[str] = "$"
```
alexsherstinsky, line 23: @Kilo59 Again, love the use of `Final` -- does it need to be imported/defined?  Thanks.

alexsherstinsky (changes requested):
@Kilo59 Great stuff.  I am mostly asking questions (with a couple of suggestions).  Thanks!

great_expectations/rule_based_profiler/config/base.py
alexsherstinsky: 🙇

great_expectations/rule_based_profiler/config/base.py
alexsherstinsky: 🙇

great_expectations/rule_based_profiler/parameter_container.py
Kilo59: It is imported on line `16`

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
Kilo59: The `V` type var can't be used here.
`Set`, `set`, and `Set[Any]` are all equivalent.

Also, I like to remove the `Any` import whenever possible to discourage people from using it 😄

great_expectations/rule_based_profiler/data_assistant_result/onboarding_data_assistant_result.py
Kilo59: >To specify a variable-length tuple of homogeneous type, use literal ellipsis, e.g. `Tuple[int, ...]`

https://docs.python.org/3/library/typing.html#typing.Tuple

great_expectations/rule_based_profiler/estimators/quantiles_numeric_range_estimator.py
Kilo59: Yes, it's imported above from `typing_extensions`.
https://mypy.readthedocs.io/en/stable/final_attrs.html?highlight=Final#final-names-methods-and-classes

https://docs.python.org/3/library/typing.html#typing.Final

great_expectations/rule_based_profiler/domain_builder/column_domain_builder.py
Kilo59: It's possible we could do something with a `TypeVar`. I'll revisit this.

great_expectations/rule_based_profiler/config/base.py
Kilo59: AFIAK this isn't going to work.
There's no mechanism for `mypy` to know about the schema fields that have been validated.
We can annotate the classes based on how they are defined in the schema, but we can't infer them from the schemas. Our runtime, business objects are still just "weakly" typed `DictDot`s underneath. 

I can show you how we could do some static enforcement, but I don't think that work should be part of this initial effort.

great_expectations/rule_based_profiler/config/base.py
Kilo59: `DictDot` doesn't inherit from `dict` or any other `MutableSequence` that defines a `pop()`.
If it did, the type-check would understand that `.pop()` is defined in one of the parent classes.

https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/types/__init__.py#L24

great_expectations/rule_based_profiler/config/base.py
Kilo59: Done

great_expectations/rule_based_profiler/estimators/exact_numeric_range_estimator.py
Kilo59: Because `build_numeric_range_estimation_result()` expects numbers for `min_value` and `max_value`. `MetricValue` is a much broader union of types.

>great_expectations/rule_based_profiler/estimators/exact_numeric_range_estimator.py:69: error: Argument "min_value" to "build_numeric_range_estimation_result" has incompatible type "Union[Any, List[Any], Set[Any], Tuple[Any, ...], Any, Any, ndarray[Any, Any], int, str, float, bool, Dict[str, Any]]"; expected "Number"  [arg-type]

great_expectations/rule_based_profiler/domain_builder/column_domain_builder.py
Kilo59: The possible `Tuple` of types complicates the use of a `TypeVar` here.
Without it, I could declare that the return type is the same type as the `property_value_type` type.

great_expectations/rule_based_profiler/estimators/numeric_range_estimator.py
Kilo59: Yes. `configuration` / `self._configuration` defaults to `None`.
```python
class NumericRangeEstimator(ABC, SerializableDictDot):
    """
    Parent class that incorporates the "get_numeric_range_estimate()" interface method, requiring all subclasses to
    implement the "_get_numeric_range_estimate()" method (for encapsulation reasons, the former calls the latter).
    """

    def __init__(
        self,
        name: str,
        configuration: Optional[Attributes] = None,
    ) -> None:
        """
        Args:
            name: the name of this estimator, which encodes the choice of the estimation algorithm: "quantiles",
            "bootstrap", "exact" (default - deterministic, incorporating entire observed value range), or "kde"
            (kernel density estimation).
            configuration: attributes needed for the estimation algorithm (subject of the inherited class) to operate.
        """
        self._name = name
        self._configuration = configuration
```

great_expectations/rule_based_profiler/config/base.py
alexsherstinsky: @Kilo59 `DictDot` fakes the dictionary behavior.  We can leave the ignore in.

great_expectations/rule_based_profiler/config/base.py
alexsherstinsky: @Kilo59 Understandable.  Thanks.

great_expectations/rule_based_profiler/estimators/exact_numeric_range_estimator.py
alexsherstinsky: @Kilo59 This warning is not showing up in my `PyCharm`, which is strange.  Regardless, what if we change the type hint to `Number`?  Thanks.

great_expectations/rule_based_profiler/estimators/numeric_range_estimator.py
alexsherstinsky: @Kilo59 You are correct.  Sorry -- and thanks.

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
alexsherstinsky: @Kilo59 Fine -- to me, having `Set` without anything in the brackets seems sloppy, but fine.  Perhaps if you make another change to this PR, you could change it to `set`; otherwise, we can leave it be (or do it later).  Thanks.

alexsherstinsky (approved):
LGTM (provided all tests pass, and a couple of other comments/remarks are read and any action is applied as you deem appropriate).  Thanks.

great_expectations/rule_based_profiler/estimators/exact_numeric_range_estimator.py
Kilo59: Done

great_expectations/rule_based_profiler/parameter_builder/value_set_multi_batch_parameter_builder.py
Kilo59: changed to `set`

Issue #6625, opened 2022-12-21
[BUG] Duplicated expectations in validation results when metric resolution fail


itaise:
Hi,
We are using great expectations on Trino, and when the expectation fails on syntax error, for example a typo in the name of a column in the expectation configuration, the validation result is duplicated 3 times.
So - our data docs, slack message and Postgres logs show this expectation 3 times instead of one. 

I investigated a bit and i see that GX is trying to resolve the metric 3 times [in this part](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/validator/validation_graph.py#L269) and getting a MetricResolutionError. 
3 times is because this is the maximum metric computation retries [as defined here](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/validator/validation_graph.py#L34).

Then, the metric is aborted, and  in [this part](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/validator/validator.py#L1189) - metric_exception_info is **a set** with 3 exceptions (instead of one), of the metric that was "aborted" because of the query error. 

By the way - interesting fact -  the reason that the set "metric_exception_info" has 3 values and not one, is because **trino returns an exception message that includes the Trino query id**. and in each try (out of the three), the trino query id is different obviously. 

Then - the "evrs" variable is a list of 3 elements, instead of one, and the validation results get it after that.. 

Expected behaviour - 
When metric resolution failed, even though it failed 3 times, it should not duplicate the validation results.

 GX version - 0.15.41

Can you please check? thanks a lot!

itaise:
A suggestion - maybe in [this part](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/validator/validator.py#L1192) it should be checked that this metric name is not already in evrs?
I can raise a PR if you think its a good solution.

austiezr:
Hey @itaise! Thanks for raising (and your patience over the holidays)! This seems reasonable and we would be happy to look at a PR to resolve if you're willing to contribute. 🚀

itaise:
Thanks @austiezr! will open a PR

talagluck:
Resolved by #7373 - thanks @itaise!

PR #6697, opened 2023-01-04
Remove add exception info, if exist already for the metric

itaise:
This PR is to solve this issue: 

https://github.com/great-expectations/great_expectations/issues/6625

When a metric fails, It adds the exception that was raised 3 times to the failed metric info. 
This cause the validation result (sent in the checkpoint actions, e.g. in our slack / custom actions) to be duplicated 3 times with the same error (e.g. column not found).

After this fix, the first exception is saved in the validation results, without duplication.


amitbend (approved):

kenwade4:
FYI @amitbend a different solution is being worked on by @itaise since he discovered that it only seems to happen on Expectations that use the query.template_values metric.

kenwade4:
@itaise please open a new PR whenever you have that new solution 👍

PR #6703, opened 2023-01-05
[MAINTENANCE] update `cli` `DataContext` types

Kilo59:
- add `FileDataContext` overload for `DataContext` function
- update CLI `data_context` types to reflect that they are currently always `FileDataContext`s
- `cli/cli.py` and `cli/datasource.py` under type-coverage


great_expectations/data_context/data_context/data_context.py
```diff
@@ -24,6 +26,43 @@
     from great_expectations.alias_types import PathStr
 
 
+@overload
```
cdkini, line 29: For each overload, should we add a docstring or comment to describe what situation they're used in? I know the answer here since I worked on it but it might be helpful for others?

great_expectations/cli/datasource.py
```diff
@@ -195,7 +203,7 @@ def _datasource_new_flow(
         selected_files_backend = _prompt_for_execution_engine()
         helper = _get_files_helper(
             selected_files_backend,
-            context_root_dir=context.root_directory,
+            context_root_dir=context.root_directory,  # type: ignore[arg-type] # could be None
```
Kilo59, line 206: Seems like we could the `root_directory` for `FileDataContext` should never be `None`.
I'll try to correct this in a followup.

great_expectations/data_context/data_context/data_context.py
Kilo59: That sounds like a good idea.
I probably won't have the best explanation but I'll try 😄

great_expectations/cli/datasource.py
```diff
@@ -204,7 +212,11 @@ def _datasource_new_flow(
         selected_database = _prompt_user_for_database_backend()
         helper = _get_sql_yaml_helper_class(selected_database, datasource_name)
     else:
-        helper = None
+        helper = None  # type: ignore[assignment] # causes error below
+        raise AttributeError(
+            "No `helper` selected"
+        )  # `None` has no `send_backend_choice_usage_message()`.
+        # Explicitly raised so that mypy understands that helper is not `None` below this line
```
Kilo59, line 219: Seems like we might want a cleaner exit here. But this keeps the existing runtime behavior.

great_expectations/cli/datasource.py
```diff
@@ -204,7 +212,11 @@ def _datasource_new_flow(
...
 
     helper.send_backend_choice_usage_message(context)
```
Kilo59, line 221: Seems like we might want a cleaner exit here.
But this preserves the existing runtime behavior.

great_expectations/data_context/data_context/data_context.py
Kilo59: @cdkini Done

cdkini (approved):

PR #6716, opened 2023-01-06
[MAINTENANCE] ID/PK result_format documentation update

Shinnnyshinshin:
## Proposed Changes
* Updated `result_format` doc to include changes to `unexpected_index_column_names` with the ID/PK work, as well as `return_unexpected_index_query`. 
  * Examples have been consolidated and updated, and now use references to result format integration test file 
* Added `result_format` integration test that shows `Validator`-level and `Checkpoint`-level configurations
### Definition of Done


docs/reference/expectations/result_format.md
```diff
@@ -4,66 +4,62 @@ title: Result format
 
 
 The `result_format` parameter may be either a string or a dictionary which specifies the fields to return in `result`.
-  * For string usage, see `result_format` values.
+  * For string usage, see [`result_format` values and fields](#result_format-values-and-fields).
```
anthonyburdi: Nit - What about limiting the link to the text after the inline code snippet? It just gets rendered a little wonky (with the underline and the underscore of `result_format`. 
image

```suggestion
  * For string usage, see `result_format` [values and fields](#result_format-values-and-fields).
```

docs/reference/expectations/result_format.md
```diff
@@ -4,66 +4,62 @@ title: Result format
...
   * For dictionary usage, `result_format` which may include the following keys:
     * `result_format`: Sets the fields to return in result.
+    * `unexpected_index_column_names`: Defines the primary key (PK) columns used to represent unexpected results.
```
anthonyburdi: How about something like the following? Just to be more general that a column does not have to be the primary key column and explicit that multiple columns can be used.
```suggestion
    * `unexpected_index_column_names`: Defines columns that can be used to identify unexpected results, for example primary key (PK) column(s) or other columns with unique identifiers. Supports multiple column names as a list.
```

docs/reference/expectations/result_format.md
```diff
@@ -4,66 +4,62 @@ title: Result format
...
+    * `return_unexpected_index_query`: When running validations, a query (or a set of indices) will be returned that will
+      allow you to retrieve the full set of unexpected results defined by the PK.  Setting this value to `False` will 
+      suppress the output (default is `True`).
```
anthonyburdi: Nit - What do you think of this?
```suggestion
    * `return_unexpected_index_query`: When running validations, a query (or a set of indices) will be returned that will
      allow you to retrieve the full set of unexpected results including any columns identified in `unexpected_index_column_names`.  Setting this value to `False` will 
      suppress the output (default is `True`).
```

docs/reference/expectations/result_format.md
```diff
@@ -4,66 +4,62 @@ title: Result format
...
-validation_result = validator.expect_column_values_to_be_between(
-    column="pickup_location_id",
-    min_value=0,
-    max_value=100,
-    result_format="COMPLETE",
-    include_unexpected_rows=True
-)
-unexpected_index_list = validation_result["result"]["unexpected_index_list"]
-unexpected_list = validation_result["result"]["unexpected_list"]
+To apply `result_format` to an Expectation, pass it into the Expectation. We will first need to obtain a Validator object instance by running the `$ great_expectations suite new` command.
```
anthonyburdi: Nit: Just set off the CLI in a parenthetical since there are other ways to set up a validator (the CLI just generating the boilerplate).
```suggestion
To apply `result_format` to an Expectation, pass it into the Expectation. We will first need to obtain a Validator object instance (e.g. by running the `$ great_expectations suite new` command).
```

docs/reference/expectations/result_format.md
```diff
@@ -4,66 +4,62 @@ title: Result format
...
+
+```python name="result_format_complete_example_set"
 ```
-When configured at the Expectation level, the `unexpected_index_list` and `unexpected_list` won't be passed through to the final Validation Result object.
-In order to see those values at the Suite level, configure `result_format` in your Checkpoint configuration.
+
 ### Checkpoint Level Config
 To apply `result_format` to every Expectation in a Suite, define it in your Checkpoint configuration under the `runtime_configuration` key.
```
anthonyburdi, line 37: Just a related question: Does this also work if there are multiple validations (e.g. each with an expectation suite)? Can you specify a validation-specific result_format?

docs/reference/expectations/result_format.md
```diff
@@ -4,66 +4,62 @@ title: Result format
...
-    }
-}
+
+```python name="result_format_checkpoint_example"
 ```
+
 The results will then be stored in the Validation Result after running the Checkpoint.
 :::note
-Regardless of where Result Format is configured, `unexpected_list` and `unexpected_index_list` are never rendered in Data Docs.
+The `unexpected_index_list`, as represented by primary key (PK) columns, is
```
anthonyburdi, line 44:
```suggestion
The `unexpected_index_list`, as represented by columns selected via `unexpected_index_column_names`, is 
```

docs/reference/expectations/result_format.md
```diff
@@ -96,112 +93,35 @@ cases for working with Great Expectations, including interactive exploratory wor
...
-```python
-my_df.expect_column_values_to_be_in_set(
-    "my_var",
-    ["B", "C", "D", "F", "G", "H"],
-    result_format={'result_format': 'BOOLEAN_ONLY'}
-)
-{
-    'success': False
-}
+```python name="pandas_df_for_result_format"
```
anthonyburdi: Totally optional nitpick - do you think this dataframe can be reduced in size so that each of the columns can fit on a single line e.g. `"my_var": ["A", "B", "B", "C", "C", "D", ... <etc>],` instead of 
```
"my_var": [
            "A",
            "B",
            "B",
            "C",
            "C",
            "C",
            "D",
            "D",
            "D",
            "D",
            "E",
            "E",
            "E",
            "E",
            "E",
        ],
```

docs/reference/expectations/result_format.md
```diff
@@ -298,71 +222,78 @@ The summary `result` includes:
         "partial_unexpected_list" : [A list of up to 20 values that violate the expectation] (also in `BASIC`)
         'missing_count': The number of missing values in the column
         'missing_percent': The total percent of missing values in the column
-        'partial_unexpected_counts': [{A list of objects with value and counts, showing the number of times each of the unexpected values occurs}]
-        'partial_unexpected_index_list': [A list of up to 20 of the indices of the unexpected values in the column]
+        'partial_unexpected_counts': [{A list of objects with value and counts, showing the number of times each of the unexpected values occurs}
+        'partial_unexpected_index_list': [A list of up to 20 of the indices of the unexpected values in the column, as defined by the primary key column]
```
anthonyburdi:
```suggestion
        'partial_unexpected_index_list': [A list of up to 20 of the indices of the unexpected values in the column, as defined by the columns in `unexpected_index_column_names`]
```

docs/reference/expectations/result_format.md
```diff
@@ -298,71 +222,78 @@ The summary `result` includes:
...
-        'unexpected_count': 6,
-        'unexpected_percent': 0.16666666666666666,
-        'unexpected_percent_nonmissing': 0.16666666666666666,
-        'missing_count': 0,
-        'missing_percent': 0.0,
-        'partial_unexpected_counts': [{'value': 'A', 'count': 1}, {'value': 'E', 'count': 5}],
-        'partial_unexpected_index_list': [0, 10, 11, 12, 13, 14],
-        'partial_unexpected_list': ['A', 'E', 'E', 'E', 'E', 'E']
-    }
+    "element_count": 15,
```
anthonyburdi: Should this example (and others for validation_result.success and .result) come from snippets too? That way they are under test and then change automatically if the test script changes? Maybe an assert statement in the script below the script tag if we don't want to show the assert in the docs? E.g. 
```
<snippet name="some name">
expected_validation_result = {
   ... (fill this in)
}
</snippet>
assert validation_result.result == expected_validation_result
```

docs/reference/expectations/result_format.md
```diff
@@ -298,71 +222,78 @@ The summary `result` includes:
...
 
 ```python
 {
     'success': False,
     'result': {
         'observed_value': The aggregate statistic computed for the column (also in `BASIC`)
-        'element_count': The total number of values in the column
-        'missing_count':  The number of missing values in the column
-        'missing_percent': The total percent of missing values in the column
-        'details': {<expectation-specific result justification fields>}
```
anthonyburdi, line 341: Oh wow - good catch.

docs/reference/expectations/result_format.md
```diff
@@ -298,71 +222,78 @@ The summary `result` includes:
...
-    "success" : Boolean,
-    "result" : {
-        "observed_value" : 1.5,
-        'element_count': 5,
-        'missing_count': 1,
-        'missing_percent': 0.2
-    }
-}
+>>> print(validation_result.result)
+{'observed_value': 3.6666666666666665}
```
anthonyburdi: Similar comment about adding these to the test script.

docs/reference/expectations/result_format.md
```diff
@@ -379,7 +310,8 @@ The complete `result` includes:
     'success': False,
     'result': {
         "unexpected_list" : [A list of all values that violate the expectation]
-        'unexpected_index_list': [A list of the indices of the unexpected values in the column]
+        'unexpected_index_list': [A list of the indices of the unexpected values in the column, as defined by primary key column]
```
anthonyburdi:
```suggestion
        'unexpected_index_list': [A list of the indices of the unexpected values in the column, as defined by the columns in `unexpected_index_column_names`]
```

docs/reference/expectations/result_format.md
```diff
@@ -391,20 +323,47 @@ The complete `result` includes:
...
+```python name="result_format_complete_example_set"
+```
 
+Will return the following output:
 ```python
+>>> print(validation_result.success)
+False
+
+>>> print(validation_result.result)
 {
```
anthonyburdi: Same comment about adding these to the test script.

docs/reference/expectations/result_format.md
```diff
@@ -419,28 +378,20 @@ The complete `result` includes:
     'success': False,
     'result': {
         'observed_value': The aggregate statistic computed for the column (also in `SUMMARY`)
-        'element_count': The total number of values in the column (also in `SUMMARY`)
-        'missing_count':  The number of missing values in the column (also in `SUMMARY`)
-        'missing_percent': The total percent of missing values in the column (also in `SUMMARY`)
```
anthonyburdi, line 424: Good catch. Just to verify - these don't appear for all expectations? Any idea why they were included here (i.e. were they available at one point but not any longer and we didn't update the docs)?

docs/reference/expectations/result_format.md
```diff
@@ -419,28 +378,20 @@ The complete `result` includes:
...
-        "observed_value" : 1.5,
-        'element_count': 5,
-        'missing_count': 1,
-        'missing_percent': 0.2
-    }
-}
-```
+>>> print(validation_result.result)
+{'observed_value': 3.6666666666666665}
+```
```
anthonyburdi: Same comment about adding these to the test script.

docs/reference/expectations/result_format.md
```diff
@@ -96,112 +93,35 @@ cases for working with Great Expectations, including interactive exploratory wor
...
-my_df.expect_column_values_to_be_in_set(
-    "possible_benefactors",
-    ["Joe Gargery", "Mrs. Gargery", "Mr. Pumblechook", "Ms. Havisham", "Mr. Jaggers", "Mr. Magwitch"]
-    result_format={'result_format': 'BOOLEAN_ONLY'}
-)
-{
-    'success': False
-}
+>>> print(validation_result.result)
+{}
```
anthonyburdi: Same comment about adding these to the test script (see below).

docs/reference/expectations/result_format.md
```diff
@@ -226,23 +146,29 @@ The basic `result` includes:
...
+    "element_count": 15,
+    "unexpected_count": 6,
+    "unexpected_percent": 40.0,
+    "partial_unexpected_list": ["A", "E", "E", "E", "E", "E"],
+    "missing_count": 0,
+    "missing_percent": 0.0,
+    "unexpected_percent_total": 40.0,
+    "unexpected_percent_nonmissing": 40.0,
 }
 ```
```
anthonyburdi: Same comment about adding these to the test script (see below).

docs/reference/expectations/result_format.md
```diff
@@ -259,21 +185,19 @@ The basic `result` includes:
...
 
-{
-    "success" : Boolean,
-    "result" : {
-        "observed_value" : 1.5
-    }
-}
+>>> print(validation_result.result)
+{'observed_value': 3.6666666666666665}
 ```
```
anthonyburdi: Same comment about adding these to the test script (see below).

great_expectations/expectations/expectation.py
```diff
@@ -1187,7 +1187,7 @@ def validate(
         if not configuration:
             configuration = deepcopy(self.configuration)
 
-        # issue warnings if necesary
+        # issue warnings if necessary
```
anthonyburdi, line 1199: 👍 💪

tests/integration/docusaurus/reference/core_concepts/result_format.py
```diff
@@ -0,0 +1,382 @@
...
+    execution_engine=engine,
+    data_context=context,
+    batches=[
+        batch,
+    ],
+)
+
+# Expectation-level Configuration
+# Snippet: result_format BOOLEAN example
+# <snippet name="result_format_boolean_example">
```
anthonyburdi: Heads up - @cdkini has been changing snippet names to use the filepath as a namespace - might be helpful here as well: https://greatexpectationslabs.slack.com/archives/C03B8DZCJ07/p1673271099464929

anthonyburdi (changes requested):
Lot of great stuff in here. Thank you for updating the document and adding a script under test! I have some suggestions and requested changes but nothing major.

docs/reference/expectations/result_format.md
Shinnnyshinshin: Love this change. thank you :)

docs/reference/expectations/result_format.md
Shinnnyshinshin: 👍🏼

docs/reference/expectations/result_format.md
```diff
@@ -4,66 +4,62 @@ title: Result format
 
 
 The `result_format` parameter may be either a string or a dictionary which specifies the fields to return in `result`.
-  * For string usage, see `result_format` values.
+  * For string usage, see `result_format` [values and fields](#result_format-values-and-fields).
   * For dictionary usage, `result_format` which may include the following keys:
```
Rachel-Reverie: This item doesn't read well -- the word "which" makes `result_format` seem insufficient of a subject.

I would change it to:
```
  * For dictionary usage, `result_format` may include the following keys:
```
or:
```
  * For dictionary usage, provide `result_format` as a dictionary, which may include the following keys:
```

(I think I prefer the first of the two.)

docs/reference/expectations/result_format.md
Shinnnyshinshin: Yes they "technically" work, but will not be persisted in the same way that a checkpoint-level configuration will. Basically everything gets munged together at the graph-validate step when the metrics are collected 😓

docs/reference/expectations/result_format.md
Shinnnyshinshin: Done. Thanks for the catch :)

docs/reference/expectations/result_format.md
Shinnnyshinshin: Done :) This was sortof bugging me too

tests/integration/docusaurus/reference/core_concepts/result_format.py
Shinnnyshinshin: Thanks for letting me know this. I made the adjustment :)

docs/reference/expectations/result_format.md
Shinnnyshinshin: Screen Shot 2023-01-09 at 11 18 52 PM

I had a stab at adjusting the output, and there is potentially a small inconsistency (in the formatting) that this introduces. It's pretty minor, so let me know if which direction you would like to go.  I'm happy either way.

anthonyburdi: How do we do this in other docs? I don't recall the repl style `>>>` in other docs. Maybe we can just show the assert statement? like:
```
assert validation_result.success
assert validation_result.result == {"observed_value": 3.6666666666666665}
```
I see we do that in a few places e.g. https://docs.greatexpectations.io/docs/guides/connecting_to_your_data/advanced/how_to_configure_a_dataconnector_for_splitting_and_sampling_tables_in_sql/#splitting-and-sampling

tests/integration/docusaurus/reference/core_concepts/result_format.py
anthonyburdi: I don't see the adjustment, maybe a commit didn't get pushed up?

docs/reference/expectations/result_format.md
Shinnnyshinshin: Great idea. I think this was the result of tension between preserving the existing docs as much as possible (which used the `>>>`) and making them accurate. I like the idea of displaying the `assert`. Update coming up ;)

Shinnnyshinshin: Screen Shot 2023-01-10 at 12 24 15 PM
Boom

anthonyburdi (approved):
Yay! More code under test. Thank you for making all those updates.

PR #6717, opened 2023-01-06
[FEATURE] ZEP - Synchronize & save XDatasources

Kilo59:
- Stash zep objects before "saving" in order to avoid the `instiate_class_from_config` logic and deepcopy/pickling issues.
- Keep track of `XDatasources` on the data_context
- dump XDatasources to yaml file as part of normal `_save_project_config()` call
- typing updates to `DataContextVariables` dataclass fields

### Definition of Done

Thank you for submitting!


great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -87,6 +87,16 @@ def _save_project_config(self) -> None:
...
+
+            zep_datasources = self._synchronize_zep_datasources()
+            if zep_datasources:
+                logger.info(
+                    f"Saving {len(self.zep_config.datasources)} ZEP Datasources to {config_filepath}"
+                )
+                with open(config_filepath, "r+") as io_file:
+                    yaml_dict = yaml.load(io_file)
+                    yaml.dump({**yaml_dict, **self.zep_config._json_dict()}, io_file)
+
```
cdkini: so the parent method just calls `self.variables.save_config()` - could we leverage that in any way?

great_expectations/data_context/data_context_variables.py
```diff
@@ -304,7 +310,7 @@ def _init_store(self) -> DataContextStore:
 
 @dataclass(repr=False)
 class FileDataContextVariables(DataContextVariables):
-    data_context: Optional[FileDataContext] = None
+    data_context: FileDataContext = None  # type: ignore[assignment] # validated in __post_init__
```
cdkini: Is there a more appropriate default arg than `None` here?

great_expectations/data_context/data_context_variables.py
```diff
@@ -336,12 +342,41 @@ def _init_store(self) -> DataContextStore:
         )
         return store
 
+    def save_config(self) -> Any:
+        """
+        Persist any changes made to variables utilizing the configured Store.
+        """
+        # overridden in order to prevent calling `instantiate_class_from_config` on ZEP objects
+        # parent class does not have access to the `data_context`
```
cdkini, line 351: Minor nit - could we encapsulate this ZEP specific logic into helpers? Does that make sense?

great_expectations/data_context/data_context/serializable_data_context.py
cdkini: Nevermind read the rest of the PR 😬

great_expectations/data_context/data_context_variables.py
```diff
@@ -336,12 +342,41 @@ def _init_store(self) -> DataContextStore:
...
+
+        return save_result
+
 
 @dataclass(repr=False)
 class CloudDataContextVariables(DataContextVariables):
-    ge_cloud_base_url: Optional[str] = None
-    ge_cloud_organization_id: Optional[str] = None
-    ge_cloud_access_token: Optional[str] = None
+    ge_cloud_base_url: str = None  # type: ignore[assignment] # validated in __post_init__
```
cdkini: Same note for these `type: ignores`

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -201,6 +202,7 @@ def __init__(self, runtime_environment: Optional[dict] = None) -> None:
             runtime_environment (dict): a dictionary of config variables that
                 override those set in config_variables.yml and the environment
         """
+        self.__xdatasource_keys: Set[str] = set()
```
cdkini: Do we want to more widely use `__` prefixes? I know they're different than `_`

great_expectations/data_context/data_context/serializable_data_context.py
Kilo59: I don't follow.

great_expectations/data_context/data_context/abstract_data_context.py
billdirks: I think we should actually just use a single underscore. The Google style guide, which we've adopted, seems to agree (see 3rd bullet):
https://google.github.io/styleguide/pyguide.html#3162-naming-conventions

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3872,3 +3881,13 @@ def _attach_zep_config_datasources(self, config: GxConfig):
...
             self._attach_datasource_to_context(datasource)
+
+    def _synchronize_zep_datasources(self) -> Dict[str, XDatasource]:
+        """
+        Update `self.zep_config.xdatasources` with any newly added datasources.
+        Should be called before serializing `zep_config`.
+        """
+        xdatasources = self.xdatasources
+        if xdatasources:
+            self.zep_config.xdatasources.update(xdatasources)
```
billdirks, line 3891: Should we file a followup ticket about supporting deletes?

great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -87,6 +87,16 @@ def _save_project_config(self) -> None:
         try:
             with open(config_filepath, "w") as outfile:
                 self.config.to_yaml(outfile)
+
+            zep_datasources = self._synchronize_zep_datasources()
+            if zep_datasources:
+                logger.info(
+                    f"Saving {len(self.zep_config.datasources)} ZEP Datasources to {config_filepath}"
+                )
+                with open(config_filepath, "r+") as io_file:
```
billdirks: Could you change the preceding, `self.config.to_yaml(outfile)` to `yaml_dict = self.config.commented_map()` (see `great_expectations/data_context/types/base.py`) or something similar to avoid writing to disk twice?

great_expectations/data_context/data_context_variables.py
```diff
@@ -336,12 +342,41 @@ def _init_store(self) -> DataContextStore:
...
+    def save_config(self) -> Any:
+        """
+        Persist any changes made to variables utilizing the configured Store.
+        """
+        # overridden in order to prevent calling `instantiate_class_from_config` on ZEP objects
+        # parent class does not have access to the `data_context`
+        config_xdatasources_stash: Dict[
+            str, XDatasource
+        ] = self.data_context._synchronize_zep_datasources()
+        if config_xdatasources_stash:
```
billdirks: I don't think you need this `if config_xdatasources_stash:` guard here and below since, if it is `{}`, the for loop will be a no-op.

great_expectations/data_context/data_context_variables.py
billdirks: I think it's good to keep the setup and teardown in 1 function so one can readily see that they both need to happen and be in sync. If we were going to make a helper and keep the setup/teardown together, it would take a callable that is the middle part we wrap which is just `save_result = super().save_config()` so I'm not sure we get much at the moment. Either way is fine with me though.

great_expectations/experimental/datasources/experimental_base_model.py
```diff
@@ -55,12 +80,55 @@ def yaml(
...
+        )
         yaml.dump(intermediate_json, stream=stream_or_path, **yaml_kwargs)
 
         if isinstance(stream_or_path, pathlib.Path):
             return stream_or_path
         return stream_or_path.getvalue()
 
+    def _json_dict(
+        self,
+        *,
```
billdirks, line 103: Not sure if adding the `*` to `yaml` would be useful or break anything. Seems like requiring named arguments might be good.

great_expectations/data_context/data_context_variables.py
Kilo59: Not that I can think of. The `__post_init__` method is checking for `None`.  Whatever it is we will still have to suppress the mypy warning. It won't be a `FileDataContext`.

I could use some magic-string or sentinel value, but IMO that doesn't feel obviously better. 
```
NOT_SET = object()

data_context: FileDataContext = NOT_SET
```

I could just alias `None` so it's more expressive 🤷 

```
NOT_SET = None

data_context: FileDataContext = NOT_SET
```

tests/experimental/datasources/test_viral_snippets.py
```diff
@@ -157,10 +158,68 @@ def test_zep_simple_validate_workflow(zep_file_context: FileDataContext):
...
+
+    context: FileDataContext = get_context(
+        context_root_dir=config_file.parent, cloud_mode=False
+    )
+
+    context.zep_config = zep_only_config
+    context._save_project_config()
+
+    final_yaml = config_file.read_text()
+    diff = difflib.ndiff(initial_yaml.splitlines(), final_yaml.splitlines())
```
billdirks, line 189: I'm not sure generating this `diff` is worthwhile if it can't be asserted on. I think your assert below is sufficient. You could also turn the yaml into a dict and assert keys there.

billdirks (commented):
Thanks for the PR! Some small comments inline.

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59: Works, for me. I'll change it to a `sunder` method 😄

great_expectations/data_context/data_context_variables.py
Kilo59: Good catch

great_expectations/data_context/data_context_variables.py
Kilo59: On second look, the comment itself could be better. 
> `validated in __post_init__`

is a bit unclear...

Maybe... ??
>  `__post_init__ ensures field always set`

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59: The other thing I could do right here, is get rid of `__xdatasource_keys` and make the `xdatasources` just do an `isinstance()` check on all the data sources.
Not very performant 😆 , but it would always be correct and handles deleted datasources.

```python
  @property
    def xdatasources(self) -> Dict[str, XDatasource]:
        return {
            k: v for k in self.datasources.items() if isinstance(v, XDatasource)
        }
```

great_expectations/data_context/data_context/serializable_data_context.py
Kilo59: Yes, not sure why I didn't do that before 😅

great_expectations/experimental/datasources/experimental_base_model.py
Kilo59: I like letting the first argument (`stream_or_path`) be positional. I can try to make a better case then "I like it" if you disagree 😆 .

I agree with making the rest keyword only.

great_expectations/data_context/data_context_variables.py
Kilo59: I am somewhat inspired to do this as a context manager.

```python
def save_config():
   with self.stash_zep_objects():
         save_result = super().save_config()
   return save_result
```

But I also like being able to easily see exactly what actions the workaround is doing without another layer of abstraction, especially given that this is meant to be temporary until we can get rid of `instantiate_class_from_config`. 🤷

billdirks: I like the context manager approach! If it's only in 1 place, I think this is fine too.

great_expectations/data_context/data_context/abstract_data_context.py
billdirks: That sounds simpler. There isn't a huge number of datasources compared to actual data so I'm not concerned with performance here.

great_expectations/experimental/datasources/experimental_base_model.py
billdirks: Sounds good me! I'm also fine with leaving this as is.

tests/experimental/datasources/test_viral_snippets.py
Kilo59: I had left it in to help debuging in the event the test fails.
But I can remove it.

great_expectations/data_context/data_context_variables.py
Kilo59: Done & pushed.

great_expectations/data_context/data_context_variables.py
Kilo59: The loop would be a no-op but there's 2 other steps that need the conditional.

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59: Got rid of the whole thing.

great_expectations/data_context/data_context_variables.py
Kilo59: I've pushed the better comment changes.

@cdkini Also, on further consideration I think I do have a better default arg. The `EllipsisType` `...`

```python
data_context: FileDataContext = ...  # type: ignore[assignment]
```

This also just happens to be a way of formally defining required pydantic `Fields`
https://docs.pydantic.dev/usage/schema/#field-customization
https://docs.python.org/3/library/types.html#types.EllipsisType

However, before I implement this I would like to float it in arch review + submit a formal proposal for this to our style-guide repo.

great_expectations/data_context/data_context/serializable_data_context.py
Kilo59: Now I remember.
I did initially just try to update to `self.config.comment_map` with the zep details and keep everything else as is but it didn't work.
I couldn't find anything in the `ruamel` docs on `CommentMap` to understand why updating it with keys from a regular dictionary doesn't work. I hadn't taken the time to look at the source code.

Because of this I added an `else` branch that retains the old behavior if there are no zep objects.
```python
               if zep_datasources:
                   ...  # do new stuff
                else:
                    self.config.to_yaml(outfile) # original method 
```


If there is some artifact of this change it should be limited to only the zep stuff and I can revisit this if I discover any issues.

great_expectations/data_context/data_context_variables.py
cdkini: sounds good to me, will defer to your judgement here

great_expectations/data_context/data_context_variables.py
```diff
@@ -359,24 +376,24 @@ def save_config(self) -> Any:
...
-        if config_xdatasources_stash:  # config_xdatasources is the superset
-            logger.info(
-                f"Replacing {len(config_xdatasources_stash)} stashed `XDatasource`s"
-            )
-            self.data_context.datasources.update(config_xdatasources_stash)
-            self.data_context.zep_config.xdatasources = config_xdatasources_stash
-
-        return save_result
+        try:
+            yield
```
billdirks, line 379: You might want to pull the preceding `if` block into the `try` because it is modifying data structures that you want to restore on exit of the context manager. That is, if you've popped some datasources and then an error occurs in the above code block, we won't ever get the to finally and the popped datasources won't be restored. Likely we'll just crash and it will be ok but there could be an enclosing try block that makes things proceed.

billdirks (approved):
One small change request. Otherwise, LGTM, thank you!

great_expectations/data_context/data_context_variables.py
Kilo59: Done

PR #6723, opened 2023-01-06
[MAINTENANCE] Refactor tests to leverage `get_context` instead of `DataContext`

cdkini:
- Change `DataContext` references to `get_context` to fit our new desired path for context instantiation

### Definition of Done


Kilo59 (approved):
non-blocking but I'd like to see more of the contexts annotated if they are something other than `AbstractDataContext`.

I don't think there's really much value in the `context: AbstractDataContext = get_context()` annotation.
But `context: File/Emphemeral/CloudDataContext = get_context()` is useful.

PR #6728, opened 2023-01-07
Add docstring for `AbstractDataContext.add_checkpoint`

abegong:
Substantive notes on the API itself:
* Some of the types in the API look like they might be incorrect. For example, shouldn't `batches` accept `List[Batch]` in addition to (maybe instead of) `List[dict]`?
* The staticmethod `Checkpoint.construct_from_config_args` looks redundant with this method to me---its existence is turning this method into a code smell. My understanding is that we want the DataContext to handle translation of loosely typed arguments into strongly typed arguments for components like Checkpoint. I recommend deprecating `Checkpoint.construct_from_config_args` and putting that logic here. This looks like it'd be a quick fix.
* The custom arguments for `LegacyCheckpoint` and `SimpleCheckpoint` concern me. Implicitly, we seem to be making this API the superset of all possible configuration APIs for subclasses of Checkpoint. That seems wrong; at some point, I think we need to revisit how we configure Checkpoints.


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1249,7 +1250,38 @@ def add_checkpoint(
...
+            run_name_template: The run name template to use in generating this checkpoint.
+            expectation_suite_name: The expectation suite name to use in generating this checkpoint.
+            batch_request: The batch request to use in generating this checkpoint.
+            action_list: The action list to use in generating this checkpoint.
+            evaluation_parameters: The evaluation parameters to use in generating this checkpoint.
+            runtime_configuration: The runtime configuration to use in generating this checkpoint.
+            validations: The validations to use in generating this checkpoint.
+            profilers: The profilers to use in generating this checkpoint.
+            validation_operator_name: The validation operator name to use in generating this checkpoint. This is only used for LegacyCheckpoint configuration.
+            batches: The batches to use in generating this checkpoint. This is only used for LegacyCheckpoint configuration.
```
anthonyburdi, line 1309: Does a deprecation warning make sense here for `validation_operator_name` and `batches`?

anthonyburdi:
> Substantive notes on the API itself:
> 
> * Some of the types in the API look like they might be incorrect. For example, shouldn't `batches` accept `List[Batch]` in addition to (maybe instead of) `List[dict]`?
> * The staticmethod `Checkpoint.construct_from_config_args` looks redundant with this method to me---its existence is turning this method into a code smell. My understanding is that we want the DataContext to handle translation of loosely typed arguments into strongly typed arguments for components like Checkpoint. I recommend deprecating `Checkpoint.construct_from_config_args` and putting that logic here. This looks like it'd be a quick fix.
> * The custom arguments for `LegacyCheckpoint` and `SimpleCheckpoint` concern me. Implicitly, we seem to be making this API the superset of all possible configuration APIs for subclasses of Checkpoint. That seems wrong; at some point, I think we need to revisit how we configure Checkpoints.

🙇 Thanks for these notes.
1. Agree on types. I think we can be more permissive here. I made two edits so far.
2. I think we wanted to go the other way and move more logic out of the Data Context, making it mainly a facade class with our public API, calling out to each of its subcomponents. E.g. similar to `test_yaml_config()` and `_YamlConfigValidator`. The logic to translate loosely typed arguments into a Checkpoint would then be the responsibility of a class or method related to the Checkpoint like `_YamlConfigValidator`. Then the subcomponents can be tested independently from one another, and Data Context tests are simpler.
3. That is true. We could instead split this into multiple methods, `add_checkpoint()`, `add_legacy_checkpoint()` and `add_simple_checkpoint()`. There are probably other options as well.

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: I think so - just added.

anthonyburdi (approved):
Thank you for these changes 🙇

PR #6731, opened 2023-01-08
[MAINTENANCE] `core/expectation_` type checking 

Kilo59:
- bring code under type-checking coverage
  - `core/expectation_*` modules
  - `expectations/registry.py`
- Define `KWargDetailsDict` `TypedDict`
  - Update `KWargDetailsDict` `domain_kwargs` and `success_kwargs` to be tuples instead of lists
- Define `ExpectationDiagnosticCheckMessageDict` `TypedDict` 

### Definition of Done


great_expectations/expectations/registry.py
```diff
@@ -355,10 +355,11 @@ def get_expectation_impl(expectation_name: str):
         )
         expectation_name = renamed[expectation_name]
 
-    if expectation_name not in _registered_expectations:
+    expectation = _registered_expectations.get(expectation_name)
+    if not expectation:
         raise ge_exceptions.ExpectationNotFoundError(f"{expectation_name} not found")
 
-    return _registered_expectations.get(expectation_name)
+    return expectation
```
Kilo59, line 362: `expectation_class`

great_expectations/core/expectation_configuration.py
```diff
@@ -93,13 +105,19 @@ def make_expectation_context(self, data, **kwargs):
         return ExpectationContext(**data)
 
 
+class KWargDetailsDict(TypedDict):
+    domain_kwargs: tuple[str, ...]
+    success_kwargs: tuple[str, ...]
```
Kilo59, line 110: This is the new `KWargDetailsDict` `TypedDict`

```python
class KWargDetailsDict(TypedDict):
    domain_kwargs: tuple[str, ...]
    success_kwargs: tuple[str, ...]
    default_kwarg_values: dict[str, str | bool | float | RuleBasedProfilerConfig | None]
```


-----

AFIAK these `(domain_kwargs, success_kwargs)` are never meant to be mutated, so by making them `tuples` instead of lists we assure they are not unintentally modified.
A small side benefit is we reduce the size of this `kwarg_lookup_dict` by about `15%` `61,016 -> 51,496 bytes`.

great_expectations/core/expectation_configuration.py
```diff
@@ -1187,6 +1206,7 @@ def get_runtime_kwargs(self, runtime_configuration: Optional[dict] = None) -> di
         expectation_kwargs_dict = self.kwarg_lookup_dict.get(
             self.expectation_type, None
         )
+        runtime_keys: list[str] | tuple[str, ...]
```
Kilo59: this might only be `tuple[str, ...]`

great_expectations/core/expectation_validation_result.py
```diff
@@ -360,14 +355,14 @@ def get_metric(self, metric_name, **kwargs):
 class ExpectationValidationResultSchema(Schema):
     success = fields.Bool(required=False, allow_none=True)
     expectation_config = fields.Nested(
-        lambda: ExpectationConfigurationSchema, required=False, allow_none=True
+        lambda: ExpectationConfigurationSchema, required=False, allow_none=True  # type: ignore[arg-type,return-value]
```
Kilo59, line 358: It seems like the Marshmallow annotations for `fields.Nested` are just wrong.
I think `fields.Nested()` should at minimum accept a `Callable[..., Schema]` that returns a mm `Schema`. 🤷

great_expectations/core/expectation_configuration.py
```diff
@@ -1327,15 +1351,15 @@ def to_json_dict(self) -> dict:
         return myself
 
     def get_evaluation_parameter_dependencies(self) -> dict:
-        parsed_dependencies = {}
+        parsed_dependencies: dict = {}
```
NathanFarmer, line 1354: Doesn't `{}` imply `dict`?

great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
```diff
@@ -27,15 +32,18 @@
...
+    import black  # noqa: F401
 except ImportError:
-    black = None
+    BLACK_IMPORTED = False  # type: ignore[misc] # Final assignment
 
 try:
     import isort
 except ImportError:
-    isort = None
+    ISORT_IMPORTED = False  # type: ignore[misc] # Final assignment
```
NathanFarmer, line 46: I like this. Is there any reason why we shouldn't follow this pattern throughout the codebase?

great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
```diff
@@ -242,7 +250,7 @@ def _check_core_logic_for_all_applicable_execution_engines(
     ) -> ExpectationDiagnosticCheckMessage:
         """Check whether core logic for this Expectation exists and passes tests on all applicable Execution Engines"""
 
-        sub_messages = []
+        sub_messages: list[ExpectationDiagnosticCheckMessageDict] = []
```
NathanFarmer, line 253: Does `list` work (as opposed to `typing.List`) because of `from __future__ import __annotations__`?

great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
```diff
@@ -368,7 +376,7 @@ def _check_input_validation(
     ) -> ExpectationDiagnosticCheckMessage:
         """Check that the validate_configuration exists and doesn't raise a config error"""
         passed = False
-        sub_messages = []
+        sub_messages: list[ExpectationDiagnosticCheckMessageDict] = []
```
NathanFarmer, line 379: Again with `list` over `List`

great_expectations/core/expectation_validation_result.py
```diff
@@ -438,7 +433,7 @@ def __init__(
             meta
         )  # We require meta information to be serializable.
         self.meta = meta
-        self._metrics = {}
+        self._metrics: dict = {}
```
NathanFarmer, line 436: Again, thought `{}` implied `dict`.

great_expectations/core/expectation_configuration.py
```diff
@@ -1259,24 +1283,24 @@ def isEquivalentTo(
         if match_type == "domain":
             return all(
                 (
-                    self.expectation_type == other.expectation_type,
-                    self.get_domain_kwargs() == other.get_domain_kwargs(),
+                    self.expectation_type == other.expectation_type,  # type: ignore[union-attr] # could be dict
+                    self.get_domain_kwargs() == other.get_domain_kwargs(),  # type: ignore[union-attr] # could be dict
```
NathanFarmer, line 1287: Of other could be a dict why not use `other.expectation_type or other.get("expectation_type")`?

NathanFarmer (approved):
A few non-blocking questions/comments from me.

great_expectations/core/expectation_configuration.py
Kilo59: Yes, but `mypy` wants you to define what kind of `dict`.
I'm sure the intention is for me to say what kind of values the dict contains... but I don't know/was being lazy.

`dict == dict[Any, Any]`

great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
Kilo59: Thanks, I don't think so. This is the pattern I'm more familiar with.
I think I originally encountered this style in the `pymongo/motor` codebases.

great_expectations/core/expectation_diagnostics/expectation_diagnostics.py
Kilo59: Yes, and the alternative union syntax works with `from __future__ import __annotations__` as well.

great_expectations/core/expectation_configuration.py
Kilo59: A benefit of using this type-dict is we don't have to deal with the union of all possible values for every key.

If this was just a dict we would have to narrow the union of all possible types for every field single key/field.
But because of the `TypeDict` we know that `domain_kwargs` & `success_kwargs` are always tuples, and `default_kwarg_values` are always a dict.

great_expectations/core/expectation_configuration.py
```diff
@@ -93,13 +105,19 @@ def make_expectation_context(self, data, **kwargs):
...
+    domain_kwargs: tuple[str, ...]
+    success_kwargs: tuple[str, ...]
+    default_kwarg_values: dict[str, str | bool | float | RuleBasedProfilerConfig | None]
+
+
 class ExpectationConfiguration(SerializableDictDot):
     """ExpectationConfiguration defines the parameters and name of a specific expectation."""
 
-    kwarg_lookup_dict = {
+    kwarg_lookup_dict: ClassVar[Mapping[str, KWargDetailsDict]] = {
```
Kilo59, line 117: I was expecting making the tuple changes below to be a nightmare because missing one tuple change would cause failures.

Thankfully mypy was able to point to the specific lines of each specific nested dict instead of an unhelpful error on line `117` that the whole object was wrong.

great_expectations/core/expectation_configuration.py
Kilo59: I'd prefer to save logic changes for another PR. Also I don't totally understand if this is the same logic 😅

great_expectations/expectations/registry.py
```diff
@@ -355,10 +355,13 @@ def get_expectation_impl(expectation_name: str):
...
 
-    if expectation_name not in _registered_expectations:
+    expectation: Type[Expectation] | None = _registered_expectations.get(
+        expectation_name
+    )
+    if not expectation:
         raise ge_exceptions.ExpectationNotFoundError(f"{expectation_name} not found")
 
-    return _registered_expectations.get(expectation_name)
+    return expectation
```
Kilo59, line 364: Here I am inverting this check to make the type-checker happy but the result is the same.

PR #6745, opened 2023-01-10
[FEATURE] F/great 1313/zep pandas poc

billdirks:
This adds a new experimental pandas datasource. There are also some cleanup that occurred when I touched various files.


great_expectations/experimental/datasources/interfaces.py
```diff
@@ -65,21 +75,103 @@ def datasource(self) -> Datasource:
         return self._datasource
 
     # TODO (kilo): remove setter and add custom init for DataAsset to inject datasource in constructor??
+    # This setter is non-functional: https://github.com/pydantic/pydantic/issues/3395
```
billdirks, line 78: This setter is defined but it doesn't work because of the linked bug.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -53,7 +53,17 @@ class BatchRequest:
     options: BatchRequestOptions
 
 
+class BatchRequestError(Exception):
+    pass
+
+
 class DataAsset(ExperimentalBaseModel):
+    # To subclass a DataAsset one must define `type` as a Class literal explicitly on the sublass
```
billdirks, line 61: I've added a description about the sets required to subclass a `DataAsset`.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -65,21 +75,103 @@ def datasource(self) -> Datasource:
...
+        raise NotImplementedError
+
+    def get_batch_list_from_batch_request(
+        self, batch_request: BatchRequest
+    ) -> List[Batch]:
         raise NotImplementedError
 
+    # End Abstract Methods
 
+    def get_batch_request(
```
billdirks, line 104: The methods here use to live in `TableAsset` but were generic enough that I pulled them into `DataAsset`.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -65,21 +75,103 @@ def datasource(self) -> Datasource:
...
+
+
+# If a Datasource can have more than 1 DataAssetType, this will need to change.
 DataAssetType = TypeVar("DataAssetType", bound=DataAsset)
 
 
 class Datasource(
     ExperimentalBaseModel, Generic[DataAssetType], metaclass=MetaDatasource
 ):
+    # To subclass Datasource one needs to define:
```
billdirks, line 165: I added a description of the steps required to subclass `Datasource`.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -168,6 +254,26 @@ def get_asset(self, asset_name: str) -> DataAssetType:
                 f"'{asset_name}' not found. Available assets are {list(self.assets.keys())}"
             ) from exc
 
+    def add_asset(self, asset: DataAssetType) -> DataAssetType:
```
billdirks, line 257: This was pulled up from the `SQLDatasource`.

great_expectations/experimental/datasources/postgres_datasource.py
```diff
@@ -9,7 +9,8 @@ class PostgresDatasource(SQLDatasource):
 
     Args:
         name: The name of this postgres datasource
-        connection_str: The SQLAlchemy connection string used to connect to the postgres database.
+        connection_string: The SQLAlchemy connection string used to connect to the postgres
```
billdirks, line 12: To line this up with the actual variable names `conntection_str` got changed to `connection_string`.

great_expectations/experimental/datasources/sources.py
```diff
@@ -9,7 +9,7 @@
 
 if TYPE_CHECKING:
     from great_expectations.data_context import AbstractDataContext as GXDataContext
-    from great_expectations.experimental.datasources.context import DataContext
+    from great_expectations.experimental.context import DataContext
```
billdirks, line 12: This import path was broken.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -189,59 +183,6 @@ def _parse_order_by_sorter(
             return _batch_sorter_from_str(v)
         return v
 
-    def get_batch_request(
```
billdirks, line 192: These deleted method were pulled up to the parent class.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -470,28 +411,4 @@ def add_table_asset(
             order_by=order_by or [],  # type: ignore[arg-type]  # coerce list[str]
             # see TableAsset._parse_order_by_sorter()
         )
-        # TODO (kilo59): custom init for `DataAsset` to accept datasource in constructor?
-        # Will most DataAssets require a `Datasource` attribute?
-        asset._datasource = self
-        self.assets[name] = asset
-        return asset
-
-    def get_asset(self, asset_name: str) -> TableAsset:
```
billdirks, line 479: These methods were pulled up into the parent class.

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -9,13 +9,13 @@
 from great_expectations.execution_engine import SqlAlchemyExecutionEngine
 from great_expectations.experimental.datasources.interfaces import (
     BatchRequest,
+    BatchRequestError,
```
billdirks, line 12: `BatchRequestError` was pulled up into the parent class so the import had to change.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -65,21 +75,103 @@ def datasource(self) -> Datasource:
...
+
+        Returns:
+            A BatchRequest object that can be used to obtain a batch list from a Datasource by calling the
+            get_batch_list_from_batch_request method.
+        """
+        if options is not None and not self._valid_batch_request_options(options):
+            raise BatchRequestError(
+                "Batch request options should have a subset of keys:\n"
+                f"{list(self.batch_request_options_template().keys())}\n"
+                f"but actually has the form:\n{pf(options)}\n"
```
NathanFarmer, line 122: What is this `pf`?

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -65,21 +75,103 @@ def datasource(self) -> Datasource:
...
+            )
+        return BatchRequest(
+            datasource_name=self._datasource.name,
+            data_asset_name=self.name,
+            options=options or {},
+        )
+
+    def _valid_batch_request_options(self, options: BatchRequestOptions) -> bool:
```
NathanFarmer, line 130: Can we use a [root validator](https://docs.pydantic.dev/usage/validators/#root-validators) here?

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -65,21 +75,103 @@ def datasource(self) -> Datasource:
...
+        return set(options.keys()).issubset(
+            set(self.batch_request_options_template().keys())
+        )
+
+    def _validate_batch_request(self, batch_request: BatchRequest) -> None:
```
NathanFarmer, line 135: Can we use a [root validator](https://docs.pydantic.dev/usage/validators/#root-validators) here?

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -168,6 +254,27 @@ def get_asset(self, asset_name: str) -> DataAssetType:
...
+            asset: The DataAsset to be added to this datasource.
+        """
+        # The setter for datasource is non-functional so we access _datasource directly.
+        # See the comment in DataAsset for more information.
+        asset._datasource = self
+        self.assets[asset.name] = asset
+        return asset
+
+    # Abstract Methods
+    def execution_engine_type(self) -> Type[ExecutionEngine]:
```
NathanFarmer, line 270: Does this get implemented as a property?

It would also seem like `execution_engine_type` could just be a field on this pydantic model and this method could be a [validator](https://docs.pydantic.dev/usage/validators/).

great_expectations/experimental/datasources/interfaces.py
billdirks: See the imports at the top of the file. `pf` is an alias for `pprint.pformat`: `from pprint import pformat as pf`. `pformat` is like `pprint.pprint` except it returns a string instead of writing to a stream.

great_expectations/experimental/datasources/interfaces.py
billdirks: I'm not sure what you mean. I've validating the passed in `options` argument and not part of this model. In general `BatchRequestOptions` don't know which `DataAsset` they are going to be passed to so I can't do it there. Maybe I'm misunderstanding something. Let's chat tomorrow.

great_expectations/experimental/datasources/interfaces.py
billdirks: Oh, this is a hack to avoid circular dependencies. I don't recall the specifics but we run into this issue because of the relationship between `ExecutionEngine`'s and the `Batch*` types from outside of `Experimental`. When we try to import both, we run into issues. By allowing this to be a function we can delay importing `ExecutionEngine`. That is the implementations look like:

```
    def execution_engine_type(self) -> Type[ExecutionEngine]:
        """Return the PandasExecutionEngine unless the override is set"""
        from great_expectations.execution_engine.pandas_execution_engine import (
            PandasExecutionEngine,
        )

        return PandasExecutionEngine
```

If you think this is better as a property, I could change it to that instead of a function. I could do that here but would prefer to do it in an immediate followup PR to keep the scope of this one from getting bigger.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: No, this makes perfect sense to me!

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: A property would be nice, but that might even go away eventually if we are able to handle the circular imports in the future.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -0,0 +1,204 @@
...
+    path: pathlib.Path
+    regex: Pattern
+
+    # Internal attrs
+    _unnamed_regex_param_prefix: str = pydantic.PrivateAttr(
+        default="batch_request_param_"
+    )
+
+    def _fully_specified_batch_requests_with_path(
+        self, batch_request
```
NathanFarmer: Can we annotate this `batch_request` type?

NathanFarmer (approved):

Kilo59 (approved):

great_expectations/experimental/datasources/pandas_datasource.py
Kilo59: Not sure if this is the place to do it, but we could start using some "thin-interfaces/Protocols" here.

great_expectations/experimental/datasources/interfaces.py
billdirks: 👍 , I'll do that as a followup.

great_expectations/experimental/datasources/pandas_datasource.py
billdirks: Annotated as a `BatchRequest` for now.

PR #6747, opened 2023-01-10
[MAINTENANCE] Unset cloud vars when running pytest from a mac

kenwade4:


tests/__init__.py
```diff
@@ -0,0 +1,11 @@
+import os
```
alexsherstinsky, line 1: @cdkini @anthonyburdi @billdirks @Kilo59 -- Something tells me that OSS should not have any mentions of `GE_CLOUD` or other non-OSS type branding (similar sentiment as not including internal JIRA ticket links in the OSS code), but wanted to know what you think.  Thanks. /cc @Shinnnyshinshin

tests/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+import sys
+
+if sys.platform == "darwin":
+    print(
+        "Running tests from a dev machine, so unsetting any GE_CLOUD_* environment vars"
+    )
+    cloud_vars = [k for k in os.environ if k.startswith("GE_CLOUD")]
+    for var_name in cloud_vars:
+        print(f"Unsetting environment variable {var_name}")
+        del os.environ[var_name]
```
cdkini: I think we should probably discuss this a bit more before implementing.

We've got both `GE_` and `GX_` env vars to consider

cdkini (changes requested):

tests/__init__.py
```diff
@@ -0,0 +1,12 @@
+import os
+import sys
+
+if sys.platform == "darwin":
+    cloud_vars = [k for k in os.environ if k[:8] in ("GX_CLOUD", "GE_CLOUD")]
```
cdkini: Could we instead use the `GXCloudEnvironmentVariable` enum as a check here?

tests/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+    if cloud_vars:
+        print(
+            "Running tests from a dev machine, so unsetting any CLOUD environment vars"
+        )
+        for var_name in cloud_vars:
+            print(f"Unsetting environment variable {var_name}")
+            del os.environ[var_name]
```
cdkini:
```suggestion
if sys.platform == "darwin":
    for var in GXCloudEnvironmentVariable:
        if var in os.environ:
            logger.info(...) # or print
            os.environ.pop(var)
```

cdkini (approved):

PR #6759, opened 2023-01-11
[BUGFIX] Fix observed value

itaise:
Hi, 
here is a fix for the observed value - after the change to the metric "query.template_values" by @alexsherstinsky the returned value from the metric is not SQL Alchemy row anymore, so no need to serialize it. 

Thanks!


alexsherstinsky (approved):
LGTM -- @itaise -- thank you so much for finding this bug and fixing it!

PR #6765, opened 2023-01-11
[FEATURE] Experimental datasources `batch.head()`

NathanFarmer:
- GREAT-1404
- Add `head()` functionality to `Batch` class
- Make `Batch` an `ExperimentalModel`
- Align behavior for SQL Alchemy and Spark `table.head` metrics for negative `n_rows` which is currently supported by Pandas

### Definition of Done


ghost:
<details>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=6765&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

    [Try it now!](https://app.codesee.io/r/reviews?pr=6765&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)
</details>

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=6765&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map Legend

great_expectations/render/renderer/inline_renderer.py
```diff
@@ -61,7 +61,7 @@ def _get_atomic_rendered_content_for_object(
             A list of RenderedAtomicContent objects for a given ExpectationConfiguration or ExpectationValidationResult.
         """
         expectation_type: str
-        renderer_typs: List[AtomicRendererType]
+        renderer_types: List[AtomicRendererType]
```
NathanFarmer, line 64: Unrelated typo

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -371,94 +371,68 @@ def execution_engine_type(self) -> Type[ExecutionEngine]:
...
+    metadata: dict[str, Any] = {}
 
     # TODO: These legacy fields are currently required. They are only used in usage stats so we
     #       should figure out a better way to anonymize and delete them.
-    _legacy_batch_markers: BatchMarkers
-    _legacy_batch_spec: BatchSpec
-    _legacy_batch_definition: BatchDefinition
+    batch_markers: BatchKwargs  # Due to circular imports we can't use BatchMarkers
+    batch_spec: BatchSpec
+    batch_definition: SerializableDictDot  # Due to circular imports we can't use BatchDefinition
```
NathanFarmer, line 394: Do want to make these private so they aren't serialized?

great_expectations/expectations/metrics/table_metrics/table_head.py
```diff
@@ -91,14 +91,21 @@ def _sqlalchemy(
                         con=execution_engine.engine,
                     )
                 else:
-                    df = next(
-                        pd.read_sql_table(
-                            table_name=getattr(selectable, "name", None),
-                            schema=getattr(selectable, "schema", None),
-                            con=execution_engine.engine,
-                            chunksize=metric_value_kwargs["n_rows"],
-                        )
```
NathanFarmer, line 100: This was a SQL Alchemy bug; when `n_rows` `chunksize` was negative the entire dataframe was returned instead of an `Iterator` of partitions.

tests/experimental/datasources/test_integration_datasource.py
```diff
@@ -336,3 +347,102 @@ def _configure_and_run_data_assistant(
...
+        (
+            0,
+            True,
+        ),
+    ],
+)
+def test_batch_head(
+    datasource_test_data: tuple[
+        AbstractDataContext, Datasource, DataAsset, BatchRequest
+    ],
```
NathanFarmer, line 382: Sharing parametrized fixture to ensure Spark is also tested.

great_expectations/core/batch_manager.py
```diff
@@ -144,10 +142,7 @@ def reset_batch_cache(self) -> None:
         self._batch_cache = OrderedDict()
         self._active_batch_id = None
 
-    def load_batch_list(self, batch_list: Optional[List[Batch]]) -> None:
-        if batch_list is None:
-            batch_list = []
-
+    def load_batch_list(self, batch_list: List[Batch]) -> None:
```
cdkini, line 145: Are there are repercussions from removing the default value here?

great_expectations/expectations/metrics/table_metrics/table_head.py
```diff
@@ -29,29 +34,40 @@ def _pandas(
...
             metric_domain_kwargs, domain_type=MetricDomainTypes.TABLE
         )
         if metric_value_kwargs.get("fetch_all", cls.default_kwarg_values["fetch_all"]):
             return df
-        return df.head(metric_value_kwargs["n_rows"])
+        n_rows: int = (
+            metric_value_kwargs.get("n_rows")
+            if metric_value_kwargs.get("n_rows") is not None
+            else cls.default_kwarg_values["n_rows"]
+        )
```
cdkini, line 49:
```suggestion
        n_rows: int = (
            metric_value_kwargs.get("n_rows"), cls.default_kwarg_values["n_rows"])
        )
```
This work?

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -371,94 +380,88 @@ def execution_engine_type(self) -> Type[ExecutionEngine]:
...
+    """This represents a batch of data.
+
+    This is usually not the data itself but a hook to the data on an external datastore such as
+    a spark or a sql database. An exception exists for pandas or any in-memory datastore.
+    """
+
+    datasource: Datasource
+    data_asset: DataAsset
+    batch_request: BatchRequest
+    data: "BatchData"
```
cdkini, line 434: Could we type annotate this without quotes since we use `__future__`?

tests/experimental/datasources/test_integration_datasource.py
```diff
@@ -336,3 +351,148 @@ def _configure_and_run_data_assistant(
...
+        execution_engine.batch_manager.load_batch_list(batch_list=[batch])
+
+        metrics_calculator = MetricsCalculator(execution_engine=execution_engine)
+
+        # define metric for the total number of rows
+        row_count_metric = MetricConfiguration(
+            metric_name="table.row_count",
+            metric_domain_kwargs={"batch_id": batch.id},
+            metric_value_kwargs=None,
+        )
```
cdkini, line 409: I think this is personal preference but if the test has a number of conditionals and branching paths, it may make sense to lose some value from parameterization in favor of easier readability/ease of debugging?

I think it may make sense to break up this test and the grouped asserts based on what we're actually testing but will defer to the core ZEP folks here

great_expectations/expectations/metrics/table_metrics/table_head.py
NathanFarmer: I tried that first, but if `metric_value_kwargs["n_rows"]=None`, it will not use the default.

great_expectations/core/batch_manager.py
NathanFarmer: I just had to update two callers to ensure `batch_list` is not `None`. Those are the only two places this is called.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: No because of circular imports. The quotes avoid the circular import issue and then `pydantic` still wants to validate the actual type, so you have to pass the type to `Batch.update_forward_refs()` in the scope that you are instantiating the `Batch` class (see changes to `pandas_datasource.py` and `sql_datasource.py`.

tests/experimental/datasources/test_integration_datasource.py
NathanFarmer: Right now there are 36 test cases (will be 54 when Spark datasource is added). Almost all of the complexity you see here is because Spark returns something different than Pandas or SQL Alchemy. That would be one way to simplify, but then you lose the ability to ensure the 3 `ExecutionEngine`s are being tested similarly. We could also separate the column name logic from the row count logic if that makes readability better. My opinion is that the comments should give you enough information to debug.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: I should also say using `if TYPE_CHECKING` won't help because `pydantic` actually validates at runtime (as opposed to `mypy`).

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -371,94 +380,88 @@ def execution_engine_type(self) -> Type[ExecutionEngine]:
...
+    def head(
+        self, n_rows: Literal[0], fetch_all: Literal[False]
+    ) -> pd.DataFrame | list[pyspark_sql_Row]:
+        ...
+
+    @overload
+    def head(
+        self, n_rows: Optional[StrictInt] = None, fetch_all: StrictBool = False
+    ) -> pd.DataFrame | list[pyspark_sql_Row] | pyspark_sql_Row:
+        ...
```
Kilo59: How do these overloads help us here?
The return types are the same.

Edit: Oh wait I think I see, we are conveying that if `n_rows` is`0` then `fetch_all` must be false?

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -371,94 +380,88 @@ def execution_engine_type(self) -> Type[ExecutionEngine]:
...
+
+    @overload
+    def head(
+        self, n_rows: Literal[0], fetch_all: Literal[False]
+    ) -> pd.DataFrame | list[pyspark_sql_Row]:
+        ...
+
+    @overload
+    def head(
+        self, n_rows: Optional[StrictInt] = None, fetch_all: StrictBool = False
```
```suggestion
        self, n_rows: Optional[StrictInt] = ..., fetch_all: StrictBool = ...
```

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -371,94 +380,88 @@ def execution_engine_type(self) -> Type[ExecutionEngine]:
...
 
 
-class Batch:
-    # Instance variable declarations
-    _datasource: Datasource
-    _data_asset: DataAsset
-    _batch_request: BatchRequest
-    _data: BatchDataType
-    _id: str
+class Batch(ExperimentalBaseModel):
```
Kilo59, line 424: Why does this need to be an `ExperimentalBaseModel`?
Do we serialize it anywhere?

great_expectations/expectations/metrics/table_metrics/table_head.py
```diff
@@ -114,7 +135,8 @@ def _sqlalchemy(
         if df is None:
             # we want to compile our selectable
             stmt = sa.select(["*"]).select_from(selectable)
-            if metric_value_kwargs["fetch_all"]:
+            fetch_all = metric_value_kwargs["fetch_all"]
```
billdirks, line 138: This function seems to attempt the query in 2 stages. First it tries to query the database via pandas (above this `if` statement on line 135) and if this fails it falls back to using sqlalchemy. I'm curious why we do this. If panda's fails can sqlalchemy succeed? Why try pandas at all?

great_expectations/expectations/metrics/table_metrics/table_head.py
```diff
@@ -29,29 +34,40 @@ def _pandas(
         execution_engine: PandasExecutionEngine,
         metric_domain_kwargs: dict,
         metric_value_kwargs: dict,
-        metrics: Dict[str, Any],
+        metrics: dict[str, Any],
```
billdirks, line 37: Does lowercase `dict` work in python 3.7?

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -15,29 +15,38 @@
...
-        BatchMarkers,
-        BatchSpec,
-    )
+    # TODO: When the Batch class is moved out of the experimental directory, we should try to import the annotations
+    #       from core.batch so we no longer need to call Batch.update_forward_refs() before instantiation.
+    from great_expectations.core.batch import BatchData, BatchDefinition, BatchMarkers
     from great_expectations.execution_engine import ExecutionEngine
+    from great_expectations.validator.computed_metric import MetricValue
+
+try:
```
billdirks, line 43: Can you move this `try` into the `TYPE_CHECKING` block? 

_The rest of the comment is about this pattern of setting the import to None and isn't relevant when we only use the imported values for typing._

I know we use this pattern a bit but I think we should get away from it when we are using values from these maybe imported modules. That is we should define a `MaybeImported` type that wraps this import so we can give better error messages when something like this happens: https://greatexpectationslabs.slack.com/archives/C026YSF2JA3/p1673453243284299

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -371,94 +380,88 @@ def execution_engine_type(self) -> Type[ExecutionEngine]:
...
-        self._legacy_batch_markers = legacy_batch_markers
-        self._legacy_batch_spec = legacy_batch_spec
-        self._legacy_batch_definition = legacy_batch_definition
-
-        # computed property
-        # We need to unique identifier. This will likely change as I get more input
+    class Config:
+        allow_mutation = False
+        arbitrary_types_allowed = True
+        validate_assignment = True
```
billdirks: Do we need `validate_assignment = True` if `allow_mutation = False`? No assignment will be done if mutation isn't allowed, correct?

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -371,94 +380,88 @@ def execution_engine_type(self) -> Type[ExecutionEngine]:
...
-
-        self._legacy_batch_markers = legacy_batch_markers
-        self._legacy_batch_spec = legacy_batch_spec
-        self._legacy_batch_definition = legacy_batch_definition
-
-        # computed property
-        # We need to unique identifier. This will likely change as I get more input
+    class Config:
+        allow_mutation = False
+        arbitrary_types_allowed = True
```
billdirks, line 448: Why do we need to set `arbitrary_types_allowed = True`?

great_expectations/experimental/datasources/interfaces.py
billdirks: I believe if `n_rows == 0 and fetch_all = False` then the return type is `pd.DataFrame | list[pyspark_sql_Row]` otherwise it is `pd.DataFrame | list[pyspark_sql_Row] | pyspark_sql_Row`. Do we do any switching based on this or is it simpler to remove the `@overload`s?

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -258,16 +258,23 @@ def get_batch_list_from_batch_request(
...
-                batch_spec=batch_spec
-            )
 
             # batch_definition (along with batch_spec and markers) is only here to satisfy a
             # legacy constraint when computing usage statistics in a validator. We hope to remove
             # it in the future.
             # imports are done inline to prevent a circular dependency with core/batch.py
             from great_expectations.core import IDDict
-            from great_expectations.core.batch import BatchDefinition
+            from great_expectations.core.batch import (
```
billdirks: Can you make a function like this:
```
def update_batch_forward_refs():
    from great_expectations.core.batch import (
                BatchData,
                BatchDefinition,
                BatchMarkers,
            )
      Batch.update_forward_refs(
                BatchData=BatchData,
                BatchDefinition=BatchDefinition,
                BatchMarkers=BatchMarkers,
            )
```
and put it in a common location so we can call it here and in `pandas_datasource`? If you need some of these types for type checking you can put it in a `TYPE_CHECKING` block.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: The return types are different if you squint 😄

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: We don't serialize it now, although it's possible in the future, since this class does seem to be taking on more importance and we have plans to build a metric store (formerly Batch-Metric Store). The biggest pros in my eyes are immutability and default error messages. If you think this means I should define a custom `BaseModel` instead of using `ExperimentalBaseModel` that could be do-able, but I think this makes sense as-is.

great_expectations/expectations/metrics/table_metrics/table_head.py
NathanFarmer: According to the comment on lines 93 and 124 it fails on temp tables for `pandas < 1.4.0`. We don't fall back to using SQL Alchemy, we just use it to compile the `selectable` before passing it to `pandas` again on lines 163-171 and then temp tables work. We could compile first and just use the `read_sql` convenience method (it calls either `read_sql_query` or `read_sql_table`), but that would be an extra step/less explicit/more points of failure for non-temp tables.

In the end, we need to return a `pd.DataFrame`. Using the Pandas `read_sql` methods outsources the work of converting SQL Alchemy objects to a DataFrame, so we don't need to maintain that logic in the metric.

great_expectations/expectations/metrics/table_metrics/table_head.py
NathanFarmer: My understanding is that yes, if you use `from __future__ import annotations` it works (I have been doing this for the past week or so after @Kilo59 mentioned it).

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: Nice catch! Yes, I have removed `validate_assignment`.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: Because `pydantic` only has out-of-the-box validations for [certain common types](https://docs.pydantic.dev/usage/types/#standard-library-types). This option enables an `isinstance` check for our custom types. More detailed [validations can be written for our types](https://docs.pydantic.dev/usage/types/#custom-data-types) in the future.

[https://docs.pydantic.dev/usage/types/#arbitrary-types-allowed](https://docs.pydantic.dev/usage/types/#arbitrary-types-allowed)

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: Good idea, I will put this in a common location. This won't work in an `if TYPE_CHECKING` block since `pydantic` needs the actual types at runtime.

NathanFarmer:
@billdirks 

> I believe if `n_rows == 0` and `fetch_all = False` then the return type is `pd.DataFrame | list[pyspark_sql_Row]` otherwise it is `pd.DataFrame | list[pyspark_sql_Row] | pyspark_sql_Row`. Do we do any switching based on this or is it simpler to remove the @overloads?

I had other overloads defined, but found them to be inaccurate due to some weird return type behavior with `pyspark` `head()`. I will remove this overload since it no longer seems to be adding much value.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: I had other overloads defined, but found them to be inaccurate due to some weird return type behavior with `pyspark` `head()`. I will remove these overloads since they no longer seem to be adding much value.

NathanFarmer: ✅

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: Using `if TYPE_CHECKING` here won't work because `pydantic` validates (even the return) types at runtime due to the `@validate_arguments` decorator. Designing the new pattern seems like something that we should take to arch review and implement throughout the codebase.

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: I was able to make this an overloaded `classmethod` that imports within the method and then passes the kwargs to `super()`! Definitely something we should do if we need postponed annotations in the future.

Kilo59 (approved):

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -408,6 +406,19 @@ def execution_engine_type(self) -> Type[ExecutionEngine]:
     # End Abstract Methods
 
 
+@dataclasses.dataclass(frozen=True)
+class HeadData:
+    """
+    An immutable wrapper around pd.DataFrame for .head() methods which are intended to be used for visual inspection of
```
billdirks: nit: This line is over 100 chars long. Could you shorten it?

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -477,19 +489,29 @@ def head(
...
+            pyspark_sql_Row
+        ] | pyspark_sql_Row = self.data.execution_engine.resolve_metrics(
+            metrics_to_resolve=(metric,)
+        )[
+            metric.id
+        ]
+        table_head_df: pd.DataFrame
+        if not isinstance(table_head_metric_value, pd.DataFrame):
+            if not isinstance(table_head_metric_value, list):
+                table_head_df = pd.DataFrame(table_head_metric_value.asDict())
```
billdirks: Could you add a comment in each of these conditionals, what says the type `table_head_metric_value`. For example here is it s `pyspark_sql_Row` while in the `else` it's a list of these.

tests/integration/profiling/rule_based_profiler/test_profiler_user_workflows.py
```diff
@@ -619,7 +619,7 @@ def test_bobby_columnar_table_multi_batch_batches_are_accessible(
 
     # noinspection PyUnresolvedReferences
     pickup_datetime: datetime.datetime = pd.to_datetime(
-        validator_latest.head(n_rows=1)["pickup_datetime"][0]
+        validator_latest.head(n_rows=1).data["pickup_datetime"][0]
```
billdirks: `head` is public. Are you concerned about this api change breaking anyone?

billdirks (commented):
This looks good to me. I do have a question about `head` being public and changing the API. If you don't think this is an issue, I'm good with you merging.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: I made it shorter, although right now we state 120 char limit in the style guide (I had to look this up). Something to debate.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: I don't know why those conditionals ended up so convoluted. I simplified and added comments.

tests/integration/profiling/rule_based_profiler/test_profiler_user_workflows.py
NathanFarmer: This is just being used as a convenience for a test. I made the same change to `validator` after consulting product and Alex.

NathanFarmer: I reverted the change after seeing all of the test changes that were required. I will file a ticket for the changes to `Validator.head()`.

PR #6778, opened 2023-01-12
[MAINTENANCE] Refactor `file_relative_path` util

cdkini:
- Use `pathlib`
- Make the method test strict existence

### Definition of Done


great_expectations/data_context/util.py
```diff
@@ -153,9 +156,14 @@ def file_relative_path(dunderfile, relative_path):
     It is meant to be used like the following:
     file_relative_path(__file__, 'path/relative/to/file')
 
+    This has been modified from Dagster's utils:
     H/T https://github.com/dagster-io/dagster/blob/8a250e9619a49e8bff8e9aa7435df89c2d2ea039/python_modules/dagster/dagster/utils/__init__.py#L34
     """
-    return os.path.join(os.path.dirname(dunderfile), relative_path)
+    abs_path = pathlib.Path(source_path).joinpath(relative_path).absolute()
+    str_path = abs_path.as_posix()
```
Kilo59: `non-blocking`
Can we just return a `pathlib.Path` as well? Then this call (`.as_posix()`) is unnecessary.

Non-blocking because this might involve modifying a lot of code that calls this.
I still think we should do it... but not needed right now.

great_expectations/data_context/util.py
```diff
@@ -143,7 +144,9 @@ def format_dict_for_error_message(dict_):
     return "\n\t".join("\t\t".join((str(key), str(dict_[key]))) for key in dict_)
 
 
-def file_relative_path(dunderfile, relative_path):
+def file_relative_path(
+    source_path: str, relative_path: str, strict: bool = True
```
Kilo59: https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/alias_types.py#L8
```python
from great_expectations.alias_types import PathStr
```

```suggestion
    source_path: PathStr, relative_path: PathStr, strict: bool = True
```

Kilo59 (changes requested):
Can you update the arg types to reflect that it can take `Path` objects too?

My other request is `non-blocking`

great_expectations/data_context/util.py
```diff
@@ -2,6 +2,7 @@
 import inspect
 import logging
 import os
```
Kilo59:
```suggestion
```
🎊 good riddance

great_expectations/data_context/util.py
```diff
@@ -153,9 +156,14 @@ def file_relative_path(dunderfile, relative_path):
...
     file_relative_path(__file__, 'path/relative/to/file')
 
+    This has been modified from Dagster's utils:
     H/T https://github.com/dagster-io/dagster/blob/8a250e9619a49e8bff8e9aa7435df89c2d2ea039/python_modules/dagster/dagster/utils/__init__.py#L34
     """
-    return os.path.join(os.path.dirname(dunderfile), relative_path)
+    abs_path = pathlib.Path(source_path).joinpath(relative_path).absolute()
+    str_path = abs_path.as_posix()
+    if strict:
+        assert abs_path.exists(), f"The target relative path {str_path} does not exist!"
```
Kilo59: Might want to call `.resolve(strict=strict)` as the first step. Then I think you can skip this logic.
https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve
```python
abs_path = pathlib.Path(source_path).joinpath(relative_path).resolve(strict=strict)
return abs_path.as_posix()
```
`FileNotFoundError` will be raised if `strict=True` and it doesn't exist.

Kilo59 (approved):

great_expectations/data_context/util.py
Kilo59: Why do we need to call `as_posix()` at all? Why not just `str(abs_path)`?
Does the rest of our filepath operations expect windows paths to be converted to posix paths?

Another issue that will go away if we use `pathlib.Path` objects universally.

great_expectations/data_context/util.py
```diff
@@ -153,9 +158,12 @@ def file_relative_path(dunderfile, relative_path):
     It is meant to be used like the following:
     file_relative_path(__file__, 'path/relative/to/file')
 
+    This has been modified from Dagster's utils:
     H/T https://github.com/dagster-io/dagster/blob/8a250e9619a49e8bff8e9aa7435df89c2d2ea039/python_modules/dagster/dagster/utils/__init__.py#L34
     """
-    return os.path.join(os.path.dirname(dunderfile), relative_path)
+    dir_path = pathlib.Path(source_path).parent
+    abs_path = dir_path.joinpath(relative_path).resolve(strict=strict)
+    return abs_path.as_posix()
```
Kilo59:
```suggestion
    return str(abs_path)
```

great_expectations/data_context/util.py
Kilo59: BTW I think defaulting to `strict=False` is reasonable (for now), and then you shouldn't have to change it's usage at all.

Kilo59:
Why close? It looked like it only failing because a couple of tests were using it to point to files that didn't exist. Which could be resolved by changing the strict default to false or just passing as false for those cases.

PR #6780, opened 2023-01-12
[FEATURE] ZEP - dynamic pandas asset schema definitions

Kilo59:
The goal in this PR is to be able to dynamically create `pandas` `DataAsset` types by inspecting the function signatures of the `pandas.io` read methods (eg `read_csv`, `read_excel`, `read_json` etc.). `20` methods in total.

The names and type annotations from these function signatures will be used to define fields on the corresponding `DataAsset` example: (`CSVAsset`, `ExcelAsset` etc.).

~There are some arguments and types that we will ignore during this initial effort due to difficulties representing the types during serialization.~ Example: `Callable`. We could choose to ignore these only during serialization but then we could get into an awkward state where our `DataAsset` exhibited different behavior the first time we created in code, than it does after we round trip it from config.

If a type has no annotation or otherwise cannot be serialized, in this initial effort we will either give it an `Any` type, or a more specific primitive type. Example: `dtype` -> `dict`

- Generate pandas `DataAsset` types dynamically by inspecting the `pandas.io` read methods
- change existing `CSVAsset` -> `_DataFrameAsset` which all dynamically generated assets inherit from
- [x] plumb additional asset arguments to the `ExecutionEngine`
- [ ] ~evaluate string types to real python types~ - followup
- [ ] put json schemas under source control
- [x] make other asset types available
  - [x] CSVAsset
  - [x] ExcelAsset
  - [x] JSONAsset
  - [x] ParquetAsset
- [x] verify the generated schemas are reasonable

### Definition of Done

## Implementation Notes

`pandas` uses string annotation (from __future__ import annotations) for their io method type annotations.
This means at some point we need to evaluate/resolve these string annotations back into python type objects.
We can...
1. let `pydantic` do this as part of normal model creation/definition. This might require an `.update_forward_references()` call.
2. ~"eval" the string annotations ourselves.~ - (not yet, might be done in follow-up)
    - https://docs.python.org/3/howto/annotations.html#manually-un-stringizing-stringized-annotations
    - [This is how `pydantic` deals with this problem](https://github.com/pydantic/pydantic/blob/v1.10.4/pydantic/typing.py#L375-L404)


great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -0,0 +1,260 @@
...
+        no_annotation: bool = param.annotation is inspect._empty
+        if no_annotation:
+            logger.debug(f"`{param_name}` has no type annotation")
+            FIELD_SKIPPED_NO_ANNOTATION.add(param_name)
+            continue
+        type_ = _get_annotation_type(param)
+        if type_ is UNSUPPORTED_TYPE:
+            logger.debug(f"`{param_name}` has no supported types. Field skipped")
+            FIELD_SKIPPED_UNSUPPORTED_TYPE.add(param_name)
+            continue
```
Kilo59, line 202: TODO: annotate as an Any type and allow

tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -246,7 +302,7 @@ def test_pandas_sorter(
     @dataclass(frozen=True)
     class TimeRange:
         key: str
-        range: List[str]
+        range: list[int]
```
Kilo59:
```suggestion
        range: list[str]
```

tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -1,8 +1,13 @@
-import pathlib
+from __future__ import annotations
+
+import inspect
 from dataclasses import dataclass
-from typing import List
+from pathlib import Path
```
Kilo59:
```suggestion
import pathlib
```

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -201,19 +226,96 @@ def add_csv_asset(
...
+        **kwargs,  # TODO: update signature to have specific keys & types
+    ) -> CSVAsset:  # type: ignore[valid-type]
         """Adds a csv asset to this pandas datasource
 
         Args:
             name: The name of the csv asset
             data_path: Path to directory with csv files
             regex: regex pattern that matches csv filenames that is used to label the batches
-            order_by: one of "asc" (ascending) or "desc" (descending) -- the method by which to sort "Asset" parts.
+            kwargs: Extra keyword arguments should correspond to ``pandas.read_csv`` keyword args
```
Kilo59, line 246:
```suggestion
            order_by: one of "asc" (ascending) or "desc" (descending) -- the method by which to sort "Asset" parts.
            kwargs: Extra keyword arguments should correspond to ``pandas.read_csv`` keyword args
```

tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -148,19 +216,14 @@ def test_get_batch_list_from_fully_specified_batch_request(
 
 @pytest.mark.unit
 def test_get_batch_list_from_partially_specified_batch_request(
-    pandas_datasource: PandasDatasource, csv_path: pathlib.Path
+    pandas_datasource, csv_path: Path
```
Kilo59:
```suggestion
    pandas_datasource: PandasDatasource, csv_path: pathlib.Path
```

tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -31,6 +52,133 @@ def csv_path() -> pathlib.Path:
...
+        )
+
+    monkeypatch.setattr(
+        great_expectations.execution_engine.pandas_execution_engine.PandasExecutionEngine,
+        "_get_reader_fn",
+        lambda *_: reader_fn_spy,
+        raising=True,
+    )
+
+    yield captured_args, captured_kwargs
```
Kilo59: This is probably the most important part of these new tests.

It is currently only used by `test_data_asset_reader_options_passthrough`

tests/experimental/datasources/tasks.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@invoke.task(aliases=["gen-assets"])
+def generate_asset_files(
+    ctx: invoke.Context,
+    source_csv_dir: str | pathlib.Path = None,
+    limit: int = 12,
+    year: str = 2019,
+):
+    """Generate or re-generate various test asset files."""
```
Kilo59, line 22: This is how I generated all the new json files.
We really shouldn't need to run it again unless we want tweak it to generate other pandas formats.

The other thing we could do is just generate these other formats on the fly with some parametrized fixture.
But I didn't want to spend more time going down that rabbit hole.

We discussed before that there's limited value testing all of these various formats because then we are essentially just testing that `pandas` works. 😆 

So TLDR... 
We could do this differently, or not at all.

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -0,0 +1,272 @@
...
+def _to_pydantic_fields(
+    sig_tuple: _SignatureTuple,
+) -> Dict[str, _FieldSpec]:
+    """
+    Extract the parameter details in a structure that can be easily unpacked to
+    `pydantic.create_model()` as field arguments
+    """
+    fields_dict: Dict[str, _FieldSpec] = {}
+    for param_name, param in sig_tuple.signature.parameters.items():
+        # print(type(param), param)
```
Kilo59, line 201:
```suggestion
```

tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -31,6 +52,133 @@ def csv_path() -> pathlib.Path:
...
+
+
+@pytest.mark.unit
+class TestDynamicPandasAssets:
```
Kilo59, line 81: I could also put this in it's own test module

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -177,14 +189,35 @@ def get_batch_list_from_batch_request(
...
+        "read_json",
+        "read_excel",
+        "read_parquet",
+    ),
+)
+
+CSVAsset = _ASSET_MODELS["csv"]
+ExcelAsset = _ASSET_MODELS["excel"]
+JSONAsset = _ASSET_MODELS["json"]
+ParquetAsset = _ASSET_MODELS["parquet"]
```
Kilo59, line 205: We no longer have a model with nicely annotated fields to look at.
One thing we could do is commit the generated json schemas to source control. That way we can catch if they deviate in unexpected ways.

You can use `invoke schema csv|excel|json` etc. to view them. But it's hard to see how things are changing.
Also, it would be trivial to write a tests to make sure the schemas under source control are always in sync with the latest model.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -201,19 +234,97 @@ def add_csv_asset(
         data_path: PathStr,
         regex: Union[str, re.Pattern],
         order_by: Optional[BatchSortersDefinition] = None,
-    ) -> CSVAsset:
+        **kwargs,  # TODO: update signature to have specific keys & types
+    ) -> CSVAsset:  # type: ignore[valid-type]
```
Kilo59, line 238: I have a follow-up ticket for getting rid of the kwargs

Kilo59:
Note there are some failing integration tests I need to take a look at.
All the existing and new experimental tests pass.

tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -31,6 +52,133 @@ def csv_path() -> pathlib.Path:
...
+    captured_args: list[list] = []
+    captured_kwargs: list[dict[str, Any]] = []
+
+    def reader_fn_spy(*args, **kwargs):
+        logging.info(f"reader_fn_spy() called with...\n{args}\n{kwargs}")
+        captured_args.append(args)
+        captured_kwargs.append(kwargs)
+        raise SpyInterrupt(
+            "Reader options have been captured", {"args": args, "kwargs": kwargs}
+        )
```
Kilo59: I actually don't need to do `SpyInterrupt` this capture AND use the mutable `captured_kwargs` containers.
I could do one or the other.

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -318,6 +323,7 @@ def get_batch_data_and_markers(  # noqa: C901 - 22
             buf.seek(0)
             df = reader_fn(buf, **reader_options)
 
+        # ZEP code path
         elif isinstance(batch_spec, PathBatchSpec):
             reader_method = batch_spec.reader_method
             reader_options = batch_spec.reader_options
```
Kilo59, line 332: Lines `331` and `332` below are where I inject/monkeypatch my Spy for capturing the given `**reader_options`

alexsherstinsky:
A thought regarding this design principle:
```
The goal in this PR is to be able to dynamically create pandas DataAsset types by inspecting the function signatures of the pandas.io read methods (eg read_csv, read_excel, read_json etc.). 20 methods in total.
```

While previously, the burden was on the user to specify the correct reader method and reader options, this approach shifts the burden on us to keep up with the evolution of `Pandas` and other backends' versions and adjust our implementation accordingly as well as manage previous versions, deprecation, and end-of-life scenarios.  This will entail an appreciable increase in support and maintenance.  As long as this approach and costs have been incorporated into our business model, it is in and of itself an acceptable direction, but we should be fully cognizant of the trade-off we are making here.  Thank you.  /cc @donaldheppner @5u2ie

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -0,0 +1,284 @@
+import inspect
```
alexsherstinsky, line 4: @Kilo59 Since this module does not define a `Datasource` (unless I did not notice that class definition), then perhaps we can call it something like `dynamic_pandas_support.py` (or `dynamic_pandas_helper.py`), etc.?  Thanks.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -135,7 +135,19 @@ def get_batch_list_from_batch_request(
         for request, path in self._fully_specified_batch_requests_with_path(
             batch_request
         ):
-            batch_spec = PathBatchSpec(path=str(path))
+            batch_spec = PathBatchSpec(
```
alexsherstinsky, line 138: @Kilo59 @billdirks It looks like the direction is toward `DataFrameDataAsset` rather than `FilesystemDataAsset` -- in other words, the focus on the computation (e.g., `DataFrame`), not the source of data (e.g., local filesystem, or cloud filesystem, etc.).  I am curious about the following questions:
* How would we represent an `SQL` data asset?
* How would we represent other `PathBatchSpec` types (e.g., `S3BatchSpec`, `AzureBatchSpec`, `GCSBatchSpec`)?

Thank you.

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -0,0 +1,284 @@
...
+        except TypeError as err:
+            logger.error(f"{model_name} - {type(err).__name__}:{err}")
+            logger.warning(f"{model_name} fields\n{pf(fields)}")
+            continue
+        models[type_name] = model
+        model.update_forward_refs(**_TYPE_REF_LOCALS)
+
+    logger.debug(f"Needs extra handling\n{pf(dict(NEED_SPECIAL_HANDLING))}")
+    logger.debug(f"No Annotation\n{FIELD_SKIPPED_NO_ANNOTATION}")
+    return models
```
alexsherstinsky: I would like to see concrete examples of what is being returned that is generalized as ` Type[M]` above.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -177,14 +189,35 @@ def get_batch_list_from_batch_request(
         return batch_list
 
 
+_ASSET_MODELS = _generate_data_asset_models(
```
alexsherstinsky, line 192: @Kilo59 Per above -- what would be contained in `_ASSET_MODELS` (looking for examples)?  Thanks.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -177,14 +189,35 @@ def get_batch_list_from_batch_request(
...
+
 class PandasDatasource(Datasource):
     # class attrs
-    asset_types: ClassVar[List[Type[DataAsset]]] = [CSVAsset]
+    asset_types: ClassVar[List[Type[DataAsset]]] = [
+        CSVAsset,
+        ExcelAsset,
+        ParquetAsset,
+        JSONAsset,
+    ]
```
alexsherstinsky, line 215: @Kilo59 For my edification, are these different assets (which bear file extension names) differ based on the file extension being processed (e.g., ".csv", ".xlsx", ".parquet", ".json")?  Or some other way?  Because if it is by file extension (or, perhaps, also directory-based processing for `Parquet`), then perhaps we can consider the `FilesystemDataAsset` as their parent class (as in #6928 )?  Happy to discuss.  Thanks.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -177,14 +189,35 @@ def get_batch_list_from_batch_request(
...
 
     # instance attrs
     type: Literal["pandas"] = "pandas"
     name: str
-    assets: Dict[str, CSVAsset] = {}
+    assets: Dict[str, Union[CSVAsset, ExcelAsset, ParquetAsset, JSONAsset]] = {}  # type: ignore[valid-type]
```
alexsherstinsky, line 220: @Kilo59 This is where `FilesystemDataAsset` (from #6928 ) comes to mind.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -201,19 +234,97 @@ def add_csv_asset(
...
         asset = CSVAsset(
             name=name,
             path=data_path,  # type: ignore[arg-type]  # str will be coerced to Path
             regex=regex,  # type: ignore[arg-type]  # str with will coerced to Pattern
             order_by=order_by or [],  # type: ignore[arg-type]  # coerce list[str]
+            **kwargs,
+        )
+        return self.add_asset(asset)
+
+    def add_json_asset(
```
alexsherstinsky, line 257: @Kilo59 I believe that the `FilesystemDataAsset` can support JSON as well -- seems no different than CSV, et al.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -201,19 +234,97 @@ def add_csv_asset(
...
+    ) -> ExcelAsset:  # type: ignore[valid-type]
+        """Adds a Excel asset to this pandas datasource
+
+        Args:
+            name: The name of the csv asset
+            data_path: Path to directory with csv files
+            regex: regex pattern that matches csv filenames that is used to label the batches
+            kwargs: Extra keyword arguments should correspond to ``pandas.read_excel`` keyword args
+        """
+        asset = ExcelAsset(
```
alexsherstinsky, line 298: @Kilo59 Same observation for Excel.  I could add them all into #6928 -- if we so desire.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -201,19 +234,97 @@ def add_csv_asset(
...
+    ) -> ParquetAsset:  # type: ignore[valid-type]
+        """Adds a parquet asset to this pandas datasource
+
+        Args:
+            name: The name of the csv asset
+            data_path: Path to directory with csv files
+            regex: regex pattern that matches csv filenames that is used to label the batches
+            kwargs: Extra keyword arguments should correspond to ``pandas.read_parquet`` keyword args
+        """
+        asset = ParquetAsset(
```
alexsherstinsky, line 323: @Kilo59 We could add `Parquet` to #6928 as well; I am just not 100% certain whether or not the fact that `Parquet` is represented by entire directories "will just work" or if we are supposed to do something special for that.  Thanks.

tests/experimental/datasources/tasks.py
alexsherstinsky, line 22: @Kilo59 Are these files automatically cleaned up?  Thanks.

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -0,0 +1,284 @@
...
+    """
+    # TODO: parse the annotation string
+    annotation = param.annotation
+    if not isinstance(annotation, str):
+        logger.debug(type(annotation), annotation)
+        return annotation
+
+    types: list = []
+
+    union_parts = annotation.split(" | ")
```
billdirks: This could be fragile. Maybe split on `|` and then call `.strip()` on the parts?

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -0,0 +1,284 @@
...
+    io_methods = _extract_io_methods(whitelist)
+    io_method_sigs = _extract_io_signatures(io_methods)
+
+    models: Dict[str, Type[M]] = {}
+    for signature_tuple in io_method_sigs:
+
+        fields = _to_pydantic_fields(signature_tuple)
+
+        type_name = signature_tuple.name.split("read_")[1]
+        model_name = _METHOD_TO_CLASS_NAME_MAPPINGS.get(
```
billdirks, line 269: Can we get rid of this mapping and always do `f"{type_name.capitalize()}Asset"`?

great_expectations/experimental/datasources/dynamic_pandas.py
billdirks: See `pandas_datasource.py` below. These are `CSVAsset`, `ExcelAsset`, etc.

great_expectations/experimental/datasources/pandas_datasource.py
billdirks: For your SQL question see [TableAsset defined in sql_datasource.py](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/experimental/datasources/sql_datasource.py#L132).

In the ZEP worldview, the data assets are very much tied to the datasource. @Kilo59 is leveraging that he wants to be very pandas specific so he can leverage the reader method arguments. The `PathBatchSpec` in an implementation detail that he is leveraging so we can use our current execution engine. It is not a ZEP datasource/data asset concept.

So for your 2nd question, do we need an `S3BatchSpec` or will pandas handle this for us if we pass in an `s3` url into `read_csv`? Currently the `S3BatchSpec` implementation is empty. We do switch on the BatchSpec type with `isinstance` checks in some place in code but we should actively try to make those go away and code to a standard interface instead.

I also think it is fine to only support the file system one for now. If we care about the underlying storage layer, we will end up calling a factory here instead of directly instantiating a `PathBatchSpec`.

great_expectations/experimental/datasources/pandas_datasource.py
billdirks: These are all subclasses of `DataAsset` defined in `interfaces.py`.

great_expectations/experimental/datasources/pandas_datasource.py
billdirks: This would be good to do.

great_expectations/experimental/datasources/pandas_datasource.py
billdirks: These assets are very much tied to pandas while your needs are tied to spark. Maybe we should implement them separately for now and then converge once we have concrete examples of each?

great_expectations/experimental/datasources/pandas_datasource.py
billdirks: @Kilo59 I'm not sure how having a `Union` here will interact with the `DataAssetType` `TypeVar` defined in `interfaces.py`. Maybe you should use the parent DataAsset type here or maybe you aren't seeing any type errors from `Datasource.get_asset`?

tests/experimental/datasources/test_pandas_datasource.py
billdirks: Yeah, it seems like you should just put them in `SpyInterrupt` since you are raising it anyways and delete `captured_args` and `captured_kwargs`. The other way is fine too, I'd pick one.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -201,19 +234,97 @@ def add_csv_asset(
...
+        asset = JSONAsset(
+            name=name,
+            path=data_path,  # type: ignore[arg-type]  # str will be coerced to Path
+            regex=regex,  # type: ignore[arg-type]  # str with will coerced to Pattern
+            order_by=order_by or [],  # type: ignore[arg-type]  # coerce list[str]
+            **kwargs,
+        )
+        return self.add_asset(asset)
+
+    def add_excel_asset(
```
billdirks, line 282: Are we planning on dynamically generating these `add_*_asset` methods? It seems like we do a lot of work to parse the reader methods and dynamically generate the `DataAsset` classes so we can get them automatically. If we have to manually write the `add_*_asset` methods, that means we still have to do some work to integrate them. 
We definitely don't need to do this (and definitely in this PR) but I was wondering what your thoughts are.

great_expectations/experimental/datasources/pandas_datasource.py
billdirks: The pandas method `read_csv` and `read_json` have different parameters. If we want to get rid of `kwargs`, display them explicitly in autocomplete (juptyer notebook, for example), and have error messages around them they will need to be different functions on the datasource.
It is true they could all create a `FilesystemDataAsset` in terms of reading them, but @Kilo59 has put validation of these different arguments in the different `DataAsset` subclasses, eg `CSVAsset` validates the input parameters that are CSV specific while `JSONAsset` validates all the arguments that are JSON specific.

billdirks (commented):
Thanks for doing this @Kilo59! I've left some comments inline. I don't think anything I wrote is blocking. I need to wrap my head around it a bit more. Let's discuss tomorrow.

great_expectations/experimental/datasources/pandas_datasource.py
alexsherstinsky: @billdirks @Kilo59 How are these assets tied to `Pandas`?  Could you please point me to a module and/or code segment that illustrate that these assets are exhibiting functionality that is `Pandas` specific?  I did not see that in the code in `develop` prior to the present pull request, nor did I see it while browsing this pull request.  Thank you.

alexsherstinsky: @billdirks Thank you for this -- I think that I am finally starting to have a "Eureka" moment about this.  So, does that mean that we will also need `dynamic_spark.py` -- as a sibling of `dynamic_pandas.py`?

alexsherstinsky: @billdirks @Kilo59 -- What will happen if when executing `PandasDatasource.add_csv_asset()` one supplies the `base_directory_path` (i.e., `data_path: PathStr`) argument and/or the `RegularExpressions` pattern that yield no `CSV` files, but, accidentally, point to a number of files with the ".txt" extension, or ".xlsx" extension, etc. -- i.e., incompatible with the reader methods introspected by the framework for the given `DataAsset` (`CsvDataAsset` in the present example)?  Thank you.

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -0,0 +1,284 @@
...
+    # misc
+    "filepath": {"path": _FieldSpec(pathlib.Path, ...)},
+    "dtype": {"dtype": _FieldSpec(Optional[dict], None)},  # type: ignore[arg-type]
+    "dialect": {"dialect": _FieldSpec(Optional[str], None)},  # type: ignore[arg-type]
+    "usecols": {"usecols": _FieldSpec(Union[int, str, Sequence[int], None], None)},  # type: ignore[arg-type]
+    "skiprows": {"skiprows": _FieldSpec(Union[Sequence[int], int, None], None)},  # type: ignore[arg-type]
+}
+
+_METHOD_TO_CLASS_NAME_MAPPINGS: Final[Dict[str, str]] = {
+    "csv": "CSVAsset",
```
alexsherstinsky, line 130: @Kilo59 I would prefer that we spelled out `DataAsset` in the names -- and also formalized the capitalization as in the following examples:
`CsvDataAsset`

`JsonDataAsset`

`HtmlDataAsset`

etc.

There must be published guidelines for capitalization (I have seen some from the ".NET" literature). 

Thanks!

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -0,0 +1,284 @@
...
+
+        fields = _to_pydantic_fields(signature_tuple)
+
+        type_name = signature_tuple.name.split("read_")[1]
+        model_name = _METHOD_TO_CLASS_NAME_MAPPINGS.get(
+            type_name, f"{type_name.capitalize()}Asset"
+        )
+
+        try:
+            model = _create_pandas_asset_model(
```
alexsherstinsky: @Kilo59 I would prefer that we spelled out `data_asset` in method names (to avoid confusion down the road).  Thank you.

great_expectations/experimental/datasources/pandas_datasource.py
alexsherstinsky, line 282: @Kilo59 So this method would say `add_excel_data_asset()`.  Thanks.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -201,19 +234,97 @@ def add_csv_asset(
...
+        asset = ExcelAsset(
+            name=name,
+            path=data_path,  # type: ignore[arg-type]  # str will be coerced to Path
+            regex=regex,  # type: ignore[arg-type]  # str with will coerced to Pattern
+            order_by=order_by or [],  # type: ignore[arg-type]  # coerce list[str]
+            **kwargs,
+        )
+        return self.add_asset(asset)
+
+    def add_parquet_asset(
```
alexsherstinsky, line 307: @Kilo59 And this method would say `add_parquet_data_asset()`.  Thanks.

great_expectations/experimental/datasources/pandas_datasource.py
alexsherstinsky, line 257: @Kilo59 And this method would say `add_json_data_asset()`.  Thanks.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -201,19 +234,97 @@ def add_csv_asset(
         data_path: PathStr,
         regex: Union[str, re.Pattern],
         order_by: Optional[BatchSortersDefinition] = None,
-    ) -> CSVAsset:
```
alexsherstinsky, line 204: @Kilo59 I would prefer that this method would be called `add_csv_data_asset()` -- spelled out for clarity.  Thanks.

great_expectations/experimental/datasources/dynamic_pandas.py
Kilo59: The naming convention is to keep abbreviations capitalized.
Personally, I think adding `DataAsset` in the name doesn't add much. AFIAK there's no other `Asset` domain object that it could be confused with.

We can change the naming convention if this is what we want, but I don't want to do that in this PR.

great_expectations/experimental/datasources/pandas_datasource.py
Kilo59: Yes, I had that same thought.
We had something similar in an earlier version but we decided it was too much magic, without adding a lot of value. 

But as you said now it does dovetail nicely with the dynamic asset creation.

great_expectations/experimental/datasources/dynamic_pandas.py
Kilo59: We can, I was doing this to preserve the uppercase for abbreviations because I thought that was our convention.

@alexsherstinsky asked a similar question.

https://github.com/great-expectations/great_expectations/pull/6780#discussion_r1089017142

tests/experimental/datasources/tasks.py
Kilo59: No this is the utility I used to generate the test JSON files. See my longer explanation below.

https://github.com/great-expectations/great_expectations/pull/6780#discussion_r1087920374

great_expectations/experimental/datasources/pandas_datasource.py
Kilo59: Without the type union the schema for `PandasSource` won't contain any of the information about the specific asset subtypes.

But I should say that for our current needs, everything will still work (AFIAK) with `assets: Dict[str, DataAsset]`

Using a `generic` model for our `Datasource` class could help here. But I'd rather do that in a separate PR.
https://docs.pydantic.dev/usage/models/#generic-models

As for whether I get a typing error, I'm going to turn on type-checking for our test code so that we can start seeing these issues more easily.

There is a comment there about the TypeVar needing to change once a Datasource can have more than one asset.

https://github.com/great-expectations/great_expectations/blob/c82b5080d3d3793fd036f3053a8461e5ffea1df0/great_expectations/experimental/datasources/interfaces.py#L292-L293

I don't know if `mpy` considers this union a valid subtype of `DataAsset`, `Union[CSVAsset, ExcelAsset, ParquetAsset, JSONAsset]]` if so then we don't actually need to change the `TypeVar`.

great_expectations/experimental/datasources/dynamic_pandas.py
Kilo59: I agree that's safer, I was trying to avoid extra operations.
My thought was that since pandas is using postponed annotations and `X | Y` the whitespace around the `|` is always going to be consistent.
But I think it makes sense to be safer initially and we can optimize this if needed later.

What I was worried about is that if they forget to use their convention somewhere and we get a `Union[X, Y]` then this will fail.

great_expectations/experimental/datasources/pandas_datasource.py
Kilo59: I think some of this might be clearer once I put the schemas under source control.

great_expectations/experimental/datasources/pandas_datasource.py
Kilo59: Resolving a few of these just so it's easier to read the PR.

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -0,0 +1,299 @@
...
+class _FieldSpec(NamedTuple):
+    type: Type
+    default_value: object  # ... for required value
+
+
+@functools.lru_cache(maxsize=64)
+def _replace_builtins(input_: str | type) -> str | type:
+    if not isinstance(input_, str):
+        return input_
+    return input_.replace("list", "List").replace("dict", "Dict")
```
Kilo59, line 113: This would be faster as complied regex replace.
Was just trying to prove that this was the issue.

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -321,6 +326,7 @@ def get_batch_data_and_markers(  # noqa: C901 - 22
             buf.seek(0)
             df = reader_fn(buf, **reader_options)
 
+        # ZEP code path
```
billdirks: nit: I'd update this comment to make it clear it's not only ZEP that goes down this codepath:
`Experimental datasources will go down this code path`.

great_expectations/experimental/datasources/dynamic_pandas.py
billdirks: While I may personally prefer `CsvAsset` over `CSVAsset` we've decided to follow [PEP 8](https://peps.python.org/pep-0008/) and capitalize all acronyms. From the PEP: `Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.`

Kilo59:
~I renamed the `getting-started` test package because it's not a valid python package name and this was causing mypy to crash.~

https://github.com/great-expectations/great_expectations/pull/6780/files#diff-94ecb0e92ccc3d947db32df87bbcc3df310723e87b9f99774dfa6e052e0936f4

Reverted this

great_expectations/experimental/datasources/dynamic_pandas.py
billdirks: Oh, I misunderstood this. I thought `.captialize` would capitalize the whole world but it only capitalizes the first letter.

great_expectations/experimental/datasources/pandas_datasource.py
billdirks: We don't validate the file extension. I believe that means the reader method will attempt to read this file and we will get whatever the pandas reader method result is. If the `.txt` really is a `.csv` then we will read it otherwise we'll get a panda's error.

great_expectations/experimental/datasources/pandas_datasource.py
billdirks: Let's file a ticket and do it in a subsequent PR.

tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -1,20 +1,41 @@
...
+    from great_expectations.alias_types import PathStr
+    from great_expectations.data_context import AbstractDataContext
+    from great_expectations.experimental.datasources.interfaces import (
+        BatchSortersDefinition,
+    )
+
+logger = logging.getLogger(__file__)
+
+# TODO: use this to parametrize `test_data_asset_defined_for_io_read_method`
+PANDAS_IO_METHODS: list[str] = [
```
billdirks: Is this used anywhere, can we delete it?

great_expectations/experimental/datasources/dynamic_pandas.py
Kilo59: I've updated the variable names in this method to `asset_model` and `data_asset_models`

great_expectations/experimental/datasources/pandas_datasource.py
Kilo59: I don't want to change anything about the public API in this PR.

billdirks (approved):
It would be great if you could followup on the Union TypeVar thing we discussed. Otherwise LGTM, thank you!

great_expectations/execution_engine/pandas_execution_engine.py
Kilo59: Will do before merge.

tests/experimental/datasources/test_pandas_datasource.py
Kilo59: Will do

PR #6787, opened 2023-01-12
[FEATURE] Add pandas datasource sorter by refactoring into DataAsset

billdirks:
The implements panda datasource sorters by pulling them up from TableAsset and putting them in the parent class DataAsset. Sorters sort on `Batch.metadata` which is a property of `Batch` in general and is not `DataAsset` specific. Also in this PR I changed the `BatchSorter.metadata_key` to `BatchSorter.key` based on previous feedback.


Kilo59 (approved):

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -57,6 +61,40 @@ class BatchRequestError(Exception):
...
+def _batch_sorter_from_list(sorters: BatchSortersDefinition) -> List[BatchSorter]:
+    if len(sorters) == 0 or isinstance(sorters[0], BatchSorter):
+        # mypy gets confused here. Since BatchSortersDefinition has all elements of the
+        # same type in the list so if the first on is BatchSorter so are the others.
+        return cast(List[BatchSorter], sorters)
+    # Likewise, sorters must be List[str] here.
+    return [_batch_sorter_from_str(sorter) for sorter in cast(List[str], sorters)]
+
+
+def _batch_sorter_from_str(sort_key: str) -> BatchSorter:
```
Kilo59, line 100:
```suggestion
def _batch_sorter_from_str(sort_key: Union[str, BatchSorter]) -> BatchSorter:
```
If you do this can you remove the `cast` on line `79`?

billdirks: If I do this, there are other mypy errors in `_batch_sorter_from_str` since it assumes the parameter is a `str`. I can get rid of casting by using `TypeGuard`s which I have done. It might be cleanest to use 1 `TypeGuard` and 1 `cast` but 🤷 . The issue is that a `TypeGuard` can determine which one of the possible `Union` values explicitly but the `else` branch doesn't infer it must be the other `Union` type, so I have to explicitly check for both.

PR #6784, opened 2023-01-12
[MAINTENANCE] Clean up public api report part 1

anthonyburdi:
Changes proposed in this pull request:
- Add excludes to the public api report from `invoke public-api`
- This is part 1, which includes excludes from going through just over 50% of the items in the public api report.
- To see the full list, run `invoke public-api`, or look at [the tracker with rationales](https://docs.google.com/spreadsheets/d/1zuE1zmNuQchFCD4qcI_06p2hVMMZnhlGtZl_I2VwAi0/edit#gid=0).

### Definition of Done

- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code

PR #6792, opened 2023-01-12
[MAINTENANCE] Clean up public api report part 2

anthonyburdi:
- Add excludes to the public api report from `invoke public-api`
- This is part 2, which includes excludes from going through just about 75% of the items in the public api report.
- To see the full list, run `invoke public-api`, or look at [the tracker with rationales](https://docs.google.com/spreadsheets/d/1zuE1zmNuQchFCD4qcI_06p2hVMMZnhlGtZl_I2VwAi0/edit#gid=0).

Note: This PR builds on #6784 which should be merged first.

### Definition of Done


scripts/public_api_report.py
```diff
@@ -489,6 +503,287 @@ class CodeReferenceFilter:
...
+            reason="ValidationOperators are now run from Checkpoints: https://docs.greatexpectations.io/docs/guides/miscellaneous/migration_guide#manually-migrate-v2-checkpoints-to-v3-checkpoints",
+            filepath=pathlib.Path(
+                "great_expectations/validation_operators/validation_operators.py"
+            ),
+        ),
+        IncludeExcludeDefinition(
+            reason="ValidationActions are now run from Checkpoints: https://docs.greatexpectations.io/docs/guides/miscellaneous/migration_guide#manually-migrate-v2-checkpoints-to-v3-checkpoints",
+            name="run",
+            filepath=pathlib.Path("great_expectations/checkpoint/actions.py"),
+        ),
```
talagluck: Just to confirm, I would think we would consider the `_run` methods of individual Actions to be public, since those can be overridden using plugins, and are very user-facing. Is that compatible with this?

talagluck (approved):
Looks good! One question, but if it's satisfied, no need to gate on merge.

scripts/public_api_report.py
anthonyburdi: What if the classes of the actions themselves are part of the public API? e.g. SlackNotificationAction, PagerdutyAlertAction, etc? That way the docstring explaining how to configure the action is documented as part of API docs?

talagluck: I think that should work!

anthonyburdi: Ah I see also your point about overriding. Yes it would be good to also document the `_run` method of the abstract class along with the abstract class (ValidationAction) with a description about how to build your own action.

anthonyburdi: I'll add these to the list of methods to document in an upcoming PR, thanks for the discussion!

anthonyburdi: Upcoming PR: https://github.com/great-expectations/great_expectations/pull/6803

PR #6793, opened 2023-01-12
[MAINTENANCE] Add checkpoint and datadoc integration test for zep pandas datasource.

billdirks:
Checkpoints and datadocs work with zep pandas datasources. This PR adds an integration test for this. Since we are using the same data for pandas and sql, we create a helper test function. Hopefully when adding other datasources, we can use the same data and continue to use this helper.

I ended up pushing another commit into this PR which adds data assistant tests. It follows the same pattern as the checkpoint and datadoc tests where refactor the sql integration tests and parameterize by the datasource/data asset. The asserts are all identical because the underlying data is the same (or at least similar enough).


tests/experimental/datasources/integration_test_util.py
```diff
@@ -0,0 +1,131 @@
...
+from great_expectations.checkpoint import SimpleCheckpoint
+from great_expectations.data_context.types.base import IncludeRenderedContentConfig
+from great_expectations.experimental.datasources.interfaces import BatchRequest
+from great_expectations.render import (
+    AtomicDiagnosticRendererType,
+    AtomicPrescriptiveRendererType,
+)
+
+
+def run_checkpoint_and_datadoc_on_taxi_data_2019_01(
```
billdirks: This was basically pulled from `test_integration_sql_datasource.py::test_run_checkpoint_and_data_doc`. If we use the same underlying data for different datasource, we should get the same computed expectations.

Kilo59: `non-blocking`

I try to avoid using helpers in tests whenever possible.
IMO it becomes too easy to pull everything into helpers that obfuscate what's actually being tested.
Although I think you could argue pytest fixtures are basically the same thing. The `Arrange` phase works very naturally with fixtures.

Is there any part of this that would be useful to extract into a parametrized fixture that yields the necessary "arranged" pandas, SQL and postgres pieces?
https://docs.pytest.org/en/6.2.x/fixture.html#parametrizing-fixtures

I do think it would be easy to go down a rabbit hole here and spend way too much time on it.
I'm fine to push some or all of this work to a followup (or even not do it at all).

Kilo59 (approved):

tests/experimental/datasources/integration_test_util.py
```diff
@@ -0,0 +1,237 @@
+from typing import Tuple
+
+from great_expectations import DataContext
```
Kilo59: This is the wrong type.
We want either `AbstractDataContext` or `FileDataContext`.

+1 for renaming `AbstractDataContext` -> `DataContext`
https://www.youtube.com/watch?v=-J3wNP6u5YU&t=270s

Also we might want to consider type-checking our experimental test code? 🤔 ... not in this PR though.

billdirks: Thank you!

tests/experimental/datasources/integration_test_util.py
billdirks: Done. All the integration tests now live in `test_integration_datasource.py` and the tests are now parameterized by the datasource type. This now more clearly highlights what is under test.

There is still a helper function `_configure_and_run_data_assistant` that contains boiler plate code for running the data assistant. Folding that into a fixture would make what's under test less clear I believe.

Kilo59 (approved):

PR #6797, opened 2023-01-13
[MAINTENANCE] Only include relevant diagnostics info in gallery JSON

kenwade4:
- When run_diagnostics is called by build_gallery, only return relevant useful info

### DIFF
--- a/great_expectations/expectations/expectation.py
+++ b/great_expectations/expectations/expectation.py
@@ -1229,6 +1229,7 @@ class Expectation(metaclass=MetaExpectation):
         raise_exceptions_for_backends: bool = False,
         ignore_suppress: bool = False,
         ignore_only_for: bool = False,
+        for_gallery: bool = False,
         debug_logger: Optional[logging.Logger] = None,
         only_consider_these_backends: Optional[List[str]] = None,
         context: Optional[AbstractDataContext] = None,
@@ -1379,6 +1380,14 @@ class Expectation(metaclass=MetaExpectation):
             for test_result in test_results
             if test_result.error_diagnostics
         ]
+
+        # If run for the gallery, don't include a bunch of stuff
+        if for_gallery:
+            examples = []
+            gallery_examples = []
+            renderers = []
+            test_results = []
+            errors = []
 
         return ExpectationDiagnostics(
             library_metadata=library_metadata,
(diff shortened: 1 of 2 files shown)


great_expectations/expectations/expectation.py
```diff
@@ -1380,6 +1381,14 @@ def run_diagnostics(
             if test_result.error_diagnostics
         ]
 
+        # If run for the gallery, don't include a bunch of stuff
+        if for_gallery:
+            examples = []
+            gallery_examples = []
```
talagluck, line 1387: Why don't we want these things for the gallery?

talagluck (commented):
Hi Ken - I'm not too familiar with this code, so just one question here.

great_expectations/expectations/expectation.py
kenwade4: They are not being used by the gallery frontend at all anymore as of earlier this month. They weren't that useful.

One of the things we are going to do is come up with more real examples to put in the docstrings like https://deploy-preview-198--gxbrandupdate.netlify.app/expectations/expect_column_distinct_values_to_be_in_set

talagluck (approved):
LGTM

great_expectations/expectations/expectation.py
anthonyburdi: @kenwade4 Do we still need to run the tests in the gallery if we are throwing away the test_results and errors? Would probably speed up the build if we don't.

kenwade4: Yea we need to run them all still because the summary info is included in the `backend_test_result_counts` key

```
        "tests": [],
        "backend_test_result_counts": [
            {
                "backend": "pandas",
                "num_passed": 6,
                "num_failed": 0,
                "failing_names": null
            },
            {
                "backend": "sqlite",
                "num_passed": 6,
                "num_failed": 0,
                "failing_names": null
            }
```

kenwade4: Those power the logic behind the green, yellow, and red squares in the "datasource view" of the gallery.

If num_passed >= 1 and num_failed == 0 then green box
if num_passed ==0 and num_failed >= 1 then red box
if num_passed >=1 and num_failed >= 1 then yellow box
if num_passed == 0 and num_failed == 0 then grey empty box

PR #6803, opened 2023-01-13
[MAINTENANCE] Clean up public api report part 3

anthonyburdi:
- Add excludes & includes to the public api report from `invoke public-api`
- This is part 3, which completes the first pass of the items in the public api report.
- To see the full list, run `invoke public-api`, or look at [the tracker with rationales](https://docs.google.com/spreadsheets/d/1zuE1zmNuQchFCD4qcI_06p2hVMMZnhlGtZl_I2VwAi0/edit#gid=0).

This PR builds on #6784 and #6792.

### Definition of Done


scripts/public_api_report.py
```diff
@@ -784,6 +854,394 @@ class CodeReferenceFilter:
...
+            name="self_check",
+            filepath=pathlib.Path("great_expectations/datasource/new_datasource.py"),
+        ),
+        IncludeExcludeDefinition(
+            reason="False match for dict `.update()` method.",
+            name="update",
+            filepath=pathlib.Path(
+                "great_expectations/execution_engine/execution_engine.py"
+            ),
+        ),
```
talagluck, line 948: Would this be considered a basic CRUD method for Execution Engine, and so part of the Public API? Or do we not need that for the Execution Engine class.

scripts/public_api_report.py
```diff
@@ -784,6 +854,394 @@ class CodeReferenceFilter:
...
+                "great_expectations/rule_based_profiler/parameter_builder/parameter_builder.py"
+            ),
+        ),
+        IncludeExcludeDefinition(
+            reason="Not used directly but from Data Assistant or RuleBasedProfiler",
+            name="run",
+            filepath=pathlib.Path(
+                "great_expectations/rule_based_profiler/rule/rule.py"
+            ),
+        ),
```
talagluck, line 997: Are these needed for the Data Assistant contribution process?

scripts/public_api_report.py
```diff
@@ -784,6 +854,394 @@ class CodeReferenceFilter:
...
+        IncludeExcludeDefinition(
+            reason="to_json_dict is an internal helper method",
+            name="to_json_dict",
+            filepath=pathlib.Path("great_expectations/core/run_identifier.py"),
+        ),
+        IncludeExcludeDefinition(
+            reason="to_json_dict is an internal helper method",
+            name="to_json_dict",
+            filepath=pathlib.Path("great_expectations/data_context/types/base.py"),
+        ),
```
talagluck: I think these should be public. They are used for passing things around in the Airflow Operator, and some users use it when they need to pass around various forms of config. My hope is that the need for it goes away when we transition to New Style Datasources, but we’re not quite there yet.

scripts/public_api_report.py
```diff
@@ -784,6 +854,394 @@ class CodeReferenceFilter:
...
+        IncludeExcludeDefinition(
+            reason="to_json_dict is an internal helper method",
+            name="to_json_dict",
+            filepath=pathlib.Path("great_expectations/types/attributes.py"),
+        ),
+        IncludeExcludeDefinition(
+            reason="to_json_dict is an internal helper method",
+            name="to_json_dict",
+            filepath=pathlib.Path("great_expectations/types/base.py"),
+        ),
```
talagluck: See above. I don't think all `to_json_dict` methods need to be public, but I think any classes that are public, and/or whose config objects are public, should have public `to_json_dict` methods.

scripts/public_api_report.py
```diff
@@ -784,6 +854,394 @@ class CodeReferenceFilter:
...
+            name="to_json_dict",
+            filepath=pathlib.Path("great_expectations/validator/exception_info.py"),
+        ),
+        IncludeExcludeDefinition(
+            reason="to_json_dict is an internal helper method",
+            name="to_json_dict",
+            filepath=pathlib.Path(
+                "great_expectations/validator/metric_configuration.py"
+            ),
+        ),
```
talagluck: See above.

talagluck (commented):
Looks good overall! A couple of questions and some opinion base changes.

scripts/public_api_report.py
anthonyburdi: Good question - this method doesn't exist though on the Execution Engine. It is an artifact of the string matching we are doing to find usages in our documentation examples.

scripts/public_api_report.py
anthonyburdi: I don't see either `Rule.run()` or `ParameterBuilder.get_validator()` used in DataAssistant, OnboardingDataAssistant or the VolumeDataAssistant. It looks like it gets called internally. 
Could you point me to the Data Assistant contribution process to double check? I haven't been able to find it in our docs.

scripts/public_api_report.py
anthonyburdi: Updated (not all, see https://github.com/great-expectations/great_expectations/pull/6803/commits/5be5c7016e942973c8db0fa61ac2350b293ff0a7 for specifics)

scripts/public_api_report.py
anthonyburdi: Updated (see https://github.com/great-expectations/great_expectations/pull/6803/commits/5be5c7016e942973c8db0fa61ac2350b293ff0a7 for specifics)

scripts/public_api_report.py
anthonyburdi: Updated (not all, see https://github.com/great-expectations/great_expectations/pull/6803/commits/5be5c7016e942973c8db0fa61ac2350b293ff0a7 for specifics)

scripts/public_api_report.py
talagluck: Ah, it doesn't exist yet! Though it is on the roadmap. Perhaps @alexsherstinsky can help answer this? Otherwise we can always add this to the Public API when it's needed.

talagluck (approved):
LGTM! Thanks for the updates. For the RBP stuff, we can always add this to the Public API when we're ready for the DA contribution docs, so no need to gate on that.

PR #6806, opened 2023-01-13
[MAINTENANCE] Add overload for `get_context`

cdkini:
- Add overload for `EphemeralDataContext`

### Definition of Done

### DIFF
--- a/great_expectations/util.py
+++ b/great_expectations/util.py
@@ -95,9 +95,11 @@ if TYPE_CHECKING:
     import numpy.typing as npt
 
     from great_expectations.alias_types import PathStr
-    from great_expectations.data_context import FileDataContext
-    from great_expectations.data_context.data_context.abstract_data_context import (
+    from great_expectations.data_context import (
         AbstractDataContext,
+        CloudDataContext,
+        EphemeralDataContext,
+        FileDataContext,
     )
     from great_expectations.data_context.types.base import DataContextConfig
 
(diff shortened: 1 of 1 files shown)


great_expectations/util.py
```diff
@@ -1783,7 +1803,7 @@ def get_context(
     ge_cloud_access_token: Optional[str] = None,
     ge_cloud_organization_id: Optional[str] = None,
     ge_cloud_mode: Optional[bool] = None,
-) -> AbstractDataContext:
+) -> Union[CloudDataContext, FileDataContext, EphemeralDataContext]:
```
Kilo59, line 1806:
```suggestion
) -> Union[AbstractDataContext, CloudDataContext, FileDataContext, EphemeralDataContext]:
```
We still need the `AbstractDataContext` to handle the case where no specific subtype can be determined from the overload

great_expectations/util.py
```diff
@@ -1747,6 +1749,24 @@ def get_context(
...
+    cloud_organization_id: None = ...,
+    cloud_mode: Optional[Literal[False]] = ...,
+    # <GX_RENAME> Deprecated as of 0.15.37
+    ge_cloud_base_url: None = ...,
+    ge_cloud_access_token: None = ...,
+    ge_cloud_organization_id: None = ...,
+    ge_cloud_mode: Optional[Literal[False]] = ...,
+) -> EphemeralDataContext:
+    ...
+
```
Kilo59, line 1768: Should still add the `CloudDataContext` here. Although I don't think it's worth it to enumerate all the possible permutations of args that could create a `CloudDataContext`.
Would that be `8!` (factorial)??

I do think we can ignore the overlap error as long as the common cases are ["revealing"](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html?highlight=reveal_type#when-you-re-puzzled-or-when-things-are-complicated) the correct types. In another similar mypy issue, one of the mypy contributors even recommends an `ignore` comment as a solution.

https://github.com/python/mypy/issues/8656#issuecomment-1193170195

Kilo59: As per my above comment... there are obviously a lot more permutations that could return a `CloudDataContext` but something like these 2 seem like they might be the most common.

Also keep in mind the order of the overloads matter. So maybe `FileDataContext` -> `Ephemeral` -> `Cloud`

```python
@overload
def get_context(
    # not sure what the other items can be
    cloud_base_url: str = ...,
    cloud_access_token: str = ...,
    cloud_organization_id: str = ...,
    cloud_mode: Optional[Literal[True]] = ...,
    # <GX_RENAME> Deprecated as of 0.15.37
    ge_cloud_base_url: Optional[str] = ...,
    ge_cloud_access_token: Optional[str] = ...,
    ge_cloud_organization_id: Optional[str] = ...,
    ge_cloud_mode: Optional[Literal[True]] = ...,
) -> CloudDataContext:
    ...

@overload
def get_context(
    # not sure what the other items can be
    cloud_base_url: Optional[str] = ...,
    cloud_access_token: Optional[str] = ...,
    cloud_organization_id: Optional[str] = ...,
    cloud_mode: Literal[True] = ...,
    # <GX_RENAME> Deprecated as of 0.15.37
    ge_cloud_base_url: Optional[str] = ...,
    ge_cloud_access_token: Optional[str] = ...,
    ge_cloud_organization_id: Optional[str] = ...,
    ge_cloud_mode: Optional[Literal[True]] = ...,
) -> CloudDataContext:
    ...

```

Kilo59:
Why close?

PR #6832, opened 2023-01-18
[BUGFIX] Pandas ID/PK query was causing DataDocs error

Shinnnyshinshin:
## What was the issue?
* `unexpected_index_query` is a list in the case of Pandas, and this was causing problems with rendering. This PR includes a step to cast to `str` in cases where it isn't a string already. 

## Why wasn't the bug caught before?
* `test_render_ValidationResultsTableContentBlockRenderer.py` is intended to test whether rendering of the EVR objects was successful, but this didn't include the `unexpected_index_query` for pandas.


great_expectations/expectations/expectation.py
```diff
@@ -876,6 +876,9 @@ def _diagnostic_unexpected_table_renderer(
         )
         if result_dict.get("unexpected_index_query"):
             query = result_dict.get("unexpected_index_query")
+            # in Pandas case, this is a list
+            if not isinstance(query, str):
+                query = str(query)
```
Shinnnyshinshin, line 881: convert `list()` to `str` in the case of Pandas so that it can be rendered.

tests/checkpoint/test_checkpoint.py
```diff
@@ -4392,6 +4391,11 @@ def test_newstyle_checkpoint_instantiates_and_produces_a_validation_result_when_
     assert len(context.validations_store.list_keys()) == 1
     assert result["success"]
 
+    first_result_key: str = list(result.run_results.keys())[0]
+    assert result.run_results[first_result_key]["actions_results"][
+        "update_data_docs"
+    ].get("local_site")
+
```
Shinnnyshinshin: Testing DataDocs site is rendered for `spark`

tests/render/test_render_ValidationResultsTableContentBlockRenderer.py
```diff
@@ -133,7 +134,6 @@ def test_ValidationResultsTableContentBlockRenderer_generate_expectation_row_wit
     result = ValidationResultsTableContentBlockRenderer.render(
         [evr_failed_with_exception]
     ).to_json_dict()
-    print(result)
```
anthonyburdi, line 136: ❤️

anthonyburdi (approved):
Nice find!

PR #6844, opened 2023-01-19
[FEATURE] Experimental `Datasource` and `DataAsset` connection testing

NathanFarmer:
- GREAT-1416
- Add `test_connection` methods to `Datasource`s and `DataAsset`s
- Call `test_connection` when adding a `Datasource` or `DataAsset`

### Definition of Done


great_expectations/experimental/datasources/metadatasource.py
```diff
@@ -49,7 +49,9 @@ def __new__(
         def _datasource_factory(name: str, **kwargs) -> Datasource:
             # TODO: update signature to match Datasource __init__ (ex update __signature__)
             LOGGER.debug(f"5. Adding '{name}' {cls_name}")
-            return cls(name=name, **kwargs)
+            datasource = cls(name=name, **kwargs)
+            datasource.test_connection()
```
Kilo59, line 53: I still think `test_connection` should be an argument here. Even if it defaults to `True`.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -183,14 +191,19 @@ class PandasDatasource(Datasource):
...
-        from great_expectations.execution_engine.pandas_execution_engine import (
-            PandasExecutionEngine,
-        )
-
-        return PandasExecutionEngine
+    def test_connection(self) -> None:
+        if self.assets:
+            for asset in self.assets.values():
+                success = False
+                for filename in os.listdir(asset.path):
```
Kilo59: Use `pathlib` for this. `asset.path` is already a `pathlib.Path` object.
https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name
```python
for filepath in asset.path.iterdir():
   if asset.regex.match(filepath.name):
```

great_expectations/experimental/datasources/sqlite_datasource.py
```diff
@@ -108,7 +109,7 @@ class SqliteDatasource(SQLDatasource):
     # right side of the operator determines the type name
     # left side enforces the names on instance creation
     type: Literal["sqlite"] = "sqlite"  # type: ignore[assignment]
-    connection_string: SqliteConnectionString
+    connection_string: str
```
Kilo59:
```suggestion
    connection_string: SqliteDsn
```
??

BTW this is very slick. Nice work 🚀 
The errors will be beautiful

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -183,19 +167,33 @@ def test_construct_table_asset_directly_with_splitter(create_source):
...
+        )()
+        (
+            batch_data,
+            batch_markers,
+        ) = mock_sqlalchemy_execution_engine.get_batch_data_and_markers(
+            batch_spec=fake_batch_spec
+        )
+        with mock.patch(
+            "great_expectations.execution_engine.SqlAlchemyExecutionEngine.get_batch_data_and_markers"
+        ) as get_batch_data_and_markers:
```
Kilo59: Don't need to worry about this right now while we are discussing the design, but I'd like to see this refactored into a reusable fixture.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -136,6 +133,13 @@ class TableAsset(DataAsset):
     column_splitter: Optional[SqlYearMonthSplitter] = None
     name: str
 
+    # non-field private attrs
+    _datasource: SQLDatasource = pydantic.PrivateAttr()
+
+    @property
+    def datasource(self) -> SQLDatasource:
+        return self._datasource
```
Kilo59: Why do we need to re-implement this on all the child classes?
Is it just for the better typing?

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -29,11 +29,16 @@
     Datasource,
 )
 
+try:
+    from sqlalchemy import create_engine
+    from sqlalchemy.engine import Engine
+except ImportError:
+    create_engine = None
+    Engine = None
```
Kilo59: I'd like it if we imported the top-level packages so that it's more apparent in the annotations that `Engine` is a `sqlalchemy.engine.Engine`?
```suggestion
SQLA_IMPORTED: bool = False
try:
    import sqlalchemy
    SQLA_IMPORTED: bool = True
except ImportError:
    pass
```

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -316,12 +322,28 @@ class SQLDatasource(Datasource):
...
+    def engine(self) -> Engine:
+        return self._engine
+
+    class Config:
+        arbitrary_types_allowed = True
+        extra = pydantic.Extra.forbid
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._engine = create_engine(self.connection_string)
```
Kilo59: I think this should be defined on the parent `Datasource`
And we can use `<DATASOURCE>._excluded_eng_args` to know which arguments to pass.

Maybe the concrete `DataAsset` or `DataSource` should define a `_create_engine()` method?? 🤔 
Which would also be very handy for doing overrides ❗

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -305,38 +296,11 @@ class Datasource(
 
     # class attrs
     asset_types: ClassVar[List[Type[DataAsset]]] = []
-    # Datasource instance attrs but these will be fed into the `execution_engine` constructor
-    _excluded_eng_args: ClassVar[Set[str]] = {
-        "name",
-        "type",
-        "execution_engine",
-        "assets",
-    }
```
Kilo59: By removing this we now force each `DataAsset` (or `DS`)  definition to manually instantiate the execution engine. As opposed to just adding a field or 2 to the `Datasource` model.

We are adding extra steps to defining a `DataSource`/`DataAsset` with no obvious benefit (that I can see).
Maybe I'm wrong here 🤔 ...

tests/experimental/datasources/test_config.py
```diff
@@ -233,7 +231,7 @@ def test_catch_bad_asset_configs(
         "my_test_ds": {
             "type": "postgres",
             "name": "my_test_ds",
-            "connection_string": "postgres://userName:@hostname/dbName",
+            "connection_string": "postgresql://userName:@hostname/dbName",
```
Kilo59: You probably don't want to change this one. This is intentionally wrong.

Kilo59 (commented):
I like a lot of these ideas.
Left several comments, but I wouldn't alter the design before we've had a chance to meet.
I think I'm starting to see the light on creating the engines at the `DataAsset` level 💡 😄 .

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -316,12 +322,28 @@ class SQLDatasource(Datasource):
...
+    _engine: Engine = pydantic.PrivateAttr()
 
-        return SqlAlchemyExecutionEngine
+    @property
+    def engine(self) -> Engine:
+        return self._engine
+
+    class Config:
+        arbitrary_types_allowed = True
+        extra = pydantic.Extra.forbid
```
Kilo59: I think the configs are combined so you shouldn't need to set `extra` here.
```suggestion
```

Kilo59: Actually, why do we need  `arbitrary_types_allowed = True` at all?
`_engine` is a `PrivateAttr` so doesn't `pydantic` just ignore it?

great_expectations/experimental/datasources/metadatasource.py
NathanFarmer: I would need to default to `False` and then we pass `True` here if that's what we decide.

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: Only `SqlDatasource` needs a SQL Alchemy `engine` so it would seem to be an anti-pattern to promote this.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: As discussed in our check-in I will be adding this back.

great_expectations/experimental/datasources/metadatasource.py
Kilo59: I don't quite follow. Wouldn't `_datasource_factory("my_name")` with `test_connection=True` be the same as what you are proposing? Making it default to `False` would be a change.
```python
  def _datasource_factory(name: str, test_connection: bool = True, **kwargs) -> Datasource:
      LOGGER.debug(f"5. Adding '{name}' {cls_name}")
      datasource = cls(name=name, **kwargs)
     if test_connection:
        datasource.test_connection()
     ...
```

billdirks: We discussed in our daily meeting today. It sounds like `test_connection` is relevant to all datasources since in pandas you'd want to check the existence of the data directory. If this is true, we would want to add `test_connection` as an abstract method in `Datasource` since we are mandating that every datasource implements it. We might want to change the name since checking for the presence of a data sounds different than what most people think about a connection, though I don't have anything better at the moment.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -183,14 +191,19 @@ class PandasDatasource(Datasource):
...
 
-    @property
-    def execution_engine_type(self) -> Type[ExecutionEngine]:
-        """Return the PandasExecutionEngine unless the override is set"""
-        from great_expectations.execution_engine.pandas_execution_engine import (
-            PandasExecutionEngine,
-        )
-
-        return PandasExecutionEngine
+    def test_connection(self) -> None:
```
billdirks: Should the asset be able to verify itself? Then this could call the underlying asset method.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -44,6 +45,11 @@ class CSVAsset(DataAsset):
     _unnamed_regex_param_prefix: str = pydantic.PrivateAttr(
         default="batch_request_param_"
     )
+    _datasource: PandasDatasource = pydantic.PrivateAttr()
```
billdirks: I don't think we need to redefine this here since it's declared on the superclass. You might need to use `Generics` and a `TypeVar` which bounds `_datasource` to a subclass of `Datasource`. However `Datasource` is a `Generic` that depends on the `DataAsset` type so I'm not sure if that will work right now.  
How this works for `Datasource` may need to change soon though since right now the `DataAssetType` `TypeVar` assumes only 1 `DataAsset` type.

great_expectations/experimental/datasources/sql_datasource.py
Kilo59: Only `SqlDatasource` needs a `SQLAlchemy` engine as returned by `sqlalchemy.engine.create_engine` but everything needs an "ExectuionEngine".
I understand that the`self._engine` here is not a `GX` `ExectuionEngine`. It's just prompting me to think about our design more.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Discussed in meeting today. This is for typing. Hopefully we can find a way around this but it might be in a followup.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -258,9 +264,9 @@ def get_batch_list_from_batch_request(
                 )
             # Creating the batch_spec is our hook into the execution engine.
             batch_spec = SqlAlchemyDatasourceBatchSpec(**batch_spec_kwargs)
-            data, markers = self.datasource.execution_engine.get_batch_data_and_markers(
-                batch_spec=batch_spec
-            )
+            data, markers = SqlAlchemyExecutionEngine(
```
billdirks: We discussed leaving the `execution_engine` on the `datasource` as a property that is not created on `__init__` but instead is deferred until referenced on the property.

billdirks (commented):
Thanks for this draft @NathanFarmer! I think addressing the comments in this PR and any others from the group discussion today will put this in a good spot.

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: I removed it with the recent changes. I think this was because I didn't always have `_engine` as a `PrivateAttr`.

great_expectations/experimental/datasources/sqlite_datasource.py
NathanFarmer: I had this like this and planned to add it back once I got around to fixing the typing errors.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -316,13 +321,37 @@ class SQLDatasource(Datasource):
...
+        return self._engine
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        if SQLALCHEMY_IMPORTED:
+            self._engine = sqlalchemy.create_engine(self.connection_string)
+        else:
+            raise SQLDatasourceError(
+                "Unable to create SQLDatasource due to missing sqlalchemy dependency."
+            )
```
Kilo59: Can we get rid of the custom `__init__` and make this a validator?

NathanFarmer: Yes, that is the plan!

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -320,23 +311,15 @@ class Datasource(
...
     def execution_engine(self) -> ExecutionEngine:
-        return self._execution_engine
-
-    class Config:
-        # TODO: revisit this (1 option - define __get_validator__ on ExecutionEngine)
-        # https://pydantic-docs.helpmanual.io/usage/types/#custom-data-types
-        arbitrary_types_allowed = True
+        engine_kwargs = {
+            k: v for (k, v) in self.__dict__.items() if k not in self._excluded_eng_args
+        }
```
```suggestion
        engine_kwargs = self.dict(exclude=self._excluded_eng_args)
```
https://docs.pydantic.dev/usage/exporting_models/#modeldict

great_expectations/experimental/datasources/pandas_datasource.py
NathanFarmer: Yeah that makes sense. We can move this into `CSVAsset`.

great_expectations/experimental/datasources/pandas_datasource.py
NathanFarmer: ✅

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: This was removed after deciding I don't need the child-class type in this PR.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: ✅

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: ✅

great_expectations/experimental/datasources/pandas_datasource.py
NathanFarmer: ✅

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -396,6 +398,7 @@ def add_asset(self, asset: DataAssetType) -> DataAssetType:
         # The setter for datasource is non-functional, so we access _datasource directly.
         # See the comment in DataAsset for more information.
         asset._datasource = self
+        asset.test_connection()
```
NathanFarmer, line 413: Happy to discuss adding a `test_connection` arg to `add_asset` after consulting with product.

great_expectations/experimental/datasources/metadatasource.py
NathanFarmer: Happy to discuss adding a `test_connection` arg to the datasource factory after consulting with product.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -323,6 +351,90 @@ def execution_engine_type(self) -> Type[ExecutionEngine]:
 
         return SqlAlchemyExecutionEngine
 
+    class Config:
+        arbitrary_types_allowed = True
+
+    @pydantic.root_validator()
```
NathanFarmer: Now THIS is why you use a `root_validator`!

tests/experimental/datasources/test_metadatasource.py
```diff
@@ -162,8 +165,11 @@ def test_ds_execution_engine_type_not_defined(
         class MissingExecEngineTypeDatasource(Datasource):
             type: str = "valid"
 
+            def test_connection(self) -> None:
+                ...
+
         with pytest.raises(NotImplementedError):
-            MissingExecEngineTypeDatasource(name="name")
+            MissingExecEngineTypeDatasource(name="name").get_execution_engine()
```
NathanFarmer, line 172: `ExecutionEngine` is no longer instantiated in tandem with `Datasource`.

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -37,17 +42,24 @@ def _source(
...
-        yield PostgresDatasource(
-            name="my_datasource",
-            connection_string=connection_string,
+        with mock.patch("sqlalchemy.create_engine"):
+            postgres_datasource = PostgresDatasource(
+                name="my_datasource",
+                connection_string=connection_string,
+                engine=engine,
+            )
+        postgres_datasource.engine = MockSaEngine(
```
NathanFarmer: We could do something similar to this with `execution_engine_override`. Then we won't be able to make this immutable though.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -136,6 +141,29 @@ class TableAsset(DataAsset):
...
+        Raises:
+            TestConnectionError
+        """
+        assert isinstance(self.datasource, SQLDatasource)
+        schema: Optional[str] = None
+        table_name: str
+        if "." in self.table_name:
+            schema, table_name = self.table_name.split(".")
+        else:
+            table_name = self.table_name
```
NathanFarmer, line 159: I can harden this logic in the subsequent PR by adding a schema parameter to `TableAsset`.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -228,6 +248,20 @@ def execution_engine_type(self) -> Type[ExecutionEngine]:
...
+        Args:
+            test_assets: If assets have been passed to the PandasDatasource, whether to test them as well.
+
+        Raises:
+            TestConnectionError
+        """
+        # Only self.assets can be tested for PandasDatasource
+        if self.assets and test_assets:
+            for asset in self.assets.values():
+                asset.test_connection()  # type: ignore[union-attr]
```
Kilo59, line 263:
```suggestion
                asset.test_connection()
```
Is the union-attr ignore still relevant?
Is `test_connection` not defined on `DataAsset`?

great_expectations/experimental/datasources/sql_datasource.py
Kilo59: You'll have to use an `alias` or a different name for a `schema` arg/field.
It's a reserved field for pydantic.
Ran into this for pandas SQL assets as well.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -323,6 +353,58 @@ def execution_engine_type(self) -> Type[ExecutionEngine]:
 
         return SqlAlchemyExecutionEngine
 
+    class Config:
+        arbitrary_types_allowed = True
+
+    @pydantic.root_validator()
```
Kilo59:
```suggestion
    @pydantic.root_validator
```

tests/experimental/datasources/test_config.py
```diff
@@ -134,7 +135,8 @@
     ],
 )
 def test_load_config(inject_engine_lookup_double, load_method: Callable, input_):
-    loaded: GxConfig = load_method(input_)
+    with mock.patch("sqlalchemy.create_engine"):
+        loaded: GxConfig = load_method(input_)
```
Kilo59, line 139: Using it as a context manager shouldn't cause an issue.

But if you use `pytest-mock` or `pytest`'s built in `monkeypatch` they always automatically remove any applied patch after the test ends.
So you'll never have to worry about the future you or someone else forgetting and leaving a patch in-place.

https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html#how-to-monkeypatch-mock-modules-and-environments

Note: `pytest-mock` is a drop in replacement for `unittest.mock` but with some extra features.
https://pytest-mock.readthedocs.io/en/latest/usage.html
https://pytest-mock.readthedocs.io/en/latest/remarks.html

great_expectations/experimental/datasources/experimental_base_model.py
```diff
@@ -40,6 +40,12 @@ class ExperimentalBaseModel(pydantic.BaseModel):
     class Config:
         extra = pydantic.Extra.forbid
 
+    def __key(self):
+        return tuple(self.dict())
+
+    def __hash__(self):
+        return hash(self.__key())
```
Kilo59: Is this because of `_get_execution_engine` lru cache?
Which models specifically get can be passed to `_get_execution_engine()`?
We might not want to do this on the `ExperimentalBaseModel`.

Kilo59: It looks like instances of custom classes are normally hashable by default.
But `pydantic` models are not (by default).

https://docs.python.org/3.8/glossary.html (search `hashable`)

>An object is hashable if it has a hash value which never changes during its lifetime (it needs a [__hash__()](https://docs.python.org/3.8/reference/datamodel.html#object.__hash__) method), and can be compared to other objects (it needs an [__eq__()](https://docs.python.org/3.8/reference/datamodel.html#object.__eq__) method). Hashable objects which compare equal must have the same hash value.

>...

>Most of Python’s immutable built-in objects are hashable; mutable containers (such as lists or dictionaries) are not; immutable containers (such as tuples and frozensets) are only hashable if their elements are hashable. Objects which are instances of user-defined classes are hashable by default. They all compare unequal (except with themselves), and their hash value is derived from their [id()](https://docs.python.org/3.8/library/functions.html#id).

---------

https://docs.pydantic.dev/usage/model_config/#options (see `frozen`)

>setting `frozen=True` does everything that `allow_mutation=False` does, and also generates a `__hash__()` method for the model. This makes instances of the model potentially hashable if all the attributes are hashable. (default: False)

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -362,6 +361,16 @@ def _execution_engine_type(self) -> Type[ExecutionEngine]:
         """Returns the execution engine to be used"""
         return self.execution_engine_override or self.execution_engine_type
 
+    @functools.lru_cache(maxsize=1)
+    def _get_execution_engine(self, **engine_kwargs) -> ExecutionEngine:
```
Kilo59: You may want to create a `clear_exec_engine_cache` fixture for use in testing.
```python
@pytest.fixture
def clear_exec_engine_cache():
     # maybe you want to inspect the cache before cleaning it??
    print(_get_execution_engine.cache_info())
    _get_execution_engine.clear_cache()
   return
```

https://docs.python.org/3/library/functools.html#functools.lru_cache

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: I also ran into this for `RendererConfiguration`

great_expectations/experimental/datasources/pandas_datasource.py
NathanFarmer: Something about your new dynamic pandas assets isn't passing along the method defined on the models.

Kilo59: Ah, yes I have noticed that a lot of the typing information is getting destroyed.
I'll create a ticket for that.

great_expectations/experimental/datasources/experimental_base_model.py
NathanFarmer: Yes, this is due to `lru_cache` needing all args to be hashable. We can move this down to the datasource level if we chose to do so. For some reason the `lru_cache` also needs the `Datasource` to be hashable or it throws a `TypeError`.

billdirks: If this is not a property you need, you could manually cache it by setting a variable on the data source, eg `my_datasource._execution_engine` which get's populated the first time `execution_engine` is called. You could store the kwargs or other metadata if you need to know when to reconnect.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -98,15 +109,8 @@ def _query_for_year_and_month(
...
-    # We should make an assertion about the execution_engine earlier. Right now it is assigned to
-    # after construction. We may be able to use a hook in a property setter.
-    from great_expectations.execution_engine import SqlAlchemyExecutionEngine
-
-    assert isinstance(
-        table_asset.datasource.execution_engine, SqlAlchemyExecutionEngine
-    )
-
-    with table_asset.datasource.execution_engine.engine.connect() as conn:
+    assert isinstance(table_asset.datasource, SQLDatasource)
```
billdirks: Thank you!

billdirks (commented):
This LGTM. I'll let you and @Kilo59 hash out any remaining comments.

Kilo59 (approved):
Non of my requests are blocking.

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -282,7 +314,9 @@ def collect_batch_spec(spec: SqlAlchemyDatasourceBatchSpec) -> None:
     with create_source(
         validate_batch_spec=collect_batch_spec, dialect="postgresql"
     ) as source:
-        asset = source.add_table_asset(name="my_asset", table_name="my_table")
+        source, asset = create_and_add_table_asset_without_testing_connection(
+            source=source, name="my_asset", table_name="my_table"
+        )
```
Kilo59, line 314: This test fails.

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -258,7 +288,9 @@ def collect_batch_spec(spec: SqlAlchemyDatasourceBatchSpec) -> None:
...
```
Kilo59, line 288: This test fails

great_expectations/experimental/datasources/experimental_base_model.py
NathanFarmer: This was removed along with `lru_cache`.

PR #6868, opened 2023-01-20
[DOCS] enable running `invoke docstrings` on select modules

Kilo59:
- `check_public_api_docstrings` can take a `select_paths` which will prevent scanning the whole repo for docstring errrors
- `invoke docstrings` can multiple `--path`/`-p` args to which are passed down as `select_paths`


anthonyburdi (approved):
Thanks for splitting!

PR #6880, opened 2023-01-21
[DOCS] Add some json serialization docstrings.

billdirks:
Add 36 public api docstrings. These are mostly for implementations of `to_json_dict`. Docstrings were also added to `convert_to_json_serializable`. I've also updated the return type of these methods except for 1 instance (see comment inline) where some implementation/typing work needs to be done that is distinct from docstring work.

### Definition of Done

Thank you for submitting!


great_expectations/render/components.py
```diff
@@ -682,8 +785,13 @@ def __repr__(self) -> str:
     def __str__(self) -> str:
         return json.dumps(self.to_json_dict(), indent=2)
 
-    def to_json_dict(self) -> Optional[dict]:
-        """Returns RenderedAtomicValueGraph as a json dictionary."""
+    @public_api
+    def to_json_dict(self) -> Optional[dict[str, JSONValues]]:
```
billdirks, line 790: This looks like a potential bug. @NathanFarmer should we be calling some serialization function on `self.graph`?

great_expectations/core/expectation_validation_result.py
```diff
@@ -456,7 +464,13 @@ def __repr__(self):
     def __str__(self):
         return json.dumps(self.to_json_dict(), indent=2)
 
+    @public_api
```
billdirks, line 485: This is the only function I didn't annotate the return type on. There is some more implementation/typing work that needs to be done to get this correct because of he current implementation.

great_expectations/types/__init__.py
```diff
@@ -236,12 +238,12 @@ def assert_valid_keys(keys: Set[str], purpose: str) -> None:
 
 
 class SerializableDictDot(DictDot):
-    def to_json_dict(self) -> dict:
-        """
+    def to_json_dict(self) -> Dict[str, JSONValues]:
```
billdirks, line 241: This is a method that needs to be implemented by subclasses. I didn't know if we considered that as public (it's not the spreadsheet) so I didn't add the decorator but I still wanted to add the docstring.

great_expectations/render/components.py
NathanFarmer: Yes, as it stands right now we just use Altair (vega-lite) for graphs, so this serializes them for rendering at the client.

billdirks: I have confirmed with @NathanFarmer that this is ok for renderers.

anthonyburdi (approved):

PR #6888, opened 2023-01-22
[MAINTENANCE] switch to `ruff` linter

Kilo59:
If accepted this PR would change our primary code linter from `flake8` -> `ruff`.

https://notes.crmarsh.com/python-tooling-could-be-much-much-faster
https://github.com/charliermarsh/ruff#ruff

image

## TODO

- [x] Update documentation to reflect changes
- [x] update requirements files and dependency tracking
- [x] Update `invoke` tasks to use `ruff`
   - [x] add optional `invoke sort --ruff` usage
   - [x] `invoke lint` uses `ruff`
       - [x] `invoke lint --fix`
       - [x] `invoke lint --watch` 
   - [x] add new `invoke fix`
- [x] Update CI steps to use `ruff`
  - [x] install `ruff` instead of `flake8`
  - [x] update unused import checker to use `ruff` 
- [x] migrate `flake8` `setup.cfg` config to `ruff` `pyproject.toml` config
- [x] update pre-commit hooks (add `ruff`, remove `flake8`)
- [x] remove `flake8` dev/contrib dependency 
- ~replace `isort` with `ruff`~ (will address this at another time)
- [x] replace and remove `pyupgrade`
- [x] present to team and get acceptance 

### Rules to consider adopting
There are a lot of rules I think we should consider adopting but these ones standout.
https://github.com/charliermarsh/ruff#flake8-bugbear-b
https://github.com/charliermarsh/ruff#flake8-simplify-sim
https://github.com/charliermarsh/ruff#flake8-datetimez-dtz
https://github.com/charliermarsh/ruff#flake8-use-pathlib-pth

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,11 +26,6 @@ repos:
       - id: ruff
         files: ^(great_expectations/core)  # TODO: expand to all of gx
         args: ["--fix"]
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v2.7.2
-    hooks:
-      - id: pyupgrade
-        args: ["--py3-plus"]
 # https://pre-commit.ci/
 ci:
     autofix_commit_msg: |
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
+    rev: v4.4.0
     hooks:
       - id: check-ast
       - id: check-json
@@ -13,7 +13,7 @@ repos:
       - id: requirements-txt-fixer
         exclude: docs/.*
   - repo: https://github.com/timothycrosley/isort
-    rev: 5.10.1
+    rev: 5.11.4
     hooks:
       - id: isort
   - repo: https://github.com/psf/black
@@ -22,7 +22,7 @@ repos:
       - id: black
   - repo: https://github.com/charliermarsh/ruff-pre-commit
     # Ruff version.
-    rev: 'v0.0.228'
+    rev: 'v0.0.229'
     hooks:
       - id: ruff
         files: ^(great_expectations/core)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: 'v0.0.229'
+    rev: 'v0.0.230'
     hooks:
       - id: ruff
         files: ^(great_expectations/core)  # TODO: expand to all of gx
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,11 +21,11 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    # Ruff version.
     rev: 'v0.0.229'
     hooks:
       - id: ruff
-        files: ^(great_expectations/core)
+        files: ^(great_expectations/core)  # TODO: expand to all of gx
+        args: ["--fix"]
   - repo: https://github.com/asottile/pyupgrade
     rev: v2.7.2
     hooks:
@@ -41,8 +41,4 @@ ci:
     autoupdate_branch: 'develop'
     autoupdate_commit_msg: '[pre-commit.ci] pre-commit autoupdate'
     autoupdate_schedule: monthly
-    # skip flake8 enforcement with pre-commit CI as it is enforced elsewhere
-    # flake8 hook also doesn't have an options to auto-fix
-    # pre-commit hooks should either have mechanisms for auto-fixing or should be simple to do manually
-    skip: [flake8]
     submodules: false
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: v3.2.0
     hooks:
       - id: check-ast
       - id: check-json
@@ -13,7 +13,7 @@ repos:
       - id: requirements-txt-fixer
         exclude: docs/.*
   - repo: https://github.com/timothycrosley/isort
-    rev: 5.11.4
+    rev: 5.10.1
     hooks:
       - id: isort
   - repo: https://github.com/psf/black
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,10 +20,11 @@ repos:
     rev: 22.3.0
     hooks:
       - id: black
-  - repo: https://github.com/PyCQA/flake8
-    rev: 5.0.4
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    # Ruff version.
+    rev: 'v0.0.228'
     hooks:
-      - id: flake8
+      - id: ruff
         files: ^(great_expectations/core)
   - repo: https://github.com/asottile/pyupgrade
     rev: v2.7.2
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 [![Twitter Follow](https://img.shields.io/twitter/follow/expectgreatdata?style=social)](https://twitter.com/expectgreatdata)
 [![Slack Status](https://img.shields.io/badge/slack-join_chat-white.svg?logo=slack&style=social)](https://greatexpectations.io/slack)
 [![Contributors](https://img.shields.io/github/contributors/great-expectations/great_expectations)](https://github.com/great-expectations/great_expectations/graphs/contributors)
+[![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/charliermarsh/ruff/main/assets/badge/v1.json)](https://github.com/charliermarsh/ruff)
 
 <!-- <<<Super-quickstart links go here>>> -->
 
--- a/azure-pipelines-dev.yml
+++ b/azure-pipelines-dev.yml
@@ -103,11 +103,10 @@ stages:
             displayName: 'Use Python 3.7'
 
           - script: |
-              pip install $(grep -E '^(black|invoke|isort|ruff|pyupgrade)' reqs/requirements-dev-contrib.txt)
+              pip install $(grep -E '^(black|invoke|isort|ruff)' reqs/requirements-dev-contrib.txt)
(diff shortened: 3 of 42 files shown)


docs/contributing/style_guides/code_style.md
```diff
@@ -54,14 +54,14 @@ Options:
 
 ### Linting
 
-Our CI system will check using `black`, `isort`, `flake8` and `pyupgrade`. 
+Our CI system will check using `black`, `isort`, `ruffand `pyupgrade`.
```
Kilo59, line 57:
```suggestion
Our CI system will check using `black`, `isort`, `ruff` and `pyupgrade`.
```

great_expectations/self_check/util.py
```diff
@@ -139,8 +139,8 @@
 _BIGQUERY_MODULE_NAME = "sqlalchemy_bigquery"
 try:
     # noinspection PyPep8Naming
-    import sqlalchemy_bigquery as sqla_bigquery
     import sqlalchemy_bigquery as BigQueryDialect
+    import sqlalchemy_bigquery as sqla_bigquery
```
Kilo59, line 143: revert these changes

.pre-commit-config.yaml
```diff
@@ -13,18 +13,19 @@ repos:
...
-  - repo: https://github.com/PyCQA/flake8
-    rev: 5.0.4
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: 'v0.0.229'
     hooks:
-      - id: flake8
-        files: ^(great_expectations/core)
+      - id: ruff
+        files: ^(great_expectations/core)  # TODO: expand to all of gx
+        args: ["--fix"]
```
Kilo59, line 28: Ruff can actually fix a fair number of issues, which makes it much more useful as a pre-commit hook.
For example, if it finds an `unused-import`, it will remove it.

tasks.py
```diff
@@ -74,33 +101,58 @@ def fmt(
...
+    if watch:
+        cmds.append("--watch")
+    ctx.run(" ".join(cmds), echo=True, pty=pty)
+
+
+@invoke.task(help={"path": _PATH_HELP_DESC})
+def fix(ctx: Context, path: str = "."):
+    """Automatically fix all possible code issues."""
+    lint(ctx, path=path, fix=True)
+    fmt(ctx, path=path, sort_=True)
```
Kilo59, line 148: This is the only new `invoke` task in this PR.

great_expectations/cli/python_subprocess.py
```diff
@@ -40,9 +40,8 @@ def execute_shell_command(command: str) -> int:
             check=True,
             encoding=None,
             errors=None,
-            text=None,
             env=env,
-            universal_newlines=True,
+            text=True,
```
Kilo59, line 44: `universal_newlines` has been deprecated in favor of `text` which was added in `3.7`
https://docs.python.org/3.7/library/subprocess.html#subprocess.run

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
+    rev: v4.4.0
```
cdkini: Any reason to do these bumps in this PR? I'm okay with it, just curious

.pre-commit-config.yaml
cdkini: do we always use all of our imports? If I do an `import sqlalchemy as sa`, am I guaranteed to use `sa`? I'm not so confident about this but you've done the research and I guess you're right!

README.md
```diff
@@ -6,6 +6,7 @@
 [![Twitter Follow](https://img.shields.io/twitter/follow/expectgreatdata?style=social)](https://twitter.com/expectgreatdata)
 [![Slack Status](https://img.shields.io/badge/slack-join_chat-white.svg?logo=slack&style=social)](https://greatexpectations.io/slack)
 [![Contributors](https://img.shields.io/github/contributors/great-expectations/great_expectations)](https://github.com/great-expectations/great_expectations/graphs/contributors)
+[![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/charliermarsh/ruff/main/assets/badge/v1.json)](https://github.com/charliermarsh/ruff)
```
cdkini, line 9: ❤️ love your attention to detail with these things

docs/guides/expectations/creating_custom_expectations/how_to_create_custom_column_aggregate_expectations.md
```diff
@@ -281,15 +281,14 @@ Completeness checklist for ExpectColumnMaxToBeBetweenCustom:
 
 ### 8. Linting
 
-Finally, we need to lint our now-functioning Custom Expectation. Our CI system will test your code using `black`, `isort`, `flake8`, and `pyupgrade`. 
+Finally, we need to lint our now-functioning Custom Expectation. Our CI system will test your code using `black`, `isort`, and `ruff`.
```
cdkini, line 284: By definition, isn't a linter is something that just does static analysis but doesn't fix the issue for you? This let's us run various linting tools in parallel?

Should we really be using `--fix` in all cases? If so, would we want to add the same command for `black` and `isort` below?

docs/guides/expectations/creating_custom_expectations/how_to_create_custom_table_expectations.md
```diff
@@ -285,15 +285,14 @@ Completeness checklist for ExpectTableColumnsToBeUnique:
 
 ### 8. Linting
 
-Finally, we need to lint our now-functioning Custom Expectation. Our CI system will test your code using `black`, `isort`, `flake8`, and `pyupgrade`. 
+Finally, we need to lint our now-functioning Custom Expectation. Our CI system will test your code using `black`, `isort`, and `ruff`.
```
cdkini, line 288: Looks like we use this snippet a lot - not import here but let's maybe find a way to reuse this? cc @Rachel-Reverie I think we have the ability to save snippets through our own MDX right?

docs/guides/expectations/creating_custom_expectations/how_to_create_custom_column_aggregate_expectations.md
Kilo59: I agree that's the way I think of a linter too but our instructions for `black`,  `isort` and `pyupgrade` have contributors calling them in such a way that it **will auto-fix** the issues.

Before this change, `flake8` was the outlier in that it didn't actually fix anything.

.pre-commit-config.yaml
Kilo59: No not really. I had run `pre-commit autoupdate`

Initially, I only bumped `isort` because I wanted to know if the very minor differences with the `ruff` sorting rule were because of an old `isort` version (it wasn't).

I'll remove it so that the diff is cleaner.

README.md
dctalbot: what is this for?

.pre-commit-config.yaml
Kilo59: No there are cases where we just need to "re-export" or raise up an import like in `__init__.py`.
For that we ignore these rules for all `__init__.py` just like we did before in the `flake8` `setup.cfg` file.
Also if the import is declared in `__all__` it will also be ignored.

But just as further reassurance, if I missed one of these cases it would also fail the CI lint step. We don't run `--fix` anywhere in the CI.
The only way it would happen is with an installed pre-commit hook or manually running `invoke lint --fix` or `ruff . --fix`

pyproject.toml
```diff
@@ -186,6 +186,71 @@ convention = "google"
 [tool.darglint]
 docstring_style = "google"
 
+[tool.ruff]
+target-version = "py37"
+line-length = 88
```
dctalbot, line 191: For cloud, I decided line length style should be delegated to `black`, so we ignored this rule e.g.
```toml
ignore = [
    "E501" # line length
]
```

Not a big deal, but you need to keep ruff and black in sync with the current settings

tasks.py
```diff
@@ -39,25 +39,51 @@
...
 )
 def sort(
-    ctx: Context, path: str = ".", check: bool = False, exclude: str | None = None
+    ctx: Context,
+    path: str = ".",
+    check: bool = False,
+    exclude: str | None = None,
+    ruff: bool = False,
+    isort: bool = False,  # isort is the current default
+    pty: bool = True,
```
Kilo59, line 66: @NathanFarmer if you or anyone want to try out the `ruff` based sorting.
`invoke sort --ruff`

or 

`invoke sort --ruff --check` to see the diff of what it WOULD change.

Will do a followup PR with a proposal to make it the default.

dctalbot (approved):

pyproject.toml
Kilo59: I have it set to ignore this rule right now. We have too many violations even if we wanted to enforce it. 😅 
`black` won't actually enforce the length in certain cases like if you have comments or docstrings that go beyond the line-length.

.pre-commit-config.yaml
cdkini: no real qualms here tbh will defer to you

README.md
cdkini: just a shield in our README

docs/guides/expectations/creating_custom_expectations/how_to_create_custom_column_aggregate_expectations.md
cdkini: okay I'm fine with this

cdkini (approved):

README.md
Kilo59: The most important part of the PR 😆 
image

azure-pipelines-dev.yml
```diff
@@ -149,9 +148,9 @@ stages:
     - job: unused_import_checker
       steps:
       - script: |
-          pip install $(grep -E '^(flake8)' reqs/requirements-dev-contrib.txt)
+          pip install $(grep -E '^(ruff)' reqs/requirements-dev-contrib.txt)
           # https://www.flake8rules.com/rules/F401.html
-          flake8 --select F401 great_expectations tests
+          ruff --select F401 great_expectations tests
```
Kilo59, line 153: @anthonyburdi if you want to have ruff only check for a particular code or category of violations you can use `--select` just like with `flake8`.

So you could do `ruff --select D` to check all the pydocstyle rules
https://github.com/charliermarsh/ruff#pydocstyle-d

README.md
dctalbot: ah cool! that's a nice way to give them credit. wasn't sure if there was some integration going on behind the scenes.

PR #6908, opened 2023-01-24
[FEATURE] Add Validation Result URL to Checkpoint Result

josectobar:
Include a `validation_result_url` for GX Cloud in Checkpoint Run output:
- add a `validation_result_url` arg to `CheckpointResult` class
 
image

### Definition of Done

Thank you for submitting!


great_expectations/checkpoint/checkpoint.py
```diff
@@ -217,7 +220,16 @@ def run(
 
                 checkpoint_run_results.update(run_results)
 
+        # Generate a URL to the validation result details page in GX Cloud
+        validation_result_url = None
+        cloud_base_url: Final[str] = CLOUD_APP_DEFAULT_BASE_URL
```
Kilo59: I wouldn't create this `cloud_base_url` variable at all. I would use `CLOUD_APP_DEFAULT_BASE_URL` directly.
```python
        validation_result_url: Optional[str] = None
        for key in checkpoint_run_results:
            if isinstance(key, GXCloudIdentifier) and key.cloud_id:
                validation_result_url = (
                    f"{CLOUD_APP_DEFAULT_BASE_URL}?validationResultId={key.cloud_id}"
                )
```

### Sidenote
This note below is just FYI, you don't actually need `Final` here.
You're getting an error because we are doing type-checking on the oldest supported `python` `3.7` which doesn't have a `typing.Final`.
You can import it and other `typing` types `from typing_extensions import Final`.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -217,7 +220,16 @@ def run(
...
+        for key in checkpoint_run_results:
+            if isinstance(key, GXCloudIdentifier) and key.cloud_id:
+                validation_result_url = (
+                    f"{cloud_base_url}?validationResultId={key.cloud_id}"
+                )
```
Kilo59: If we get a `validation_result_url` can we break out of the loop early?
```python
        for key in checkpoint_run_results:
            if isinstance(key, GXCloudIdentifier) and key.cloud_id:
                validation_result_url = (
                    f"{cloud_base_url}?validationResultId={key.cloud_id}"
                break # we don't need to keep iterating
```

Kilo59 (changes requested):
2 minor suggestions

great_expectations/checkpoint/types/checkpoint_result.py
```diff
@@ -300,18 +306,21 @@ def to_json_dict(self) -> dict:
...
         }
-        serializeable_dict = recursively_convert_to_json_serializable(
-            test_obj=serializeable_dict
+        if not self.validation_result_url:
+            serializable_dict.pop("validation_result_url")
+        serializable_dict = recursively_convert_to_json_serializable(
+            test_obj=serializable_dict
         )
-        return serializeable_dict
+        return serializable_dict
```
Kilo59: I would avoid the extra assignment (`=`) operation.
```python
        if not self.validation_result_url:
            serializable_dict.pop("validation_result_url")
        return recursively_convert_to_json_serializable(serializable_dict)
```

Kilo59 (approved):
1 non-blocking suggestion

great_expectations/checkpoint/types/checkpoint_result.py
Kilo59: On second thought, you shouldn't need to call this method at all. All the members are already json serializable. Not sure why it was there before 🤷.
If you want to leave it as that's fine. But I would remove the extra assignment like in the example above ☝️  
```python
        if not self.validation_result_url:
            serializable_dict.pop("validation_result_url")
        return serializable_dict
```

great_expectations/checkpoint/types/checkpoint_result.py
```diff
@@ -106,6 +108,10 @@ def run_results(
     def run_id(self) -> RunIdentifier:
         return self._run_id
 
+    @property
+    def validation_result_url(self) -> str:
+        return self._validation_result_url
```
cdkini: This should be `Optional[str]` I think

great_expectations/checkpoint/types/checkpoint_result.py
```diff
@@ -57,8 +57,10 @@ def __init__(
             Dict[str, Union[ExpectationSuiteValidationResult, dict, str]],
         ],
         checkpoint_config: CheckpointConfig,
+        validation_result_url: Optional[str] = None,
         success: Optional[bool] = None,
     ) -> None:
+        self._validation_result_url = validation_result_url
```
cdkini, line 63: Do we need to make a corresponding change to `CheckpointResultSchema`? See `checkpoint/types/checkpoint_result.py`

great_expectations/checkpoint/types/checkpoint_result.py
```diff
@@ -300,18 +306,18 @@ def to_json_dict(self) -> dict:
...
             ),
             "checkpoint_config": self.checkpoint_config.to_json_dict(),
             "success": convert_to_json_serializable(data=self.success),
         }
-        serializeable_dict = recursively_convert_to_json_serializable(
-            test_obj=serializeable_dict
-        )
-        return serializeable_dict
+        if not self.validation_result_url:
+            serializable_dict.pop("validation_result_url")
```
cdkini, line 319:
```suggestion
serializable_dict.pop("validation_result_url", None)
```

great_expectations/checkpoint/types/checkpoint_result.py
josectobar: I don't think this schema is being used for some reason at least not in this file. We tried to implement this change with @NathanFarmer but it was giving us an error. I decided to leave this out for now.

great_expectations/checkpoint/types/checkpoint_result.py
Kilo59: We aren't type-checking this code otherwise CI would catch it.

NathanFarmer (approved):

PR #6914, opened 2023-01-24
[MAINTENANCE] partial `checkpoint` type checking

Kilo59:
- Minimal type-checking of `checkpoint`
  - bring `checkpoint/checkpoint.py` and `checkpoint/util.py` under mypy coverage


great_expectations/checkpoint/configurator.py
```diff
@@ -182,29 +191,29 @@ def _build_checkpoint_config(self) -> CheckpointConfig:
...
         return [
             ActionDicts.STORE_VALIDATION_RESULT,
             ActionDicts.STORE_EVALUATION_PARAMS,
         ]
 
     def _add_update_data_docs_action(
-        self, action_list: list[ActionDict]
-    ) -> list[ActionDict]:
+        self, action_list: List[ActionDict]
+    ) -> List[ActionDict]:
```
Kilo59:
```suggestion
        self, action_list: list[ActionDict]
    ) -> list[ActionDict]:
```

great_expectations/checkpoint/configurator.py
```diff
@@ -182,29 +191,29 @@ def _build_checkpoint_config(self) -> CheckpointConfig:
...
         update_docs_action = copy.deepcopy(ActionDicts.UPDATE_DATA_DOCS)
         if isinstance(self.site_names, list):
             update_docs_action["action"]["site_names"] = self.site_names
         action_list.append(update_docs_action)
         return action_list
 
-    def _add_slack_action(self, action_list: list[ActionDict]) -> list[ActionDict]:
+    def _add_slack_action(self, action_list: List[ActionDict]) -> List[ActionDict]:
```
Kilo59:
```suggestion
    def _add_slack_action(self, action_list: list[ActionDict]) -> list[ActionDict]:
```

great_expectations/checkpoint/configurator.py
```diff
@@ -182,29 +191,29 @@ def _build_checkpoint_config(self) -> CheckpointConfig:
...
         """
         The underlying SlackNotificationAction and SlackRenderer default to
         including links to all sites if the key notify_with is not present. We
         are intentionally hiding this from users of SimpleCheckpoint by having a
         default of "all" that sets the configuration appropriately.
         """
-        _notify_with = self.notify_with
+        _notify_with: Optional[Union[str, List[str]]] = self.notify_with
```
Kilo59, line 216:
```suggestion
        _notify_with: str | list[str] | None= self.notify_with
```

great_expectations/checkpoint/util.py
```diff
@@ -515,7 +515,8 @@ def send_sns_notification(
     sns = session.client("sns")
     try:
         response = sns.publish(**message_dict)
-    except sns.exceptions.InvalidParameterException:
+    except sns.exceptions.InvalidParameterException as err:
         logger.error(f"Received invalid for message: {validation_results}")
+        return err.__class__.__name__
```
Kilo59:
```suggestion
```

great_expectations/checkpoint/configurator.py
```diff
@@ -182,7 +191,7 @@ def _build_checkpoint_config(self) -> CheckpointConfig:
         return CheckpointConfig(**config_kwargs)
 
     @staticmethod
-    def _default_action_list() -> list[ActionDict]:
+    def _default_action_list() -> List[ActionDict]:
```
Kilo59, line 194:
```suggestion
    def _default_action_list() -> list[ActionDict]:
```

great_expectations/checkpoint/util.py
```diff
@@ -525,6 +525,8 @@ def send_sns_notification(
     try:
         response = sns.publish(**message_dict)
     except sns.exceptions.InvalidParameterException:
-        logger.error(f"Received invalid for message: {validation_results}")
+        error_msg = f"Received invalid for message: {validation_results}"
+        logger.error(error_msg)
+        return error_msg
```
Kilo59, line 530: This IS a runtime behavior change.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -911,7 +917,7 @@ def construct_from_config_args(
         config_version: Optional[Union[int, float]] = None,
         template_name: Optional[str] = None,
         module_name: Optional[str] = None,
-        class_name: Optional[str] = None,
+        class_name: Literal["Checkpoint", "SimpleCheckpoint"] = "Checkpoint",
```
Kilo59, line 920:
```suggestion
        class_name: str = "Checkpoint",
```
This causes warnings in `data_context` because the type there is just a `str`

great_expectations/checkpoint/checkpoint.py
```diff
@@ -804,7 +807,8 @@ def __init__(
             template_name=template_name,
             run_name_template=run_name_template,
             expectation_suite_name=expectation_suite_name,
-            batch_request=batch_request,
+            batch_request=batch_request,  # type: ignore[arg-type] # FluentBatchRequest is not a dict
+            # TODO: check if `pydantic.BaseModel` and call `batch_request.dict()`??
```
Kilo59, line 813: Wanted to call this out because I would expect for things to break here if we passed a `FluentBatchRequest` here as is.

pyproject.toml
```diff
@@ -67,9 +67,7 @@ exclude = [
     # TODO: complete typing for the following modules and remove from exclude list
     # number is the current number of typing errors for the excluded pattern
     'checkpoint/actions\.py',  # 18
-    'checkpoint/checkpoint\.py',  # 22
     'checkpoint/types/checkpoint_result\.py',  # 34
```
Kilo59, line 70: Planning followup PR for these remaining 2 modules.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -321,9 +321,9 @@ def run(  # noqa: C901 - complexity 19
                 meta: ExpectationSuiteValidationResultMeta
                 validation_result_url: str | None = None
                 for run_result in run_results.values():
-                    validation_result = run_result.get("validation_result")
+                    validation_result = run_result.get("validation_result")  # type: ignore[assignment] # could be dict
                     if validation_result:
-                        meta = validation_result.meta
+                        meta = validation_result.meta  # type: ignore[assignment] # could be dict
```
cdkini, line 328: If its a dict, can we just use `[]` notation? I believe whatever rich type we use allows for access the object like that?

great_expectations/checkpoint/checkpoint.py
```diff
@@ -357,7 +357,7 @@ def get_substituted_config(
         if runtime_kwargs is None:
             runtime_kwargs = {}
 
-        config_kwargs: dict = self.get_config(mode=ConfigOutputModes.JSON_DICT)
+        config_kwargs: dict = self.get_config(mode=ConfigOutputModes.JSON_DICT)  # type: ignore[assignment] # always returns a dict
```
cdkini, line 362: If it always return a dict, do we need this? Does the explicit annotation cause an issue? Should we cast instead?

great_expectations/checkpoint/checkpoint.py
```diff
@@ -640,11 +640,14 @@ def action_list(self) -> List[Dict]:
             return []
 
     @property
-    def validations(self) -> List[CheckpointValidationConfig]:
+    def validations(
+        self,
+    ) -> list[CheckpointValidationConfig] | list[dict]:  # always a list[dict]
```
cdkini, line 647: If this is always `list[dict]`, why the union? Could we maybe add a line adding the context we discussed (if you deem in applicable)?

great_expectations/checkpoint/checkpoint.py
cdkini: hmm good call - could maybe be a good ticket for an FDS cleanup epic?

great_expectations/checkpoint/checkpoint.py
```diff
@@ -840,7 +844,7 @@ def run_with_runtime_args(
         run_id: Optional[Union[str, int, float]] = None,
         run_name: Optional[str] = None,
         run_time: Optional[datetime.datetime] = None,
-        result_format: Optional[str] = None,
+        result_format: str | dict | None = None,  # TODO: type-dict?
```
cdkini, line 849: Would be great 🙇🏽 - maybe not necessary here to begin with

great_expectations/checkpoint/checkpoint.py
```diff
@@ -911,7 +915,7 @@ def construct_from_config_args(
         config_version: Optional[Union[int, float]] = None,
         template_name: Optional[str] = None,
         module_name: Optional[str] = None,
-        class_name: Optional[str] = None,
+        class_name: str = "Checkpoint",
```
cdkini: Is this a literal or can it be any string?

great_expectations/checkpoint/checkpoint.py
```diff
@@ -1036,7 +1040,10 @@ def instantiate_from_config_with_runtime_args(
             if value is not None:
                 config[key] = value
 
-        config = filter_properties_dict(properties=config, clean_falsy=True)
+        config = filter_properties_dict(  # type: ignore[assignment] # filter could return None
+            properties=config,
+            clean_falsy=True,
+        )
```
cdkini, line 1048: oof I'm going to look into this whole nullable return type stuff separately

cdkini (approved):

great_expectations/checkpoint/checkpoint.py
Kilo59: I don't want to cast because that's like ignoring all errors.

I want to make `ConfigPeer`  a generic and define `overloads` for `.get_config()` but that won't work until `ConfigPeer` is a generic.

I can add a comment explaining these aspirational goals.

great_expectations/checkpoint/checkpoint.py
Kilo59: I tried to make it a `Literal` but I started get this warning in `absDataContext`
```
great_expectations/data_context/data_context/abstract_data_context.py:2109: error:
Argument "class_name" to "construct_from_config_args" of "Checkpoint"
has incompatible type "Optional[str]"; expected "Literal['Checkpoint', 'SimpleCheckpoint']"  [arg-type]
```
Although if that's the only place maybe it's better to just ignore/fix the problem there.

Kilo59: Oh actually I'm still getting something very similar so let me take a look. I may add this back.
```
great_expectations/data_context/data_context/abstract_data_context.py:2109: error:
Argument "class_name" to "construct_from_config_args" of "Checkpoint" has
incompatible type "Optional[str]"; expected "str"  [arg-type]
                    class_name=class_name,
                               ^~~~~~~~~~
```

great_expectations/checkpoint/checkpoint.py
```diff
@@ -153,7 +153,7 @@ def run(  # noqa: C901 - complexity 19
         profilers: Optional[List[dict]] = None,
         run_id: Optional[Union[str, RunIdentifier]] = None,
         run_name: Optional[str] = None,
-        run_time: Optional[Union[str, datetime.datetime]] = None,
+        run_time: Optional[datetime.datetime] = None,
```
alexsherstinsky, line 158: @Kilo59 Would this not lead to type checking errors?..

great_expectations/checkpoint/checkpoint.py
alexsherstinsky: If the result is a conf object extending `DictDot` then it supports the dot notation (and also acts dict-like).  Hence, I am not certain that the `ignore` statement is appropriate (and am curious as to what it will take to resolve type checking errors currently showing up in Azure CI).  Thanks.

great_expectations/checkpoint/checkpoint.py
alexsherstinsky: It is not clear to me either why the `ignore` statement is needed; hence, an explanation would be useful.  Thanks.

great_expectations/checkpoint/checkpoint.py
Kilo59: It didn't cause any mypy errors.
That doesn't mean it won't ever, just that it didn't based on our current type-checking coverage.

great_expectations/checkpoint/checkpoint.py
alexsherstinsky: Because our Checkpoint code is unjustifiably complex, unfortunately `Union` is needed (I filed a comprehensive JIRA ticket, which poses more questions than offering solutions; hence, it should be discussed).  Thanks.

great_expectations/checkpoint/checkpoint.py
alexsherstinsky: The Config objects that extend `DictDot` only accept primitive types -- by design; hence `batch_request` is always a `dict`, meaning that the `ignore` statement may not be appropriate.  However, I would like to point out that this `Marshmallow`-based design is overdue for being implemented on `Pydantic` instead, where typed objects will be the norm.  Thanks.

great_expectations/checkpoint/checkpoint.py
alexsherstinsky: @Kilo59 and @cdkini This is one of the key problematic areas with the current design (and part of the JIRA ticket I mentioned above).  Our design of `SimpleCheckpoint` is anything but simple -- in fact, it is extremely and unjustifiably complex, because we have violated the basic design principal of API sophistication continuity.  Thanks.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -1378,6 +1384,7 @@ class SimpleCheckpoint(Checkpoint):
     """
 
     _configurator_class = SimpleCheckpointConfigurator
+    name: str
```
alexsherstinsky, line 1389: Is this `name` needed?  By the way, the "Configurator" is a very complex solution to the problem that was not posed correctly.

great_expectations/checkpoint/configurator.py
```diff
@@ -24,21 +24,30 @@
 logger = logging.getLogger(__name__)
 
 
+class ActionDetails(TypedDict):
```
alexsherstinsky, line 30: This is very nice as a bandaid.  Ultimately, we should obviate `Configurator` by a more elegant design, based on the principle of broader `Checkpoint` API being offered based on the user's growth in sophistication.

great_expectations/checkpoint/checkpoint.py
Kilo59: Yes, otherwise `mypy` thinks the `self.name` attribute could be `None`, but it's always present.

great_expectations/core/batch.py
```diff
@@ -914,7 +914,7 @@ def _is_fluent_batch_request(args: dict[str, Any]) -> bool:
 
 
 def batch_request_contains_batch_data(
-    batch_request: Optional[Union[BatchRequestBase, dict]] = None
+    batch_request: Optional[Union[BatchRequestBase, FluentBatchRequest, dict]] = None
```
alexsherstinsky, line 917: @Kilo59 I am not sure if this is needed; do we already have `FluentBatchRequest` ability to accept `batch_data`?  I do recall a PR being in the works to allow `batch_request.dataframe` to be set, but was not sure.  Thanks.

great_expectations/checkpoint/checkpoint.py
Kilo59: I added it back and added an type ignore with a comment in `AbstractDataContext._resolve_add_checkpoint_args()`

alexsherstinsky (changes requested):
@Kilo59 this looks approve-ready pending answers to some questions I have raised.  Thank you very much for doing this!

great_expectations/core/batch.py
Kilo59: I'm not sure about the other points you raise but this is needed, we call it from `checkpoint.py` and pass it a `FluentBatchRequest`

great_expectations/checkpoint/checkpoint.py
Kilo59: because `get_config()` is a union of `dict` `BaseConfigYaml` and `str`

great_expectations/checkpoint/checkpoint.py
Kilo59: The ignore is needed unless we do as @cdkini suggested and use `__getitem__` .
`run_results` is a dict of dicts and mypy has no way of knowing that the `validation_result` key is a `DictDoct`. I could make `run_results` a `TypeDict` and then it could know what the `validation_result` is but the goal of this PR is just to get `checkpoint` under coverage quickly and then iterate from there.

great_expectations/checkpoint/checkpoint.py
Kilo59: The ignore IS needed, `batch_request` is NOT always a dict.

alexsherstinsky: @Kilo59 This sounds very concerning.  I am looking for a test where we submit a non-`dict` entity to Config.

alexsherstinsky: @Kilo59 Could you please take a look at line 2810 in `great_expectations/data_context/types/base.py` -- do you see why I am concerned?  Thank you.

alexsherstinsky: It is unclear why `CheckpointConfig` in that module allows only plain Python `dict` as `batch_request` while yet the `ignore` seems to be needed.  We will need to understand what is happening there in detail later.

alexsherstinsky (approved):
Approving with a request to troubleshoot later.

great_expectations/checkpoint/checkpoint.py
Kilo59: @alexsherstinsky yes, this looks like it's currently (and has been broken).
But I'm not exactly sure how it manifests itself in the typical GX user workflows yet.

Running a checkpoint works with FDS right now in OSS, but not Cloud.

PR #6915, opened 2023-01-24
[MAINTENANCE] Add test to ensure that all types in the `DataContext` hierarchy emit expected usage stats

cdkini:
- Add test to ensure that `EphemeralDataContext`, `FileDataContext`, and `CloudDataContext` all emit the same usage stats events

### Definition of Done


tests/core/usage_statistics/test_usage_statistics_data_context.py
```diff
@@ -0,0 +1,125 @@
...
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture
+def enable_usage_stats(monkeypatch):
+    monkeypatch.delenv("GE_USAGE_STATS")
+
+
+@pytest.fixture
+def usage_stats_decorated_methods_on_abstract_data_context() -> list[str]:
```
cdkini, line 33: Maybe there's a better way to do this?

tests/core/usage_statistics/test_usage_statistics_data_context.py
```diff
@@ -0,0 +1,125 @@
...
+    "great_expectations.core.usage_statistics.usage_statistics.UsageStatisticsHandler.emit"
+)
+@pytest.mark.integration
+def test_all_relevant_context_methods_emit_usage_stats(
+    mock_emit: mock.MagicMock,
+    usage_stats_decorated_methods_on_abstract_data_context: list[str],
+    enable_usage_stats,  # Needs to be before context fixtures to ensure usage stats handlers are attached
+    in_memory_runtime_context: EphemeralDataContext,
+    empty_data_context: FileDataContext,
+    empty_data_context_in_cloud_mode: CloudDataContext,
```
cdkini: As it stands, this test takes 30.1 seconds - 30 seconds for the setup of these 3 fixtures and 0.1 seconds for the actual logic below.

I'm working to clean this up.

tests/core/usage_statistics/test_usage_statistics_data_context.py
```diff
@@ -0,0 +1,122 @@
...
+            logger.info(f"Testing {context.__class__}.{method_name}")
+
+            # As we only care about the decorator and not the underlying method being decorated,
+            # we use the following try/except pattern. All method calls will generally fail due
+            # to having no input args but we still manage to trigger our target decorator.
+            try:
+                method = getattr(context, method_name)
+                method()
+            except Exception:
+                pass
```
cdkini: Does this feel bad? Maybe a little but it works pretty well? We only really care about triggering the event

tests/core/usage_statistics/test_usage_statistics_data_context.py
```diff
@@ -0,0 +1,122 @@
...
+@pytest.fixture
+def enable_usage_stats(monkeypatch):
+    monkeypatch.delenv("GE_USAGE_STATS")
+
+
+@pytest.fixture
+def usage_stats_decorated_methods_on_abstract_data_context() -> list[str]:
+    """
+    Uses AST parsing to collect the names of all methods that are decorated with
+    `usage_statistics_enabled_method`.
```
Kilo59: The `@usage_statistics_enabled_method` decorator could collect method names and append them to a global list/dict etc. (Which you could later inspect)

That should be much faster then walking an ast tree.

tests/core/usage_statistics/test_usage_statistics_data_context.py
```diff
@@ -0,0 +1,98 @@
...
+        logger.info(f"Testing {context.__class__}.{method_name}")
+
+        # As we only care about the decorator and not the underlying method being decorated,
+        # we use the following try/except pattern. All method calls will generally fail due
+        # to having no input args but we still manage to trigger our target decorator.
+        try:
+            method = getattr(context, method_name)
+            method()
+        except Exception:
+            pass
```
cdkini: This pattern unfortunately results in a side effect - `build_data_docs` runs through and actually opens a site in the browser. We need to determine a better way to do this.

Perhaps I could mock every function except the ones decorated above?

tests/core/usage_statistics/test_usage_statistics_data_context.py
```diff
@@ -0,0 +1,101 @@
...
+        UsageStatsEvents.DATA_CONTEXT_BUILD_DATA_DOCS,
+        UsageStatsEvents.DATA_CONTEXT_GET_BATCH_LIST,
+        UsageStatsEvents.DATA_CONTEXT_OPEN_DATA_DOCS,
+        UsageStatsEvents.DATA_CONTEXT_RUN_CHECKPOINT,
+        UsageStatsEvents.DATA_CONTEXT_RUN_RULE_BASED_PROFILER_ON_DATA,
+        UsageStatsEvents.DATA_CONTEXT_RUN_RULE_BASED_PROFILER_WITH_DYNAMIC_ARGUMENTS,
+        UsageStatsEvents.DATA_CONTEXT_RUN_VALIDATION_OPERATOR,
+        UsageStatsEvents.DATA_CONTEXT_SAVE_EXPECTATION_SUITE,
+    )
+    assert len(relevant_methods) == len(
```
billdirks: Can we test for equality instead of the `len` equality here? That is does each method map to 1 event? If we know this mapping, instead of a list we could have a `Dict` of method to event and then test the lengths are the same and we see the correct event when we call the method.

tests/core/usage_statistics/test_usage_statistics_data_context.py
```diff
@@ -0,0 +1,101 @@
...
+        expected_events
+    ), "Please update the `expected_events` list to account for all usage stats decorated methods in AbstractDataContext"
+
+    for method_name, expected_event in zip(relevant_methods, expected_events):
```
billdirks: Oh, you do my suggestion here. How do you control the order of `revelevant_methods` and `expected_events` here? It seems like you may want to put them in a dict to more explicitly associate them.

tests/core/usage_statistics/test_usage_statistics_data_context.py
billdirks: Can you save the event with the method name in `ENABLED_METHODS`. Then you can check directly that the method is associated with the correct event without calling the method.

You could then have a separate test for the decorator where you wrap a no-op function with the decorator configured however, and test you see the correct event emitted. That is you have a separate test that the decorator works and this test ensures the right event is associated with the right method.

tests/core/usage_statistics/test_usage_statistics_data_context.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def filter_enabled_methods(
+    enabled_methods: dict[str, UsageStatsEvents]
+) -> dict[str, UsageStatsEvents]:
+    return {
+        k.split(".")[1]: v
+        for k, v in enabled_methods.items()
+        if k.startswith("AbstractDataContext") and not k.endswith("__init__")
+    }
```
cdkini: I wanted to make this filtered dict a fixture but didn't because I need it for parameterization below - a global dict would work as well but I thought a function was a happy middle ground

tests/core/usage_statistics/test_usage_statistics_data_context.py
```diff
@@ -0,0 +1,111 @@
...
+    method_to_patch = module_path.replace(method_name, f"_{method_name}")
+
+    with mock.patch(method_to_patch) as mock_fn:
+        method = getattr(context, method_name)
+
+        # Generate a set of dummy values to trigger the target method
+        # and invoke the usage stats decorator without causing side-effects
+        signature = inspect.signature(method)
+        kwargs = {param: None for param in signature.parameters}
+        method(**kwargs)
```
cdkini, line 111: This is a bit of magic I don't love (but it works) - dynamically determine what the arg names are and supply some dummy values. With this pattern, we don't need a `try/except` block

great_expectations/core/usage_statistics/usage_statistics.py
```diff
@@ -276,6 +276,11 @@ def get_usage_statistics_handler(args_array: list) -> Optional[UsageStatisticsHa
     return handler
 
 
+# Mapping between method's qualified name and the event name it emits
+# Used to esnure proper usage stats coverage in tests
+ENABLED_METHODS = {}
```
billdirks: Can you add typing to this dict?

tests/core/usage_statistics/test_usage_statistics_data_context.py
billdirks: Do you also need to check that `k.split(".")` has length 2? Is the `1` here suppose to be grabbing the last element?

tests/core/usage_statistics/test_usage_statistics_data_context.py
```diff
@@ -0,0 +1,111 @@
...
+    [(name, event) for name, event in filter_enabled_methods(ENABLED_METHODS).items()],
+)
+@pytest.mark.parametrize(
+    "data_context_fixture_name",
+    [
+        # In order to leverage existing fixtures in parametrization, we provide
+        # their string names and dynamically retrieve them using pytest's built-in
+        # `request` fixture.
+        # Source: https://stackoverflow.com/a/64348247
+        pytest.param("in_memory_runtime_context", id="EphemeralDataContext"),
```
billdirks, line 70: Do you want to ensure that if more `DataContext` subclasses get added, they are tested here? That is you could have a test that looks for all subclasses of `DataContext` and verify you only see these 3. 
This comment is not blocking just something to think about.

tests/core/usage_statistics/test_usage_statistics_data_context.py
```diff
@@ -0,0 +1,111 @@
...
+        pytest.param("empty_data_context", id="FileDataContext"),
+        pytest.param(
+            "empty_data_context_in_cloud_mode",
+            id="CloudDataContext",
```
billdirks, line 74: I haven't used `id` before. What does it do here?

billdirks (approved):
This LGTM. I'd add typing and clarify the `split` in tests but otherwise 🚀!

great_expectations/core/usage_statistics/usage_statistics.py
cdkini: done!

tests/core/usage_statistics/test_usage_statistics_data_context.py
cdkini: allows us to reference the specific parameterized test using `-k`

tests/core/usage_statistics/test_usage_statistics_data_context.py
cdkini: hmm I don't envision us having to deal with this but a good idea! Per conversations with Don, I believe we're at the tail end of our context work

PR #6916, opened 2023-01-24
[MAINTENANCE] Typing `histogram_single_batch_parameter_builder`

Shinnnyshinshin:
Mob Typing


great_expectations/rule_based_profiler/helpers/util.py
```diff
@@ -1080,3 +1091,17 @@ def sanitize_parameter_name(
...
+class _NumericIterableWithDtype(Iterable, Protocol):
+    @property
+    def dtype(self) -> Any:
+        ...
+
+
+def _is_iterable_numeric_and_has_dtype_attr(
+    obj: Any,
+) -> TypeGuard[_NumericIterableWithDtype]:
+    if hasattr(obj, "dtype") and not np.issubdtype(obj.dtype, np.number):
```
Kilo59:
```suggestion
    if hasattr(obj, "dtype") and np.issubdtype(obj.dtype, np.number):
```
Shouldn't both these be true?

great_expectations/rule_based_profiler/helpers/util.py
```diff
@@ -1080,3 +1091,17 @@ def sanitize_parameter_name(
...
     return name.replace(FULLY_QUALIFIED_PARAMETER_NAME_SEPARATOR_CHARACTER, "_")
+
+
+class _NumericIterableWithDtype(Iterable, Protocol):
+    @property
+    def dtype(self) -> Any:
+        ...
+
+
+def _is_iterable_numeric_and_has_dtype_attr(
```
Kilo59:
```suggestion
def _is_iterable_of_numeric_dtypes(
```
Just thinking about this a bit more. We are doing more than just checking if it has the `dtype` attr aren't we? 
If we were just checking for the existence of the attr and that it's iterable the type-guard would just be...
```python
if hasattr(obj, "dtype") and isinstance(obj, Iterable):
   return True
return False
```

Shinnnyshinshin:
@Kilo59 asking for re-review when you get a chance.

Kilo59 (approved):
Sorry for the late review.

I was going to propose an alternative idea but I keep getting distracted. This works as is.

PR #6930, opened 2023-01-26
[MAINTENANCE] Fix flakey complexity warning

Kilo59:
The C901 mccabe complexity warnings seem to be somewhat flakey in CI when checking methods with decorators.
Sometimes it will complain about the decorator, and sometimes the function. Applying the `# noqa: C901` comment to both decorator and function resolves the issue.

https://stackoverflow.com/questions/51127170/flake8-ignore-warnings-for-a-function


Kilo59:
closing for now.
Can be re-opened if needed

PR #6934, opened 2023-01-26
[DOCS] Remove the `great_expectations` path prefix for API docs

anthonyburdi:
- Remove the parent `great_expectations` path prefix for API docs.
- Note: there are other issues to address e.g. consistency with Checkpoint that will be addressed separately.
- New sidebar looks like:
image


Shinnnyshinshin (approved):
love it. Thanks for the quick turnaround on this

PR #6947, opened 2023-01-27
[DOCS] Add public API docstring for `RuleBasedProfiler`

dctalbot:
- add a public docstring for `RuleBasedProfiler`.
- I'm unsure if the `"--ge-feature-maturity-info--"` that follows this docstring is still needed. Maybe we should ignore it in the generated documentation?


anthonyburdi:
> * add a public docstring for `RuleBasedProfiler`.
> * I'm unsure if the `"--ge-feature-maturity-info--"` that follows this docstring is still needed. Maybe we should ignore it in the generated documentation?

Good question - ignoring the feature maturity info for now. It already gets stripped out during pre-processing of the API docs.

anthonyburdi (approved):

PR #6928, opened 2023-01-25
[FEATURE] Refactor Filesystem DataAsset for applicability to CSV and other filesystem DataAsset types and multiple ExecutionEngine backends

alexsherstinsky:
### Scope
Currently, `CSV` `DataAsset` is enclosed in `PandasDatasource`.  However, it (and its parent, filesystem `DataAsset`) can be utilized for `SparkDFExecution` engine as well.  Hence, the code refactoring is proposed, subject to the following important observation, which applies to all new experimental `DataAsset` types under development (certainly the file path based ones; `SQL` ones potentially as well): 

```
    """
    # TODO: <Alex>
        This temporary placeholder implementation of "Local Filestem based DataAsset" pertains to local filesystem only.
        Fully-functional implementation should be architecturally analogous to general inheritance hierarchy similar to:
        - ConfiguredAssetFilesystemDataConnector<-ConfiguredAssetFilePathDataConnector<-FilePathDataConnector
        - ConfiguredAssetS3DataConnector<-ConfiguredAssetFilePathDataConnector<-FilePathDataConnector
        - ConfiguredAssetAzureDataConnector<-ConfiguredAssetFilePathDataConnector<-FilePathDataConnector
        - ConfiguredAssetGCSDataConnector<-ConfiguredAssetFilePathDataConnector<-FilePathDataConnector
        with attention to cloud storage access protocols and corresponding "ExecutionEngine" support for retrieving
        "Batch" data as depicted in "ExecutionEngine.resolve_data_reference()" as well as to data reference caching.
    # TODO: </Alex>
    """
```

### Notes
Review this pull request only after #6918 is merged.

Please annotate your PR title to describe what the PR does, then give a brief bulleted description of your PR below. PR titles should begin with [BUGFIX], [FEATURE],  [DOCS], or [MAINTENANCE]. If a new feature introduces breaking changes for the Great Expectations API or configuration files, please also add [BREAKING]. You can read about the tags in our [contributor checklist](https://docs.greatexpectations.io/docs/contributing/contributing_checklist).

Changes proposed in this pull request:
- JIRA: GREAT-579/GREAT-1607
-
-

After submitting your PR, CI checks will run and @cla-bot will check for your CLA signature.

For a PR with nontrivial changes, we review with both design-centric and code-centric lenses.

[...]

PR #6981, opened 2023-01-28
[FEATURE] Implement Experimental SparkDatasource with CSVDataAsset

alexsherstinsky:
### Scope
This change introduces Experimental `SparkDatasource` with `CSVDataAsset` (for now, `CSVAsset`).  Here are some notes:
* We need to improve type/name/extension management for `DataAsset` registration so as to make user-accessible classes/methods consistent.  (There is a "TODO" note for it.)
* We need to figure out a formal way to handle `reader_options` for `Spark` in particular and different options for various `DataAsset` implementations in particular; the current "work-around" hard-codes the needed ones into the `CSVSparkAsset` implementation.  (There is a "TODO" note for it.)
* Integration tests have been refactored to use a utility module and common `DataContext` and `DataAsset` loading methods from stored fixtures were moved into `conftest.py` and turned into `PyTest` fixtures, where appropriate.

### Additional changes
* Rename `logger` to `LOGGER` per `Python` style guidelines.
* Delete unused `MetricFunctionTypes` `Enum` constants (only `VALUE` is used).

- JIRA: GREAT-579/GREAT-1603
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


tests/experimental/datasources/test_spark_datasource.py
```diff
@@ -0,0 +1,275 @@
...
+    CSVSparkAsset,
+    SparkDatasource,
+)
+
+LOGGER = logging.getLogger(__name__)
+
+
+@pytest.fixture
+def spark_datasource() -> SparkDatasource:
+    return SparkDatasource(name="spark_datasource")
```
cdkini: We should `pytest.skip` if we don't have Spark in our `test_backends`

tests/experimental/datasources/test_spark_datasource.py
```diff
@@ -0,0 +1,275 @@
...
+def csv_path() -> pathlib.Path:
+    return pathlib.Path(
+        file_relative_path(
+            __file__,
+            pathlib.Path("..", "..", "test_sets", "taxi_yellow_tripdata_samples"),
+        )
+    )
+
+
+@pytest.mark.unit
```
cdkini, line 40: If using a true Spark connection, I don't believe these are unit tests?

cdkini (dismissed):

great_expectations/core/util.py
```diff
@@ -864,11 +862,20 @@ def get_or_create_spark_session(
...
 
-    except AttributeError:
+    # TODO: <Alex>ALEX</Alex>
+    # except AttributeError:
+    #     logger.error(
+    #         "Unable to load spark context; install optional spark dependency for support."
+    #     )
+    #     spark_session = None
+    # TODO: <Alex>ALEX</Alex>
+    # TODO: <Alex>ALEX</Alex>
```
cdkini: Are these comments necessary? Can we remove them?

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -496,6 +498,7 @@ def head(
             HeadData
         """
         self.data.execution_engine.batch_manager.load_batch_list(batch_list=[self])
+        # TODO: <Alex>ALEX_NOTE-1/27/2023: This should be accomplished through calling "MetricsCalculator.head()"; however, circular-imports problem currently prevents importing "MetricsCalculator".</Alex>
```
cdkini: Is there a ticket to capture this? If not, let's do so because code comments tends to get lost in the mix

great_expectations/experimental/datasources/spark_datasource.py
```diff
@@ -0,0 +1,229 @@
...
+    BatchSortersDefinition,
+    DataAsset,
+    Datasource,
+)
+
+if TYPE_CHECKING:
+    from great_expectations.execution_engine import ExecutionEngine
+
+LOGGER = logging.getLogger(__name__)
+
```
cdkini: Nit - why do we use `LOGGER` some places and `logger` elsewhere? Can we standardize?

great_expectations/experimental/datasources/spark_datasource.py
```diff
@@ -0,0 +1,229 @@
...
+
+class SparkDatasourceError(Exception):
+    pass
+
+
+# TODO: <Alex>ALEX_NOTE-1/27/2023: We did not add "_pandas" suffix (nor qualified class name) for CSVSparkAsset in PandasDatasource module (we might need to do that at a later date, or come up with a different, more general, solution altogether).</Alex>
```
cdkini: Same note as before - for any code comments, do they make more sense as Jira tickets or discussion points at arch review or in our Slack? We have many outdated comments in our codebase - I'm wary of adding more if they aren't necessary

great_expectations/experimental/datasources/spark_datasource.py
```diff
@@ -0,0 +1,229 @@
...
+            This list will be empty if no files exist on disk that correspond to the input
+            batch request.
+        """
+        option_to_group_id = self._option_name_to_regex_group_id()
+        group_id_to_option = {v: k for k, v in option_to_group_id.items()}
+        batch_requests_with_path: List[Tuple[BatchRequest, pathlib.Path]] = []
+
+        all_files: List[pathlib.Path] = list(
+            pathlib.Path(self.base_directory).iterdir()
+        )
```
cdkini, line 85: Nit - we don't need to explicitly type here (or anywhere else where the type can be implied. `mypy` dislikes these annotations unless there is ambiguity present

great_expectations/experimental/datasources/spark_datasource.py
```diff
@@ -0,0 +1,229 @@
...
+            name: The name of the csv asset
+            base_directory: base directory path, relative to which CSV file paths will be collected
+            regex: regex pattern that matches csv filenames that is used to label the batches
+            order_by: one of "asc" (ascending) or "desc" (descending) -- the method by which to sort "Asset" parts.
+        """
+        asset = CSVSparkAsset(
+            name=name,
+            base_directory=base_directory,  # type: ignore[arg-type]  # str will be coerced to Path
+            regex=regex,  # type: ignore[arg-type]  # str with will coerced to Pattern
+            order_by=order_by or [],  # type: ignore[arg-type]  # coerce list[str]
```
cdkini, line 259: Are all these `type: ignore` comments necessary? Is there a way to get around it?

tests/experimental/datasources/test_utils.py
```diff
@@ -0,0 +1,377 @@
+from __future__ import annotations
```
cdkini: Do we really need a utils file? Could these not be moved to fixture or ported over to the test files where they are used?

I personally think its very difficult to know what utils exist

great_expectations/core/util.py
alexsherstinsky: @cdkini This was debug -- deleting (in next commit).

great_expectations/experimental/datasources/interfaces.py
alexsherstinsky: @cdkini Good idea -- I will create the JIRA ticket.  Thanks.

great_expectations/experimental/datasources/spark_datasource.py
alexsherstinsky: @cdkini I am not sure, but in ZEP the style has been capitalized "LOGGER" -- perhaps @Kilo59 and/or @billdirks could provide more background.  Thank you.

great_expectations/experimental/datasources/spark_datasource.py
alexsherstinsky: @cdkini Good idea -- I will create the JIRA ticket.  Thanks.

great_expectations/experimental/datasources/spark_datasource.py
alexsherstinsky: @cdkini Could you please suggest a better coding pattern?  Thank you.

alexsherstinsky: @cdkini Thank you -- I will take a look (including `Pandas`).  This code is temporary anyway.  Thanks.

great_expectations/experimental/datasources/spark_datasource.py
alexsherstinsky: @cdkini I will check (they were in the `Pandas` version).  Thanks.

tests/experimental/datasources/test_utils.py
alexsherstinsky: @cdkini I was thinking of them as utilities independent of fixtures, but I will double-check and combine if it can be done.  Thanks.

great_expectations/experimental/datasources/spark_datasource.py
alexsherstinsky: @cdkini I checked and feel that the type hint is useful here (and in `Pandas`), because of what the list contains (i.e., `pathlib.Path` is the non-obvious part here).  Thanks.

tests/experimental/datasources/test_spark_datasource.py
alexsherstinsky: @cdkini Applied (committed).  Thanks!

tests/experimental/datasources/test_spark_datasource.py
alexsherstinsky: @cdkini I am not sure whether or not using the `pyspark` module in our local (non-distributed) tests (even in Azure) qualifies as a "true" connection.  So I am inclined to keep as is.  Thanks.

great_expectations/experimental/datasources/spark_datasource.py
alexsherstinsky: @cdkini I checked.  Without these "ignore" comments, the following happened:

```
mypy great_expectations/experimental/datasources/spark_datasource.py
great_expectations/experimental/datasources/spark_datasource.py:225: error: Argument "base_directory" to "CSVSparkAsset" has incompatible type "Union[str, Path]"; expected "Path"  [arg-type]
great_expectations/experimental/datasources/spark_datasource.py:226: error: Argument "regex" to "CSVSparkAsset" has incompatible type "Union[str, Pattern[Any]]"; expected "Pattern[Any]"  [arg-type]
great_expectations/experimental/datasources/spark_datasource.py:227: error: Argument "order_by" to "CSVSparkAsset" has incompatible type "Union[List[BatchSorter], List[str], List[<nothing>]]"; expected "List[BatchSorter]"  [arg-type]
Found 3 errors in 1 file (checked 1 source file)
```

With these "ignores" in place, `mypy` ran cleanly.

Thanks.

tests/experimental/datasources/conftest.py
```diff
@@ -115,3 +123,139 @@ def inject_engine_lookup_double(monkeypatch: MonkeyPatch) -> ExecutionEngineDoub
...
+    ).resolve()
+    pandas_ds = context.sources.add_pandas(name="my_pandas")
+    asset = pandas_ds.add_csv_asset(
+        name="csv_asset",
+        base_directory=csv_path,
+        regex=r"yellow_tripdata_sample_(?P<year>\d{4})-(?P<month>\d{2})\.csv",
+        order_by=["year", "month"],
+    )
+    batch_request = asset.get_batch_request({"year": "2019", "month": "01"})
+    return context, pandas_ds, asset, batch_request
```
cdkini: Should these be fixtures by the way?

tests/experimental/datasources/conftest.py
```diff
@@ -115,3 +123,139 @@ def inject_engine_lookup_double(monkeypatch: MonkeyPatch) -> ExecutionEngineDoub
...
+        .add_year_and_month_splitter(column_name="pickup_datetime")
+        .add_sorters(["year", "month"])
+    )
+    batch_request = asset.get_batch_request({"year": 2019, "month": 1})
+    return context, datasource, asset, batch_request
+
+
+def spark_data(
+    context: AbstractDataContext,
+) -> tuple[AbstractDataContext, Datasource, DataAsset, BatchRequest]:
```
cdkini: Since we have this return type everywhere, maybe we use a namedtuple? Not super necessary but might be a nice quality-of-life improvement

tests/experimental/datasources/test_spark_datasource.py
cdkini: I don't know the intricacies of this new datasource (or ZEP in general) so I'm going to defer to Gabe and Bill - I want to say these should be integration though

tests/experimental/datasources/test_utils.py
```diff
@@ -0,0 +1,377 @@
...
+        data_assistant_result._batch_id_to_batch_identifier_display_name_map
+    )
+    assert batch_num == 12, f"Expected exactly 12 batches but found {batch_num}"
+
+    expectation_num = len(data_assistant_result.expectation_configurations)
+    if isinstance(execution_engine, (PandasExecutionEngine, SparkDFExecutionEngine)):
+        assert expectation_num == 109
+
+    if isinstance(execution_engine, SqlAlchemyExecutionEngine):
+        assert expectation_num == 111
```
cdkini: These numbers should probably be constants and defined at the top of the file?

tests/experimental/datasources/test_utils.py
cdkini: there feels like there is quite a lot of logic here - I think all of these calls will result in integration tests. Please ensure that any uses are appropriately marked

tests/experimental/datasources/test_utils.py
```diff
@@ -0,0 +1,377 @@
...
+    assert docs_dict["local_site"].startswith(
+        "file://"
+    ), "build_data_docs returns file path in unexpected form"
+    path = docs_dict["local_site"][7:]
+    with open(path) as f:
+        data_doc_index = f.read()
+
+    # Checking for ge-success-icon tests the result table was generated and it was populated with a successful run.
+    assert "ge-success-icon" in data_doc_index
+    assert "ge-failed-icon" not in data_doc_index
```
cdkini: Is all of this logic necessary? This feels like so much to assert against - what is the purpose of our tests? Can we narrow our scope at all so we don't have brittle tests?

tests/experimental/datasources/test_utils.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+def run_data_assistant_and_checkpoint(
+    datasource_test_data: tuple[
+        AbstractDataContext, Datasource, DataAsset, BatchRequest
+    ]
```
cdkini: On second thought, I'm going to push for a namedtuple here - this is difficult to reason about as a len 4 tuple in my opinion

tests/experimental/datasources/test_utils.py
alexsherstinsky: @cdkini This is not new (came from before Spark).  However, I believe that it is temporary and will be replaced with a more fully-featured and performant implementation soon.  Thanks.

tests/experimental/datasources/test_utils.py
alexsherstinsky: @cdkini Again, this is not new.  We can update this implementation in a subsequent effort -- however, it is temporary anyway and will be replaced by a fuller file_path handling implementation soon.

tests/experimental/datasources/test_spark_datasource.py
cdkini: The very requirement on connecting to Spark (even if that connection is not used) makes me think that unit is not applicable. Since we're taking a lot of care in crafting our unit tests, let's be confident in our approach here and be consistent

alexsherstinsky: @cdkini @Kilo59 @billdirks -- Fine with me either way; I would just want to appeal for consistency (if makes sense).  If this test is integration, then so should be the same tests for the other backends.  Thanks.

tests/experimental/datasources/test_utils.py
alexsherstinsky: @cdkini This is refactoring (not new code).  I will look to advice from @Kilo59 and @billdirks in terms of which tests should be marked as `@integration`.  Thanks!

tests/experimental/datasources/conftest.py
alexsherstinsky: @cdkini This function, and the appropriately-named functions for the other backends and their multi-batch counterparts, are all used in fixtures.  However, making them -- in addition -- fixtures themselves causes problems with `pytest` which is notoriously known for not accepting fixtures as parameters to other fixtures.  (I tried, and after error messages explaining the above, reverted to just declaring them as methods to be used in fixtures.)  If you know of a way, please do let me know.  Thank you.

tests/experimental/datasources/conftest.py
alexsherstinsky: @cdkini This was written in the same style as the analogous `pandas_data()` method.

tests/experimental/datasources/test_utils.py
alexsherstinsky: @cdkini We have not done that in other tests that assert on specific counts (e.g., in the `DataAssistant` tests).  Thanks.

tests/experimental/datasources/test_spark_datasource.py
Kilo59: I'm fine with changing it.
This is just an old `pylint` habit.
https://pylint.pycqa.org/en/latest/user_guide/messages/convention/invalid-name.html#invalid-name-c0103

great_expectations/experimental/datasources/spark_datasource.py
Kilo59: I'm fine with changing it but...
This is an old `pylint` habit. Pylint wants any model level constants to be `UPPER_CASE`.
>`const` - Module-level constants, any variable defined at module level that is not bound to a class object.
https://pylint.pycqa.org/en/latest/user_guide/messages/convention/invalid-name.html#invalid-name-c0103

This is also the google style recommendation, but it doesn't appear that we've been following it.
https://google.github.io/styleguide/pyguide.html#254-decision

alexsherstinsky: @Kilo59 @cdkini @billdirks -- I am happy to convert all of our `logger.` usage to `LOGGER.` in an immediately following pull request -- after the present pull request is merged with `logger.` spelled as is in this pull request's working branch right now.  Thanks.

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -123,11 +123,11 @@ def _replace_builtins(input_: str | type) -> str | type:
 
 FIELD_SUBSTITUTIONS: Final[Dict[str, Dict[str, _FieldSpec]]] = {
     # CSVAsset
-    "filepath_or_buffer": {"path": _FieldSpec(pathlib.Path, ...)},
+    "filepath_or_buffer": {"base_directory": _FieldSpec(pathlib.Path, ...)},
```
billdirks, line 126: Tagging @Kilo59 so he knows this is changing.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -240,14 +242,14 @@ def add_csv_asset(
 
         Args:
             name: The name of the csv asset
-            data_path: Path to directory with csv files
+            base_directory: base directory path, relative to which CSV file paths will be collected
```
billdirks, line 279: nit: remove 'relative' so it is not confused with 'relative path'. That is change it to something like: `The path to the directory where the csv files will be collected.`

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -316,13 +320,14 @@ def add_parquet_asset(
...
             regex: regex pattern that matches csv filenames that is used to label the batches
+            order_by: one of "asc" (ascending) or "desc" (descending) -- the method by which to sort "Asset" parts.
```
billdirks: This `order_by` description doesn't match with the type, `Optional[BatchSortersDefinition]`, above.

great_expectations/experimental/datasources/spark_datasource.py
billdirks: +1 remove this comment and file a jira ticket instead.

great_expectations/experimental/datasources/spark_datasource.py
```diff
@@ -0,0 +1,229 @@
...
+        Args:
+            batch_request: A batch request
+
+        Returns:
+            A list of pairs (batch_request, path) where 'batch_request' is a fully specified
+            batch request and 'path' is the path to the corresponding file on disk.
+            This list will be empty if no files exist on disk that correspond to the input
+            batch request.
+        """
+        option_to_group_id = self._option_name_to_regex_group_id()
```
billdirks, line 79: The file system, regex matching code is datasource agnostic so it feels like it should be moved to some shared module that the pandas and pyspark implementations can reuse. I think I'll like that to be down now, but we could wait until we need to do edits in both places.

great_expectations/experimental/datasources/spark_datasource.py
```diff
@@ -0,0 +1,229 @@
...
+    def get_batch_list_from_batch_request(
+        self, batch_request: BatchRequest
+    ) -> List[Batch]:
+        self._validate_batch_request(batch_request)
+        batch_list: List[Batch] = []
+
+        for request, path in self._fully_specified_batch_requests_with_path(
+            batch_request
+        ):
+            # TODO: <Alex>ALEX_NOTE-1/27/2023: Adding "reader_options" is temporary work-around to enable integration tests to pass.  Ultimately, this layer will be replaced with fully-functional solution.</Alex>
```
billdirks: Could you file a JIRA and delete this comment? Why is this required now? What does a fully-functional solution do that this doesn't? ie what are the acceptance criteria for a fix.

great_expectations/experimental/datasources/spark_datasource.py
```diff
@@ -0,0 +1,229 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class SparkDatasourceError(Exception):
+    pass
+
+
+# TODO: <Alex>ALEX_NOTE-1/27/2023: We did not add "_pandas" suffix (nor qualified class name) for CSVSparkAsset in PandasDatasource module (we might need to do that at a later date, or come up with a different, more general, solution altogether).</Alex>
+class CSVSparkAsset(DataAsset):
```
billdirks, line 35: This whole class is a lot of copy and paste from the pandas code. Could we refactor this functionality into some shared code that gets called from this class.

great_expectations/experimental/datasources/spark_datasource.py
```diff
@@ -0,0 +1,229 @@
...
+        regex: Union[str, re.Pattern],
+        order_by: Optional[BatchSortersDefinition] = None,
+    ) -> CSVSparkAsset:
+        """Adds a csv asset to this Spark datasource
+
+        Args:
+            name: The name of the csv asset
+            base_directory: base directory path, relative to which CSV file paths will be collected
+            regex: regex pattern that matches csv filenames that is used to label the batches
+            order_by: one of "asc" (ascending) or "desc" (descending) -- the method by which to sort "Asset" parts.
```
billdirks: This doesn't appear to be correct. It looks like it should be a list of fields that one wants to sort by with an optional prefix of `+` or `-` if one wants to specify sort direction.

great_expectations/experimental/datasources/spark_datasource.py
billdirks: This has to do with pydantic run time type coercion that mypy doesn't know about.

tests/experimental/datasources/conftest.py
```diff
@@ -1,26 +1,34 @@
+from __future__ import annotations
```
billdirks, line 1: I'm curious why you broke these out of the `test_integration_datasources.py` file. These functions are about connecting to real datasources and should only be called integration tests. We should actively discourage there use elsewhere.
If we want to break up integration tests maybe we should have an `integration` directory and put integration tests there and shared utilities under that directory.

tests/experimental/datasources/test_spark_datasource.py
```diff
@@ -0,0 +1,278 @@
...
+def spark_datasource(test_backends) -> SparkDatasource:
+    if "spark" not in test_backends:
+        pytest.skip("No spark backend selected.")
+
+    return SparkDatasource(name="spark_datasource")
+
+
+@pytest.fixture
+def csv_path() -> pathlib.Path:
+    return pathlib.Path(
```
billdirks: Could you get rid of the `file_relative_path` dependency and change this line to:
```
return pathlib.Path(__file__) / "../../../test_sets/taxi_yellow_tripdata_samples"
```

tests/experimental/datasources/test_utils.py
billdirks: Both the code here and in the new `conftest.py` came from `test_integration_datasource.py`. All the functionality should only be used in integration tests and we should discourage its use elsewhere. If that file was getting to big, please see my comment there for a strategy for doing this.

tests/experimental/datasources/test_spark_datasource.py
billdirks: I don't believe any of these tests actually need the spark backend. We might instantiate one when the datasource comes up but the tests check the `Batch` (a dataclass of data describing the data in the batch) which might involve listing the csv files in a directory. We could mock the spark execution engine to more strongly separate this from using actual spark. For SQL tests, we do mock out the sql execution engine. For pandas everything is in memory and there is no external service to mock out. I believe you can use `execution_engine_override` as a hook to use a fake execution engine.

tests/experimental/datasources/test_utils.py
billdirks: This should be our e2e test so the goal here is to test that we can go from a zep datasource and generate datadocs and they look correct (ie we see no expectations failing, metrics look correct, the datadocs look right, etc). It is a lot of stuff but I wanted something that was comprehensive since this was a new entry point to this workflow.
The goal was to have 1 e2e test for each major workflow (computing expectations, checkpoints, data assistants) and then lots of unit tests for particular functionality.

billdirks (commented):
Thanks for doing this Alex! Overall it looks fine. There is some code copy/pasted which is ok but it may be nice to find a shared location for code shared between pandas and spark. The regex matching and filesystem code seems datasource agnostic.
I also left some questions about tests. I would like to make it more obvious which code should be used in integration tests only. Since you are breaking up the integration test code, I have a suggestion in `conftest.py` about organization.

great_expectations/experimental/datasources/spark_datasource.py
alexsherstinsky: @cdkini This has been done.  Thanks.

great_expectations/experimental/datasources/interfaces.py
alexsherstinsky: @cdkini This has been done.  Thanks.

great_expectations/experimental/datasources/dynamic_pandas.py
alexsherstinsky: @billdirks We did indeed discuss this during our sync up (after the prioritization meeting) and are in agreement.  Thanks.

great_expectations/experimental/datasources/spark_datasource.py
alexsherstinsky: @billdirks The JIRA ticket has been filed; the comment will be deleted (in next commit).  Thanks.

great_expectations/experimental/datasources/pandas_datasource.py
alexsherstinsky: @billdirks Good point -- fixing (will be in next commit).  Thanks.

great_expectations/experimental/datasources/spark_datasource.py
alexsherstinsky: @billdirks This has been done.

great_expectations/experimental/datasources/spark_datasource.py
alexsherstinsky: @billdirks I am not comfortable doing it yet; perhaps soon (when moving towards build file-path data assets as I had attempted in #6928).  Thanks.

great_expectations/experimental/datasources/spark_datasource.py
alexsherstinsky: @billdirks Good point; I fixed it (in next commit).  Thanks.

tests/experimental/datasources/test_spark_datasource.py
alexsherstinsky: @billdirks Let us discuss -- it seems that `file_relative_path()` is a useful general utility that accomplishes this.  Thanks.

tests/experimental/datasources/conftest.py
billdirks: @alexsherstinsky This still needs to be moved to the integration folder.

tests/experimental/datasources/tasks.py
```diff
@@ -9,7 +9,14 @@
 import invoke
 import pandas as pd
 
-TEST_ROOT = pathlib.Path(__file__).parent.parent.parent.resolve(strict=True)
+from great_expectations.data_context.util import file_relative_path
+
+TEST_ROOT = pathlib.Path(
```
billdirks: Can you revert this back? One thing I really don't like in our codebase is the transformation of objects into strings and then back into objects again. If we want a Path object we should avoid turning it into a string unless that is needed by a third party lib. Alternatively you coud update `file_relative_path` to return a `Path` object but that is probably better done outside of this PR.

tests/experimental/datasources/test_config.py
```diff
@@ -23,7 +24,12 @@
 
 p = pytest.param
 
-EXPERIMENTAL_DATASOURCE_TEST_DIR = pathlib.Path(__file__).parent
+EXPERIMENTAL_DATASOURCE_TEST_DIR = pathlib.Path(
```
billdirks: Similar comment here as above. We should avoid turning things into strings.

tests/experimental/datasources/tasks.py
alexsherstinsky: @billdirks No, I should not modify `file_relative_path()` -- so I will revert the change -- that is more proper.  One moment.

tests/experimental/datasources/test_config.py
alexsherstinsky: @billdirks Reverting (commit coming soon).  Thanks.

tests/experimental/datasources/conftest.py
alexsherstinsky: @billdirks This has been done and a higher level "conftest.py" created (containing the fixtures for the unit tests that need them).  Thanks.

tests/experimental/datasources/test_spark_datasource.py
alexsherstinsky: @billdirks This has been reverted and standardized in an elegant, consistent way (now I can clearly see that it is indeed amenable to a utility function that would stop short of outputting a string -- for another time).  Thanks.

tests/experimental/datasources/tasks.py
alexsherstinsky: @billdirks This has been taken care of.  Thanks.

tests/experimental/datasources/test_config.py
alexsherstinsky: @billdirks This has been taken care of. Thanks.

billdirks (approved):
Thanks @alexsherstinsky!

Please do the follow up items we discussed after this PR merges:
* Coordinate with @NathanFarmer to make sure the changes from his PR apply to this.
* Look at mocking the pyspark engine in the unit tests so we can run them without pyspark installed.
* logger -> LOGGER (this wasn't in a discussion with me but was in a slack thread)

great_expectations/experimental/datasources/spark_datasource.py
Kilo59: That seems like unnecessary churn that's going to lead to conflicts. IMO it would be better to drop/revert the changes to `LOGGER`.

alexsherstinsky: @Kilo59 @billdirks @cdkini While we previously had a motion to replace `logger` (all lowercase) with `LOGGER` (all uppercase) and even though doing so has not caused any checks to fail, not everybody is on-board with this change at the present time.  Hence, per the recommendation by @Kilo59 I am reverting it (we can always apply it at a later date, should we rationalize that this aspect of adhering to the Google Python style is beneficial to GX OSS.  Thank you. /cc @donaldheppner @5u2ie

PR #6999, opened 2023-01-31
[FEATURE] Add SQL query data asset for new experimental datasources

billdirks:

### DIFF
--- a/great_expectations/execution_engine/sqlalchemy_execution_engine.py
+++ b/great_expectations/execution_engine/sqlalchemy_execution_engine.py
@@ -95,7 +95,7 @@ except ImportError:
 try:
     from sqlalchemy.engine import Dialect, Row
     from sqlalchemy.exc import OperationalError
-    from sqlalchemy.sql import Selectable, Subquery
+    from sqlalchemy.sql import Selectable
     from sqlalchemy.sql.elements import (
         BooleanClauseList,
         Label,
--- a/great_expectations/execution_engine/sqlalchemy_execution_engine.py
+++ b/great_expectations/execution_engine/sqlalchemy_execution_engine.py
@@ -95,7 +95,7 @@ except ImportError:
 try:
     from sqlalchemy.engine import Dialect, Row
     from sqlalchemy.exc import OperationalError
-    from sqlalchemy.sql import Selectable
+    from sqlalchemy.sql import Selectable, Subquery
     from sqlalchemy.sql.elements import (
         BooleanClauseList,
         Label,
@@ -1147,6 +1147,15 @@ class SqlAlchemyExecutionEngine(ExecutionEngine):
     def _build_selectable_from_batch_spec(
         self, batch_spec: BatchSpec
     ) -> Union[Selectable, str]:
+        if (
+            batch_spec.get("query") is not None
+            and batch_spec.get("sampling_method") is not None
+        ):
+            raise ValueError(
+                "Sampling is not supported on query data. "
+                "It is currently only supported on table data."
+            )
+
         if "splitter_method" in batch_spec:
             splitter_fn: Callable = self._get_splitter_method(
                 splitter_method_name=batch_spec["splitter_method"]
@@ -1162,7 +1171,7 @@ class SqlAlchemyExecutionEngine(ExecutionEngine):
             else:
                 split_clause = sa.true()
 
-        table_name: str = batch_spec["table_name"]
+        selectable: Selectable = self._subselectable(batch_spec)
         sampling_method: Optional[str] = batch_spec.get("sampling_method")
         if sampling_method is not None:
             if sampling_method in [
@@ -1181,9 +1190,7 @@ class SqlAlchemyExecutionEngine(ExecutionEngine):
                 sampler_fn = self._data_sampler.get_sampler_method(sampling_method)
                 return (
                     sa.select("*")
-                    .select_from(
-                        sa.table(table_name, schema=batch_spec.get("schema_name", None))
-                    )
+                    .select_from(selectable)
                     .where(
                         sa.and_(
                             split_clause,
@@ -1192,13 +1199,23 @@ class SqlAlchemyExecutionEngine(ExecutionEngine):
                     )
                 )
 
-        return (
-            sa.select("*")
-            .select_from(
-                sa.table(table_name, schema=batch_spec.get("schema_name", None))
+        return sa.select("*").select_from(selectable).where(split_clause)
+
+    def _subselectable(self, batch_spec: BatchSpec) -> Selectable:
+        table_name = batch_spec.get("table_name")
+        query = batch_spec.get("query")
+        selectable: Selectable
+        if table_name:
+            selectable = sa.table(
+                table_name, schema=batch_spec.get("schema_name", None)
             )
-            .where(split_clause)
-        )
+        else:
+            if not isinstance(query, str):
+                raise ValueError(f"SQL query should be a str but got {query}")
+            # Query is a valid SELECT query that begins with r"\w+select\w"
+            selectable = sa.select(sa.text(query.lstrip()[6:].lstrip())).subquery()
+
+        return selectable
 
     def get_batch_data_and_markers(
         self, batch_spec: BatchSpec
@@ -1210,6 +1227,18 @@ class SqlAlchemyExecutionEngine(ExecutionEngine):
                 f"""SqlAlchemyExecutionEngine accepts batch_spec only of type SqlAlchemyDatasourceBatchSpec or
         RuntimeQueryBatchSpec (illegal type "{str(type(batch_spec))}" was received).
                         """
+            )
+        if (
+            sum(
+                1 if x else 0
+                for x in [batch_spec.get("query"), batch_spec.get("table_name")]
+            )
+            != 1
+        ):
+            raise InvalidBatchSpecError(
+                "SqlAlchemyExecutionEngine only accepts a batch_spec where exactly 1 of "
+                "'query' or 'table_name' is specified. "
+                f"table_name={batch_spec.get('table_name')}, query={batch_spec.get('query')}"
             )
 
         batch_data: Optional[SqlAlchemyBatchData] = None
--- a/great_expectations/experimental/datasources/sql_datasource.py
+++ b/great_expectations/experimental/datasources/sql_datasource.py
@@ -5,6 +5,7 @@ import itertools
 from datetime import datetime
 from typing import (
     TYPE_CHECKING,
+    Any,
     Callable,
     Dict,
     List,
@@ -17,6 +18,7 @@ from typing import (
 
 import pydantic
 from pydantic import dataclasses as pydantic_dc
+from sqlalchemy import func
 from typing_extensions import ClassVar, Literal
 
 from great_expectations.core.batch_spec import SqlAlchemyDatasourceBatchSpec
@@ -56,7 +58,7 @@ class ColumnSplitter:
     method_name: str
     param_names: Sequence[str]
 
-    def param_defaults(self, table_asset: TableAsset) -> Dict[str, List]:
+    def param_defaults(self, sql_asset: _SQLAsset) -> Dict[str, List]:
         raise NotImplementedError
 
     @pydantic.validator("method_name")
@@ -91,33 +93,34 @@ class SqlYearMonthSplitter(ColumnSplitter):
         default_factory=lambda: ["year", "month"]
     )
 
-    def param_defaults(self, table_asset: TableAsset) -> Dict[str, List]:
+    def param_defaults(self, sql_asset: _SQLAsset) -> Dict[str, List]:
         """Query sql database to get the years and months to split over.
 
         Args:
-            table_asset: A TableAsset over which we want to split the data.
+            sql_asset: A SQL DataAsset over which we want to split the data.
         """
         return _query_for_year_and_month(
-            table_asset, self.column_name, _get_sql_datetime_range
+            sql_asset, self.column_name, _get_sql_datetime_range
         )
 
 
 def _query_for_year_and_month(
-    table_asset: TableAsset,
+    sql_asset: _SQLAsset,
     column_name: str,
     query_datetime_range: Callable[
-        [sqlalchemy.engine.base.Connection, str, str], DatetimeRange
+        [sqlalchemy.engine.base.Connection, sqlalchemy.sql.Selectable, str],
+        DatetimeRange,
     ],
 ) -> Dict[str, List]:
     from great_expectations.execution_engine import SqlAlchemyExecutionEngine
 
-    execution_engine = table_asset.datasource.get_execution_engine()
+    execution_engine = sql_asset.datasource.get_execution_engine()
     assert isinstance(execution_engine, SqlAlchemyExecutionEngine)
 
     with execution_engine.engine.connect() as conn:
         datetimes: DatetimeRange = query_datetime_range(
             conn,
-            table_asset.table_name,
+            sql_asset.as_selectable(),
             column_name,
         )
     year: List[int] = list(range(datetimes.min.year, datetimes.max.year + 1))
@@ -130,17 +133,25 @@ def _query_for_year_and_month(
 
 
 def _get_sql_datetime_range(
-    conn: sqlalchemy.engine.base.Connection, table_name: str, col_name: str
+    conn: sqlalchemy.engine.base.Connection,
+    selectable: sqlalchemy.sql.Selectable,
+    col_name: str,
 ) -> DatetimeRange:
-    q = f"select min({col_name}), max({col_name}) from {table_name}"
-    min_max_dt = list(conn.execute(q))[0]
+    import sqlalchemy as sa
+
+    column = sa.column(col_name)
+    query = sa.select([func.min(col_name), func.max(column)]).select_from(selectable)
+    min_max_dt = list(conn.execute(query))[0]
+    if min_max_dt[0] is None or min_max_dt[1] is None:
+        raise SQLDatasourceError(
+            f"Data date range can not be determined for the query: {query}. The returned range was {min_max_dt}."
+        )
     return DatetimeRange(min=min_max_dt[0], max=min_max_dt[1])
 
 
-class TableAsset(DataAsset):
+class _SQLAsset(DataAsset):
     # Instance fields
-    type: Literal["table"] = "table"
-    table_name: str
+    type: Literal["sqlasset"] = "sqlasset"
     column_splitter: Optional[SqlYearMonthSplitter] = None
     name: str
 
@@ -186,14 +197,14 @@ class TableAsset(DataAsset):
     def add_year_and_month_splitter(
         self,
         column_name: str,
-    ) -> TableAsset:
+    ) -> _SQLAsset:
         """Associates a year month splitter with this DataAsset
 
         Args:
             column_name: A column name of the date column where year and month will be parsed out.
 
         Returns:
-            This TableAsset so we can use this method fluently.
+            This DataAsset so we can use this method fluently.
         """
         self.column_splitter = SqlYearMonthSplitter(
             column_name=column_name,
@@ -272,12 +283,7 @@ class TableAsset(DataAsset):
         column_splitter = self.column_splitter
         for request in self._fully_specified_batch_requests(batch_request):
             batch_metadata = copy.deepcopy(request.options)
-            batch_spec_kwargs = {
-                "type": "table",
-                "data_asset_name": self.name,
-                "table_name": self.table_name,
-                "batch_identifiers": {},
-            }
+            batch_spec_kwargs = self._create_batch_spec_kwargs()
             if column_splitter:
                 batch_spec_kwargs["splitter_method"] = column_splitter.method_name
                 batch_spec_kwargs["splitter_kwargs"] = {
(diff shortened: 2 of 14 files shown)


great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1147,6 +1147,15 @@ def get_data_for_batch_identifiers(
     def _build_selectable_from_batch_spec(
         self, batch_spec: BatchSpec
     ) -> Union[Selectable, str]:
+        if (
+            batch_spec.get("query") is not None
+            and batch_spec.get("sampling_method") is not None
+        ):
+            raise ValueError(
```
billdirks, line 1154: I am going to file a ticket to add sampling support to new experimental datasources. We should be able to use the same strategy/helper method I use here to add query support.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -121,18 +120,27 @@ def _query_for_year_and_month(
...
+SQL_DATETIME_COUNTER = 0
+
+
 def _get_sql_datetime_range(
-    conn: sqlalchemy.engine.base.Connection, table_name: str, col_name: str
+    conn: sqlalchemy.engine.base.Connection, data_from: str, col_name: str
 ) -> DatetimeRange:
-    q = f"select min({col_name}), max({col_name}) from {table_name}"
+    global SQL_DATETIME_COUNTER
+    from_name = f"_gx_sql_datetime_range_data_{SQL_DATETIME_COUNTER}"
```
billdirks: I use the `SQL_DATETIME_COUNTER` to make sure every subquery alias is unique. I could use a generated uuid in the function itself. Let me know if you prefer this or a different strategy.

great_expectations/experimental/datasources/sqlite_datasource.py
```diff
@@ -27,19 +27,37 @@
...
+    QueryAsset,
+    SQLAsset,
     SQLDatasource,
     SQLDatasourceError,
     TableAsset,
     _query_for_year_and_month,
 )
 
 
+def _add_year_and_month_splitter(
```
billdirks: In addition to this private helper method, I tried was defining a mixin class that only defined the `add_year_and_month_splitter` that could be mixed in with the `SqliteTableAsset` and `SqliteQueryAsset`. This let me sidestep defining `add_year_and_month_splitter` on these classes. I ran into issues the with type of `self` unless I made the mixin class a subclass of `SQLAsset` but it felt strange mixing that into `SqliteQueryAsset` which is derived from `QueryAsset`.  It also made the return type of this function a generic `SQLAsset` since the mixin wouldn't know about `[Table|Query]Asset`.

The main issue is the inheritance tree looks like:

```
SQLAsset -> TableAsset -> SqliteTableAsset
                 |
                 |-> QueryAsset -> SqliteQueryAsset
```

So the `Sqlite*Asset`s earliest common ancestor is `SQLAsset`. The mixin would let me spread the functionality across the hierarchy without worrying about the structure of this inheritance tree with the drawback of losing some type specificity.

I could play with that again if that is preferred. This way is a bit more verbose but is also more specific so I decided to go with that for now.

billdirks: It just occurred to me I could probably dynamically generate the `Sqlite*Asset` subclasses to reduce the boilerplate.
I don't want sqlite to block this PR though since the main value comes from the `SQLDatasource` changes so unless someone feels strongly or there are a lot of other changes required I may save this more a followup PR to prep for adding more splitters.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1192,13 +1199,23 @@ def _build_selectable_from_batch_spec(
...
+            selectable = sa.table(
+                table_name, schema=batch_spec.get("schema_name", None)
             )
-            .where(split_clause)
-        )
+        else:
+            if not isinstance(query, str):
+                raise ValueError(f"SQL query should be a str but got {query}")
+            # Query is a valid SELECT query that begins with r"select\w"
+            selectable = sa.select(sa.text(query[6:].lstrip()))
```
alexsherstinsky: @billdirks Would it make sense to utilize some kind of an SQL query parser into tokens -- or something equivalent on our part -- to allow for a flexible number `\s`, `\t`, `\r`, `\n`, etc. characters in the query string?  Thanks.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1181,9 +1190,7 @@ def _build_selectable_from_batch_spec(
                 sampler_fn = self._data_sampler.get_sampler_method(sampling_method)
                 return (
                     sa.select("*")
-                    .select_from(
-                        sa.table(table_name, schema=batch_spec.get("schema_name", None))
-                    )
+                    .select_from(subselectable)
```
alexsherstinsky: @billdirks While not part of this effort, it would be great to investigate the feasibility of using the formal table columns in the `sa.select("*")` statement (analogous to the way it is done as part of introspection functions in "great_expectations/expectations/metrics/util.py") -- the advantage is that this will provide objects for columns, which could then be utilized for other operations (e.g., adding labels, etc.), which are more difficult (if not outright impossible) when using the string `"*"`.  Thanks.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1147,6 +1147,15 @@ def get_data_for_batch_identifiers(
     def _build_selectable_from_batch_spec(
         self, batch_spec: BatchSpec
     ) -> Union[Selectable, str]:
+        if (
```
alexsherstinsky, line 1150: @billdirks -- ❤️ For a simple and elegant way of bringing back the `SQL` query as a legitimate way to specify a selectable!

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1211,6 +1228,14 @@ def get_batch_data_and_markers(
         RuntimeQueryBatchSpec (illegal type "{str(type(batch_spec))}" was received).
                         """
             )
+        has_query = 1 if batch_spec.get("query") else 0
```
alexsherstinsky: @billdirks While equivalent, I think that the following style, used in other parts of the codebase, is more elegant in the sense that it always uses booleans to compute the mutual exclusion as a boolean (i.e., without the integers):

```
        if (
            sum(
                bool(x)
                for x in [
                    batch is not None,
                    batch_list is not None,
                    batch_request is not None,
                    batch_request_list is not None,
                ]
            )
            > 1 
        ):
            raise ValueError(
                "No more than one of batch, batch_list, batch_request, or batch_request_list can be specified"
            )
```

Your call.  Thanks for considering.

great_expectations/experimental/datasources/sql_datasource.py
alexsherstinsky: @billdirks In my opinion, something that ends on `_{str(uuid.uuid4())[:8]}` might be slightly more elegant.  We have a function in `great_expectations/util.py` called `generate_temporary_table_name()`; one can create a better-named one here by calling `generate_temporary_table_name()` with `default_table_name_prefix=""` and leaving the `num_digits` same.  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -121,18 +120,27 @@ def _query_for_year_and_month(
...
+    SQL_DATETIME_COUNTER += 1
+    q = f"select min({col_name}), max({col_name}) from ({data_from}) as {from_name}"
```
alexsherstinsky: @billdirks Should we not be building a formal `SQLAlchemy` query object for the given DBMS backend, and then utilizing something like the following:

```
            query: str = str(
                selectable.compile(
                    dialect=execution_engine.dialect,
                    compile_kwargs={"literal_binds": True},
                )
            )
```

in order to obtain the string-valued query version for printing?  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -295,6 +298,72 @@ def get_batch_list_from_batch_request(
...
+
+
+class QueryAsset(SQLAsset):
+    # Instance fields
+    type: Literal["query"] = "query"  # type: ignore[assignment]
+    query: str
+
+    @pydantic.validator("query")
+    def query_must_start_with_select(cls, v: str):
+        if not (v.upper().startswith("SELECT") and v[6].isspace()):
```
alexsherstinsky: @billdirks I think that this should be flexible in terms of blank spaces, tabs, etc. (please see an earlier comment).  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -295,6 +298,72 @@ def get_batch_list_from_batch_request(
...
+    def data_from(self) -> str:
+        """Returns the from clause that would be used to query this data
+
+        Returns:
+            The from clause which might be a table name or a query.
+        """
+        raise NotImplementedError
+
+
+class QueryAsset(SQLAsset):
```
alexsherstinsky: @billdirks -- I ❤️ the idea of the `SQLDataAsset` defining an interface for query, table, etc. to implement their specifics for!  Thanks!

tests/experimental/datasources/test_integration_datasource.py
```diff
@@ -42,17 +43,24 @@
...
+        / f"../../../test_sets/taxi_yellow_tripdata_samples/sqlite/{db_filename}"
     ).resolve()
     datasource = context.sources.add_sqlite(
         name="test_datasource",
         connection_string=f"sqlite:///{db_file}",
     )
+    return datasource
+
+
+def sql_data(
```
alexsherstinsky: @billdirks In #6981 `sql_data` is moved to a util module (the merge should not be too difficult).

tests/experimental/datasources/test_integration_datasource.py
```diff
@@ -257,13 +265,8 @@ def test_run_data_assistant_and_checkpoint(datasource_test_data):
...
     context: AbstractDataContext,
 ) -> tuple[AbstractDataContext, Datasource, DataAsset, BatchRequest]:
-    db_file = (
-        pathlib.Path(__file__)
-        / "../../../test_sets/taxi_yellow_tripdata_samples/sqlite/yellow_tripdata_sample_2020_all_months_combined.db"
-    ).resolve()
-    datasource = context.sources.add_sqlite(
-        name="test_datasource",
-        connection_string=f"sqlite:///{db_file}",
+    datasource = sqlite_datasource(
```
alexsherstinsky: @billdirks In #6981 `multibatch_sql_data` is moved to a util module (the merge should not be too difficult).

tests/experimental/datasources/test_integration_datasource.py
```diff
@@ -515,3 +518,26 @@ def test_batch_head(
...
+
+
+def test_sql_query_data_asset(empty_data_context):
+    context = empty_data_context
+    datasource = sqlite_datasource(context, "yellow_tripdata.db")
+    passenger_count_value = 5
+    asset = (
+        datasource.add_query_asset(
+            name="query_asset",
+            query=f"SELECT * from yellow_tripdata_sample_2019_02 WHERE passenger_count = {passenger_count_value}",
```
alexsherstinsky: @billdirks Per earlier comments, I would want to be able to include a free amount of blank spaces in the `SQL` query specification.  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -102,14 +103,12 @@ def _query_for_year_and_month(
...
-        table_asset.datasource.execution_engine, SqlAlchemyExecutionEngine
-    )
+    assert isinstance(sql_asset.datasource.execution_engine, SqlAlchemyExecutionEngine)
 
-    with table_asset.datasource.execution_engine.engine.connect() as conn:
+    with sql_asset.datasource.execution_engine.engine.connect() as conn:
         datetimes: DatetimeRange = query_datetime_range(
             conn,
-            table_asset.table_name,
+            sql_asset.data_from(),
```
alexsherstinsky: @billdirks While not part of this change, the calculated variable names `year` and `month` are misleading.  They should be `years` and `months` based on the fact that they represent ranges.  Thank you.

great_expectations/experimental/datasources/sqlite_datasource.py
alexsherstinsky: @billdirks I would be interested in better understanding this explanation, even if the implementation is not changed.  One thought that came to me was that in order to avoid the same code that seemingly appears in the parent `SQLAsset` class for `add_year_and_month_splitter()` would be to call `super()` here and instantiate the output field by field with the present class (similarly to the way it is done in `DataAssistant` implementations for returning the corresponding `DataAssistantResult` type (something like this technique I thought. might help here).  Or could something as simple as `cast()` work here (it seems that the output is always the same and it is just a matter of returning the object with the correct type)?  Thanks.

alexsherstinsky (changes requested):
@billdirks This is great stuff; I am requesting only that my recommendations are considered and questions answered.  In other words, no particular changes are required, unless you agree with the suggestions and/or any JIRA tickets for future enhancements are created.  And, of course, all required tests in Azure must pass.  Then I will be happy to approve.  Thank you!

great_expectations/execution_engine/sqlalchemy_execution_engine.py
billdirks: I can look more into if it's possible to do this. I would like to know also. One possible difficulty is we are selecting from an arbitrary query and there is no sql table or sqlalchemy model defined. We will get a dictionary returned with the key being the column name currently.
In past experience with django, I often found the dict preferable because, object creation there was orders of magnitude slower and with a biggish data set this really mattered. Don't know the situation here though.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
billdirks: Added an lstrip to remove any initial whitespace characters.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
billdirks: Done.
I didn't use `bool` since I prefer to sum `int`s but otherwise similar.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: I can update this is a followup PR if you don't mind since this isn't part of this change and will involve some plumbing throughout this file.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Done

great_expectations/experimental/datasources/sql_datasource.py

tests/experimental/datasources/test_integration_datasource.py
billdirks: Added prepended spaces.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Can you provide a code snippet and/or point me to docs about how to create a formal `SQLALchemy` object with a subselect in the from clause?

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -121,18 +120,27 @@ def _query_for_year_and_month(
...
     min_max_dt = list(conn.execute(q))[0]
+    if min_max_dt[0] is None or min_max_dt[1] is None:
+        raise SQLDatasourceError(
+            f"Data date range can not be determined for the query: {q}. The returned range was {min_max_dt}."
+        )
     return DatetimeRange(min=min_max_dt[0], max=min_max_dt[1])
 
 
-class TableAsset(DataAsset):
+class SQLAsset(DataAsset):
```
Kilo59: 1. Is the class meant to be instantiated?
2. Will a user ever have a config file with `type: sqlasset`?

If the answer is no, then it probably doesn't need to be registered by the `MetaDatasource`.
You could make this `_SQLAsset(DataAsset)`: and this registration step would be skipped.

https://github.com/great-expectations/great_expectations/blob/1e91b803778309a4d18c27fadb156148b7109760/great_expectations/experimental/datasources/sources.py#L127-L132

However, it might be better to be more explicit about this and have some `ClassVar` `_SKIP_ASSET_REGISTRATION` attribute that gets checked. But that doesn't need to be done in this PR.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
alexsherstinsky: @billdirks All I can think about is if the example(s) from `great_expectations/expectations/metrics/util.py:column_reflection_fallback()` and/or from `great_expectations/expectations/metrics/multicolumn_map_metrics/compound_columns_unique.py` and/or `great_expectations/execution_engine/sqlalchemy_execution_engine.py::ExecutionEngine.get_domain_records()` apply, then there may be a way; otherwise, we keep the `sa.select("*")` -- actually it looks like we resort to `sa.select([sa.text("*")])` as the catch-all in the latter.  So, perhaps worth a quick check, just in case.  Thanks!

great_expectations/experimental/datasources/sql_datasource.py
alexsherstinsky: @billdirks Examples that come to mind are:
* In `great_expectations/expectations/metrics/map_metric_provider.py`: `_sqlalchemy_map_condition_unexpected_count_value()` and `_sqlalchemy_map_condition_unexpected_count_aggregate_fn()` (among others).
* In `great_expectations/expectations/metrics/column_aggregate_metrics/column_quantile_values.py`, methods called by `_sqlalchemy()`.

Happy to discuss.  Thank you!

great_expectations/experimental/datasources/sql_datasource.py
alexsherstinsky: 🙇

tests/experimental/datasources/test_integration_datasource.py
alexsherstinsky: 🙇

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Good call out. I had sketched this before I saw you exclude the `_` prefix assets and have added it.

great_expectations/experimental/datasources/sqlite_datasource.py
billdirks: Yes the issue is with typing. I want to add a common function to the 2 sqlite data assets but they don't share a direct superclass. I also want mypy to throw an error when this function is used incorrectly. I really don't like the duplication though. Instead of a `cast`, I've updated this to use a mixin where I can make an assert about types so we at least get a runtime check with a good error message.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
billdirks: Those examples don't work here and they fall back to `select(*)` for similar use cases.
I think it would be hard to do because we would need to parse the subselect query, which could have a `*` in it, and then map the selected columns back to their origin to infer types.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Thank you! This is done.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1211,6 +1228,18 @@ def get_batch_data_and_markers(
         RuntimeQueryBatchSpec (illegal type "{str(type(batch_spec))}" was received).
                         """
             )
+        if (
+            sum(
```
alexsherstinsky, line 1232: Nice "XOR"! :)

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -130,43 +134,35 @@ def _query_for_year_and_month(
...
-    q = f"select min({col_name}), max({col_name}) from {table_name}"
-    min_max_dt = list(conn.execute(q))[0]
+    import sqlalchemy as sa
+
+    column = sa.column(col_name)
+    query = sa.select([func.min(col_name), func.max(column)]).select_from(selectable)
+    min_max_dt = list(conn.execute(query))[0]
+    if min_max_dt[0] is None or min_max_dt[1] is None:
+        raise SQLDatasourceError(
+            f"Data date range can not be determined for the query: {query}. The returned range was {min_max_dt}."
```
alexsherstinsky, line 151: @billdirks This is great now how it is all `SQLAlchemy` object form.  One last thing on this, please: I believe that it is safer to explicitly convert `query` into `str`, for example:
```
query_string = query.compile(compile_kwargs={"literal_binds": True})
```

Thank you.

alexsherstinsky (changes requested):
@billdirks Thank you for this -- really good stuff.  Here are a couple of remaining things:
* Static Type Checks are currently failing (probably easy to address);
* I made a suggestion to explicitly compile the query into a string (I feel that this is important);
* Optional: make one of the SQL query tests include some blank spaces/tabs around "SELECT" (sorry if you already added one, and I missed it in the code). 😄 

Thank you!

Kilo59 (approved):

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Done. Could you explain why you think calling compile directly is better?

billdirks (commented):
@alexsherstinsky Thanks for your comments. This is ready for re-review.

* static checking now succeeds. It was passing locally so must be a mypy version issue or something related.
* I now compile the sql query. I'm wondering what makes this safer? I googled around but couldn't find any conversations about this.
* There is one: https://github.com/great-expectations/great_expectations/pull/6999/files#diff-b97e96dba34432dc6b4f3bb19ce37d8e23693cbbcf5a6e244f4803bc494fa2c9R100

great_expectations/experimental/datasources/sql_datasource.py
alexsherstinsky: As per earlier versions of `SQLAlchemy`, it has been utilized as a more explicit way to control what happens in "to string" operation.  It is certainly worth seeing how this approach has changed as the newer `SQLAlchemy` versions have been including more and more improvements.  Thanks.

alexsherstinsky (approved):
Approving provisionally -- for when all the test failures have been resolved.  Thanks!

PR #7009, opened 2023-01-31
[BUGFIX ] ZEP - pandas serde fix

Kilo59:
Follow-up to #6780 

- [x] change default serialization method behavior to `exclude_unset` fields by default
    * modify `__fields_set__` as needed to make this work.
- [x] change the `_load_asset_subtype` validator to be no longer be a `pre` validator and also validate `each`
- [x] thorough config serialization testing of pandas data assets

### Definition of Done


great_expectations/experimental/datasources/interfaces.py
```diff
@@ -341,22 +346,24 @@ class Config:
...
-        LOGGER.debug(f"Loaded 'assets' ->\n{repr(loaded_assets)}")
-        return loaded_assets
+    @pydantic.validator("assets", each_item=True)
+    def _load_asset_subtype(
+        cls: Type[Datasource[DataAssetType]], data_asset: DataAsset
+    ) -> DataAssetType:
+        global CALLS
+        CALLS.append(cls.__name__)
+        LOGGER.warning(f"Loading 'assets' ->\n{pf(data_asset, depth=4)}")
+        asset_type_name: str = data_asset.type
```
Kilo59: catch error that could happen here??

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -137,6 +137,9 @@ class DataAsset(ExperimentalBaseModel):
     # non-field private attrs
     _datasource: Datasource = pydantic.PrivateAttr()
 
+    class Config:
+        extra = pydantic.Extra.forbid
+
```
Kilo59, line 142:
```suggestion
```

great_expectations/experimental/datasources/experimental_base_model.py
```diff
@@ -37,6 +37,19 @@
...
+    Base model for most ZEP pydantic models.
+
+    Adds yaml dumping and parsing methods.
+
+    Extra fields are not allowed.
+
+    Serialization methods default to `exclude_unset = True` to prevent serializing
+    configs full of mostly unset default values.
+    Also prevents passing along unset kwargs to BatchSpec.
+    https://docs.pydantic.dev/usage/exporting_models/
```
Kilo59, line 52: Making this the default for all of our models does not work.
It leads to weird behavior where a `Datasource` doesn't serialize its own assets.

I believe this is due to things that are being set in `pre` validators being treated as if they were default values. This actually makes sense when thinking about using `pre` validators to set dynamic default values like in the example below.
https://docs.pydantic.dev/usage/validators/#validate-always

There are ways I could make this work but I believe the simpler method is to make this the default for all `_DataFrameAsset`s and keep the rest of the models alone.

tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -129,6 +133,32 @@ def test_data_asset_defined_for_io_read_method(self, method_name: str):
...
+        """
+        kwargs: dict[str, Any] = {
+            "name": "test",
+            "path": pathlib.Path(__file__),
+            "regex": re.compile(r"yellow_tripdata_sample_(\d{4})-(\d{2})"),
+        }
+        kwargs.update(extra_kwargs)
+        print(f"extra_kwargs\n{pf(extra_kwargs)}")
+        asset_instance = asset_model(**kwargs)
+        assert asset_instance.dict(exclude_unset=True) == kwargs
```
Kilo59:
```suggestion
        assert asset_instance.dict() == kwargs
```

pyproject.toml
```diff
@@ -295,4 +294,4 @@ mock_use_standalone_module = false
 # https://docs.pytest.org/en/7.1.x/how-to/logging.html#how-to-manage-logging
 # uncomment to adjust captured log levels
 # WARNING some cli v012 tests may fail if this is altered
-# log_level = "info"
+log_level = "info"
```
Kilo59:
```suggestion
# log_level = "info"
```

great_expectations/experimental/datasources/experimental_base_model.py
Kilo59: The above comment is wrong.
It does work, we just need to make use of the pydantic `__fields_set__` attribute if we add items to a model post-init.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -237,7 +249,7 @@ class PandasDatasource(Datasource):
     # instance attrs
     type: Literal["pandas"] = "pandas"
     name: str
-    assets: Dict[str, Union[CSVAsset, ExcelAsset, ParquetAsset, JSONAsset]] = {}  # type: ignore[valid-type]
+    assets: Dict[str, _DataFrameAsset] = {}  # type: ignore[valid-type]
```
Kilo59:
```suggestion
    assets: Dict[str, Union[_DataFrameAsset, CSVAsset, ExcelAsset, ParquetAsset, JSONAsset]] = {}  # type: ignore[valid-type]
```

tests/experimental/datasources/test_config.py
```diff
@@ -204,7 +284,13 @@ def test_catch_bad_top_level_config(
...
                 },
             },
-            ("xdatasources", "assets", "column_splitter", "method_name"),
+            (
+                "xdatasources",
+                "assets",
+                "unknown splitter",
+                "column_splitter",
+                "method_name",
+            ),
```
Kilo59, line 293: These more specific location errors are because of the change to `_load_asset_subtype` which results in asset errors being raised by the asset validators rather than the datasource validators.

great_expectations/experimental/datasources/experimental_base_model.py
```diff
@@ -132,5 +207,36 @@ def _json_dict(
...
+        skip_defaults: bool | None = None,
+    ) -> dict[str, Any]:
+        """
+        Generate a dictionary representation of the model, optionally specifying which
+        fields to include or exclude.
+
+        Deviates from pydantic `exclude_unset` `True` by default instead of `False` by
+        default.
+        """
+        self.__fields_set__.add("assets")
```
Kilo59: This would probably be better as a constant.
`_FIELDS_ALWAYS_SET` or something
I'm now using it in 3 different places/

alexsherstinsky (approved):
LGTM

tests/experimental/datasources/test_config.py
```diff
@@ -69,7 +71,28 @@
...
-        }
+        },
+        "my_pandas_ds": {
+            "type": "pandas",
+            "name": "my_pandas_ds",
+            "assets": {
+                "my_csv_asset": {
+                    "name": "my_csv_asset",
+                    "type": "csv",
+                    "path": __file__,
```
Kilo59: Need to update this to `base_directory`

great_expectations/experimental/datasources/experimental_base_model.py
Kilo59: I've done this.

tests/experimental/datasources/test_config.py
Kilo59: Done

PR #7021, opened 2023-02-01
[BUGFIX] Patch logic error in new `add_or_update` methods

cdkini:
- Previously, changes did not persist - this was an oversight during the initial implementation.
- Additional tests to ensure this now works.

### Definition of Done

### DIFF
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -2549,6 +2549,8 @@ class AbstractDataContext(ConfigPeer, ABC):
             if variables is None:
                 variables = convert_variables_to_dict(existing_profiler.variables)
 
+        config_version = config_version or 1.0
+        rules = rules or {}
         return self.add_profiler(
             name=name, config_version=config_version, rules=rules, variables=variables
         )
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -124,6 +124,9 @@ from great_expectations.rule_based_profiler.config.base import (
 from great_expectations.rule_based_profiler.data_assistant.data_assistant_dispatcher import (
     DataAssistantDispatcher,
 )
+from great_expectations.rule_based_profiler.helpers.util import (
+    convert_variables_to_dict,
+)
 from great_expectations.rule_based_profiler.rule_based_profiler import RuleBasedProfiler
 from great_expectations.util import load_class, verify_dynamic_loading_support
 from great_expectations.validator.validator import BridgeValidator, Validator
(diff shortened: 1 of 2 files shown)


tests/data_context/test_data_context_state_management.py
```diff
@@ -381,24 +381,48 @@ def test_add_or_update_expectation_suite_adds_successfully(
...
 
     assert context.expectations_store.save_count == 1
 
-    suite.expectations = [
-        ExpectationConfiguration(
-            expectation_type="expect_column_values_to_be_in_set",
-            kwargs={"column": "x", "value_set": [1, 2, 4]},
-        ),
-    ]
-    _ = context.add_or_update_expectation_suite(expectation_suite_name=suite_name)
```
cdkini, line 400: This made no sense before - I was updating the state of the object and then only referencing it by name in the `add_or_update` call. This did not persist the changes

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2300,10 +2303,22 @@ def add_or_update_expectation_suite(
...
+            if expectations is None:
+                expectations = existing_suite.expectations
+            if evaluation_parameters is None:
+                evaluation_parameters = existing_suite.evaluation_parameters
+            if data_asset_type is None:
+                data_asset_type = existing_suite.data_asset_type
+            if execution_engine_type is None:
+                execution_engine_type = existing_suite.execution_engine_type
+            if meta is None:
+                meta = existing_suite.meta
```
cdkini, line 2321: Prefer the runtime args for updating but use the existing values as backup. To simplify things, we always delete and write a new value to the store.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2522,13 +2537,18 @@ def add_or_update_profiler(
...
                 name=existing_profiler.name, ge_cloud_id=existing_profiler.ge_cloud_id
             )
+            if id is None:
+                id = existing_profiler.ge_cloud_id
+            if config_version is None:
+                config_version = existing_profiler.config_version
+            if rules is None:
+                rules = {rule.name: rule.to_dict() for rule in existing_profiler.rules}
+            if variables is None:
+                variables = convert_variables_to_dict(existing_profiler.variables)
```
cdkini, line 2550: Special cases since the input objects go through transformations in the __init__

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2496,10 +2511,10 @@ def update_profiler(self, profiler: RuleBasedProfiler) -> None:
     def add_or_update_profiler(
         self,
         name: str,
+        id: str | None = None,
         config_version: float | None = None,
         rules: dict[str, dict] | None = None,
         variables: dict | None = None,
```
Kilo59, line 2517: Will these ever all be none?
And could we represent the different ways this is called without a 3! or 4! (factorial) number of overloads? If so I think it's worth doing.

In other words, if we could represent how this gets called in like 3 `overloads` or less I think they should be added. But if it's going to take more than that it a. might not be worth it or b. could be done in a followup.

Kilo59 (approved):
Have a non-blocking question about adding `@overload`s.

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: I don't think you'll ever have a situation where they'll be `None` - since this method is essentially upsert, you'll want to pass some kind of value to either instantiate a new object or update the attribute of an existing one.

Kilo59: So could this be represented with 2 overloads?
Each one represents a call where one of fields is required and the rest optional.

Apologies for the shorthand here...
```python
    @overload
    def add_or_update_profiler(
        self,
        name: str,
        id: str | None = ...,
        config_version: float | None  = ...,
        rules: dict[str, dict] = ...,  # rules passed
        variables: dict | None = ...,
       )
       ...

    @overload
    def add_or_update_profiler(
        self,
        name: str,
        id: str | None = ...,
        config_version: float | None  = ...,
        rules: dict[str, dict] | None = ...,
        variables: dict = ..., # variables passed
       )
       ...
```

PR #7022, opened 2023-02-01
[DOCS] Fix metric provider and reorganize sidebar

anthonyburdi:
- Methods and classes in metric_provider.py are now able to be rendered. Fixes https://github.com/great-expectations/great_expectations/pull/6997
- Classes with names that matched their containing folder name are now nested appropriately, e.g. Checkpoint (see before/after images below).
    - Note, this is implemented with the addition of `_class` and `.py` to class and module URLs. If this is undesirable happy to revisit.
- This PR also reorganizes and refactors the code that builds the API docs in preparation for removing the duplicated class definitions (currently class definitions show at both the module level and as separate pages for the class if there are module level methods in the same file as a file with class definitions). In this PR I was unable to get Sphinx to respect the `:exclude-members:` option or include explicitly to remove the classes from the module level pages, but I left in the code to track which classes were already documented if we are able to address this in the future. Alternatively, this code can be removed.

Note that all class pages still use the shortest available dotted import path e.g. `great_expectations.checkpoint.Checkpoint` instead of `great_expectations.checkpoint.checkpoint.Checkpoint`:
image

Before, Checkpoint is at a higher level and other classes are nested under it:
image

After, Checkpoint is nested under checkpoint/checkpoint as it is in the codebase:
image

### Definition of Done


great_expectations/expectations/metrics/metric_provider.py
```diff
@@ -21,6 +22,7 @@
 logger = logging.getLogger(__name__)
 
 
+@public_api
```
Shinnnyshinshin, line 25: ❤️

docs/sphinx_api_docs_source/build_sphinx_api_docs.py
```diff
@@ -229,7 +174,129 @@ def _process_and_create_docusaurus_mdx_files(self) -> None:
 
         logger.info("Created mdx files for serving with docusaurus.")
 
-    def _get_mdx_file_path(self, html_file_path: pathlib.Path) -> pathlib.Path:
+    def _get_generated_html_file_paths(self):
```
Shinnnyshinshin, line 177: I like that this was broken up :) Thanks

Shinnnyshinshin (approved):
Thank you very much for this work. LGTM

PR #7044, opened 2023-02-03
[FEATURE] ZEP - generate pandas assets

Kilo59:
- [x] generate `DataAsset` models for all pandas assets unless specifically blacklisted
- [x] test that these new assets have been registered and do some minimal input validation

Note: this PR does not define `add_<ASSET_NAME>()` methods for these new assets because a followup PR will be adding these with metaprogramming.

### Definition of Done


great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -29,30 +29,75 @@ class PandasDatasourceError(Exception):
...
 
+ClipboardAsset = _ASSET_MODELS["clipboard"]
 CSVAsset = _ASSET_MODELS["csv"]
 ExcelAsset = _ASSET_MODELS["excel"]
+FeatherAsset = _ASSET_MODELS["feather"]
+GBQAsset = _ASSET_MODELS["gbq"]
+HDFAsset = _ASSET_MODELS["hdf"]
+HTMLAsset = _ASSET_MODELS["html"]
 JSONAsset = _ASSET_MODELS["json"]
+ORCAsset = _ASSET_MODELS["orc"]
```
Kilo59, line 68: 😢 unfortunately `ORC` is an acronym, so no `OrcAsset` 😆

tests/experimental/datasources/test_schemas.py
```diff
@@ -51,6 +51,8 @@ def test_vcs_schemas_match(zep_ds_or_asset_model: Type[pydantic.BaseModel]):
     if schema_path.name in (
         "SqliteDatasource.json",
         "SqliteTableAsset.json",
+        "SASAsset.json",
+        "PandasDatasource.json",
```
Kilo59, line 56: You'll notice we are ignoring the pandas datasource and it's been deleted from the repo because the new `SASAsset` schema cannot be generated.

I could also temporarily blacklist the `SASAsset` and add this back but... the `PandasDatasource.json` is now almost 2k lines due to it containing all the asset sub definitions.
We could tweak the schema generation so that it uses external `$ref`s instead of internal ones but that's a bit of a rabbit hole that doesn't really impact our launch.

great_expectations/experimental/datasources/pandas_datasource.py
NathanFarmer: 🤣

tasks.py
```diff
@@ -512,19 +513,24 @@ def type_schema(
         schema_dir = GX_ROOT_DIR / "experimental" / "datasources" / "schemas"
         for name in _SourceFactories.type_lookup.type_names():
             model = _SourceFactories.type_lookup[name]
+
             try:
                 schema_path = schema_dir.joinpath(f"{model.__name__}.json")
                 json_str: str = model.schema_json(indent=indent) + "\n"
 
+                if issubclass(model, Datasource):
+                    print(f"🙈  {name} - is a Datasource skipping")
```
NathanFarmer: "is a Datasource - skipping" or "is a Datasource; skipping"?

tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -132,6 +146,23 @@ def test_data_asset_defined_for_io_read_method(self, method_name: str):
 
         assert type_name in asset_class_names
 
+    @pytest.mark.parametrize("asset_class", PandasDatasource.asset_types)
+    def test_minimal_validation(self, asset_class: Type[_FilesystemDataAsset]):
+        with pytest.raises(pydantic.ValidationError) as exc_info:
```
NathanFarmer, line 157: Why does this raise a `ValidationError`? Could we add a comment?

tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -132,6 +146,23 @@ def test_data_asset_defined_for_io_read_method(self, method_name: str):
...
+                regex=re.compile(r"yellow_tripdata_sample_(\d{4})-(\d{2})"),
+                foo="bar",
+            )
+
+        errors_dict = exc_info.value.errors()
+        assert {
+            "loc": ("foo",),
+            "msg": "extra fields not permitted",
+            "type": "value_error.extra",
+        } == errors_dict[-1]
```
NathanFarmer: Can we use `any` or `all` instead of indexing?

tests/experimental/datasources/test_schemas.py
NathanFarmer: What's the issue with the `SASAsset`? Can we add a comment here as well? Same question for `Sqlite`.

tests/experimental/datasources/test_pandas_datasource.py
Kilo59: I can add a comment, but it's because of the extra "forbidden" keyword arg `foo`

tasks.py
```suggestion
                    print(f"🙈  {name} - is a Datasource; skipping")
```

tasks.py
```diff
@@ -512,19 +519,33 @@ def type_schema(
...
             model = _SourceFactories.type_lookup[name]
+
+            if (
+                model in {*PandasDatasource.asset_types, PandasDatasource}
+                and _PANDAS_SCHEMA_VERSION != pandas.__version__
+            ):
+                print(
+                    f"🙈  {name} - was generated with pandas {_PANDAS_SCHEMA_VERSION}; skipping"
+                )
+                continue
```
Kilo59, line 530: @NathanFarmer @billdirks this logic will stop `invoke schemas --sync` from generating "bad" pandas schemas.

tests/experimental/datasources/test_schemas.py
Kilo59: For sqlite I think it's the `SqliteDsn`, pydantic doesn't know what type to map it to. Which is a bit surprising, I would have just assumed it would treat it the same as its parent `AnyUrl` type.
https://docs.pydantic.dev/usage/schema/#json-schema-types
I'd need to look more into the pydantic internals.
https://docs.pydantic.dev/usage/types/#custom-data-types

For `SASAsset` it's something, similar, there's a field type somewhere that pydantic doesn't know how to map to a JSONSchema type. I'll update this comment with specifics once I know. If it's a simple fix I'll make it, if it requires much special handling I'll likely save it for a follow-up PR.

I've created tickets for both.

tests/experimental/datasources/test_pandas_datasource.py
Kilo59: I don't want to use `any` or `all` because I want to see the nice assert dict diff if it fails.

There's ~one~ a couple of special cases that prevents me from doing
```python
assert [{
            "loc": ("foo",),
            "msg": "extra fields not permitted",
            "type": "value_error.extra",
        }] == error_dicts
```
Which is any asset that has more than 1 has required field. For those, this error will be the last error in the list. For all others the last item in the list is the same as the first one.
I'll add a comment to explain.

Kilo59: On second thought.... `kwargs` doesn't make sense as a required field. I'll make it optional and then this test can be simpler.

great_expectations/experimental/datasources/schemas/ParquetAsset.json
```diff
@@ -1,6 +1,6 @@
...
...rrow', falling back to 'fastparquet' if\n    'pyarrow' is unavailable.\ncolumns : list, default=None\n    If not None, only these columns will be read from the file.\n\nstorage_options : dict, optional\n    Extra options that make sense for a particular storage connection, e.g.\n    host, port, username, password, etc. For HTTP(S) URLs the key-value pairs\n    are forwarded to ``urllib`` as header options. For other URLs (e.g.\n    starting with \"s3://\", and \"gcs://\") the key-value pairs are forwarded to\n    ``fsspec``. Please see ``fsspec`` and ``urllib`` for more details.\n\n    .. versionadded:: 1.3.0\n\nuse_nullable_dtypes : bool, default False\n    If True, use dtypes that use ``pd.NA`` as missing value indicator\n    for the resulting DataFrame. (only applicable for the ``pyarrow``\n    engine)\n    As new dtypes are added that support ``pd.NA`` in the future, the\n    output with this option will change to use those dtypes.\n    Note: this is an experimental option, and behaviour (e.g. additional\n    support dtypes) may change without notice.\n\n    .. versionadded:: 1.2.0\n\n**kwargs\n    Any additional kwargs are passed to the engine.\n\nReturns\n-------\nDataFrame",
```
Kilo59, line 3: All these descriptions are coming from the `__doc__` (docstring) of the related pandas `read` method. It might be enough to just include the summary line (split on first `/n`) and then include a link to the pandas `io` API docs.
https://pandas.pydata.org/docs/reference/io.html

tests/experimental/datasources/test_pandas_datasource.py
Kilo59: I'd added comments to the test to explain.

great_expectations/experimental/datasources/__init__.py
```diff
@@ -1,5 +1,7 @@
 import pathlib
 
+import pandas
+
```
Kilo59:
```suggestion
```

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -312,12 +349,15 @@ def _generate_pandas_data_asset_models(
                 type_field=(f"Literal['{type_name}']", type_name),
                 fields_dict=fields,
                 extra=pydantic.Extra.forbid,
+                model_docstring=signature_tuple.docstring.partition("\n\nParameters")[0]
```
Kilo59, line 362: This is safer than split because if the string doesn't contain the target characters the first item in the result tuple is the original string.
https://docs.python.org/3/library/stdtypes.html#str.partition

NathanFarmer (approved):

PR #7001, opened 2023-01-31
[FEATURE] ID/PK `ColumnPairExpectations` and `MultiColumnMapExpectations` - Spark

Shinnnyshinshin:
# What is this PR? 
- Enables ID/PK for `ColumnPairExpectations` and `MultiColumnMapExpectations` by.
   1. Adding `UNEXPECTED_INDEX_LIST` and `UNEXPECTED_INDEX_QUERY` to validation dependencies
   2. Modifying metric functions to take in domain column **list** rather than just domain column. 
- Enables JSON-defined tests to run against Spark for ID/PK
- Adds tests at Checkpoint level for MultiColumn and ColumnPair Expectations

### Definition of Done
Please delete options that are not relevant.

- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added [unit tests](https://docs.greatexpectations.io/docs/contributing/contributing_test#writing-unit-and-integration-tests) where applicable and made sure that new and existing tests are passing.
- [x] I have run any local integration tests and made sure that nothing is broken.

Thank you for submitting!

PR #7046, opened 2023-02-04
[FEATURE] ID/PK ColumnPairExpectations and MultiColumnMapExpectations - SQL

Shinnnyshinshin:
## What is this PR?
* Similar to #7001, ID/PK for `ColumnPairExpectations` and` MultiColumnMapExpectations` is enabled for SQL by.
   * Adding `UNEXPECTED_INDEX_LIST` and `UNEXPECTED_INDEX_QUERY` to validation dependencies.
   * Modifying metric functions to take in domain column list rather than just domain column.
   * As mentioned during StandUp, ID/PK currently doesn't work for `compound_columns.unique` metric only. To be addressed in follow-up PR. 
* Enables JSON-defined tests to run against SQL for ID/PK
  * This one required some more modifications than Spark because each backend seems to handle datetime differently 😓 
* Adds tests at Checkpoint level for MultiColumn and ColumnPair Expectations


great_expectations/expectations/expectation.py
```diff
@@ -3374,63 +3343,34 @@ def get_validation_dependencies(
...
-            )
-        elif isinstance(execution_engine, SparkDFExecutionEngine):
-            metric_kwargs = get_metric_kwargs(
-                metric_name=f"{self.map_metric}.{SummarizationMetricNameSuffixes.UNEXPECTED_INDEX_LIST.value}",
-                configuration=configuration,
-                runtime_configuration=runtime_configuration,
-            )
-            validation_dependencies.set_metric_configuration(
+        if self.map_metric == "compound_columns.unique":
+            return validation_dependencies
```
Shinnnyshinshin, line 3345: This ensure that `UNEXPECTED_INDEX_LIST` and `UNEXPECTED_INDEX_QUERY` are not computed for the `compound_columns.unique` metric.  Will be removed once it's fixed

great_expectations/expectations/expectation.py
```diff
@@ -3074,63 +3074,32 @@ def get_validation_dependencies(
...
-            )
-            validation_dependencies.set_metric_configuration(
+        metric_kwargs = get_metric_kwargs(
+            metric_name=f"{self.map_metric}.{SummarizationMetricNameSuffixes.UNEXPECTED_INDEX_LIST.value}",
+            configuration=configuration,
+            runtime_configuration=runtime_configuration,
+        )
+        validation_dependencies.set_metric_configuration(
+            metric_name=f"{self.map_metric}.{SummarizationMetricNameSuffixes.UNEXPECTED_INDEX_LIST.value}",
+            metric_configuration=MetricConfiguration(
```
Shinnnyshinshin, line 3084: Simplified this section (no more `if` statements needed) since we are enabling `UNEXPECTED_INDEX_LIST` and `UNEXPECTED_INDEX_QUERY` for all ExecutionEngines.

tests/expectations/metrics/column_map_metrics/test_unexpected_indices_and_query.py
```diff
@@ -270,8 +270,9 @@ def test_sa_unexpected_index_query_metric_with_id_pk(
     )
     for (key, val) in results.items():
         assert (
-            val
-            == "SELECT animals, pk_1 \nFROM test \nWHERE animals IS NOT NULL AND (animals NOT IN ('cat', 'fish', 'dog'));"
+            "SELECT pk_1, animals \n"
+            "FROM test \n"
+            "WHERE animals IS NOT NULL AND (animals NOT IN ('cat', 'fish', 'dog'));"
         )
```
anthonyburdi: Looks like this will always be true? Does this need to be compared to val like in the previous version?

anthonyburdi (changes requested):
Really strong work! Just one small change requested.

tests/expectations/metrics/column_map_metrics/test_unexpected_indices_and_query.py
Shinnnyshinshin: @anthonyburdi yes thank you for spotting this. Yes, this will always be `True`. The reason why we had to compare to other versions in the Pandas PR is because the method had to take into account default indices for Pandas Dataframes (which SQL doesn't have), which depended on a `groupby()` stage that had multiple potential outputs. Therefore, the list of columns is always going to be in the same order.

anthonyburdi: But shouldn't we still compare to the query (`val` I think here) that we are getting in the results to the expected query? To make sure the output query is correct? The proposed change means we are just asserting that it is a non-empty string.

Shinnnyshinshin: oh nooooo. You're right. Let me fix this real quick :)

anthonyburdi (approved):
Thanks for the quick fix! Excited for this new functionality :)

PR #7049, opened 2023-02-04
[MAINTENANCE] ruff - enable `TCH` flake8-type-checking

Kilo59:
https://github.com/charliermarsh/ruff#flake8-type-checking-tch

This PR partially enables the `TCH` `flake8-type-checking` rules.
These rules help identify unnecessary imports if they are only being used for `TYPE_CHECKING`.
This reduces runtime imports, which means faster startup and reduces or eliminates cyclic import dependencies.

See https://github.com/snok/flake8-type-checking#flake8-type-checking for more details

- enables `TCH` rules
   - fixes all `TCH002` `third-party-imports`  violations
- ignore  `TCH003` - typing-only-standard-library-import
  - importing from the standard library introduces minimal runtime cost
  - making stdlib imports move to a `TYPE_CHECKING` block is tedious and may discourage the use of type annotations
- adds directory specific `ruff.toml` configuration files for the directories below...
   - `tests/`
   - `contrib/`
   - Without excluding `tests/`, `contrib/` we are approaching 1K errors for `TCH001`

image

## Questions
Is `TCH002` `third-party-imports` really that useful for us?
Do we want to force devs to always use an `if TYPE_CHECKING` block to import `pandas` or `numpy`?
The main benefit of these rules seems to be for `first-party` imports and the reduction of circular dependencies.
- [ ] I'm tempted to drop all the `TCH002` changes and disable the rule.


Kilo59:
Closing in favor of 
https://github.com/great-expectations/great_expectations/pull/7072
https://github.com/great-expectations/great_expectations/pull/7073

PR #6907, opened 2023-01-24
[MAINTENANCE] replace `isort` with `ruff` sorting rules

Kilo59:
### Changes proposed in this pull request:

- [x] - remove `isort` dependencies
- [x] - remove `isort` pre-commit hook
- [x] - update documentation
- [ ] - ~ruff pre-commit hook should run on full code-base (at least all of `great_expecatations/`)~
  - [ ] - ~for this we might want to wait until all of `great_expectations` is being linted (lint `great_expecations/dataset`)~
- ~update `expectation_diagnostics.py` to use `ruff` sorting~ - [see comment](https://github.com/great-expectations/great_expectations/pull/6907#discussion_r1084775270)
- [x] remove expectation_diagnostics `_check_linting` method
- [x] - `contrib/cli/great_expectations_contrib/commands.py` to use `ruff` sorting

### Definition of Done
Please delete options that are not relevant.

- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code
- [x] I have made corresponding changes to the documentation

Follow-up to https://github.com/great-expectations/great_expectations/pull/6888

PR #7050, opened 2023-02-04
[MAINTENANCE] finish linting `great_expectations`

Kilo59:
With this change, all of `great_expectations` is now completely under linting coverage.
This makes it easier to proceed with the `isort` -> `ruff` move #6907 without gaps in coverage.

1. remove `great_expectations/dataset` from the ruff exclude list
2. update the `pre-commit` `ruff` hook to lint anything in `great_expectations/` and `contrib/` instead of only `great_expectations/core`
3. create a `great_expectations/dataset/ruff.toml` file for overriding `ruff` linting rules for the `dataset/` package.


.pre-commit-config.yaml
```diff
@@ -24,7 +24,7 @@ repos:
     rev: 'v0.0.238'
     hooks:
       - id: ruff
-        files: ^(great_expectations/core)  # TODO: expand to all of gx
+        files: ^(great_expectations|contrib) # TODO: add tests/ scripts/ etc.
```
Kilo59, line 23: Now that all of `great_expectations` is being linted we can use the pre-commit hook on more than just `great_expectations/core`

NathanFarmer (approved):

billdirks (commented):
LGTM, thanks!

PR #7051, opened 2023-02-05
[FEATURE] Experimental Splitter connection testing

NathanFarmer:
- GREAT-1490
- Add `connection_test` for `ColumnSplitter`
- Leverage `Generic` `TypeVar` `ColumnSplitterType` in anticipation of additional Splitter classes/methods and the need for connection testing for `TableAsset` splitters, but not `QueryAsset` splitters
- Refactor Sqlite splitters to bring in line with other splitter patterns and remove `isinstance`/type ignores
- Introduce `Generic` `TypeVar` `ExecutionEngineType` for passing along `Datasource.get_execution_engine()` typing information
- Make `Datasource` `test_connection` error message unit test less brittle

### Definition of Done


tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -713,64 +721,36 @@ def test_validate_invalid_postgres_connection_string(
...
 
 
-def bad_connection_string_config() -> tuple[
-    str, str, str, str | None, TestConnectionError
-]:
+def bad_connection_string_config() -> tuple[str, str, str]:
     connection_string = "postgresql+psycopg2://postgres:@localhost/bad_database"
     table_name = "good_table"
     schema_name = "good_schema"
-    mock_sqlalchemy_engine_connect_error_raised = (
```
NathanFarmer, line 722: After comments from @billdirks, I decided to remove any assertions against error message strings, even though this strictly tests GX messages.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -334,7 +339,7 @@ class Datasource(
     }
     # Setting this in a Datasource subclass will override the execution engine type.
     # The primary use case is to inject an execution engine for testing.
-    execution_engine_override: ClassVar[Optional[Type[ExecutionEngine]]] = None
+    execution_engine_override: ClassVar[Optional[Type[ExecutionEngineType]]] = None  # type: ignore[misc]
```
NathanFarmer: It is not recommended to use `Generic` types in `ClassVar`s, but this attribute seems like a code smell to me, and type ignores related to this attribute are the only thing required to get `ExecutionEngineType` implemented.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -80,6 +82,30 @@ def _splitter_method_exists(cls, v: str):
...
+    def test_connection(self, table_asset: TableAsset) -> None:
+        """Test the connection for the ColumnSplitter.
+
+        Args:
+            table_asset: The TableAsset to which the ColumnSplitter will be applied.
+
+        Raises:
+            TestConnectionError: If the connection test fails.
+        """
+        assert isinstance(table_asset.datasource, SQLDatasource)
```
NathanFarmer, line 94: A `TypeVar` for `Datasource` would get rid of this check, but circularity between `Datasource` <-> `DataAsset` is an issue.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -178,18 +203,15 @@ def add_year_and_month_splitter(
...
         Args:
             column_name: A column name of the date column where year and month will be parsed out.
-
         Returns:
-            This DataAsset so we can use this method fluently.
+            This _SQLAsset so we can use this method fluently.
         """
-        self.column_splitter = SqlYearMonthSplitter(
-            column_name=column_name,
+        raise NotImplementedError(
```
NathanFarmer, line 216: If we want to avoid `isinstance` checks in superclasses `_SQLAsset` and `ColumnSplitter` we need to make this abstract and implement in the subclasses due to the diverging logic for `test_connection` on splitters.

NathanFarmer: This also narrows types at the API-level without needing to introduce another `TypeVar`.

great_expectations/experimental/datasources/sqlite_datasource.py
```diff
@@ -16,62 +16,74 @@
...
     SQLDatasourceError,
+    SqlYearMonthSplitter,
     TableAsset,
     _query_for_year_and_month,
-    _SQLAsset,
 )
 
 
-class _SplitterMixin:
+class SqliteTableAsset(TableAsset):
```
NathanFarmer, line 29: Hopefully I'm not missing anything regarding the benefits of these mixins, but it doesn't seem as scalable as we will likely continue to require dialect-specific `SQLAsset`s.

NathanFarmer: Also, moving to inheritance gets rid of the `isinstance` checks and `type: ignore`s.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -311,8 +310,14 @@ def sort_batches(self, batch_list: List[Batch]) -> None:
 DataAssetType = TypeVar("DataAssetType", bound=DataAsset)
 
 
+# It would be best to bind this to ExecutionEngine, but we can't now due to circular imports
+ExecutionEngineType = TypeVar("ExecutionEngineType")
```
Kilo59: `non-blocking`
You don't need to actually do this but...
We could bind it to an `ExecutionEngine` protocol that defines whatever methods we actually call within `zep` code.

great_expectations/experimental/datasources/interfaces.py
Kilo59: Yeah, I'm not sure how this would work. 🤔 

This might need to be left as is with a `type` ignore.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -81,6 +83,33 @@ def _splitter_method_exists(cls, v: str):
...
+        Raises:
+            TestConnectionError: If the connection test fails.
+        """
+        # A TypeVar for Datasource would get rid of this assertion,
+        # but circularity between Datasource <-> DataAsset is an issue.
+        assert isinstance(table_asset.datasource, SQLDatasource)
+        engine: sqlalchemy.engine.Engine = table_asset.datasource.get_engine()
+        inspector: sqlalchemy.engine.Inspector = sqlalchemy.inspect(engine)
+
+        columns: list[dict[str, Any]] = inspector.get_columns(
```
alexsherstinsky, line 101: @NathanFarmer Observation: Inspect will not always work.  While I am not certain of the best path forward, perhaps considering our fallback methods in metrics/util.py might help.  Thank you.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -278,7 +303,9 @@ def get_batch_list_from_batch_request(
                 )
             # Creating the batch_spec is our hook into the execution engine.
             batch_spec = SqlAlchemyDatasourceBatchSpec(**batch_spec_kwargs)
-            execution_engine: ExecutionEngine = self.datasource.get_execution_engine()
+            execution_engine: SqlAlchemyExecutionEngine = (
```
alexsherstinsky, line 306: ❤️ for type narrowing here as well as in the file-path (Pandas, Spark) cases!

great_expectations/experimental/datasources/interfaces.py
Kilo59: `non-blocking` - I'll create a ticket

Also not necessarily needed in this PR but I think we should change how we name our `TypeVar` so that they are more obviously `TypeVar`s.

Here's the same issue being discussed in `typeshed`
https://github.com/python/typeshed/issues/1872
https://github.com/python/typeshed/pull/4553/files

So this might be `EET`  `_EET`, or `ExecEngineT`
And `DataAssetType` -> `DAT`, `_DAT`, `DataAssetT`

I prefer the shorter names, because the longer one still looks like it's regular type/class

https://peps.python.org/pep-0008/#type-variable-names.

alexsherstinsky (approved):
@NathanFarmer I am approving, but please do not merge until someone else with more experience in the ZEP world provides feedback as well.  I left a note for consideration.  For a slightly longer term (within a couple of days), I am also working on a hypothesis for leveraging Data Connectors within the new Asset in the file path modules (Pandas, Spark), which, if works, might entail some beneficial refactoring opportunities in the SQL ZEP modules as well.  Thanks!

tests/experimental/datasources/test_postgres_datasource.py
Kilo59: I'm fine with this but just wanted to flag that `pytest.raises` can do a partial string match that would be less brittle than an exact match.
https://docs.pytest.org/en/7.2.x/reference/reference.html#pytest.raises

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: Thanks @alexsherstinsky! What I was able to glean from `metrics/util.py` is that this might not work for `mssql` or `trino`. We will want to handle that logic when we implement subclasses such as `MSSQLDatasource` and `TrinoDatasource`. A good thing to keep in mind.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -426,6 +495,9 @@ def _create_batch_spec_kwargs(self) -> dict[str, Any]:
         }
 
 
+SQLAsset: TypeAlias = Union[_SQLAsset, QueryAsset, TableAsset]
```
Kilo59: I don't think we should be doing this.
If we don't like the way `_SQLAsset` looks we can change the mechanism the `Metadatasource` uses to register types and then we can rename `_SQLAsset` -> `SQLAsset`.

In my head, it would just be a `_register_type: ClassVar[bool] = True/Flase` attribute that gets checked by the `Metadatasource`.
I can create a ticket for that.

Kilo59:
Everything looks good.
I especially like all the improved generics work 🚀 .

The only thing I might consider `blocking` is the `SQLAsset` type-alias, which I'd like to handle differently. 
https://github.com/great-expectations/great_expectations/pull/7051#discussion_r1097715950

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: Per our offline conversation I have reverted this `TypeAlias` change.

Kilo59 (approved):

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -335,22 +338,22 @@ class Datasource(
     }
     # Setting this in a Datasource subclass will override the execution engine type.
     # The primary use case is to inject an execution engine for testing.
-    execution_engine_override: ClassVar[Optional[Type[ExecutionEngine]]] = None
+    execution_engine_override: ClassVar[Optional[Type[_ExecutionEngineT]]] = None  # type: ignore[misc]
```
billdirks, line 341: What are the details behind this new type ignore? That is, I am curious what the error message is if you don't have this.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Maybe we don't want to test the connection on splitters by default. That is, since this works on postgres but not on generic splitters, maybe we want to not check the connection here but instead check the connection on the classes we know this works for.
That is, what type of errors would a `mssql` and `trino` user see here? Also, if this would work without for them if we didn't do this check, it seems like this check is too high up the hierarchy.

great_expectations/experimental/datasources/sqlite_datasource.py
billdirks: It was about removing verbosity but this looks better to me.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: After internal discussion name was changed to `_ExecutionEngineT`.

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: Following up from our internal discussion, the forks seen in `metrics/util.py` may be due to temp table use by `SqlAlchemyExecutionEngine` and we will be adding testing for various dialects to ensure this works.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: Message: `ClassVar cannot contain type variables`. I added a comment in this PR: https://github.com/great-expectations/great_expectations/pull/7010

PR #7054, opened 2023-02-05
[MAINTENANCE] format notebooks with `black`

Kilo59:
- format all jupyter notebooks
- install `black[jupyter]` in ci to enforce notebook formatting
- use `black_jupyter` pre-commit hook
- exclude invalid notebooks from formatting check

In Arch review, we had talked about [excluding this from the git-blame](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame), but as explained in the post below, we can't actually do this easily because we always squash and merge (which means we won't know what commits to ignore until after the merge).
https://github.com/community/community/discussions/16880

We likely don't want to ignore the related dependency and configuration changes, we only want to ignore the formatting changes. So if we want to do this it would require 3 total PR's.
1. Just the formatting changes
2. Add entries to `.git-blame-ignore-revs`
3. Add the configuration and dependency changes that actually enforce the formatting changes from PR 1.

In this case, it wouldn't require any force-merges but for many other scenarios, it would.
For example, the newest version of black introduces some additional formatting changes. There's no way to introduce that (and ignore-revs) without a force-merge. If we make the formatting changes first the CI will fail because the old black will format differently. If we make the `black` version change first the CI will fail for the same reason.

--------

Although we talked about ignoring these changes in the git-blame. Notebooks are really hard to view in git, so I don't know that all these extra steps to ignore the blame are worth it here.


contrib/cli/requirements.txt
```diff
@@ -1,4 +1,4 @@
-black==22.3.0        # Linting / code style
+black[jupyter]==22.3.0 # Linting / code style
```
Kilo59, line 1: Initially, I updated the version of black to `v23` but that introduced a bunch of formatting changes to `.py` files so I reverted.

cdkini (approved):

PR #7055, opened 2023-02-05
[MAINTENANCE] fix - `FutureWarning: pandas.Float64Index is deprecated` when importing `great_expectations`

Kilo59:
```
>>> import great_expectations
FutureWarning: pandas.Float64Index is deprecated and will be removed from pandas in a future version.
Use pandas.Index with the appropriate dtype instead.
```

This happens because of the `pandas` io method introspection that happens at import time.

Using the `catch_warnings` context-manager to temporarily suppress this warning. While allowing pandas to raise the warning under other conditions.
https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings

- filter `FutureWarning` during zep pandas method inspection

### DIFF
--- a/great_expectations/experimental/datasources/dynamic_pandas.py
+++ b/great_expectations/experimental/datasources/dynamic_pandas.py
@@ -175,9 +175,9 @@ def _extract_io_methods(
         warnings.simplefilter(action="ignore", category=FutureWarning)
 
         member_functions = inspect.getmembers(pd, predicate=inspect.isfunction)
-        if whitelist:
-            return [t for t in member_functions if t[0] in whitelist]
-        return [t for t in member_functions if t[0].startswith("read_")]
+    if whitelist:
+        return [t for t in member_functions if t[0] in whitelist]
+    return [t for t in member_functions if t[0].startswith("read_")]
 
 
 def _extract_io_signatures(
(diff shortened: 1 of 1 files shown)


great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -168,7 +169,12 @@ def _replace_builtins(input_: str | type) -> str | type:
 def _extract_io_methods(
     whitelist: Optional[Sequence[str]] = None,
 ) -> List[Tuple[str, DataFrameFactoryFn]]:
-    member_functions = inspect.getmembers(pd, predicate=inspect.isfunction)
+    # temporarily suppress pandas future warnings that may be emitted by collecting
+    # pandas io methods
+    with warnings.catch_warnings():
+        warnings.simplefilter(action="ignore", category=FutureWarning)
```
NathanFarmer, line 178: So am I correct in thinking that this is a pandas warning that is being emitted due to some usage in the pandas code itself? Is there no way to narrow this category further?

Kilo59: Yes, it's being emitted simply from the act of importing and inspecting these pandas methods.
We could narrow the category further but I don't want to do that because I don't want to have to tweak this code for every slight variation in the `FutureWarning` message.

But I also don't want to suppress any of these warnings throughout our codebase because it probably is a relevant warning in other places. Hence the temporary suppression of the warning.

NathanFarmer (approved):

PR #7056, opened 2023-02-06
[BUGFIX] zep - always serialize type field

Kilo59:
https://github.com/great-expectations/great_expectations/pull/7009 changed the default behavior of all ZEP class serialization methods to exclude any fields that haven't been explicitly set.
Without this change, we would end up with configs full of default field values.
However, we always need the value of `type` in order to determine the correct asset class/type to use.

- always serialize the `Datasource` and `DataAsset` `type` field


tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -647,6 +649,8 @@ def test_data_source_yaml_has_properties(create_source: CreateSourceFixture):
         )
         source_str = source.__str__()
         assert "order_by:" in source_str
+        # type should be in dumped str even if not explicitly set
+        assert f"type: {asset.type}" in source_str
```
Kilo59, line 662: Note. These 3 `test_postgres_datasource.py` tests are the only ones that I could actually get to catch the error.
This only happens when an asset is created in code and no `type` field is set.

I couldn't reproduce it with any of the config tests (as expected).

tests/experimental/datasources/test_config.py
```diff
@@ -508,3 +509,32 @@ def test_custom_sorter_serialization(
...
+            PG_COMPLEX_CONFIG_DICT["xdatasources"]["my_pg_ds"],
+            id=PG_COMPLEX_CONFIG_DICT["xdatasources"]["my_pg_ds"]["type"],
+        ),
+        p(
+            PG_COMPLEX_CONFIG_DICT["xdatasources"]["my_pandas_ds"],
+            id=PG_COMPLEX_CONFIG_DICT["xdatasources"]["my_pandas_ds"]["type"],
+        ),
+    ],
+)
+def test_type_field_always_dumped(ds_w_assets: dict):
```
Kilo59: This test didn't actually catch the error so it's of questionable value.

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -629,9 +629,11 @@ def test_data_source_json_has_properties(create_source: CreateSourceFixture):
         asset.add_year_and_month_splitter(column_name="my_col").add_sorters(
             ["year", "month"]
         )
-        source_json = source.json(indent=4, sort_keys=True)
+        source_json = source.json(indent=4, sort_keys=True, exclude_unset=True)
```
Kilo59:
```suggestion
        source_json = source.json(indent=4, sort_keys=True)
```

tests/experimental/datasources/test_postgres_datasource.py
```diff
@@ -660,9 +664,11 @@ def test_datasource_dict_has_properties(create_source):
         asset.add_year_and_month_splitter(column_name="my_col").add_sorters(
             ["year", "month"]
         )
-        source_dict = source.dict()
+        source_dict = source.dict(exclude_unset=True)
```
Kilo59:
```suggestion
        source_dict = source.dict()
```

tests/experimental/datasources/test_config.py
billdirks: Is this because you are using a dictionary with the `type` field defined and checking that the type field roundtrips?
If you don't think this adds value then feel free to remove it.

billdirks (approved):
Thank you! LGTM. One question responding to your comment.

tests/experimental/datasources/test_config.py
Kilo59: It's because the only way to reproduce the `type` field not being serialized error is when adding assets "in-code". Because that's the only place where the default `type` field value is used. And because the default is used, pydantic doesn't consider the field as having been "set".

So when we dump the model with `exclude_unset=True` it gets left out.
https://docs.pydantic.dev/usage/exporting_models/#modeldict
Note in the above docs it says `exclude_unset` defaults to `False`, but we've changed this behavior to default to `True`

When we read from config (dict, yaml or json) everything (including `type`) gets loaded into the `DataAsset` class constructer and so pydantic considers it `set`, that's why the `test_config.py` tests didn't catch the issue.

So.... I'll delete these new `test_config.py` tests and leave the changes to `test_postgres_datasource.py`

tests/experimental/datasources/test_config.py
```diff
@@ -3,14 +3,14 @@
...
-from typing import Callable, List
+from typing import Callable, List, Type
 
 import pydantic
 import pytest
 
 from great_expectations.data_context import FileDataContext
 from great_expectations.experimental.datasources.config import GxConfig
-from great_expectations.experimental.datasources.interfaces import Datasource
+from great_expectations.experimental.datasources.interfaces import DataAsset, Datasource
```
Kilo59:
```suggestion
from great_expectations.experimental.datasources.interfaces import Datasource
```

tests/experimental/datasources/test_config.py
```diff
@@ -3,7 +3,7 @@
 import pathlib
 import re
 from pprint import pformat as pf
-from typing import Callable, List
+from typing import Callable, List, Type
```
Kilo59:
```suggestion
from typing import Callable, List
```

PR #7071, opened 2023-02-07
[MAINTENANCE] Cache dependency installation during build of `Dockerfile.tests`

NathanFarmer:
- GREAT-1677
- Reduce test image build times for `SOURCE=local` as much as 90%
- Move `COPY . .` to after `pip install` ensuring caching if requirements are unchanged
- 813.0s -> 74.4s

### Before:
813.0s
image

### After

Run 1: 987.0s
image

Run 2 - After non-dependency code change: 74.4s
image

### Definition of Done

### DIFF
--- a/docker/Dockerfile.tests
+++ b/docker/Dockerfile.tests
@@ -15,21 +15,27 @@ ARG BRANCH=develop
 RUN git clone --filter=tree:0 --branch ${BRANCH} https://github.com/great-expectations/great_expectations.git
 WORKDIR /great_expectations
 
+FROM base as build_local
+WORKDIR /great_expectations
+COPY ./requirements.txt ./requirements.txt
+COPY ./requirements-dev.txt ./requirements-dev.txt
+COPY ./constraints-dev.txt ./constraints-dev.txt
+COPY ./reqs ./reqs
+
 FROM build_${SOURCE} AS dev
 ARG GE_USAGE_STATISTICS_URL="https://dev.stats.greatexpectations.io/great_expectations/v1/usage_statistics"
 env GE_USAGE_STATISTICS_URL="${GE_USAGE_STATISTICS_URL}"
 RUN pip install --no-cache-dir --requirement requirements-dev.txt --constraint constraints-dev.txt
-RUN pip install .
 CMD [ "python", "--version"]
 
 FROM dev as test
 RUN pip install --no-cache-dir --requirement requirements-dev.txt --requirement reqs/requirements-dev-test.txt \
     --constraint constraints-dev.txt
+COPY . .
+RUN pip install .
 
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.tests
```diff
@@ -33,3 +29,7 @@ RUN pip install --no-cache-dir --requirement requirements-dev.txt --requirement
 FROM dev as dev-tools
 RUN pip install --no-cache-dir --requirement requirements-dev.txt --requirement reqs/requirements-dev-test.txt \
     --requirement reqs/requirements-dev-tools.txt --constraint constraints-dev.txt
+
+FROM base as build_local
```
billdirks: The stage `FROM build_${SOURCE} AS dev` either reads from `build_local` or `build_github` depending on whether you are building an image with your local environment or from github (controlled by the ARG `SOURCE` passed into the `docker buildx` command). Moving this stage lower will break this.

docker/Dockerfile.tests
```diff
@@ -8,28 +8,34 @@ FROM python:${PYTHON_VERSION} AS base
...
 FROM base AS build_github
 # BRANCH is the branch name you want to build
 ARG BRANCH=develop
 # We use --filter=tree:0 instead of --depth 1 because we really on git to get version information
 RUN git clone --filter=tree:0 --branch ${BRANCH} https://github.com/great-expectations/great_expectations.git
 WORKDIR /great_expectations
 
+FROM base as build_local
+WORKDIR /great_expectations
+COPY ./requirements.txt ./requirements.txt
```
billdirks: I believe you can copy multiple files all at once like this:
`COPY ./requirements.txt ./requirements-dev.txt ./constraints-dev.txt ./reqs ./`
You might need to break out `reqs` into it's own line because it is a directory.

The reason we want to put it all one 1 line is because docker will make a new layer for each line that makes a filesystem modification. This makes the docker image a little bit bigger.

Also, it doesn't matter if this block comes before or after the `FROM base AS build_github` block since either one or the other will build but not both.

Note, this change means we don't copy the whole local directory to the docker image, just the requirement files, but that is likely fine for local development, since you will be mounting your own directory.

billdirks (approved):
I have 1 comment inline but otherwise LGTM.

Kilo59 (approved):
🙌

docker/Dockerfile.tests
NathanFarmer: @billdirks I made the changes you described, thanks for the explanation.

> Note, this change means we don't copy the whole local directory to the docker image, just the requirement files, but that is likely fine for local development, since you will be mounting your own directory.

We are still doing `COPY . .` in the subsequent steps, so I think this would only be true if you use `--target dev`, correct?

PR #7078, opened 2023-02-08
[MAINTENANCE] Move experimental `base_directory` from `_FilesystemDataAsset` to `PandasDatasource` and `SparkDatasource`

NathanFarmer:
- GREAT-1678
- Move `base_directory` from `_FilesystemDataAsset` to `PandasDatasource` and `SparkDatasource`
- Rename `add_pandas` factory function to `add_pandas_filesystem`
- Introduce abstract classes `_PandasDatasource` and `_SparkDatasource`
- Introduce `_DatasourceT` and push methods referencing `self.datasource` down into subclasses for more specific typing information
- Use `TypeVars` and `alias` to kill type ignores at API-level
- Change default serialization method for `by_alias` to `True`
- Update `DataAsset` schemas

### Definition of Done


great_expectations/experimental/datasources/file_path_data_asset.py
```diff
@@ -16,21 +20,31 @@
...
-        "regex",
-        "order_by",
+        "regex_setter",
+        "order_by_setter",
         "type",
     }
 
     # General file-path DataAsset pertaining attributes.
-    regex: Pattern
+    regex_setter: Union[str, Pattern] = pydantic.Field(alias="regex")
```
NathanFarmer: If we don't use an alias here, `mypy` will complain that you can't set the `regex` property. This enables us to pass along the typing information from here to the API-level.

great_expectations/experimental/datasources/file_path_data_asset.py
```diff
@@ -65,6 +79,16 @@ def __init__(self, **data):
         )
         self._all_group_names = self._regex_parser.get_all_group_names()
 
+    @property
+    def _base_directory(self) -> pathlib.Path:
```
NathanFarmer, line 82: This is to pass `base_directory` to the various path fields in the dynamic pandas assets.

great_expectations/experimental/datasources/file_path_data_asset.py
```diff
@@ -80,12 +104,46 @@ def get_batch_request(
...
+            expect_batch_request_form = BatchRequest(
+                datasource_name=self.datasource.name,
+                data_asset_name=self.name,
+                options=self.batch_request_options_template(),
+            )
+            raise gx_exceptions.InvalidBatchRequestError(
+                "BatchRequest should have form:\n"
+                f"{pf(dataclasses.asdict(expect_batch_request_form))}\n"
+                f"but actually has form:\n{pf(dataclasses.asdict(batch_request))}\n"
+            )
```
NathanFarmer, line 139: These were pushed down from `DataAsset` because they reference `self.datasource`.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -77,7 +80,7 @@ class BatchSorter:
     reverse: bool = False
 
 
-BatchSortersDefinition: TypeAlias = Union[List[BatchSorter], List[str]]
+BatchSortersDefinition: TypeAlias = Union[List[BatchSorter], List[str], list]
```
NathanFarmer: This typing info is now passed to the API-level. We need to add the empty `list` since we use `order_by = sorters or []`.

tests/experimental/datasources/test_schemas.py
```diff
@@ -62,15 +62,19 @@ def _sort_required_lists(schema_as_dict: dict) -> None:
...
         "SqliteTableAsset.json",
         "SqliteQueryAsset.json",
         "SASAsset.json",
-        "PandasDatasource.json",
+        "SQLDatasource.json",
+        "PostgresDatasource.json",
+        "PandasFilesystemDatasource.json",
+        "SparkDatasource.json",
+        "_PandasDatasource.json",
+        "_SparkDatasource.json",
```
NathanFarmer, line 77: `invoke schema --sync` does not update `Datasource` schemas, so we opted to ignore rather than update by hand for now.

great_expectations/experimental/datasources/file_path_data_asset.py
Kilo59: The order of types here needs to be swapped.
Pydantic will attempt to parse types according to a union in the order the union is defined, so almost any object will be coercible into a string and then pydantic will just stop there and not attempt to parse it into the more specific type.
https://docs.pydantic.dev/usage/types/#unions

Why do we want to make this a `str | Pattern` anyways? Is this just because mypy complains about regex strings?

great_expectations/experimental/datasources/file_path_data_asset.py
Kilo59: Why can't this just be a normal pydantic field?

great_expectations/experimental/datasources/interfaces.py
Kilo59: I would have expected assigning an empty list to a `List[str]` would be fine.
```python
my_list: List[str] = [] # this should be ok
```

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -139,22 +146,27 @@ class DataAsset(ExperimentalBaseModel):
     # * type: Literal["csv"] = "csv"
     name: str
     type: str
-    order_by: List[BatchSorter] = Field(default_factory=list)
+
+    order_by_setter: BatchSortersDefinition = Field(
+        default_factory=list,
+        alias="order_by",
+    )
```
Kilo59: I think I disagree or maybe just don't understand the use of `setter` fields and properties. It seems to complicate the code without an obvious benefit other than satisfying our internal type-checking. Is this something that's necessary as part of the `base_directory` change?

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -72,18 +74,18 @@ class PandasDatasourceError(Exception):
...
 
-class PandasDatasource(Datasource):
+class _PandasDatasource(Datasource):
     # class attributes
     asset_types: ClassVar[List[Type[DataAsset]]] = list(_ASSET_MODELS.values())
 
     # instance attributes
-    type: Literal["pandas"] = "pandas"
-    name: str
+    type: str = pydantic.Field("_pandas")
```
Kilo59, line 83: If this is a `_private` datasource it shouldn't need a `type` field as it won't be "registered" by the `MetaDatasource` `__new__` method.

Kilo59 (changes requested):
Most everything looks really good.
Can you revert the `@property` changes to fields like `_base_directory` `regex`/`regex_setter` and `order_by_setter`?
It's unrelated to the rest of the PR and adds unneeded complexity.

great_expectations/experimental/datasources/file_path_data_asset.py
NathanFarmer: I swapped the ordering here and in `add_csv_asset()`. Thanks for the info.

Yes, this is just about `mypy`. We could have different types for `add_csv_asset()` `regex` and `CSVAsset.regex` like it was before, but then we should do type conversion (instead of `type: ignore`) in `add_csv_asset` instead of converting in the `regex` property below here. I opted to handle it internally, abstracting this away from the end user/contributor.

[According to pydantic](https://docs.pydantic.dev/usage/types/) they just use `re.compile` for coercion, so it's not like we are getting something special by having them do the type conversion.

```
typing.Pattern
will cause the input value to be passed to re.compile(v) to create a regex pattern
```

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: It has something to do with the `Union` in the `TypeAlias`:

```
my_list_alias: TypeAlias = List[str]
my_list: Optional[my_list_alias] = None
my_list_arg = my_list or []
reveal_type(my_list_arg)  # Revealed type is "builtins.list[builtins.str]"
```

```
my_list_alias: TypeAlias = Union[List[str], List[int]]
my_list: Optional[my_list_alias] = None
my_list_arg = my_list or []
reveal_type(my_list_arg)  # Revealed type is "Union[builtins.list[builtins.str], builtins.list[builtins.int], builtins.list[Any]]"
```

Possibly a `mypy` bug...

great_expectations/experimental/datasources/pandas_datasource.py
NathanFarmer: If I change this to just `type: str` I get: `great_expectations.experimental.datasources.sources.TypeRegistrationError: '_PandasDatasource' is missing a 'type' attribute with an assigned string value`

Also, I had to change the backticks in that error message to single quotes to show it above. Do we want backticks in error messages?

NathanFarmer: Also note I had to ignore `_PandasDatasource` in `tests/experimental/datasources/test_schemas.py`.

great_expectations/experimental/datasources/file_path_data_asset.py
NathanFarmer: I'm happy to do a `re.compile` in each `add_*_asset` method if we would prefer that. I just don't like the `type: ignore`.

Kilo59: >I'm happy to do a re.compile in each add_*_asset method if we would prefer that. I just don't like the type: ignore.

Yes, let's do that instead of having property setters/getters.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: This is the same as `regex`. Not required for this PR, but results in the removal of 10 `type: ignore`s (20 total). Another case of the types revealed in the IDE being different than what we actually pass to the model. It seems sloppy to just rely on `pydantic` to coerce types and sprinkle all these `ignore`s everywhere. They will continue to proliferate. In this case we actually use a `validator` to convert to `List[BatchSorter]`, but why don't we just pass the correct type to the model?

great_expectations/experimental/datasources/file_path_data_asset.py
NathanFarmer: This one really nagged me. There are multiple issues going on here and now I'm seeing a `RuntimeWarning` due to the leading underscore field being passed to the dynamic assets.

1) A `DataAsset` can be instantiated without a `Datasource`, so we can't pull the `Datasource.base_directory` at instantiation time. Maybe this is wrong and we should be requiring a `datasource` in the `CSVAsset` args, because after all, what does it mean to have a `CSVAsset` without a `Datasource`? The property goes and looks for it when it is requested. One possible alternative solution would be to make this a `pydantic.PrivateAttr` and use a `Validator` for `datasource` that sets `_base_directory` upon `datasource` assignment. It is also important to note that `pydantic.Field`s [can't have leading underscores](https://github.com/pydantic/pydantic/issues/1476).
2) This is all complicated by the fact that changing this to `base_directory` from `_base_directory` and also updating the `FIELD_SUBSTITUTIONS` in `dynamic_pandas.py` to get rid the `RuntimeWarning` results in:
```
E   pydantic.error_wrappers.ValidationError: 1 validation error for _FilesystemDataAsset
E   type
E     field required (type=value_error.missing)
```
which is quite confusing. Maybe I'm just doing something wrong/missed a `base_directory` attribute somewhere, but I haven't figured it out.

great_expectations/experimental/datasources/file_path_data_asset.py
NathanFarmer: ✅

Kilo59 (approved):

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: Reverted this and handled it a different way. Let's discuss the `TypeAlias` `mypy` issue separately.

great_expectations/experimental/datasources/pandas_datasource.py
NathanFarmer: Per our offline conversation @Kilo59 will handle this issue in the registry.

great_expectations/experimental/datasources/interfaces.py
NathanFarmer: Decision was made to stop using the `TypeAlias` and use `List[BatchSorter] | List[str]` where the alias is currently being used.

NathanFarmer: @Kilo59 I figured it out. Instead of:
`TypeAlias = Union[List[BatchSorter], List[str]]` it should be `TypeAlias = List[Union[BatchSorter, str]]`

great_expectations/experimental/datasources/file_path_data_asset.py
NathanFarmer: Agreed to handle this in Alex's PR, which leverages DataConnector here.

PR #7081, opened 2023-02-08
[MAINTENANCE] Refactor `Store` and `StoreBackend` to leverage new CRUD methods

cdkini:
- Add `add`, `update`, and `add_or_update` logic to `Store` and `StoreBackend`
- Refactor context CRUD to leverage these changes

### Definition of Done

### DIFF
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -56,9 +56,6 @@ from great_expectations.core.config_provider import (
     _EnvironmentConfigurationProvider,
     _RuntimeEnvironmentConfigurationProvider,
 )
-from great_expectations.core.expectation_configuration import (
-    ExpectationConfiguration,  # noqa: TCH001
-)
 from great_expectations.core.expectation_validation_result import get_metric_kwargs_id
 from great_expectations.core.id_dict import BatchKwargs
 from great_expectations.core.run_identifier import RunIdentifier
@@ -73,19 +70,7 @@ from great_expectations.data_asset import DataAsset
 from great_expectations.data_context.config_validator.yaml_config_validator import (
     _YamlConfigValidator,
 )
-from great_expectations.data_context.data_context_variables import (
-    DataContextVariables,  # noqa: TCH001
-)
 from great_expectations.data_context.store import Store, TupleStoreBackend
-from great_expectations.data_context.store.expectations_store import (
-    ExpectationsStore,  # noqa: TCH001
-)
-from great_expectations.data_context.store.profiler_store import (
-    ProfilerStore,  # noqa: TCH001
-)
-from great_expectations.data_context.store.validations_store import (
-    ValidationsStore,  # noqa: TCH001
-)
 from great_expectations.data_context.templates import CONFIG_VARIABLES_TEMPLATE
 from great_expectations.data_context.types.base import (
     CURRENT_GX_CONFIG_VERSION,
@@ -123,7 +108,6 @@ from great_expectations.datasource.datasource_serializer import (
     NamedDatasourceSerializer,
 )
 from great_expectations.datasource.new_datasource import BaseDatasource, Datasource
-from great_expectations.execution_engine import ExecutionEngine  # noqa: TCH001
 from great_expectations.experimental.datasources.config import GxConfig
 from great_expectations.experimental.datasources.interfaces import Batch as XBatch
 from great_expectations.experimental.datasources.interfaces import (
@@ -162,13 +146,25 @@ except ImportError:
 if TYPE_CHECKING:
     from great_expectations.checkpoint import Checkpoint
     from great_expectations.checkpoint.types.checkpoint_result import CheckpointResult
+    from great_expectations.core.expectation_configuration import (
+        ExpectationConfiguration,
+    )
+    from great_expectations.data_context.data_context_variables import (
+        DataContextVariables,
+    )
     from great_expectations.data_context.store import (
         CheckpointStore,
         EvaluationParameterStore,
     )
+    from great_expectations.data_context.store.expectations_store import (
+        ExpectationsStore,
+    )
+    from great_expectations.data_context.store.profiler_store import ProfilerStore
+    from great_expectations.data_context.store.validations_store import ValidationsStore
     from great_expectations.data_context.types.resource_identifiers import (
         GXCloudIdentifier,
     )
+    from great_expectations.execution_engine import ExecutionEngine
     from great_expectations.render.renderer.site_builder import SiteBuilder
     from great_expectations.rule_based_profiler import RuleBasedProfilerResult
     from great_expectations.validation_operators.validation_operators import (
@@ -2742,16 +2738,14 @@ class AbstractDataContext(ConfigPeer, ABC):
 
         expectation_suite_name = expectation_suite.expectation_suite_name
         key = ExpectationSuiteIdentifier(expectation_suite_name=expectation_suite_name)
-        if (
-            self.expectations_store.has_key(key)  # noqa: W601
-            and not overwrite_existing
-        ):
-            raise gx_exceptions.DataContextError(
-                f"expectation_suite with name {expectation_suite_name} already exists."
-                " If you would like to overwrite this expectation_suite, please delete or"
-                " update it using `delete_expectation_suite` or `update_expectation_suite`, respectively."
-            )
-        self.expectations_store.set(key, expectation_suite, **kwargs)
+
+        fn: Callable
+        if overwrite_existing:
+            fn = self.expectations_store.add_or_update
+        else:
+            fn = self.expectations_store.add
+
+        fn(key=key, value=expectation_suite, **kwargs)
         return expectation_suite
 
     @public_api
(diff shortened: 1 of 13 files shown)


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2770,15 +2760,8 @@ def update_expectation_suite(
         """
         expectation_suite_name: str = expectation_suite.expectation_suite_name
         key = ExpectationSuiteIdentifier(expectation_suite_name=expectation_suite_name)
-        if not self.expectations_store.has_key(key):  # noqa: W601
-            raise gx_exceptions.DataContextError(
-                f"expectation_suite with name {expectation_suite_name} does not exist."
-            )
```
cdkini, line 2776: None of these `has_key` checks should happen in the context - these should happen at the store level

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3001,58 +2984,16 @@ def add_profiler(
         Returns:
             The persisted Profiler constructed by the input arguments.
         """
-        return self._add_profiler(
+        return RuleBasedProfiler.add_profiler(
+            data_context=self,
+            profiler_store=self.profiler_store,
```
cdkini, line 2989: I really want to migrate this logic from static RBP methods to ProfilerStore methods - this will have to be done in a later effort as it is too cumbersome at the moment

great_expectations/data_context/store/_store_backend.py
```diff
@@ -130,6 +130,32 @@ def set(self, key, value, **kwargs):
             logger.debug(str(e))
             raise StoreBackendError("ValueError while calling _set on store backend.")
 
+    def add(self, key, value, **kwargs):
```
cdkini, line 133: Left untyped as the rest of the file is untyped - I think we should definitely type this but it'll cause a bunch of cascading `mypy` issues. Can come back to this in a subsequent PR

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -177,22 +177,76 @@ def get_checkpoint(
...
+            checkpoint: The Checkpoint to persist.
+
+        Returns:
+            The persisted Checkpoint (possibly modified state based on store backend).
+
+        Raises:
+            CheckpointError: If a Checkpoint with the given name already exists.
+        """
         key = self._construct_key_from_checkpoint(checkpoint)
-        return self._add_checkpoint(key=key, checkpoint=checkpoint)
```
cdkini, line 181: Found no need to use a private here since the logic is isolated to this one func

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -1153,12 +1234,54 @@ def add_profiler(
...
 
-        response = profiler_store.set(key=key, value=config)
+        response = persistence_fn(key=key, value=config)
         if isinstance(response, GXCloudResourceRef):
             new_profiler.ge_cloud_id = response.cloud_id
 
         return new_profiler
 
+    @staticmethod
+    def _resolve_profiler_config_for_store(
```
cdkini, line 1251: Ported over from `AbstractDataContext`

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -1153,12 +1234,54 @@ def add_profiler(
                 configuration_key=config.name,
             )
 
-        response = profiler_store.set(key=key, value=config)
+        response = persistence_fn(key=key, value=config)
```
cdkini, line 1244: The only difference between `add`, `update`, and `add_or_update` is the callable we invoke at the end of this method

tests/data_context/store/test_store.py
```diff
@@ -69,3 +71,65 @@ def test_build_store_from_config_failure(store_config: dict, module_name: str):
             store_config=store_config,
             module_name=module_name,
         )
+
+
+@pytest.mark.unit
+def test_store_add_success():
```
cdkini, line 77: Testing the base class

tests/data_context/store/test_store_backends.py
```diff
@@ -1572,3 +1572,65 @@ def test_InMemoryStoreBackend_build_Key() -> None:
     assert store_backend.build_key(name=name) == DataContextVariableKey(
         resource_name=name
     )
+
+
+@pytest.mark.unit
+def test_InMemoryStoreBackend_add_success():
```
cdkini, line 1578: Testing the store backend - using `InMemory` to unit test

tests/data_context/test_data_context_state_management.py
```diff
@@ -30,9 +30,20 @@ def __init__(self) -> None:
         self.save_count = 0
         super().__init__()
 
-    def set(self, key, value, **kwargs):
+    def add(self, key, value, **kwargs):
+        ret = super().add(key=key, value=value, **kwargs)
```
cdkini, line 34: `save_count` needs to be incremented after the operation just in case it fails - this was previously erroneously signalling persistence when it didn't make sense

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -177,22 +177,76 @@ def get_checkpoint(
...
+        except gx_exceptions.StoreBackendError:
             raise gx_exceptions.CheckpointNotFoundError(
-                f"Could not find a Checkpoint named {checkpoint.name}"
+                f"Could not find an existing Checkpoint named {checkpoint.name}."
             )
-        return self._add_checkpoint(key=key, checkpoint=checkpoint)
+
+        if isinstance(checkpoint_ref, GXCloudIDAwareRef):
+            cloud_id = checkpoint_ref.cloud_id
+            checkpoint.ge_cloud_id = uuid.UUID(cloud_id)  # type: ignore[misc]
```
Kilo59: What's the misc issue here?

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -177,22 +177,76 @@ def get_checkpoint(
...
+
+        Returns:
+            The persisted Checkpoint (possibly modified state based on store backend).
+        """
         key = self._construct_key_from_checkpoint(checkpoint)
-        if self.has_key(key):
-            return self.update_checkpoint(checkpoint)
-        return self._add_checkpoint(key=key, checkpoint=checkpoint)
+
+        checkpoint_ref = self.add_or_update(key=key, value=checkpoint.get_config())  # type: ignore[func-returns-value]
```
Kilo59: If this is actually returning something shouldn't we just update the return type on the `add_or_update` method?
https://mypy.readthedocs.io/en/stable/error_code_list.html?highlight=func-returns-value#check-that-called-function-returns-a-value-func-returns-value

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -1117,15 +1118,95 @@ def run_profiler_on_data(
...
+        profiler_store: ProfilerStore,
+        data_context: AbstractDataContext,
+        name: str | None = None,
+        id: str | None = None,
+        config_version: float | None = None,
+        rules: dict[str, dict] | None = None,
+        variables: dict | None = None,
+        profiler: RuleBasedProfiler | None = None,
     ) -> RuleBasedProfiler:
+        return RuleBasedProfiler._persist_profiler(
```
Kilo59: Why are these static methods as opposed to classmethods? Aren't you using the class anyway?
```python
   @classmethod
    def update_profiler(
        cls,
        profiler_store: ProfilerStore,
        data_context: AbstractDataContext,
        name: str | None = None,
        id: str | None = None,
        config_version: float | None = None,
        rules: dict[str, dict] | None = None,
        variables: dict | None = None,
    ) -> RuleBasedProfiler:
        return cls._persist_profiler(
```

Edit: Oh I think I see, it takes `profiler` which is an instance of a `RuleBasedProfiler` 🤔 not a class?

tests/data_context/store/test_store.py
```diff
@@ -69,3 +71,65 @@ def test_build_store_from_config_failure(store_config: dict, module_name: str):
...
+    store = Store()
+    key = StringKey("foo")
+    value = "bar"
+
+    store.add(key=key, value=value)
+    assert store.has_key(key)
```
Kilo59, line 83: `non-blocking`/out-of-scope
I see this a lot... 
Why do we use `.has_key()`? 
The canonical python way to do this is with the `in` / `__contains__` operator.
https://docs.python.org/3/library/operator.html#operator.__contains__

Seems like our `stores` should implement/inherit the `Collections.ABC`.

Kilo59 (approved):
This looks great.
A couple of comments.
If it's not too difficult I'd like to see the `func-returns-value` mypy error resolved before merge.
Consider the others non-blocking.

great_expectations/rule_based_profiler/rule_based_profiler.py
cdkini: updated!

tests/data_context/store/test_store.py
cdkini: ugh yeah I'm not sure tbh - this is a super legacy API. Don and I want to change things but not sure if this is the time to do so

great_expectations/data_context/store/checkpoint_store.py
cdkini: ah yeah you are right. This is bad design here - normally return `None` but in the Cloud case, we return this ref. I'm wary of adding that annotation to the base `Store` so where might this go?

great_expectations/data_context/store/checkpoint_store.py
cdkini: This is a bug - `ge_cloud_id` is read-only. I've changed this so the value is actually updated

PR #7100, opened 2023-02-09
[FEATURE] Add polars `ColumnValuesValueLengthEquals` and `ColumnValuesValueLength` metrics

anthonyburdi:
- Add polars metrics for `great_expectations/expectations/metrics/column_map_metrics/column_value_lengths.py`
- Partial update to `great_expectations/expectations/metrics/map_metric_provider.py` to enable polars tests for `expect_column_value_lengths_to_equal.json`
- Additional tests will need to be added for conditional expectations, metrics, and more of the polars related methods in map_metric_provider.py will need to be modified. Planning to do those in a PR tomorrow.

### Definition of Done


great_expectations/expectations/metrics/column_map_metrics/column_value_lengths.py
```diff
@@ -146,6 +160,55 @@ def _pandas(
...
+        elif min_value is None and max_value is not None:
+            if strict_max:
+                metric_series = column_lengths < max_value
+            else:
+                metric_series = column_lengths <= max_value
+        elif min_value is not None and max_value is None:
+            if strict_min:
+                metric_series = column_lengths > min_value
+            else:
+                metric_series = column_lengths >= min_value
```
anthonyburdi, line 203: If this logic appears in multiple metrics I'd love to refactor it into a central place e.g. with the execution engine.

great_expectations/self_check/util.py
```diff
@@ -1237,34 +1243,30 @@ def get_test_validator_with_data(  # noqa: C901 - 31
...
                 try:
-                    type_ = np.dtype(value)
+                    type_ = np.dtype(col_type)
                 except TypeError:
                     # noinspection PyUnresolvedReferences
-                    type_ = getattr(pd, value)()
-                pandas_schema[key] = type_
+                    type_ = getattr(pd, col_type)()
+                pandas_schema[col_name] = type_
             # pandas_schema = {key: np.dtype(value) for (key, value) in schemas["pandas"].items()}
```
anthonyburdi, line 1267: Just a rename.

great_expectations/self_check/util.py
```diff
@@ -1278,47 +1280,39 @@ def get_test_validator_with_data(  # noqa: C901 - 31
...
+                        dtype = getattr(pl_datatypes, col_type)
+                    except AttributeError as e:
+                        raise AttributeError(
+                            f"{col_type} for column {col_name} is not a valid polars data type."
+                        ) from e
+
+                    pl_df.replace(
+                        col_name,
+                        pl_df.get_column(col_name).cast(dtype=dtype, strict=False),
+                    )
```
anthonyburdi, line 1313: Replacing columns with the correct data type. Polars doesn't infer a datetime from a string so we need to standardize on format in tests or we can alternatively use `dateutil.parser.parse()` to datetime. I'd rather be explicit if possible.

reqs/requirements-dev-polars.txt
```diff
@@ -1 +1 @@
-polars>=0.13.62
+polars>=0.16.2
```
anthonyburdi, line 1: Let's use latest and only downgrade if there is demand.

great_expectations/expectations/metrics/column_map_metrics/column_value_lengths.py
Shinnnyshinshin: Great idea

tests/test_definitions/test_expectations_v3_api.py
```diff
@@ -1,3 +1,11 @@
+"""Run expectation tests for expectations implemented in the v3 API.
+
+For example:
+pytest -vv -rxAP --pandas -k "[pandas/column_map_expectations/expect_column_value_lengths_to_equal:positive_test_string_one_character_length]" tests/test_definitions/test_expectations_v3_api.py
```
Shinnnyshinshin, line 4: ❤️

Shinnnyshinshin:
I'm seeing an import error for the `polars` package. I've been seeing the `try-except` approach that we see with sql and spark. Do we want to use it here too?

```
try:
    import polars as pl
    from polars import DataFrame as pl_DataFrame

except ImportError:
    pl_DataFrame = None
    pl = None
```

great_expectations/self_check/util.py
Shinnnyshinshin: I want to think about this a little bit more... I'm running into an issue with the `in_set` tests right now with this, and it's because we create the validator (which goes through this conversion) even though we suppress the test ( funny, I know 😓)

Shinnnyshinshin: On second thought, I'm with you on this. I'd rather be explicit as well. I'll tag the tests that had to be modified in the `in_set` PR.. could you let me know what you think?

Shinnnyshinshin (approved):
Thank you for getting us started ✅

anthonyburdi:
> I'm seeing an import error for the `polars` package. I've been seeing the `try-except` approach that we see with sql and spark. Do we want to use it here too?
> 
> ```
> try:
>     import polars as pl
>     from polars import DataFrame as pl_DataFrame
> 
> except ImportError:
>     pl_DataFrame = None
>     pl = None
> ```

You are absolutely right. And it passed locally because I have polars installed 🤦 . It looks like we can use the import manager as-is in metrics without running into circular import issues so I used that instead.

PR #7101, opened 2023-02-10
F/great 1393/date splitters 1

billdirks:
This is very much a work in progress but I wanted to create a PR in case it helps our discussion @NathanFarmer. Eventually this will be a refactor PR that lets us more easily add additional splitters. It does a few things:

* It defines the `test_connection` method as part of the interface for `_SQLAsset`. This defines the contract explicitly on the base class. It also makes it so we don't need to define it on all the subclasses since the default implementation is a no-op.
* This PR allows removal on code the the SQL datasource subclasses, namely sqlite here.
* It also fixes a bug where we are not currently testing the sqlite connection when we add a sqlite table asset.
* Adds some utility methods for stepping through time to make it easier to split by different time intervals.

Work I currently know is outstanding:
* I need to look into how we are currently choosing the correctly splitter when we deserialize from config. It could be we aren't doing anything now since we only have 1. 
* Testing

I haven't done any linting, typing, etc either.

### DIFF
--- a/great_expectations/experimental/datasources/sql_datasource.py
+++ b/great_expectations/experimental/datasources/sql_datasource.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import copy
+import dataclasses
 from datetime import datetime
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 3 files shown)


great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -116,14 +117,18 @@ class DatetimeRange(NamedTuple):
     max: datetime
 
 
+# BDIRKS this seems like a good model.
```
billdirks, line 120: This is just a note to myself that will get removed.

PR #7102, opened 2023-02-10
<DO_NOT_MERGE>[FEATURE] Utilize FilePathDataAsset in PandasDatasource implementations generically.</DO_NOT_MERGE>

alexsherstinsky:

-
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -94,41 +97,75 @@ def execution_engine_type(self) -> Type[PandasExecutionEngine]:
...
             base_directory=base_directory,  # type: ignore[arg-type]
             regex=regex,  # type: ignore[arg-type]  # type: ignore[arg-type]
+            glob_directive=glob_directive,
             order_by=order_by or [],  # type: ignore[arg-type]
+            data_connector=self._build_filesystem_data_connector(
+                name=name,
+                base_directory=base_directory,
+                regex=regex,
+                glob_directive=glob_directive,
+            ),
```
Kilo59, line 168: Even without looking at the rest of the code. This kind of DI/IOC is exactly what I had in mind 😄 .

alexsherstinsky: @Kilo59 Upon some thinking about this approach, what I had to do in order to make it work does not feel right.  If you look at the modules where lines are marked with my name in capitals, I had to soften the validation requirements and make the models more permissive.  But that defeats tight typing and validation.  It seems to me that the combinatoric enumeration would be a more clear and extendable way to go.  Specifically, once #7094 is merged, we would create the S3, ABS, GCS, etc. siblings of the `_FilesystemDataAsset`, each with its corresponding child of `FilePathDataConnector` (siblings of `FilesystemDataConnector`).  Then, we would use these leaf-level `DataAsset` classes in their corresponding renditions of `PandasDatasource`, inside of which `_generate_pandas_data_asset_models()` will accept them as arguments to generate the file extension specific `DataAsset` classes.  In other words, for `FilesystemPandasDatasource`, we would use
```
_ASSET_MODELS = _generate_pandas_data_asset_models(
    _FilesystemDataAsset,
    blacklist=_BLACK_LIST,
    use_docstring_from_method=True,
)
```
(which is what we have today in `PandasDatasource`, because it is hard-wired to `_FilesystemDataAsset`).  For `S3PandasDatasource`, we would use
```
_ASSET_MODELS = _generate_pandas_data_asset_models(
    _S3DataAsset,
    blacklist=_BLACK_LIST,
    use_docstring_from_method=True,
)
```
and similar pairings for GCS and ABS.  And, of course, analogously for `Spark` (i.e., `FilesystemSparkDatasource`, `S3SparkDatasource`, etc.).

This means that the total number of modules would be `4^2 = 16` (`4` cloud storage services and `2` `DataFrame` types).  While this may seem a lot, it actually is manageable as the number of cloud storage systems hosting files and computational engines that area applicable to loading files with different extensions grows very infrequently.  The advantage is that it will be easy to extend and maintain, and, quite importantly, the specifics of accessing the different storage systems (`base_directory`, `bucket`, etc.) will be validated as part of each one's dedicated `Pydantic` model.

As it stands in the present working branch, the access attributes of all storage systems are attributed to the `_FilePathDataAsset`, which is the parent class and should not have its descendants' specifics all added to it permissively.  This is, in fact, the shortcoming of the V3 `DataConnector` Marshmallow Schema class -- it holds access attributes of all of its descendants, which makes it very difficult to interpret and to use for tight validation.

Please let me know what you think.

Thank you!

Kilo59: I think that is certainly a very workable approach but I still think we can prevent this model explosion AND keep the tight typing after @NathanFarmer's https://github.com/great-expectations/great_expectations/pull/7078 work to move `base_directory` up to the `Datasource` level.

If `base_directory` and `bucket` are at the `Datasource` level, we can do the tight typing and validation there, and then the `DataAsset` can be ignorant of these details.

I didn't totally understand your last paragraph about the V3 `DataConnector`. I'm going to read some more code and then maybe it will click.

alexsherstinsky:
This work is being implemented as part of other efforts.

PR #7103, opened 2023-02-10
[FEATURE] `ToBeinSet` and `ToNotBeinSet` Metric - Polars

Shinnnyshinshin:
- Adding Metrics `column_values.in_set` and  `column_values.not_in_set` for Polars.
- JSON tests updated


tests/test_definitions/column_map_expectations/expect_column_values_to_be_in_set.json
```diff
@@ -232,7 +235,10 @@
     "tests" : [
       {
         "title": "basic_positive_test_case_datetime_set",
-        "suppress_test_for": ["sqlite"],
+        "suppress_test_for": [
+          "sqlite",
+          "polars"
+        ],
```
Shinnnyshinshin, line 241: Needing to suppress tests for `Datetime` tests. I also couldn't set the `schema` values (see line `233` of this file). It honestly feels a bit messy, but I'll think a bit more about it tonight.

great_expectations/execution_engine/polars_execution_engine.py
```diff
@@ -628,6 +628,10 @@ class MetricDomainTypes.
 
         return data, split_domain_kwargs.compute, split_domain_kwargs.accessor
 
+    def head(self, n=5):
```
Shinnnyshinshin, line 637: Noticed this was missing

great_expectations/expectations/metrics/column_map_metrics/column_value_lengths.py
```diff
@@ -1,6 +1,9 @@
 from typing import Optional
 
-import polars as pl
+try:
```
anthonyburdi: 🙇  Thank you! But let's use the import manager instead when it doesn't cause circular import issues (see https://github.com/great-expectations/great_expectations/pull/7100 for the fix).

great_expectations/expectations/metrics/column_map_metrics/column_values_in_set.py
```diff
@@ -19,6 +20,14 @@
 except ImportError:
     sa = None
 
+try:
+    import polars as pl
+    from polars import DataFrame as pl_DataFrame
+
+except ImportError:
+    pl_DataFrame = None
+    pl = None
```
anthonyburdi: Looks like you can use the import manager for polars and sqlalchemy, mind switching? Might help to centralize the import logic.

great_expectations/expectations/metrics/column_map_metrics/column_values_not_in_set.py
```diff
@@ -14,6 +15,14 @@
...
 from great_expectations.expectations.metrics.util import parse_value_set
 
+try:
+    import polars as pl
+    from polars import DataFrame as pl_DataFrame
+
+except ImportError:
+    pl_DataFrame = None
+    pl = None
+
```
anthonyburdi: If possible let's use the import manager here too.

tests/test_definitions/column_map_expectations/expect_column_values_to_be_in_set.json
anthonyburdi: Hmm - I think once my PR is merged you should be able to use `Datetime`. But it looks like the value set here is a mixed format. For other backends do we coerce whatever string we get to a datetime? AH I see - this is for `parse_strings_as_datetimes=True`. I believe that is deprecated so we should not try to support for polars.

anthonyburdi (commented):
Just a few suggestions and comments - most of it looks great though!

great_expectations/expectations/metrics/column_map_metrics/column_values_not_in_set.py
Shinnnyshinshin: done :)

great_expectations/expectations/metrics/column_map_metrics/column_values_in_set.py
Shinnnyshinshin: great catch. I'll do it now

anthonyburdi (approved):
LGTM!

PR #7109, opened 2023-02-10
[BUGFIX] Remove version from versions.json

anthonyburdi:
- versions.json gets overridden during the build process when the static version is copied from s3, it should be blank to allow local `yarn build` and `yarn start` without versions.


Rachel-Reverie (approved):
LGTM

PR #7116, opened 2023-02-10
[FEATURE] Update polars map metric providers (partial)

anthonyburdi:
- Update some polars map metric providers


Shinnnyshinshin (approved):
LGTM

PR #7123, opened 2023-02-12
[MAINTENANCE] ZEP - Realign pandas asset args for `Datasource` level `base_directory`

Kilo59:
- remove the now unused references to `base_directory` defined at the `DataAsset` level
- skip field generation for the first positional argument on the pandas `read_<type>` that corresponds to the filepath `path`
- blacklist the non-path based `read_<type>` dataframe reader methods
  -   `read_sql`, `read_sql_query`,  `read_sql_table`, `read_table`,  `read_gbq`, `read_clipboard`
  -  remove their schemas
- special case asset `kwargs` `pydantic.Field` and pass them along to the `PathBatchSpec ` in the expected way. 

### Definition of Done

Thank you for submitting!


great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -352,20 +362,26 @@ def _generate_pandas_data_asset_models(
...
         model_name = _METHOD_TO_CLASS_NAME_MAPPINGS.get(
             type_name, f"{type_name.capitalize()}Asset"
         )
 
+        if type_name in non_filebased_asset_types:
+            fields = _to_pydantic_fields(signature_tuple, skip_first_param=False)
+        else:
+            # skip the first parameter as this corresponds to the path/buffer/io field
+            # paths to specific files are provided by the batch building logic
+            fields = _to_pydantic_fields(signature_tuple, skip_first_param=True)
```
Kilo59: In essence, this is to distinguish between assets that take some kind of path (which is the vast majority of assets) and those that don't. 
As of `pandas` `1.5` ~that's only 4 types~ here are some examples.
```
("clipboard", "gbq", "sql_query", "sql_table")
```
We can generate models for them but currently, our pandas-based architecture assumes there is always some kind of path that needs to be passed so they are not actually useable. If we want to support these non-path based pandas asset types we have two options.
1. A Datasource that just takes already created in-memory `DataFrame` objects. We already have a story for this.
2. A new non-path based `PandasDatasource` class that uses a different mechanism for batching. This might also involve changes to the `PandasExecutionEngine` to support this.


Given that `1` ☝️ is the likely path forward, this extra complexity in the model generation logic doesn't seem worth it. I think we should just say that we only support path-based dataframe assets in config and blacklist the ~4~ non path-based asset types from the model generation logic.
If users really need to use a `cliboard`, `sql_query` etc. based dataframe they can create the dataframe themselves and use our runtime `DataFrame` type datasource/data-asset.

great_expectations/experimental/datasources/metadatasource.py
```diff
@@ -36,9 +36,9 @@ def __new__(
 
         cls = super().__new__(meta_cls, cls_name, bases, cls_dict)
 
-        if cls_name == "Datasource":
+        if cls_name == "Datasource" or cls_name.startswith("_"):
             # NOTE: the above check is brittle and must be kept in-line with the Datasource.__name__
-            logger.debug("1c. Skip factory registration of base `Datasource`")
+            logger.debug(f"1c. Skip factory registration of base `{cls_name}`")
```
Kilo59: move this to a small QOL PR

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -138,10 +138,6 @@ def _replace_builtins(input_: str | type) -> str | type:
 
 
 FIELD_SUBSTITUTIONS: Final[Dict[str, Dict[str, _FieldSpec]]] = {
-    # CSVAsset
-    "filepath_or_buffer": {"_base_directory": _FieldSpec(pathlib.Path, ...)},
-    # JSONAsset
-    "path_or_buf": {"_base_directory": _FieldSpec(pathlib.Path, ...)},
```
NathanFarmer, line 144: ❤️

NathanFarmer: This should get rid of our `RuntimeWarning` correct?

NathanFarmer (commented):
Before approving this, can you please explain how the `base_directory` from the `Datasource` gets passed to `pd.read_csv`, for example? It's always the first positional argument for all of these reader methods?

great_expectations/experimental/datasources/dynamic_pandas.py
Kilo59: Yes, it does! 🎊

Kilo59:
>Before approving this, can you please explain how the base_directory from the Datasource gets passed to pd.read_csv, for example? It's always the first positional argument for all of these reader methods?

Yes `base_directory` along with the `regex` is used by `_fully_specified_batch_requests_with_path()` to find matching files and then these are passed to the `PathBatchSpec`. The `path` is then given positionally by the `PandasExecutionEngine` to the actual pandas reader method.

https://github.com/great-expectations/great_expectations/blob/9506ce48a0c3fb09c3a294b606080e64b6f63a2f/great_expectations/experimental/datasources/filesystem_data_asset.py#L81-L87


https://github.com/great-expectations/great_expectations/blob/9506ce48a0c3fb09c3a294b606080e64b6f63a2f/great_expectations/experimental/datasources/filesystem_data_asset.py#L149-L159

https://github.com/great-expectations/great_expectations/blob/9506ce48a0c3fb09c3a294b606080e64b6f63a2f/great_expectations/execution_engine/pandas_execution_engine.py#L333-L339

NathanFarmer (approved):

great_expectations/experimental/datasources/dynamic_pandas.py
Kilo59: ☝️ ended up just black-listing all the non-path based asset types for this PR, and removed the above conditional.

PR #7142, opened 2023-02-14
[BUGFIX] Copy previous versions after checking out the the current commit

anthonyburdi:
- Prior versions were not appearing when building from the develop branch, but were appearing in deploy previews.
- The bug appears to be from overwriting the `versions.json` file in the repo root when checking back out the latest commit. We are now downloading and overwriting from s3 after checking out the current version.


Shinnnyshinshin (approved):
✅

PR #7151, opened 2023-02-15
[FEATURE] Update zep to use sqlalchemy_data_splitter.py

billdirks:
This refactors our sql zep splitters to use `sqlalchemy_data_splitter.py` which let's us simplify our code. We also add splitters for `year` and `year_and_month_and_day`.

### Definition of Done

Thank you for submitting!


tests/experimental/datasources/test_config.py
```diff
@@ -291,31 +290,9 @@ def test_catch_bad_top_level_config(
                 "column_splitter",
                 "method_name",
             ),
-            "unexpected value; permitted: 'split_on_year_and_month'",
+            "unexpected value; permitted:",
             id="unknown splitter method",
         ),
-        p(
```
billdirks, line 297: The `param_names` field is now gone so I've removed this test.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -207,73 +161,84 @@ def batch_request_options_template(
...
+        return self._add_splitter(
+            method_name="split_on_year_and_month_and_day", column_name=column_name
         )
 
-    def _fully_specified_batch_requests(self, batch_request) -> List[BatchRequest]:
+    def test_connection(self) -> None:
+        pass
+
+    def test_column_splitter_connection(self) -> None:
+        pass
```
NathanFarmer, line 216: I would think this shouldn't be necessary if we define `test_connection` on the splitter itself. The existing pattern is something like:
```
    ...
    if self.column_splitter:
        self.column_splitter.test_connection()
```

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -510,24 +453,9 @@ def test_connection(self) -> None:
...
-        """
-        column_splitter = SqlYearMonthSplitter(
-            column_name=column_name,
-        )
-        column_splitter.test_connection(table_asset=self)
-        self.column_splitter = column_splitter
-        return self
+    def test_column_splitter_connection(self) -> None:
+        if self.column_splitter:
+            self.column_splitter.test_connection(table_asset=self)
```
NathanFarmer, line 458: I don't see why it is necessary to refactor this out and create a new method name when right now we use `test_connection` across the board. In `Datasource.test_connection()` I added a parameter called `test_assets: bool` which can determine if the datasource assets should also be tested. Maybe we want to add `test_column_splitter: bool` as well? I feel like it should be necessary to first confirm your `DataAsset` connection is valid before you can ever test the splitter connection.

NathanFarmer (commented):
I love the pattern being set in this PR! I just have two comments/questions regarding `test_column_splitter_connection()`.

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: [This](https://github.com/great-expectations/great_expectations/blob/ad95ef6d03a5302d5b4359ebe361537df39c3b9a/great_expectations/experimental/datasources/sql_datasource.py#L605) is what I'm referring to for reference. We can do it either way, but I like how clean it is to have the same method name on each class.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: I almost removed ColumnSplitter.test_connection all together to put that logic in the TableAsset. I feel that `test_connection` on the column splitter is in the wrong level on abstraction since it doesn't know about the database or it's connection. The same column splitter should also work on arbitrary/all asset types. If an implementer is adding a new `_SQLDataAsset` subclass, they should be able to subclass `_SQLDataAsset` look at the methods that need to be implemented and be done. They shouldn't have to peer at the rest of our code and it's implementation. Currently, they have to subclass _`SQLDataAsset`, realize `ColumnSplitter.test_connection` is called and takes a parameter of the `DataAsset` type. They then have to modify `ColumnSplitter` and overload `test_connect` to be `test_connection(self, my_new_asset)` likely by adding a new `@overload` decorated empty method and updating the actual implementation. By making `test_column_splitter_connection` explicit on the `DataAsset`, the subclasser will know what they have to implement. I didn't remove `test_connection` on the ColumnSplitter since we are only testing it on table assets currently but I wanted to be explicit about what a DataAsset subclasser would have to implement.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: On the data asset level, I feel like it should be testing its own splitter. Or we could make this static.

Kilo59 (approved):

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -63,10 +56,45 @@ class SQLDatasourceWarning(UserWarning):
 class ColumnSplitter:
     column_name: str
     method_name: str
-    param_names: Sequence[str]
```
Kilo59, line 66: So we don't want to serialize this field anymore?
If we do still want this we could make this a normal field again and put the `param_defaults` logic into an `always=True` validator.
https://docs.pydantic.dev/usage/validators/#validate-always

BTW I think it's totally fine as is.

Kilo59 (approved):

great_expectations/experimental/datasources/sql_datasource.py
NathanFarmer: This makes sense to me, but it means we probably want to make a similar change in `Datasource.test_connection` for `assets`.

NathanFarmer (approved):

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Yes, it's no longer being serialized. I realized that since we've incorporated `method_name` this can be inferred. In a following PR, I when I add different non-date column splitters, the pattern to add new column splitters will be more clear.

PR #7158, opened 2023-02-15
[MAINTENANCE] Experimental `PandasDatasource`, single-batch `_PandasDataAsset`s, related schemas

NathanFarmer:
- GREAT-1623
- PR 1 of 3
- Implement `PandasDatasource` and `_PandasDataAsset` and update schemas

### Definition of Done


great_expectations/experimental/datasources/sources.py
```diff
@@ -122,25 +122,27 @@ def _register_assets(cls, ds_type: Type[Datasource], asset_type_lookup: TypeLook
...
+                        f"{asset_type.__name__} `type` field must be assigned and cannot be `None`"
                     )
                 logger.debug(
-                    f"2b. Registering `DataAsset` `{t.__name__}` as {asset_type_name}"
+                    f"2b. Registering `DataAsset` `{asset_type.__name__}` as {asset_type_name}"
                 )
-                asset_type_lookup[t] = asset_type_name
+                asset_type_lookup[
+                    (ds_type, asset_type)
+                ] = f"{ds_type.__name__}.{asset_type_name}"
```
Kilo59: I would solve this a different way, by special casing the `in-memory` `PandasDatsource` asset names rather than changing the format for every ZEP asset type.
https://github.com/great-expectations/great_expectations/pull/7158#discussion_r1108540089

great_expectations/experimental/datasources/dynamic_pandas.py
```diff
@@ -365,6 +365,7 @@ def _generate_pandas_data_asset_models(
     base_model_class: M,
     blacklist: Optional[Sequence[str]] = None,
     use_docstring_from_method: bool = False,
+    skip_first_param: bool = False,
 ) -> Dict[str, M]:
```
Kilo59: Rather than touching the `sources.py` code I would just add an optional `type_prefix`/`type_suffix` here.
```python
def _generate_pandas_data_asset_models(
    base_model_class: M,
    blacklist: Optional[Sequence[str]] = None,
    use_docstring_from_method: bool = False,
    skip_first_param: bool = False,
   # new kwarg
    type_prefix: Optional[str] = None
) -> Dict[str, M]:
    io_methods = _extract_io_methods(blacklist)
    io_method_sigs = _extract_io_signatures(io_methods)

    data_asset_models: Dict[str, M] = {}
    for signature_tuple in io_method_sigs:

        # skip the first parameter as this corresponds to the path/buffer/io field
        # paths to specific files are provided by the batch building logic
        fields = _to_pydantic_fields(signature_tuple, skip_first_param=skip_first_param)

        type_name = signature_tuple.name.split("read_")[1]
        model_name = _METHOD_TO_CLASS_NAME_MAPPINGS.get(
            type_name, f"{type_name.capitalize()}Asset"
        )

       # new code *************************************************************
       # TODO: remove this special case once we have namespace type-lookups
        if type_prefix:
            type_name = type_prefix + type_name
       # example pandas_csv / quick_csv / singlebatch_csv
      # new code *************************************************************

        try:
            asset_model = _create_pandas_asset_model(
                model_name=model_name,
                model_base=base_model_class,
                type_field=(f"Literal['{type_name}']", type_name),
```

NathanFarmer: @Kilo59 can you explain why we should use a workaround now? I assume you think we should do the namespacing some other way?

Kilo59: Because I AFIAK this is the most minimal change that allows us to achieve the goal of this PR.

The initial attempt would have changed the `type` field/name format for every zep asset class (to avoid name conflicts), and would have required changes throughout the codebase. So every asset's type field would have been `SQLiteDatasource.table`, `PandasFileSystem.csv` etc. 

And then we would change everything back once we have namespace type-lookups.

But with this change, we are just special casing the `PandasDatasource` type names so we minimize the churn within the codebase.

NathanFarmer: I see thank you!

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -54,18 +127,50 @@ class PandasDatasourceError(Exception):
...
+)
+
+_PANDAS_BLACK_LIST = (
+    # "read_csv",
+    # "read_json",
+    # "read_excel",
+    # "read_parquet",
+    # "read_clipboard",
+    # "read_feather",
+    "read_fwf",  # unhandled type
```
NathanFarmer, line 144: Tried and failed to figure out the issue here, but we are dealing with it in `PandasFilesystemDatasource` as well.

tests/experimental/datasources/test_schemas.py
```diff
@@ -69,6 +69,7 @@ def _sort_required_lists(schema_as_dict: dict) -> None:
         "SqliteTableAsset.json",
         "SqliteQueryAsset.json",
         "SASAsset.json",
+        "PandasSASAsset.json",
```
NathanFarmer, line 71: These models work, but `invoke` is failing to generate the schema.

great_expectations/experimental/datasources/pandas_datasource.py
Kilo59: There's a ticket for this. Also I think I solved this awhile back. I just need to find the stash which contained it.

Kilo59 (approved):

PR #7174, opened 2023-02-17
[MAINTENANCE] Raise store config `ClassInstantiationError` from original `DataContextError`

Kilo59:
- raise `ClassInstantiationError` from the original `DataContextError` exception
- simplify exception handling logic
- [avoid key error during exception handling](https://greatexpectationstalk.slack.com/archives/CUTCNHN82/p1676355827874889)
- create a `StoreConfigTypedDict` type

We have a lot of instances where we catch an exception and then raise another without using the `raise <NEW_EXCEPTION> from <ORGINAL_EXCEPTION>` syntax.
We should go through our codebase and fix these. See `TRY200` below.

```sh
$ ruff great_expectations --select TRY --statistics
   2	TRY002	[ ] Create your own exception
1023	TRY003	[ ] Avoid specifying long messages outside the exception class
  34	TRY004	[*] Prefer `TypeError` exception for invalid type
 221	TRY200	[ ] Use `raise from` to specify exception cause
  40	TRY201	[ ] Use `raise` without specifying exception name
  86	TRY300	[ ] Consider moving this statement to an `else` block
  34	TRY301	[ ] Abstract `raise` to an inner function
  39	TRY400	[ ] Use `logging.exception` instead of `logging.error`
```

### Definition of Done


great_expectations/data_context/store/store.py
```diff
@@ -271,16 +279,15 @@ def build_store_from_config(
...
             )
         except gx_exceptions.DataContextError as e:
-            new_store = None
             logger.critical(
                 f"Error {e} occurred while attempting to instantiate a store."
             )
-        if not new_store:
             class_name: str = store_config["class_name"]
-            module_name = store_config["module_name"]
+            module_name = store_config["module_name"]  # TODO: use .get() ???
```
Kilo59: I think this should be a `get()` but I'm not sure what the default should be.
1. `module_name`
2. `None`

What's more helpful for debugging purposes with the error that will be raised below??

great_expectations/data_context/store/store.py
```diff
@@ -20,6 +22,12 @@
 logger = logging.getLogger(__name__)
 
 
+class StoreConfigTypedDict(TypedDict):
+    class_name: str
+    module_name: str
+    store_backend: dict
```
Kilo59: I can drop these typing changes if they are not useful.

great_expectations/data_context/store/store.py
```diff
@@ -271,16 +279,15 @@ def build_store_from_config(
...
             class_name: str = store_config["class_name"]
-            module_name = store_config["module_name"]
+            module_name = store_config["module_name"]  # TODO: use .get() ???
             raise gx_exceptions.ClassInstantiationError(
                 module_name=module_name,
                 package_name=None,
                 class_name=class_name,
-            )
+            ) from e
+
```
Kilo59, line 295: This should make is easier for user's to understand what went wrong.
Looking at user error messages in the support channel, the `logger.critical` message above doesn't always make it to the user.

great_expectations/data_context/store/store.py
```diff
@@ -271,16 +282,15 @@ def build_store_from_config(
                 config_defaults=config_defaults,
             )
         except gx_exceptions.DataContextError as e:
-            new_store = None
             logger.critical(
                 f"Error {e} occurred while attempting to instantiate a store."
             )
-        if not new_store:
```
Kilo59, line 278: This conditional is not needed.

great_expectations/data_context/store/store.py
cdkini: are we always guaranteed these values? or these are just possible keys?

cdkini (approved):

great_expectations/data_context/store/store.py
Kilo59: They are not guaranteed, unfortunately.
I created this list by just creating this type and then adding fields based on the typing errors `mypy` gave me.

For example based on errors users are reporting we know that `module_name` is not always defined. What I'm not sure about is... should `module_name` always be present? In which case this `TypeDict` could help us enforce this behavior and or should we expect the `module_name` to sometimes not be present.

I'm going to add some comments and look at some more tests before merging to help answer these questions.

PR #7175, opened 2023-02-17
[FEATURE] `EphemeralDataContext.convert_to_file_context()`

cdkini:
- Add conversion mechanism between `EphemeralDataContext` and `FileDataContext` (as part of ZEP)

### Definition of Done


great_expectations/cli/datasource.py
```diff
@@ -270,7 +270,7 @@ def create_notebook(self, context: FileDataContext) -> str:
         """Create a datasource_new notebook and save it to disk."""
         renderer = self.get_notebook_renderer(context)
         notebook_path = os.path.join(
-            context.root_directory,  # type: ignore[arg-type] # could be None
```
Kilo59, line 274: 🔥 😄 🎊

NathanFarmer: 🕺🏻

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -278,7 +281,10 @@ def __init__(self, runtime_environment: Optional[dict] = None) -> None:
         )
         # Init stores
         self._stores: dict = {}
-        self._init_stores(self.project_config_with_variables_substituted.stores)
+        self._init_primary_stores(self.project_config_with_variables_substituted.stores)
+        # The DatasourceStore is inherent to all DataContexts but is not an explicit part of the project config.
```
NathanFarmer, line 285: Do we know if there is a reason for this?

great_expectations/data_context/data_context/file_data_context.py
```diff
@@ -114,17 +117,7 @@ def _init_datasource_store(self) -> None:
...
-
-    @property
-    def root_directory(self) -> Optional[str]:
-        """The root directory for configuration objects in the data context; the location in which
-        ``great_expectations.yml`` is located.
-
-        Why does this exist in AbstractDataContext? CloudDataContext and FileDataContext both use it
-
-        """
-        return self._context_root_directory
```
NathanFarmer, line 127: Love that we can remove these

great_expectations/data_context/migrator/file_migrator.py
```diff
@@ -0,0 +1,165 @@
...
+    Only takes in the necessary dependencies for conversion:
+        - context.stores
+        - context._datasource_store
+        - context.variables
+    """
+
+    def __init__(
+        self,
+        primary_stores: dict[str, Store],
+        datasource_store: DatasourceStore,
```
NathanFarmer, line 39: I definitely would like to find out why this dichotomy is needed.

great_expectations/data_context/migrator/file_migrator.py
```diff
@@ -0,0 +1,165 @@
...
+        self._primary_stores = primary_stores
+        self._datasource_store = datasource_store
+        self._variables = variables
+
+    def migrate(self, path: PathStr = pathlib.Path.cwd()) -> FileDataContext:
+        """Migrate your in-memory Data Context to a file-backed one.
+
+        Takes the following steps:
+            1. Scaffolds filesystem
+            2. Migrates primary stores (only works with default named stores)
```
NathanFarmer: Why does this only work with default named stores? Is it more accurate to say it "only creates default named stores"?

great_expectations/data_context/migrator/file_migrator.py
```diff
@@ -0,0 +1,165 @@
...
+            3. Migrates datasource store
+            4. Migrate data docs sites (both physical files and config)
+
+        Returns:
+            A FileDataContext with an updated config to reflect the state of the current context.
+        """
+        dst_context = self._scaffold_filesystem(path)
```
NathanFarmer: What does `dst` represent?

great_expectations/data_context/migrator/file_migrator.py
```diff
@@ -0,0 +1,165 @@
...
+    def _migrate_data_docs_site(
+        self, site_name: str, site_config: dict, dst_base_directory: pathlib.Path
+    ) -> dict:
+        store_backend = site_config["store_backend"]
+
+        src_site = pathlib.Path(store_backend["base_directory"])
+        dst_site = dst_base_directory.joinpath(site_name)
+
+        if src_site.exists():
+            shutil.copytree(src=str(src_site), dst=str(dst_site), dirs_exist_ok=True)
```
NathanFarmer: Why do we need `shutil.copytree` is this copying HTML?

great_expectations/data_context/store/datasource_store.py
```diff
@@ -57,18 +56,6 @@ def __init__(
...
-        """
-        See parent 'Store.list_keys()' for more information
-        """
-        keys_without_store_backend_id: List[str] = list(
-            filter(
-                lambda k: k != StoreBackend.STORE_BACKEND_ID_KEY,
-                self._store_backend.list_keys(),
-            )
-        )
-        return keys_without_store_backend_id
```
NathanFarmer, line 70: ?

NathanFarmer: I guess this does the same as in `Store`?

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -130,20 +130,22 @@ def list_keys(self, prefix: Tuple[str, ...] = ()) -> List[str]:
             A list of string keys from the user's project config.
         """
         config_section: Optional[str] = None
+        if self._resource_type is not DataContextVariableSchema.ALL_VARIABLES:
+            config_section = self._resource_type
         if prefix:
             config_section = prefix[0]
 
-        keys: List[str]
+        keys: List[Tuple]
```
NathanFarmer: Don't see why you couldn't use built-in types here.

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: We didn't want to make it a point of extensibility like the other stores - we added it as a way to interact with YAML in a clean way but users shouldn't interact with it

great_expectations/data_context/migrator/file_migrator.py
cdkini: Yeah that makes more sense - trying to figure out how to support non-default ones but it might not be worth the headache?

great_expectations/data_context/migrator/file_migrator.py
cdkini: I've seen `src` and `dst` as convention for source and destination in both Python and C source code but if this feels iffy, I'm happy to be more verbose

great_expectations/data_context/migrator/file_migrator.py
cdkini: yeah so I'm copying the site builds - is there a better approach here? Maybe copying config and running `context.build_data_docs`?

great_expectations/data_context/store/datasource_store.py
cdkini: yup

great_expectations/data_context/store/inline_store_backend.py
cdkini: true I think this whole file could use that cleanup

great_expectations/data_context/migrator/file_migrator.py
cdkini: Normal stores are configurable and sit in a dictionary (`self.stores`). The datasource store is a recent addition that is intentionally excluded from this list as we don't want users to mess around with it. As such, it sits in its own private attr (`self._datasource_store`) as is only ever configured during `__init__`.

I could go ahead and add the datasource store the the store dict but if the context treats them as two separate entities, I think we can continue that here.

Kilo59:
On my phone so it's hard to check, but do we test that this works if there is an existing ZEP/Fluent datasource on the context?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -4455,7 +4461,7 @@ def xdatasources(self) -> Dict[str, XDatasource]:
     def data_context_id(self) -> str:
         return self.variables.anonymous_usage_statistics.data_context_id  # type: ignore[union-attr]
 
-    def _init_stores(self, store_configs: Dict[str, StoreConfigTypedDict]) -> None:
```
Kilo59, line 4458: Can we keep the `StoreConfigTypedDict`?

cdkini: of course! didn't realize I removed it

tests/experimental/integration/test_migration.py
```diff
@@ -0,0 +1,31 @@
...
+)
+from great_expectations.data_context.data_context.file_data_context import (
+    FileDataContext,
+)
+from tests.test_utils import working_directory
+
+
+@pytest.mark.integration
+def test_convert_to_file_context(
+    tmpdir: py.path.local,
```
Kilo59: Can you use `tmp_path` instead of `tmpdir`

https://pytest.org/en/7.2.x/how-to/tmp_path.html#the-tmp-path-fixture

https://pytest.org/en/7.2.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures

>The tmpdir and tmpdir_factory fixtures are similar to tmp_path and tmp_path_factory, but use/return legacy [py.path.local](https://py.readthedocs.io/en/latest/path.html) objects rather than standard [pathlib.Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path) objects.

Kilo59 (approved):
Look's good to me.

1 minor thing I'd like to see updated before merging.

I'd like to see the `tmp_path` fixture used in place of `tmpdir`
https://github.com/great-expectations/great_expectations/pull/7175#discussion_r1117584658

NathanFarmer (approved):

PR #7183, opened 2023-02-19
[FEATURE] F/great 1393/add initial non datetime sql splitters

billdirks:
This adds a few column splitters for sql datasources: `ColumnSplitterColumnValue`, `ColumnSplitterDividedInteger`, `ColumnSplitterDatetimePart`.

This PR looks big by line count but most lines by far are autogenerated schemas and I've added lost of test:

* autogenerated json schemas: `5 files changed, 853 insertions(+), 28 deletions(-)`
* tests: `6 files changed, 333 insertions(+), 27 deletions(-)`
* functional code: ` 2 files changed, 282 insertions(+), 60 deletions(-)`

### Definition of Done

### DIFF
--- a/great_expectations/experimental/datasources/interfaces.py
+++ b/great_expectations/experimental/datasources/interfaces.py
@@ -275,7 +275,9 @@ class DataAsset(ExperimentalBaseModel, Generic[_DatasourceT]):
         for sorter in reversed(self.order_by):
             try:
                 batch_list.sort(
-                    key=functools.cmp_to_key(_sort_batches_with_none(sorter.key)),
+                    key=functools.cmp_to_key(
+                        _sort_batches_with_none_metadata_values(sorter.key)
+                    ),
                     reverse=sorter.reverse,
                 )
             except KeyError as e:
@@ -285,7 +287,7 @@ class DataAsset(ExperimentalBaseModel, Generic[_DatasourceT]):
                 ) from e
 
 
-def _sort_batches_with_none(
+def _sort_batches_with_none_metadata_values(
     key: str,
 ) -> Callable[[Batch, Batch], int]:
     def _compare_function(a: Batch, b: Batch) -> int:
--- a/great_expectations/experimental/datasources/interfaces.py
+++ b/great_expectations/experimental/datasources/interfaces.py
@@ -287,8 +287,8 @@ class DataAsset(ExperimentalBaseModel, Generic[_DatasourceT]):
 
 def _sort_batches_with_none(
     key: str,
-) -> Callable[[Optional[Batch], Optional[Batch]], int]:
-    def _compare_function(a: Optional[Batch], b: Optional[Batch]) -> int:
+) -> Callable[[Batch, Batch], int]:
+    def _compare_function(a: Batch, b: Batch) -> int:
         if a.metadata[key] is not None and b.metadata[key] is not None:
             if a.metadata[key] < b.metadata[key]:
                 return -1
--- a/great_expectations/experimental/datasources/interfaces.py
+++ b/great_expectations/experimental/datasources/interfaces.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
 import dataclasses
+import functools
 import logging
 from pprint import pformat as pf
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     ClassVar,
     Dict,
     Generic,
@@ -273,7 +275,7 @@ class DataAsset(ExperimentalBaseModel, Generic[_DatasourceT]):
         for sorter in reversed(self.order_by):
             try:
                 batch_list.sort(
-                    key=lambda b: b.metadata[sorter.key],
+                    key=functools.cmp_to_key(_sort_batches_with_none(sorter.key)),
                     reverse=sorter.reverse,
                 )
             except KeyError as e:
@@ -281,6 +283,27 @@ class DataAsset(ExperimentalBaseModel, Generic[_DatasourceT]):
                     f"Trying to sort {self.name} table asset batches on key {sorter.key} "
                     "which isn't available on all batches."
                 ) from e
+
+
+def _sort_batches_with_none(
+    key: str,
+) -> Callable[[Optional[Batch], Optional[Batch]], int]:
+    def _compare_function(a: Optional[Batch], b: Optional[Batch]) -> int:
+        if a.metadata[key] is not None and b.metadata[key] is not None:
+            if a.metadata[key] < b.metadata[key]:
+                return -1
+            elif a.metadata[key] > b.metadata[key]:
+                return 1
+            else:
+                return 0
+        elif a.metadata[key] is None and b.metadata[key] is None:
+            return 0
+        elif a.metadata[key] is None:  # b.metadata[key] is not None
+            return -1
+        else:  # b.metadata[key] is None, a.metadata[key] is not None
+            return 0
+
+    return _compare_function
 
 
 # If a Datasource can have more than 1 _DataAssetT, this will need to change.
--- a/great_expectations/experimental/datasources/interfaces.py
+++ b/great_expectations/experimental/datasources/interfaces.py
@@ -302,8 +302,8 @@ def _sort_batches_with_none_metadata_values(
             return 0
         elif a.metadata[key] is None:  # b.metadata[key] is not None
             return -1
-        else:  # b.metadata[key] is None, a.metadata[key] is not None
-            return 0
+        # b.metadata[key] is None, a.metadata[key] is not None
+        return 0
 
     return _compare_function
 
--- a/great_expectations/experimental/datasources/sql_datasource.py
+++ b/great_expectations/experimental/datasources/sql_datasource.py
@@ -16,11 +16,13 @@ from typing import (
 )
 
 import pydantic
-from pydantic import dataclasses as pydantic_dc
 from typing_extensions import Literal
 
 import great_expectations.exceptions as gx_exceptions
 from great_expectations.core.batch_spec import SqlAlchemyDatasourceBatchSpec
+from great_expectations.experimental.datasources.experimental_base_model import (
+    ExperimentalBaseModel,
+)
 from great_expectations.experimental.datasources.interfaces import (
     Batch,
     BatchRequest,
@@ -52,10 +54,32 @@ class SQLDatasourceWarning(UserWarning):
     pass
 
 
-@pydantic_dc.dataclass(frozen=True)
-class ColumnSplitter:
+class _ColumnSplitter(ExperimentalBaseModel):
     column_name: str
     method_name: str
+
+    @property
+    def param_names(self) -> List[str]:
+        raise NotImplementedError
+
+    def splitter_method_kwargs(self) -> Dict[str, Any]:
+        raise NotImplementedError
+
+    def request_options_to_batch_spec_kwarg_identifiers(
+        self, options: BatchRequestOptions
+    ) -> Dict[str, Any]:
+        """Translate the batch request options to the dictionary needed to by our execution engine
+
+        Arguments:
+            options: A BatchRequest.options dictionary that specified all the fields necessary
+                     to specify a batch with respect to this column splitter.
+
+        Returns:
+            A dictionary that can be added to batch_spec_kwargs["batch_identifiers"] when we
+            make a SqlAlchemyDatasourceBatchSpec and run execution_engine.get_batch_data_and_markers
+            for a particular batch specified by options.
+        """
+        raise NotImplementedError
 
     def param_defaults(self, sql_asset: _SQLAsset) -> List[Dict]:
         from great_expectations.execution_engine.split_and_sample.sqlalchemy_data_splitter import (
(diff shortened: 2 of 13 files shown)


great_expectations/experimental/datasources/interfaces.py
```diff
@@ -283,6 +287,27 @@ def sort_batches(self, batch_list: List[Batch]) -> None:
...
+) -> Callable[[Batch, Batch], int]:
+    def _compare_function(a: Batch, b: Batch) -> int:
+        if a.metadata[key] is not None and b.metadata[key] is not None:
+            if a.metadata[key] < b.metadata[key]:
+                return -1
+            elif a.metadata[key] > b.metadata[key]:
+                return 1
+            else:
+                return 0
+        elif a.metadata[key] is None and b.metadata[key] is None:
```
Kilo59, line 301: I know next to nothing about how our `metadata` works but should we be accessing these values with `.get()`?
```python
if a.metadata.get(key) is None and b.metadata.get(key) is None:
```

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -283,6 +287,27 @@ def sort_batches(self, batch_list: List[Batch]) -> None:
...
+            return 0
+        elif a.metadata[key] is None:  # b.metadata[key] is not None
+            return -1
+        else:  # b.metadata[key] is None, a.metadata[key] is not None
+            return 0
```
```suggestion
        # b.metadata[key] is None, a.metadata[key] is not None
        return 0
```

great_expectations/experimental/datasources/interfaces.py
alexsherstinsky: @billdirks I am just curious -- it this is a splitter-focused contribution, how come work on sorters is also included?  Is it needed for the new splitter functionality, or are we just including it because it is needed anyway.  Thank you.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -52,69 +54,63 @@ class SQLDatasourceWarning(UserWarning):
...
     method_name: str
 
+    @property
+    def param_names(self) -> List[str]:
+        raise NotImplementedError
+
+    def splitter_method_kwargs(self) -> Dict[str, Any]:
+        raise NotImplementedError
+
+    def request_options_to_batch_spec_kwarg_identifiers(
```
alexsherstinsky: @billdirks Would it make sense for us to spell "batch_request" out in names here for an easier interpretation so that "request" is not confused with something else (e.g., Web request)?  Otherwise, there is some risk of cognitive overload and extra time spent searching through documentation.  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -52,69 +54,63 @@ class SQLDatasourceWarning(UserWarning):
...
 
 
-@pydantic_dc.dataclass(frozen=True)
-class ColumnSplitter:
+class _ColumnSplitter(ExperimentalBaseModel):
     column_name: str
     method_name: str
 
+    @property
+    def param_names(self) -> List[str]:
```
alexsherstinsky, line 62: @billdirks What would be a helpful docstring explaining this interface method for developers needing to extend it?  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -52,69 +54,63 @@ class SQLDatasourceWarning(UserWarning):
...
+        raise NotImplementedError
+
+    def splitter_method_kwargs(self) -> Dict[str, Any]:
```
alexsherstinsky, line 70: @billdirks What would be a helpful docstring explaining this interface method for developers needing to extend it?  Since this is method outputs "kwargs", perhaps an illustrative example would make it concrete.  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -52,69 +54,63 @@ class SQLDatasourceWarning(UserWarning):
...
+        raise NotImplementedError
+
+    def request_options_to_batch_spec_kwarg_identifiers(
+        self, options: BatchRequestOptions
```
alexsherstinsky, line 83: @billdirks Not for immediately now, but in a near future, I am thinking of putting the sorting for files into DataConnector (as it was in V3).  In the DataConnector interface set, there is a method that converts `BatchDefition` to `BatchSpec`, and a private method that it uses that does exactly what this one here.  I will investigate at an opportune time if there are any commonalities and bring up any ideas for consideration (no action for now, just thinking ahead).  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -52,69 +54,63 @@ class SQLDatasourceWarning(UserWarning):
...
+            splitter_kwargs=self.splitter_method_kwargs(),
         )
         params: List[Dict] = []
-        for batch_identifier in batch_identifiers:
-            # batch_identifiers is a list of dicts where each dict contains
-            # a column name key whose value is the batch parameters
-            params.append(batch_identifier[self.column_name])
+        for identifier_data in batch_identifier_data:
+            # batch_identifiers is a list of dicts that is keyed by the column name. The value is either:
+            #  * A dict whose keys are batch request parameter names and values are the batch values.
```
alexsherstinsky, line 114: @billdirks What would be an example of `BatchRequest` parameter name and `Batch` value in this case?  Perhaps you could point to a documentation or design specifications that give some examples of when we have dictionaries and when we have single values.  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -52,69 +54,63 @@ class SQLDatasourceWarning(UserWarning):
...
+            #  * A single value for the 1 batch request parameter.
+            value = identifier_data[self.column_name]
+            if isinstance(value, dict):
+                params.append(value)
+            elif len(self.param_names) == 1:
+                params.append({self.param_names[0]: value})
+            else:
+                raise SQLDatasourceError(
+                    f"identifier_data is in an unexpected form for {self.__class__} "
+                    f"on column {self.column_name}: {identifier_data}"
```
alexsherstinsky, line 124: @billdirks Same question as just above -- where would the user go to learn the expected form of the "identifier_data"?  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -138,6 +134,158 @@ def test_connection(self, table_asset: TableAsset) -> None:
...
+    ColumnSplitterDatetimePart,
+]
+
+
+def _request_options_to_datetime_batch_spec_kwargs(
+    splitter: DatetimeColumnSplitter, options: BatchRequestOptions
+) -> Dict[str, Any]:
+    identifiers: Dict = {}
+    for part in splitter.param_names:
+        if part not in options:
```
alexsherstinsky, line 244: @billdirks As a small potential optimization, there may be a way to ensure at the top of this method that the parts (i.e., `splitter.param.names`) are a subset of options and point out the outliers (we have code like that in `ExecutionEngine` implementations).  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -138,6 +134,158 @@ def test_connection(self, table_asset: TableAsset) -> None:
...
+        self, options: BatchRequestOptions
+    ) -> Dict[str, Any]:
+        if "quotient" not in options:
+            raise ValueError(
+                "'quotient' must be specified in the batch request options to create a batch identifier"
+            )
+        return {self.column_name: options["quotient"]}
+
+
+ColumnSplitter = Union[
```
alexsherstinsky, line 293: @billdirks There seems to be some overlap between the members of `ColumnSplitter` and `DatetimeColumnSplitter`; would it be useful to somehow combine them in a purely composable way?  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -206,7 +355,37 @@ def add_splitter_year_and_month_and_day(
...
+
+    def add_splitter_column_value(self, column_name: str) -> _SQLAsset:
+        return self._add_splitter(
+            ColumnSplitterColumnValue(
+                method_name="split_on_column_value",
+                column_name=column_name,
+            )
+        )
+
+    def add_splitter_divided_integer(self, column_name: str, divisor: int) -> _SQLAsset:
```
alexsherstinsky, line 396: @billdirks There is also a "mod integer" splitter in V3 (we would need to check the code and documentation for any others that could be useful candidates for adding here as well).  Thanks.

tests/experimental/datasources/integration/test_integration_datasource.py
```diff
@@ -111,3 +111,125 @@ def test_sql_query_data_asset(empty_data_context):
...
+    last_specified_batch_metadata,
+):
+    context = empty_data_context
+    datasource = sqlite_datasource(context, database)
+    asset = datasource.add_table_asset(
+        name="table_asset",
+        table_name=table_name,
+    )
+    getattr(asset, splitter_name)(**splitter_kwargs)
+    asset.add_sorters(sorter_args)
```
alexsherstinsky, line 226: @billdirks It seems that the assertions below would work even without adding the sorters; is this correct?  If affirmative, then perhaps we could remove that line (or add a comment for those, like me, wondering).  If not, where should I look in order to see what is actually happening?  Thanks.

tests/experimental/datasources/conftest.py
```diff
@@ -88,7 +91,15 @@ def __init__(self, attributes):
                     for k, v in attributes.items():
                         setattr(self, k, v)
 
-            return [Row(row_dict) for row_dict in splitter_query_response]
+            # We know that splitter_query_response is non-empty because of validation
+            # at the top of the outer function.
+            # In some cases, such as in the datetime splitters,
+            # a dictionary is returned our from out splitter query with the key as the parameter_name.
+            # Otherwise, a list of values is returned.
+            if isinstance(splitter_query_response[0], dict):
```
alexsherstinsky, line 99: @billdirks Perhaps for stylistic purposes, if there is a `return` as part of the `if` clause, then the `else` is not needed -- just another `return` underneath (because it will logically follow).  Thank you.

great_expectations/experimental/datasources/interfaces.py
billdirks: The splitters determine which fields can be sorted on. Some entries in our non-datetime columns are null which can cause integration tests to fail. This fixes that bug so I can include the non-datetime splitters in the integration tests.

great_expectations/experimental/datasources/interfaces.py
billdirks: Done.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Please see the [test here](https://github.com/great-expectations/great_expectations/pull/7183/files#diff-d17bd5ef95c5457abc6a63074ed41560a076e1f8209dd0f3fc95feb6fddeb8f4R1035). 
This is a shim to [our current splitter code](https://github.com/great-expectations/great_expectations/blob/f0e89bdbaa491f39ba06f3d7af3d0b4d31d425bf/great_expectations/execution_engine/split_and_sample/sqlalchemy_data_splitter.py) so you can look there for details. If there are design documents, they predate me and I haven't seen them.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: No user should ever need to worry about the form of `identifier_data`. This is an internal implementation detail and is a shim to our current system splitter code which is located [here](https://github.com/great-expectations/great_expectations/blob/f0e89bdbaa491f39ba06f3d7af3d0b4d31d425bf/great_expectations/execution_engine/split_and_sample/sqlalchemy_data_splitter.py)

great_expectations/experimental/datasources/sql_datasource.py
billdirks: These are distinct types. How do you propose we compose them? Note a union of union types is not equivalent to unioning the subtypes together.

tests/experimental/datasources/conftest.py
billdirks: Done.

tests/experimental/datasources/integration/test_integration_datasource.py
billdirks: The assertion at line 235 will fail. I am grabbing the last batch to make assertions. To guarantee which batch is last, I sort them.

tests/experimental/datasources/test_config.py
```diff
@@ -323,6 +321,7 @@ def test_catch_bad_asset_configs(
         if expected_error_loc == all_errors[0]["loc"]:
             test_msg = error["msg"]
             break
+    # BDIRKS this error message is now different
```
billdirks: I have removed this line.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Done. Also, please see the note above as the this not currently being a point of extensibility.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: Done.

Note, we decided in an early discussion that column splitters are not a point of extension for contributors since this is tightly couple to non-public api code, namely, [this file](https://github.com/great-expectations/great_expectations/blob/f0e89bdbaa491f39ba06f3d7af3d0b4d31d425bf/great_expectations/execution_engine/split_and_sample/data_splitter.py) and the implementations in that directory. So the typical developer extending this is the core developers.

great_expectations/experimental/datasources/sql_datasource.py

great_expectations/experimental/datasources/sql_datasource.py
billdirks: I am not sure this would be an optimization. It would slow it down in the case where it is a subset which is likely the most common use case since it is configured once then run multiple times. In the case where is it is not a subset, it prevents a handful (at most 5) dictionary assignments.

great_expectations/experimental/datasources/interfaces.py
billdirks: Actually, it may have been a postgres dataset with nulls in it that hit this crasher (our integration tests currently only use a sqlite dataset). I don't recall which dataset it was.

billdirks: @Kilo59 The keys are determined by the splitter (in sql, in pandas/spark it is the regex) so should always be present. If they aren't then that is a bug.

The value may be `None` if that column was `NULL` in a database row (or empty in the csv).

great_expectations/experimental/datasources/sql_datasource.py
billdirks: I plan to add the remaining splitters in the next PR. There are a few more.

Kilo59 (approved):

tests/experimental/datasources/conftest.py
alexsherstinsky: 🙇

tests/experimental/datasources/integration/test_integration_datasource.py
alexsherstinsky: 🙇

alexsherstinsky (approved):
LGTM -- thank you!

PR #7197, opened 2023-02-22
[FEATURE] Provide S3 access/operations to new PandasDatasource

alexsherstinsky:
### Scope
This change refactors the New Datasources codebase to introduce Filesystem and S3 versions of PandasDatasource all operating through `FilePathDataAsset` (and eliminates the need for `FilesystemDataAsset`).

A `Datasource` can initialize a `DataConnector` for any of its `DataAsset` constituents.  A `DataAsset` can then use its `DataConnector` to do connection tests; convert `BatchRequest` to series of `BatchSpec` objects; and so on through the high-level `DataConnector` interfaces.  Adding cloud storage support using `DataConnector` abstraction is very low effort. 

- JIRA: GREAT-1692/GREAT-1690/GREAT-1691/GREAT-1619
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


tests/experimental/datasources/test_pandas_s3_datasource.py
```diff
@@ -0,0 +1,267 @@
...
+def _get_region_name() -> str:
+    return "us-east-1"
+
+
+def _get_boto3_client() -> BaseClient:
+    client = boto3.client("s3", region_name=_get_region_name())
+    return client
+
+
+def _build_pandas_s3_datasource() -> PandasS3Datasource:
```
alexsherstinsky, line 57: I could not use fixtures here, because no matter in which order I tried them, the `@fixture` decorator and the `@mock_s3` decorator would not co-exist.  Hence, I converted fixtures into the private "_build" type utility methods.

tests/experimental/datasources/test_pandas_s3_datasource.py
```diff
@@ -0,0 +1,267 @@
...
+        r"(?P<name>.+)_(?P<ssn>\d{9})_(?P<timestamp>.+)_(?P<price>\d{4})\.csv"
+    )
+    data_connector: S3DataConnector = cast(S3DataConnector, asset._data_connector)
+    test_connection_error = TestConnectionError(
+        f"""No file in bucket "{asset.datasource.bucket}" with prefix "{data_connector._prefix}" matched regular expressions pattern "{regex.pattern}" using deliiter "{data_connector._delimiter}" for DataAsset "{asset.name}"."""
+    )
+    return regex, test_connection_error
+
+
+@pytest.mark.integration
```
alexsherstinsky, line 109: Please note that all tests are marked "integration" here, because of the "S3" access (albeit a "mock" one).  Thanks.

great_expectations/experimental/datasources/file_path_data_asset.py
```diff
@@ -47,7 +46,14 @@
 class _FilePathDataAsset(DataAsset):
     _EXCLUDE_FROM_READER_OPTIONS: ClassVar[Set[str]] = {
         "name",
-        "regex",
+        "regex",  # file_path argument
+        "base_directory",  # filesystem argument
```
alexsherstinsky: Based on discussions with @Kilo59 we determined that we can avoid combinatorial enumeration of data assets for every storage system by sharing the settings between the `Datasource` and the `add_<type>_asset()` method, such that the `DataConnector` for the `DataAsset` could be instantiated in order to provide (via the top-level `DataConnector` interface methods) all utilities needed by `DataAsset` to access files on different storage environments.  As a consequence, the exclusions in `FilePathDataAsset` have to account for the settings of all supported storage environments.  In addition, top-level `Datasource.add_asset()` method must accept additional arguments (such as instantiated `DataConnector`).

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -306,9 +306,6 @@ class _SQLAsset(DataAsset):
     column_splitter: Optional[ColumnSplitter] = None
     name: str
 
-    # Internal attributes
```
alexsherstinsky, line 309: This extra declaration is not needed -- top-level `_DataAsset` already declares a `_datasource` private variable (and we are not counting on any special methods that might be found in the `SQLDatasource` extension).  Hence, deleting this is acceptable (and tests pass to confirm this).

great_expectations/experimental/datasources/file_path_data_asset.py
```diff
@@ -65,7 +71,8 @@ class _FilePathDataAsset(DataAsset):
     _all_group_index_to_group_name_mapping: Dict[int, str] = pydantic.PrivateAttr()
     _all_group_names: List[str] = pydantic.PrivateAttr()
 
-    _data_connector: DataConnector | None = pydantic.PrivateAttr()
+    _data_connector: DataConnector = pydantic.PrivateAttr()
```
alexsherstinsky, line 73: We redeclare `_data_connector` here, because in `_DataAsset`, it is optional by design (but here it is required, because all File-Path DataAsset implementations rely on a `DataConnector`).

great_expectations/experimental/datasources/file_path_data_asset.py
```diff
@@ -65,7 +71,8 @@ class _FilePathDataAsset(DataAsset):
...
+    _test_connection_error_message: str = pydantic.PrivateAttr()
```
alexsherstinsky, line 74: We redeclare `_test_connection_error_message` here, because in _DataAsset, it is optional by design (but here it is required, because all File-Path DataAsset implementations utilize a `_test_connection_error_message`).

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -339,7 +344,13 @@ class Datasource(
...
         "execution_engine",
         "assets",
-        "base_directory",
+        "base_directory",  # filesystem argument
+        "glob_directive",  # filesystem argument
+        "data_context_root_directory",  # filesystem argument
+        "bucket",  # s3 argument
+        "prefix",  # s3 argument
+        "delimiter",  # s3 argument
+        "max_keys",  # s3 argument
```
alexsherstinsky, line 350: Based on discussions with @Kilo59 we determined that we can avoid combinatorial enumeration of data assets for every storage system by sharing the settings between the `Datasource` and the add_<type>_asset() method, such that the `DataConnector` for the `DataAsset` could be instantiated in order to provide (via the top-level `DataConnector` interface methods) all utilities needed by `DataAsset` to access files on different storage environments.  As a consequence, the exclusions here in `Datasource` have to account for the settings of all supported storage environments.  In addition, the `Datasource.add_asset()` method here, in addition to the `asset` argument, must accept additional arguments (such as instantiated `DataConnector` and `test_connection_error_message`).

great_expectations/experimental/datasources/data_asset/data_connector/s3_data_connector.py
```diff
@@ -13,19 +13,26 @@
...
+    from botocore.client import BaseClient
+except ImportError:
+    logger.debug(
+        "Unable to load boto3 or botocore; install optional boto3 and botocore dependencies for support."
+    )
+    boto3 = None
+    botocore = None
+    BaseClient = None
+
+
```
Kilo59: I really dislike this style of importing and then setting things to null if they aren't importable.
Are any of these actually used at runtime?
Can we get rid of this block and only import during `TYPE_CHECKING`
```python
if TYPE_CHECKING:
    from great_expectations.core.batch import BatchDefinition
    import boto3
    import botocore
    from botocore.client import BaseClient
```

great_expectations/experimental/datasources/file_path_data_asset.py
Kilo59: I don't undestand why this is necessary. All these are arguments for the injected `DataConnector`, they don't need to be fields/attributes on the `DataAsset`.

great_expectations/experimental/datasources/interfaces.py
Kilo59: Same response as above - https://github.com/great-expectations/great_expectations/pull/7197/files#r1114474090
These fields should not exist on the `DataAsset` they are simply use to create the appropriate `DataConnector`.

great_expectations/experimental/datasources/interfaces.py
```diff
@@ -424,20 +435,33 @@ def get_asset(self, asset_name: str) -> _DataAssetT:
                 f"'{asset_name}' not found. Available assets are {list(self.assets.keys())}"
             ) from exc
 
-    def add_asset(self, asset: _DataAssetT) -> _DataAssetT:
+    def add_asset(
+        self,
+        asset: _DataAssetT,
+        data_connector: Optional[DataConnector] = None,
```
Kilo59, line 438: Where is the code that actually creates the `DataConnector` instance?

great_expectations/experimental/datasources/data_asset/data_connector/filesystem_data_connector.py
```diff
@@ -36,8 +36,8 @@ def __init__(
         self,
         datasource_name: str,
         data_asset_name: str,
-        base_directory: pathlib.Path,
         regex: re.Pattern,
+        base_directory: pathlib.Path,
```
Kilo59, line 40: I would try to resist the urge to make these minor stylistic changes unless there is a compelling reason.
It makes it harder to follow the git history, and is more likely to lead to git conflicts.

Kilo59 (changes requested):
I'd like to see the issue of having to enumerate all the possible `DataConnector` arguments in `_EXCLUDE_FROM_READER_OPTIONS` addressed.

I think this is a critical part of the DI/IOC design that we need to get right.

Also please update with `develop`, I know it's tedious but it should make this PR smaller and easier to review.

great_expectations/experimental/datasources/pandas_filesystem_datasource.py
```diff
@@ -133,12 +78,24 @@ def add_csv_asset(
...
 
-        return self.add_asset(asset)
+        data_connector: DataConnector = FilesystemDataConnector(
+            datasource_name=self.name,
+            data_asset_name=name,
+            regex=regex,
+            base_directory=self.base_directory,
+            glob_directive=glob_directive,
+            data_context_root_directory=self.data_context_root_directory,
+        )
```
Kilo59: Mostly a note for myself. This is where the `DataConnector` is being created.

great_expectations/experimental/datasources/file_path_data_asset.py
alexsherstinsky: @Kilo59 I placed those exclusions with comments, to prevent such strange errors as `ExecutionEngine.__init__()` is not aware of argument “glob_directive” (I am paraphrasing).  If there is a better way to address this issue?  I cleaned up as much as I could (with now a better understanding with how we are using `Pydantic` -- and it concerns me a little but that we use `extra = pydantic.Extra.allow` here and there, which defeats the premise of strong validation; however, we can defer any actions regarding this for now).

great_expectations/experimental/datasources/interfaces.py
alexsherstinsky: @Kilo59 Please see above and let me know if you have guidance; otherwise, I keep running into "constructor does not know about this argument" type runtime errors.  Thank you.

great_expectations/experimental/datasources/interfaces.py
alexsherstinsky: @Kilo59 It is in our `Datasource` subclasses, which are the ones that know exactly which `DataConnector` variant they should instantiate and have all the lowest-level arguments to construct them.

great_expectations/experimental/datasources/data_asset/data_connector/filesystem_data_connector.py
alexsherstinsky: @Kilo59 If you re referring to the changed order, then it had to be done for better clarity and comprehension of the story that this code tells (this early in the development, unfortunately an earlier contribution did not get it right).  Having said that, I agree that we should have a justification for all changes, including cosmetic ones, such as this.  Thanks.

great_expectations/experimental/datasources/pandas_filesystem_datasource.py
alexsherstinsky: @Kilo59 Yes -- and I believe that it is in line with our idea that using this coding pattern we can reduce the combinatorial enumeration of leave `DataAsset` classes.  Thank you.

great_expectations/experimental/datasources/data_asset/data_connector/s3_data_connector.py
alexsherstinsky: @Kilo59 I agree with you (please see the latest version -- we still need an isolated import for `boto3`, but I was able to put the other ones into `TYPE_CHECKING`, where they belong).  Thank you.

alexsherstinsky:
> I'd like to see the issue of having to enumerate all the possible `DataConnector` arguments in `_EXCLUDE_FROM_READER_OPTIONS` addressed.
> 
> I think this is a critical part of the DI/IOC design that we need to get right.
> 
> Also please update with `develop`, I know it's tedious but it should make this PR smaller and easier to review.

@Kilo59 Not tedious for me at all -- I am just working on another change (path inclusion), and then will update.

Regarding, `_EXCLUDE_FROM_READER_OPTIONS`, I would also like to get to the bottom of it.  In my opinion, it can be done as a follow-up pull request after this one is merged, because I am just following the existing coding pattern.  So if all other goals of this pull request are satisfied, I would rather merge it, because it already brings quite a few enhancements.  Thank you.

great_expectations/experimental/datasources/data_asset/data_connector/file_path_data_connector.py
```diff
@@ -73,13 +76,15 @@ def __init__(
             data_asset_name=data_asset_name,
         )
 
-        self._regex: re.Pattern = regex
+        self._unnamed_regex_group_prefix: str = unnamed_regex_group_prefix
+        self._regex: re.Pattern = self._insure_regex_groups_include_data_reference_key(
```
NathanFarmer: Nitpick: I think this should be `ensure` not `insure`.

great_expectations/experimental/datasources/data_asset/data_connector/file_path_data_connector.py
```diff
@@ -247,7 +252,30 @@ def _generate_batch_spec_parameters_from_batch_definition(
 
         path = self._get_full_file_path(path=path)
 
-        return {"path": path}
+        return {FilePathDataConnector.FILE_PATH_BATCH_SPEC_KEY: path}
+
+    def _insure_regex_groups_include_data_reference_key(
```
NathanFarmer: `ensure`

great_expectations/experimental/datasources/data_asset/data_connector/file_path_data_connector.py
```diff
@@ -247,7 +252,30 @@ def _generate_batch_spec_parameters_from_batch_definition(
...
+    ) -> re.Pattern:
+        """
+        Args:
+            regex: regex pattern for filtering data references; if reserved group name "path" (FILE_PATH_BATCH_SPEC_KEY)
+            is absent, then it is added to enclose original regex pattern, supplied on input.
+
+        Returns:
+            Potentially modified Regular Expression pattern (with enclosing FILE_PATH_BATCH_SPEC_KEY reserved group)
+        """
+        regex_parser: RegExParser = RegExParser(
```
NathanFarmer: Did you need to put this type annotation here due to the type checker? Otherwise I would remove it.

great_expectations/experimental/datasources/file_path_data_asset.py
```diff
@@ -49,17 +51,20 @@
...
+        "bucket",  # s3 argument
+        "prefix",  # s3 argument
+        "delimiter",  # s3 argument
+        "max_keys",  # s3 argument
         "kwargs",  # kwargs need to be unpacked and passed separately
     }
 
-    _ALWAYS_INCLUDE_IN_BATCH_REQUEST_OPTIONS_TEMPLATE: ClassVar[Set[str]] = {
-        "path",
-    }
```
NathanFarmer, line 61: 👍🏻

great_expectations/experimental/datasources/pandas_s3_datasource.py
```diff
@@ -0,0 +1,290 @@
...
+    def add_parquet_asset(
+        self,
+        name: str,
+        regex: Union[str, re.Pattern],
+        prefix: str = "",
+        delimiter: str = "/",
+        max_keys: int = 1000,
+        order_by: Optional[BatchSortersDefinition] = None,
+        **kwargs,  # TODO: update signature to have specific keys & types
+    ) -> ParquetAsset:  # type: ignore[valid-type]
```
NathanFarmer: Is this just for IDE support, because I'm seeing tab completion in jupyter without this I think.

NathanFarmer (approved):

great_expectations/experimental/datasources/data_asset/data_connector/file_path_data_connector.py
alexsherstinsky: @NathanFarmer No, my old habit (deleting -- in next commit).  Thanks.

great_expectations/experimental/datasources/data_asset/data_connector/file_path_data_connector.py
alexsherstinsky: @NathanFarmer Oops!  I did it again!  Fixing!  Thank you!

great_expectations/experimental/datasources/data_asset/data_connector/file_path_data_connector.py
alexsherstinsky: @NathanFarmer This I should have known -- typo; thanks!

great_expectations/experimental/datasources/pandas_s3_datasource.py
alexsherstinsky: @NathanFarmer Which part of this method definition is raising questions?  Sorry if I am staring at the obvious.  Thanks.

alexsherstinsky:
@Kilo59 I have a hunch regarding what is happening with `_EXCLUDE_FROM_READER_OPTIONS` -- the problem lies in the fact that we are using the `Pydantic` `Model` method `self.dict()` to put together `reader_options` for `ExecutionEngine`.  So `self.dict()` gathers all properties, subject to its inclusion/exclusion directives.  Could we instead create an interface that `Datasource` subclasses would implement, since `PandasDatasource` knows what reader options the `pd.` methods accept, `SparkDatasource` knows what reader options the `pyspark.sql.` methods accept, and so on?  In any case, we can do it in a subsequent pull request (this one should not be gated on this, because we simply increased the already existing `_EXCLUDE_FROM_READER_OPTIONS` list.  In fact, what would be even better is that each appropriate `Datasource` subclass declares its own exclusion list, and the `DataAsset` combines it with its own.  I may implement this as part of this PR (or the very next one if this one is merged). Thank you.

great_expectations/experimental/datasources/file_path_data_asset.py
Kilo59: This issue is that these should not be fields/attrs on the `DataAsset` at all. They are only needed by the `DataConnector`.

great_expectations/experimental/datasources/pandas_filesystem_datasource.py
Kilo59: The `DataConnector` takes in the needed parameters from the `DataSource`, the `DataAsset` itself does not need a `base_directory` or `s3_bucket`, etc. parameter, so you shouldn't need to exclude them from any `dict()` serialization.

Kilo59 (changes requested):
Still need to see this merged with develop and the removal of the extra `_EXCLUDE_FROM_READER_OPTIONS`

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -195,13 +195,35 @@ def _validate_batch_request(self, batch_request: BatchRequest) -> None:
...
+    # "read_pickle",
+    # "read_sas",  # invalid json schema
+    # "read_spss",
+    "read_sql",  # not path based & type-name conflict
+    "read_sql_query",  # not path based
+    "read_sql_table",  # not path based
+    "read_table",  # type-name conflict
+    # "read_xml",
+)
+
```
Kilo59: Why did you move this here?
I think this belongs in `pandas_file_path_datasource.py`

alexsherstinsky: Yes -- good catch -- must be from a merge.  I will fix.

tests/experimental/datasources/test_pandas_filesystem_datasource.py
```diff
@@ -418,24 +419,17 @@ def test_get_batch_list_from_fully_specified_batch_request(
...
-    assert batch.metadata == {
-        "year": "2018",
-        "month": "04",
-        "base_directory": pandas_filesystem_datasource.base_directory,
-        "path": pathlib.Path(path),
-    }
+
+    path = "yellow_tripdata_sample_2018-04.csv"
+    assert batch.batch_request.options == {"path": path, "year": "2018", "month": "04"}
+    assert batch.metadata == {"path": path, "year": "2018", "month": "04"}
```
Kilo59, line 431: `non-blocking`
This feels like we are going a bit backwards. Why is this no a `str` rather then a `pathlib.Path` object?

tests/experimental/datasources/data_asset/data_connector/test_s3_data_connector.py
```diff
@@ -19,6 +20,20 @@
...
+    import boto3
+    import botocore
+    from botocore.client import BaseClient
+except ImportError:
+    logger.debug(
+        "Unable to load boto3 or botocore; install optional boto3 and botocore dependencies for support."
+    )
+    boto3 = None
+    botocore = None
+    BaseClient = None
```
Kilo59: Why do we need this try import for tests?
Also I presume some of these are type-checking only imports?

great_expectations/experimental/datasources/sources.py
```diff
@@ -154,7 +154,6 @@ def _register_datasource_and_factory_method(
 
     @classmethod
     def _register_assets(cls, ds_type: Type[Datasource], asset_type_lookup: TypeLookup):
-
```
Kilo59, line 157: @alexsherstinsky this is the kind of change we should avoid making.
It needlessly adds this file to the list of changed files, and isn't obviously better than what existed previously. 
All these little changes add up and make it harder to navigate the git history of a file.

tests/experimental/datasources/test_pandas_s3_datasource.py
Kilo59: You should be able to use `mock_s3` as a context manager.
https://docs.getmoto.org/en/latest/docs/getting_started.html#context-manager

If it were me I would probably use the context manager to define a `mock_s3` fixture which could then be combined with other fixtures (or tests).

great_expectations/experimental/datasources/pandas_filesystem_datasource.py
```diff
@@ -140,12 +84,24 @@ def add_csv_asset(
...
-        return self.add_asset(asset)
+        data_connector: DataConnector = FilesystemDataConnector(
+            datasource_name=self.name,
+            data_asset_name=name,
+            batching_regex=batching_regex_pattern,
+            base_directory=self.base_directory,
+            glob_directive=glob_directive,
+            data_context_root_directory=self.data_context_root_directory,
+        )
+        test_connection_error_message: str = f"""No file at base_directory path "{self.base_directory.resolve()}" matched regular expressions pattern "{batching_regex_pattern.pattern}" and/or glob_directive "{glob_directive}" for DataAsset "{name}"."""
```
Kilo59, line 99: `non-blocking`
This seems like the kind of thing that could be defined as an attribute on the `FilesystemDataConnector` class itself.
Then we wouldn't need to pass it.

Maybe not though 🤔

great_expectations/experimental/datasources/interfaces.py
Kilo59: `non-blocking`
On further review, I understand why this is on the `Datasource`.
These are `_EXCLUDED_EXEC_ENG_ARGS`  on the `Datasource`
This is not the `_EXCLUDE_FROM_READER_OPTIONS` attribute on the pandas `DataAsset`.

Still, I think we could make this unnecessary by using a `data_connecter_options` / `data_connecter_parameter` object. I'm happy to save that for a followup PR though.

alexsherstinsky:
@Kilo59 Yes, you and I agree on the reasoning, captured in `https://github.com/great-expectations/great_expectations/pull/7197#discussion_r1116098706`, and I am happy to finish it in my next increment.  Thank you.

great_expectations/experimental/datasources/pandas_filesystem_datasource.py
alexsherstinsky: @Kilo59 Interesting idea, and technically -- obviously -- easy to do.  However, the `DataConnector` only does the low-level access work; it does not have the responsibility to manage helpful messages to users.  In addition, turning this message, albeit tiny in terms of computation work, is borderline not a good fit for a property (properties must not build or compute anything; they should just retrieve a result, with only minor work, best avoided entirely).  Thanks.

great_expectations/experimental/datasources/sources.py
alexsherstinsky: @Kilo59 Sometimes it is worth incurring these changes early on in an effort that is still maturing so that we (or contributors) do not have to do a lot of them later.  Stylistic tidiness and logical organization along with verbose naming can reduce cognitive strain on those trying to learn the system.  Of course these are only my own aesthetic preferences and I can see how others may find more value in fewer changes; having said that, small changes can be understood and checked off quickly.   Not sure if these kind of formatting items need to be in a style guide, but happy to discuss.  Thank you!

tests/experimental/datasources/data_asset/data_connector/test_s3_data_connector.py
alexsherstinsky: @Kilo59 Good catch -- sorry again, this is due to a bad merge.  Fixed (will be committed after the release).  Thank you.

tests/experimental/datasources/test_pandas_filesystem_datasource.py
alexsherstinsky: @Kilo59 My understanding is that this is metadata, where all values were strings up to now, because they resulted from partitioning the data reference using RegEx.  Whether or the `path` addition should also be a string may need some discussion.  I would say that it should be a string, because to the user the data references for files are strings.

tests/experimental/datasources/test_pandas_s3_datasource.py
alexsherstinsky: @Kilo59 This sounds like a good idea -- I will include the improvement you are recommending as part of the next pull request.  Thank you.

great_expectations/experimental/datasources/file_path_data_asset.py
alexsherstinsky: @Kilo59 As of now, we have solved the problem, subject to a potential refactor (as per https://github.com/great-expectations/great_expectations/pull/7197#discussion_r1116098706).  Thanks.

Kilo59 (approved):
I'd like to see the use of `mock_s3` adjusted.
But I think that's fine to do as a followup.
https://github.com/great-expectations/great_expectations/pull/7197#discussion_r1116071832

alexsherstinsky:
> I'd like to see the use of `mock_s3` adjusted. But I think that's fine to do as a followup. [#7197 (comment)](https://github.com/great-expectations/great_expectations/pull/7197#discussion_r1116071832)

@Kilo59 I can show you what I have done for this as part of the next PR (at first commit), but it is not looking promising.

PR #7203, opened 2023-02-22
[FEATURE] F/great 1393/add more zep column splitters

billdirks:
This adds the following 3 column splitters:
`ColumnSplitterModInteger`
`ColumnSplitterHashedColumn`
`ColumnSplitterConvertedDateTime`


great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -402,6 +484,53 @@ def add_splitter_divided_integer(self, column_name: str, divisor: int) -> _SQLAs
...
+        )
+
+    def add_splitter_hashed_column(
+        self, column_name: str, hash_digits: int
+    ) -> _SQLAsset:
+        from great_expectations.experimental.datasources.sqlite_datasource import (
+            SqliteDatasource,
+        )
+
+        if not isinstance(self._datasource, SqliteDatasource):
```
billdirks: This isn't ideal but is the cleanest way to fail early here. A similar exception is raised inside of the `sqlalchemy_data_splitter.py` module and I wanted to raise earlier. Alternatives:

* raise late
* build out some sqlite specific infrastructure. Since this is at the `_SQLAsset` level, it is more involved than if it was at the `Datasource` level. While doable, it was more of a refactor than I wanted to do at the moment.

billdirks: Maybe the refactor is worth it if we don't want this to appear in autocomplete for postgres and other sql backends?

billdirks: Yeah, I've converted this PR to a draft and will fix this now.

great_expectations/experimental/datasources/sqlite_datasource.py
```diff
@@ -27,8 +141,34 @@ class SqliteDatasource(SQLDatasource):
...
+            super().add_table_asset(name, table_name, schema_name, order_by),
+        )
+
+    def add_query_asset(  # type: ignore[override]  # override return type
+        self,
+        name: str,
+        query: str,
+        order_by: Optional[BatchSortersDefinition] = None,
+    ) -> SqliteQueryAsset:
+        return cast(SqliteQueryAsset, super().add_query_asset(name, query, order_by))
```
NathanFarmer: Is this strictly for IDE support?

billdirks: Yes.

great_expectations/experimental/datasources/sqlite_datasource.py
```diff
@@ -1,9 +1,100 @@
...
+)
+
+if TYPE_CHECKING:
+    from great_expectations.experimental.datasources.interfaces import (
+        BatchRequestOptions,
+        BatchSortersDefinition,
+        DataAsset,
+    )
+
+# This module serves as an example of how to extend _SQLAssets for specific backends. The steps are:
```
billdirks, line 27: I've added explicit instructions for subclassing a SQL datasource here.

tests/experimental/datasources/test_sqlite_datasource.py
```diff
@@ -66,3 +71,101 @@ def test_connection_string_that_does_not_start_with_sqlite():
...
+    all_batches_cnt,
+    specified_batch_request,
+    specified_batch_cnt,
+    last_specified_batch_metadata,
+):
+    with create_sqlite_source(
+        splitter_query_response=[response for response in splitter_query_responses],
+    ) as source:
+        asset = source.add_query_asset(name="query_asset", query="SELECT * from table")
+        getattr(asset, splitter_name)(**splitter_kwargs)
```
alexsherstinsky: @billdirks Suggestion: Not sure if anybody else would agree, but for clarity purposes, I would keep splitter name to the actual name of the splitter and put the string "add_" explicitly in the `getattr()` calls.  Thanks.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -290,13 +257,32 @@ def batch_request_options_to_batch_spec_kwarg_identifiers(
...
+
+    def splitter_method_kwargs(self) -> Dict[str, Any]:
+        return {"column_name": self.column_name, "mod": self.mod}
+
+    def batch_request_options_to_batch_spec_kwarg_identifiers(
+        self, options: BatchRequestOptions
+    ) -> Dict[str, Any]:
+        if "remainder" not in options:
+            raise ValueError(
+                "'remainder' must be specified in the batch request options to create a batch identifier"
```
alexsherstinsky: @billdirks I was under the impression that while "batch_identifier" is still an implementation detail (at least for the immediate term), this phrase is no longer present in the New Datasources vocabulary.  Thanks.

great_expectations/experimental/datasources/sqlite_datasource.py
```diff
@@ -1,9 +1,100 @@
...
+# 1. Create a plain class with the extensions necessary for the specific backend.
+# 2. Make 2 classes XTableAsset and XQueryAsset by mixing in the class created in step 1 with
+#    sql_datasource.TableAsset and sql_datasource.QueryAsset.
+#
+# See SqliteDatasource, SqliteTableAsset, and SqliteQueryAsset below.
+
+
+class ColumnSplitterHashedColumn(_ColumnSplitter):
+    """Splite on hash value of a column.
```
alexsherstinsky: @billdirks Typo: "Splite" -> "Split".  Thanks.

great_expectations/experimental/datasources/sql_datasource.py

tests/experimental/datasources/test_sqlite_datasource.py
billdirks: I've changed the variable name to `add_splitter_method` since the name without the `add_` isn't the exact name. The prefix `add_splitter_*` was requested for these methods but the old gx names are like `split_on_*`.

alexsherstinsky (changes requested):
@billdirks One typo found and a few questions asked for clarification.  Thank you!

tests/experimental/datasources/test_sqlite_datasource.py
billdirks: The old `split_on_*` names refer to methods but the new splitters are objects, hence the different prefixes.

alexsherstinsky: ❤️

great_expectations/experimental/datasources/sql_datasource.py
alexsherstinsky: 🙇

alexsherstinsky (approved):
LGTM!

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -16,10 +15,15 @@
 )
 
 import pydantic
-from typing_extensions import Literal
```
NathanFarmer, line 19: ❤️

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -709,12 +724,14 @@ class SQLDatasource(Datasource):
     # private attrs
     _cached_connection_string: str = pydantic.PrivateAttr("")
     _engine: Union[sqlalchemy.engine.Engine, None] = pydantic.PrivateAttr(None)
+    # These are instance var because ClassVars can't contain Type variables. See
+    # https://peps.python.org/pep-0526/#class-and-instance-variable-annotations
+    _TableAsset: Type[TableAsset] = pydantic.PrivateAttr(TableAsset)
+    _QueryAsset: Type[QueryAsset] = pydantic.PrivateAttr(QueryAsset)
```
NathanFarmer, line 730: Why set this here? I see it being used down below, but I don't understand why.

great_expectations/experimental/datasources/sqlite_datasource.py
```diff
@@ -16,6 +107,45 @@ class SqliteDsn(pydantic.AnyUrl):
...
+                method_name="split_on_converted_datetime",
+                column_name=column_name,
+                date_format_string=date_format_string,
+            )
+        )
+
+
+class SqliteTableAsset(_SQLiteAssetMixin, SqlTableAsset):
+    type: Literal["sqlite_table"] = "sqlite_table"  # type: ignore[assignment]  # override superclass value
+    column_splitter: Optional[SqliteColumnSplitter] = None  # type: ignore[assignment]  # override superclass type
```
NathanFarmer, line 141: What is driving going back to this mixin pattern? I'm not a fan of the `type: ignore`s in extended classes, but I guess its fine if we gain something somewhere else.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: This is a point of extension for subclasses and let's us decouple the creation of a [Table|Query]Asset from the datasource. That way sql specific engines can set these to construct their particular versions of the asset. See the sqlite datasource to see this in action.

Alternatively, I could have defined 2 methods `create_table_asset` and `create_query_asset` that the subclass could override and I could call these where I instantiate the asset.

great_expectations/experimental/datasources/sqlite_datasource.py
billdirks: Any new functionality we want to add, adding the new column splitters here, apply to both the table asset and the query asset. They don't share a common sqlite asset parent however, so I use this to prevent code duplication. You write the functionality once and mix it in to the engine specific table and query asset.

I just thought of this.  I could use dependency injection where I have a member variable in `_SQLAsset` called something like `engine_extensions`. I could override `__getattr__` or `__getattribute__` and delegate calls to `engine_extensions`.  I wouldn't need the sqlite datasset subclasses since I could inject in the sqlite engine extensions. However, we would need to update how we serialize/deserialize from config since we would need to know what `engine_extensions` to inject. Or I could still make subclasses and set the correct `engine_extensions` in the subclass. 

In terms of the type ignore, I agree I don't like it. Mypy doesn't do a good job with mixins since they are so dynamic and recommends using [protocols](https://mypy.readthedocs.io/en/latest/more_types.html#mixin-classes). I want to return `Self` here to be fluent and I don't think these interact well. There are a few other ways to avoid it:
* The dependency injection strategies above would avoid this. 
* I could define a property that calls super to avoid it (I don't think mypy yet checks the superclass hierarchy to see if it exists) or I could possibly declare an attribute. These seemed a little like tricking the type checker.
These all brought in forms of complication and I opted to ignore.

PR #7219, opened 2023-02-23
[BUGFIX] Ensure that Marshmallow UUIDs are converted to strings before usage

cdkini:
- Keep UUID type for validation but cast to string before usage

### Definition of Done


great_expectations/checkpoint/checkpoint.py
```diff
@@ -690,8 +689,8 @@ def __init__(
         profilers: Optional[List[dict]] = None,
         validation_operator_name: Optional[str] = None,
         batches: Optional[List[dict]] = None,
-        ge_cloud_id: Optional[UUID] = None,
-        expectation_suite_ge_cloud_id: Optional[UUID] = None,
+        ge_cloud_id: Optional[str] = None,
+        expectation_suite_ge_cloud_id: Optional[str] = None,
```
Kilo59, line 693: If these are being converted with a mm hook then shouldn't this be a union of `str | UUID` (and `None`)

cdkini: I guess we technically could allow both but I think it's cleaner to just use a string? Open to either?

Kilo59 (approved):

great_expectations/data_context/types/resource_identifiers.py
```diff
@@ -524,10 +523,10 @@ def from_tuple(cls, tuple_):
 
 
 class ConfigurationIdentifier(DataContextKey):
-    def __init__(self, configuration_key: Union[str, UUID]) -> None:
+    def __init__(self, configuration_key: str) -> None:
         super().__init__()
-        if isinstance(configuration_key, UUID):
-            configuration_key = str(configuration_key)
+        # if isinstance(configuration_key, UUID):
```
dctalbot: should we delete this?

cdkini: yup, good call!

PR #7235, opened 2023-02-25
Contrib  time series generators

abegong:
- Adds better type hinting and testing for time series generators
- Converts generators from utility scripts to a shared TimeSeriesGenerator class
- Extends functionality by implementing HourlyTimeSeriesGenerator, DailyTimeSeriesGenerator, WeeklyTimeSeriesGenerator, and MonthlyTimeSeriesGenerator.

Note: I mistakenly built this off my previous branch, without pulling the squashed commits. As a result, this branch shows a long commit history, but only the commits from Feb 25 and later are actually new.


alexsherstinsky (approved):
LGTM

contrib/time_series_expectations/time_series_expectations/generator/time_series_generator.py
```diff
@@ -0,0 +1,40 @@
...
+            *args will differ depending on the specific generator class.
+
+        Keyword Args:
+            *kwargs will differ depending on the specific generator class.
+
+        Returns:
+            pd.DataFrame: A two-column pandas DataFrame with a datetime index and a column for the time series values
+
+        """
+        raise NotImplementedError()
```
alexsherstinsky: Non-blocking comment.  When we extend ABC, we can also `import  from abc abstractmethod` and then decorate `generate_df()` with `@abstractmethod` (then the `generate_df()` does not need to raise `NotImplementedError()` but can simple contain `pass`).  Thank you.

abegong: Got it. I'll make that change before merging.

PR #7244, opened 2023-02-27
[BUGFIX] Use timezone-aware datetimes in checkpoint run information by default.

jcampbell:

- Updates datetime in the run_time used by checkpoints to be timezone-aware
-

### Definition of Done

Thank you for submitting!


ghost:
<details open>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=7244&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

#### 

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=7244&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map legend
</details>

great_expectations/checkpoint/checkpoint.py
```diff
@@ -170,7 +170,7 @@ def run(
         # and action_list are considered the "default" validation.
         using_default_validation = not self.validations and not validations
 
-        run_time = run_time or datetime.datetime.now()
+        run_time = run_time or datetime.datetime.now().astimezone()
```
Kilo59, line 173:
```suggestion
        run_time = run_time or datetime.datetime.now(tz=datetime.timezone.utc)
```
I think this is equivalent to what you have but with 1 less object creation.
https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

great_expectations/data_context/types/base.py
```diff
@@ -3136,7 +3136,7 @@ def resolve_config_using_acceptable_arguments(
             run_id and run_time
         ), "Please provide either a run_id or run_name and/or run_time."
 
-        run_time = run_time or datetime.datetime.now()
+        run_time = run_time or datetime.datetime.now().astimezone()
```
Kilo59, line 3139:
```suggestion
        run_time = run_time or datetime.datetime.now(tz=datetime.timezone.utc)
```

great_expectations/checkpoint/checkpoint.py
Kilo59:
```suggestion
        run_time = run_time or datetime.datetime.now(tz=dateutil.tz.gettz())
```

We could also use `dateutil.tz.gettz()` to get the local timezone, and then skip the extra object creation step.
This would require an `import dateutil.tz` for this module.
https://dateutil.readthedocs.io/en/stable/tz.html#dateutil.tz.gettz

I think both methods are fine.

`dateutil` is part of our core requirements, so we should always have access to it.
https://github.com/great-expectations/great_expectations/blob/bcc6e8564a743d58db5abcaa161f016e0981ead0/requirements.txt#L27

Kilo59 (approved):
Approved with a non-blocking comment.

great_expectations/data_context/types/base.py
DrewHoo: Is there specific reason to attach local timezone information here vs creating an aware datetime with utc as its timezone, e.g. `datetime.now(timezone.utc)`?

jcampbell: I think it's that we *do* want things to display in local timezone where possible; the current behavior uses the local time. It's just that this is the tz-aware local timezone.

great_expectations/core/run_identifier.py
```diff
@@ -54,9 +54,7 @@ def __init__(
         if not run_time.tzinfo:
             # this takes the given time and just adds timezone (no conversion)
             run_time = run_time.replace(tzinfo=datetime.timezone.utc)
```
DrewHoo: Wait, this is the bug, isn't it? if run_time is naive and you replace the timezone with utc, you fundamentally change the time if the naive time isn't utc. we should be making this case impossible by only making (preferably utc) aware times or calling `run_time = run_time.astimezone(tz=timezone.utc)`

great_expectations/checkpoint/checkpoint.py
jcampbell: Not a really strong preference here, but my thinking was that we're using dateutil in expectations. I still have hope that eventually the core library dependencies are isolated from expectation dependencies, so I liked keeping to the standard library here.

DrewHoo (changes requested):
Got a couple questions re: root cause of the bug + preference for using utc tz in aware datetimes instead of local

great_expectations/data_context/types/base.py
DrewHoo: Where are we using this information to display in local timezones? If it's in a browser, don't we want the browser's idea of local time instead of whatever the execution environment thinks local time is?

great_expectations/core/run_identifier.py
jcampbell: Does that mean we should just remove this do you think? So if somehow it was naive it stays naive?

jcampbell: Alternatively, we could assume that if it's naive it is local I guess...

DrewHoo: this definitely has to go because it's destructive--I also think we should stick to utc aware times, especially if the string output doesn't contain full tz information (looks like we're just including offsets)--that will lead to confusing behavior when offsets are reinterpreted by browsers or anything with more specific timezone info

jcampbell: I'm not sure what you mean by being destructive (or the difference of full tz info vs. offsets)...can you clarify the concern a bit more?

great_expectations/data_context/types/base.py
jcampbell: It comes up in the MS Teams notification that is changed in this test, and in the dictionary-view of results in a notebook. I don't really have a dogmatic view on whether the internal representation is UTC or not as long as it's timezone aware.

If we were to go with `run_time = run_time or datetime.datetime.now(tz=datetime.timezone.utc)` here instead, for example, we'd get the same (correct-as-this-fix) behavior in cloud, but the dictionary view would show the correct UTC time instead of the correct local time.

jcampbell: ...and, what-the-heck that makes more sense to me, so I switched to that.

great_expectations/core/run_identifier.py
Kilo59: As long as we always include the timezone when saving/creating `datetime` objects we should always be able to compare `datetimes` across timezones and translate any `datetime`s to the user's local time.

That's why IMO we should turn on the linting rules that disallow any naive `datetimes` from ever being created.

But I think maybe we don't actually do that final step of translating the `datetime` during the "last mile" of serialization/presentation to the user's local time (not just showing the tz offset)??

jcampbell: Agreed...curious what Drew thinks here. End of the day as long as it's an aware datetime, the underlying problem is fixed from my perspective, but happy to only do the localization at the end.

great_expectations/core/run_identifier.py
```diff
@@ -50,13 +53,11 @@ def __init__(
             except (ValueError, TypeError):
                 run_time = None
 
-        run_time = run_time or datetime.datetime.now(datetime.timezone.utc)
+        run_time = run_time or datetime.datetime.now(tz=datetime.timezone.utc)
         if not run_time.tzinfo:
             # this takes the given time and just adds timezone (no conversion)
             run_time = run_time.replace(tzinfo=datetime.timezone.utc)
```
DrewHoo: I think we want something like this--if run_time is aware, don't touch it. If it is naive, make it aware w/ utc as its timezone. If it doesn't exist, create it as a utc-aware datetime
```
        if not run_time:
           run_time = datetime.datetime.now(datetime.timezone.utc)
        elif not run_time.tzinfo:
          run_time = run_time.astimezone(tz=datetime.timezone.utc)
```
crucial (destructive) difference between 
```
run_time.replace(tzinfo=datetime.timezone.utc)
```
and 
```
run_time = run_time.astimezone(tz=datetime.timezone.utc)
```
is that the former changes the underlying data by the offset between naive time and utc time.
Also: I checked datadocs just to be sure (maybe this was obvious) datadocs happily displays inferred local time from browser when given a utc time.

jcampbell: Here's how I'm thinking about this: 
- If naive time is in the user's local time, the right behavior is to use `astimezone`
- If naive time is in UTC time, the right behavior is to use `replace`

I don't know if the original logic, but I suspect that since the schema we were using to serialize/deserialize was defaulting to UTC but was NOT aware (???), the `replace` option was sensical. That maybe overly optimistic, and I didn't verify it. Net/net, with the other changes here, going to `astimezone` makes sense to me, so I'll make that change now.

great_expectations/core/run_identifier.py
```diff
@@ -70,13 +72,17 @@ def run_time(self):
...
+            self._run_time.astimezone(tz=datetime.timezone.utc).strftime(
+                "%Y%m%dT%H%M%S.%fZ"
+            ),
         )
 
     def to_fixed_length_tuple(self):
         return (
             self._run_name or "__none__",
-            self._run_time.strftime("%Y%m%dT%H%M%S.%fZ"),
+            self._run_time.astimezone(tz=datetime.timezone.utc).strftime(
```
jcampbell, line 83: @DrewHoo / @Kilo59 is this change right given the other things?

DrewHoo: LGTM!

great_expectations/core/run_identifier.py
```diff
@@ -70,13 +72,17 @@ def run_time(self):
...
+                "%Y%m%dT%H%M%S.%fZ"
+            ),
         )
 
     def to_fixed_length_tuple(self):
         return (
             self._run_name or "__none__",
-            self._run_time.strftime("%Y%m%dT%H%M%S.%fZ"),
```
DrewHoo, line 79: Actually, why is this `%fZ` and not `%f%Z` (actually `%f%z` might be more consistent)? This also seems like a bug, because the former prints an apparently UTC timezone datetime regardless of whether conversion was performed (with your change, conversion is performed):
```
>>> datetime.now().strftime("%Y%m%dT%H%M%S.%fZ") # you're specifying timezone so this is not an issue for your proposed change but it illustrates a bug that was probably plaguing us
'20230306T163511.786825Z' # it is not currently 11:35AM local time
```
and the latter prints the timezone name:
```
>>> datetime.now(tz=timezone.utc).strftime("%Y%m%dT%H%M%S.%f%Z")
'20230306T212900.056263UTC'
```
tl;dr I think it makes sense to change that format string to `"%Y%m%dT%H%M%S.%f%z"` so that we get strings that look like this: `'20230306T212905.119843+0000'`

IDK how much it makes sense to change here but `.isoformat()`'s  output is even more readable:
`'2023-03-06T21:30:50.685481+00:00'`

DrewHoo (approved):

great_expectations/core/run_identifier.py
DrewHoo: On second thought, PR as is abides by "no worse than before", would like to deal with the formatting weirdness in a different venue

NathanFarmer (approved):

PR #7255, opened 2023-03-01
[FEATURE] Develop SparkGoogleCloudStorageDatasource SparkAzureBlobStorageDatasource (as part of New Datasources)

alexsherstinsky:
### Scope
* Develop Google Cloud Storage support for new SparkDatasource
* Develop Microsoft Azure Blob Storage support for new SparkDatasource

- JIRA: GREAT-1462/GREAT-1692/GREAT-1690/GREAT-1691/GREAT-1619
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!

### DIFF
--- a/great_expectations/experimental/datasources/pandas_google_cloud_storage_datasource.py
+++ b/great_expectations/experimental/datasources/pandas_google_cloud_storage_datasource.py
@@ -42,13 +42,12 @@ logger = logging.getLogger(__name__)
 
 GCS_IMPORTED = False
 try:
-    from google.cloud import storage
-    from google.oauth2 import service_account
+    from google.cloud import storage  # noqa: disable=E0602
+    from google.oauth2 import service_account  # noqa: disable=E0602
 
     GCS_IMPORTED = True
 except ImportError:
-    storage = None
-    service_account = None
+    pass
 
 
 class PandasGoogleCloudStorageDatasourceError(PandasDatasourceError):
--- a/great_expectations/experimental/datasources/pandas_google_cloud_storage_datasource.py
+++ b/great_expectations/experimental/datasources/pandas_google_cloud_storage_datasource.py
@@ -51,7 +51,7 @@ except ImportError:
     service_account = None
 
 
-class PandasGoogleCloudStorageError(PandasDatasourceError):
+class PandasGoogleCloudStorageDatasourceError(PandasDatasourceError):
     pass
 
 
@@ -96,11 +96,11 @@ class PandasGoogleCloudStorageDatasource(_PandasFilePathDatasource):
                     )
                 except Exception as e:
                     # Failure to create "gcs_client" is most likely due invalid "gcs_options" dictionary.
-                    raise PandasGoogleCloudStorageError(
+                    raise PandasGoogleCloudStorageDatasourceError(
                         f'Due to exception: "{str(e)}", "gcs_client" could not be created.'
                     ) from e
             else:
-                raise PandasGoogleCloudStorageError(
+                raise PandasGoogleCloudStorageDatasourceError(
                     'Unable to create "PandasGoogleCloudStorageDatasource" due to missing google dependency.'
                 )
 
@@ -139,7 +139,7 @@ class PandasGoogleCloudStorageDatasource(_PandasFilePathDatasource):
         """Adds a CSV DataAsst to the present "PandasGoogleCloudStorageDatasource" object.
 
         Args:
-            name: The name of the csv asset
+            name: The name of the CSV asset
             batching_regex: regex pattern that matches csv filenames that is used to label the batches
             prefix (str): Google Cloud Storage object name prefix
             delimiter (str): Google Cloud Storage object name delimiter
@@ -192,7 +192,7 @@ class PandasGoogleCloudStorageDatasource(_PandasFilePathDatasource):
         """Adds an Excel DataAsst to the present "PandasGoogleCloudStorageDatasource" object.
 
         Args:
-            name: The name of the csv asset
+            name: The name of the Excel asset
             batching_regex: regex pattern that matches csv filenames that is used to label the batches
             prefix (str): Google Cloud Storage object name prefix
             delimiter (str): Google Cloud Storage object name delimiter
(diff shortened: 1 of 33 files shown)


great_expectations/experimental/datasources/spark_azure_blob_storage_datasource.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+ABS_IMPORTED = False
+try:
+    from azure.storage.blob import BlobServiceClient, ContainerClient
+
+    ABS_IMPORTED = True
+except ImportError:
+    BlobServiceClient = None
+    ContainerClient = None
```
Kilo59: Do you really need to set these to `None`?

great_expectations/experimental/datasources/spark_google_cloud_storage_datasource.py
```diff
@@ -0,0 +1,173 @@
...
+
+GCS_IMPORTED = False
+try:
+    from google.cloud import storage
+    from google.oauth2 import service_account
+
+    GCS_IMPORTED = True
+except ImportError:
+    storage = None
+    service_account = None
```
Kilo59: Why do you need to set these to `None` at all?

alexsherstinsky: @Kilo59 Otherwise, we get a `PyCharm` waning, "service_account in the try block with except ImportError should also be defined in the except block."  What should we do about all these situations?  In the past, we have handled them using "variable = None" in the except block.  Should we simply disable those `PyCharm` warnings?  Thanks.

Kilo59: If the only reason is "because PyCharm" complained then yes, I would get rid of them.
You can know if these values are available by checking `GCS_IMPORTED`.
And if we only need them for type-checking they should be part of the `if TYPE_CHECKING` block.

great_expectations/experimental/datasources/spark_azure_blob_storage_datasource.py
alexsherstinsky: @Kilo59 Please see https://github.com/great-expectations/great_expectations/pull/7255#discussion_r1121853785 (and `https://youtrack.jetbrains.com/issue/PY-8933/Import-in-try-block-should-also-be-defined-in-except-block-false-positive-when-name-not-used`).  Thank you.

great_expectations/experimental/datasources/spark_google_cloud_storage_datasource.py
alexsherstinsky: @Kilo59 I am looking for a way to disable these warnings in `PyCharm` -- thanks.

alexsherstinsky: @Kilo59 I found a way to disable these warnings: `  # noqa: disable=E0602`.  Thanks.

great_expectations/experimental/datasources/pandas_google_cloud_storage_datasource.py
```diff
@@ -42,13 +42,12 @@
 
 GCS_IMPORTED = False
 try:
-    from google.cloud import storage
-    from google.oauth2 import service_account
+    from google.cloud import storage  # noqa: disable=E0602
+    from google.oauth2 import service_account  # noqa: disable=E0602
```
Kilo59, line 46: `non-blocking`
I think you could also use the fully qualified import, and that would get rid of this.
```suggestion
    import google.cloud
    import google.oauth2
```

alexsherstinsky: @Kilo59 How can this work, given that we do need the `storage` and `service_account` references in the logic?  Thanks.

alexsherstinsky: @Kilo59 I see -- by fully qualifying it in the code.  Let me try that.

great_expectations/experimental/datasources/spark_azure_blob_storage_datasource.py
```diff
@@ -0,0 +1,167 @@
...
+            datasource_name=self.name,
+            data_asset_name=name,
+            batching_regex=batching_regex_pattern,
+            azure_client=self._azure_client,
+            account_name=self._account_name,
+            container=container,
+            name_starts_with=name_starts_with,
+            delimiter=delimiter,
+            file_path_template_map_fn=AzureUrl.AZURE_BLOB_STORAGE_HTTPS_URL_TEMPLATE.format,
+        )
```
Kilo59, line 161: I think this was brought up before but we should have a `_get_data_connector` method on the datasource.
As it stands right now, all the dynamic `add_<ASSET_TYPE>_methods()` have no concept of a data connector and will need something like this in order to work.

We'll need to do this for the launch.

Kilo59 (approved):

great_expectations/experimental/datasources/pandas_google_cloud_storage_datasource.py
alexsherstinsky: @Kilo59 This did not work -- the warning complains now about "cloud" and "oauth2" references.  So leaving as recent.  Thanks.

great_expectations/experimental/datasources/spark_azure_blob_storage_datasource.py
alexsherstinsky: @Kilo59 this is exactly the focus of my next pull request (mentioned in Slack).  Thanks.

alexsherstinsky: @Kilo59 Please note that I had suggested having a `_get_data_connector()` method on the `Datasource` a long time ago and had is a variation of the original `DataConnector` introduction to "New Datasources" pull request.  This idea was rejected, and hence we arrived at the current approach.  In fact, I do believe that this is a good idea after all (for the reasons. you mentioned and also for code brevity and elegance).  However, there are nuances.  It may be more fitting to define a `_data_connector` property on `Datasource`, rather than the `_get_data_connector()` method, because the actual  `_data_connector` reference depends on the specific `add_<file_type>_asset()` arguments, which differ among `Datasource` types (e.g., depending on the settings of different cloud file storage mechanisms).  I will have a proposal soon, and then we can iterate it to the best outcome.  Thank you. /cc @donaldheppner

alexsherstinsky: @Kilo59 and @alexsherstinsky believe that a private `Datasource._build_data_connector(self, <common_args>, **kwargs)`  interface method is the best trade-off (and in this case `kwargs` is acceptable, since it is a private interface method, and the subclasses indeed use different arguments, depending on the storage environment).  Thanks. /cc @donaldheppner

PR #7260, opened 2023-03-01
[MAINTENANCE] Keep repo root consistent

anthonyburdi:
- Keep the repo root clean and consistent by limiting the number of files and directories. The limit can be changed but this is intended to be a reminder that we should do so only when necessary. It is also set as a not equal (rather than a greater than) limit so that we remember to lower this number when removing items.

### Definition of Done

### DIFF
--- a/scripts/check_repo_root_size.sh
+++ b/scripts/check_repo_root_size.sh
@@ -5,7 +5,7 @@
 # Please take care to only add files or directories to the repo root unless they are
 # required to be in the repo root, otherwise please find a more appropriate location.
 
-NUM_ITEMS_SHOULD_BE=47
+NUM_ITEMS_SHOULD_BE=40
 NUM_ITEMS=$(ls -la | wc -l)
 
 echo "Items found in repo root:"
--- a/scripts/check_repo_root_size.sh
+++ b/scripts/check_repo_root_size.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# If you need to add or remove from the repo root please change NUM_ITEMS_SHOULD_BE
+# Please take care to only add files or directories to the repo root unless they are
+# required to be in the repo root, otherwise please find a more appropriate location.
+
+NUM_ITEMS_SHOULD_BE=27
+NUM_ITEMS=$(ls -l | wc -l)
+
+if [ $NUM_ITEMS -ne $NUM_ITEMS_SHOULD_BE ];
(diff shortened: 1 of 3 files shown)


scripts/check_repo_root_size.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# If you need to add or remove from the repo root please change NUM_ITEMS_SHOULD_BE
+
+# Please take care to only add files or directories to the repo root unless they are
+# required to be in the repo root, otherwise please find a more appropriate location.
+
+NUM_ITEMS_SHOULD_BE=40
```
cdkini, line 8: Should we ignore directories and files that start with `.`?

Kilo59: There are a lot of configuration files that do start with `.`

cdkini: How do we ignore files/directories that are captured in our `.gitignore`? What if this stage runs after something else and we have a `__pycache__` or `.ruff_cache`?

Kilo59: That's a good point. We should exclude items in `.gitignore`.

anthonyburdi: 💯 I included all files because we do have configuration files starting with `.`. Since this is its own job, I believe it gets a fresh copy of the repo so I don't think we need to worry about something else running first, but please correct me if I'm wrong. See the Checkout great-expectations step in this job: https://dev.azure.com/great-expectations/great_expectations/_build/results?buildId=208968&view=logs&j=ab9c2685-3130-5a8e-cfa2-f6f8ba5ee4fb&t=a3589824-89d9-564f-3a8c-9b0da884204e

Kilo59: Can always just read the items in`.gitignore` and compare them or use something like this.
https://stackoverflow.com/a/467053/6304433

Could even do something simple like count the number of items returned by ` git check-ignore *` and subtract it from `NUM_ITEMS`

Kilo59 (approved):

scripts/check_repo_root_size.sh
anthonyburdi: Hoping to keep this a really simple check using a clean checkout of the repo, but if we need to worry about doing this check after more steps that sounds like a good approach 🙇

PR #7265, opened 2023-03-01
[BUGFIX] Patch Cloud workflow for `add_or_update_checkpoint`

cdkini:
- Ensure that `add_or_update_checkpoint` works with Cloud

Say I have the following code:
```python
checkpoint = context.add_or_update_checkpoint(
    name="experimental",
    config_version=1,
    class_name="Checkpoint",
    validations=[
        ...
    ]
)
```
What happens if I already have "experimental" saved in the backend? Because I'm using `add_or_update`, I'd expect to overwrite the existing value - however, due to the fact that I'm not passing in the associated id, we'll run into an error.

Instead of requiring the user to find out the id, let's find it for them when they reference by name and use that to determine when to add or update.

### Definition of Done


NathanFarmer (approved):
LGTM

PR #7267, opened 2023-03-01
[DOCS] Scan yaml examples for class names that should be part of the public API

anthonyburdi:
- Scan yaml examples for class names that should be a part of the public API.
- Bump the threshold to 87 to include the new items.

### Definition of Done


Kilo59:
@anthonyburdi we are starting to rely on these script files more and more.
Can we bring them under our `ruff` linter coverage?
The sooner we do it, the easier it will be.

scripts/public_api_report.py
```diff
@@ -164,7 +170,11 @@ def _get_names_of_all_usages_in_file(self, file_contents: FileContents) -> Set[s
         import_names = self._get_non_private_gx_import_names(imports=gx_imports)
         logger.debug(f"import_names: {import_names}")
 
-        return function_names | import_names
+        pattern = r"class_name: (\w+)"
```
Kilo59: Should get a slight performance boost by compiling this regex.

https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile

Opinion seems a bit split on how much this improves performance. But it should be at least somewhat faster

donaldheppner (approved):

Kilo59 (approved):

scripts/public_api_report.py
anthonyburdi: Good call! Done :)

anthonyburdi:
> @anthonyburdi we are starting to rely on these script files more and more. Can we bring them under our `ruff` linter coverage? The sooner we do it, the easier it will be.

Good call - I like this idea. I'm sure we will end up adding more scripts... though this feels like a utils directory I'm not sure where else to put it. Maybe these files should be moved into the ci directory. And also put under linter coverage. I was wondering why it wasn't being more strict. Do we remove `scripts` from ruff `extend-excludes`: https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L245 and fix the issues? Or is there something else we need to change as well? (E.g. add `scripts` to mypy files? https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L29)

anthonyburdi:
> > @anthonyburdi we are starting to rely on these script files more and more. Can we bring them under our `ruff` linter coverage? The sooner we do it, the easier it will be.
> 
> Good call - I like this idea. I'm sure we will end up adding more scripts... though this feels like a utils directory I'm not sure where else to put it. Maybe these files should be moved into the ci directory. And also put under linter coverage. I was wondering why it wasn't being more strict. Do we remove `scripts` from ruff `extend-excludes`: https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L245 and fix the issues? Or is there something else we need to change as well? (E.g. add `scripts` to mypy files? https://github.com/great-expectations/great_expectations/blob/develop/pyproject.toml#L29)

Done! https://github.com/great-expectations/great_expectations/pull/7270

PR #7273, opened 2023-03-02
[FEATURE] Editor support for Fluent Datasources dynamic `.add_<DS_TYPE>()` methods

Kilo59:
- add `sources.pyi` stubs file
- explicitly add `Type[<BASE_MODEL>]` type annotations for dynamically created pandas types saved to variables `CSVAsset`, `ExcelAsset` etc.

image

### DIFF
--- a/great_expectations/experimental/datasources/pandas_datasource.py
+++ b/great_expectations/experimental/datasources/pandas_datasource.py
@@ -213,25 +213,37 @@ _PANDAS_ASSET_MODELS = _generate_pandas_data_asset_models(
 )
 
 
-ClipboardAsset = _PANDAS_ASSET_MODELS.get("clipboard", _PandasDataAsset)
-CSVAsset = _PANDAS_ASSET_MODELS.get("csv", _PandasDataAsset)
-ExcelAsset = _PANDAS_ASSET_MODELS.get("excel", _PandasDataAsset)
-FeatherAsset = _PANDAS_ASSET_MODELS.get("feather", _PandasDataAsset)
-GBQAsset = _PANDAS_ASSET_MODELS.get("gbq", _PandasDataAsset)
-HDFAsset = _PANDAS_ASSET_MODELS.get("hdf", _PandasDataAsset)
-HTMLAsset = _PANDAS_ASSET_MODELS.get("html", _PandasDataAsset)
-JSONAsset = _PANDAS_ASSET_MODELS.get("json", _PandasDataAsset)
-ORCAsset = _PANDAS_ASSET_MODELS.get("orc", _PandasDataAsset)
-ParquetAsset = _PANDAS_ASSET_MODELS.get("parquet", _PandasDataAsset)
-PickleAsset = _PANDAS_ASSET_MODELS.get("pickle", _PandasDataAsset)
-SQLAsset = _PANDAS_ASSET_MODELS.get("sql", _PandasDataAsset)
-SQLQueryAsset = _PANDAS_ASSET_MODELS.get("sql_query", _PandasDataAsset)
-SQLTableAsset = _PANDAS_ASSET_MODELS.get("sql_table", _PandasDataAsset)
-SASAsset = _PANDAS_ASSET_MODELS.get("sas", _PandasDataAsset)
-SPSSAsset = _PANDAS_ASSET_MODELS.get("spss", _PandasDataAsset)
-StataAsset = _PANDAS_ASSET_MODELS.get("stata", _PandasDataAsset)
-TableAsset = _PANDAS_ASSET_MODELS.get("table", _PandasDataAsset)
-XMLAsset = _PANDAS_ASSET_MODELS.get(
+ClipboardAsset: Type[_PandasDataAsset] = _PANDAS_ASSET_MODELS.get(
+    "clipboard", _PandasDataAsset
+)
+CSVAsset: Type[_PandasDataAsset] = _PANDAS_ASSET_MODELS.get("csv", _PandasDataAsset)
+ExcelAsset: Type[_PandasDataAsset] = _PANDAS_ASSET_MODELS.get("excel", _PandasDataAsset)
+FeatherAsset: Type[_PandasDataAsset] = _PANDAS_ASSET_MODELS.get(
+    "feather", _PandasDataAsset
+)
+GBQAsset: Type[_PandasDataAsset] = _PANDAS_ASSET_MODELS.get("gbq", _PandasDataAsset)
+HDFAsset: Type[_PandasDataAsset] = _PANDAS_ASSET_MODELS.get("hdf", _PandasDataAsset)
+HTMLAsset: Type[_PandasDataAsset] = _PANDAS_ASSET_MODELS.get("html", _PandasDataAsset)
+JSONAsset: Type[_PandasDataAsset] = _PANDAS_ASSET_MODELS.get("json", _PandasDataAsset)
+ORCAsset: Type[_PandasDataAsset] = _PANDAS_ASSET_MODELS.get("orc", _PandasDataAsset)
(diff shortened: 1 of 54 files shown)


great_expectations/experimental/datasources/sources.pyi
```diff
@@ -0,0 +1,61 @@
...
+        cls, ds_type: Type[Datasource], factory_fn: SourceFactoryFn
+    ) -> None: ...
+    @property
+    def pandas_default(self) -> PandasDatasource: ...
+    @property
+    def factories(self) -> List[str]: ...
+    def __getattr__(self, attr_name: str): ...
+    def __dir__(self) -> List[str]: ...
+    def add_sqlite(
+        self, name: str, *, connection_string: SqliteDsn
```
Kilo59: Should expect to see a new mypy error because we are actually passing `str` here.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -213,25 +213,37 @@ def _validate_batch_request(self, batch_request: BatchRequest) -> None:
...
-PickleAsset = _PANDAS_ASSET_MODELS.get("pickle", _PandasDataAsset)
-SQLAsset = _PANDAS_ASSET_MODELS.get("sql", _PandasDataAsset)
-SQLQueryAsset = _PANDAS_ASSET_MODELS.get("sql_query", _PandasDataAsset)
-SQLTableAsset = _PANDAS_ASSET_MODELS.get("sql_table", _PandasDataAsset)
-SASAsset = _PANDAS_ASSET_MODELS.get("sas", _PandasDataAsset)
-SPSSAsset = _PANDAS_ASSET_MODELS.get("spss", _PandasDataAsset)
-StataAsset = _PANDAS_ASSET_MODELS.get("stata", _PandasDataAsset)
-TableAsset = _PANDAS_ASSET_MODELS.get("table", _PandasDataAsset)
-XMLAsset = _PANDAS_ASSET_MODELS.get(
+ClipboardAsset: Type[_PandasDataAsset] = _PANDAS_ASSET_MODELS.get(
```
Kilo59, line 216: Need this because without this the new `soures.pyi` causes a bunch of additional `has-type` errors
https://mypy.readthedocs.io/en/stable/error_code_list.html?highlight=has-type#check-that-type-of-target-is-known-has-type

tests/experimental/datasources/test_pandas_filesystem_datasource.py
```diff
@@ -289,7 +289,7 @@ def test_data_asset_reader_options_passthrough(
         extra_kwargs: dict,
     ):
         batch_request = (
-            empty_data_context.sources.add_pandas_filesystem(
+            empty_data_context.sources.add_pandas_filesystem(  # type: ignore[attr-defined]
```
Kilo59: This type-ignore should go away once the stubs for all the `add_asset_methods` have been defined too.

great_expectations/experimental/datasources/sources.py
```diff
@@ -32,7 +32,7 @@
     )
     from great_expectations.validator.validator import Validator
 
-SourceFactoryFn = Callable[..., "Datasource"]
+SourceFactoryFn: TypeAlias = Callable[..., "Datasource"]
```
alexsherstinsky, line 35: @Kilo59 Question out of curiosity: How come the return type of `Callable` is the string `"Datasource"` value, rather than just the type (i.e., `Datasource` without the quotes)?  Thank you.

great_expectations/experimental/datasources/sources.pyi
```diff
@@ -0,0 +1,129 @@
+from __future__ import annotations
```
alexsherstinsky, line 1: @Kilo59 Question out of curiosity: must the `.pyi` file always have the same root name as the concrete implementation module ("sources" in the present example)?  Thanks.

great_expectations/experimental/datasources/sources.py
Kilo59: Because of the assignment. Anything to the right operator is happening at runtime, so we can't use the non-string postponed annotation we do for other type hints.

great_expectations/experimental/datasources/sources.pyi
Kilo59: Yes

great_expectations/experimental/datasources/sources.pyi
```diff
@@ -0,0 +1,129 @@
...
+DEFAULT_PANDAS_DATA_ASSET_NAME: Final[str]
+
+class DefaultPandasDatasourceError(Exception): ...
+class TypeRegistrationError(TypeError): ...
+
+class _FieldDetails(NamedTuple):
+    default_value: Any
+    type_annotation: Type
+
+def _get_field_details(
```
alexsherstinsky, line 61: @Kilo59 Non-blocking: Due to reliance on meta-programming, it would be so nice to have docstrings explaining the purpose of these (even private) methods.  Thank you.

alexsherstinsky (approved):
LGTM (subject to non-blocking questions and request).  Thank you!

great_expectations/experimental/datasources/sources.pyi
Kilo59: There's nothing I could add to this particular docstring that isn't captured by the method's name and its return type.
It gets the pydantic field details for a field. It's `type` and the default value for it.

tests/experimental/datasources/integration/conftest.py
```diff
@@ -155,7 +155,7 @@ def multibatch_pandas_data(
         batching_regex=r"yellow_tripdata_sample_(?P<year>\d{4})-(?P<month>\d{2})\.csv",
         order_by=["year", "month"],
     )
-    batch_request = asset.build_batch_request({"year": "2020"})
+    batch_request = asset.build_batch_request({"year": "2020"})  # type: ignore[attr-defined]
```
billdirks, line 158: Why do we need this now?

tests/experimental/datasources/test_pandas_filesystem_datasource.py
billdirks: Oh, I believe this answers my question from above?

Kilo59, line 292:
```suggestion
            empty_data_context.sources.add_pandas_filesystem(  # type: ignore[attr-defined] # .build_batch_request
```

tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -281,7 +281,7 @@ def test_data_asset_reader_options_passthrough(
             {"filepath_or_buffer": csv_path / "yellow_tripdata_sample_2018-04.csv"}
         )
         batch_request = (
-            empty_data_context.sources.add_pandas(
+            empty_data_context.sources.add_pandas(  # type: ignore[attr-defined]
```
Kilo59, line 284:
```suggestion
            empty_data_context.sources.add_pandas(  # type: ignore[attr-defined] # .build_batch_request
```

billdirks (approved):
I believe you mentioned a following PR will have a tool for generating and syncing the pyi files going forward?

Thanks for doing this!

great_expectations/experimental/datasources/sources.py
alexsherstinsky: @Kilo59 And rearranging positions of the type definitions involved and/or breaking up the module can help here?  Or can it?

great_expectations/experimental/datasources/sources.pyi
alexsherstinsky: @Kilo59 My comment is a general sentiment, alluding to the fact that sometimes following the code in the IDE (`PyCharm`) in my case often does not work for these dynamically created methods, adding challenges to studying the implementation.

PR #7274, opened 2023-03-02
[MAINTENANCE] Add maybe imports

billdirks:


great_expectations/util.py
```diff
@@ -2152,3 +2152,37 @@ def numpy_quantile(
...
+        raise ModuleNotFoundError(self.__dict__["gx_error_message"])
+
+    def __call__(self, *args, **kwargs) -> Any:
+        raise ModuleNotFoundError(self.__dict__["gx_error_message"])
+
+    def __str__(self) -> str:
+        return self.__dict__["gx_error_message"]
+
+
+def maybe_import_module(module_name: str) -> Union[ModuleType, NotImport]:
```
alexsherstinsky: @billdirks I am wondering if the approach we used elsewhere already can help with higher-level methods (please see `head()` in New Datasources (also present in `MetricsCalculator` -- it was a generalization of the approach you showed me, when I found that we already had that in `great_expectations/util.py`.  Thanks.

tests/test_maybe_importing.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_maybe_import_existing_value_attr():
+    attr = maybe_import_attr("datetime", "datetime")
+    # This should not error
+    attr.now()
+
+
+def test_maybe_import_nonexisting_attr():
+    with pytest.raises(AttributeError):
```
Kilo59: Can we verify that the error message gives the intended helpful info?

tests/test_maybe_importing.py
```diff
@@ -0,0 +1,33 @@
...
+        attr = maybe_import_attr("datetime", "not_a_real_attribute")
+
+
+def test_maybe_import_uninstalled_attr():
+    attr = maybe_import_attr("this_module_does_not_exist", "an_attr")
+    with pytest.raises(ModuleNotFoundError):
```
Kilo59:
```suggestion
    with pytest.raises(ModuleNotFoundError, match="an_attr is not available because this_module_does_not_exist is not installed. Please install."):
```

Kilo59 (approved):
Looks good.
Could we update some of the tests to ensure the error message is what we expect?

Can do this with the `match` argument to `pytest.raises` or by using the `ExceptionInfo` object that the context manager provides.

Example below in the comments

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.raises

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -37,14 +38,13 @@
...
-try:
-    import sqlalchemy
+sqlalchemy = maybe_import_module("sqlalchemy")
 
-    SQLALCHEMY_IMPORTED = True
-except ImportError:
-    pass
+if TYPE_CHECKING:
+    # sqlalchemy is a hard requirement for type checking
+    import sqlalchemy  # noqa: TCH004
```
Kilo59: If this is going to be a recurring problem with the `maybe_import_module()` we might want to just disable this rule (`TCH004`).
It's only ever "caught" these false positives where we have a try import and an `If TYPE_CHECKING` import.

billdirks:
I ran into some difficulties because I thought something like this would work with mypy but it does not:
```
sqlalchemy = maybe_import_module("sqlalchemy")
Table = maybe_import_attr("sqlalchemy", "Table")

if TYPE_CHECKING:
    # sqlalchemy is a hard requirement for type checking
    import sqlalchemy  # noqa: TCH004
    from sqlalchemy import Table
```
That is, I thought that when mypy ran the `TYPE_CHECKING` block would make `sqlalchemy` and `Table` resolve to their proper types but this does not happen. I am going to explore a different path forward.

tests/test_maybe_importing.py
billdirks: I had to change how I am doing these imports, see https://github.com/great-expectations/great_expectations/pull/7274#issuecomment-1452701350.

I verify the error message are as expected.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: The strategy is different now, see https://github.com/great-expectations/great_expectations/pull/7274#issuecomment-1452701350.

great_expectations/util.py
```diff
@@ -2152,3 +2152,20 @@ def numpy_quantile(
         )
 
     return quantile
+
+
+class NotImported:
+    def __init__(self, message: str):
```
billdirks, line 2158: I was unsure whether to have default error message or leave it open like this. One may want different error messages depending on if one is importing a module or a symbol from a module.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -37,14 +37,12 @@
...
+from great_expectations.util import NotImported
 
-SQLALCHEMY_IMPORTED = False
 try:
     import sqlalchemy
-
-    SQLALCHEMY_IMPORTED = True
 except ImportError:
-    pass
+    sqlalchemy = NotImported("sqlalchemy not found, please install.")
```
billdirks, line 46: The import strategy looks similar to what we do now except instead of setting the module to `None` you set it to `NotImported` which will result in better error messages. I was hoping one could avoid the try/except block but that proved hard because of how mypy handles the interplay of dynamic types that might not be types when not type-checking but will be types when type-checking.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -827,7 +823,12 @@ class SQLDatasource(Datasource):
 
     # private attrs
     _cached_connection_string: str = pydantic.PrivateAttr("")
-    _engine: Union[sqlalchemy.engine.Engine, None] = pydantic.PrivateAttr(None)
+    # The sqlalchemy is in quotes (hence the # noqa) here because otherwise pydantic
+    # will try to resolve this annotation at gx import time which will cause the
+    # process die if sqlalchemy is not installed.
+    _engine: Union["sqlalchemy.engine.Engine", None] = pydantic.PrivateAttr(  # noqa
```
alexsherstinsky: @billdirks I would like the opinion of @Kilo59 on this one; I may be wrong, but somehow it seems that in the past we handled it somehow without the quotes, but I am not a hundred percent certain.  I was under the impression that by importing from "future" and using "TYPE_CHECKING" we can keep the actual types without the quotes.  In addition, if we end up going with the quotes, it would be nice to provide more information for the `# noqa` macro (e.g., `# noqa: E711`).  Thank you.

great_expectations/experimental/datasources/sql_datasource.py
```diff
@@ -840,23 +841,17 @@ def execution_engine_type(self) -> Type[SqlAlchemyExecutionEngine]:
...
-                    # connection_string has passed pydantic validation, but still fails to create a sqlalchemy engine
-                    # one possible case is a missing plugin (e.g. psycopg2)
-                    raise SQLDatasourceError(
-                        "Unable to create a SQLAlchemy engine from "
-                        f"connection_string: {self.connection_string} due to the "
-                        f"following exception: {str(e)}"
-                    ) from e
-                self._cached_connection_string = self.connection_string
-            else:
+            try:
```
alexsherstinsky, line 844: @billdirks If we indeed dispense with the `<MY_PACKAGE>_IMPORTED` paradigm and this pull request gets merged, then how should we go about updating all the other `Datasource` subclasses?  Perhaps create a ticket for each?  As it stands right now, all `Datasource` implementations use this `<MY_PACKAGE>_IMPORTED` pattern (I followed the recommendation by @NathanFarmer and made all File Path type `Datasource` modules follow it).  Thank you.

tests/test_not_imported.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def test_set_attr_on_not_installed(not_imported):
+    with pytest.raises(ModuleNotFoundError, str(not_imported)):
+        not_imported.thing = 5
+
+
+def test_calling_not_installed(not_imported):
+    with pytest.raises(ModuleNotFoundError, str(not_imported)):
+        not_imported()
```
alexsherstinsky, line 23: @billdirks I am curious -- how come we need the `NotImported` class to be "callable"?  Thanks.

alexsherstinsky (changes requested):
@billdirks I have asked a few questions and made (minor) suggestions.  I would like to learn the opinions of @NathanFarmer and @Kilo59 as well on some of these points.  (By the way, the failing tests might be shedding some useful information about `NotImported` implementation/usage as well.)  Thank you!

tests/test_not_imported.py
billdirks: For this use case:
```import fxn from module```

File: great_expectations/experimental/datasources/sql_datasource.py (hunk 7 above)
Comment (billdirks) [reply great_expectations/experimental/datasources/sql_datasource.py]: Please see comment above. The quotes aren't for static analysis and type checking, they are there because of pydantic. I am not sure what the correct code to use is for the `noqa` annotation. Is there a definitive ruff list somewhere?

Comment (billdirks) [reply great_expectations/experimental/datasources/sql_datasource.py]: Oh I see it now, it's `UP037`. I'll update it.

File: great_expectations/experimental/datasources/sql_datasource.py (hunk 8 above)
Comment (billdirks) [reply great_expectations/experimental/datasources/sql_datasource.py:844]: I think we should do it across the whole code base. We set `module = None` all over the place. We can fix it as we go or do it all at once. I'm more worried about the cases where we don't catch it at all than catching it an alternative way.

**Review by Kilo59 — APPROVED** [review]

File: great_expectations/experimental/datasources/sql_datasource.py [hunk 10]
```diff
@@ -827,7 +823,12 @@ class SQLDatasource(Datasource):
 
     # private attrs
     _cached_connection_string: str = pydantic.PrivateAttr("")
-    _engine: Union[sqlalchemy.engine.Engine, None] = pydantic.PrivateAttr(None)
+    # The sqlalchemy is in quotes (hence the # noqa) here because otherwise pydantic
+    # will try to resolve this annotation at gx import time which will cause the
+    # process die if sqlalchemy is not installed.
+    _engine: Union[
+        "sqlalchemy.engine.Engine", None  # noqa: UP037
+    ] = pydantic.PrivateAttr(None)
```
Kilo59: I would expect this to work the same as long as you are doing a `IF TYPE_CHECKING: import sqlalchemy` and using `from __future__ import annotations`

Kilo59 (changes requested):
I think we need to think about this a bit more. I'm worried about how easily string annotations can turn into `Any`s.

great_expectations/experimental/datasources/sql_datasource.py
billdirks: This fails in pydantic at runtime, and not at type_checking time, during a call to `resolve_annotations`. I can walk you through the backtrace to show you what is happening if you want.

Kilo59: So is this going to force us to always use strings anytime we use this `NotImported`? That seems less than ideal.

The problem with string type annotations is that it's very easy for them to turn into `Any`'s.

If `sqlalchemy` isn't available when the type-checker is run then this type will turn into an `Any` and it will be hard to catch when and if this has happened and easy for typing bugs to get through.
Example: `sqlalchemy` isn't installed as part of the type-checking step, or if this import gets removed.

When we use non-string postponed annotations it's easy to catch, ruff will flag that symbol/name is not available.

Discussed this same point regarding some ruff auto-fixes a little while ago.
https://github.com/charliermarsh/ruff/issues/2329#issuecomment-1415930582

great_expectations/experimental/datasources/sql_datasource.py
alexsherstinsky: @billdirks We have a lot less of `module = None` now, because of the effort already conducted (the one mentioned in my comment).  But whatever is our best approach going forward, I am happy for us to apply universally.  Thanks.

alexsherstinsky (approved):

Kilo59 (approved):

PR #7280, opened 2023-03-03
[FEATURE] `PandasDatasource` in-memory `DataFrameAsset`

NathanFarmer:
- GREAT-1719
- `PandasDatasource` in-memory `DataFrameAsset`
- Exclude `dataframe` attribute from serialization
- Include `dataframe` attribute in json schema (without type)

### Definition of Done


tests/experimental/datasources/test_pandas_datasource.py
```diff
@@ -403,18 +403,16 @@ def test_default_pandas_datasource_get_and_set(
     assert pandas_datasource.assets[DEFAULT_PANDAS_DATA_ASSET_NAME]
 
     # ensure we overwrite the ephemeral data asset if no name is passed
-    validator = pandas_datasource.read_csv(filepath_or_buffer=valid_file_path)
-    assert isinstance(validator, Validator)
```
NathanFarmer: No need to check this more than once.

great_expectations/experimental/datasources/pandas_datasource.py
```diff
@@ -294,17 +409,34 @@ def _get_validator(self, asset: _PandasDataAsset) -> Validator:
...
         return self.add_asset(asset=asset)
 
-    def read_clipboard(self, name: Optional[str] = None, **kwargs) -> Validator:
-        if not name:
-            name = DEFAULT_PANDAS_DATA_ASSET_NAME
-        asset: ClipboardAsset = self.add_clipboard_asset(name=name, **kwargs)  # type: ignore[valid-type]
+    def read_clipboard(self, asset_name: Optional[str] = None, **kwargs) -> Validator:
+        if not asset_name:
+            asset_name = DEFAULT_PANDAS_DATA_ASSET_NAME
+        asset: ClipboardAsset = self.add_clipboard_asset(name=asset_name, **kwargs)  # type: ignore[valid-type]
```
NathanFarmer: This parameter name change was missed in a previous PR

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -340,7 +341,7 @@ class Datasource(
     # data context method `data_context.sources.add_my_datasource` method.
 
     # class attrs
-    asset_types: ClassVar[List[Type[DataAsset]]] = []
+    asset_types: ClassVar[Sequence[Type[DataAsset]]] = []
```
alexsherstinsky, line 344: @NathanFarmer Quick question: going forward, when do we use `List`, `Set`, etc. versus `Sequence`?  Thanks.

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -80,7 +86,7 @@ def _get_reader_method(self) -> str:
         )
 
     def test_connection(self) -> None:
-        pass
+        ...
```
alexsherstinsky, line 89: @NathanFarmer Quick question -- what is the reason for the ellipsis?  Is there a `test_connection()` with a different signature that this method falls down to?  Thanks.

great_expectations/datasource/fluent/interfaces.py
NathanFarmer: It depends on the use case. I was running into issues when trying to use two lists in `asset_types` because lists are invariant. I wanted to make this a `tuple`, but that would require updating every `Datasource` in all of `fluent`. Using the more general `Sequence` allows us to use `list` where it already exists and enables covariance for `asset_types`.

great_expectations/datasource/fluent/pandas_datasource.py
NathanFarmer: The ellipsis can be thought of as the same as `pass` when it is used like this. I just like how it looks.

alexsherstinsky (approved):
LGTM (only a couple of non-blocking questions).  Thanks!

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -248,9 +254,76 @@ def _validate_batch_request(self, batch_request: BatchRequest) -> None:
...
+        # batch_definition (along with batch_spec and markers) is only here to satisfy a
+        # legacy constraint when computing usage statistics in a validator. We hope to remove
+        # it in the future.
+        # imports are done inline to prevent a circular dependency with core/batch.py
+        from great_expectations.core import IDDict
+        from great_expectations.core.batch import BatchDefinition
+
+        batch_definition = BatchDefinition(
+            datasource_name=self.datasource.name,
+            data_connector_name="experimental",
```
Kilo59, line 297: `non-blocking`
Can we use a constant for this?
Also, should we change the name to `fluent`?

☝️ these might be better as followups

Kilo59 (approved):

great_expectations/datasource/fluent/pandas_datasource.py
NathanFarmer: I will do this in a follow-up PR next.

PR #7285, opened 2023-03-04
[BUGFIX] AssertError Duing Great Expectations Installation

richardohara:
This PR is submitted to resolve issue 7239.

While installing Great Expectations on a machine with Windows OS, the following AssertError is raised:

The extras requirements dir (req) contains files that do not adhere to the following format: requirements-dev-*.txt

Investigation revealed the the use of forward slashes ("/") do not work on Windows without the use of Path() from pathlib.

- Wrap variables that contain a "/" with Path()
- Change several variable names to eliminate issues raised by PEP.


setup.py
```diff
@@ -110,4 +119,4 @@ def get_extras_require():
     ],
 }
 
-setup(**config)
+setup(**config)
```
anthonyburdi: Minor nit: Let's make sure to leave the newline at the end of the file.

anthonyburdi (changes requested):
Hi @richardohara! Thank you so much for the contribution! Thank you for helping our Windows users :) The changes in setup.py look great though it looks like a few unrelated changes creeped in related to Athena, would you mind merging develop into your branch and handling the conflicts?

setup.py
```diff
@@ -34,18 +35,26 @@ def get_extras_require():
...
     requirements_dir = "reqs"
-    rx_fname_part = re.compile(rf"{requirements_dir}/requirements-dev-(.*).txt")
-    for fname in glob(f"{requirements_dir}/*.txt"):
-        match = rx_fname_part.match(fname)
+    glob_path = Path(f"{requirements_dir}/*.txt")
+    rx_name_part = re.compile(rf"requirements-dev-(.*).txt")
+
+    # Loop through each requirement file and verify they are named
+    # correctly and are in the right location.
+    for glob_file_name in glob(f"{glob_path}"):
```
Kilo59, line 46: If we are switching to using `pathlib` for this (which I strongly agree with 😄 ), we can do the `glob` operation directly from the `pathlib.Path` object.
https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob

We don't even need to import `glob`.

anthonyburdi (approved):
Thanks again for this contribution! Our Windows users will thank you 🙇 Thanks to @Kilo59 we are moving to using pathlib everywhere which should enable more seamless use on Windows, though we don't officially support it yet. I just updated this PR with our develop branch and fixed the newline issue. Let's merge this but I agree with @Kilo59's suggestion to use pathlib glob if you'd like to include that in a follow up contribution it would be very welcome!

richardohara:
Sorry for the delay in finishing this, but I took a few days off for some interview prep. 😊

I have a code change ready to commit ready that incorporates the change suggested by @Kilo59.

Windows is not officially supported? Oh dear. Time to set up WSL.

PR #7305, opened 2023-03-08
Fix a bug in the hourly generator. Add hourly functionality to genera…

abegong:
- Small PR to add methods to generate hourly data, in preparation for creating time series models that function at grains other than dates.
- Also fixes a bug where the Hourly generator would fail if size wasn't an exact multiple of 24.
-


contrib/time_series_expectations/assets/generate_test_time_series_data.py
```diff
@@ -19,30 +20,55 @@ def generate_time_series_data_and_plot(
...
-            }
-        ],
-        weekday_dummy_params=[
-            weekly_seasonality * x for x in [0, 0.5, 3.5, 4, 5, -3, -3.5]
-        ],
-        annual_seasonality_params=[],
-        holiday_alpha=1000,
-        outlier_alpha=outlier_alpha,
-    )
+    if grain == "daily":
```
anthonyburdi, line 31: Non blocking suggestion to make this an enum and use the enum as the type hint for the grain param. This just makes it more clear to the caller what values are allowable.

anthonyburdi (approved):
LGTM - just one small optional suggestion.

contrib/time_series_expectations/assets/generate_test_time_series_data.py
abegong: yeah, that would be the more maintainable thing to do. This is just a utility script, so I'm not going to polish it up to that level, at least not yet.

PR #7315, opened 2023-03-09
[MAINTENANCE] Fluent Datasource stubs for dynamic Pandas add_asset methods

Kilo59:
- add stub files for all Datasource that rely on dynamically defined assets (Pandas)
- create a minimal script for helping to generate the `add_<asset_name>_asset()` method stubs


gen_stub.py
```diff
@@ -0,0 +1,45 @@
+from inspect import Parameter, Signature
```
Kilo59, line 1: move this to `/scripts` or delete it...

tests/datasource/fluent/test_pandas_filesystem_datasource.py
```diff
@@ -54,7 +54,7 @@ def pandas_filesystem_datasource() -> PandasFilesystemDatasource:
         .parent.joinpath(base_directory_rel_path)
         .resolve(strict=True)
     )
-    return PandasFilesystemDatasource(
+    return PandasFilesystemDatasource(  # type: ignore[call-arg]
```
NathanFarmer, line 57: Why did this need to get added?

Kilo59: This is because it's missing the `type` field.
But I'm not sure 
1. Why it wasn't complaining before
2. Why it's even an issue because the type field isn't required, it has a default.

great_expectations/datasource/fluent/pandas_file_path_datasource.pyi
```diff
@@ -0,0 +1,26 @@
...
+class ORCAsset(_FilePathDataAsset): ...
+class ParquetAsset(_FilePathDataAsset): ...
+class FeatherAsset(_FilePathDataAsset): ...
+class HDFAsset(_FilePathDataAsset): ...
+class HTMLAsset(_FilePathDataAsset): ...
+class PickleAsset(_FilePathDataAsset): ...
+class SASAsset(_FilePathDataAsset): ...
+class SPSSAsset(_FilePathDataAsset): ...
+class STATAAsset(_FilePathDataAsset): ...
+class XMLAsset(_FilePathDataAsset): ...
```
NathanFarmer, line 22: What purpose does this serve?

NathanFarmer (approved):
One non-blocking question

scripts/gen_stub.py
```diff
@@ -0,0 +1,45 @@
+from inspect import Parameter, Signature
+from typing import ForwardRef, Type
+
+from great_expectations.datasource.fluent import Datasource, PandasFilesystemDatasource
+
```
cdkini, line 5: Not necessary now but would be nice to have a top level docstring explaining what this is

scripts/gen_stub.py
```diff
@@ -0,0 +1,45 @@
...
+
+def print_add_asset_method_signatures(datasource_class: Type[Datasource]):
+    type_lookup = datasource_class._type_lookup
+
+    for asset_type_name in type_lookup.type_names():
+        asset_type = type_lookup[asset_type_name]
+        method_name = f"add_{asset_type_name}_asset"
+        method = getattr(PandasFilesystemDatasource, method_name)
```
cdkini, line 13: This should be `datasource_class`?

scripts/gen_stub.py
```diff
@@ -0,0 +1,45 @@
...
+                    default = "..."
+                else:
+                    default = param.default
+                    if isinstance(default, str):
+                        default = f"'{default}'"
+                print(f" = {default}", end="")
+            print(",")
+
+
+print_add_asset_method_signatures(PandasFilesystemDatasource)
```
cdkini, line 45: if name == main?

great_expectations/datasource/fluent/pandas_file_path_datasource.pyi
Kilo59: This lets' `mypy` and the IDE properly understand these types. The pydantic generate model doesn't properly pass the type information.

We could do this better and also define all the specific attrs each asset has, but at least this lets the IDE and mypy know about the shared attributes that come from `_FilePathDataAsset`

scripts/gen_stub.py
Kilo59: Yes, good catch.

PR #7318, opened 2023-03-09
[BUGFIX] Scope `PandasDatasource` reader method types and serialization (2 of 2)

NathanFarmer:
- GREAT-1735
- Enable passing of URLs to pandas reader methods
- Make `_data_context` a `PrivateAttr` on `Datasource` to satisfy LSP

### Definition of Done


great_expectations/datasource/fluent/interfaces.py
```diff
@@ -373,6 +373,7 @@ class Datasource(
     assets: MutableMapping[str, _DataAssetT] = {}
 
     # private attrs
+    _data_context = pydantic.PrivateAttr()
```
Kilo59, line 376: `non-blocking`
Sorry I've already forgotten why we need this. But could we just store whatever the needed attribute/object is from the datacontext instead of the whole context?

Kilo59 (approved):

great_expectations/datasource/fluent/interfaces.py
NathanFarmer: We need it to eventually call `_data_context.get_validator()`, but we can't do that until we build a batch request from an asset. I don't know of another way than this.

great_expectations/datasource/fluent/sources.py
```diff
@@ -273,28 +273,21 @@ def factories(self) -> List[str]:
...
-            wrapped.__signature__ = ds_constructor.__signature__  # type: ignore[attr-defined]
-            return wrapped
         except KeyError:
             raise AttributeError(f"No factory {attr_name} in {self.factories}")
 
+        def wrapped(name: str, **kwargs):
+            datasource = ds_constructor(name=name, **kwargs)
+            datasource._data_context = self._data_context
+            # TODO (bdirks): _attach_datasource_to_context to the AbstractDataContext class
+            self._data_context._attach_datasource_to_context(datasource)
```
NathanFarmer, line 283: Try/except removed from here now that `_data_context` exists on all `Datasource`s.

NathanFarmer: Sorry actually on line 281

great_expectations/datasource/fluent/pandas_datasource.pyi
```diff
@@ -639,7 +639,7 @@ class PandasDatasource(_PandasDatasource):
     ) -> Validator: ...
     def read_html(
         self,
-        io: str | os.PathLike,
+        io: os.PathLike | str,
```
Kilo59, line 642: Order shouldn't matter here.
Consistency is nice, but it won't change runtime behavior like it does on a pydantic field.

NathanFarmer: I wanted everything to match, but in theory this could use `@validate_arguments` although I don't think we would want to.

PR #7323, opened 2023-03-09
 [BUGFIX] Fluent Datasources: Ensure that Datasource and DataAsset Model Serialization Does Not Include `"name"` Field

alexsherstinsky:
### Scope
Ensuring that `Datasource` and `DataAsset` model serialization does not include `"name"` field involved modifying serialization methods so as to filter out `"name"` fields (this is accomplished using generalized constants and methods).

- JIRA: GREAT-1709/GREAT-1726
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/datasource/fluent/fluent_base_model.py
```diff
@@ -39,6 +42,50 @@
 _Self = TypeVar("_Self", bound="FluentBaseModel")
 
 
+def _exclude_name_fields_from_fluent_datasources(
+    config: Dict[str, Any]
+) -> Dict[str, Any]:
+    if "fluent_datasources" in config:
+        fluent_datasources: dict = config["fluent_datasources"]
```
Kilo59: Why is this in the `FluentBaseModel`?
I think this only applies when `GxConfig` model tries to serialize.

I think this will affect the `repr` of our `Datasource` and `DataAsset` which is not desirable.

alexsherstinsky: @Kilo59 I put these utilities here, because they are used only in this module.  But they can be relocated.  I will let you know.  Thank you.

great_expectations/datasource/fluent/fluent_base_model.py
```diff
@@ -101,6 +102,7 @@ def yaml(
     ) -> pathlib.Path:
         ...
 
+    @public_api
```
NathanFarmer, line 105: @Kilo59 is this something we would want in the public API?

NathanFarmer: I don't expect that it will show up yet given that the class is not decorated.

great_expectations/datasource/fluent/config.py
```diff
@@ -139,3 +177,239 @@ def parse_yaml(
...
+        # custom
+        config_provider: _ConfigurationProvider | None = None,
+    ) -> dict[str, Any]:
+        """
+        Generate a dictionary representation of the model, optionally specifying which
+        fields to include or exclude.
+        Deviates from pydantic `exclude_unset` `True` by default instead of `False` by
+        default.
+        """
+        intermediate_dict: dict = super().dict(
```
NathanFarmer, line 353: Something about calling `super()` first feels wrong.

great_expectations/datasource/fluent/config.py
```diff
@@ -139,3 +177,239 @@ def parse_yaml(
...
+                exclude=exclude,
+                by_alias=by_alias,
+                exclude_unset=exclude_unset,
+                exclude_defaults=exclude_defaults,
+                exclude_none=exclude_none,
+                encoder=encoder,
+                models_as_dict=models_as_dict,
+                **dumps_kwargs,
+            )
+        )
```
NathanFarmer, line 329: How does this differ from the superclass implementation?

NathanFarmer (approved):
Merging for release, but we should revisit my comments.

great_expectations/datasource/fluent/config.py
```diff
@@ -139,3 +177,239 @@ def parse_yaml(
...
+
+    @public_api
+    def dict(
```
NathanFarmer, line 332: I also don't think we should need to touch `dict`. This should be handled in `json` or maybe also `yaml`, but I think using `dict` means it can be dropped from the model internally.

great_expectations/datasource/fluent/fluent_base_model.py
Kilo59: Yeah I agree. `FluentBaseModel` is also an implementation detail. It has no meaning to a user of `gx`.
Perhaps it should be `_FluentBaseModel` to make this more apparent.

alexsherstinsky: @Kilo59 @NathanFarmer The only reason for the `@public_api` was to deal with the public API checker, which kept giving errors; I am happy to remove it as long as that we can agree that I can also modify the public API count accordingly.  I ended up lowering the count by 1, but if we do not use `@public_api`, the could would need to be adjusted by 1 higher.  Adding @anthonyburdi -- thank you.

great_expectations/datasource/fluent/config.py
alexsherstinsky: @NathanFarmer Here is the rationale: we must modify the configuration dictionary.  Our starting point is the original.  The original is in `super()`.  I believe (now after some experience with it), there could be a way to do it all through `json()` as @Kilo59 has been pointing out; I will attempt this in the next pull request and if it is possible, then it will show up as part of the changes.  Thank you.

great_expectations/datasource/fluent/config.py
alexsherstinsky: @NathanFarmer This is a good point; by not repeating it here, superclass will call `json()`, which will be called from this class (as it is a more specific implementation), which will then run into infinite recursion.  So we need an independent `json_dict()` method in the subclass.  Thanks.

great_expectations/datasource/fluent/config.py
alexsherstinsky: @NathanFarmer As per above, I initially agree and will attempt it in the next pull request (it is a good idea to try, even if it does not work!).  Thanks.

great_expectations/datasource/fluent/fluent_base_model.py
NathanFarmer: I don't think this class is referenced anywhere in the docs, so either there is something weird going on with the method names, or the report was misinterpreted.

PR #7325, opened 2023-03-09
[BUGFIX] BigQuery performance uses updated `add_or_update_expectation_suite()` method

Shinnnyshinshin:
- Fixes failure in `cloud_db_integration` pipeline
- Updates util used by the performance test to use `add_or_update_expectation_suite()`


anthonyburdi (approved):
Good catch!

Issue #7239, opened 2023-02-27
AssertError During Great Expectations Installation 


richardohara:
While setting up a development environment I ran the following installation command:
`pip install -e .`
And received the following error message:
`The extras requirements dir (req) contains files that do not adhere to the following format: requirements-dev-*.txt`

1. Fork and clone the development branch
2. Follow the instructions to set up a development environment
3. Execute the command pip install -e .
4. See the error message

The expected behavior is the install will successfully complete

Environment
 - Operating System: Windows]
 - Great Expectations Version: 

Additional Info
The bug is due to code in Setup.py. The code that sets up the directory path does not work with Windows machines. I have a fix and will submit a PR.

cdkini:
Hey @richardohara, thanks for reaching out about this! A fix here would be greatly appreciated - we look forward to your PR.

richardohara:
I submitted a pull request today. I somehow ended up with extra commits in the PR, three of which are mine and one that is not. I am not sure how to remove the extra commits. My commit that should be use is 6bd7a68. Sorry for making a mess of this.

cdkini:
Apologies for the delayed response here @richardohara - looks like the changes are in! I'll go ahead and close this issue, thanks for your help!

PR #7327, opened 2023-03-09
[BUGFIX] Use Path().glob Instead of Path(). Issue #7239

richardohara:
This is a follow on pull request to PR #7285

- Use Path().glob instead of Path()
- Remove glob import statement

NOTE: Please delete commit a4da10b. I was not successful in reverting this commit


setup.py
```diff
@@ -34,18 +34,23 @@ def get_extras_require():
...
     requirements_dir = "reqs"
-    rx_fname_part = re.compile(rf"{requirements_dir}/requirements-dev-(.*).txt")
-    for fname in glob(f"{requirements_dir}/*.txt"):
-        match = rx_fname_part.match(fname)
+    rx_name_part = re.compile(rf"requirements-dev-(.*).txt")
+
+    # Use Path() from pathlib so we can make this section of the code OS agnostic.
+    # Loop through each requirement file and verify they are named
+    # correctly and are in the right location.
+    for path_and_file_name in Path().glob(f"{requirements_dir}/*.txt"):
```
Kilo59:
```suggestion
    for file_path in Path().glob(f"{requirements_dir}/*.txt"):
```
Minor point.
`file_name` makes me think it's a string, but it's a `Path` object.

setup.py
```diff
@@ -34,18 +34,23 @@ def get_extras_require():
...
         assert (
             match is not None
         ), f"The extras requirements dir ({requirements_dir}) contains files that do not adhere to the following format: requirements-dev-*.txt"
         key = match.group(1)
         if key in ignore_keys:
             continue
-        with open(fname) as f:
-            parsed = [str(req) for req in pkg_resources.parse_requirements(f)]
+        with open(path_and_file_name) as parsed_file:
+            parsed = [str(req) for req in pkg_resources.parse_requirements(parsed_file)]
```
Kilo59:
```suggestion
        with open(file_path) as f:
            parsed = [str(req) for req in pkg_resources.parse_requirements(f)]
```
I think `f` or maybe `f_in`/`f_out` is a good (and conventional) name for `open` file objects.

I don't really like `parsed_file` because it implies it's already been read and parsed.

richardohara: Thank you for the feedback. File is updated with y'alls suggestions.

Shinnnyshinshin (approved):
Thank you @richardohara for this PR and for addressing the comments / suggestions. LGTM

setup.py
```diff
@@ -1,11 +1,11 @@
 import re
-from glob import glob
 from pathlib import Path
 
 import pkg_resources
 from setuptools import find_packages, setup
 
 import versioneer
+from pathlib import Path
```
Kilo59: This is imported twice.

Kilo59 (changes requested):
Need to fix the double path import

setup.py
Shinnnyshinshin: Oh oops, this is my fault. Let me do a quick fix

Shinnnyshinshin (commented):
removing double import which is being caught by the `linter`

Kilo59 (approved):

PR #7334, opened 2023-03-10
[FEATURE] add optional `id` to Fluent Datasources and DataAsset schemas

Kilo59:
- adds `id` as an optional field for Fluent `Datasource` and `DataAsset`
- ensures that the `id` field is only serialized if it is present
- updates all schemas
- Put the base `DataSource` schema under source control.


great_expectations/datasource/fluent/schemas/Datasource.json
```diff
@@ -0,0 +1,87 @@
+{
+    "title": "Datasource",
+    "description": "Base model for most fluent datasource related pydantic models.\n\nAdds yaml dumping and parsing methods.\n\nExtra fields are not allowed.\n\nSerialization methods default to `exclude_unset = True` to prevent serializing\nconfigs full of mostly unset default values.\nAlso prevents passing along unset kwargs to BatchSpec.\nhttps://docs.pydantic.dev/usage/exporting_models/",
```
Kilo59, line 3: These descriptions are pulled from the class docstrings.
But because neither `Datasource` or `DataAsset` declare their own docstring we get the docstring from `FluentBaseModel`

tests/datasource/fluent/test_schemas.py
```diff
@@ -28,6 +28,8 @@ def _models_and_schema_dirs() -> Generator[
     datasource: Type[Datasource] = Datasource
     ds_type_name: str = ""
 
+    yield Datasource, _SCHEMAS_DIR, f"datasource:{Datasource.__name__}"
```
Kilo59, line 31:
```suggestion
    yield Datasource, _SCHEMAS_DIR, Datasource.__name__
```

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -151,6 +152,7 @@ class DataAsset(FluentBaseModel, Generic[_DatasourceT]):
     # * type: Literal["csv"] = "csv"
     name: str
     type: str
+    id: Optional[uuid.UUID] = Field(default=None, description="DataAsset id")
```
NathanFarmer, line 155: Why a description here, but nowhere else?

NathanFarmer: I would expect to use descriptions for other fields that are repeated throughout `FluentBaseModel`s (e.g. `name` and `type`).

tests/datasource/fluent/test_config.py
```diff
@@ -283,6 +292,48 @@ def test_from_gx_config(self, asset_dict: dict):
         )
 
 
+def test_id_only_serialized_if_present(ds_dict_config: dict):
```
NathanFarmer, line 295: Why are we allowing `id` to be `Optional`?

great_expectations/datasource/fluent/interfaces.py
Kilo59: I agree we should have descriptions for other fields as well, but I don't want to do that as part of this PR.

tests/datasource/fluent/test_config.py
Kilo59: Because nothing relies on it.
Cloud will generate IDs for the objects, but that will be done server-side, not by the client.

NathanFarmer (approved):
LGTM!

Kilo59:
>One thing to look out for is UUID / str type mismatch. Generally, we are using str everywhere except the database and the REST endpoints.

@dctalbot 

Yeah, I've experienced that exact problem many times.

Pydantic will attempt to coerce a string uuid into an actual `UUID` object, so it will only fail if it's not a valid uuid format.
It also will always serialize the item to a string if `.yaml()` or `.json()` is called.

`mypy` should also help ensure we don't forget to convert to/from a `str` when comparing `id`s.


------------

We could use a stricter `UUID4` type (assuming that's the uuid type we use) and ensure that only `uuid4`s (or `str`/`bytes` of `uuid4`) are possible. But that doesn't seem very useful.

https://docs.pydantic.dev/usage/types/#pydantic-types

PR #7336, opened 2023-03-10
[BUGFIX] Ensure Correct JSON and Dictionary Dump Output of Serialized Fluent Objects

alexsherstinsky:
### Scope
* Ensure that serialized Fluent `Datasource` and `DataAsset` objects contain the `name` key in their `JSON` and Dictionary output dumps.
* Add validation checks to ensure that `name` keys in `Datasource` and `DataAsset` are consistent with `name` attribute values in their configuration sections.
* Ensure that `YAML` dump output of Fluent `Datasource` objects that do not have `assets` does not contain an empty `assets` section.
* New tests are created for the above listed behaviors.

### Remarks
* While output schema of `JSON` and Dictionary Fluent output dumps with this change is correct, there is still duplication of `name` attribute value between configuration section keys and `name` attribute values in the corresponding sections for Fluent `Datasource` and `DataAsset` configurations.  Despite validation, currently implemented to prevent `name` inconsistency, this duplication should be corrected as part of a future enhancement.

- JIRA: GREAT-1709/GREAT-1726
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


great_expectations/datasource/fluent/config.py
```diff
@@ -111,13 +107,25 @@ def _load_datasource_subtype(cls, v: Dict[str, dict]):
                     f"'{ds_name}' has unsupported 'type' - {type_lookup_err}"
                 ) from type_lookup_err
 
-            config["name"] = ds_name
+            if "name" in config:
+                if config["name"] != ds_name:
```
Kilo59:
```suggestion
            ds_name = config.get("name",  MISSING) # MISSING = object()
            if ds_name is not MISSING and ds_name != ds_key:
```

In the suggestion above `MISSING` is a global constant sentinel value that can be used to determine whether or not the `"name"` key is actually present.

Also in this example, I've renamed the existing `ds_name` to `ds_key` and use `ds_name` to represent the value of the `Datasource.name` if it's present.

great_expectations/datasource/fluent/config.py
```diff
@@ -111,13 +107,25 @@ def _load_datasource_subtype(cls, v: Dict[str, dict]):
...
+                    raise ValueError(
+                        f'Datasource name "{ds_name}" is different from value in its "name" entry.'
```
Kilo59: I would show both values.
```suggestion
                    raise ValueError(
                        f'Datasource name "{ds_name}" is different from value in its "name" entry "{<OTHER_VALUE>}".'
```

great_expectations/datasource/fluent/config.py
```diff
@@ -246,127 +254,16 @@ def yaml(
         intermediate_json_dict = self._exclude_name_fields_from_fluent_datasources(
             config=intermediate_json_dict
         )
+        deep_filter_properties_iterable(
+            properties=intermediate_json_dict, clean_falsy=True, inplace=True
```
Kilo59: Why are we doing this?
This seems wrong and feels like a hidden foot gun waiting to go off.

My assumption is that it's to remove an empty `assets` key, but this is not how I would solve that. In fact, I don't think that issue necessarily needs to be done as part of this PR. It seems unrelated to the other fixes.

Kilo59 (changes requested):
Thanks for doing this!

There's a least 1 foot-gun in this PR I would like to see disarmed before the merge.
Also made a couple of unrelated suggestions.

great_expectations/datasource/fluent/config.py
alexsherstinsky: @Kilo59 I incorporated your suggestion for the name change -- for both `Datasource` (key/name) and `DataAsset` key/name as well as your suggestion to report both the key and the name when they do not match (and updated the tests accordingly).  However, I did not create `MISSING` -- we have not used this pattern, and I did not want to create a global sentinel just for this; I feel that the current pattern of test/act (test for key's presence, act if it exists) is easy to read and works well.  Thank you.

great_expectations/datasource/fluent/config.py
alexsherstinsky: @Kilo59 Indeed, we were asked to not include `assets` in the `YAML` representation, if there are no data assets in the given `Datasource` configuration, rather than to show an empty dictionary in the `YAML` string.  I am not sure if this is documented, but I distinctly remember being asked to do that (otherwise, I would not have done it).

So your assumption was correct.  The only times that using `deep_filter_properties_iterable()` is not appropriate is if there are possibilities of having keys in `YAML` with `NULL` (or empty) values, whereby the keys must still be shown.  If this will not be the case with the `Fluent` Datasources, then we can keep using `deep_filter_properties_iterable()`.  However, if you feel that there will be such cases potentially in the future, then `deep_filter_properties_iterable()` will not work.  Could you please describe the approach you had in mind for solving it?  Is it looping through the configuration of every `Datasource` and removing only the `assets` key (if it has a `Falsy`) value?  Please let me know so that we can finalize the implementation.  Thank you.

NathanFarmer: I think the point @Kilo59 is making is that we can handle this logic via `pydantic` with the overloaded `yaml()` method, like we did to remove `name`.

Kilo59 (changes requested):
Please remove `deep_filter_properties_iterable()`
I'll create a separate story for the empty `assets` key.

great_expectations/datasource/fluent/config.py
Kilo59: @alexsherstinsky I think it's okay if you decide not to use this pattern but it is very common in programming.
I would be surprised if we don't do this at least once in the `great_expectations` codebase.

There's even a draft pep right now that proposes a convenience function/class for creating these sentinel types. The abstract does a good job of explaining the use of the pattern. However, the discussion about its implementation is not relevant to this conversation.
https://peps.python.org/pep-0661/

----

The thing it would do for us, in this case, is let us use `1` conditional instead of `2` when checking for the `ds_name`. You don't have to check if the `name` key is present and then check the name value, you can do both checks in one conditional by testing for the presence of the `sentinel` value.

alexsherstinsky: @Kilo59 I have used this pattern in other projects; however, I have not seen it in GreatExpectations (at least not in a widely utilized way -- if you have seen a usage, please provide the link; thanks).  Introducing it in a truly global way seems like should be part of a separate change (which would need to be discussed with the team).  Thanks.

great_expectations/datasource/fluent/config.py
alexsherstinsky: @NathanFarmer I would be curious to see the description (and happy to implement it) -- I handled the `name` using the validator, though the overloaded methods differ only in their method signature; is this not the case?  Thanks. /cc @Kilo59

tests/datasource/fluent/test_config.py
```diff
@@ -402,15 +497,23 @@ def test_general_splitter_errors(
...
-def from_dict_gx_config() -> GxConfig:
+def from_simple_dict_gx_config() -> GxConfig:
+    gx_config = GxConfig.parse_obj(SIMPLE_DS_DICT)
+    assert gx_config
+    return gx_config
+
+
+@pytest.fixture
+@functools.lru_cache(maxsize=1)
+def from_complex_dict_gx_config() -> GxConfig:
```
Kilo59: Can you remove these `from_complex_dict_gx_config` vs `from_simple_dict_gx_config` extra fixtures.
We gain nothing from having a `from_simple_dict_gx_config` fixture

alexsherstinsky: @Kilo59 I deleted  `from_simple_dict_gx_config` fixture -- it is not currently used, now that `assets` removal is not part of this change; good catch.  Thanks.

tests/datasource/fluent/test_config.py
```diff
@@ -402,15 +497,15 @@ def test_general_splitter_errors(
 
 @pytest.fixture
 @functools.lru_cache(maxsize=1)
-def from_dict_gx_config() -> GxConfig:
+def from_complex_dict_gx_config() -> GxConfig:
```
Kilo59: Please revert the name change as well

great_expectations/datasource/fluent/config.py
```diff
@@ -96,28 +91,42 @@ def _load_datasource_subtype(cls, v: Dict[str, dict]):
...
-                asset_config["name"] = asset_name
+            for asset_key, asset_config in config["assets"].items():
+                if "name" in asset_config:
+                    asset_name: str = asset_config["name"]
+                    if asset_name != asset_key:
+                        raise ValueError(
+                            f'DataAsset key "{asset_key}" is different from name "{asset_name}" in its configuration.'
+                        )
+                else:
+                    asset_config["name"] = asset_key
```
Kilo59, line 129: IMO these belong in a different validator method. This check has nothing to do with `_load_datasource_subtype`.
Maybe call it `_check_datasource_names_and_keys_match()`

Also, the check for `DataAsset` names should be on the `Datasource` class, not the `GxCofig` class.
`_check_asset_names_and_keys_match()`

You don't need to change the logic.

great_expectations/datasource/fluent/config.py
Kilo59: Here's a draft PR for the `assets` key fix.
- https://github.com/great-expectations/great_expectations/pull/7341

tests/datasource/fluent/test_config.py
alexsherstinsky: @Kilo59 Done -- assuming that you will create any fixtures with appropriate name(s) for testing cleaned out `assets` key (if any are needed, that is).

tests/datasource/fluent/test_config.py
```diff
@@ -443,16 +531,16 @@ def test_dict_config_round_trip(
 
 
 def test_json_config_round_trip(
-    inject_engine_lookup_double, from_json_gx_config: GxConfig
+    inject_engine_lookup_double, from_dict_gx_config: GxConfig
```
Kilo59, line 534:
```suggestion
    inject_engine_lookup_double, from_json_gx_config: GxConfig
```

tests/datasource/fluent/test_config.py
```diff
@@ -489,9 +577,9 @@ def test_yaml_file_config_round_trip(
 
 
 def test_splitters_deserialization(
-    inject_engine_lookup_double, from_json_gx_config: GxConfig
+    inject_engine_lookup_double, from_dict_gx_config: GxConfig
```
Kilo59, line 580:
```suggestion
    inject_engine_lookup_double, from_json_gx_config: GxConfig
```

tests/datasource/fluent/test_config.py
```diff
@@ -512,9 +600,9 @@ def test_yaml_config_round_trip_ordering(
 
 @pytest.mark.xfail(reason="Custom Sorter serialization logic needs to be implemented")
 def test_custom_sorter_serialization(
-    inject_engine_lookup_double, from_json_gx_config: GxConfig
+    inject_engine_lookup_double, from_dict_gx_config: GxConfig
```
Kilo59, line 603:
```suggestion
    inject_engine_lookup_double, from_json_gx_config: GxConfig
```

Kilo59 (approved):
Thanks for doing this.

I'm okay with this being done as a followup.
https://github.com/great-expectations/great_expectations/pull/7336/files#r1134157575

PR #7341, opened 2023-03-13
[BUGFIX] Fluent Datasources - empty "assets" key on serialization

Kilo59:
- Remove `'assets'` from `_FIELDS_ALWAYS_SET`
- Only consider the `assets` part of `__fields_set__` if the `Datasource` actually contains `assets`.
- Add private function `_update__fields_set__on_truthyness()`

### Definition of Done


ghost:
<details open>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=7341&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

#### 

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=7341&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map legend
</details>

tests/datasource/fluent/test_config.py
```diff
@@ -488,6 +489,28 @@ def test_yaml_file_config_round_trip(
...
+):
+    ds_wo_assets = None
+    ds_with_assets = None
+    for ds in from_yaml_gx_config.datasources.values():
+        if not ds.assets:
+            ds_wo_assets = ds
+        else:
+            ds_with_assets = ds
+    assert ds_with_assets, "Need at least one Datasource without assets for this test"
+    assert ds_wo_assets, "Need at least one Datasource with assets for this test"
```
Kilo59, line 503:
```suggestion
    assert ds_with_assets, "Need at least one Datasource with assets for this test"
    assert ds_wo_assets, "Need at least one Datasource without assets for this test"
```

Kilo59:
Wait for merge from
- https://github.com/great-expectations/great_expectations/pull/7336

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -463,10 +462,6 @@ def add_asset(self, asset: _DataAssetT) -> _DataAssetT:
 
         self.assets[asset.name] = asset
 
-        # pydantic needs to know that an asset has been set so that it doesn't get excluded
-        # when dumping to dict, json, yaml etc.
-        self.__fields_set__.update(_FIELDS_ALWAYS_SET)
```
NathanFarmer, line 468: Why don't we need this anymore?

tests/datasource/fluent/test_metadatasource.py
```diff
@@ -409,7 +409,6 @@ def context_with_fluent_datasource(
         config_file_path,
         {
             DEFAULT_CRUD_DATASOURCE_NAME: {
-                "assets": {},
```
NathanFarmer, line 412: ❤️

NathanFarmer (approved):

great_expectations/datasource/fluent/interfaces.py
Kilo59: Because we do this same logic before serialization. It's very likely we didn't actually need this before.

great_expectations/datasource/fluent/fluent_base_model.py
```diff
@@ -294,3 +301,28 @@ def _recursively_set_config_value(
                 data[i] = v.get_config_value(config_provider)
             elif isinstance(v, (MutableMapping, MutableSequence)):
                 return _recursively_set_config_value(v, config_provider)
+
+
+def _update__fields_set__on_truthyness(model: FluentBaseModel, field_name: str) -> None:
```
alexsherstinsky, line 306: @Kilo59 Do we need the double underscore between `set` and `on` in this method's name?  Thanks.

great_expectations/datasource/fluent/constants.py
```diff
@@ -6,8 +6,9 @@
 # methods are called. Otherwise it could be excluded.
 # https://docs.pydantic.dev/usage/exporting_models/#modeldict
 _FIELDS_ALWAYS_SET: Final[set[str]] = {
```
alexsherstinsky, line 8: ❤️

great_expectations/datasource/fluent/fluent_base_model.py
```diff
@@ -294,3 +301,28 @@ def _recursively_set_config_value(
...
+    """
```
alexsherstinsky, line 307: @Kilo59 I am approving, because I do not have a better proposal at the moment, other than to express that being overly invested in the inner workings of a framework (`Pydantic` in this case) may cause maintenance difficulties in the future (e.g., if the framework's behavior changes).  In addition, the use of `getattr()` masks the specifics of the the field being used, and while appropriate in certain situation may cause unforeseen results, if used on unintended field names (e.g., by accident) -- @cdkini may wish to offer an opinion on `getattr()`.  Finally, a question: is the absence of `assets` a normal situation, or a corner case?  If the former, then perhaps the `debug` log messages may not be needed.  Thank you.

tests/datasource/fluent/test_config.py
```diff
@@ -488,6 +489,28 @@ def test_yaml_file_config_round_trip(
     assert from_yaml_gx_config == re_loaded
 
 
+def test_assets_key_presence(
+    inject_engine_lookup_double, from_yaml_gx_config: GxConfig
+):
+    ds_wo_assets = None
+    ds_with_assets = None
+    for ds in from_yaml_gx_config.datasources.values():
+        if not ds.assets:
```
alexsherstinsky: @Kilo59 Non-blocking: stylistically, when the `if` clause contains the `else`, it might be less verbose to put the negative branch into the `else` part.  Thanks.

tests/datasource/fluent/test_config.py
```diff
@@ -488,6 +489,28 @@ def test_yaml_file_config_round_trip(
...
+            ds_wo_assets = ds
+        else:
+            ds_with_assets = ds
```
alexsherstinsky, line 587: @Kilo59 Nice way to test all variations without additional fixtures!

alexsherstinsky (changes requested):
Will LGTM (after resolving the method name with the double underscore and acknowledging the non-blocking comments).  Thanks.

great_expectations/datasource/fluent/fluent_base_model.py
Kilo59: It is referencing the `__fields_set__` attribute of a pydantic model.

great_expectations/datasource/fluent/fluent_base_model.py
Kilo59: Yes, we are relying on some implementation details for `pydantic` here. But all of the things we rely on are part of it's public documentation (example `__fields_set__`- https://docs.pydantic.dev/usage/models/#basic-model-usage )
I'm not worried about the framework changing because we are pinned to the last major version. If `pydantic` makes breaking changes (for us) in `v2` we don't have to use it.

>In addition, the use of getattr() masks the specifics of the the field being used, and while appropriate in certain situation may cause unforeseen results, if used on unintended field names (e.g., by accident)

I don't follow.
It is true that this logic doesn't need to be applied to `FluentBaseModel` and could be isolated to `Datasource.dict()`/`Datasource.json()`. But this is no worse than what we had previously and I wanted to keep these changes small and address the bug.

Kilo59: >Finally, a question: is the absence of assets a normal situation, or a corner case?

I would say it's very normal. A Datasource can be created and then saved or serialized without any `assets`.

tests/datasource/fluent/test_config.py
Kilo59, line 501:
```suggestion
        if ds.assets:
            ds_with_assets = ds
        else:
            ds_wo_assets = ds
```

tests/datasource/fluent/test_config.py
Kilo59: I agree.  Updated.

great_expectations/datasource/fluent/fluent_base_model.py
alexsherstinsky: Hence, the log messages are not needed perhaps?

Kilo59: I would prefer to keep the debug logs, at least for now.
We log at `warning` and above, users won't see this.

tests/datasource/fluent/test_config.py
alexsherstinsky: 🙇

great_expectations/datasource/fluent/fluent_base_model.py
alexsherstinsky: @Kilo59 Good point -- thank you.

alexsherstinsky (approved):
LGTM

tests/datasource/fluent/test_config.py
```diff
@@ -736,8 +759,7 @@ def test_config_substitution_retains_original_value_on_save(
         "fluent_datasources"
     ]["my_sqlite_ds_w_subs"]
 
-    # FIXME: serialized items should not have name or assets
-    round_tripped.pop("assets")
+    # FIXME: serialized items should not have name
     round_tripped.pop("name")
```
Kilo59, line 763:
```suggestion
```
This should now give us a key error and needs to be removed.

PR #7343, opened 2023-03-13
[MAINTENANCE] Test Pandas 2.0 prerelease in CI/CD

cdkini:
- Test Pandas 2.0 in async pipeline

### Definition of Done


ci/constraints-test/pandas2-min-install.txt
```diff
@@ -0,0 +1,2 @@
+numpy==1.23.0
+pandas==2.0.0rc0
```
Kilo59: Would be better if this was 
```
pandas>=2.0.0rc0
```
It would great if this just worked because if we have to add the `--pre` in our install step that might break the other matrix tests.

This SO answer makes me think this should just work without the `--pre` flag.
https://stackoverflow.com/a/54330251/6304433

Kilo59 (approved):

ci/azure-pipelines.yml
```diff
@@ -443,6 +443,9 @@ stages:
             Python310:
               python.version: '3.10'
               min_constraints: 'ci/constraints-test/py310-min-install.txt'
+            Pandas20:
+              python.version: '3.9'
+              min_constraints: 'ci/constraints-test/pandas2-min-install.txt'
```
NathanFarmer: The only possible issue I see with this PR, is this stage references "min versions" and here we are adding a pandas step instead of a python step. It probably doesn't belong here.

NathanFarmer (commented):
This seems to be passing all of a sudden. I just have one concern about the appropriateness of using this Azure stage.

ci/constraints-test/pandas2-min-install.txt
```diff
@@ -0,0 +1 @@
+pandas>=2.0.0rc0
```
billdirks: I'm not sure of the timing on things but if this wasn't passing before but now is, we may want to bump this to `2.0.0rc1`. 
The releases are here:
https://pypi.org/project/pandas/#history

ci/azure-pipelines.yml
billdirks: We might want to change this jobs name to something about testing `test_dependency_versions` and we can update the names of the strategy matrix to indicate what is being tested.

PR #7373, opened 2023-03-16
[BUGFIX] Remove Query ID from exception for query.template_values metric

itaise:

- Remove query_id from the exception raised by query.template_values metric:
The query.template_values metric may cause duplicated validation_results. 
In our case, when using Trino, it returns the TrinoUserError exception which contains query_id, so for each attempt to resolve the metric a different exception is returned, which cause the validation results to be duplicated 3 times - according to the number of attempts to resolve the metric. 

I have tested the change - When removing the query id from the exception the results are not duplicated. 

Here is a previous attempt to resolve this - We agreed that I will resolve this in the level of the query.template_values metric. 

https://github.com/great-expectations/great_expectations/pull/6697

Thank you!


kenwade4 (approved):
Looks good, thanks for these 2 new PRs @itaise!

PR #7377, opened 2023-03-16
updates to quickstart, to create working code

donaldheppner:
- Added a second expectation, so that the results look more interesting
- Removed the "convert to file" since I think this is extra and unnecessary for a quickstart
- Updated the code so it works
  - The file exists
  - The parameter is `data_context` (not `context`)
  - SimpleCheckpoint reference now resolves.

### DIFF
--- a/docs/docusaurus/docs/tutorials/quickstart/quickstart.md
+++ b/docs/docusaurus/docs/tutorials/quickstart/quickstart.md
@@ -60,6 +60,9 @@ checkpoint_result = checkpoint.run()
 # View results
 validation_result_identifier = checkpoint_result.list_validation_result_identifiers()[0]
 context.open_data_docs(resource_identifier=validation_result_identifier)
+
+# Save the Data Context for future use
+context.convert_to_file_context()
 ```
 
 In the following steps we'll break down exactly what is happening here so that you can follow along and perform a Validation yourself.
--- a/docs/docusaurus/docs/tutorials/quickstart/quickstart.md
+++ b/docs/docusaurus/docs/tutorials/quickstart/quickstart.md
@@ -41,17 +41,18 @@ import great_expectations as gx
 context = gx.get_context()
 
 # Connect to data
-validator = context.datasources.pandas_default.read_csv(
-    filepath_or_buffer="https://raw.githubusercontent.com/great_expectations/taxi_data.csv"
+validator = context.sources.pandas_default.read_csv(
+    "https://raw.githubusercontent.com/great-expectations/gx_tutorials/main/data/yellow_tripdata_sample_2019-01.csv"
 )
 
 # Create Expectations
 validator.expect_column_values_to_not_be_null("pickup_datetime")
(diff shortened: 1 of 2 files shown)


Rachel-Reverie (commented):
Looks good to me -- I approve the changes to the prose content.  I'll leave this as a comment since I don't have the ownership to approve the code changes (though I think they are also good).

docs/docusaurus/docs/tutorials/quickstart/quickstart.md
```diff
@@ -158,15 +157,6 @@ validation_result_identifier = checkpoint_result.list_validation_result_identifi
...
 
-#### 4.3 Save the Data Context for future use
-Because we did not previously initialize a Filesystem Data Context or specify a path at which to create one, the Data Context we recieved from `gx.get_context()` was a temporary, in-memory Ephemeral Data Context.  To save this Data Context for future use, we will convert it to a Filesystem Data Context:
-
-```python title="Python code"
-context = context.convert_to_file_context()
-```
-
-You can provide the path to a specific folder when you convert your Ephemeral Data Context to a Filesystem Data Context.  If you do, your Filesystem Data Context will be initialized at that location.  If you do not, your new Filesystem Data Context will be initialized in the folder that your script is executed in.
-
```
talagluck: @donaldheppner I can see why you would want to remove this from Quickstart, but I do think it should be readily available for users who want to know what comes next. Where would you expect this to live instead?

talagluck (commented):
LGTM overall. One question on the removal of `convert_to_file_context()`.

docs/docusaurus/docs/tutorials/quickstart/quickstart.md
donaldheppner: Do we have a page that describes an ephemeral data context? I'd rather it go there.

talagluck: I can see that, but I think we either need it here, or we need a link here to the next part of the Getting Started Tutorial where this is the first step. Otherwise, I'm afraid that users will have a really easy first 30 seconds and not know where to go next. I'd defer to @Rachel-Reverie though.

Rachel-Reverie: We have two guides for the Ephemeral Data Context that have been back-burnered due to the need to get all of the fluent Datasource guides in place.  One is the guide for explicitly creating an Ephemeral Data Context, the other is the guide for converting an Ephemeral Data Context to a Filesystem Data Context.  We could link the second of those to the end of the Quickstart, in the "Next steps" section.

Rachel-Reverie: I would suggest that we leave this section in for now, with the understanding that we will remove it and link to the "How to convert an Ephemeral Data Context to a Filesystem Data Context" guide in the "Next Steps" section once we have that guide through the review process.

PR #7387, opened 2023-03-16
[FEATURE] Make SQL datasources add asset methods public.

billdirks:
### Definition of Done


great_expectations/datasource/fluent/sqlite_datasource.py
```diff
@@ -170,6 +172,7 @@ class SqliteDatasource(SQLDatasource):
     _TableAsset: Type[SqlTableAsset] = pydantic.PrivateAttr(SqliteTableAsset)
     _QueryAsset: Type[SqlQueryAsset] = pydantic.PrivateAttr(SqliteQueryAsset)
 
+    @public_api
```
alexsherstinsky, line 175: @billdirks If we are making a method public, does it not also mean that we have to add a properly-formatted docstring?  Thanks.

alexsherstinsky (changes requested):
@billdirks If the answer to my comment about docstrings is "not necessary", then I will approve -- that was the only thing.  Thank you.

great_expectations/datasource/fluent/sqlite_datasource.py
billdirks: Done. I was under the mistaken impression that we would inherit these but apparently not. 
```
In [1]: import great_expectations as gx

In [2]: print(gx.datasource.fluent.sqlite_datasource.SqliteDatasource.add_query_asset.__doc__)
--Public API--Adds a query asset to this datasource.

        Args:
            name: The name of this table asset.
            query: The SELECT query to selects the data to validate. It must begin with the "SELECT".
            order_by: A list of Sorters or Sorter strings.

        Returns:
            The query asset that is added to the datasource.
            The type of this object will match the necessary type for this datasource.
            eg, it could be a QueryAsset or a SqliteQueryAsset.
```

great_expectations/datasource/fluent/schemas/SqliteDatasource.json
```diff
@@ -1,6 +1,6 @@
 {
     "title": "SqliteDatasource",
-    "description": "Adds a sqlite datasource to the data context.\n\nArgs:\n    name: The name of this sqlite datasource.\n    connection_string: The SQLAlchemy connection string used to connect to the sqlite database.\n        For example: \"sqlite:///path/to/file.db\"\n    assets: An optional dictionary whose keys are TableAsset names and whose values\n        are TableAsset objects.",
+    "description": "--Public API--Adds a sqlite datasource to the data context.\n\nArgs:\n    name: The name of this sqlite datasource.\n    connection_string: The SQLAlchemy connection string used to connect to the sqlite database.\n        For example: \"sqlite:///path/to/file.db\"\n    assets: An optional dictionary whose keys are TableAsset names and whose values\n        are TableAsset objects.",
```
NathanFarmer, line 3: Why do we need to add this?

billdirks: [The `@public_api` decorator adds this prefix (and that's all it does)](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/core/_docs_decorators.py#L16). I believe another step of the doc building process looks for this prefix when generating the docs? I haven't read that code.

great_expectations/datasource/fluent/sqlite_datasource.py
alexsherstinsky: @billdirks The content looks really good, but for some reason, I am not seeing this documentation in the actual `Python` module.  Please let me know if I am misunderstanding something here.  Thank you.

alexsherstinsky (approved):
LGTM

PR #7392, opened 2023-03-17
[FEATURE] `BatchMetadata` for all fluent `DataAsset`s

NathanFarmer:
- GREAT-1729
- Add `BatchMetadata` to all fluent `DataAsset`s
- Update pandas `add_*_asset` methods to accept `BatchMetadata`
- Change instances of `STATAAsset` to `StataAsset`

### Definition of Done


great_expectations/datasource/fluent/interfaces.py
```diff
@@ -159,6 +162,7 @@ class DataAsset(FluentBaseModel, Generic[_DatasourceT]):
     id: Optional[uuid.UUID] = Field(default=None, description="DataAsset id")
 
     order_by: List[Sorter] = Field(default_factory=list)
+    batch_metadata: Optional[BatchMetadata] = pydantic.Field(default_factory=dict)
```
billdirks, line 165: We can remove the `Optional` here. If the user doesn't pass this in the `default_factory` will make an empty dict.

NathanFarmer: For some reason I thought this would effect the schemas, but it doesn't so I made this change. ✅

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -372,7 +377,8 @@ def get_batch_list_from_batch_request(
             batch_spec_passthrough=None,
         )
 
-        batch_metadata = copy.deepcopy(batch_request.options)
+        batch_metadata = copy.deepcopy(self.batch_metadata) or {}
+        batch_metadata.update(copy.deepcopy(batch_request.options))
```
NathanFarmer, line 379: I think this is where the config variable substitution will also happen, so I'm planning to revisit this in a subsequent PR.

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -503,49 +507,85 @@ class PandasDatasource(_PandasDatasource):
...
 
+    @staticmethod
+    def _validate_asset_name(asset_name: Optional[str] = None) -> str:
+        if asset_name == DEFAULT_PANDAS_DATA_ASSET_NAME:
+            raise PandasDatasourceError(
+                f"""An asset_name of {DEFAULT_PANDAS_DATA_ASSET_NAME} cannot be passed because it is a reserved name."""
+            )
+        if not asset_name:
+            asset_name = DEFAULT_PANDAS_DATA_ASSET_NAME
+        return asset_name
```
Kilo59, line 518: `non-blocking`
If we do this as a standard pydantic validator method we will get the nice pydantic error locations/`loc` details
And then we don't have to remember to call it.

NathanFarmer: I started to do that, but the issue is that we need to do type conversion `Optional[str]` -> `str`. There is no way to do type conversion in a validator and have `mypy` pick-up on the types correctly.

NathanFarmer: @Kilo59 see: https://github.com/pydantic/pydantic/discussions/5050

great_expectations/datasource/fluent/dynamic_pandas.py
```diff
@@ -193,7 +193,7 @@ def _replace_builtins(input_: str | type) -> str | type:
     "spss": "SPSSAsset",
     "sql_query": "SQLQueryAsset",
     "sql_table": "SQLTableAsset",
-    "stata": "STATAAsset",
+    "stata": "StataAsset",
```
Kilo59:
```suggestion
```
Sorry, this `_METHOD_TO_CLASS_NAME_MAPPINGS` variable could have been named better.

You can just remove this entry. This mapping is only if we want to override the default naming convention.

https://github.com/great-expectations/great_expectations/blob/15af89bc1fc4b2a4ccfc4eb52f0f7cfa4b478433/great_expectations/datasource/fluent/dynamic_pandas.py#L397-L398

great_expectations/datasource/fluent/interfaces.py
Kilo59: Have you pushed this change? 
To me it still shows up.
```suggestion
    batch_metadata: BatchMetadata = pydantic.Field(default_factory=dict)
```

great_expectations/datasource/fluent/pandas_dbfs_datasource.pyi
```diff
@@ -53,6 +54,7 @@ class PandasDBFSDatasource(PandasFilesystemDatasource):
         batching_regex: Optional[Union[str, re.Pattern]] = ...,
         glob_directive: str = ...,
         order_by: Optional[SortersDefinition] = ...,
+        batch_metadata: Optional[BatchMetadata] = ...,
```
Kilo59, line 57: Is `None` ever an acceptable value here?
Not a question specifically for this method but all of `batch_metadata: Optional[BatchMetadata] = ...,`

great_expectations/datasource/fluent/interfaces.py
NathanFarmer: So the issue is that in-practice the `batch_metadata` gets passed in via `add` method as `Optional[BatchMetadata]`. Unless we handle the `None` case, the type checker will not allow this field to be strict `BatchMetadata` with a `dict` default. Handling the `None` case would require us to concretely define the `add` methods on all datasources and perform the type conversion. You can't do the type conversion in a validator and have `mypy` pick up on the types correctly (see: https://github.com/pydantic/pydantic/discussions/5050).

NathanFarmer: @billdirks What this also means is that `order_by` is not currently working for any Datasources other than `SQLDatasource`. For `order_by` we do type conversion in the `SQLDatasource` `add` methods, but nowhere else.

great_expectations/datasource/fluent/pandas_datasource.pyi
```diff
@@ -142,25 +143,35 @@ class PandasDatasource(_PandasDatasource):
...
+        self,
+        dataframe: pd.DataFrame,
+        *,
+        asset_name: Optional[str] = ...,
+        batch_metadata: Optional[BatchMetadata] = ...,
     ) -> Validator: ...
     def add_clipboard_asset(
         self,
         name: str,
-        order_by: typing.List[Sorter] = ...,
```
NathanFarmer, line 155: `order_by` does not apply to `PandasDatasource`, as it is a single-batch `Datasource`.

tests/datasource/fluent/test_pandas_datasource.py
```diff
@@ -471,3 +471,24 @@ def test_dataframe_asset(empty_data_context: AbstractDataContext, test_df_pandas
             for asset in empty_data_context.sources.pandas_default.assets.values()
         ]
     )
+
+
+def test_dynamic_pandas_batch_metadata(
```
NathanFarmer, line 476: More detail to be added to this test in my subsequent PR enabling config variables.

great_expectations/datasource/fluent/interfaces.py
Kilo59: But `batch_metadata` gets passed in via kwargs in the implementation, doesn't it? So it won't actually ever be `None` it just won't be provided???

Exception being for `pandas_default` because those are concrete `add_<ASSET>` methods.

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -503,49 +507,85 @@ class PandasDatasource(_PandasDatasource):
...
+            batch_metadata=batch_metadata,
         )
         return self._get_validator(asset=asset)
 
-    def add_clipboard_asset(self, name: str, **kwargs) -> ClipboardAsset:  # type: ignore[valid-type]
+    def add_clipboard_asset(
+        self,
+        name: str,
+        batch_metadata: Optional[BatchMetadata] = None,
+        **kwargs,
```
Kilo59, line 555:
```suggestion
        **kwargs,
```
I think `batch_metadata` could be part of the `**kwargs` and then it would never be `None`.

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -503,49 +507,85 @@ class PandasDatasource(_PandasDatasource):
...
+    ) -> ClipboardAsset:  # type: ignore[valid-type]
         asset = ClipboardAsset(
             name=name,
+            batch_metadata=batch_metadata,
```
Kilo59, line 559:
```suggestion
            batch_metadata=batch_metadata or {},
```
Or this, and that should prevent the mypy error around `None`

great_expectations/datasource/fluent/pandas_dbfs_datasource.pyi
NathanFarmer: No, but it is an immutable default.

great_expectations/datasource/fluent/pandas_datasource.py
NathanFarmer: We would have to concretely define all `add` methods to do this.

Kilo59 (approved):

great_expectations/datasource/fluent/pandas_datasource.py
NathanFarmer: I will make this change in a follow-up PR per our offline conversation.

great_expectations/datasource/fluent/interfaces.py
NathanFarmer: I am changing `batch_metadata` to always be passed in via `kwargs` in a follow-up PR. The comments about `order_by` may not actually be founded, because it does look like we are doing type conversion in the `order_by` validator, so if it gets passed in via `kwargs` it should work.

NathanFarmer: The `order_by` stubs are "covering up" what would typically get surfaced as a typing error because the validator typing information does not get passed along.

great_expectations/datasource/fluent/pandas_datasource.py
billdirks: Do we want to put the data asset meta data in a sub-dictionary with some specially defined key? Like this, it is possible that the batch_request options will overwrite the asset metadata.

PR #7393, opened 2023-03-17
[BUGFIX] Fix bug with case sensitive execution env

anthonyburdi:
- Package names are not case sensitive but we are treating them as they are. This caused issues since SQLAlchemy is listed in our requirements as sqlalchemy. Now we convert every package name to lowercase before comparing to see if it is installed before getting the version number. This will help us find a better lower bound for SQLAlchemy.

### Definition of Done


alexsherstinsky (approved):
LGTM!  Thanks!

PR #7396, opened 2023-03-17
[MAINTENANCE] Test against minimum SQLAlchemy versions

anthonyburdi:
- Add tests against previously supported versions of SQLAlchemy to ensure compatibility with them as we make changes to support later versions.
- This PR adds testing for 1.4.x where x is latest for that minor version. It also currently only tests against postgresql. We can enhance / optimize this test but I would like to have something in place to make sure we retain compatibility with earlier versions as we are working on SQLAlchemy 2.0 compatibility.
- In preparation of v1.3.x compatibility, it also changes SQLAlchemy 2.0 warnings to `DeprecationWarning` (which is a parent class of `sqlalchemy.exc.RemovedIn20Warning`) since `sqlalchemy.exc.RemovedIn20Warning` only exists starting with v1.4.x. I think this may be good practice anyway so that warning ignores don't cause errors if run in an environment without the version of the library with the specific warning.

### Definition of Done

### DIFF
--- a/ci/azure-pipelines-sqlalchemy-compatibility.yml
+++ b/ci/azure-pipelines-sqlalchemy-compatibility.yml
@@ -1,0 +1,194 @@
+# This file is responsible for configuring the `sqlalchemy-compatibility` pipeline (https://dev.azure.com/great-expectations/great_expectations/_build)
+#
+# The pipeline is run under the following conditions:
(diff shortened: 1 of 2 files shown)


alexsherstinsky (approved):
LGTM -- thank you very much for this!

PR #7400, opened 2023-03-17
[MAINTENANCE] SqlAlchemy2 Compatibility - implicit autocommit

Shinnnyshinshin:
- Remove implicit autocommit by using connection and runtime context (with the `with` statement), [that we don't rely on autocommit for DML and DDL](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-to-2-0-step-two-turn-on-removedin20warnings)
- So we are basically using either `engine.begin()` or `connection.begin()` wherever we have an `connection.execute()` function. 
- Ran into trouble with `pandas<1.4` and created `add_dataframe_to_db()` helper method for `to_sql()` method as a result. Docstring should explain the motivation behind its creation. 
- Moved `MockSaEngine` and related classes from `fluent` folder into more general `test_utils.py` file. Tests have been updated to reflect the change. 

### Definition of Done


tests/test_utils.py
```diff
@@ -1040,3 +1042,46 @@ def working_directory(directory: PathStr):
...
+
+class MockSaInspector:
+    def get_columns(self) -> list[dict[str, Any]]:  # type: ignore[empty-body]
+        ...
+
+    def get_schema_names(self) -> list[str]:  # type: ignore[empty-body]
+        ...
+
+    def has_table(self, table_name: str, schema: str) -> bool:  # type: ignore[empty-body]
+        ...
```
Shinnnyshinshin: Mock `SA` classes are moved to `test_utils` and tests have been updated as a result.

tests/datasource/fluent/test_postgres_datasource.py
```diff
@@ -865,13 +857,9 @@ def test_test_connection_failures(
...
     inspect.return_value = MockSaInspector()
-    get_schema_names = mocker.patch(
-        "tests.datasource.fluent.conftest.MockSaInspector.get_schema_names"
-    )
+    get_schema_names = mocker.patch("tests.test_utils.MockSaInspector.get_schema_names")
     get_schema_names.return_value = ["good_schema"]
-    has_table = mocker.patch(
-        "tests.datasource.fluent.conftest.MockSaInspector.has_table"
-    )
+    has_table = mocker.patch("tests.test_utils.MockSaInspector.has_table")
```
Shinnnyshinshin: Had to be updated because the location of the `Mock` classes have changed.

great_expectations/util.py
```diff
@@ -2152,3 +2153,84 @@ def numpy_quantile(
...
+            if_exists=if_exists,  # type: ignore
+            index=True,
+            index_label=index_label,
+            chunksize=chunksize,
+            dtype=dtype,
+            method=method,  # type: ignore
+        )
+
+
+class NotImported:
```
Shinnnyshinshin: This is a strange diff. It was created by a PR 3 weeks ago but showing up as "new".

great_expectations/util.py
```diff
@@ -2152,3 +2153,84 @@ def numpy_quantile(
         )
 
     return quantile
+
+
+def add_dataframe_to_db(
```
Shinnnyshinshin: Helper method that is used to wrap `to_sql()` and suppress warnings.

great_expectations/util.py
alexsherstinsky: That is not a good sign.  I wonder if things have been merged properly or something got mixed in from before.  Maybe we need to start with a new fresh branch off `develop` and merge in specific changes into it from the present work.  This happened to me once before when working with a contributor.  Sorry that this is happening.  Thanks.

great_expectations/util.py
alexsherstinsky: ❤️

anthonyburdi: I like generalizing this method but could we move this somewhere other than a general util.py file? Maybe `database_test_fixture_loader.py` or something similar?

great_expectations/util.py
```diff
@@ -2152,3 +2153,84 @@ def numpy_quantile(
...
+    chunksize: int | None = None,
+    dtype: dict | None = None,
+    method: str | Callable | None = None,
+) -> None:
+    """Write records stored in a DataFrame to a SQL database.
+
+    Wrapper for `to_sql()` method in Pandas. Created as part of the effort to allow GX to be compatible
+    with SqlAlchemy 2, and is used to suppress warnings that arise from implicit auto-commits.
+
+    The need for this function will eventually go away once we migrate to Pandas 1.4.0.
```
anthonyburdi: :bow: This is a super helpful comment, our future selves thank you!

great_expectations/util.py
```diff
@@ -2152,3 +2153,84 @@ def numpy_quantile(
...
+        method (str | Callable | None):
+            Controls the SQL insertion clause used:
+                * None : Uses standard SQL ``INSERT`` clause (one per row).
+                * 'multi': Pass multiple values in a single ``INSERT`` clause.
+                * callable with signature ``(pd_table, conn, keys, data_iter)``.
+    """
+    if isinstance(con, sa.engine.Engine):
+        con = con.connect()
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
```
anthonyburdi: Can we be more specific here with which warning to ignore? If other warnings are popping up we'll want to know.

great_expectations/data_context/store/database_store_backend.py
```diff
@@ -110,9 +110,10 @@ def __init__(  # noqa: C901 - 16
             table = Table(table_name, meta, *cols)
             try:
                 if self._schema_name:
-                    self.engine.execute(
-                        sa.text(f"CREATE SCHEMA IF NOT EXISTS {self._schema_name};")
-                    )
+                    with self.engine.begin() as connection:
```
alexsherstinsky, line 118: @Shinnnyshinshin As mentioned in Slack, maybe this pattern will help in some ways?
```
    @contextmanager
    def _get_managed_scoped_db_session(self):
        session = self._scoped_db_session
        try:
            yield session
            session.commit()
        except Exception as e:
            session.rollback()
            raise
        finally:
            session.close()
            session.remove()
```

or this:
```
def _make_query(self, query: str):
  with self.engine.connect() as connection:
    return connection.execute(query)

def some_function(self):
  query = "some query"
  return _make_query(query)
```

Happy to discuss -- thanks!

great_expectations/execution_engine/sqlalchemy_batch_data.py
```diff
@@ -287,9 +287,16 @@ def _create_temporary_table(  # noqa: C901 - 18
...
+                        stmt = stmt_1
+                    except DatabaseError:
+                        connection.execute(sa.text(stmt_2))
+                        stmt = stmt_2
         else:
-            self._engine.execute(sa.text(stmt))
+            with self._engine.connect() as connection:
+                with connection.begin():
+                    connection.execute(sa.text(stmt))
+        return str(stmt)
```
alexsherstinsky: @Shinnnyshinshin I wonder why we now need to return the query, rather than just execute it as before.  Also, it may be problematic in the sense that the statement should probably be compiled, rather than just turn into a string with `str()`.  Thanks.

great_expectations/util.py
anthonyburdi: Very interesting. This was moved to optional_imports.py recently: https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/optional_imports.py#LL18C16-L18C16. I would delete this code in your PR, maybe it got in via a merge conflict resolution gone awry.

tests/test_utils.py
anthonyburdi: Could we move this instead to a file with a name that is more descriptive? Something like `sqlalchemy_test_doubles.py` or something like that? Will help with finding / organizing things later.

tests/execution_engine/test_sqlalchemy_batch_data.py
```diff
@@ -155,26 +157,35 @@ def test_instantiation_with_unknown_dialect(sqlite_view_engine):
...
         engine.dialect.name = dialect
-        SqlAlchemyBatchData(
+        batch_data = SqlAlchemyBatchData(
             execution_engine=execution_engine,
             query="test_query",
             create_temp_table=True,
             temp_table_schema_name="test_schema",
         )
-        assert "test_schema" in str(engine.execute.call_args[0][0])
+        created_table_name: str = batch_data._create_temporary_table(
```
anthonyburdi: Is this created_table_name or the query used to create the table? If the query, I would suggest renaming this and the other instances to `query_to_create_temp_table` or similar.

pyproject.toml
```diff
@@ -467,10 +467,9 @@ filterwarnings = [
...
-    'ignore: The Engine.execute\(\) method is considered legacy as of the 1.x series of SQLAlchemy and will be removed in 2.0. All statement execution in SQLAlchemy 2.0 is performed by the Connection.execute\(\) method of Connection, or in the ORM by the Session.execute\(\) method of Session.:DeprecationWarning',
+    'ignore: The legacy calling style of select\(\) is deprecated and will be removed in SQLAlchemy 2.0.  Please use the new calling style described at select\(\).:sqlalchemy.exc.RemovedIn20Warning',
+    'ignore: The "whens" argument to case\(\), when referring to a sequence of items, is now passed as a series of positional elements, rather than as a list.:sqlalchemy.exc.RemovedIn20Warning',
+    'ignore: The Engine.execute\(\) method is considered legacy as of the 1.x series of SQLAlchemy and will be removed in 2.0. All statement execution in SQLAlchemy 2.0 is performed by the Connection.execute\(\) method of Connection, or in the ORM by the Session.execute\(\) method of Session.:sqlalchemy.exc.RemovedIn20Warning',
```
anthonyburdi: Can you revert these back to the more general `DeprecationWarning`? I was mistaken when first creating these and if they are not importable (e.g. if sqlalchemy isn't installed or the installed version doesn't have the specific warning) I was getting errors. So I changed these back to use the relevant built in parent warning classes.

anthonyburdi (changes requested):
Looking great! I love using a method to encapsulate the to_df functionality to ignore the warnings. I have some minor requested changes, mostly around code organization but otherwise it looks great! The only other thing is Alex's awesome suggestion to wrap the with blocks in a function with a yield so that we can encapsulate those as well.

tests/test_utils.py
Shinnnyshinshin: Great idea. I'll do this now

great_expectations/util.py
Shinnnyshinshin: mystery solved. Thank you @anthonyburdi

great_expectations/util.py
Shinnnyshinshin: Good idea @anthonyburdi. Doing now

great_expectations/execution_engine/sqlalchemy_batch_data.py
Shinnnyshinshin: @alexsherstinsky this one was a bit tricky. Because the `connection` object is now scoped within the `with` statement, we don't have access to the `stmt` once we break out of the method. This clashed with how we were testing whether the `schema` was being added correctly. 

If there is a better way to do this, I'd love to discuss. This change was kindof bugging me too, to be honest

pyproject.toml
Shinnnyshinshin: Sorry I may not have all the necessary context: is this for all of the `SQLAlchemy` support warnings?

great_expectations/execution_engine/sqlalchemy_batch_data.py
Shinnnyshinshin: Update: so the `statement.compile()` step, (which is very helpful) would only be relevant for the query portion of stmt here, and it is already done prior to calling the `_create_temporary_table()` function. So I think the correct update to this method would be to declare a `str` up top (to ensure the `stmt` is declared as a string) and change this portion to `return stmt`. 

The discussion earlier really helped clarify this for me. Thank you @alexsherstinsky 


https://github.com/great-expectations/great_expectations/blob/cf3053deba218a06f8c6a6041274decfd79463f2/great_expectations/execution_engine/sqlalchemy_batch_data.py#L156

great_expectations/util.py
Shinnnyshinshin: update: we are now only filtering `warnings.filterwarnings(action="ignore", category=RemovedIn20Warning)`

tests/execution_engine/test_sqlalchemy_batch_data.py
Shinnnyshinshin: great catch. Fixed now.

great_expectations/data_context/store/database_store_backend.py
Shinnnyshinshin: update here : The scope of the changes in this PR don't have a large number of connections that need to be managed (it's only 2 max) so the pattern may not be as helpful at the moment. However, I think it will come up very soon as we tackle subsequent SqlAlchemy 2 issues. Will definitely keep your method and branch in mind.

great_expectations/util.py
Shinnnyshinshin: This is now in the `df_to_database_loader.py` file

alexsherstinsky (approved):
LGTM -- thank you!

pyproject.toml
anthonyburdi: Correct (well I think this applies to all warnings but let's keep the scope to SQLAlchemy warnings for this PR). I changed these in https://github.com/great-expectations/great_expectations/pull/7415, I think they may just be reverting in this PR. I'll check the latest.

great_expectations/util.py
```diff
@@ -81,7 +81,9 @@
     import sqlalchemy as sa
     from sqlalchemy import Table
     from sqlalchemy.engine import reflection
+    from sqlalchemy.exc import RemovedIn20Warning
```
anthonyburdi: Can you get rid of this import now that you've moved to `great_expectations/df_to_database_loader.py`?

pyproject.toml
```diff
@@ -468,7 +468,7 @@ filterwarnings = [
     # SQLAlchemy 2.x support warnings. These warnings should be ignored until sqlalchemy 2.x is fully supported.
     # To get SQLAlchemy 2.x supported, remove one of these ignores and then fix the resulting errors.
     'ignore: The legacy calling style of select\(\) is deprecated and will be removed in SQLAlchemy 2.0.  Please use the new calling style described at select\(\).:DeprecationWarning',
-    'ignore: The current statement is being autocommitted using implicit autocommit, which will be removed in SQLAlchemy 2.0. Use the .begin\(\) method of Engine or Connection in order to use an explicit transaction for DML and DDL statements.:DeprecationWarning',
+    'ignore: The "whens" argument to case\(\), when referring to a sequence of items, is now passed as a series of positional elements, rather than as a list.:DeprecationWarning',
```
anthonyburdi: This was removed in https://github.com/great-expectations/great_expectations/pull/7416, looks like it's coming back in.

anthonyburdi (approved):
Thank you for making these changes! Approving but please clean up the two items noted in the review before merge.

pyproject.toml
Shinnnyshinshin: Thank you for catching this 🙇🏼

great_expectations/util.py
Shinnnyshinshin: Thank you for catching this 🙇🏼 too

PR #7401, opened 2023-03-17
[FEATURE] F/great 1463/add updates with datasource obj

billdirks:
This allows one to use the datasource crud methods with either a datasource object or a set of constructor arguments. Previously only a set of constructor arguments was possible. This is useful because one way add assets or otherwise modify a datasource but that won't be saved to config until the user calls an update on the datasource.

The main complication I ran into was the previously allowed the `name` to be passed in as a positional or keyword argument. To maintain this or allow the datasource to be passed in positionally I had to do some extra validation to make sure the arguments were sensible.


billdirks:
Before this PR is merged I need to:
1. Update the stub file
2. Rename the parameter `first` to `name_or_datasource`

great_expectations/datasource/fluent/sources.py
```diff
@@ -333,31 +334,101 @@ def pandas_default(self) -> PandasDatasource:
...
+            current_datasource = None
+        if current_datasource and not isinstance(current_datasource, datasource_type):
             raise ValueError(
                 f"Trying to update datasource {name} but it is not the correct type. "
-                f"Expected {datasource_type} but got {type(old_datasource)}"
+                f"Expected {datasource_type} but got {type(current_datasource)}"
             )
 
+    def _datasource_passed_in_as_only_argument(
+        self, datasource_type, first: Optional[str, Datasource], **kwargs
```
Kilo59:
```suggestion
        self, datasource_type: Type[Datasource], first: Optional[str, Datasource], **kwargs
```

great_expectations/datasource/fluent/sources.py
```diff
@@ -333,31 +334,101 @@ def pandas_default(self) -> PandasDatasource:
...
+            datasource = kwargs["datasource"]
+        if datasource and not isinstance(datasource, datasource_type):
+            raise ValueError(
+                f"Trying to modify datasource {datasource.name} but it is not the correct type. "
+                f"Expected {datasource_type} but got {type(datasource)}"
+            )
+        return datasource
+
+    def _datasource_passed_in(
+        self, datasource_type, first: Optional[str, Datasource], **kwargs
```
Kilo59:
```suggestion
        self, datasource_type: Type[Datasource], first: Optional[str, Datasource], **kwargs
```

great_expectations/datasource/fluent/sources.py
```diff
@@ -333,31 +334,101 @@ def pandas_default(self) -> PandasDatasource:
...
                     f"There is no datasource {name} in the data context."
                 ) from e
-            old_datasource = None
-        if old_datasource and not isinstance(old_datasource, datasource_type):
+            current_datasource = None
+        if current_datasource and not isinstance(current_datasource, datasource_type):
             raise ValueError(
                 f"Trying to update datasource {name} but it is not the correct type. "
-                f"Expected {datasource_type} but got {type(old_datasource)}"
+                f"Expected {datasource_type} but got {type(current_datasource)}"
```
Kilo59:
```suggestion
                f"Expected {datasource_type.__name__} but got {type(current_datasource).__name__}"
```
`non-blocking`
Using the `class` `__name__` attribute might be slightly more readable.

great_expectations/datasource/fluent/sources.py
```diff
@@ -333,31 +334,101 @@ def pandas_default(self) -> PandasDatasource:
...
+                        a datasource is not passed in and no name argument is present.
+        """
+        new_datasource = self._datasource_passed_in_as_only_argument(
+            datasource_type, first, **kwargs
+        )
+        if new_datasource:
+            return new_datasource
+        if (first and isinstance(first, str) and "name" not in "kwargs") or (
+            first is None and "name" in kwargs and isinstance(kwargs["name"], str)
+        ):
```
Kilo59: `non-blocking`

@billdirks This line makes me think this might be a good case for `single-dispatch`.
Dispatching on the value of `first` being either `str` or a `Datasource`.
Or if neither, throwing the `ValueError`.

https://rednafi.github.io/digressions/python/2020/04/05/python-singledispatch.html

Kilo59 (approved):
I'm approving but I do think that the stubs should be updated before merging.
As it could expose typing or issues with auto-completion.

Also `non-blocking` but I think `single-dispatch` might make this implementation simpler and reduce the number of conditionals.
https://github.com/great-expectations/great_expectations/pull/7401/files#r1143332512

great_expectations/datasource/fluent/sources.py
billdirks: You are right, this is a prime candidate for single dispatch. Depending on time, I may or may not update this for this PR but will definitely have this more on my radar.

billdirks:
I've updated the stub files and renamed `first` to `name_or_datasource`.

great_expectations/datasource/fluent/sources.py
billdirks: I'm resolving this conversation since github keeps opening a new comment box here and it's making the UI unusable.

PR #7402, opened 2023-03-18
[MAINTENANCE] : split up map_metric_provider.py

abegong:
- Splits up map_metric_provider.py into smaller files.
-
-


great_expectations/expectations/metrics/map_metric_provider/map_metric_provider.py
```diff
@@ -628,25 +622,4 @@ def _get_evaluation_dependencies(
...
-        return (
-            hasattr(map_metric_provider, "condition_metric_name")
-            and map_metric_provider.condition_metric_name
-            in MapMetricProvider.SQLALCHEMY_SELECTABLE_METRICS
-        ) or (
-            hasattr(map_metric_provider, "function_metric_name")
-            and map_metric_provider.function_metric_name
-            in MapMetricProvider.SQLALCHEMY_SELECTABLE_METRICS
-        )
+        return dependencies
```
abegong: I don't think there's any harm in moving this method outside the class, since it's a static method.

However, this is technically a breaking change. 🤔

great_expectations/expectations/metrics/map_metric_provider/spark_methods.py
```diff
@@ -427,7 +428,7 @@ def _spark_column_pair_map_condition_filtered_row_count(
 
 def _spark_multicolumn_map_condition_values(
     cls,
-    execution_engine: SqlAlchemyExecutionEngine,
+    execution_engine: SparkDFExecutionEngine,
```
abegong: I'm pretty sure this was a mistake in the previous code.

alexsherstinsky: @abegong You definitely caught a typo (a subtle one, because it was in the type hint, causing no crashes).  Thanks.

alexsherstinsky (changes requested):
@abegong I like the separation of partials into their respective separate modules and the separation of different Map Metric Providers (Column, Column-Pair, and Multi-Column) classes into their respective separate modules.  However, instead of separating the implementation methods for the auxiliary metrics that are being registered according to their `ExecutionEngine` backend (`pandas_methods.py`, `sqlalchemy_methods.py`, and `spark_methods.py`), I feel that it would be more theme-appropriate and hence intuitive to separate them based on the kind of Map Metric these methods pertain to.  Thus, some of these methods for all `ExecutionEngine` backends will end up in the `map_metric_provider.py` module, others will end up in the `column_map_metric_provider.py`, still others will end up in the `column_pair_map_metric_provider.py` module, and finally, the rest will end up in the `multicolumn_map_metric_provider.py` (note, no underscore between "multi" and "column" for consistency) module.  Otherwise, I really like this quality of life refactor!  Thank you!

abegong:
@alexsherstinsky , I got all tests passing, then made the changes that you suggested.

* Edited pyproject.toml to include comments with type error counts
* Redistributed auxilliary methods by MetricProvider class, rather than ExecutionEngine
* Renamed `multi_column_map_metric_provider.py` to `multicolumn_map_metric_provider.py`

Assuming tests pass, this should be ready to merge!

tests/expectations/metrics/test_map_metric.py
```diff
@@ -370,12 +370,22 @@ def test_get_map_metric_dependencies():
...
 
 def test_is_sqlalchemy_metric_selectable():
-    assert MapMetricProvider.is_sqlalchemy_metric_selectable(
-        map_metric_provider=CompoundColumnsUnique
+    with pytest.warns(DeprecationWarning) as record:
+        assert MapMetricProvider.is_sqlalchemy_metric_selectable(
+            map_metric_provider=CompoundColumnsUnique
+        )
+
+    assert "MapMetricProvider.is_sqlalchemy_metric_selectable is deprecated." in str(
```
alexsherstinsky, line 378: @abegong Non-blocking: I need to learn more about what led to the `MapMetricProvider.is_sqlalchemy_metric_selectable()` method becoming deprecated.

Adding @anthonyburdi and @Shinnnyshinshin who might be able to offer more detail regarding this deprecation.  

Thank you.

great_expectations/expectations/metrics/map_metric_provider/map_metric_provider.py
```diff
@@ -0,0 +1,686 @@
...
+                dependencies["unexpected_condition"] = MetricConfiguration(
+                    metric_name=f"{metric_name[:-len(metric_suffix)]}.{MetricPartialFunctionTypeSuffixes.CONDITION.value}",
+                    metric_domain_kwargs=metric.metric_domain_kwargs,
+                    metric_value_kwargs=base_metric_value_kwargs,
+                )
+
+        return dependencies
+
+    @staticmethod
+    def is_sqlalchemy_metric_selectable(
```
alexsherstinsky, line 676: @abegong @Shinnnyshinshin @anthonyburdi -- I wonder if our ability to deprecate the method `is_sqlalchemy_metric_selectable()` is thanks to the effort aimed at supporting `SQLAlchemy-2.0`; regardless, if we still need this method, it should be converted to be "private" (this is an entirely internal utility).  Please let me know (happy to discuss).  Thank you.

alexsherstinsky (approved):
LGTM -- thank you very much, @abegong -- the outcome of this quality-of-life refactor nudges the "Map Metric" code further to the right on the spectrum toward "a thing of beauty"!  I only had essentially one non-blocking question.

great_expectations/expectations/metrics/map_metric_provider/map_metric_provider.py
abegong: @alexsherstinsky , I think you're misreading this. I'm deprecating `is_sqlalchemy_metric_selectable` as a pulbic method within the `MapMetricProvider` class, because it needs to be accessible from methods that don't have access to that class.

This wasn't a problem (other than being a weird code pattern) when all the classes and methods were stored in a single file. But when I started separating the code out into smaller files, it created a circular import.

`_is_sqlalchemy_metric_selectable` continues to exist as a utility function in a separate script. Eventually, I think the right answer will be to bring it back into the `MetricProvider` class---but that will require a refactor that's out fo scope for this PR.

PR #7403, opened 2023-03-19
[CONTRIB] Limit results for two expectations

itaise:
Hi,
For two tests which we contributed, we return the full results in the observed value - when the results are too large it failes our slack message (which contains a reference to the results) and persist results which are really large in our DB. 
I added a limit of 10 values to it, so the results will show an example of the unexpected values. 

Can you please review?
Thanks a lot!


kenwade4 (approved):
Looks good, thanks for these 2 new PRs @itaise!

PR #7414, opened 2023-03-21
[BUGFIX] Fluent `PandasFilesytemDatasources` data_connector fixes

Kilo59:
Proof of concept for "fixing" the DataConnector abstraction

- [x] implement `_build_data_connector`
- [x] build data_connectors when loading from config
   - [x] test for use of data_connectors after loading from config 
- [x] add `connect_options ` field
- [x] pipe through `connect_options ` field to where it's needed
- [x] remove manual `add_<ASSET_TYPE>_asset()` methods
- [x] ~update stubs~ done in #7428 

## Notes
These changes are initially only for `PandasFilesytemDatasource`s. Followup PR's re-implement this pattern for the remaining datasources which use `DataConnectors. This will mostly entail deleting code 😄 .


great_expectations/datasource/fluent/file_path_data_asset.py
```diff
@@ -59,7 +61,12 @@ class _FilePathDataAsset(DataAsset):
     }
 
     # General file-path DataAsset pertaining attributes.
-    batching_regex: Pattern
+    batching_regex: Pattern = MATCH_ALL_PATTERN
```
Kilo59, line 64: Need this default here because before this default was being set in the manually defined `add_<ASSET_TYPE>_asset()` methods

great_expectations/datasource/fluent/file_path_data_asset.py
```diff
@@ -59,7 +61,12 @@ class _FilePathDataAsset(DataAsset):
...
+    # TODO: better name
+    dc_options: Mapping = pydantic.Field(
```
alexsherstinsky: @Kilo59 If we are going to go with this (not sure yet, but supposing that we will), then I would propose fully spelling out "data_connector" rather than "dc".  According to modern best practices published by a few large software developer houses, the recommendation is to spell literals out, rather than use acronyms (where confusion may arise).  In other words, we do not need to spell out HTTP or XML 😄 -- thanks for your consideration.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -350,6 +351,8 @@ class Datasource(
 
     # class attrs
     asset_types: ClassVar[Sequence[Type[DataAsset]]] = []
+    # Not all Datasources require a DataConnector
+    data_connector_type: ClassVar[Optional[Type[DataConnector]]] = None
```
alexsherstinsky, line 358: @Kilo59 If this addition is toward including `DataConnector` in Fluent configuration, then we are not supposed to move in that direction (because in Fluent, `DataConnector` is an implementational utility, not a core concept).  If this is for some other purpose, then please let me know and I will study more carefully.  Thanks.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -508,7 +517,7 @@ def parse_batching_regex_string(
     ) -> re.Pattern:
         pattern: re.Pattern
         if not batching_regex:
-            pattern = re.compile(".*")
+            pattern = MATCH_ALL_PATTERN
```
alexsherstinsky: ❤️

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -538,13 +547,13 @@ def test_connection(self, test_assets: bool = True) -> None:
...
 
-    def _build_data_connector(self, data_asset_name: str, **kwargs) -> None:
+    def _build_data_connector(self, data_asset: _DataAssetT, **kwargs) -> None:
         """Any Datasource subclass that utilizes DataConnector should overwrite this method.
 
         Specific implementations instantiate appropriate DataConnector class and set "self._data_connector" to it.
 
         Args:
-            data_asset_name: The name of the DataAsset using this DataConnector instance
+            data_asset: DataAsset using this DataConnector instance
```
alexsherstinsky, line 559: @Kilo59 The reason we had `data_asset_name` before is that the `DataConnector` only needed the name (i.e., a string), rather than reference to the entire `DataAsset` object (to which `_DataAssetT` appears to be bound).  Thanks.

great_expectations/datasource/fluent/sources.py
```diff
@@ -269,6 +269,7 @@ def _bind_asset_factory_method_if_not_present(
             def _add_asset_factory(
                 self: Datasource, name: str, **kwargs
             ) -> pydantic.BaseModel:
+                # TODO: identify dc kwargs and pass them to `add_asset`
```
alexsherstinsky: @Kilo59 Not sure if this is relevant here, but just wanted to note that differently-purposed `DataConnector` subclasses accept their corresponding (most likely different) "kwargs".  Thanks.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -538,13 +547,13 @@ def test_connection(self, test_assets: bool = True) -> None:
             """One needs to implement "test_connection" on a Datasource subclass."""
         )
 
-    def _build_data_connector(self, data_asset_name: str, **kwargs) -> None:
+    def _build_data_connector(self, data_asset: _DataAssetT, **kwargs) -> None:
```
alexsherstinsky, line 553: @Kilo59 I checked, and am confirming that `Datasource. _build_data_connector()` interface definition should be deleted (its presence is a typographical error -- accidental left-over from an earlier iteration).  Even if a reason for this method may arise as part of a future architectural enhancement, it should be deleted for now.  Thank you.

great_expectations/datasource/fluent/interfaces.py
Kilo59: No, it's about ensuring we don't need `DataConnector` in the configuration.
At least directly. We do need to be able to provide options that the data connector can ingest in config (even if we don't call it a data_connector). But before this change you could not provide any asset level data_connector params in config (`glob_directive`, `prefix`, `delimiter`, etc.)

great_expectations/datasource/fluent/pandas_filesystem_datasource.py
```diff
@@ -57,199 +57,21 @@ def test_connection(self, test_assets: bool = True) -> None:
             for asset in self.assets.values():
                 asset.test_connection()
 
-    @public_api
```
alexsherstinsky, line 60: @Kilo59 Please let me know whether or not the following assessment is correct:  If we delete these `add_<filetype>_asset()` methods, then `PandasFilesystemDatasource` will inherit them from `_PandasFilePathDatasource`, which will inherit them from `_PandasDatasource`.  In the latter, these methods use all-`**kwargs`, while in the present class, `PandasFilesystemDatasource`, these methods specify at least some of the `DataAsset` creation arguments.  My understanding is that in Fluent, reducing the reliance on all-`**kwargs` is preferred.  Happy to discuss.  Thank you.

alexsherstinsky (changes requested):
@Kilo59 Thank you for this exploratory work.  I identified one action item and raised a number of questions.  Happy to discuss.  Thanks!

great_expectations/datasource/fluent/file_path_data_asset.py
Kilo59: ~I'm happy to make this name this whatever.~
On second thought though my inclination is for something like `connect_options`.
DataConnectors are meant to know be an internal implementation detail so including it in the name runs counter to that effort.

Edit:
I made it `connect_options` after this idea was seconded by @billdirks

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -415,6 +418,28 @@ def _load_asset_subtype(
         logger.debug(f"{asset_type_name} - {repr(asset_of_intended_type)}")
         return asset_of_intended_type
 
+    # TODO: do this inside of the add_asset_factory methods? If so remove this
+    # @pydantic.validator("assets", each_item=True)
+    # @classmethod
+    # def _validate_data_connector_options(
```
Kilo59: TODO: remove this before merge

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -458,7 +483,9 @@ def get_asset(self, asset_name: str) -> _DataAssetT:
                 f"'{asset_name}' not found. Available assets are {list(self.assets.keys())}"
             ) from exc
 
-    def add_asset(self, asset: _DataAssetT) -> _DataAssetT:
+    def add_asset(
+        self, asset: _DataAssetT, dc_options: dict | None = None
```
billdirks: I like your suggestion in another thread to rename `dc_options` to `connect_options` even if we always will pass this to `_build_data_connector`.

great_expectations/datasource/fluent/interfaces.py
billdirks: Can we move `MATCH_ALL_PATTERN` to the default argument and remove this if branch? That is `def parse_batching_regex_string(batching_regex: Union[re.Pattern, str] = MATCH_ALL_PATTERN)`.

great_expectations/datasource/fluent/pandas_filesystem_datasource.py
```diff
@@ -57,199 +57,20 @@ def test_connection(self, test_assets: bool = True) -> None:
...
-        )
-        return self.add_asset(asset=asset)
-
-    # attr-defined issue
-    # https://github.com/python/mypy/issues/12472
-    add_csv_asset.__signature__ = _merge_signatures(add_csv_asset, CSVAsset, exclude={"type"})  # type: ignore[attr-defined]
-    add_excel_asset.__signature__ = _merge_signatures(add_excel_asset, ExcelAsset, exclude={"type"})  # type: ignore[attr-defined]
-    add_json_asset.__signature__ = _merge_signatures(add_json_asset, JSONAsset, exclude={"type"})  # type: ignore[attr-defined]
-    add_parquet_asset.__signature__ = _merge_signatures(add_parquet_asset, ParquetAsset, exclude={"type"})  # type: ignore[attr-defined]
+        return None
```
billdirks: You can remove this explicit `return None` if you want.

great_expectations/datasource/fluent/pandas_filesystem_datasource.pyi
```diff
@@ -52,6 +52,7 @@ class PandasFilesystemDatasource(_PandasFilePathDatasource):
     def add_csv_asset(
         self,
         name: str,
+        *,
```
billdirks: Thank you for adding the `*`!

great_expectations/datasource/fluent/sources.py
```diff
@@ -271,8 +271,27 @@ def _bind_asset_factory_method_if_not_present(
             def _add_asset_factory(
                 self: Datasource, name: str, **kwargs
             ) -> pydantic.BaseModel:
+
+                # if the data_connector uses a data_connector we need to identify the
+                # asset level attributes needed by the data_connector
+                # push them to `dc_options` field
+                if self.data_connector_type:
+                    dc_options = {
```
billdirks: You could combine these loops:
```
dc_options = {}
for k, v in kwarg.items():
  if k in self.data_connector_type.asset_level_option_keys and v:
    dc_options[k] = kwargs.pop(k)
if dc_options:
  self.data_connector_type.asset_options_type(**dc_options)
```

billdirks (approved):
LGTM! I left a few non-blocking comments and you have some comments about removing commented out code.

great_expectations/datasource/fluent/interfaces.py
Kilo59: Yes we can! 😄 
But some of the other `Datasource` types rely on it. Once they've all been updated to follow the pattern in this PR we can remove it.

great_expectations/datasource/fluent/sources.py
Kilo59: I did this initially but I can't `pop()` while iterating over kwargs.
I'd have to make a copy, which I could do.

great_expectations/datasource/fluent/interfaces.py
Kilo59: This wasn't being called anywhere in the code.
The way I'm calling it now we do need the full `data_asset`, not just the `name`

great_expectations/datasource/fluent/interfaces.py
Kilo59: I'm using it now.

great_expectations/datasource/fluent/pandas_filesystem_datasource.py
Kilo59: No. Deleting these methods will mean we start using the dynamically generated `add_<ASSET_TYPE>_asset()` that are created with the `_add_asset_factory()` closures in `sources.py`.

These `_add_asset_factory()` closures now make use of the `build_data_connector()` methods for any Datasource that provides a `data_connector_type` class level attribute.
This means we will also have working data_connectors for all the other asset types not manually defined (`excel`, `hdfs`, `feather` etc.)

tests/datasource/fluent/test_metadatasource.py
```diff
@@ -528,7 +528,6 @@ def test_update_datasource_with_datasource_object(
                     "csv_asset": {
                         "batching_regex": "(?P<file_name>.*).csv",
                         "name": "csv_asset",
-                        "order_by": [],
```
Kilo59, line 531: This was present because this value was being set in the manually defined `add_csv_asset()` method.
Fields that are not set should be excluded from serialization, so this is the desired behavior.

great_expectations/datasource/fluent/interfaces.py
Kilo59: Done

great_expectations/datasource/fluent/sources.py
Kilo59: I've done this in a roundabout way with the class level attributes `DataConnector.asset_level_option_keys` and `DataConnector.asset_options_type`

tests/datasource/fluent/test_viral_snippets.py
```diff
@@ -125,8 +98,28 @@ def test_serialize_fluent_config(fluent_file_context: FileDataContext):
             assert asset_name in dumped_yaml
 
 
+def test_data_connectors_are_built_on_config_load(fluent_file_context: FileDataContext):
+    dc_datasource_count: dict[str, list[str]] = defaultdict(list)
```
Kilo59, line 102:
```suggestion
    dc_datasources: dict[str, list[str]] = defaultdict(list)
```

NathanFarmer (approved):
✅ following our synchronous review

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -5424,10 +5426,19 @@ def _load_fluent_config(self, config_provider: _ConfigurationProvider) -> GxConf
         )
         return GxConfig(fluent_datasources={})
 
-    def _attach_fluent_config_datasources(self, config: GxConfig):
+    def _attach_fluent_config_datasources_and_build_data_connectors(
```
alexsherstinsky, line 5429: ✔️

alexsherstinsky: This looks a little too much bundled, but I do not have a better suggestion for now.

Kilo59: I agree, it feels bad. There may be opportunities to unbundle this as part of the work I'm doing to get this working in a `CloudDataContext`.

great_expectations/datasource/fluent/data_asset/data_connector/data_connector.py
```diff
@@ -40,6 +40,10 @@ class DataConnector(ABC):
         data_asset_name: The name of the DataAsset using this DataConnector instance
     """
 
+    # needed to select the asset level kwargs needed to build the DataConnector
+    asset_level_option_keys: ClassVar[tuple[str, ...]] = ()
+    asset_options_type: ClassVar[Type] = dict
```
alexsherstinsky, line 45: @Kilo59 If possible, could you please provide a rich docstring for `asset_options_type` and `asset_level_option_keys` here in the top-level `DataConnector` module?  The reason for this request is that whenever a reader sees a class variable, which does not appear to be used anywhere in the present module, the experience of interpreting the code slows down appreciably.  We have an analogous situation in the Metrics component (I have made some effort at documenting the class variables used, but a lot more is in order).  On a separate thought, is the class variable the best way of achieving our goal here?  Could this be an indicator that a better / cleaner design is lurking somewhere and we just have not yet nailed it?  Thank you.

great_expectations/datasource/fluent/data_asset/data_connector/filesystem_data_connector.py
```diff
@@ -33,6 +39,9 @@ class FilesystemDataConnector(FilePathDataConnector):
         file_path_template_map_fn: Format function mapping path to fully-qualified resource on filesystem (optional)
     """
 
+    asset_level_option_keys: ClassVar[tuple[str, ...]] = ("glob_directive",)
+    asset_options_type: ClassVar[Type[FilesystemOptions]] = FilesystemOptions
```
alexsherstinsky, line 43: @Kilo59 Yes, at least documenting for how `asset_level_option_keys` and `asset_level_option_keys` will be utilized would be nice (here, rather than have to search for it elsewhere).  Thanks.

great_expectations/datasource/fluent/file_path_data_asset.py
```diff
@@ -11,6 +11,7 @@
     ClassVar,
     Dict,
     List,
+    Mapping,
```
alexsherstinsky, line 14: @Kilo59 I see `Mapping` defined here, but not used in the present module; am I missing something here?  Thanks.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -354,6 +354,8 @@ class Datasource(
 
     # class attrs
     asset_types: ClassVar[Sequence[Type[DataAsset]]] = []
+    # Not all Datasources require a DataConnector
```
alexsherstinsky, line 357: @Kilo59 Side comment -- I suspect that down the road all `Datasource` would do well by utilizing some kind of a `DataConnector` in order to achieve simpler, cleaner, and consistent method interfaces.  Thanks.

great_expectations/datasource/fluent/interfaces.py
alexsherstinsky: For my edification: Sounds like if a property is `None`, it will not end up in Configuration?  Or if this is incorrect, how does one specify a property "to be part of configuration" and "to not be part of configuration"?  Thanks.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -462,7 +464,9 @@ def get_asset(self, asset_name: str) -> _DataAssetT:
                 f"'{asset_name}' not found. Available assets are {list(self.assets.keys())}"
             ) from exc
 
-    def add_asset(self, asset: _DataAssetT) -> _DataAssetT:
+    def add_asset(
+        self, asset: _DataAssetT, connect_options: dict | None = None
```
alexsherstinsky, line 468: @Kilo59 Unless it is available elsewhere, some documentation and/or examples about `connect_options` would be great.  Thanks.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -472,6 +476,10 @@ def add_asset(self, asset: _DataAssetT) -> _DataAssetT:
         # See the comment in DataAsset for more information.
         asset._datasource = self
 
+        if not connect_options:
+            connect_options = {}
+        self._build_data_connector(asset, **connect_options)
```
alexsherstinsky, line 481: @Kilo59 I am not seeing how it works whereby asset is of type `_DataAssetT`, which is bound to the `DataAsset` type, whereas the `DataConnector` constructor accepts the `data_asset_name` (whose type is a string).  Please let me know where/how this is reconciled.  Thank you.

great_expectations/datasource/fluent/sources.py
```diff
@@ -271,8 +271,27 @@ def _bind_asset_factory_method_if_not_present(
             def _add_asset_factory(
                 self: Datasource, name: str, **kwargs
             ) -> pydantic.BaseModel:
+
+                # if the data_connector uses a data_connector we need to identify the
```
alexsherstinsky: @Kilo59 This sentence reads "if the data_connector uses a data_connector" -- does it need re-phrasing?  If you are documenting the `connect_options` here, that is fine; I just wanted to make sure I understood it.  Thanks.

tests/datasource/fluent/great_expectations.yml
```diff
@@ -27,6 +27,31 @@ fluent_datasources:
         order_by:
           - year
           - -month
+  sqlite_taxi:
```
alexsherstinsky, line 30: ❤️ Nice new use cases in the configuration!  Thanks!

alexsherstinsky (changes requested):
@Kilo59 Thank you for this thrust for a massive improvement.  I spotted an oddity, requested docstrings, and asked some questions for my own understanding.  Happy to discuss!  Thanks!

great_expectations/datasource/fluent/sources.py
Kilo59: Yes thank you. It should be `# if the Datasource uses a data_connector we need to identify the...`

great_expectations/datasource/fluent/file_path_data_asset.py
Kilo59: It is used on line `65`

great_expectations/datasource/fluent/interfaces.py
Kilo59: We check this class attribute to know whether or not the Datasource and data assets need to utilize a data_connector. If it does, we need to build that specific `DataConnector` with this attribute.

great_expectations/datasource/fluent/interfaces.py
Kilo59: See the concrete implementation of `_build_data_connector()` in `pandas_filesystem_datasource.py`

great_expectations/datasource/fluent/sources.py
Kilo59:
```suggestion
                # if the Datasource uses a data_connector we need to identify the
```

great_expectations/data_context/data_context/abstract_data_context.py
alexsherstinsky: 🙇

great_expectations/datasource/fluent/pandas_filesystem_datasource.py
```diff
@@ -57,199 +57,29 @@ def test_connection(self, test_assets: bool = True) -> None:
...
-            FilesystemDataConnector.build_test_connection_error_message(
-                data_asset_name=name,
-                batching_regex=batching_regex_pattern,
-                glob_directive=glob_directive,
-                base_directory=self.base_directory,
+    def _build_data_connector(
+        self, data_asset: _FilePathDataAsset, glob_directive: str = "**/*", **kwargs
+    ) -> None:
+        """Builds and attaches the `FilesystemDataConnector` to the asset."""
+        if kwargs:
```
alexsherstinsky, line 64: ❤️ I like this technique!

great_expectations/datasource/fluent/interfaces.py
alexsherstinsky: @Kilo59 Got it -- thanks; looks great (did not see before among all the red of the deleted code).

alexsherstinsky (approved):
LGTM -- awesome!  Thanks!  P.S.: Really looking forward to making time to disentangle the `DataConnector` now covering both connecting to data and partitioning the assets).

great_expectations/datasource/fluent/data_asset/data_connector/data_connector.py
Kilo59: Yes, I agree this needs to be well explained.

Unfortunately, we do have to do something like this (even if not exactly this) if we want to preserve the simultaneous goals of...
1. No exponential explosion of DataAsset classes.
2. Auto-complete for specific DataAsset Dataconnector connect options.

PR #7418, opened 2023-03-21
[MAINTENANCE] Disable incorrectly generated PyCharm warnings in Fluent PandasDatasource modules

alexsherstinsky:
### Scope
`PyCharm` generates frivolous warnings, due to its inability to infer method signatures built using meta-programming automatically.  We disable these warnings in order to avoid confusion between warnings that require attention and those that do not.

- JIRA: GREAT-1709/GREAT-1766
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!

### DIFF
--- a/great_expectations/datasource/fluent/dynamic_pandas.py
+++ b/great_expectations/datasource/fluent/dynamic_pandas.py
@@ -309,8 +309,7 @@ def _to_pydantic_fields(
     `pydantic.create_model()` as field arguments
     """
     fields_dict: Dict[str, _FieldSpec] = {}
-    # noinspection PyTypeChecker
-    all_parameters: Iterator[tuple[str, inspect.Parameter]] = iter(
+    all_parameters: Iterator[tuple[str, inspect.Parameter]] = iter(  # type: ignore[valid-type]
         sig_tuple.signature.parameters.items()
     )
     if skip_first_param:
@@ -322,8 +321,7 @@ def _to_pydantic_fields(
         if substitution:
             fields_dict.update(substitution)
         else:
-            # noinspection PyUnresolvedReferences
-            no_annotation: bool = param.annotation is inspect._empty
+            no_annotation: bool = param.annotation is inspect._empty  # noqa: F823
             if no_annotation:
                 logger.debug(f"`{param_name}` has no type annotation")
                 FIELD_SKIPPED_NO_ANNOTATION.add(param_name)  # TODO: not skipped
@@ -370,8 +368,7 @@ def _create_pandas_asset_model(
     def _get_reader_method(self) -> str:
         return f"read_{self.type}"
 
-    # noinspection PyUnusedLocal
-    def _get_reader_options_include(self) -> set[str] | None:
+    def _get_reader_options_include(self) -> set[str] | None:  # noqa: F841
         return None
 
     setattr(model, "_get_reader_method", _get_reader_method)
--- a/great_expectations/datasource/fluent/dynamic_pandas.py
+++ b/great_expectations/datasource/fluent/dynamic_pandas.py
@@ -309,6 +309,7 @@ def _to_pydantic_fields(
     `pydantic.create_model()` as field arguments
     """
     fields_dict: Dict[str, _FieldSpec] = {}
+    # noinspection PyTypeChecker
     all_parameters: Iterator[tuple[str, inspect.Parameter]] = iter(
         sig_tuple.signature.parameters.items()
     )
@@ -321,6 +322,7 @@ def _to_pydantic_fields(
         if substitution:
             fields_dict.update(substitution)
         else:
+            # noinspection PyUnresolvedReferences
             no_annotation: bool = param.annotation is inspect._empty
             if no_annotation:
                 logger.debug(f"`{param_name}` has no type annotation")
@@ -368,6 +370,7 @@ def _create_pandas_asset_model(
     def _get_reader_method(self) -> str:
         return f"read_{self.type}"
 
+    # noinspection PyUnusedLocal
     def _get_reader_options_include(self) -> set[str] | None:
         return None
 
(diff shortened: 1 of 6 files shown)


great_expectations/datasource/fluent/dynamic_pandas.py
```diff
@@ -321,7 +323,7 @@ def _to_pydantic_fields(
         if substitution:
             fields_dict.update(substitution)
         else:
-            no_annotation: bool = param.annotation is inspect._empty
+            no_annotation: bool = param.annotation is inspect._empty  # noqa: F823
```
Kilo59: @alexsherstinsky 
I don't like these `noqa` comments for something that's only an issue in your IDE.
IMO we should update your IDE so that you don't get these false positives.

We do check for this violation with `ruff` - `undefined-local`
https://beta.ruff.rs/docs/rules/#pyflakes-f

Kilo59 (commented):
Personally, I don't think we should be doing this at all.
We should solve this via IDE configuration.

But If you can get this to pass CI I won't block it.

great_expectations/datasource/fluent/dynamic_pandas.py
alexsherstinsky: @Kilo59 I will try to resolve this one through the IDE (because clearly if the property is unavailable, tests should fail, and the IDE should not be impeding us.  However, the other ones, where all IDEs will not be able to handle the complexity of meta-programming, I would like to stay.  I will update you.  Thanks.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -383,7 +384,7 @@ class Datasource(
     assets: MutableMapping[str, _DataAssetT] = {}
 
     # private attrs
-    _data_context = pydantic.PrivateAttr()
+    _data_context: GXDataContext = pydantic.PrivateAttr()
```
NathanFarmer, line 390: Why rename this type?

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -500,7 +499,7 @@ def test_connection(self, test_assets: bool = True) -> None:
         ...
 
     def _get_validator(self, asset: _PandasDataAsset) -> Validator:
-        batch_request: BatchRequest = asset.build_batch_request()
+        batch_request: BatchRequest | None = asset.build_batch_request()
```
NathanFarmer: I don't believe this is correct.

NathanFarmer (changes requested):

great_expectations/datasource/fluent/interfaces.py
alexsherstinsky: @NathanFarmer The original import came from `interface.py` where line 387 contains `_data_context: GXDataContext = pydantic.PrivateAttr()`.  It is then used in #7425 as well.  Thank you.

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @NathanFarmer Since in the next line, `return self._data_context.get_validator(batch_request=batch_request)` the `get_validator()` method accepts `batch_request: Optional[BatchRequestBase] = None,` as an argument, the type definition has to include `None`.  Thanks.

NathanFarmer: @alexsherstinsky that doesn't make sense to me. Does this raise a `mypy` error? It didn't before.

NathanFarmer (changes requested):

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @NathanFarmer Regarding `https://github.com/great-expectations/great_expectations/pull/7418#discussion_r1145191420`, I can show you what `PyCharm` is flagging -- as soon as we have a free minute.  It looks right to me, because indeed the `get_validator()` method treats `BatchRequest` as `Optional` (defaulted to `None`).  Thanks.

Kilo59: I think `mypy` may be only running the type-checking on the stubfile version of this file.
Which might explain why the `type-ignore`'s Alex added are not getting flagged by the `--warn-unsused-ignores`
https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-unused-ignores

I think we should hold off on merging this PR until we understand what's going on there.

Kilo59 (changes requested):
I think we need to figure out why we aren't getting mypy warning in the CI here.
We should be getting at least some `unused-ignores` here.

I think what's happening is that `mypy` is checking the stubfile for these modules and not the module itself.
Which would explain why we aren't seeing `unused-ignores` or other warnings.

alexsherstinsky:
> I think we need to figure out why we aren't getting mypy warning in the CI here. We should be getting at least some `unused-ignores` here.
> 
> I think what's happening is that `mypy` is checking the stubfile for these modules and not the module itself. Which would explain why we aren't seeing `unused-ignores` or other warnings.

@Kilo59 I corrected this by collaborating with @NathanFarmer and adding a Fluent `BatchRequest` to `AbstractDataContext.get_validator()`.  This cleared away the `PyCharm` warning in the most legitimate way to date.  Thanks.

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @Kilo59 Based on the latest changes, what do you suggest we do next?  Happy to meet together with @NathanFarmer when all are free.  Thanks.

alexsherstinsky: @Kilo59 The stub file "great_expectations/datasource/fluent/pandas_datasource.pyi" declares `_get_validator()` which does not go into the internals of how it is implemented (hence no visibility to any errors that were showing for me in the actual "great_expectations/datasource/fluent/pandas_datasource.py" module.  Thanks.

alexsherstinsky:
This pull request is being closed, because the relevant work was completed elsewhere.

PR #7428, opened 2023-03-22
[MAINTENANCE] PandasFilesystemDatasource stubs and Schema corrections

Kilo59:
- fix [pandas_filesystem_datasource.pyi](https://github.com/great-expectations/great_expectations/pull/7428/files#diff-2736a0c1973043209b36d496980cc43570b465ea95d7b2eff44887b5570dc055), `pandas_dbfs` stubs
  - all methods should have a `glob_directive` keyword argument
  - add `*` keyword-only symbol to reflect that `name` is the only value that be passed positionally. 
     - https://peps.python.org/pep-3102/ 
- use a `MATCH_ALL_PATTERN` constant for `".*"` regex
- Formally give `batching_regex` a default value so that our schemas reflect the current behavior
  - re-generate schemas via `invoke schema --sync`

### Notes

Breaking this PR up from our DataSource abstraction update to lessen the scope.
 - #7414 
 - Note these changes will still be correct even if the above PR is not accepted.

The stubs and schema updates could be 2 separate PRs as they aren't directly connected.

### DIFF
--- a/great_expectations/datasource/fluent/pandas_dbfs_datasource.pyi
+++ b/great_expectations/datasource/fluent/pandas_dbfs_datasource.pyi
@@ -49,6 +49,7 @@ class PandasDBFSDatasource(PandasFilesystemDatasource):
     def add_csv_asset(
         self,
         name: str,
+        *,
         batching_regex: Optional[Union[str, re.Pattern]] = ...,
         glob_directive: str = ...,
         order_by: Optional[SortersDefinition] = ...,
@@ -106,6 +107,7 @@ class PandasDBFSDatasource(PandasFilesystemDatasource):
     def add_excel_asset(
         self,
         name: str,
+        *,
         batching_regex: Optional[Union[str, re.Pattern]] = ...,
         glob_directive: str = ...,
         order_by: Optional[SortersDefinition] = ...,
@@ -136,8 +138,10 @@ class PandasDBFSDatasource(PandasFilesystemDatasource):
     def add_feather_asset(
         self,
         name: str,
+        *,
         order_by: typing.List[Sorter] = ...,
         batching_regex: typing.Pattern = ...,
+        glob_directive: str = ...,
         columns: Union[Sequence[Hashable], None] = ...,
         use_threads: bool = ...,
         storage_options: StorageOptions = ...,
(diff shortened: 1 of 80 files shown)


great_expectations/datasource/fluent/schemas/PandasAzureBlobStorageDatasource.json
```diff
@@ -86,14 +86,14 @@
...
+                    "default": ".*",
                     "type": "string",
                     "format": "regex"
                 }
             },
             "required": [
                 "name",
-                "type",
-                "batching_regex"
+                "type"
```
Kilo59, line 96: `batching_regex` wasn't really a `required` field (according to jsonschema) before because we always set it in the manually defined `add_<ASSET_TYPE>_method()`.

By using the pydantic field default, the jsonschema now correctly expresses these constraints and default value.

alexsherstinsky (approved):
LGTM -- thanks!

billdirks (approved):
LGTM!

great_expectations/datasource/fluent/pandas_dbfs_datasource.pyi
```diff
@@ -49,6 +49,7 @@ class PandasDBFSDatasource(PandasFilesystemDatasource):
     def add_csv_asset(
         self,
         name: str,
+        *,
```
NathanFarmer, line 52: I have struggled to find documentation on how this works. Does this denote that there are no more positional args after `name`?

great_expectations/datasource/fluent/schemas/PandasAzureBlobStorageDatasource.json
NathanFarmer: 🙇🏻

NathanFarmer (approved):

great_expectations/datasource/fluent/pandas_dbfs_datasource.pyi
Kilo59: Yes exactly.
PEP 3102
https://peps.python.org/pep-3102/

It works similarly to `*args` except it doesn't actually capture any items, anything after the `*` can never be passed positionally.

So these 2 functions are equivalent (in terms of HOW parameters can be passed to them).

```python
def foo(a, b, *, c = None):
   ...

def bar(a, b, **kwargs):
   ...


# this works
foo("a", "b", c="c")
bar("a", "b", c="c")

# but this would fail in the same way for both
foo("a", "b", "c")
bar("a", "b", "c")
```

image

PR #7438, opened 2023-03-23
[MAINTENANCE] only provide `py.typed` files for fully typed sub-packages

### LINKED ISSUE #7436: type-checking problems Upgrading from 0.15.50 to 0.16.1.
**Description**
Hey all, I am trying to migrate GX from 0.15.50 to 0.16.1 but getting errors on linting my code with flake8 (4.0.1) and mypy (0.961).

- E.G.: BaseDataContext is not valid as a type -> change my logic to use great_expectation.get_context and use AbstractDataContext and that was solved.
- assistant.py:57: error: "DataAssistantRunner" has no attribute "run" -> This one I am not being able to solve and not seeing any change in documentation referring the need to change the way we call the assistant. I am using: **context.assistants.onboarding.run(...)** which is what is mentioned in the documentation
- (Error on **map_metric** defining custom expectation) -> error: Incompatible types in assignment (expression has type "str", base class "MulticolumnMapExpectation" defined the type as "None")
- (Error on **success_keys** defining custom expectation) -> error: Incompatible types in assignment (expression has type "Tuple[str, str, str]", base class "MulticolumnMapExpectation" defined the type as "Tuple[str]")

Any help or idea how to solve this? Any place where I can see the expected changes while migrating?
For context, I am using Spark, reading from a runtime DataFrame.

**Expected behavior**
No complains from the lint.

 - Operating System: MacOS
 - Great Expectations Version: 0.16.1

Kilo59:
- Fixes https://github.com/great-expectations/great_expectations/issues/7436

- remove top-level `py.typed` from the top-level `great_expectations` package
- only provide `py.typed` file markers for sub-packages that are fully typed
   - `datasource/fluent` - `datasource/` has one untyped module that prevents us from putting it at the top-level
   - `execution_engine`

https://peps.python.org/pep-0561/#packaging-type-information

See below for the modules we currently don't type-check.
Note: This doesn't mean that anything not in this list is "typed" it just means we run the type-checker against it. So any types we do declare should be correct. See `mypy` section on ["dynamically typed code"](https://mypy.readthedocs.io/en/stable/dynamic_typing.html)
https://github.com/great-expectations/great_expectations/blob/95f1c371846c0179d2558ef656b268bb080272ac/pyproject.toml#L50-L202

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -93,7 +93,7 @@ config = {
     "entry_points": {
         "console_scripts": ["great_expectations=great_expectations.cli:main"]
     },
-    "package_data": {"great_expectations": ["py.typed", "**/*.pyi"]},
+    "package_data": {"great_expectations": ["**/py.typed", "**/*.pyi"]},
     "name": "great_expectations",
     "long_description": long_description,
     "license": "Apache-2.0",


rdodev (approved):
LGTM

PR #7469, opened 2023-03-27
[MAINTENANCE] SqlAlchemy 2 Compatibility - `engine.execute()`

Shinnnyshinshin:
# What does this PR do? 
* Removes `engine.execute()` from the codebase, which is deprecated in SqlAlchemy 2
* In cases where `connection` is scoped, we use a `with` 
* In cases where Engine is passed along as an attribute, we create a connection with `engine.connect()` before executing query. 

### Definition of Done

Thank you for submitting!


ghost:
<details open>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=7469&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

####

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=7469&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map legend
</details>

great_expectations/core/expectation_validation_result.py
```diff
@@ -515,11 +515,11 @@ def __init__(
         if statistics is None:
             statistics = {}
         self.statistics = statistics
+
         if meta is None:
             meta = {}
-        ensure_json_serializable(
-            meta
-        )  # We require meta information to be serializable.
+
```
Shinnnyshinshin: This was just a change by the linter. Nothing has changed the code.

great_expectations/core/util.py
```diff
@@ -427,6 +427,13 @@ def convert_to_json_serializable(  # noqa: C901 - complexity 32
     if StructType is not None and isinstance(data, StructType):
         return dict(data.jsonValue())
 
+    if sqlalchemy is not None and isinstance(data, sqlalchemy.sql.elements.TextClause):
+        # TextClause is converted to str manually
+        return str(data)
+    if sqlalchemy is not None and isinstance(data, sqlalchemy.engine.base.Connection):
```
anthonyburdi: Let's use TextClause that is imported above, and use the same pattern for Connection. This way we are consistently handling imports from within the sqlalchemy library and can handle issues (like moving / removed imports between sqlalchemy versions) in one place.

great_expectations/data_context/store/database_store_backend.py
```diff
@@ -355,7 +362,7 @@ def remove_key(self, key):
             )
         )
         try:
-            return self.engine.execute(delete_statement)
+            return self._get_connection().execute(delete_statement)
```
anthonyburdi: I'm not seeing _get_connection() defined - were you able to get that to work or should this revert to using the context manager?

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1144,7 +1147,10 @@ def execute_split_query(self, split_query: Selectable) -> List[Row]:
             pattern = re.compile(r"(CAST\(EXTRACT\(.*?\))( AS STRING\))", re.IGNORECASE)
             split_query = re.sub(pattern, r"\1 AS VARCHAR)", split_query)
 
-        return self.engine.execute(split_query).fetchall()
+        if isinstance(self.engine, Engine):
+            self.engine = self.engine.connect()
+        query_result: List[Row] = self.engine.execute(split_query).fetchall()
+        return query_result
```
anthonyburdi: What if instead here we use the connection instead of reassigning to self.engine? That would remove the side effect.

```suggestion
        if isinstance(self.engine, Engine):
            connection = self.engine.connect()
        query_result: List[Row] = connection.execute(split_query).fetchall()
        return query_result
```

great_expectations/expectations/metrics/util.py
```diff
@@ -501,6 +501,8 @@ def column_reflection_fallback(
                 columns_table_query.c.column_id.asc(),
             )
         )
+        if isinstance(sqlalchemy_engine, Engine):
+            sqlalchemy_engine = sqlalchemy_engine.connect()
```
anthonyburdi: How about we call this connection instead of `sqlalchemy_engine` and then change the references below it? This comment applies to all instances in this file.

tests/test_utils.py
```diff
@@ -151,6 +153,9 @@ def validate_uuid4(uuid_string: str) -> bool:
 
 
 def get_sqlite_temp_table_names(engine):
+    if isinstance(engine, Engine):
+        engine = engine.connect()
+
     result = engine.execute(
```
anthonyburdi: Let's call this connection

```suggestion
        connection = engine.connect()

    result = connection.execute(
```

anthonyburdi (changes requested):
Thank you so much for these changes! I have a few suggestions that I think we should address before merging. Happy to discuss any of them!

great_expectations/core/util.py
Shinnnyshinshin: Great suggestion. Thanks :)

great_expectations/data_context/store/database_store_backend.py
Shinnnyshinshin: This one is strange. May be a merging error 😓 thanks for catching.

great_expectations/expectations/metrics/util.py
```diff
@@ -501,7 +507,7 @@ def column_reflection_fallback(
                 columns_table_query.c.column_id.asc(),
             )
         )
-        col_info_tuples_list: List[tuple] = sqlalchemy_engine.execute(
+        col_info_tuples_list: List[tuple] = connection.execute(
```
Shinnnyshinshin: renamed

great_expectations/core/util.py
```diff
@@ -427,6 +429,13 @@ def convert_to_json_serializable(  # noqa: C901 - complexity 32
     if StructType is not None and isinstance(data, StructType):
         return dict(data.jsonValue())
 
+    if sqlalchemy is not None and isinstance(data, TextClause):
+        # TextClause is converted to str manually
+        return str(data)
+    if sqlalchemy is not None and isinstance(data, Connection):
+        # Connection is a module, which is non-serializable. Return module name instead.
+        return "sqlalchemy.engine.base.Connection"
```
anthonyburdi: I think these need to be 
```suggestion
    if sqlalchemy and isinstance(data, TextClause):
        # TextClause is converted to str manually
        return str(data)
    if sqlalchemy and isinstance(data, Connection):
        # Connection is a module, which is non-serializable. Return module name instead.
        return "sqlalchemy.engine.base.Connection"
```
Since the NotImported type returns false (not None).

great_expectations/core/util.py
```diff
@@ -534,7 +543,11 @@ def ensure_json_serializable(data):  # noqa: C901 - complexity 21
         return
 
     if sqlalchemy is not None and isinstance(data, sqlalchemy.sql.elements.TextClause):
-        return str(data)
+        # TextClause is handled manually by convert_to_json_serializable()
+        return
+    if sqlalchemy is not None and isinstance(data, sqlalchemy.engine.base.Connection):
+        # Connection module is handled manually by convert_to_json_serializable()
+        return
```
anthonyburdi: Same comment here (and anywhere else in this file)

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: I didn't see this change implemented - would you be able to make this change (in the other part of the file as well) or do we need to set the engine property to a connection? If so can you add a comment that we are overriding with a connection?

great_expectations/expectations/metrics/util.py
```diff
@@ -405,6 +405,12 @@ def column_reflection_fallback(
     selectable: Select, dialect: Dialect, sqlalchemy_engine: Engine
 ) -> List[Dict[str, str]]:
     """If we can't reflect the table, use a query to at least get column names."""
+
+    if isinstance(sqlalchemy_engine, Engine):
+        connection: Connection = sqlalchemy_engine.connect()
```
anthonyburdi: We should probably use the context manager here so that the connection is closed after use (https://docs.sqlalchemy.org/en/20/core/connections.html#basic-usage)

tests/conftest.py
```diff
@@ -1703,10 +1703,11 @@ def titanic_sqlite_db(sa):
...
         engine = create_engine(f"sqlite:///{titanic_db_path}")
-        assert engine.execute(sa.text("select count(*) from titanic")).fetchall()[
-            0
-        ] == (1313,)
-        return engine
+        with engine.begin() as connection:
+            assert connection.execute(
+                sa.text("select count(*) from titanic")
+            ).fetchall()[0] == (1313,)
+            return connection
```
anthonyburdi: Just looking closer at this - here we are beginning a transaction with engine.begin(). And the original method returns an engine. Should this instead be:
```suggestion
        with engine.connect() as connection:
            assert connection.execute(
                sa.text("select count(*) from titanic")
            ).fetchall()[0] == (1313,)
        return engine
```

tests/conftest.py
```diff
@@ -1719,9 +1720,10 @@ def titanic_sqlite_db_connection_string(sa):
 
         titanic_db_path = file_relative_path(__file__, "./test_sets/titanic.db")
         engine = create_engine(f"sqlite:////{titanic_db_path}")
-        assert engine.execute(sa.text("select count(*) from titanic")).fetchall()[
-            0
-        ] == (1313,)
+        with engine.begin() as connection:
```
anthonyburdi, line 1725: Is there a reason to use engine.begin() instead of engine.connect() here?

tests/conftest.py
```diff
@@ -1759,7 +1761,8 @@ def empty_sqlite_db(sa):
         from sqlalchemy import create_engine
 
         engine = create_engine("sqlite://")
-        assert engine.execute(sa.text("select 1")).fetchall()[0] == (1,)
+        with engine.begin() as connection:
```
anthonyburdi, line 1766: Same comment.

tests/data_context/test_data_context_test_yaml_config.py
```diff
@@ -41,31 +41,35 @@ def test_connectable_postgresql_db(sa, test_backends, test_df):
         database="test_ci",
     )
     engine = sa.create_engine(url)
-
-    schema_check_results = engine.execute(
-        sa.text(
-            "SELECT schema_name FROM information_schema.schemata WHERE schema_name = 'connection_test';"
-        )
-    ).fetchall()
+    with engine.begin() as connection:
```
anthonyburdi, line 46: Same comment (engine.connect() vs engine.begin())

tests/expectations/test_null_filters.py
```diff
@@ -61,5 +61,7 @@ def test_sa_null_filters(sa):
     # Demonstrate that spark's max aggregate function can tolerate null values
     df = pd.DataFrame({"a": [1, 2, 3, None, None, 4]})
     df.to_sql(name="test", con=eng, index=False)
-
-    assert eng.execute(sa.text(f"SELECT MAX(a) FROM test;")).fetchone()[0] == 4
+    with eng.begin() as connection:
```
anthonyburdi, line 69: same comment

tests/test_utils.py
```diff
@@ -151,7 +154,12 @@ def validate_uuid4(uuid_string: str) -> bool:
 
 
 def get_sqlite_temp_table_names(engine):
-    result = engine.execute(
+    if isinstance(engine, Engine):
+        connection: Connection = engine.connect()
```
anthonyburdi: Same comment about putting this in a with block to ensure the connection is closed

anthonyburdi (changes requested):
Sorry I missed a few things the first time through.

great_expectations/core/util.py
Shinnnyshinshin: Thank you for catching this @anthonyburdi. This is because the `NotImported` class in `optional_imports.py` has a `__bool__`, and this make it different from the pattern followed by `StructType` in like 429 of this current file?

great_expectations/core/util.py
Shinnnyshinshin: Thank you. These should be the only two spots.

tests/conftest.py
Shinnnyshinshin: you're absolutely right :) thanks for the catch here.

tests/execution_engine/test_sqlalchemy_execution_engine.py
```diff
@@ -993,8 +993,8 @@ def validate_tmp_tables():
         "unexpected_condition": condition_metric,
     }
     # noinspection PyUnusedLocal
-    results = engine.resolve_metrics(
+    results = execution_engine.resolve_metrics(
```
Shinnnyshinshin, line 996: A more-descriptive parameter

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1058,6 +1058,9 @@ def resolve_metric_bundle(
                     )
 
                 logger.debug(f"Attempting query {str(sa_query_object)}")
+
+                if isinstance(self.engine, Engine):
```
anthonyburdi: Just a heads up that if `Engine` is `None` this will raise a `TypeError`

great_expectations/core/util.py
anthonyburdi: Exactly! We'll be migrating over to the new style.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1058,6 +1060,9 @@ def resolve_metric_bundle(
                     )
 
                 logger.debug(f"Attempting query {str(sa_query_object)}")
+
+                if sqlalchemy_Engine and isinstance(self.engine, sqlalchemy_Engine):
+                    self.engine = self.engine.connect()
```
Shinnnyshinshin, line 1065: This lines bug me a bit. I wonder if it causes some inconsistencies, since `self.engine` can be either an `Engine` or`Connection`

great_expectations/expectations/metrics/util.py
```diff
@@ -405,214 +405,220 @@ def column_reflection_fallback(
...
-            schema="sys",
-        ).alias("sys_tables_table_clause")
-        tables_table_query: Select = (
-            sa.select(
-                tables_table_clause.c.object_id.label("object_id"),
-                sa.func.schema_name(tables_table_clause.c.schema_id).label(
-                    "schema_name"
-                ),
-                tables_table_clause.c.name.label("table_name"),
+    with sqlalchemy_engine.begin() as connection:
```
Shinnnyshinshin: 🚀

great_expectations/core/util.py
```diff
@@ -533,8 +544,12 @@ def ensure_json_serializable(data):  # noqa: C901 - complexity 21
     if isinstance(data, RunIdentifier):
         return
 
-    if sqlalchemy is not None and isinstance(data, sqlalchemy.sql.elements.TextClause):
-        return str(data)
+    if sqlalchemy and isinstance(data, TextClause):
```
anthonyburdi: Thinking about this further, I think we need to make sure that TextClause is imported. For example, in the case where we are running a version of sqlalchemy that doesn't contain TextClause. In that case sqlalchemy would be truthy but this would fail since TextClause doesn't exist.
```suggestion
    if TextClause and isinstance(data, TextClause):
```

great_expectations/core/util.py
```diff
@@ -533,8 +544,12 @@ def ensure_json_serializable(data):  # noqa: C901 - complexity 21
...
+        # TextClause is handled manually by convert_to_json_serializable()
+        return
+    if sqlalchemy and isinstance(data, sqlalchemy_Connection):
```
anthonyburdi:
```suggestion
    if sqlalchemy_Connection and isinstance(data, sqlalchemy_Connection):
```

great_expectations/core/util.py
```diff
@@ -427,6 +431,13 @@ def convert_to_json_serializable(  # noqa: C901 - complexity 32
     if StructType is not None and isinstance(data, StructType):
         return dict(data.jsonValue())
 
+    if sqlalchemy and isinstance(data, TextClause):
```
anthonyburdi:
```suggestion
    if sqlalchemy and TextClause and isinstance(data, TextClause):
```

great_expectations/data_context/store/database_store_backend.py
```diff
@@ -170,7 +172,7 @@ def store_backend_id(self) -> str:
             self._store_backend_id = f"{self.STORE_BACKEND_ID_PREFIX}{store_id}"
         return self._store_backend_id.replace(self.STORE_BACKEND_ID_PREFIX, "")
 
-    def _build_engine(self, credentials, **kwargs) -> "sa.engine.Engine":
+    def _build_engine(self, credentials, **kwargs) -> sa.engine.Engine:
```
anthonyburdi: Let's import this instead from `optional_imports` (e.g. `sqlalchemy_Engine`)

great_expectations/data_context/store/database_store_backend.py
```diff
@@ -198,7 +200,7 @@ def _build_engine(self, credentials, **kwargs) -> "sa.engine.Engine":
     @staticmethod
     def _get_sqlalchemy_key_pair_auth_url(
         drivername: str, credentials: dict
-    ) -> Tuple["URL", Dict]:
+    ) -> Tuple[URL, Dict]:
```
anthonyburdi: Let's import this instead from optional_imports

great_expectations/core/util.py
anthonyburdi: Actually how about just
```suggestion
    if TextClause and isinstance(data, TextClause):
```

great_expectations/core/util.py
```diff
@@ -427,6 +431,13 @@ def convert_to_json_serializable(  # noqa: C901 - complexity 32
...
+        # TextClause is converted to str manually
+        return str(data)
+    if sqlalchemy and isinstance(data, sqlalchemy_Connection):
```
anthonyburdi:
```suggestion
    if sqlalchemy_Connection and isinstance(data, sqlalchemy_Connection):
```

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: I've been thinking about this a bit more. Maybe we persist the connection in a private variable depending on dialect (i.e. persist if the dialect needs the connection to keep the temp tables around) and then preferentially use that connection if it exists, otherwise use the engine to create a new connection. What do you think?

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1144,7 +1149,10 @@ def execute_split_query(self, split_query: Selectable) -> List[Row]:
             pattern = re.compile(r"(CAST\(EXTRACT\(.*?\))( AS STRING\))", re.IGNORECASE)
             split_query = re.sub(pattern, r"\1 AS VARCHAR)", split_query)
 
-        return self.engine.execute(split_query).fetchall()
+        # self.engine can be a Connection or Engine, but a engine.engine will always be the Engine object
+        with self.engine.engine.begin() as connection:
+            query_result: List[Row] = connection.execute(split_query).fetchall()
+            return query_result
```
anthonyburdi: I see the pattern here but I would prefer a conditional depending on `self.engine` type. Though maybe that isn't necessary with https://github.com/great-expectations/great_expectations/pull/7469#discussion_r1161981772

tests/expectations/test_util.py
```diff
@@ -415,7 +415,7 @@ def test_table_column_reflection_fallback(test_backends, sa):
         reflected_columns_list = column_reflection_fallback(
             selectable=selectable,
             dialect=sqlalchemy_engine.engine.dialect,
-            sqlalchemy_engine=sqlalchemy_engine.engine,
+            sqlalchemy_engine=sqlalchemy_engine.engine.engine,
         )
```
anthonyburdi: I think it might be better to be explicit here with an isinstance check on the type of `sqlalchemy_engine` to determine if it is an engine or connection.

tests/conftest.py
anthonyburdi: Sorry to re-hash this one, but can we return the engine instead of the connection (as it was previously?)

tests/test_utils.py
```diff
@@ -36,6 +36,7 @@
 
 try:
     import sqlalchemy as sa
+    from sqlalchemy.engine.base import Connection, Engine
```
anthonyburdi: Let's import instead from `optional_imports`

anthonyburdi (changes requested):
Sorry - just a few more items to address. Happy to discuss.

great_expectations/core/util.py
Shinnnyshinshin: I do think this is a "cleaner" way of checking whether `TextClause`, but wonder how often we are going to run into a case where the `sqlalchemy` version won't contain `TextClause`...`https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.TextClause` doesn't seem to indicate that it is a new / recent feature, so maybe it has been around for a while

great_expectations/core/util.py
Shinnnyshinshin: Sounds good :)

great_expectations/data_context/store/database_store_backend.py
Shinnnyshinshin: Love it. Thanks :)

great_expectations/data_context/store/database_store_backend.py
Shinnnyshinshin: Added to `optional_imports` and updated :)

great_expectations/execution_engine/sqlalchemy_execution_engine.py
Shinnnyshinshin: It makes sense I think. where would the logic for the "choice" between Engine and Connection live? As the code currently is, I think it live at the `SqlAlchemyExecutionEngine` (ie. here) + tests that deal with `SqlAlchemyBatchData`. 

I wonder if a helper method could be used here.

great_expectations/datasource/fluent/spark_google_cloud_storage_datasource.pyi
```diff
@@ -13,14 +13,14 @@ from great_expectations.datasource.fluent.config_str import (
...
-    SortersDefinition,
-)
 
 if TYPE_CHECKING:
     from google.cloud.storage.client import Client as GoogleCloudStorageClient
 
     from great_expectations.datasource.fluent import BatchMetadata
+    from great_expectations.datasource.fluent.interfaces import (
+        SortersDefinition,
+    )
```
Shinnnyshinshin: `ruff` was complaining on this one. Updated to push to branch.

great_expectations/expectations/metrics/util.py
```diff
@@ -405,214 +406,226 @@ def column_reflection_fallback(
...
-                columns_table_clause.c.user_type_id.label("user_type_id"),
-                columns_table_clause.c.column_id.label("column_id"),
-                columns_table_clause.c.name.label("column_name"),
-                columns_table_clause.c.max_length.label("column_max_length"),
-                columns_table_clause.c.precision.label("column_precision"),
+
+    if isinstance(sqlalchemy_engine.engine, sqlalchemy_Engine):
+        connection = sqlalchemy_engine.engine.connect()
+    else:
+        connection = sqlalchemy_engine.engine
```
Shinnnyshinshin, line 413: Added `Engine` / `Connection` check.

pyproject.toml
```diff
@@ -460,10 +460,6 @@ filterwarnings = [
     # Example Actual Warning: (tests/datasource/fluent/integration/conftest.py)
     # sqlalchemy.exc.SADeprecationWarning: Table.tometadata() is renamed to Table.to_metadata() (deprecated since: 1.4)
     'ignore: Table.tometadata\(\) is renamed to Table.to_metadata\(\):DeprecationWarning',
-
-    # SQLAlchemy 2.x support warnings. These warnings should be ignored until sqlalchemy 2.x is fully supported.
-    # To get SQLAlchemy 2.x supported, remove one of these ignores and then fix the resulting errors.
-    'ignore: The Engine.execute\(\) method is considered legacy as of the 1.x series of SQLAlchemy and will be removed in 2.0. All statement execution in SQLAlchemy 2.0 is performed by the Connection.execute\(\) method of Connection, or in the ORM by the Session.execute\(\) method of Session.:DeprecationWarning',
```
Shinnnyshinshin, line 466: finally

great_expectations/data_context/store/database_store_backend.py
Shinnnyshinshin: update on this one: the `static-type-check` stage doesn't seem to allow variables as a type.  
```
Variable "great_expectations.optional_imports.sqlalchemy_URL" is not valid as a
type  [valid-type]
        ) -> Tuple[sqlalchemy_URL, Dict]:
```

I'll keep the `URL` that we had in the PR for now

Shinnnyshinshin: Do you know why this might be the case? it seems to be handling return values fine.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
Shinnnyshinshin: This has been updated.

great_expectations/data_context/store/database_store_backend.py
Shinnnyshinshin: This also doesn't seem to work for typing: 

```
great_expectations/data_context/store/database_store_backend.py:179: error:
Variable "great_expectations.optional_imports.sqlalchemy_Engine" is not valid as
a type  [valid-type]
    ...  def _build_engine(self, credentials, **kwargs) -> sqlalchemy_Engine:

```

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1149,12 +1149,8 @@ def execute_split_query(self, split_query: Selectable) -> List[Row]:
...
             split_query = re.sub(pattern, r"\1 AS VARCHAR)", split_query)
 
-        if isinstance(self.engine, sqlalchemy_Engine):
-            connection = self.engine.connect()
-        else:
-            connection = self.engine
-
-        with connection:
+        # ensuring that we are executing the split query with an sa.Engine and not sa.Connection with engine.engine.
+        with self.engine.engine.begin() as connection:
```
Shinnnyshinshin: This had to be a guaranteed to be an `sa.Engine`, or else we would close the Connection

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: Yeah I agree. Something like the `get_connection()` that yields a connection that we discussed previously. And returns the same connection if we are using a specific dialect that requires the connection to stay open to keep temp tables alive.

great_expectations/core/util.py
```diff
@@ -72,7 +72,12 @@
     MultiPolygon = None
     LineString = None
 
-from great_expectations.optional_imports import SQLALCHEMY_NOT_IMPORTED, sqlalchemy
+from great_expectations.optional_imports import (
+    SQLALCHEMY_NOT_IMPORTED,
+    sqlalchemy,
+    sqlalchemy_Connection,
+    sqlalchemy_TextClause,
```
anthonyburdi, line 79: :heart:

great_expectations/core/util.py
```diff
@@ -427,6 +425,13 @@ def convert_to_json_serializable(  # noqa: C901 - complexity 32
     if StructType is not None and isinstance(data, StructType):
         return dict(data.jsonValue())
 
+    if sqlalchemy_TextClause and isinstance(data, sqlalchemy_TextClause):
+        # TextClause is converted to str manually
+        return str(data)
```
anthonyburdi: Is this a duplicate of the above? Looks like this is here twice

great_expectations/data_context/store/database_store_backend.py
anthonyburdi: Let's revert then for now. I am just concerned this will raise an AttributeError if sa is NotImported

great_expectations/data_context/store/database_store_backend.py
```diff
@@ -198,7 +201,7 @@ def _build_engine(self, credentials, **kwargs) -> "sa.engine.Engine":
     @staticmethod
     def _get_sqlalchemy_key_pair_auth_url(
         drivername: str, credentials: dict
-    ) -> Tuple["URL", Dict]:
+    ) -> Tuple[sa.engine.Engine.url.URL, Dict]:
```
anthonyburdi: Same here, let's revert for now and address in a followup.

great_expectations/data_context/store/database_store_backend.py
```diff
@@ -260,7 +263,9 @@ def _get(self, key):
             )
         )
         try:
-            return self.engine.execute(sel).fetchone()[0]
+            with self.engine.begin() as connection:
```
anthonyburdi, line 268: Can this not be self.engine.connect() as connection?

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: Can we use something more like this pattern here? (copied from map_condition_auxiliary_methods.py) or will that not work?
```
if sqlalchemy_Engine and isinstance(execution_engine.engine, sqlalchemy_Engine):
            connection = execution_engine.engine.connect()
        else:
            # execution_engine.engine is already a Connection. Use it directly
            connection = execution_engine.engine
```

anthonyburdi (approved):
Left some comments and suggestions. I'm going to approve to not block but please address especially the potential attribute errors (e.g. types).

great_expectations/data_context/store/database_store_backend.py
Shinnnyshinshin: Confirmed that this will break if we dont use `.begin()`

great_expectations/core/util.py
Shinnnyshinshin: You're absolutely right. Thank you for the catch

great_expectations/data_context/store/database_store_backend.py
Shinnnyshinshin: Got it. Sounds like a good idea

great_expectations/execution_engine/sqlalchemy_execution_engine.py
Shinnnyshinshin: I'm not 100% sure, so I'm submitting one more time with the proposed changes. If things break, then I'll revert back to this `begin()` pattern.

great_expectations/data_context/store/database_store_backend.py
Shinnnyshinshin: Done

Shinnnyshinshin:
@anthonyburdi regarding the `great_expectations/execution_engine/sqlalchemy_execution_engine.py` change you requested, I am testing with the `engine.connect()` pattern. If it works, I'll merge with the "better" pattern. If we get failures, I will revert back to the `engine.begin()` pattern. 

Update: It worked :)

great_expectations/core/util.py
```diff
@@ -82,13 +87,6 @@
 ):  # We need to catch an AttributeError since sqlalchemy>=2 does not have LegacyRow
     LegacyRow = SQLALCHEMY_NOT_IMPORTED
 
-# This is a separate try/except than the LegacyRow one since TextClause exists in sqlalchemy 2. This means LegacyRow
```
Shinnnyshinshin, line 85: Duplicate

PR #7472, opened 2023-03-27
[MAINTENANCE] Enable TCH251 - banned-api for sqlalchemy

Kilo59:
- gradual adoption for `TCH251` banned-api
-
-

### DIFF
--- a/great_expectations/datasource/simple_sqlalchemy_datasource.py
+++ b/great_expectations/datasource/simple_sqlalchemy_datasource.py
@@ -10,7 +10,9 @@ from great_expectations.datasource.data_connector.configured_asset_sql_data_conn
 from great_expectations.datasource.new_datasource import BaseDatasource
 
 if TYPE_CHECKING:
-    from sqlalchemy.engine import Engine as SaEngine
+    from sqlalchemy.engine import (  # noqa: TID251 # type-checking only
+        Engine as SaEngine,
+    )
 
     from great_expectations.execution_engine import SqlAlchemyExecutionEngine
 
--- a/great_expectations/optional_imports.py
+++ b/great_expectations/optional_imports.py
@@ -77,11 +77,12 @@ SQLALCHEMY_NOT_IMPORTED = NotImported(
 )
 try:
     import sqlalchemy
+    from sqlalchemy.exc import SQLAlchemyError
 
     sqlalchemy_version_check(sqlalchemy.__version__)
 except ImportError:
     sqlalchemy = SQLALCHEMY_NOT_IMPORTED
-
+    SQLAlchemyError = SQLALCHEMY_NOT_IMPORTED
 
 SPARK_NOT_IMPORTED = NotImported(
     "pyspark is not installed, please 'pip install pyspark'"
--- a/great_expectations/optional_imports.py
+++ b/great_expectations/optional_imports.py
@@ -77,12 +77,13 @@ SQLALCHEMY_NOT_IMPORTED = NotImported(
 )
 try:
     import sqlalchemy
-    from sqlalchemy.exc import SQLAlchemyError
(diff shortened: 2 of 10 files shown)


great_expectations/datasource/simple_sqlalchemy_datasource.py
```diff
@@ -10,7 +10,9 @@
 from great_expectations.datasource.new_datasource import BaseDatasource
 
 if TYPE_CHECKING:
-    from sqlalchemy.engine import Engine as SaEngine
+    from sqlalchemy.engine import (  # noqa: TID251 # type-checking only
+        Engine as SaEngine,
```
Kilo59, line 14: This was a bit annoying. Would be nice if we could configure this rule to not flag `TYPE_CHECKING` only imports

great_expectations/optional_imports.py
```diff
@@ -77,11 +77,13 @@ def is_version_greater_or_equal(
...
 try:
     import sqlalchemy
+    from sqlalchemy.exc import OperationalError, SQLAlchemyError
 
     sqlalchemy_version_check(sqlalchemy.__version__)
 except ImportError:
     sqlalchemy = SQLALCHEMY_NOT_IMPORTED
-
+    SQLAlchemyError = SQLALCHEMY_NOT_IMPORTED
+    OperationalError = SQLALCHEMY_NOT_IMPORTED
```
Kilo59, line 86: This is a trial balloon, not sure if this is a pattern we want to continue with.

pyproject.toml
```diff
@@ -286,6 +286,51 @@ extend-exclude = [
 "great_expectations/optional_imports.py" = [
     "TID251",  # flake8-banned-api
 ]
+"great_expectations/dataset/*" = [ # dataset is deprecated
+    "TID251",  # flake8-banned-api
+]
+
+# #########################################
+# BEGIN temporary banned-api exclusion list
+# TODO: remove packages from this list as they are addressed
```
Kilo59, line 295: We could also just make ruff add `noqa` comments for all these violations.
But then we would be touching a lot of files.

tests/ruff.toml
```diff
@@ -4,10 +4,13 @@
 extend = "../pyproject.toml"
 
 extend-ignore = [
-    # https://github.com/charliermarsh/ruff#flake8-type-checking-tch
+    # https://beta.ruff.rs/docs/rules/#flake8-type-checking-tch
     # This is likely to be a high-touch rule. Doing this in `tests` doesn't help circular imports.
     # Let's differ this for tests until there are auto-fixes
     "TCH001",
+    # https://beta.ruff.rs/docs/rules/#flake8-tidy-imports-tid
+    "TID251",  # banned-api - is used to prevent meanigless attribute errors caused by missing optional depedencies
```
Kilo59:
```suggestion
    "TID251",  # banned-api - is used to prevent opaque attribute errors caused by missing optional dependencies
```

great_expectations/optional_imports.py
billdirks: `OperationalError` is so general that I'm hesitant to add that here without any namespacing.

pyproject.toml
```diff
@@ -296,7 +341,7 @@ convention = "google"
 [tool.ruff.flake8-tidy-imports]
 [tool.ruff.flake8-tidy-imports.banned-api]
 # Uncomment this to prevent importing sqlalchemy everywhere but optional_imports.py
-# "sqlalchemy".msg = "Don't import sqlalchemy directly, please use the great_expectations.import_manager."
+"sqlalchemy".msg = "Don't import sqlalchemy directly, please use the great_expectations.optional_imports."
```
Kilo59, line 344:
```suggestion
"sqlalchemy.warnings.Removedin20".msg = "Removedin20 is only available in SQLAchemy1.4"
```

PR #7482, opened 2023-03-28
[MAINTENANCE] Lower pydantic requirement to v1.9.2 or greater

### LINKED ISSUE #7209: GX doesn't support airflow 2.2.2 anymore due to a new release (PR#7108)
**Describe the bug**
Even though I created a clean environment with just airflow2.2.2 and one dependency "airflow-provider-great-expectations==0.2.4"
I still get the following error (when I changed to 2.4.3 it worked fine):

Broken DAG: [/opt/airflow/dags/GREAT_EXPECTATION_TEST.py] Traceback (most recent call last):
  File "pydantic/fields.py", line 804, in pydantic.fields.ModelField.populate_validators
  File "pydantic/validators.py", line 723, in find_validators
RuntimeError: no validator found for <class 're.Pattern'>, see `arbitrary_types_allowed` in Config

**To Reproduce**
1. Open the airflow docker-compse community file (https://airflow.apache.org/docs/apache-airflow/**2.2.2**/docker-compose.yaml)
2. Add the "GX" dependency on the _PIP_ADDITIONAL_REQUIREMENTS: ${_PIP_ADDITIONAL_REQUIREMENTS:-airflow-provider-great-expectations}
3. Lunch airflow with docker-compse (docker-compose up airflow-init) -> (docker-compose up)

**Expected behavior**
Even though we hadn't created any dags, the RuntimeError mentioned above should raise.

 - Operating System: Linux
 - Great Expectations Version: latest

### LINKED ISSUE #7461: [BUG] Package conflict - pydantic with Airflow 2.2.3
**Describe the bug**
Hi, 
We are using GX on Airflow 2.2.3, and trying to upgrade to the latest version - 0.16.3 we have a package conflict. 

Since [this PR](https://github.com/great-expectations/great_expectations/commit/68752af349bfa5b8d7b654390efe790b7097f787), GX depends on pydantic 1.10.4 which depends on typing-extensions>=4.2.0. 

The issue is that[ Airflow 2.2.3 ](https://raw.githubusercontent.com/apache/airflow/constraints-2.2.3/constraints-3.7.txt)depends on typing-extensions 3.10.0.2, so we get a conflict when building our env. 

Can you loosen the range of pydantic so it will be compatible with this old-ish Airflow version? 
Thanks a lot!

Kilo59:
- loosen `pydantic` lower bound to `v1.9.2`
- make necessary changes needed for earlier versions of `pydantic` and `typing-extensions`
  - use `typing.Pattern` in place of `re.Pattern` for pydantic field annotations
  - stop using `typing_extensions.Self` at runtime
  - stop using `typing_extensions.NotRequired` at runtime
- `min_airflow` test steps for airflow `2.2.0`
   - part of normal min constrains job that runs in async pipeline
   - part of the airflow providers job that blocks PR merges

pydantic `v1.9.2` uses `'typing-extensions>=3.7.4.3'` so this should be compatible with airflow `2.2.3`
https://github.com/pydantic/pydantic/blob/1d37ffd89c9e6607db2553eb830fea2fc028dc2f/setup.py#L134
https://raw.githubusercontent.com/apache/airflow/constraints-2.2.3/constraints-3.7.txt

### Definition of Done


ci/azure-pipelines-dev.yml
```diff
@@ -239,6 +239,16 @@ stages:
...
 
+          - script: |
+              # Run unit-tests with select minamum versions
+              if $(run_unit_tests); then
+                # TODO: grep+sed >=VERSION -> ==VERSION
+                pip install typing-extentions==typing-extentions==3.10.0.0
+                invoke tests --ci --cloud --timeout=3.0
+              fi
+
+            displayName: 'Min Versions Unit Tests'
```
Kilo59: This probably isn't exactly what we want to do.
I would like this to be a blocking CI test, but this probably belongs somewhere else.

I could make this an async test initially but ideally, I would like these tests to block merges.

### Edit:
Going to move this (or something equivalent) to either the [min test constraints](https://github.com/great-expectations/great_expectations/tree/develop/ci/constraints-test) or the airflow providers section

ci/azure-pipelines-dev.yml
```diff
@@ -239,6 +239,16 @@ stages:
               GE_USAGE_STATISTICS_URL: ${{ variables.GE_USAGE_STATISTICS_URL }}
               SQLALCHEMY_WARN_20: true
 
+          - script: |
+              # Run unit-tests with select minamum versions
+              if $(run_unit_tests); then
+                # TODO: grep+sed >=VERSION -> ==VERSION
+                pip install typing-extentions==typing-extentions==3.10.0.0
```
Kilo59, line 246:
```suggestion
                pip install typing-extensions==3.10.0.0
```

ci/checks/get_context.py
```diff
@@ -0,0 +1,9 @@
+from pprint import pprint
+
```
Kilo59, line 2:
```suggestion
```

ci/checks/get_context.py
```diff
@@ -0,0 +1,9 @@
...
+import great_expectations as gx
+
+if __name__ == "__main__":
+    context = gx.get_context(cloud_mode=False)
+    assert context, "get_context() failed or result evaluated falsey"
+
+    print(f"Hello {type(context).__name__}!!")
```
Kilo59, line 9:
```suggestion
    print(f"Hello {type(context).__name__}!!")
```

ci/azure-pipelines-dev.yml
```diff
@@ -558,5 +558,8 @@ stages:
               versionSpec: '3.7'
             displayName: 'Use Python 3.7'
 
+          - script: ./ci/checks/check_min_airflow_dependency_compatibility.sh
+            name: CheckMinAirflowDependencyCompatibility.sh
+
```
Kilo59: @anthonyburdi 
The goal here is a minimal test that block PR merges to develop if we have a fatal error during import time or when `get_context()` is called.

There might be a better place to put this. 🤷

ci/azure-pipelines.yml
```diff
@@ -441,6 +441,9 @@ stages:
             Python310:
               python.version: '3.10'
               min_constraints: 'ci/constraints-test/py310-min-install.txt'
+            AirflowMin:
+              python.version: '3.7'
+              min_constraints: 'ci/constraints-test/py37-min-airflow.txt'
```
Kilo59: @billdirks 
This is the more comprehensive set of tests that should run in the async pipeline

ci/constraints-test/py37-min-airflow.txt
```diff
@@ -0,0 +1 @@
+--constraints https://raw.githubusercontent.com/apache/airflow/constraints-2.2.3/constraints-3.7.txt
```
Kilo59: This works with a constraints file
https://pip.pypa.io/en/stable/reference/requirements-file-format/#requirements-file-format

ci/checks/check_min_airflow_dependency_compatibility.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Install great_expectations using the airflow constraints file for our minimum supported version
+
+AIRFLOW_VERSION=2.2.3
```
Kilo59: @anthonyburdi @talagluck 
Do we say anything about the minimum version of airflow that we support?

This is the version the issue was raised about.
 - https://github.com/great-expectations/great_expectations/issues/7461

ci/checks/check_min_airflow_dependency_compatibility.sh
```diff
@@ -0,0 +1,20 @@
...
+PYTHON_VERSION="$(python --version | cut -d " " -f 2 | cut -d "." -f 1-2)"
+
+CONSTRAINT_URL="https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-${PYTHON_VERSION}.txt"
+# 3.7 example
+# https://raw.githubusercontent.com/apache/airflow/constraints-2.2.3/constraints-3.7.txt
+
+echo "Install gx dependencies using airflow constraint file"
+echo "Python:${PYTHON_VERSION}\nAirflow:${AIRFLOW_VERSION}\n${CONSTRAINT_URL}\n"
+
+pip install -e . --constraint "${CONSTRAINT_URL}"
```
anthonyburdi: Don't we also need to install using our constraint file? Can we pass both or merge the two first?

ci/checks/get_context.py
```diff
@@ -0,0 +1,6 @@
+import great_expectations as gx
+
+if __name__ == "__main__":
+    context = gx.get_context(cloud_mode=False)
+    assert context, "get_context() failed or result evaluated falsey"
+    print(f"Hello {type(context).__name__}!!")
```
anthonyburdi, line 6: ❤️

ci/checks/check_min_airflow_dependency_compatibility.sh
anthonyburdi: Looks like we can pass both using multiple constraint flags https://pip.pypa.io/en/stable/cli/pip_install/#install-constraint

ci/azure-pipelines-dev.yml
anthonyburdi: I think this is good for now. I have been thinking as our pipelines get more complex we will need to use something like [pipeline triggers](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/pipeline-triggers?view=azure-devops) to DRY our pipelines more - e.g. have a pipeline that just does linting and have that as a prerequisite for other pipelines. Instead of including linting steps in all of our pipelines. So for example there would be an airflow compatibility pipeline where this would live (similarly sqlalchemy, pandas, spark etc compatibility pipelines) triggered after prereqs like linting.

ci/checks/check_min_airflow_dependency_compatibility.sh
anthonyburdi: We support 2.x.latest. So we would want to test with the latest patch version of each minor version starting with 2.0.latest.

anthonyburdi (commented):
Left some comments for discussion. Appreciate the work here, thank you!

ci/checks/check_min_airflow_dependency_compatibility.sh
Kilo59: Supporting `2.0` might require more work than I what I've done in this PR.
I'll add a comment about this and lower it as much as possible, but going much lower will likely require a followup.

ci/checks/check_min_airflow_dependency_compatibility.sh
Kilo59: I'll add it. Thanks!

Kilo59: Thinking about this some more, I don't think we want to use our `constraints-dev.txt` as that's not how users install us. Doing this could potentially cause us to miss the dependency conflicts that our users experience.

ci/checks/check_min_airflow_dependency_compatibility.sh
donaldheppner: I'd like to know how low we can go with this PR, and lock it there

ci/azure-pipelines-dev.yml
billdirks: I wonder if we could use terraform instead of this and leverage its module composition functionality to DRY our pipelines while avoiding multiple pipelines and triggers.

ci/azure-pipelines.yml
billdirks: I wonder if we can combine this strategy with the other `3.7` strategy so we don't have to run all these tests twice.

billdirks (commented):
Thanks for doing this! One question inline about being able to combine the 2 python 3.7 min version steps in 1.

ci/checks/check_min_airflow_dependency_compatibility.sh
Kilo59: Going to update this comment with all the dependency conflicts or errors we get.

### `2.0.2 -> 2.1.4`
We have a conflict with `pytz` which prevents installation.

https://github.com/great-expectations/great_expectations/blob/1ca8948eb4fd4d192dad834cadde09932031d732/requirements.txt#L28

```
ERROR: Cannot install great-expectations==0.16.3+55.g57750997f.dirty because these package versions have conflicting dependencies.

The conflict is caused by:
    great-expectations 0.16.3+55.g57750997f.dirty depends on pytz>=2021.3
    The user requested (constraint) pytz==2021.1
```


### `2.2.0`

Looks like we can go as far back as `2.2.0` without any more code or dependency changes.


```
Install gx dependencies using airflow constraint file
Python:3.7
Airflow:2.2.0
https://raw.githubusercontent.com/apache/airflow/constraints-2.2.0/constraints-3.7.txt

...

Call 'get_context()' to check for show-stopping errors

python ci/checks/get_context.py
Hello EphemeralDataContext!!
```

ci/azure-pipelines-dev.yml
anthonyburdi: Interesting - that might also avoid lock in to azure.

ci/checks/check_min_airflow_dependency_compatibility.sh
anthonyburdi: Good point - forgot that our constraints are only dev.

ci/constraints-test/py37-min-airflow.txt
Kilo59, line 1:
```suggestion
--constraints https://raw.githubusercontent.com/apache/airflow/constraints-2.2.0/constraints-3.7.txt
```

ci/azure-pipelines.yml
Kilo59: @billdirks 
Done

ci/constraints-test/py37-min-install.txt
```diff
@@ -1,2 +1,3 @@
 numpy==1.18.5
 pandas==1.1.0
+--constraints https://raw.githubusercontent.com/apache/airflow/constraints-2.2.3/constraints-3.7.txt
```
Kilo59, line 3:
```suggestion
--constraints https://raw.githubusercontent.com/apache/airflow/constraints-2.2.0/constraints-3.7.txt
```

donaldheppner (approved):

PR #7483, opened 2023-03-28
[DOCS] Add a note about installing version of pyspark that matches Spark version

kenwade4:
- Add a note about installing version of pyspark that matches Spark version


Rachel-Reverie (approved):
LGTM

PR #7500, opened 2023-03-29
[DOCS] Adds guides on using Ephemeral Data Contexts and updates Quickstart Next Steps

Rachel-Reverie:
### Changes proposed in this PR:
- Adds new guide: "How to explicitly instantiate an Ephemeral Data Context"
- Adds new guide: "How to convert an Ephemeral Data Context to a Filesystem Data Context"
- Edits the Quickstart to remove the section about saving the Data Context for future use and adds a section to the Quickstart's "Next steps" on initializing, instantiating, and saving Data Contexts instead

### Definition of Done


rdodev (approved):

docs/docusaurus/docs/tutorials/quickstart/quickstart.md
```diff
@@ -179,20 +165,12 @@ To access GX Cloud, you need to join our Cloud Early Access program. During this
...
+Now that you've seen how easy it is to implement the GX workflow, it is time to customize that workflow to suit your specific use cases! To help with this we have prepared some more detailed guides on setting up and installing GX and getting an initial Data Context that are tailored to specific environments and resources.
 
-For information on installing GX for use with SQL databases, see:
+### Installing GX for specific environments and source data systems
 
-<SetupAndInstallForSqlData />
+<SetupAndInstallGx />
 
-And for instructions on installing GX for use with hosted data systems, read:
+### Initializing, instantiating, and saving a Data Context
```
talagluck, line 174: I really like the Next Steps section overall, but it still feels like a tough gap to bridge for the user to persist an ephemeral context. That doc is the last one in the list, and we don't mention Ephemeral Data Context anywhere in Quickstart, so for a user it would not be clear how to go from what they just did to something more persistent unless they start all over again.

docs/docusaurus/docs/guides/setup/configuring_data_contexts/how_to_convert_an_ephemeral_data_context_to_a_filesystem_data_context.md
```diff
@@ -0,0 +1,109 @@
...
+
+</details>
+
+## Steps
+
+### 1. Verify that your current working directory does not already contain a GX Filesystem Data Context
+
+The method for converting an Ephemeral Data Context to a Filesystem Data Context initializes the new Filesystem Data Context in the current working directory of the Python process that is being executed.  If a Filesystem Data Context already exists at that location, the process will fail.
+
+You can determine if your current working directory already has a Filesystem Data Context by looking for a `great_expectations.yml` file.  The presence of that file indicates that a Filesystem Data Context has already been initialized in the corresponding directory.
```
talagluck, line 80: I get that we want a separate doc on how to quickly instantiate a Data Context, but it feels odd to have this doc which is essentially a single line of code to run. I also know that I specifically requested this doc as a bridge from Quickstart.

One proposal would be to have a doc called something like "How to get and save a Data Context quickly" which has two steps, one for getting and one for saving. I don't love that either, tbh, but having the user need to click to an extra doc to run one line of code doesn't sit right with me. I'm open to brainstorming here.

rdodev (changes requested):
On a second thought, Tal does bring a good point. We should think about the way split/fragments exist.

docs/docusaurus/docs/guides/setup/configuring_data_contexts/how_to_convert_an_ephemeral_data_context_to_a_filesystem_data_context.md
Rachel-Reverie: I'm making an edit that adds an abbreviated explanation of the `convert_to_file_context()` method to an admonition box that can be included in both "How to quickly get a Data Context" and "How to explicitly instantiate an Ephemeral Data Context".  The admonition box directs people to "How to convert an Ephemeral Data Context to a Filesystem Data Context" if they are interested in a more detailed explanation of the `convert_to_file_context()` method.

I think this might be the best option -- I don't want to add steps to "How to quickly get a Data Context" that would essentially be flagged as "optional, only if you are using an Ephemeral Data Context" when we can assume that a fair number of people won't be using an Ephemeral Data Context.

One guide = one process, so the idea is to minimize the explicitly optional steps while still including the information that may be relevant.

talagluck (approved):
Overall, I like the changes. I still wonder whether there is anything we can do to help bridge the gap from the Quickstart more smoothly. But for now this works for me.

rdodev (approved):

donaldheppner (approved):

PR #7507, opened 2023-03-29
[MAINTENANCE] Use YAMLHandler in tests and docs

kenwade4:
This un-breaks 6 stages of the docs-integration pipeline. The remaining fixes (to get the pipeline fully green again) should go in new small PRs since their solutions (for the different backends) will not be too similar.

- Use YAMLHandler in tests and docs
- Don't rename any snippets so build_docs script doesn't error
- Add required dataset_name to example data for 4 custom Expectations

Future will need a similar PR making sure we use YAMLHandler only in the main code too.


Kilo59 (approved):

anthonyburdi:
This is great, thanks for doing this! Once all are removed we can set a tidy-imports directive in pyproject.toml to raise errors if we try to import ruamel directly (except in the yaml handler).

Kilo59:
@anthonyburdi actually can set the `banned-api` setting now?
I'll take to @kenwade4 about setting that up.

PR #7513, opened 2023-03-30
[FEATURE] Fluent `DataAsset` `batch_metadata` config variables

NathanFarmer:
- GREAT-1729
- Add variable substitution for `batch_metadata` that references config variables

### Definition of Done


great_expectations/datasource/fluent/interfaces.py
```diff
@@ -241,6 +238,25 @@ def _validate_batch_request(self, batch_request: BatchRequest) -> None:
...
+        return set(options.keys()).issubset(set(self.batch_request_options))
+
+    def _get_batch_metadata_from_batch_request(
+        self, batch_request: BatchRequest
+    ) -> BatchMetadata:
+        """Performs config variable substitution and populates batch request options for
+        Batch.metadata at runtime.
+        """
+        batch_metadata = copy.deepcopy(self.batch_metadata)
+        config_variables = self._datasource._data_context.config_variables
```
NathanFarmer: This means `_data_context` really is needed on all `Datasource`s.

tests/datasource/fluent/test_spark_datasource.py
```diff
@@ -53,3 +72,48 @@ def test_dataframe_asset(
             for asset in datasource.assets.values()
         ]
     )
+
+
+@pytest.mark.xfail(
+    strict=True,
+    reason="this will fail until we add batch_metadata to Spark add_*_asset methods",
+)
```
NathanFarmer, line 80: @billdirks your PR for Spark `batch_metadata`, should cause you to remove this xfail.

great_expectations/datasource/fluent/pandas_datasource.pyi
```diff
@@ -143,12 +143,14 @@ class PandasDatasource(_PandasDatasource):
     type: Literal["pandas"]
     assets: Dict[str, _PandasDataAsset]
     def test_connection(self, test_assets: bool = ...) -> None: ...
-    def _get_validator(self, asset: _PandasDataAssetT) -> Validator: ...
+    @staticmethod
+    def _validate_asset_name(self, asset_name: Optional[str] = ...) -> str: ...
+    def _get_validator(self, asset: _PandasDataAsset) -> Validator: ...
```
Kilo59: Why do these need to be part of the stub?
Are they being called from outside the class?

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -241,6 +238,25 @@ def _validate_batch_request(self, batch_request: BatchRequest) -> None:
...
+        batch_metadata = _ConfigurationSubstitutor().substitute_all_config_variables(
+            data=batch_metadata, replace_variables_dict=config_variables
+        )
```
Kilo59, line 252: Does this also automatically prevent these values from getting serialized to config?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1455,12 +1456,12 @@ def get_datasource(
         substituted_config = self.config_provider.substitute_config(raw_config_dict)
 
         # Instantiate the datasource and add to our in-memory cache of datasources, this does not persist:
-        datasource_config = datasourceConfigSchema.load(substituted_config)
```
NathanFarmer, line 1458: This variable was unused.

great_expectations/datasource/fluent/interfaces.py
NathanFarmer: Yes, the variables are never substituted on the `DataAsset` itself. This method only gets called in `get_batch_list_from_batch_request` and then the `BatchMetadata` is attached to each `Batch.metadata`.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -462,6 +478,7 @@ def get_asset(self, asset_name: str) -> _DataAssetT:
         """Returns the DataAsset referred to by name"""
         # This default implementation will be used if protocol is inherited
         try:
+            self.assets[asset_name]._datasource = self
```
NathanFarmer, line 477: The `get_datasource` and `get_asset` methods need to propagate updated `DataContext` -> `Datasource` -> `DataAsset` attributes. Otherwise you end up with missing `_data_context`s that are needed for config variable substitution and `get_validator` calls.

great_expectations/datasource/fluent/pandas_datasource.pyi
NathanFarmer: @Kilo59 I forgot to follow up with you about this. I think they can actually be removed from the stubs (I removed them). It just results in the IDE complaining about "Unresolved attribute", but I think the stubs should be reserved for public methods.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1443,6 +1443,7 @@ def get_datasource(
             )
 
         if datasource_name in self._cached_datasources:
+            self._cached_datasources[datasource_name]._data_context = self
```
NathanFarmer, line 1446: The `get_datasource` and `get_asset` methods need to propagate updated `DataContext` -> `Datasource` -> `DataAsset` attributes. Otherwise you end up with missing `_data_context`s that are needed for config variable substitution and `get_validator` calls.

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59: I would think the linter would have flagged this 🤷

great_expectations/datasource/fluent/interfaces.py
Kilo59: You might have just fixed an edge case I found with my `ConfigStr` logic 😍 
It was happening because of a missing `_data_contexts.config_provider` which I didn't understand at the time.

Kilo59 (approved):
🙌

tests/datasource/fluent/test_spark_datasource.py
billdirks: This actually still fails because the `DataFrameAsset` has not implemented `build_batch_request`. Would you like me to implement this here or can I do it in a follow up PR?

PR #7518, opened 2023-03-30
[MAINTENANCE] Dedicated airflow 2.2.0 async test

Kilo59:
- add a dedicated `Airflow2` test to `test_min_versions` async test matrix
- ~commit [airflow/constraints-2.2.0/constraints-3.7.txt](https://raw.githubusercontent.com/apache/airflow/constraints-2.2.0/constraints-3.7.txt) to source control~

### DIFF
--- a/ci/azure-pipelines.yml
+++ b/ci/azure-pipelines.yml
@@ -431,9 +431,12 @@ stages:
         condition: succeeded()
         strategy:
           matrix:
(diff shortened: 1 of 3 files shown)


billdirks (approved):
Thank you, LGTM! I would prefer to vendor the constraints instead of relying on an external link but we can do that later.

PR #7227, opened 2023-02-24
[MAINTENANCE] Repo Cleanup: Move docs related files under docs dir

anthonyburdi:
Changes proposed in this pull request:
- Move docs related files into docs/ dir instead of the repo root and reorganize.
- Please see https://63f8f49f4f56480008a9a46e--anthony-docs-base-dir-change-testing.netlify.app/docs/ for a deploy preview. I've done some review but please check this preview for any issues before we merge this PR!
- This PR doesn't address `docs_rtd` and `readme_assets`. Those will be handled in future work.
- To coincide with this PR merging, we will also need to change the build settings on the main site to the following:
image

### Definition of Done

- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/docs/contributing/style_guides/code_style)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/docs/contributing/contributing_checklist) of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have run any local integration tests and made sure that nothing is broken.

PR #7526, opened 2023-03-30
[MAINTENANCE] SqlAlchemy 2 Compatibility - Autoload Parameter deprecation 

Shinnnyshinshin:
- Removes use of `autoload=True` from creating a `Table` object. 
- Documentation asks that `autoload_with` be used instead, and it turns out, the `Table` is created with `autoload_with` already :) 
- Folding in changes from #7227 

### Definition of Done

Thank you for submitting!


great_expectations/data_context/store/database_store_backend.py
```diff
@@ -103,7 +103,7 @@ def __init__(  # noqa: C901 - 16
             cols.append(Column(column_, String, primary_key=True))
         cols.append(Column("value", String))
         try:
-            table = Table(table_name, meta, autoload=True, autoload_with=self.engine)
```
Shinnnyshinshin, line 106: This was the only occurrence of `autoload=True`. We are asked to use the `autoload_with` parameter instead, which we already do.

anthonyburdi (approved):
Awesome! LGTM!

PR #7552, opened 2023-04-03
[BUGFIX] Render prescriptive `ExpectationConfiguration`s with evaluation parameters inline

NathanFarmer:
- GREAT-73
- Handle evaluation parameter dictionary for prescriptive rendering
- Add evaluation parameter attribute to params for use after diagnostic substitution

### Definition of Done


tests/validator/test_validator.py
```diff
@@ -1007,45 +1007,53 @@ def test_validator_include_rendered_content(
...
+                    "schema": {"type": "number"},
+                    "value": 8,
+                    "evaluation_parameter": {
+                        "schema": {"type": "object"},
+                        "value": {"$PARAMETER": "upstream_column_max"},
+                    },
                 },
             },
-            template="$column maximum value must be greater than or equal to $min_value and less than or equal to $max_value.   $eval_param__0, $eval_param__1",
+            template="$column maximum value must be greater than or equal to $min_value and less than or equal to $max_value.",
```
NathanFarmer, line 1050: @josectobar I will hold off on merge until you confirm this works for the diagnostic view.

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
```diff
@@ -607,3 +611,72 @@ def test_add_or_update_expectation_suite_updates_existing_obj(
...
+                    "min_value": {
+                        "schema": {"type": "object"},
+                        "value": {"$PARAMETER": "upstream_column_min"},
+                    },
+                    "max_value": {
+                        "schema": {"type": "object"},
+                        "value": {"$PARAMETER": "upstream_column_max"},
+                    },
+                },
+            ),
```
NathanFarmer, line 668: @josectobar this is the rendering that was failing before (we weren't testing it).

great_expectations/render/renderer_configuration.py
```diff
@@ -487,7 +483,7 @@ def _choose_param_type_for_value(
                 pass
 
         raise RendererConfigurationError(
-            f"None of the param_types: {param_types} match the value: {value}"
+            f"None of the param_types: {[param_type.value for param_type in param_types]} match the value: {value}"
```
NathanFarmer, line 490: Tests for this error message will continue to pass, but without the `.value` we were seeing the string representation of the `Enum` class name instead of the string value itself in the message.

great_expectations/render/renderer_configuration.py
```diff
@@ -463,11 +464,6 @@ def _set_template_str(cls, v: str, values: dict) -> str:
             )
             v = f"{row_condition_str}, then {v}"
 
-        if "_raw_kwargs" in values and values["_raw_kwargs"]:
-            v += "  "
-            for evaluation_parameter_count in range(len(values["_raw_kwargs"])):
-                v += f" $eval_param__{evaluation_parameter_count},"
-            v = v[:-1]
```
NathanFarmer, line 470: The evaluation parameter expression is no longer being appended to the diagnostic templated string.

great_expectations/render/renderer_configuration.py
josectobar: Just a question, I thought `param_types` was of type `List[RendererValueType]` which is an enum, how does it contain a key `value` when you say `param_type.value for param_type in param_types`

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
```diff
@@ -607,3 +611,72 @@ def test_add_or_update_expectation_suite_updates_existing_obj(
...
+            name="atomic.prescriptive.summary",
+        )
+    ]
+
+    expectation_suite_include_rendered_content: ExpectationSuite = (
+        context.get_expectation_suite(
+            expectation_suite_name=expectation_suite_name,
+            include_rendered_content=True,
```
josectobar, line 676: Do we need to do this manually for now because of the other bug where it doesn't include_rendered_content=True by default or what is the reasoning here?

tests/validator/test_validator.py
```diff
@@ -1007,45 +1007,53 @@ def test_validator_include_rendered_content(
...
 
-    expected_expectation_validation_result_rendered_content = RenderedAtomicContent(
-        name="atomic.diagnostic.observed_value",
-        value=RenderedAtomicValue(
-            schema={"type": "com.superconductive.rendered.string"},
-            params={},
-            template="6",
-        ),
-        value_type="StringValueType",
+    expected_expectation_validation_result_diagnostic_rendered_content = (
```
josectobar, line 1010: Just curious why you added `diagnostic` to the name, to me a `validation_result` is implicitly a diagnostic-type expectation. It feels a bit repetitive.

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
NathanFarmer: The `include_rendered_content` bug is a mercury bug that I'm working on now. I'm just testing that the `context.get_expectation_suite()` CRUD operation triggers the `InlineRenderer` and produces `rendered_content` here.

tests/validator/test_validator.py
NathanFarmer: I think that is correct, but I feel that we need to be precise about these "modes", or we could start to get confused in the future. The `ExpectationConfiguration` in the `ExpectationValidationResult` does have both a "prescriptive" and "diagnostic" mode due to the difference in the way we render the evaluation parameters, but that this not what we are testing on this line.

josectobar (approved):
LGTM. There's a small question but otherwise great work!

great_expectations/render/renderer_configuration.py
NathanFarmer: Answered in an offline thread.

PR #7553, opened 2023-04-03
[BUGFIX] Address `pandas==2.0.0` test failures 

Shinnnyshinshin:
# Changes proposed in this pull request. 
### Handles the following errors coming from `pandas==2.0.0`
* `pd.to_datetime()` parameter changes: 
   1. `format` is now a required parameter (can be format string `mixed` or `ISO8061`)
   2. `infer_datetime_format` is deprecated. 
* `pd.to_datetime()` is now handled by a helper method `execute_pandas_to_datetime()` which wraps the method call and calls it differently according to pandas version. 
* Moves `ConfigStr` to a `TYPE_CHECKING` block in `great_expectations/datasource/fluent/pandas_s3_datasource.pyi`
* converts 2 tests in `tests/expectations/metrics/test_core.py` to use `pd.Index()` rather than `pd.Int64Index`, which has been deprecated. 

### The following tests are skipped
* Uses `pytest.skip()` to skip 3 tests that are failing in `tests/expectations/test_expectation_atomic_renderers.py` due to a incompatibility between `pandas==2.0.0` and `altair`
* Uses `pytest.skip()` to skip 2 tests in `tests/cli/test_suite.py` due to an incompatibility in a Notebook test. These tests are also being further investigated. 

### Definition of Done


great_expectations/compatibility/sqlalchemy_and_pandas.py
```diff
@@ -45,3 +45,84 @@ def execute_pandas_reader_fn(
     else:
         reader_fn_result = reader_fn(**reader_options)
     return reader_fn_result
+
+
+def execute_pandas_to_datetime(
```
anthonyburdi: Totally optional nit - since this is pandas specific can it go in a new `compatibility/pandas.py`? I was intending this file for compatibility issues that required changes based on both the sqlalchemy and the pandas versions.

tests/expectations/metrics/test_core.py
```diff
@@ -1991,9 +1991,7 @@ def test_map_column_values_increasing_pd():
     )
     metrics.update(results)
 
-    assert metrics[unexpected_rows_metric.id]["a"].index == pd.Int64Index(
-        [4], dtype="int64"
-    )
+    assert metrics[unexpected_rows_metric.id]["a"].index == pd.Index([4], dtype="int64")
```
anthonyburdi, line 1994: Will these fail on our earliest supported pandas versions? If so we may need to bifurcate based on installed pandas version, if not then 👍

tests/cli/test_suite.py
```diff
@@ -1975,7 +1986,6 @@ def test_suite_edit_multiple_datasources_with_sql_with_no_additional_args_with_c
 
     The command should:
     - NOT open Data Docs
-    - NOT open jupyter
```
Shinnnyshinshin, line 1978: comment is inaccurate

tests/expectations/test_expectation_atomic_renderers.py
```diff
@@ -160,6 +162,10 @@ def test_atomic_prescriptive_summary_expect_column_distinct_values_to_equal_set(
     snapshot.assert_match(res)
 
 
+@pytest.mark.skipif(
+    is_version_greater_or_equal(pd.__version__, "2.0.0"),
+    reason="Altair is currently not compatible with pandas 2.0.0",
```
anthonyburdi: Can we make these three tests xfail strict instead of skip if the pandas version is >2.0.0? That way when altair is compatible these tests will get updated or re-enabled.

tests/cli/test_suite.py
```diff
@@ -1692,6 +1695,10 @@ def test_suite_edit_multiple_datasources_with_no_additional_args_with_citations_
     )
 
 
+@pytest.mark.skipif(
+    is_version_greater_or_equal(pd.__version__, "2.0.0"),
+    reason="Test is currently not compatible with pandas 2.0.0",
```
anthonyburdi: Same comment as the other tests, can we make xfail strict if pandas >2.0.0?

anthonyburdi (commented):
Just a few comments / questions

great_expectations/compatibility/sqlalchemy_and_pandas.py
Shinnnyshinshin: Ah sure that isn't a problem

tests/expectations/test_expectation_atomic_renderers.py
Shinnnyshinshin: Good call. Updating now

great_expectations/compatibility/sqlalchemy_and_pandas.py
```diff
@@ -45,3 +45,84 @@ def execute_pandas_reader_fn(
...
+    arg,
+    errors: str = "raise",
```
Kilo59: non-blocking

Could use a `typing_extensions.Literal` to help narrow the possible values. I don't think it's that useful here since this is mostly just a passthrough but wanted to bring it up regardless.
```suggestion
    errors: Literal["raise", "coerce", "ignore"] = "raise",
```

Kilo59 (approved):

tests/expectations/test_expectation_atomic_renderers.py
Kilo59: But won't this xpass in parts of our CI? Which means it fails if it's a `strict` xfail

Shinnnyshinshin: Oh is that the case @kilo59? Is there a way to do a "conditional" xfail?

great_expectations/compatibility/sqlalchemy_and_pandas.py
Shinnnyshinshin: updating now :)

great_expectations/compatibility/sqlalchemy_and_pandas.py
Shinnnyshinshin: done

tests/expectations/metrics/test_core.py
Shinnnyshinshin: Yes this will work with earlier versions 👍🏼

tests/expectations/test_expectation_atomic_renderers.py
Shinnnyshinshin: Found it:  I think we can be more specific with the xfail as @anthonyburdi is describing. Will update / run the pipeline with the described changes.

Kilo59: @Shinnnyshinshin If we are saying that this "test could fail and that's ok" that's just normal `xfail` `strict=False`.

As @anthonyburdi said `xfail(strict=True)` can be useful if we want to force some code change once the test no-longer fails.

But I would assume this test will `XPASS` in our CI when `pandas<2.0.0` right? And if that's the case we can't use `xfail(strict=True)` because it will fail the CI step .

If we are saying "this test only makes sense under condition A"  `skipif` is the right choice.

method |pandas| CI passes?
--------|--------|-------
strict xfail | 2.0 | ❌  - CI fails if the test passes
 strict xfail | 1.5.3 | ✅ - CI passes if the test fails
normal xfail | 2.0 | ✅  - doesn't matter if pass or fail
normal xfail | 1.5.3 | ✅ - doesn't matter if pass or fail
skipif | 2.0 | ✅ 
skipif | 1.5.3 | ✅ 

The above table is assuming we run this test with both pandas `2.0` and earlier versions.


Should also note you could put a conditional inside the test with xfail like in the example in the link below.
https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail

PR #7561, opened 2023-04-04
[MAINTENANCE] Simplify `GXCloudStoreBackend._has_key` check

cdkini:
- Instead of listing all available keys and iterating through them, let's try to retrieve and use the success of that call to determine existence.

### Definition of Done


great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -503,17 +503,12 @@ def remove_key(self, key):
...
-
-        key = _shorten_key(key)
-        all_keys = set(map(_shorten_key, self.list_keys()))
-        return key in all_keys
+        try:
+            self._get(key)
+        except StoreBackendError as e:
+            logger.debug(f"Could not find object associated with key {key}: {e}")
+            return False
+        return True
```
Kilo59: I think we only really want this if we get a `404` right?
Is it okay for this to silently fail due to a temporary `500` outage?

At minimum could we move this up to an `info` level log?

Kilo59 (approved):

PR #7565, opened 2023-04-04
[MAINTENANCE] Add ignore revs

anthonyburdi:
- Ignore from git blame some commits that just move files around.

### Definition of Done

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -9,3 +9,9 @@ f5e7df1846102d9a62cc9b9110387925ffae60cc
 # Apply noqa markes for all PTH (use-pathlib) violations
 # https://github.com/great-expectations/great_expectations/pull/7290
 597b2b625569b6f5f110f8230ac26ab405167da6
+# [MAINTENANCE] Move some scripts into ci folder #7271
+# https://github.com/great-expectations/great_expectations/pull/7271
+035d479a56484a0b8f4475f57dc6814b4757daee
+# [MAINTENANCE] Repo Cleanup: Move docs related files under docs dir #7227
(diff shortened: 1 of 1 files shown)


.git-blame-ignore-revs
```diff
@@ -9,3 +9,9 @@ f5e7df1846102d9a62cc9b9110387925ffae60cc
 # Apply noqa markes for all PTH (use-pathlib) violations
 # https://github.com/great-expectations/great_expectations/pull/7290
 597b2b625569b6f5f110f8230ac26ab405167da6
+# [MAINTENANCE] Move some scripts into ci folder #7271
+# https://github.com/great-expectations/great_expectations/pull/7271
+035d479a56484a0b8f4475f57dc6814b4757daee
+# [MAINTENANCE] Repo Cleanup: Move docs related files under docs dir #7227
+https://github.com/great-expectations/great_expectations/pull/7227
+0c9e4502e7dfd4984a21ca4c8da0d0fcfff37aed
```
Kilo59, line 17: I don't think we should ignore these. What's the benefit?
If nothing changed in the file as part of this move the blame should already reflect that.

Imagine there's an asset somewhere whose path has changed as a result of this move and we had to track down some bug or issue that resulted, it would be important that this move was included in our git blame.

Contrast that with the common uses for `.git-blame-ignore-revs` which is for formatting changes.
Seeing every formatting change in our source code is not useful, it will never help us determine when a regression was introduced. And it makes it harder to see when substantive changes actually happened.

anthonyburdi: Noted - I erroneously thought that moving the files would change the git blame to me. This makes sense, closing the PR.

anthonyburdi:
Created in error.

PR #7570, opened 2023-04-05
[FEATURE] add Fluent Datasources to `CloudDataContext`

Kilo59:
- Allow for Fluent Datasource to be added to and loaded from a `CloudDataContext`
- [add `responses` as testing dependency for more easily mocking cloud API endpoints](https://github.com/getsentry/responses)
   - [x] add this to requirements

### Followups
1. Delete Fluent Datasources
2. Persist assets immediately on `datasources.add_<ASSET_TYPE>()` calls.
3. Update Fluent Datasources

### Definition of Done


great_expectations/data_context/store/datasource_store.py
```diff
@@ -62,10 +63,14 @@ def remove_key(self, key: Union[DataContextVariableKey, GXCloudIdentifier]) -> N
...
-    def serialize(self, value: DatasourceConfig) -> Union[str, dict, DatasourceConfig]:
+    def serialize(
+        self, value: DatasourceConfig | FluentBaseModel
+    ) -> Union[str, dict, DatasourceConfig]:
         """
         See parent 'Store.serialize()' for more information
         """
+        if isinstance(value, FluentBaseModel):  # TODO: this is gross, do something else
+            return value._json_dict()
         return self._serializer.serialize(value)
```
Kilo59: Should probably do this in the `Serializer`

great_expectations/data_context/store/datasource_store.py
```diff
@@ -146,8 +165,8 @@ def set(  # type: ignore[override]
             DatasourceConfig retrieved from the DatasourceStore.
         """
         if not key:
-            key = self._build_key_from_config(value)
-        return self._persist_datasource(key=key, config=value)
+            key = self._build_key_from_config(value)  # type: ignore[arg-type]
+        return self._persist_datasource(key=key, config=value)  # type: ignore[arg-type]
```
Kilo59: Deal with these typing issues before merge
`_build_key_from_config` could accept a thinner interface/protocol.

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -100,6 +100,10 @@ def _set(self, key: tuple[str, ...], value: Any, **kwargs: dict) -> None:
 
         if resource_type is DataContextVariableSchema.ALL_VARIABLES:
             config_commented_map_from_yaml = yaml.load(value)
+            # TODO: instead prevent these from being set??
+            # TODO: revert this before merge
+            for name in config_commented_map_from_yaml.get("fluent_datasources", {}):
+                config_commented_map_from_yaml["datasources"].pop(name)
```
Kilo59, line 106: When we call `self._datasource_store.set(key=None, value=datasource)` this causes the `fluent` datasources to be added to the `datasources` config entry, which will then be validated using the old "Datasource" marshmallow schemas.

This temporary hack is just to demonstrate the problem.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -709,9 +709,12 @@ def _add_fluent_datasource(self, datasource: FluentDatasource) -> None:
                 f"Can not write the fluent datasource {datasource.name} because a datasource of that "
                 "name already exists in the data context."
             )
+        # store update here?
         self.datasources[datasource.name] = datasource
+        self._datasource_store.set(key=None, value=datasource)
 
     def _update_fluent_datasource(self, datasource: FluentDatasource) -> None:
+        # store update here?
         self.datasources[datasource.name] = datasource
```
Kilo59: We don't seem to actually call this anywhere in our code right now.
The `context.sources.add_or_update_<DATASOURCE_TYPE>()` method always calls `_add_fluent_datasource`.

great_expectations/data_context/store/datasource_store.py
```diff
@@ -62,20 +64,34 @@ def remove_key(self, key: Union[DataContextVariableKey, GXCloudIdentifier]) -> N
...
+        self, value: DatasourceConfig | FluentDatasource
+    ) -> Union[str, dict, DatasourceConfig]:
         """
         See parent 'Store.serialize()' for more information
         """
+        if isinstance(
+            value, FluentDatasource
+        ):  # TODO: this is gross, do something else (or do it somewhere else)
+            return value._json_dict()
         return self._serializer.serialize(value)
```
Kilo59: It seems natural to move this logic to the `serializer.serialize()` method, but the problem is that there's not an obvious way to move the `deserialization` away from this class as well.

Open to suggestions here.
I'd like to avoid changing too much at once if possible, my goal is to get this working so as to unblock the rest of Cloud FluentDatasources work, and followup with a better abstraction if it presents itself.

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -100,6 +100,10 @@ def _set(self, key: tuple[str, ...], value: Any, **kwargs: dict) -> None:
 
         if resource_type is DataContextVariableSchema.ALL_VARIABLES:
             config_commented_map_from_yaml = yaml.load(value)
+            # TODO: instead prevent these from being set??
+            # TODO: revert this before merge
+            for name in config_commented_map_from_yaml.get("fluent_datasources", {}):  # type: ignore[union-attr]
+                config_commented_map_from_yaml["datasources"].pop(name)  # type: ignore[union-attr]
```
Kilo59, line 106: When we call `self._datasource_store.set(key=None, value=datasource)` this causes the `fluent` datasources to be added to the `"datasources"` config entry, which will then be validated using the old "Datasource" marshmallow schemas when we call `context.variables.save_config()`

This temporary hack is just to demonstrate the problem. 

The cleanest solution might be to also `stash` these values or simply remove fluent datasource types from `datasources` during the `_fluent_objects_stash()` call.

https://github.com/great-expectations/great_expectations/blob/2367d978b85ebec76c7e9bd81b6e562b0dc7f9c3/great_expectations/data_context/data_context_variables.py#L364-L398

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -100,6 +100,10 @@ def _set(self, key: tuple[str, ...], value: Any, **kwargs: dict) -> None:
 
         if resource_type is DataContextVariableSchema.ALL_VARIABLES:
             config_commented_map_from_yaml = yaml.load(value)
+            # TODO: instead prevent these from being set??
+            # TODO: revert this before merge
+            for name in config_commented_map_from_yaml.get("fluent_datasources", {}):  # type: ignore[union-attr]
+                config_commented_map_from_yaml["datasources"].pop(name)  # type: ignore[union-attr,arg-type]
```
Kilo59: When we call `self._datasource_store.set(key=None, value=datasource)` this causes the `fluent` datasources to be added to the `"datasources"` config entry, which will then be validated using the old "Datasource" marshmallow schemas when we call `context.variables.save_config()`

This temporary hack is just to demonstrate the problem. 

The cleanest solution might be to also `stash` these values or simply remove fluent datasource types from `datasources` during the `_fluent_objects_stash()` call.

https://github.com/great-expectations/great_expectations/blob/2367d978b85ebec76c7e9bd81b6e562b0dc7f9c3/great_expectations/data_context/data_context_variables.py#L364-L398

reqs/requirements-dev-lite.txt
```diff
@@ -15,5 +15,6 @@ pytest-order>=0.9.5
 pytest-random-order>=1.0.4
 pytest-timeout>=2.1.0
 requirements-parser>=0.2.0
+responses>=0.23.1 # requests mocking
```
cdkini, line 18: If we go this route, can we make some noise so the team knows to use this moving forward?

great_expectations/data_context/store/datasource_store.py
cdkini: I think this is fine for now - as long as the branching is `FluentDatasource` vs `Datasource`, it doesn't feel like the worst violation. We should maybe add a ticket or brainstorm how to clean it up after the fact though

great_expectations/data_context/store/datasource_store.py
```diff
@@ -62,20 +64,34 @@ def remove_key(self, key: Union[DataContextVariableKey, GXCloudIdentifier]) -> N
...
         """
         See parent 'Store.deserialize()' for more information
         """
         # When using the InlineStoreBackend, objects are already converted to their respective config types.
-        if isinstance(value, DatasourceConfig):
+        if isinstance(value, (DatasourceConfig, FluentDatasource)):
             return value
         elif isinstance(value, dict):
+            # TODO: do this better... or elsewhere
+            type_ = value.get("type")
```
cdkini: Same note - if we use `type`, can we be clear that its the differentiating factor between `FluentDatasource` and `Datasource`

great_expectations/data_context/store/inline_store_backend.py
cdkini: this whole `InlineStoreBackend` class is quite messy tbh - I'm not sure what the stash flow would be like but we can discuss synchronously

great_expectations/data_context/store/datasource_store.py
```diff
@@ -62,20 +64,34 @@ def remove_key(self, key: Union[DataContextVariableKey, GXCloudIdentifier]) -> N
...
+            if type_:
+                datasource_model = _SourceFactories.type_lookup[type_]
+                return datasource_model(**value)
```
Kilo59, line 94: assert the type is valid

great_expectations/data_context/store/datasource_store.py
Kilo59: Yes.
Also added an assert here as discussed.

great_expectations/data_context/store/inline_store_backend.py
Kilo59: The stash strategy won't work without additional refactoring. May revisit it in a followup.

great_expectations/data_context/store/datasource_store.py
```diff
@@ -127,15 +149,19 @@ def delete(self, datasource_config: DatasourceConfig) -> None:
...
+        self, datasource_config: DatasourceConfig | FluentDatasource
     ) -> Union[GXCloudIdentifier, DataContextVariableKey]:
-        return self.store_backend.build_key(
-            name=datasource_config.name,
-            id=datasource_config.id,
-        )
+        if datasource_config.id:
+            id_: str | None = str(datasource_config.id)
+        else:
+            id_ = datasource_config.id  # type: ignore[assignment] # None or empty string
```
cdkini: Should the `id_` type declaration go above the conditional?

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -100,6 +100,10 @@ def _set(self, key: tuple[str, ...], value: Any, **kwargs: dict) -> None:
 
         if resource_type is DataContextVariableSchema.ALL_VARIABLES:
             config_commented_map_from_yaml = yaml.load(value)
+            # NOTE: fluent datasources may be present under both the `fluent_datasources` & `datasources` key
+            # if fluent datasource is part of `datasources` it will attempt to validate using a marshmallow Datasource schema and fail
+            for name in config_commented_map_from_yaml.get("fluent_datasources", {}):  # type: ignore[union-attr]
+                config_commented_map_from_yaml.get("datasources", {}).pop(name, None)  # type: ignore[union-attr,arg-type,call-arg]
```
cdkini, line 106: Do we have plans of removing this logic? Forgot where we landed

tests/datasource/fluent/test_viral_snippets.py
```diff
@@ -216,5 +222,40 @@ def test_add_and_save_fluent_datasource(
...
+):
+    context = empty_contexts
+
+    datasource_name = "save_ds_test"
+
+    context.sources.add_sqlite(
+        name=datasource_name, connection_string=f"sqlite:///{db_file}"
+    )
+
+    assert datasource_name in context.datasources
```
cdkini, line 237: Do we want to assert against the filesystem as well? We know the cache is updated but we want to make sure the store works as intended

great_expectations/data_context/store/inline_store_backend.py
Kilo59: Yes my goal is to remove this logic before resolving the ticket.

tests/datasource/fluent/test_viral_snippets.py
Kilo59: Yes, that's a good idea. I want to break that out into a separate test suite though since these tests are meant to be high level.

great_expectations/data_context/store/datasource_store.py
Kilo59: Done

tests/datasource/fluent/test_cloud.py
```diff
@@ -0,0 +1,85 @@
...
+    if isinstance(empty_contexts, CloudDataContext):
+        assert datasource.id
+        cloud_api_fake.assert_call_count(
+            f"{GX_CLOUD_MOCK_BASE_URL}/organizations/{FAKE_ORG_ID}/datasources",
+            1,
+        )
+    else:
+        yaml_path = pathlib.Path(empty_contexts.root_directory, empty_contexts.GX_YML)
+        assert yaml_path.exists()
+        assert datasource_name in yaml.load(yaml_path.read_text())["fluent_datasources"]  # type: ignore[operator]
```
Kilo59: @cdkini is there a better way to check that items have been added to the stores?
`empty_contexts.stores` was just an empty dict but I know that items are being added.

cdkini: we could use a spy or use `store.get`?

Kilo59: Yeah, will do one or both of those. Thanks

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -709,6 +709,11 @@ def _add_fluent_datasource(self, datasource: FluentDatasource) -> None:
                 f"Can not write the fluent datasource {datasource.name} because a datasource of that "
                 "name already exists in the data context."
             )
+        if self._datasource_store.cloud_mode:
+            set_datasource = self._datasource_store.set(key=None, value=datasource)
+            if set_datasource.id:
+                logger.debug(f"Assigning `id` to '{datasource.name}'")
+                datasource.id = set_datasource.id
```
Kilo59, line 720: Temporary workaround while we update `stores` to work better with Fluent Datasources.
Without this we end up with duplicate entries for fluent datasources in both `"fluent_datasources"` and `"datasources"` config/yaml entries.

cdkini: Could we add this context as a comment above the block?

tests/datasource/fluent/conftest.py
```diff
@@ -114,6 +152,176 @@ def inject_engine_lookup_double(
...
+
+@pytest.fixture
+def empty_cloud_context_fluent(cloud_api_fake) -> CloudDataContext:
+    context = gx.get_context(
+        cloud_access_token=DUMMY_JWT_TOKEN,
+        cloud_organization_id=FAKE_ORG_ID,
+        cloud_base_url=GX_CLOUD_MOCK_BASE_URL,
+        cloud_mode=True,
+    )
+    return context
```
cdkini, line 322: Can you check with @billdirks to make sure this doesn't overlap with his work? I know he was looking for a proper Cloud mock so maybe y'all can work together

tests/datasource/fluent/test_contexts.py
```diff
@@ -0,0 +1,103 @@
...
+            1,
+        )
+        cloud_api_fake.assert_call_count(
+            f"{GX_CLOUD_MOCK_BASE_URL}/organizations/{FAKE_ORG_ID}/datasources{datasource.id}",
+            1,
+        )
+
+
+if __name__ == "__main__":
+    pytest.main([__file__, "-vv"])
```
cdkini, line 103: What's the purpose of this in a test file?

cdkini (approved):

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59: Will do.

tests/datasource/fluent/test_contexts.py
Kilo59: I like to do it for 2 reasons.

1. It lets me run pytest for this test file by just running it with `python`. Which makes running the test from an IDE with `run python file` shortcut possible.
2. This is a very minor side benefit but it anchors the bottom of the file which makes git diffs much more readable.

great_expectations/data_context/data_context/abstract_data_context.py
billdirks: This seems to be syncing `datasource` with the returned `set_datasource`. It just happens that the only change is the id. Maybe this sync/copy method should live on datasource? This is fine for now, especially if it's changing soon per your comment.

great_expectations/data_context/store/datasource_store.py
```diff
@@ -62,20 +64,40 @@ def remove_key(self, key: Union[DataContextVariableKey, GXCloudIdentifier]) -> N
...
         # When using the InlineStoreBackend, objects are already converted to their respective config types.
-        if isinstance(value, DatasourceConfig):
+        if isinstance(value, (DatasourceConfig, FluentDatasource)):
             return value
         elif isinstance(value, dict):
+            # presence of a 'type' field means it's a fluent datasource
+            type_ = value.get("type")
+            if type_:
+                datasource_model = _SourceFactories.type_lookup.get(type_)
+                assert datasource_model, f"Unknown Datasource 'type': '{type_}'"
```
billdirks: I think raising may be preferable to an assert here.

great_expectations/data_context/store/datasource_store.py
```diff
@@ -127,16 +149,39 @@ def delete(self, datasource_config: DatasourceConfig) -> None:
...
     ) -> Union[GXCloudIdentifier, DataContextVariableKey]:
-        return self.store_backend.build_key(
-            name=datasource_config.name,
-            id=datasource_config.id,
-        )
+        id_: str | None
+        if datasource_config.id:
+            id_ = str(datasource_config.id)
+        else:
+            id_ = datasource_config.id  # type: ignore[assignment] # None or empty string
```
billdirks: `id_: str | None  = str(datasource_config.id) if datasource_config.id else datasource_config.id` may be more succinct.

great_expectations/data_context/store/inline_store_backend.py
```diff
@@ -100,6 +100,10 @@ def _set(self, key: tuple[str, ...], value: Any, **kwargs: dict) -> None:
 
         if resource_type is DataContextVariableSchema.ALL_VARIABLES:
             config_commented_map_from_yaml = yaml.load(value)
+            # NOTE: fluent datasources may be present under both the `fluent_datasources` & `datasources` key
```
billdirks, line 103: How do fluent datasources get added to the `datasources` key? Could we make this never happen?

great_expectations/util.py
```diff
@@ -1704,7 +1704,7 @@ def convert_ndarray_decimal_to_float_dtype(data: np.ndarray) -> np.ndarray:
 
 
 @overload
-def get_context(
+def get_context(  # type: ignore[misc]
```
billdirks: What's this ignore for?

tests/datasource/fluent/conftest.py
billdirks: Maybe I'll be able to use this. Currently, I'm mocking our cloud store calls (such as get) which wrap calls to cloud so I'm mocking above the response dependency.

tests/datasource/fluent/conftest.py
```diff
@@ -29,9 +46,30 @@
...
+    from requests import PreparedRequest
+
+    from great_expectations.data_context import CloudDataContext
+
+
 FLUENT_DATASOURCE_TEST_DIR: Final = pathlib.Path(__file__).parent
 PG_CONFIG_YAML_FILE: Final = FLUENT_DATASOURCE_TEST_DIR / FileDataContext.GX_YML
 
+
+GX_CLOUD_MOCK_BASE_URL: Final[str] = "https://app.greatexpectations.fake.io"
```
billdirks, line 60: My preference would be to move gx cloud specific fixtures and helpers into their own file. This isn't blocking though.

billdirks (commented):
Thanks for doing this. This looks good. I haven't looked at tests super closely. My only real outstanding question is how to fluent datasources get added to the `datasources` (nonfluent) key. The answer may be outside this PR but I'm curious.

great_expectations/data_context/store/datasource_store.py
Kilo59: I'll raise `LookupError` or a `TypeError` here

great_expectations/data_context/store/datasource_store.py
Kilo59: Will do

great_expectations/data_context/store/inline_store_backend.py
Kilo59: Because of the `self._datasource_store.set(key=None, value=datasource)` call that's needed to invoke the CloudAPI `DatasourceStore` webservice calls on a `CloudDataContext`

@cdkini and I are working on making this not a problem and then we can remove this but it may require some not insignificant changes to our `Stores`. But this unblocks the rest of the cloud work while we make those changes.

great_expectations/util.py
Kilo59: mypy thinks there are overlapping overloads for this method.
I'll fix it or remove it from this PR.

tests/datasource/fluent/conftest.py
Kilo59: Yeah if it's okay I'll make these changes in followup.

great_expectations/data_context/store/inline_store_backend.py
billdirks: Sounds good, thanks for the explanation.

billdirks (approved):
LGTM, thanks!

Kilo59:
This will create a `great_expectations.yml` testing artifact if this bugfix is not merged first.
- https://github.com/great-expectations/great_expectations/pull/7657

PR #7572, opened 2023-04-05
[MAINTENANCE] Add split/join logic to build_gallery process

kenwade4:
- Test results per backend go to their own files
    - Only "full runs" (that include EVERY Expectation) generate **`{backend}_full.json`** files (ONLY THESE will be considered when creating the combined JSON file)
    - Full runs with non-standard flags passed (i.e. `--ignore-suppress` and `--ignore-only-for`) will generate `{backend}_full_nonstandard.json`
    - Partial runs (passing in specific Expectation names or using `--no-core` / `--no-contrib`) will generate `{backend}_partial.json` or `{backend}_partial_nonstandard.json` files
- Before the combined JSON file is allowed to be created, ALL `*_full.json` MUST exist
- Update build_gallery CI pipeline to use a stage per backend
- Refactor build_gallery


assets/scripts/build_gallery.py
```diff
@@ -228,8 +255,25 @@ def build_gallery(
         None
 
     """
-    gallery_info = dict()
     requirements_dict = {}
+    backend_outfile_suffix = "partial"
+    if include_core and include_contrib and not only_these_expectations:
+        backend_outfile_suffix = "full"
+    if ignore_suppress or ignore_only_for:
+        backend_outfile_suffix += "_nonstandard"
```
anthonyburdi, line 464: Can we move some of this new code out to separate functions instead of including them in `build_gallery()`? This function is already a few hundred lines.
For example in this instance (also maybe the GalleryBuildOptions can be used elsewhere in the function as well):

```
@dataclass
class GalleryBuildOptions():
    include_core: bool
    include_contrib: bool
    only_these_expectations: bool
    ignore_suppress: bool
    ignore_only_for: bool

def _backend_outfile_suffix(options: GalleryBuildOptions) -> str:
    backend_outfile_suffix = "partial"
    if options.include_core and options.include_contrib and not options.only_these_expectations:
        backend_outfile_suffix = "full"
    if options.ignore_suppress or options.ignore_only_for:
        backend_outfile_suffix += "_nonstandard"
    return backend_outfile_suffix
```

assets/scripts/build_gallery.py
```diff
@@ -219,6 +248,7 @@ def build_gallery(
     Build the gallery object by running diagnostics for each Expectation and returning the resulting reports.
 
     Args:
+        only_combine: if true, only combine the various backend_full.json files
```
anthonyburdi, line 442: Could we refactor the combining out into a separate method instead of using a param here?

assets/scripts/build_gallery.py
```diff
@@ -596,8 +750,9 @@ def _disable_progress_bars() -> Tuple[str, DataContext]:
     "--outfile-name",
     "-o",
     "outfile_name",
-    default="expectation_library_v2--staging.json",
-    help="Name for the generated JSON file",
+    # default="expectation_library_v2--staging.json",
+    default="expectation_library_v2--PLAY.json",
```
anthonyburdi: Should this be reverted or is this intentional?

assets/scripts/build_gallery.py
```diff
@@ -618,11 +773,13 @@ def main(**kwargs):
...
-    gallery_info = build_gallery(
+    build_gallery(
+        only_combine=kwargs["only_combine"],
         include_core=not kwargs["no_core"],
         include_contrib=not kwargs["no_contrib"],
         ignore_suppress=kwargs["ignore_suppress"],
         ignore_only_for=kwargs["ignore_only_for"],
+        outfile_name=kwargs["outfile_name"],
         only_these_expectations=kwargs["args"],
         only_consider_these_backends=backends,
```
anthonyburdi, line 739: Is `backends` what we will use in the pipeline to generate the gallery data file for each individual backend?

assets/scripts/build_package_gallery.py
```diff
@@ -108,7 +108,9 @@ def write_results_to_disk(path: str, package_manifests: List[dict]) -> None:
             len(payload) > 0
         ), "Something went wrong; there should packages in the payload!"
         write_results_to_disk(
-            os.path.join(pwd, "./package_manifests--staging.json"), payload
+            # os.path.join(pwd, "./package_manifests--staging.json"), payload
+            os.path.join(pwd, "./package_manifests--PLAY.json"),
```
anthonyburdi: Is this intentional?

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
```diff
@@ -75,23 +75,116 @@ stages:
...
+
+          - script: |
+              pip install \
+                "cryptography==38.0.4" \
+                --requirement reqs/requirements-dev-all-contrib-expectations.txt \
+                --editable ".[spark]" \
+                --constraint constraints-dev.txt
+            displayName: 'Install dependencies'
+
+          - bash: python ./build_gallery.py --backends "spark, pandas" 2>&1 | tee output--build_gallery--dataframes.txt ; grep -o "Took .* seconds to .*" output--build_gallery--dataframes.txt | sort -k2,2nr > testing-times--dataframes.txt ; grep -o "ERROR - (.*" output--build_gallery--dataframes.txt | sort > testing-error-messages--dataframes.txt; touch gallery-tracebacks.txt; grep -o "Expectation type.*" output--build_gallery--dataframes.txt | sort > gallery-exp-types--dataframes.txt ; mv checklists.txt checklists--dataframes.txt ; mv gallery-tracebacks.txt gallery-tracebacks--dataframes.txt
```
anthonyburdi: Is there a benefit to doing multiple backends in a single job? Could we parametrize these instead and run each backend separately with the output file using the name of the backend? I think that would be clearer when looking at the output and probably faster as well since it can be parallelized.

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
```diff
@@ -162,76 +539,230 @@ stages:
...
+          - script: |
+              pip install \
+                "cryptography==38.0.4" \
+                --requirement reqs/requirements-dev-all-contrib-expectations.txt \
+                --editable ".[spark, redshift, athena]" \
+                --constraint constraints-dev.txt
+            displayName: 'Install dependencies'
+
+            #
+            # Update to `--backends "redshift, athena"` when athena testing is real
```
anthonyburdi: How about we make a separate redshift and athena job instead of trying to merge them into aws?

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
anthonyburdi: You would also be able to parametrize the diagnostics so they are not repeated multiple times in this file.

anthonyburdi (changes requested):
Glad to see this coming along! I have some requests, happy to pair and discuss.

assets/scripts/build_gallery.py
kenwade4: Intentional. In the PR description, there is a **BEFORE MERGE** checklist that is to revert this. Just don't want to accidentally overwrite anything.

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
kenwade4: @anthonyburdi I think we should parametrize every backend while also making sure the `pip install` step for each ONLY adds the drivers needed for that backend (not all dev requirements like what was happening before).

I discovered a bug with a handful of contrib Expectations that try to do spark things even if spark isn't installed and will fail because of it. That's part of why I had to re-add spark to all those.

assets/scripts/build_gallery.py
kenwade4: Would be nice, but in order to do the combine, it has to iterate over every Expectation to get all that metadata (and the in-memory objects for the Expectation instance and diagnostics object). The Expectation instance and some attributes of the diagnostic object need to get passed to a static method to recalculate the maturity checklist and final maturity status.

Since it has to do this one final run, we do sqllite here (so sqlite wouldn't be included in the parametrized calls that will get pushed later).

When a CI stage writes whatever _full.json file, it could have written some metadata.json file that got uploaded (and re-uploaded) to S3, but those in-memory objects are needed anyway, so no point in a metadata.json

assets/scripts/build_gallery.py
kenwade4: The `backends` are the backends to attempt. Also good for local testing all or a handful of Expectations on some backends`--backends "pandas, spark, sqlite" expect_column_max_to_be_between expect_column_min_to_be_between` 

It's what's passed into [self_check.util.generate_expectation_tests](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/self_check/util.py#L1830) func as `only_consider_these_backends`

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
kenwade4: @anthonyburdi decided not to parametrize since the differences between the stages are significant enough. But all backends are in their own stages now.

assets/scripts/build_gallery.py
anthonyburdi: I'm not sure I'm following here - why couldn't the final merge be a separate function like `merge_expectation_metadata_for_gallery()`? `build_gallery()` already has too many responsibilities.

assets/scripts/download_json_from_s3.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
anthonyburdi, line 1: Let's keep these files together in `docs/expectation_gallery/`. Trying to move away from `scripts` folders.

assets/scripts/build_gallery.py
```diff
@@ -163,6 +214,148 @@ def get_expectation_file_info_dict(
...
+    parsed_requirements = pkg_resources.parse_requirements(requirements)
+    installed = []
+    for req in parsed_requirements:
+        is_satisfied = any(
+            [installed_pkg in req for installed_pkg in installed_packages]
+        )
+        if not is_satisfied:
+            logger.debug(f"Executing command: 'pip install \"{req}\"'")
+            status_code = execute_shell_command(f'pip install "{req}"')
+            if status_code == 0:
```
anthonyburdi, line 256: Do we need to do any handling here if the install fails?

assets/scripts/build_gallery.py
```diff
@@ -163,6 +214,148 @@ def get_expectation_file_info_dict(
...
+    )
+    for req in requirements:
+        logger.debug(f"Executing command: 'pip uninstall -y \"{req}\"'")
+        execute_shell_command(f'pip uninstall -y "{req}"')
+
+
+def get_expectation_instances(expectations_info):
+    """Return a dict of Expectation instances
+
+    Any contrib Expectations must be imported so they appear in the Expectation registry
```
anthonyburdi, line 276: Could you add an example here of what the arg `expectations_info` should be and also an example of the return value?

assets/scripts/build_gallery.py
```diff
@@ -79,7 +104,7 @@ def execute_shell_command(command: str) -> int:
     return status_code
 
 
-def get_expectation_file_info_dict(
+def get_expectations_info_dict(
```
anthonyburdi, line 107: Could you add a docstring here with an example and description of params + return value?

assets/scripts/build_gallery.py
```diff
@@ -117,12 +144,33 @@ def get_expectation_file_info_dict(
 
     for file_path in sorted(files_found):
         file_path = file_path.replace(f"{repo_path}{os.path.sep}", "")
-        package_name = os.path.basename(os.path.dirname(os.path.dirname(file_path)))
-        if package_name == "expectations":
-            package_name = "core"
         name = os.path.basename(file_path).replace(".py", "")
```
anthonyburdi: Just a small suggestion for readability.
```suggestion
        expectation_name = os.path.basename(file_path).replace(".py", "")
```

assets/scripts/build_gallery.py
```diff
@@ -163,6 +214,148 @@ def get_expectation_file_info_dict(
...
+
+        expectation_instances[
+            expectation
+        ] = great_expectations.expectations.registry.get_expectation_impl(expectation)()
+    return expectation_instances
+
+
+def combine_backend_results(
+    expectations_info, expectation_instances, diagnostic_objects, outfile_name
+):
```
anthonyburdi, line 308: Would you mind adding a docstring with args descriptions / examples?

assets/scripts/build_gallery.py
```diff
@@ -228,132 +424,67 @@ def build_gallery(
...
                 only_consider_these_backends=only_consider_these_backends,
                 context=context,
             )
+            diagnostic_objects[expectation] = diagnostics
             checklist_string = diagnostics.generate_checklist()
             expectation_checklists.write(
-                f"\n\n----------------\n{expectation} ({group})\n"
+                f"\n\n----------------\n{expectation} ({expectations_info[expectation]['package']})\n"
             )
             expectation_checklists.write(f"{checklist_string}\n")
```
anthonyburdi, line 507: Could `expectation_checklists` be defined closer to the point of use? It would help readability.

anthonyburdi: What do you think about factoring this out (from `diagnostic_objects[expectation] = diagnostics` to the `except:`) into a separate method? `write_expectation_diagnostics()` or something similar? Then could this step and the `run_diagnostics()` step be skipped when only combining all the results from the different backends?

assets/scripts/build_gallery.py
```diff
@@ -365,113 +496,59 @@ def build_gallery(
...
-        for expectation in sorted(failed_to_import_set):
-            expectation_tracebacks.write(f"- {expectation}\n")
-    if non_matched_filenames:
-        expectation_tracebacks.write(f"\n\n----------------\n(Not a traceback)\n")
-        expectation_tracebacks.write(
-            "Expectation filenames that don't match their defined Expectation name:\n"
-        )
-        for fname in sorted(non_matched_filenames):
-            expectation_tracebacks.write(f"- {fname}\n")
+    # Iterate the gallery_info_by_backend dict and write the backend-specific files
```
anthonyburdi, line 566: This is great, so one or many backends can be passed in.

assets/scripts/build_gallery.py
```diff
@@ -365,113 +496,59 @@ def build_gallery(
...
+            with open(f"{_backend}_{backend_outfile_suffix}.json", "w") as outfile:
+                json.dump(gallery_info_by_backend[_backend], outfile, indent=4)
 
-        bad_names = sorted(
-            list(registered_expectations_set - expectation_filenames_set)
-        )
-        expectation_tracebacks.write(
-            f"\nRegistered Expectation names that don't match:\n"
+    # Only attempt to combine and write to file when no Expectations are skipped
+    if backend_outfile_suffix == "full":
```
anthonyburdi, line 573: Just trying to understand - is this not checking the `only_combine` flag because we are also combining on each run of build_gallery()?

assets/scripts/build_gallery.py
```diff
@@ -163,6 +214,148 @@ def get_expectation_file_info_dict(
...
+                continue
+            maturity_checklist_object = Expectation._get_maturity_checklist(
+                expectation_instance=expectation_instance,
+                library_metadata=diagnostic_object.library_metadata,
+                description=diagnostic_object.description,
+                examples=diagnostic_object.examples,
+                tests=diagnostic_object.tests,
+                backend_test_result_counts=backend_test_result_counts_object,
+                execution_engines=diagnostic_object.execution_engines,
+            )
```
anthonyburdi: Could we use the instance to get the maturity checklist since we already have access to it? Then we can eliminate some of the changes in `expectation.py`:
```suggestion
            maturity_checklist_object = expectation_instance._get_maturity_checklist(
                library_metadata=diagnostic_object.library_metadata,
                description=diagnostic_object.description,
                examples=diagnostic_object.examples,
                tests=diagnostic_object.tests,
                backend_test_result_counts=backend_test_result_counts_object,
                execution_engines=diagnostic_object.execution_engines,
            )
```

great_expectations/expectations/expectation.py
```diff
@@ -2188,8 +2172,9 @@ def _get_augmented_library_metadata(self):
         augmented_library_metadata["problems"] = problems
         return AugmentedLibraryMetadata.from_legacy_dict(augmented_library_metadata)
 
+    @staticmethod
     def _get_maturity_checklist(
-        self,
+        expectation_instance,
```
anthonyburdi: See note in `build_gallery.py` about potentially reverting this to `self` and calling these methods on the instance itself.

great_expectations/expectations/expectation.py
```diff
@@ -2239,6 +2228,47 @@ def _get_maturity_checklist(
...
+        all_beta = all([check.passed for check in maturity_checklist.beta])
+        all_production = all([check.passed for check in maturity_checklist.production])
+        if all_production and all_beta and all_experimental:
+            maturity = Maturity.PRODUCTION
+        elif all_beta and all_experimental:
+            maturity = Maturity.BETA
+        else:
+            maturity = Maturity.EXPERIMENTAL
+
+        return maturity
```
anthonyburdi: What do you think about changing these to module level methods?

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
```diff
@@ -75,24 +75,126 @@ stages:
...
+            inputs:
+              versionSpec: '$(python.version)'
+            displayName: 'Use Python $(python.version)'
+
+          - bash: python -m pip install --upgrade pip
+            displayName: 'Update pip'
+
+          - script: |
+              pip install \
+                "cryptography==38.0.4" \
```
anthonyburdi: Do we need this pin (for this and all other jobs)? Hopefully we can remove it, especially now that the jobs are split up.

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
```diff
@@ -173,65 +601,164 @@ stages:
...
+              AWS_DEFAULT_REGION: $(AWS_DEFAULT_REGION)
 
-          - bash: cat checklists.txt
+          - bash: for fname in $(ls -1 *_full.json); do echo -e "\n\n==========\n$fname"; cat "$fname"; done
             workingDirectory: $(Build.SourcesDirectory)/assets/scripts/
-            displayName: 'Show full checklist summary'
+            displayName: 'cat *_full.json files in assets/scripts'
 
-          - bash: grep -E "(^expect|Completeness checklist|^ *[A-z]|^ *-|-----)" checklists.txt | grep -vE '(No validate_configuration|Using default validate_configuration|Has a full suite|Has passed a manual)'
+          - bash: python ./build_gallery.py --backends "sqlite" --only-combine --outfile-name "expectation_library_v2--staging.json" 2>&1 | tee output--build_gallery.txt ; grep -o "ERROR - (.*" output--build_gallery.txt | sort > testing-error-messages.txt ; grep -o "Expectation type.*" output--build_gallery.txt | sort > gallery-exp-types.txt ; touch gallery-tracebacks.txt
```
anthonyburdi: Does there need to be a backend specified here for the combine step or can it be omitted?

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
```diff
@@ -243,6 +770,70 @@ stages:
...
               filesAcl: 'public-read'
 
+
+# Don't do build_package_gallery.py stuff...
+#   - Currently, this script will call the run_diagnostics method on every Expectation
+#     that exists in each package for all backends (even though this was already done
+#     by the build_gallery.py script)
+#   - The build_gallery.py script is now package aware
+#   - The Expectation Gallery site can now accept a URL that has a filter applied to
+#     limit results to a specific package, so no need to bundle them in the "packages" tab
```
anthonyburdi, line 769: Should this commented out code and `build_package_gallery.py` be removed based on this comment?

anthonyburdi (commented):
Thank you for the refactoring! I have some additional comments, happy to discuss if it would be helpful.

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
kenwade4: Just for clarity so someone knows it's happening. If `--backends` isn't specified, it's still gonna do sqlite-only (the fastest way to get an [ExpectationDiagnostics](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/core/expectation_diagnostics/expectation_diagnostics.py#L32) object for each Expectation).

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
kenwade4: Dunno, it was needed in all the pipelines several months ago because of some issue.

assets/scripts/build_gallery.py
kenwade4: No the traceback details end up in the gallery-tracebacks.txt file that gets shown in the pipeline. Later in the process when that Expectation attempts to run_diagnostics, it will also have a traceback for not being able to import whatever thing that failed to install.

Most things are included in the venv already via https://github.com/great-expectations/great_expectations/blob/develop/reqs/requirements-dev-all-contrib-expectations.txt

assets/scripts/build_gallery.py
kenwade4: In the refactor, this [combine_backend_results](https://github.com/great-expectations/great_expectations/blob/misc/gallery-split-join-logic/assets/scripts/build_gallery.py#L280) func was added, but it needs to be passed in ExpectationDiagnostic objects and instances of the Expectations so the maturity_checklist, library_metadata, and coverage_score can be recalculated. The files being combined don't have a ton of info in them (snippets below).

Also see this comment https://github.com/great-expectations/great_expectations/pull/7572#discussion_r1172437866

https://superconductive-public.s3.us-east-2.amazonaws.com/static/gallery/pandas_full.json

```
{
    "expect_batch_row_count_to_match_prophet_date_model": {
        "backend_test_result_counts": [
            {
                "backend": "pandas",
                "num_passed": 2,
                "num_failed": 0,
                "failing_names": null
            }
        ]
    },
    "expect_column_average_lat_lon_pairwise_distance_to_be_less_than": {
        "backend_test_result_counts": [
            {
                "backend": "pandas",
                "num_passed": 4,
                "num_failed": 0,
                "failing_names": null
            }
        ]
    },
    "expect_column_average_to_be_within_range_of_given_point": {
        "backend_test_result_counts": [
            {
                "backend": "pandas",
                "num_passed": 2,
                "num_failed": 0,
                "failing_names": null
            }
        ]
    },
    ...
}
```

https://superconductive-public.s3.us-east-2.amazonaws.com/static/gallery/spark_full.json

```
{
    "expect_column_discrete_entropy_to_be_between": {
        "backend_test_result_counts": [
            {
                "backend": "spark",
                "num_passed": 4,
                "num_failed": 0,
                "failing_names": null
            }
        ]
    },
    "expect_column_distinct_values_to_be_in_set": {
        "backend_test_result_counts": [
            {
                "backend": "spark",
                "num_passed": 10,
                "num_failed": 0,
                "failing_names": null
            }
        ]
    },
    "expect_column_distinct_values_to_contain_set": {
        "backend_test_result_counts": [
            {
                "backend": "spark",
                "num_passed": 15,
                "num_failed": 0,
                "failing_names": null
            }
        ]
    },
    ...
}
```

assets/scripts/build_gallery.py
kenwade4: I'll change it to "expectation" to be consistent with how it's called in other places.

assets/scripts/download_json_from_s3.sh
kenwade4: As asked in a different PR comment a week or 2 ago, why was the docs/expectation_gallery/ directory even created, just to move the 2 gallery yml files into? That directory shouldn't exist and those 2 files should be in the ci/ directory with everything else. The `azure-pipelines-docs-integration.yml` and `azure-pipelines-docs.yml` files (which are actually docs related) are in the new ci/ directory instead of some subdirectory of docs/

Why are we moving away from the scripts/ and assets/scripts/ directories? Coincidentally, everything in assets/scripts/ (aside from a cloud-onboarding script added 2 weeks ago) is gallery related.

I'd be ok moving all that stuff to the docs/expectation_gallery/ directory, but moving the 2 yml files from there to the ci/ directory. 

Has it been decided where all the other things in these various scripts directories will be moved to?

assets/scripts/build_gallery.py
kenwade4: No, it makes sense for those 3 StringIO instances to be globals defined together.

kenwade4: How so? Sounds interesting.

kenwade4: Might be worth a separate follow-up ticket so it's not lumped in with all these changes.

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
kenwade4: You should come to the Tuesday meeting next week. In a soon-ish later ticket, I'm going to make that thing not run any diagnostics or include the expectations in the package_manifests.json file.

For now, we just want whatever existing old package_manifests.json file on s3 to stay there, since the change will need to be coordinated with some frontend changes.

After some frontend changes that haven't been merged yet, we can link to the Expectations page with a specific filter applied (say for a particular package). https://deploy-preview-230--staging-great-expectations.netlify.app/expectations/?filterType=Package&viewType=Summary&showFilters=true&subFilterValues=capitalone_dataprofiler_expectations

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
anthonyburdi: We shouldn't pin this here. If there are expectations that require a specific version of a library that should be defined in the expectation itself. Pinning here would give us a false sense of security that the expectation would work, but most likely the user's environment would have the latest version of `cryptography`. The expectation should fail the test and need to be updated. Also pinning this specific package that is intimately related to security is something we should avoid. 

As an aside - and not relevant in this case for expectation tests but worth mentioning - in general, if there is a specific version that is causing an issue, we should use the `!=` modifier to avoid that particular version but not future versions.

assets/scripts/build_gallery.py
anthonyburdi: Ok - that makes sense, thank you for clarifying. Could gathering the ExpectationDiagnostic objects and instances of the Expectations be called also though in a `merge_expectation_metadata_for_gallery()` function? There already is a `get_expectation_instances()` function.

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
anthonyburdi: I disagree, but this is something we can address later.

assets/scripts/build_gallery.py
anthonyburdi: @kenwade4 - just making sure you saw this suggestion.

assets/scripts/download_json_from_s3.sh
anthonyburdi: The short answer is we need to do more reorganization, I wasn't able to do it all at once. `scripts` and `util` and `helper` directories and files are an anti-pattern, we should strive to organize our code better and name it according to what it is. Keeping code related to the expectation gallery in a folder called `expectation_gallery` makes it easier to find. Although the two `yml` files are run via our azure CI pipeline, they are building documentation so they should be organized with the documentation.

assets/scripts/build_gallery.py
anthonyburdi: I think that is confusing. You also have the expectation implementation as well. This variable is referring to a string that represents the name of the expectation. So this should be called `expectation_name` and the expectation implementation should be called `expectation`

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
anthonyburdi: Ok - but if we are throwing out the test results, couldn't we just create the `ExpectationDiagnostics` objects directly rather than re-running tests? I.e. with an empty list for `backend_test_result_counts` for example.

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
anthonyburdi: I don't understand - are you saying that this code will be re-enabled at that point?

assets/scripts/build_gallery.py
anthonyburdi: I'm trying to figure out ways to separate this large function up so that we can separate responsibilities better. Something like a `build_gallery_data_for_backend()` and a `merge_gallery_data()`. Pulling out code like `write_expectation_diagnostics()` helps me understand what is actually happening at each stage and makes it easier to make sure the right steps are happening. The reason I'm mentioning it here is because there is significant modification to this part of the code already.

assets/scripts/build_gallery.py
kenwade4: get_expectation_instances is the only thing that gets the instances and it's called one time. What is wrong with passing in the result to combine_backend_results when the time is right?

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
kenwade4: Some require services and extra scripts, some extra environment variables. Not sure it's worth the effort https://learn.microsoft.com/en-us/azure/devops/pipelines/process/runtime-parameters?view=azure-devops&tabs=script

assets/scripts/build_gallery.py
kenwade4: The Expectation implementation is not saved anywhere. An instance is immediately created and added to the dict via `great_expectations.expectations.registry.get_expectation_impl(expectation)()`

I renamed expectation to expectation_name and instance to expectation_instance.

great_expectations/expectations/expectation.py
kenwade4: Why? They should probably stay with all these other hidden methods that are being used by run_diagnostics.

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
kenwade4: It was for a now resolved snowflake issue https://github.com/great-expectations/great_expectations/pull/6755

Removed.

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
kenwade4: Re-enabled once it doesn't calculate the expectation_diagnostics or include expectations in it's output JSON. There is still all that unchanging metadata about the packages in the package_info.yml files (like https://github.com/great-expectations/great_expectations/blob/develop/contrib/capitalone_dataprofiler_expectations/package_info.yml).

docs/expectation_gallery/azure-pipelines-expectation-gallery.yml
kenwade4: The test results aren't getting tossed out. This is the time that the [sqlite_full.json](https://superconductive-public.s3.us-east-2.amazonaws.com/static/gallery/sqlite_full.json) file is generated.

I don't think there is a way to get a diagnostics object without running tests on at least one backend.

assets/scripts/build_gallery.py
kenwade4: The definition of acceptable build-gallery options are defined in the [`click` decorators on the main func](https://github.com/great-expectations/great_expectations/blob/develop/assets/scripts/build_gallery.py#L562-L608)

Also after the refactor, these args aren't sloppily all over the place in various conditional statements, so there should be way less confusion.

anthonyburdi (approved):
Most pressing issues addressed in https://github.com/great-expectations/great_expectations/pull/7572/commits/50e5db415dd2af00bb50d7912c03d351442007e6, we can take any followups in future PRs.

PR #7580, opened 2023-04-06
[DOCS] Update some docs_rtd requirements so the venv can be created successfully

kenwade4:
- Update some docs_rtd requirements so the venv can be created successfully


anthonyburdi (approved):
Thank you for doing this!

Shinnnyshinshin (approved):
🚀

PR #7587, opened 2023-04-07
[MAINTENANCE] Remove pip pins in CI and in contributing_setup.md

anthonyburdi:
- We had pinned pip in our pipelines due to some issues resolving dependencies with earlier versions of pip. These have since been fixed and we should remove these pins.
- Additionally this removes the suggestion to pin in our contributor guidelines. Our contributor guidelines are not yet under test, so this will need to be tested manually.

### Definition of Done

### DIFF
--- a/docs/docusaurus/docs/contributing/contributing_setup.md
+++ b/docs/docusaurus/docs/contributing/contributing_setup.md
@@ -56,8 +56,6 @@ python3 -m venv gx_dev
 
 source gx_dev/bin/activate
 
-pip install --upgrade pip==21.3.1
-
 pip install -c constraints-dev.txt -e ".[test]"
 ```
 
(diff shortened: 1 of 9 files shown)


docs/docusaurus/docs/contributing/contributing_setup.md
```diff
@@ -56,8 +56,6 @@ python3 -m venv gx_dev
 
 source gx_dev/bin/activate
```
billdirks, line 57: Do we want to leave the pip upgrade command here without the pin?

billdirks (approved):
I left 1 small comment but otherwise LGTM!

docs/docusaurus/docs/contributing/contributing_setup.md
anthonyburdi: I've been going back and forth on this since pip will show a warning if you are not on the latest version. But I think better to upgrade than not. I'll put the command back in (without the pin).

PR #7604, opened 2023-04-10
[MAINTENANCE] Polish and ratchet requirements pins and upper bounds

anthonyburdi:
- Fix bug in `test_requirements_files` where additional specs weren't being considered.
- Add a new `test_polish_and_ratchet_pins_and_upper_bounds` to create more visibility around what we have pinned or have set upper bounds on. We can use this as more of a report to help reduce our pins and upper bounds for packages.

### Definition of Done


alexsherstinsky (approved):
LGTM -- thanks!

tests/test_packaging.py
```diff
@@ -15,21 +18,57 @@ def collect_requirements_files() -> List[pathlib.Path]:
...
+    dict of the form {"filename": {"package_name_with_specs_as_str"}}
+
+    Args:
+        files: Paths to files that should be parsed
+    """
+
+    req_set_dict = {}
+    for req_file in files:
+        abs_path = req_file.absolute().as_posix()
+        key = abs_path.rsplit(os.path.sep, 1)[-1]
```
Kilo59, line 56: If I understand correctly, this can be simpler using `pathlib.Path` `.stem` or `.name` attributes.
You should never need to use `os.path` if you already have a `Path` object.

https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name
```python
    for req_file in files:
        key = req_file.name
```

or...

```python
    for req_file in files:
        with open(req_file) as f:
            req_set_dict[req_file.name] = {
                line.name: line.specs for line in rp.parse(f) if line.specs
            }

```

anthonyburdi: Good call, I just used the existing logic from `test_requirements_files()` with minor mods. This would be a good improvement.

PR #7607, opened 2023-04-11
[MAINTENANCE] Persist connection for backends that need a connection to keep temporary tables alive

anthonyburdi:
- Main changes are in: great_expectations/execution_engine/sqlalchemy_execution_engine.py where:
  - If the database drops temporary tables if the connection is closed (sqlite, mssql) so we have to keep the connection alive:
    - We use a `StaticPool` which holds a single connection 
    - We store a reference the connection on the execution engine so that the connection isn't closed automatically
    - We run queries from the execution engine instead of grabbing the sqlalchemy engine and creating our own connection (so management of the engine lifecycle can be handled by the execution engine - in a followup, this can be moved to a separate class to reduce the responsibilities of the execution engine).
- Most of the rest of the changes are using the new `execution_engine.execute_query()` and `execute_query_in_transaction()` methods.


great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -275,6 +276,10 @@ def _get_dialect_type_module(dialect):
     return dialect
 
 
+def _dialects_requiring_persisted_connection() -> tuple[GXSqlDialect, ...]:
```
billdirks: If you want to avoid the function call you could make this a private constant:
`_PERSISTED_CONNECTION_DIALECTS = (GXSqlDialect.SQLITE, GXSqlDialect.MSSQL)`

No real preference.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1318,3 +1322,38 @@ def get_batch_data_and_markers(
...
+            query:
+
+        Returns:
+
+        """
+
+        # TODO: Add docstring
+
+        if self.dialect_name in _dialects_requiring_persisted_connection():
+            # sqlite/mssql temp tables only persist within a connection,
```
billdirks: nit: you can remove `sqlite/mssql` and refer to the function that returns these so this comment doesn't go out of date.

billdirks (commented):
This strategy looks good to me!

great_expectations/execution_engine/sqlalchemy_execution_engine.py
Kilo59: The function call just seems like unnecessary indirection.
I would go ahead and use the constant.

All else being equal, python function calls are expensive. Although less so in recent versions.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1329,3 +1324,38 @@ def get_batch_data_and_markers(
...
+        """Close any open connections on shutdown."""
+        if self._connection:
+            self._connection.close()
+
+    # TODO: What is return type, docstring
+    def execute_query(self, query: Selectable):
+        """Execute a query using the underlying database engine.
+
+        Some databases sqlite/mssql temp tables only persist within a connection,
+        so we need to keep the connection alive.
```
Kilo59: @anthonyburdi
I think this is fine.
But just to throw out some more ideas... you could still do this as a context manager function.
And conditionally close the connection on exit if it's not `GXSqlDialect.SQLITE` or `GXSqlDialect.MSSQL`

And/or you could move that control to the caller so it's more explicit and just make this `excecute_query(query: Selectable, close_connection_on_exit: bool)`

Although returning the result might be more awkward there.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: Good call, changed to a constant.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: Good call, changed to be more generic.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: @Kilo59 I gave this a try in the most recent commit. I needed to keep the `__del__` to make sure that the connection is closed when the instance of the engine is removed, but in the test I need to call this via `execution_engine.__del__()` and not via `del execution_engine` which is surprising.

Open to any suggestions here.

anthonyburdi: I'm also considering trying to use a [StaticPool](https://docs.sqlalchemy.org/en/20/core/pooling.html#sqlalchemy.pool.StaticPool) instead for databases that need to use the same connection each time.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -96,7 +95,6 @@
 
 
 if sa:
-    sqlalchemy_version_check(sa.__version__)
```
anthonyburdi, line 99: This is done in `compatibility/sqlalchemy.py`

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -394,10 +442,6 @@ def _on_connect(dbapi_con, connection_record):
                 # sqlite3.Connection (distinct from a sqlalchemy Connection).
                 _add_sqlite_functions(self.engine.raw_connection())
             self._engine_backup = self.engine
-            # sqlite/mssql temp tables only persist within a connection so override the engine
-            # but only do this if self.engine is an Engine and isn't a Connection
-            if sqlalchemy.Engine and isinstance(self.engine, sqlalchemy.Engine):
-                self.engine = self.engine.connect()
```
anthonyburdi, line 400: Excited to remove this line :)

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1072,7 +1163,8 @@ def close(self) -> None:
         More background can be found here: https://github.com/great-expectations/great_expectations/pull/3104/
         """
         if self._engine_backup:
-            self.engine.close()
+            if self._connection:
```
anthonyburdi, line 1166: I think we can get rid of this method - in a followup PR since this PR is already quite large (see method docstring).

great_expectations/expectations/metrics/column_map_metrics/column_values_unique.py
```diff
@@ -62,14 +63,7 @@ def _sqlalchemy_window(cls, column, _table, **kwargs):
...
             )
-            if sqlalchemy.Engine and isinstance(sql_engine, sqlalchemy.Engine):
-                with sql_engine.connect() as connection:
-                    with connection.begin():
-                        connection.execute(sa.text(temp_table_stmt))
-            else:
-                # sql_engine is a connection
-                with sql_engine.begin():
-                    sql_engine.execute(sa.text(temp_table_stmt))
+            execution_engine.execute_query_in_transaction(sa.text(temp_table_stmt))
```
anthonyburdi, line 66: This pattern can be extended to other metrics.

great_expectations/expectations/metrics/map_metric_provider/column_condition_partial.py
```diff
@@ -189,6 +189,7 @@ def inner_func(
                     **metric_value_kwargs,
                     _dialect=dialect,
                     _table=selectable,
+                    _execution_engine=execution_engine,
```
anthonyburdi, line 192: This pattern can be extended to other condition partials (passing the execution engine object).

great_expectations/expectations/metrics/map_metric_provider/map_condition_auxilliary_methods.py
```diff
@@ -330,31 +330,29 @@ def _sqlalchemy_map_condition_unexpected_count_value(
...
-                        "condition", sa.Integer, primary_key=False, nullable=False
-                    ),
-                )
-                temp_table_obj.create(execution_engine.engine, checkfirst=True)
-
-                inner_case_query: sqlalchemy.Insert = (
-                    temp_table_obj.insert().from_select(
-                        [count_case_statement],
-                        count_selectable,
+            with execution_engine.get_connection() as connection:
```
anthonyburdi, line 333: Thank you @Shinnnyshinshin and @alexsherstinsky!! This issue was hard to find!!

great_expectations/self_check/util.py
```diff
@@ -1029,24 +1028,41 @@ def build_sa_validator_with_data(  # noqa: C901 - 39
...
-        dtype=sql_dtypes,
-        if_exists="replace",
-        method=sql_insert_method,
-    )
-    _end = time.time()
-    _debug(
-        f"Took {_end - _start} seconds to df.to_sql for {sa_engine_name} {extra_debug_info}"
+    execution_engine = SqlAlchemyExecutionEngine(caching=caching, engine=engine)
+    batch_data = SqlAlchemyBatchData(
+        execution_engine=execution_engine, table_name=table_name
```
anthonyburdi, line 1033: Passing execution engine and not engine to batch_data.

great_expectations/self_check/util.py
```diff
@@ -1029,24 +1028,41 @@ def build_sa_validator_with_data(  # noqa: C901 - 39
...
     )
+    with execution_engine.get_connection() as connection:
```
anthonyburdi, line 1035: Use the same connection (for sqlite and mssql) when populating the database.

tests/conftest.py
```diff
@@ -799,11 +800,9 @@ def postgresql_engine(test_backend):
             import sqlalchemy as sa
 
             db_hostname = os.getenv("GE_TEST_LOCAL_DB_HOSTNAME", "localhost")
-            engine = sa.create_engine(
-                f"postgresql://postgres@{db_hostname}/test_ci"
-            ).connect()
```
anthonyburdi, line 805: Now yields the engine (as per the fixture name), not the connection.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -223,6 +221,52 @@ def _get_dialect_type_module(dialect):
...
+    dialect_name isn't available yet since the engine isn't yet created when we call this method
+    so we determine the dialect from the creds/url/params:
+
+    Args:
+        connection_string:
+        credentials:
+        url:
+
+    Returns:
+
```
TrangPham: Can you add the return  and param descriptions to the docstring?

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -223,6 +221,52 @@ def _get_dialect_type_module(dialect):
...
+    """
+    if sum(bool(x) for x in [connection_string, credentials, url is not None]) != 1:
```
TrangPham, line 245: Is the URL allowed to be an empty string?

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1293,3 +1379,70 @@ def get_batch_data_and_markers(
...
+        then we get errors like sqlite3.ProgrammingError: Cannot operate on a closed database.
+
+        Returns:
+            Sqlalchemy connection
+        """
+        if self.dialect_name in _PERSISTED_CONNECTION_DIALECTS:
+            try:
+                if not self._connection:
+                    self._connection = self.engine.connect()
+                yield self._connection
```
TrangPham, line 1400: When/where does the connection get closed?

great_expectations/expectations/metrics/map_metric_provider/map_condition_auxilliary_methods.py
```diff
@@ -330,31 +330,29 @@ def _sqlalchemy_map_condition_unexpected_count_value(
 
     try:
         if execution_engine.engine.dialect.name.lower() == GXSqlDialect.MSSQL:
```
TrangPham: Should this be updated to `execution_engine.dialect_name`?

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: Good question! I think we can make future improvements here to close the connection deliberately, but to keep the PR scope here more limited (to enabling compatibility with sqlalchemy & pandas 2.0+) we are continuing with the existing method of the connection closing during garbage collection (for connections that are persisted i.e. `sqlite` and `mssql`). 
For future improvements, I think closing the connection during the checkpoint validation workflow is more straightforward than the interactive `validator` workflow. There are also different types of temporary tables in mssql that we may want to investigate that may be less ephemeral. Happy to discuss further!

great_expectations/expectations/metrics/map_metric_provider/map_condition_auxilliary_methods.py
anthonyburdi: Good catch! Changing.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
anthonyburdi: I think it can be, but it will either error here if multiple params are passed or later when trying to parse it. I added some additional tests. Let me know if that works!

contrib/capitalone_dataprofiler_expectations/capitalone_dataprofiler_expectations/tests/expectations/metrics/test_core.py
```diff
@@ -214,7 +214,7 @@ def test_data_profiler_column_list_metric_same_as_profile_report_pd(
     table_columns_metric: MetricConfiguration
     results: dict
 
-    table_columns_metric, results = get_table_columns_metric(engine=engine)
+    table_columns_metric, results = get_table_columns_metric(execution_engine=engine)
```
Shinnnyshinshin, line 217: I must say, I really do like this more-explicit name change. I do hope it the additional distinction between sa.engine and execution engine will continue to pay us dividends.

great_expectations/compatibility/sqlalchemy.py
```diff
@@ -31,7 +10,6 @@ def sqlalchemy_version_check(version: str | Version) -> None:
 try:
     import sqlalchemy
```
Shinnnyshinshin, line 12: :rocket:

great_expectations/compatibility/sqlalchemy_compatibility_wrappers.py
```diff
@@ -111,8 +111,6 @@ def add_dataframe_to_db(
                 * 'multi': Pass multiple values in a single ``INSERT`` clause.
                 * callable with signature ``(pd_table, conn, keys, data_iter)``.
     """
```
Shinnnyshinshin, line 113: very happy to see this going away

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -278,6 +325,13 @@ def __init__(  # noqa: C901 - 17
         self._create_temp_table = create_temp_table
         os.environ["SF_PARTNER"] = "great_expectations_oss"
 
+        # sqlite/mssql temp tables only persist within a connection, so we need to keep the connection alive by
```
Shinnnyshinshin, line 328: :heart:

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -484,7 +573,13 @@ def _build_engine(self, credentials: dict, **kwargs) -> sa.engine.Engine:
             options = get_sqlalchemy_url(drivername, **credentials)
 
         self.drivername = drivername
-        engine = sa.create_engine(options, **create_engine_kwargs)
+        if _dialect_requires_persisted_connection(credentials=credentials):
```
Shinnnyshinshin, line 576: :rocket:

great_expectations/execution_engine/sqlalchemy_execution_engine.py
Shinnnyshinshin: Yes I agree. Let's get rid of it soon.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1293,3 +1379,70 @@ def get_batch_data_and_markers(
...
+        """
+
+        with self.get_connection() as connection:
+            result = connection.execute(query)
+
+        return result
+
+    @public_api
+    @new_method_or_class(version="0.16.14")
+    def execute_query_in_transaction(
```
Shinnnyshinshin, line 1430: :rocket:

great_expectations/expectations/metrics/column_map_metrics/column_values_unique.py
Shinnnyshinshin: Interesting that this an `sqlalchemy_batch_data` are our few `execute_query_in_transaction` usages. I'm just thinking to myself here: I wonder what makes theses ones different than the other metrics

great_expectations/self_check/util.py
Shinnnyshinshin: YES

tests/datasource/conftest.py
```diff
@@ -159,7 +160,8 @@ def test_cases_for_sql_data_connector_sqlite_execution_engine(
         raise ValueError("SQL Database tests require sqlalchemy to be installed.")
 
     engine: sa.engine.Engine = sa.create_engine(
-        test_cases_for_sql_data_connector_sqlite_connection_url
+        test_cases_for_sql_data_connector_sqlite_connection_url,
+        poolclass=sqlalchemy.StaticPool,
```
Shinnnyshinshin, line 164: :rocket:

tests/execution_engine/test_sqlalchemy_execution_engine.py
```diff
@@ -1003,3 +1059,159 @@ def validate_tmp_tables(execution_engine):
...
     validate_tmp_tables(execution_engine=execution_engine)
+
+
+@pytest.fixture
+def pd_dataframe() -> pd.DataFrame:
+    return pd.DataFrame({"a": [1, 2], "b": [4, 4]})
+
+
+@pytest.mark.unit
+class TestExecuteQuery:
```
Shinnnyshinshin, line 1070: omg this is great

tests/execution_engine/test_sqlalchemy_execution_engine.py
```diff
@@ -1003,3 +1059,159 @@ def validate_tmp_tables(execution_engine):
...
+        select_temp_table = "SELECT * FROM temp_table;"
+
+        res = execution_engine.execute_query(sa.text(select_temp_table)).fetchall()
+        res2 = execution_engine.execute_query(sa.text(select_temp_table)).fetchall()
+
+        # This assert is here just to make sure when we assert res == res2 we are not comparing None == None
+        expected = [(1, 4), (2, 4)]
+        assert [r for r in res] == expected
+
+        assert res == res2
```
Shinnnyshinshin, line 1142: @anthonyburdi I might just be missing this, but which part tests that the `con` is the same as the one used from the static pool?

Shinnnyshinshin (commented):
one question, but otherwise we are good to go

Shinnnyshinshin (commented):
one question, but otherwise we are good to go

tests/execution_engine/test_sqlalchemy_execution_engine.py
anthonyburdi: Oh good question. If memory serves, I believe that the id of the actual connection pulled out of the static pool is different even though it is the same connection so I removed that test and instead tested by attempting to access the same temp table twice. If a different connection were used the second query would not be successful.

That was when I had removed storing the connection as an instance variable in `self._connection` before realizing that sqlite (or sqlalchemy) would still close the connection if there was no reference to it.

I think I can add that test back now, thank you for asking this question!

anthonyburdi: Updated and added in https://github.com/great-expectations/great_expectations/pull/7607/commits/a1a8b07371490af399011b6820265c620c78d006

tests/execution_engine/test_sqlalchemy_execution_engine.py
```diff
@@ -1141,6 +1141,40 @@ def test_same_connection_used_from_static_pool_sqlite(
...
+            add_dataframe_to_db(df=pd_dataframe, name="test", con=con, index=False)
+            connection = con
+        assert (
+            execution_engine.dialect_name == GXSqlDialect.SQLITE
+        ), "Error here means test setup failed."
+
+        create_temp_table = "CREATE TEMPORARY TABLE temp_table AS SELECT * FROM test;"
+        execution_engine.execute_query_in_transaction(sa.text(create_temp_table))
+
+        with execution_engine.get_connection() as test_con:
```
Shinnnyshinshin, line 1165: 🚀

Shinnnyshinshin (approved):
🚀🚀🚀🚀🚀🚀

Shinnnyshinshin (approved):
🚀🚀🚀🚀🚀🚀

PR #7609, opened 2023-04-12
[MAINTENANCE]  Deprecate ColumnExpectation in favor of ColumnAggregateExpectation

abegong:
- Adds `ColumnAggregateExpectation` as a new class, with `ColumnExpectation` as a no-op child class.
- Adds a deprecation warning to `ColumnExpectation`.
- Changes all references from `ColumnExpectation` to `ColumnAggregateExpectation`

Questions:
* I added the deprecation warning in `ColumnExpectation.__init__`. Is that the right place?
* There are two .rst files that reference `ColumnExpectation`. I haven't changed them to `ColumnAggregateExpectation`. Should I?
* Should I add any tests for `ColumnExpectations`? Proposal: assuming the docs for creating custom `ColumnAggregateExpectations` are under test, I don't think it's necessary, since `ColumnExpectations` don't have any independent functionality.


great_expectations/expectations/expectation.py
```diff
@@ -2573,6 +2573,49 @@ def validate_configuration(
...
+    Raises:
+        InvalidExpectationConfigurationError: If no `column` is specified
+    """
+
+    def __init__(
+        self,
+        configuration: Optional[ExpectationConfiguration] = None,
+    ):
+        # deprecated-v0.16.6
+        warnings.warn(
```
donaldheppner: If we introduce a deprecation warning here now, it will raise warnings for any suite containing a custom (non-contributed) expectation. I think we should remove this warning for now and add it back in after we've had a chance to communicate the change (ie: after the conceptual guide is out)

donaldheppner (changes requested):

great_expectations/expectations/expectation.py
abegong: Good point. Yes, I agree that's better.

donaldheppner (approved):

PR #7619, opened 2023-04-13
[BUGFIX] Remove spark from bic Expectations since it never worked for them

kenwade4:
- Remove spark from bic Expectations since it never worked them


alexsherstinsky (approved):
@kenwade4 Maybe you and I can make Spark work later.  Thanks!

PR #7626, opened 2023-04-14
[BUGFIX] MapCondition Memory Inefficiencies in Spark

### LINKED ISSUE #7608: Memory problems in map_condition_auxillary_methods._spark_map_condition_index() in 0.16.6
**Describe the bug**
I was upgrading our GX stack to the latest version (0.16.6) from 0.15.26. Some of our benchmarks on medium-sized data (~ 5 million rows) were running out of memory where they had run fine before. I did some debugging and it looks like there are two problems in `map_condition_auxillary_methods._spark_map_condition_index()`.

Apologies for jamming two problems into one issue, but they're literally in two consecutive lines in the code.

**Problem 1** 
The following code attempts to limit the number of results based on the unexpected count. However, it doesn't re-assign the the dataframe.
```
if result_format["result_format"] != "COMPLETE":
        filtered.limit(result_format["partial_unexpected_count"])
```

After changing it to following:

```
    if result_format["result_format"] != "COMPLETE":
        filtered = filtered.limit(result_format["partial_unexpected_count"])
```

The number of objects created by `run_checkpoint` dropped dramatically. On a pretty small dataset (500K rows), the number of python dictionaries created went from 

```
                                     types |   # objects |   total size
========================================== | =========== | ============
                                      dict |     1601406 |    354.50 MB
...
```

to

```
                                    types |   # objects |   total size
========================================== | =========== | ============
                                     dict |        5600 |      1.23 MB
...
```

The difference was even more pronounced on larger datasets.

**Problem 2**

In the original code, I noticed that this line was consuming the bulk of our runtime per the spark UI.

```
    if result_format["result_format"] != "COMPLETE":
        filtered.limit(result_format["partial_unexpected_count"])

    for row in filtered.collect():  <<<----- This was the culprit
        dict_to_add: dict = {}
        for col_name in columns_to_keep:
            dict_to_add[col_name] = row[col_name]
        unexpected_index_list.append(dict_to_add)

    return unexpected_index_list
```

After looking at this in the debugger, the problem was that the dataframe `filtered` had all 260 columns present even though we only cared about a few in `columns_to_keep`. The code inside the `collect()` for loop ignores anything that isn't in `columns_to_keep`.  Adding the following line dramatically reduced memory (I couldn't even run our 5 million row test set without kube's OOM killer stomping our container).

```
    if result_format["result_format"] != "COMPLETE":
        filtered.limit(result_format["partial_unexpected_count"])

    # Prune the dataframe down only the columns we care about
    filtered = filtered.select(columns_to_keep)  <<<------ This is the line I added

    for row in filtered.collect():
        dict_to_add: dict = {}
        for col_name in columns_to_keep:
            dict_to_add[col_name] = row[col_name]
        unexpected_index_list.append(dict_to_add)

    return unexpected_index_list
```

**To Reproduce**
1. You should be able to see a difference running sufficiently large data with some expectations that land in this part of the code. For our suite, a good 1/3 of the expectations landed here.

**Expected behavior**
Be as frugal as we can be with memory. There's still a gap in runtime between 0.15.26 and 0.16.6 that I'm hunting down. It's about 25% slower even with the above changes, but there could be some API changes that we're not utilizing properly. Still, I believe the above changes are good general changes to consider.

 - Operating System:  MacOS (running on an Ubuntu container)
 - Great Expectations Version: 0.16.6

**Additional context**
For now, I'll just build these in as overrides into our container an press on, but I thought it might be good to share these findings with the community for consideration.

Shinnnyshinshin:
- Adds changes proposed in #7608
- closes #7608 
- Adds missing test that would have caught the error. 

### Definition of Done


tests/checkpoint/test_checkpoint_result_format.py
```diff
@@ -2085,6 +2085,63 @@ def test_spark_result_format_in_checkpoint_pk_defined_one_expectation_summary_ou
     assert evrs[0]["results"][0]["result"].get("unexpected_index_query") is None
 
 
+@pytest.mark.integration
+def test_spark_result_format_in_checkpoint_pk_defined_one_expectation_summary_output_limit_1(
```
Shinnnyshinshin, line 2089: This test exists for Sql and Pandas, and was missing for Spark

great_expectations/expectations/metrics/map_metric_provider/map_condition_auxilliary_methods.py
```diff
@@ -772,7 +772,10 @@ def _spark_map_condition_index(
             )
 
     if result_format["result_format"] != "COMPLETE":
-        filtered.limit(result_format["partial_unexpected_count"])
+        filtered = filtered.limit(result_format["partial_unexpected_count"])
+
+    # Prune the dataframe down only the columns we care about
+    filtered = filtered.select(columns_to_keep)
```
Shinnnyshinshin, line 778: Suggested lines in #7608

alexsherstinsky (approved):
LGTM

PR #7635, opened 2023-04-15
[MAINTENANCE] Update NotImported mechanism to use scoped compatibility modules

alexsherstinsky:
### Scope
* This change moves all imports into scoped modules (`sqlalchemy`, `pyspark`, `azure`, `google`, etc.).
* Some code cleanup, facilitated by this change.

## Next To Be Done
Update imports to use the new `NotImported` style for the packages AWS S3, Trino, AWS Athena, Google BigQuery, Snowflake, and AWS Redshift (and any others as needed).
 
- JIRA: DX-1/DX-440
-
-

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


contrib/experimental/great_expectations_experimental/tests/rule_based_profiler/data_assistant/test_growth_numeric_data_assistant.py
```diff
@@ -413,9 +413,9 @@ def test_spark_happy_path_growth_numeric_data_assistant(
     4. Configuring BatchRequest to load 2020 January data
     5. Configuring and running Checkpoint using BatchRequest for 2020-01, and 'taxi_data_2019_suite'.
     """
-    from pyspark.sql.types import StructType
```
anthonyburdi, line 416: Why not import StructType here from `great_expectations.compatibility.pyspark`? (this applies to anywhere that this pattern shows up).

great_expectations/compatibility/pyspark.py
```diff
@@ -0,0 +1,72 @@
...
+
+from great_expectations.compatibility.not_imported import NotImported
+
+SPARK_NOT_IMPORTED = NotImported(
+    "pyspark is not installed, please 'pip install pyspark'"
+)
+
+try:
+    import pyspark  # noqa: TID251
+except (ImportError, AttributeError):
```
anthonyburdi: No need for AttributeError here. Only if we are accessing something within a package (like `from pyspark.sql import functions`). (This is applicable to the other files as well)

great_expectations/compatibility/pyspark.py
```diff
@@ -0,0 +1,72 @@
+from __future__ import annotations
+
+from great_expectations.compatibility.not_imported import NotImported
+
+SPARK_NOT_IMPORTED = NotImported(
+    "pyspark is not installed, please 'pip install pyspark'"
+)
+
+try:
+    import pyspark  # noqa: TID251
```
anthonyburdi: You shouldn't need the # noqa: TID251 for any of the compatibility files. That is what the `[tool.ruff.per-file-ignores]` directives are for. In other words, by adding the per file ignore for TID251 we are allowing import in the compatibility files and disallowing it elsewhere (to make sure we use the compatibility version).

anthonyburdi: (This is applicable to the other files as well)

great_expectations/expectations/metrics/column_aggregate_metrics/column_value_counts.py
```diff
@@ -2,6 +2,27 @@
 
 import pandas as pd
 
+from great_expectations.compatibility.pyspark import (
+    DataFrame as pyspark_sql_DataFrame,
```
anthonyburdi: One of the main goals of breaking these compatibility files up into package specific files (like pyspark.py) is so we don't need to add the prefixes like `pyspark_sql_` in this case (unless there is a conflict).

anthonyburdi (changes requested):
I haven't gone through everything yet but I did notice some items to address.

contrib/experimental/great_expectations_experimental/tests/rule_based_profiler/data_assistant/test_growth_numeric_data_assistant.py
alexsherstinsky: @anthonyburdi The style for this was in the name of generality -- covers everything under `types.`.

great_expectations/compatibility/pyspark.py
alexsherstinsky: @anthonyburdi I agree, but it did not work.  I can try again.

great_expectations/compatibility/pyspark.py
alexsherstinsky: @anthonyburdi I agree -- let me see how far I can go with removing `AttributeError` (it was extra precaution).  Thanks.

great_expectations/expectations/metrics/column_aggregate_metrics/column_value_counts.py
alexsherstinsky: @anthonyburdi This was a matter of style -- in particular, I wanted to underscore it for modules that can have multiple `ExecutionEngine` implementations (e.g., most metrics).  So this was a move for "let us be super strict and careful" (even though we could "get away" with being a bit more lax -- I feel that this was would reduce confusion long term).  Thanks.

great_expectations/compatibility/pyspark.py
alexsherstinsky: @anthonyburdi This was done -- I deleted `AttributeError` wherever it was superfluous.  Thanks.

great_expectations/compatibility/pyspark.py
alexsherstinsky: @anthonyburdi Amazing -- I tried removing `# noqa: TID251` "escapes" and it seems to be working fine now.  I wonder if also deleting the `AttributeError` from the applicable `except` clauses had something to do with it.  Thanks.

great_expectations/expectations/metrics/column_aggregate_metrics/column_value_counts.py
anthonyburdi: How about instead, now that we have the `pyspark` module we do `from great_expectations.compatibility import pyspark` and then in usage `pyspark.DataFrame`. This would make it clear which DataFrame we are using in metric implementations.
This style is also compatible with our style guide: https://google.github.io/styleguide/pyguide.html#22-imports which is to "Use import statements for packages and modules only, not for individual classes or functions." (The link to the style guide gives more info for motivation and other cases).
It would be nice to move towards that style for all of our code over time, and especially here since we are changing so many import statements.

contrib/experimental/great_expectations_experimental/tests/rule_based_profiler/data_assistant/test_growth_numeric_data_assistant.py
anthonyburdi: Wouldn't this raise an `AttributeError` though if pyspark isn't installed and `sparktypes` is `NotImported` since there is no `StructType` on `NotImported`?

great_expectations/expectations/metrics/column_aggregate_metrics/column_value_counts.py
alexsherstinsky: @anthonyburdi The idea is nice, but I am not seeing how it will work, because `DataFrame` is imported not from `pyspark` but from `pyspark.sql`.  This situation is not limited to `DataFrame` in `pyspark` -- there are many other occurrences across the packages we are importing.  Do you have a solution in mind?  Thanks.

contrib/experimental/great_expectations_experimental/tests/rule_based_profiler/data_assistant/test_growth_numeric_data_assistant.py
alexsherstinsky: @anthonyburdi It seems that it potentially could -- you are right, but it has not happened in our tests.  I need to check to see if the code involved when we use `sparktypes` is in general executed only when `pyspark` is imported and exists (then we should be fine).  More soon.  Thanks.

great_expectations/expectations/metrics/column_aggregate_metrics/column_value_counts.py
anthonyburdi: In usages we import the new pyspark compatibility package (`compatibility/pyspark.py` via `from great_expectations.compatibility import pyspark`) which imports `DataFrame`. Then in metrics etc we use that `DataFrame` from the `compatibility/pyspark.py`. Here is an example:

`great_expectations/compatibility/pyspark.py`:
```
try:
    from pyspark.sql import DataFrame
except (ImportError, AttributeError):
    DataFrame = SPARK_NOT_IMPORTED
```

then in usage (note this `pyspark` import references the `great_expectations/compatibility/pyspark.py` module):

```
from great_expectations.compatibility import pyspark

my_df: pyspark.DataFrame
```

contrib/experimental/great_expectations_experimental/tests/rule_based_profiler/data_assistant/test_growth_numeric_data_assistant.py
alexsherstinsky: @anthonyburdi I did a thorough check -- in all ~26 files where we use `sparktypes`, it is within methods/modules where a valid `pyspark` installation is expected, so we should be good here.  I just added a couple of tiny cosmetic changes (a type hint; a couple of commented imports for readability), based on my observations while checking these files.  Thanks!

great_expectations/expectations/metrics/column_aggregate_metrics/column_value_counts.py
alexsherstinsky: @anthonyburdi My approach focused on not having users/developers import `SPARK_NOT_IMPORTED` and other similar "`_NOT_IMPORTED`" constants.  I was pursuing the idea of "things just working" in application modules, whereby all "`_NOT_IMPORTED`" constants would be treated as "internal implementation" and stayed encapsulated within the appropriate compatibility modules.  I was concerned that types and other issues could happen if users had to follow the try/except pattern and import/utilize the "`_NOT_IMPORTED`" constants in application modules.  Happy to discuss.  Thank you.

alexsherstinsky: @anthonyburdi I affected the change as per our conversation.  Thanks!

great_expectations/core/batch.py
```diff
@@ -652,10 +649,10 @@ def head(self, *args, **kwargs):
         return pd.DataFrame({})
 
 
-if SPARK_NOT_IMPORTED:
-    BatchDataType = Union[BatchData, pd.DataFrame]
+if pyspark:
```
anthonyburdi, line 652: Love it! Much clearer and now no need for SPARK_NOT_IMPORTED.

great_expectations/data_context/types/base.py
```diff
@@ -460,15 +456,19 @@ def prepare_dump(self, data, **kwargs):
         This method calls the schema's jsonValue() method, which translates the object into a json
         """
         # check whether spark exists
-        if StructType is None:
+        if (not pyspark.types) or (pyspark.types.StructType is None):
```
anthonyburdi, line 459: StructType won't be None so I think you can remove the right side of the `or`

great_expectations/data_context/types/base.py
```diff
@@ -933,15 +933,19 @@ def prepare_dump(self, data, **kwargs):
...
```
anthonyburdi: Same comment, we have to be careful comparing to None since we are now using the NotImported type

anthonyburdi (approved):
Thank you for making all these changes! I have a few small requests, and a question but I'll approve and leave to your judgement. The question is - why are there type ignore statements in some `compatibility/*` files but not others? E.g. they are in `great_expectations/compatibility/google.py` but not `sqlalchemy.py`? Can we remove more of them?

great_expectations/data_context/types/base.py
alexsherstinsky: @anthonyburdi I see what you mean.  I fixed to utilize the boolean of `NotImported` and guard against `AttributeError` -- thanks!

alexsherstinsky:
> Thank you for making all these changes! I have a few small requests, and a question but I'll approve and leave to your judgement. The question is - why are there type ignore statements in some `compatibility/*` files but not others? E.g. they are in `great_expectations/compatibility/google.py` but not `sqlalchemy.py`? Can we remove more of them?

@anthonyburdi It is not entirely clear why this inconsistency happens.  (It was observed even before the large refactoring work, aimed at utilizing `NotImported` in the best possible way, got under way.). One possibility is that the internals of the `SQLAlchemy` package behave differently than those of `PySpark` and others, thus being more compatible with the `_NOT_IMPORTED` constants of the `NotImported` object type, thus not requiring these `ignore` statements.  Thanks.

PR #7640, opened 2023-04-17
[MAINTENANCE] Pandas 2.0 compatibility - Azure dependencies

Shinnnyshinshin:
- `Minimal Standard` pipeline Stage is run with `pip install --constraint constraints-dev.txt ".[test]" pytest-azurepipelines` -- 
    - `https://github.com/great-expectations/great_expectations/blob/2367d978b85ebec76c7e9bd81b6e562b0dc7f9c3/ci/azure-pipelines-dev.yml#L346`

- This means that `lite`, `contrib` and `docs_test`-related packages are installed, which does not include Azure
     - `https://github.com/great-expectations/great_expectations/blob/2367d978b85ebec76c7e9bd81b6e562b0dc7f9c3/setup.py#L61`

- Therefore the Azure DataConnector tests are marked to skip if `azure_storage` is not imported, using the same pattern used by Google Cloud Storage tests.

- 65 -> 21 tests failed on `Azure Pipelines / pr (required minimal standard)`

### Definition of Done


tests/datasource/data_connector/test_configured_asset_azure_data_connector.py
```diff
@@ -228,6 +229,10 @@ def expected_batch_definitions_sorted():
     return expected
 
 
+@pytest.mark.skipif(
+    not azure_storage,
+    reason="Could not import 'storage' from azure in configured_asset_gcs_data_connector.py",
```
cdkini: accidental GCS ref here?

tests/datasource/data_connector/test_configured_asset_azure_data_connector.py
```diff
@@ -261,6 +266,10 @@ def test_instantiation_with_account_url_and_credential(
     assert my_data_connector.get_unmatched_data_references() == []
 
 
+@pytest.mark.skipif(
+    not azure_storage,
+    reason="Could not import 'storage' from azure in configured_asset_gcs_data_connector.py",
```
cdkini: Same note here

anthonyburdi (changes requested):
Although we use it elsewhere in the codebase, I have a few concerns with this pattern and some questions around how we install the optional dependencies.

My concern with this pattern is if a package is not installed or fails to install that makes our tests irrelevant (the stage would pass since they are all skipped). Also, in which pipeline would these tests run?

I think we also need to add the pandas optional azure reqs to our azure specific requirements `reqs/requirements-dev-azure.txt` and probably change or add instructions to our docs.

Shinnnyshinshin:
> Although we use it elsewhere in the codebase, I have a few concerns with this pattern and some questions around how we install the optional dependencies.
> 
> My concern with this pattern is if a package is not installed or fails to install that makes our tests irrelevant (the stage would pass since they are all skipped). Also, in which pipeline would these tests run?
> 
> I think we also need to add the pandas optional azure reqs to our azure specific requirements `reqs/requirements-dev-azure.txt` and probably change or add instructions to our docs.

You're absolutely right @anthonyburdi, esp your question about "when are these tests ever going to be run?" gave me pause. 

* I was thinking that we can be better about splitting up `unittest` stage and `comprehensive minimal` stage.
* The Azure tests that are being updated in this PR are strictly `unittests`, because the `DataConnector` methods are mocked, and there is no actual connection to an external cloud resource.
* The issue right now is that the `UnitTest` stage is also installed with the `[.test]` flag, which means it has the same set of packages (without `azure`) are installed. 


One way that we can address this is to mark these tests as `unit` and run them as part of the unittests stage.  The issue is that we would need to add more packages to be installed for that stage. 
* How strict do we want to be that the current `unittest` stage only be the `lite`, `contrib` and `docs_test`. 
* What about other cloud resources that are able to be mocked? (I'm thinking of AWS).  
* Would a `Unittest Minimal`  + `Unittest Comprehensive` split be helpful here?

tests/datasource/data_connector/test_configured_asset_azure_data_connector.py
Shinnnyshinshin: Sorry about this 😅 . Silly mistake on my part.  PR has been updated

tests/datasource/data_connector/test_configured_asset_azure_data_connector.py
Shinnnyshinshin: PR has been updated 😓

Kilo59 (approved):
Approving to unblock the release

Shinnnyshinshin:
Checked to see this was handled by https://github.com/great-expectations/great_expectations/pull/7818 closing PR :)

PR #7651, opened 2023-04-18
[MAINTENANCE] The 'sklearn' PyPI package is deprecated, use 'scikit-learn'

kenwade4:
- Don't use deprecated PyPi package sklearn, only scikit-learn

### DIFF
--- a/great_expectations/core/usage_statistics/package_dependencies.py
+++ b/great_expectations/core/usage_statistics/package_dependencies.py
@@ -195,7 +195,6 @@ class GXDependencies:
         "scikit-learn",
         "shapely",
         "simple_icd_10",
-        "sklearn",
         "sympy",
         "tensorflow",
         "timezonefinder",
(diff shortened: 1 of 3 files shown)


great_expectations/core/usage_statistics/package_dependencies.py
```diff
@@ -195,7 +195,6 @@ class GXDependencies:
         "scikit-learn",
         "shapely",
         "simple_icd_10",
-        "sklearn",
         "sympy",
```
Shinnnyshinshin, line 198: Just wanting to check : should `sklearn` here be replaced with `scikit-learn`?

reqs/requirements-dev-all-contrib-expectations.txt
```diff
@@ -38,7 +38,6 @@ schwifty
 scikit-learn
 shapely
 simple_icd_10
-sklearn
```
Shinnnyshinshin, line 41: same question as above

great_expectations/core/usage_statistics/package_dependencies.py
kenwade4: Yes, but scikit-learn is already in that list. Ditto for other comment.

Shinnnyshinshin (approved):
LGTM. Thanks for the clarification

PR #7667, opened 2023-04-20
[MAINTENANCE] list_datasources should return FDS configs as well

jshaikGX:
- The list_datasources api should also return fluent datasources config, this fix does that

### Definition of Done

Thank you for submitting!

### DIFF
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -1564,6 +1564,12 @@ class AbstractDataContext(ConfigPeer, ABC):
                 )
             )
             datasources.append(masked_config)
+
+        for (
+            datasource_name,
+            fluent_datasource_config,
+        ) in self.fluent_datasources.items():
+            datasources.append(fluent_datasource_config.dict())
         return datasources
 
     @public_api
--- a/tests/data_context/test_data_context_datasources.py
+++ b/tests/data_context/test_data_context_datasources.py
@@ -1,5 +1,6 @@
 import pathlib
 from unittest import mock
+import great_expectations as gx
 
 import pytest
 
@@ -17,6 +18,7 @@ from great_expectations.data_context.types.base import (
     DataContextConfig,
     DatasourceConfig,
     GXCloudConfig,
+    InMemoryStoreBackendDefaults,
 )
 from great_expectations.datasource import Datasource
 from great_expectations.util import get_context
@@ -427,3 +429,27 @@ def test_BaseDataContext_delete_datasource_updates_cache(
 
     assert not mock_delete.called
     assert name not in context.datasources
+
+
+@pytest.mark.unit
+def test_list_datasources() -> None:
+    project_config = DataContextConfig(
+        store_backend_defaults=InMemoryStoreBackendDefaults()
+    )
+    project_config.datasources = {
+        "my_datasource_name": {
+            "class_name": "Datasource",
+            "data_connectors": {},
+            "execution_engine": {
+                "class_name": "PandasExecutionEngine",
+                "module_name": "great_expectations.execution_engine",
+            },
+            "module_name": "great_expectations.datasource",
+        }
+    }
+    context = gx.get_context(project_config=project_config)
+
+    datasource_name = "my_experimental_datasource_awaiting_migration"
+    context.sources.add_pandas(datasource_name)
+
+    assert len(context.list_datasources()) == 2
--- a/great_expectations/data_context/types/base.py
+++ b/great_expectations/data_context/types/base.py
@@ -2471,6 +2471,7 @@ class DataContextConfig(BaseYamlConfig):
         if datasources is None:
             datasources = {}  # type: ignore[assignment]
         self.datasources = datasources
+        self.fluent_datasources = fluent_datasources or {}
         self.expectations_store_name = expectations_store_name
         self.validations_store_name = validations_store_name
         self.evaluation_parameter_store_name = evaluation_parameter_store_name
--- a/tests/data_context/store/test_store_backends.py
+++ b/tests/data_context/store/test_store_backends.py
@@ -1406,6 +1406,7 @@ def test_InlineStoreBackend(empty_data_context: DataContext) -> None:
         ("datasources",),
         ("evaluation_parameter_store_name",),
         ("expectations_store_name",),
+        ("fluent_datasources",),
         ("include_rendered_content",),
         ("notebooks",),
         ("plugins_directory",),
--- a/tests/data_context/test_data_context_v013.py
+++ b/tests/data_context/test_data_context_v013.py
@@ -224,6 +224,7 @@ def test_get_config(empty_data_context):
         "plugins_directory",
         "stores",
         "expectations_store_name",
+        "fluent_datasources",
         "validations_store_name",
         "evaluation_parameter_store_name",
         "checkpoint_store_name",


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1558,11 +1565,16 @@ def list_datasources(self) -> List[dict]:
...
+            serialized_config: dict = (
+                self._serialize_datasource_config(
                     serializer, datasource_config
                 )
             )
+            masked_config = self._substitute_and_sanitize_datasource_config(serialized_config)
+            datasources.append(masked_config)
+
+        for datasource_name, datasource_config in self.config.fluent_datasources.items():
+            masked_config = self._substitute_and_sanitize_datasource_config(datasource_config)
```
Kilo59: We do config substitution differently for fluent datasources.
For BDS we eagerly substitute configs values but for FDS we do it lazily for special `ConfigStr` types, so I don't know that `_substitute_and_sanitize_datasource_config` makes sense here.

https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/datasource/fluent/config_str.py

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1475,26 +1475,33 @@ def get_datasource(
...
+            self, serializer: AbstractConfigSerializer, datasource_config: DatasourceConfig) -> dict:
+        """Serialize and return as dict.
+        Args:
+            serializer: Serializer to use when converting config to dict for substitutions.
+            datasource_config: Datasource config to process.
+        Returns:
+            Dict of config serialization applied.
+        """
+        datasource_dict: dict = serializer.serialize(datasource_config)
+        return datasource_dict
```
cdkini: Nit
```suggestion
        return serializer.serialize(datasource_config)
```

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: Ah okay I may have led @jshaikGX astray here then.

Given how substitution is done for fluent datasources, are we ever concerned about credential leakage? I was going to recommend a test that ensures credential obfuscation but if it isn't relevant lmk

tests/data_context/abstract_data_context/test_list_datasources.py
```diff
@@ -1,7 +1,8 @@
 from typing import List
 
 from great_expectations import DataContext
-from great_expectations.data_context.types.base import DatasourceConfig
+from great_expectations.data_context import EphemeralDataContext
+from great_expectations.data_context.types.base import DatasourceConfig, DataContextConfig
```
cdkini: I don't think any of the changes in this file are necessary right? Let's revert this back to its original state

tests/data_context/test_data_context_datasources.py
```diff
@@ -16,7 +17,7 @@
 from great_expectations.data_context.types.base import (
     DataContextConfig,
     DatasourceConfig,
-    GXCloudConfig,
+    GXCloudConfig, InMemoryStoreBackendDefaults,
```
cdkini: I think the linter may get mad here - you'll want to run `ruff --fix .` or set up pre-commit hooks to cover this

tests/data_context/test_data_context_datasources.py
```diff
@@ -427,3 +428,32 @@ def test_BaseDataContext_delete_datasource_updates_cache(
...
+            "my_datasource_1": {
+                "type": "sqlite",
+                "name": "my_datasource_1",
+                "connection_string": "sqlite: // /"
+            }
+        }
+    }
+    context = gx.get_context(project_config=project_config)
+
+    assert len(context.list_datasources()) == 2
```
cdkini, line 455: If my earlier comment about credential obfuscation is relevant, could we add an additional test here that asserts against the actual output (checking to make sure that sensitive values are covered)

cdkini (commented):
Looks great! Left some minor comments but the core of the logic is sound - will defer to other team members to see this across the line but let me know if you want to talk through anything

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59: Because of the lazy loading, there are no credentials to actually leak. So it's inherently safer. 
You have to call `.get_config_value()` on the `ConfigStr` type to do the substitution. 

Here are some tests for `ConfigStr` and demonstrations of how it works.
https://github.com/great-expectations/great_expectations/blob/develop/tests/datasource/fluent/test_config_str.py

tests/data_context/test_data_context_datasources.py
jshaikGX: yup

great_expectations/data_context/data_context/abstract_data_context.py
jshaikGX: @Kilo59 Let me know if this isnt relevant

tests/data_context/abstract_data_context/test_list_datasources.py
NathanFarmer: +1

great_expectations/data_context/data_context/abstract_data_context.py
NathanFarmer: Also, if you remove line 1577 I think your type checking problem will go away.

NathanFarmer (commented):
I also agree that this looks good pending what @Kilo59 thinks about `ConfigStr`.

tests/data_context/test_data_context_datasources.py
```diff
@@ -427,3 +429,33 @@ def test_BaseDataContext_delete_datasource_updates_cache(
...
+    }
+    project_config.fluent_datasources = {
+        "fluent_datasources": {
+            "my_datasource_1": {
+                "type": "sqlite",
+                "name": "my_datasource_1",
+                "connection_string": "sqlite: // /",
+            }
+        }
+    }
```
cdkini: I think the top level `fluent_datasources` key is not needed here - see how on L440 our key corresponds to the datasource name?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1564,6 +1564,9 @@ def list_datasources(self) -> List[dict]:
                 )
             )
             datasources.append(masked_config)
+
+        for datasource_name, datasource_config in self.config.fluent_datasources.items():
```
```suggestion
        for datasource_name, datasource_config in self.fluent_datasources.items():
```
Excuse my ignorance here but why are we pulling from `self.config.fluent_datasources` instead of just `self.fluent_datasources`?

I admit I don't know how or when `self.config.fluent_datasources` is set. But `self.config.fluent_datasources` seems like a safer bet and less likely to ever be "stale".

https://github.com/great-expectations/great_expectations/blob/4e9199122c4f9fb12b298d104b35ee8f4d1b0005/great_expectations/data_context/data_context/abstract_data_context.py#L4454-L4459

cdkini: `self.fluent_datasources` is a dict of hydrated business logic objects no? We need config because that's whats emitted to the user in `list_datasources`

jshaikGX: ^ and the self.config.fluent_datasources is initialized in the 

[data_context/types/base.py]

(https://github.com/great-expectations/great_expectations/pull/7667/files#diff-2e0903f2ba2f628b8645e3bb6decf8d451c943e724fc3852f555e94e92b0430f)

Kilo59: Maybe I just missed it but I've never written or reviewed any code that sets fluent datasource objects to `self.config`.
So I worry the test in this PR may be misleading for how this would actually work.

tests/data_context/test_data_context_datasources.py
```diff
@@ -427,3 +429,31 @@ def test_BaseDataContext_delete_datasource_updates_cache(
 
     assert not mock_delete.called
     assert name not in context.datasources
+
+
+@pytest.mark.unit
+def test_list_datasources() -> None:
```
Kilo59, line 435: I'm worried this test may not reflect how fluent datasources interact with `self.config` in the wild.

Can @jshaikGX you create another version of this test that adds a fluent datasource via `context.sources.add_<ANY_DATASOURCE>()` and then checks if it is part of `context.list_datasources()`?

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: hmm how do we actually persist fluent datasources? In YAML or in the Cloud DB right? Those values are provided back to us in the form of `DataContextConfig` no?

jshaikGX: yea when I tested this locally with a sqllite datasource (this FDS) was persisted in the YAML

```

context = gx.get_context()

context.datasources
# {}

context.sources.add_sqlite(name="my_datasource_1", connection_string="sqlite:///", kwargs={"isolation_level": "SERIALIZABLE"},)
# SqliteDatasource(type='sqlite', name='my_datasource_1', id=None, assets={}, connection_string=SqliteDsn('sqlite:///', scheme='sqlite', path='/'), create_temp_table=True, kwargs={'isolation_level': 'SERIALIZABLE'})

context.datasources
# {'my_datasource_1': SqliteDatasource(type='sqlite', name='my_datasource_1', id=None, assets={}, connection_string=SqliteDsn('sqlite:///', scheme='sqlite', path='/'), create_temp_table=True, kwargs={'isolation_level': 'SERIALIZABLE'})}

In [5]: context.list_datasources()
# []
```

Kilo59 (changes requested):
I think we need another test to confirm that this works with `self.config.fluent_datasources`.
I've proposed what that test should look like.
And if it fails the fix should be simple.

tests/data_context/test_data_context_datasources.py
jshaikGX: Sure

great_expectations/data_context/data_context/abstract_data_context.py
Kilo59: @cdkini 
>hmm how do we actually persist fluent datasources? In YAML or in the Cloud DB right? Those values are provided >back to us in the form of DataContextConfig no?

For cloud that might be true but for `FileDataContext` we use a separate code path to avoid the `instantiate_class_from_config` logic.

https://github.com/great-expectations/great_expectations/blob/4e9199122c4f9fb12b298d104b35ee8f4d1b0005/great_expectations/data_context/data_context/file_data_context.py#L199-L214

cdkini (approved):
LGTM - really nice work 🎉

Kilo59 (approved):

PR #7674, opened 2023-04-20
[MAINTENANCE] Only attempt docs-integration pipeline when manually triggered

kenwade4:
Even though the pipeline has been unpaused, we still don't want it to automatically run it until the issues are resolved that caused it to start failing in the first place

- Only attempt docs-integration pipeline when manually triggered


Shinnnyshinshin (approved):
LGTM

anthonyburdi (approved):
Thank you!

PR #7703, opened 2023-04-23
[DOCS] Light update to How to add validations data or suites to a Checkpoint

donaldheppner:
Updated YAML config to be inline with Fluent Datasources. Additional work required to bring this long-out-of-date doc under test.


billdirks (approved):
Thanks, this LGTM. I haven't tried serializing/deserializing batch request objects but this looks to conform to the object model.

PR #7720, opened 2023-04-25
[MAINTENANCE] Iterate over the regex_pattern characters too in 

kenwade4:
There was a bug in the _invert_regex_to_data_reference_template func.

The filename `yellow_tripdata_sample_2019-01.csv` as a pattern was ending up as `yellow_tripdata_sample_2019-01*csv` in batches.

```
FileNotFoundError: [Errno 2] No such file or directory: 'yellow_tripdata_sample_2019-01*csv'
```

- Iterate over the regex_pattern characters too in _invert_regex_to_data_reference_template


great_expectations/datasource/data_connector/util.py
```diff
@@ -249,7 +249,8 @@ def _invert_regex_to_data_reference_template(
 
     # print("-"*80)
     parsed_sre = sre_parse.parse(str(regex_pattern))
-    for token, value in parsed_sre:  # type: ignore[attr-defined]
+    for thing, char in zip(parsed_sre, list(str(regex_pattern))):  # type: ignore[attr-defined]
+        token, value = thing
```
anthonyburdi: Couldn't think of a great name but how about this?
```suggestion
    for parsed_sre_tuple, char in zip(parsed_sre, list(str(regex_pattern))):  # type: ignore[attr-defined]
        token, value = parsed_sre_tuple
```

anthonyburdi (changes requested):
This sounded like a journey to figure out! Thanks for fixing. Can you add some test cases to `test__invert_regex_to_data_reference_template` that would have failed and are now passing because of the change? That would help avoid future regressions.

anthonyburdi (approved):
Thank you for fixing this!

PR #7723, opened 2023-04-25
[MAINTENANCE] fix get available data assets names for fds

jshaikGX:
- get_available_data_asset_names() did not return fds asset names before. This fix here changes that

### Definition of Done

Thank you for submitting!


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3917,27 +3917,28 @@ def get_available_data_asset_names(
             ):  # Iterate over both together
                 for idx, datasource_name in enumerate(datasource_names):
                     datasource = self.get_datasource(datasource_name)
-                    assert not isinstance(
```
alexsherstinsky, line 3920: @jshaikGX Before we can confirm this approach, we need to make sure that we are not breaking anything.  The problem is that the existing implementation is based on GX V2!  So the question is -- will anything break that we are still supposed to support?  Perhaps something needs to be done to ensure that whoever is relying on the V2 construct `batch_kwargs_generator_names` (or anything to do with `batch_kwargs_generator`) should not be.  Maybe some kind of a warning or a deprecation notice is in order?  We never supported this call `get_available_data_asset_names()` at the DataContext level properly in V3 (only in Datasource) which is what you are using, but this will leave the V2 implementation non-existent.  Recently, @anthonyburdi deleted quite a bit of V2 functionality, please add him as a reviewer.  Thank you.

jshaikGX: @anthonyburdi

tests/data_context/test_data_context_datasources.py
```diff
@@ -453,3 +453,34 @@ def test_list_datasources() -> None:
...
+                    "class_name": "RuntimeDataConnector",
+                    "module_name": "great_expectations.datasource.data_connector",
+                    "batch_identifiers": ["default_identifier_name"],
+                },
+            },
+        }
+    }
+    context = gx.get_context(project_config=project_config)
+
+    datasource_name = "my_experimental_datasource_awaiting_migration"
```
NathanFarmer: I would just call this `my_fluent_pandas_datasource`.

tests/data_context/test_data_context_datasources.py
```diff
@@ -453,3 +453,34 @@ def test_list_datasources() -> None:
...
+    context.sources.add_pandas(datasource_name)
```
NathanFarmer: This just creates a datasource for you. You need to also add an asset like:
```
datasource = context.sources.add_pandas(datasource_name)
data_asset = datasource.add_dataframe_asset(df)
```

tests/data_context/test_data_context_datasources.py
```diff
@@ -453,3 +453,34 @@ def test_list_datasources() -> None:
...
+
+    assert len(context.get_available_data_asset_names()) == 2
```
NathanFarmer: What happens when you pass `datasource_names`? Do you get the same behavior for the block-config datasource as the fluent datasource?

NathanFarmer (changes requested):

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: I haven't yet looked at the code in this PR, but just to respond to this comment - yes we have deprecated the V2 api several minor versions ago and have started to delete some of the code per our deprecation policy (with the rest being deleted hopefully soon).

anthonyburdi: Also, I would suggest adding the `deprecated_argument` decorator to this method for `batch_kwargs_generator_names`, looks like we missed adding that when adding this method to the public API.

anthonyburdi: Also, this does not look like it will break existing v2 api workflows yet, there are isinstance checks for `FluentDatasource`.

anthonyburdi (approved):
This looks like it will work. The get_available_data_asset_names() method is a bit hard to follow with all the nesting, I am going to approve but if you are able to refactor I think it could be made easier to separate the v2 / v3 / fluent logic so that we can remove it later.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3917,27 +3917,28 @@ def get_available_data_asset_names(
...
-                        batch_kwargs_generator_names[idx]
-                    )
+                    if isinstance(datasource, FluentDatasource):
+                        data_asset_names[datasource_name] = datasource.get_asset_names()
+                    else:
+                        data_asset_names[
+                            datasource_name
+                        ] = datasource.get_available_data_asset_names(
+                            batch_kwargs_generator_names[idx]
+                        )
```
Kilo59, line 3933:
```
great_expectations/data_context/data_context/abstract_data_context.py:3925: error:
Incompatible types in assignment (expression has type
"Union[Dict[str, List[str]], Any]", target has type "Set[str]")  [assignment]
                            ] = datasource.get_available_data_asset_names(
                                ^
  ```
You are assigning a `Set[str]` to a `List[str]`. Easiest fix here is to convert the `Set` to a `List`.
```python
data_asset_names[
     datasource_name
    ] = list(datasource.get_available_data_asset_names(
     batch_kwargs_generator_names[idx]
                        ))
```

NathanFarmer (approved):

tests/data_context/test_data_context_datasources.py
```diff
@@ -453,3 +453,21 @@ def test_list_datasources() -> None:
     context.sources.add_pandas(datasource_name)
 
     assert len(context.list_datasources()) == 2
+
+
+@pytest.mark.unit
+def test_get_available_data_assets_names(test_df_pandas, empty_data_context) -> None:
```
cdkini: As `empty_data_context` scaffolds the filesystem as part of setup, I wouldn't consider this a unit test but rather an integration test - getting true, pure unit tests with the context is difficult work

tests/data_context/test_data_context_datasources.py
```diff
@@ -453,3 +453,21 @@ def test_list_datasources() -> None:
...
+
+    assert len(empty_data_context.get_available_data_asset_names()) == 1
+
+    data_asset_names = dict(
+        empty_data_context.get_available_data_asset_names(
+            datasource_names=datasource_name
+        )
+    )
+
+    assert "test_data_frame" in str(data_asset_names.values())
```
cdkini: Nit - instead of using two separate strings, can we do something like this:

```
asset_name = "test_data_frame"
...
datasource.add_dataframe_asset(name=asset_name)
...
assert asset_name in data_asset_names.values()
```

cdkini: also this casting to a string on L473 shouldn't be necessary

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3897,7 +3897,7 @@ def get_available_data_asset_names(
         Raises:
             ValueError: `datasource_names` is not None, a string, or list of strings.
         """
-        data_asset_names = {}
+        data_asset_names: dict[str, list[str]] = {}
```
cdkini: Could we annotate the return value on L3886? Do we know what to expect now?

Our typing of the OSS codebase is gradual so let's chip away at it if its not too big of a lift

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3917,27 +3917,31 @@ def get_available_data_asset_names(
...
-                    ), 'Method "get_available_data_asset_names" not implemented for FluentDatasource'
-                    data_asset_names[
-                        datasource_name
-                    ] = datasource.get_available_data_asset_names(
-                        batch_kwargs_generator_names[idx]
-                    )
+                    if isinstance(datasource, FluentDatasource):
+                        data_asset_names[datasource_name] = list(
+                            datasource.get_asset_names()
+                        )
```
cdkini: Nit - probably not super necessary but if we're converting from a set to a list, can we use `sorted()`? It would be nice to get consistent ordering here since we're not guaranteed that in a set.

Same throughout

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3948,15 +3952,17 @@ def get_available_data_asset_names(
...
+                    else:
+                        data_asset_names[datasource_name] = list(
+                            datasource.get_available_data_asset_names()
+                        )
                 except ValueError:
                     # handle the edge case of a non-existent datasource
-                    data_asset_names[datasource_name] = {}
+                    data_asset_names[datasource_name] = list()
```
cdkini: Nit - can we use `[]`? Either works but we tend to use this syntax in our codebases

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: yeah this shouldn't break V2 - that being said, I don't want to add scope creep to @jshaikGX 's work here.

Are we confident in this deprecation? Should we surface in arch to confirm?

tests/data_context/test_data_context_datasources.py
```diff
@@ -453,3 +453,22 @@ def test_list_datasources() -> None:
...
+
+    assert len(empty_data_context.get_available_data_asset_names()) == 1
+
+    data_asset_names = dict(
+        empty_data_context.get_available_data_asset_names(
+            datasource_names=datasource_name
+        )
+    )
+
+    assert asset_name in data_asset_names["my_fluent_pandas_datasource"]
```
cdkini: Nit - let's use `datasource_name` as the key

PR #7724, opened 2023-04-25
[MAINTENANCE] Use same version of mypy in contrib tool

kenwade4:
Let gx and the contrib tool be installed together again.

```
ERROR: Cannot install great-expectations-contrib==0.1.0 and great-expectations[dev]==0.16.8+76.g8e7dd99db.dirty because these package versions have conflicting dependencies.

The conflict is caused by:
    great-expectations-contrib 0.1.0 depends on mypy==1.1.1
    great-expectations[dev] 0.16.8+76.g8e7dd99db.dirty depends on mypy==1.2.0; extra == "dev"
```

- Use same version of mypy in contrib tool


anthonyburdi (approved):
Good catch!

PR #7745, opened 2023-04-27
[BUGFIX] Patch expectation registry

cdkini:
- Move core registration closer to where it is actually used

### Definition of Done

### DIFF
--- a/great_expectations/expectations/registry.py
+++ b/great_expectations/expectations/registry.py
@@ -169,7 +169,7 @@ def register_expectation(expectation: Type[Expectation]) -> None:
     _registered_expectations[expectation_type] = expectation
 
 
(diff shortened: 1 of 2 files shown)


great_expectations/expectations/registry.py
```diff
@@ -399,6 +399,9 @@ def get_expectation_impl(expectation_name: str) -> Type[Expectation]:
         )
         expectation_name = renamed[expectation_name]
 
+    if not _registered_expectations:
+        _register_core_expectations()
```
cdkini, line 404: Move this logic closer to where it is used - if we try to access `_registred_expectations`, this should be run before

Kilo59 (approved):

PR #7754, opened 2023-04-27
[FEATURE] Spark parquet reader support for fluent datasources

anthonyburdi:
- Add `ParquetAsset` and `add_parquet_asset` method with available read options from spark: https://spark.apache.org/docs/latest/sql-data-sources-parquet.html.
- TODO or considered in future PRs:
    - The rest of the spark methods after this pattern is verified.
    - Parametrizing tests and adding additional unit tests.
    - Adding **kwargs for as-yet defined parameters (e.g. future versions).

### Definition of Done


great_expectations/datasource/fluent/spark_file_path_datasource.py
```diff
@@ -35,10 +36,26 @@ def _get_reader_method(self) -> str:
     def _get_reader_options_include(self) -> set[str] | None:
         return {"header", "infer_schema"}
 
+class ParquetAsset(_FilePathDataAsset):
```
Kilo59, line 39: Looks like you are defining this asset in two places.
I would keep the one defined here and get rid of the `from great_expectations.datasource.fluent.pandas_datasource import ParquetAsset` import

anthonyburdi: All cleaned up - I'll add the rest of the options and then submit. Thanks for the keen eyes on the wrong literal 🙇

great_expectations/datasource/fluent/spark_file_path_datasource.py
```diff
@@ -36,9 +36,36 @@ def _get_reader_options_include(self) -> set[str] | None:
...
+    def _get_reader_options_include(self) -> set[str] | None:
+        """These options are available as of spark v3.4.0
+
+        See https://spark.apache.org/docs/latest/sql-data-sources-parquet.html for more info.
+        """
+        return {
+            "datetimeRebaseMode",
+            "int96RebaseMode",
+            "mergeSchema"
+        }
```
anthonyburdi: Should this instead use the snake case version e.g. `merge_schema`?

anthonyburdi: No I don't think so, they should use the original spark names like in this PR. These options are passed on to spark directly via the batch_spec here: https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/execution_engine/sparkdf_execution_engine.py#L303

tests/datasource/fluent/test_spark_filesystem_datasource.py
```diff
@@ -83,6 +83,25 @@ def test_add_csv_asset_to_datasource(
     assert m1 is not None
 
 
+# TODO: Parametrize with all asset types
```
alexsherstinsky, line 86: @anthonyburdi Do we need to add a test for other storage possibilities (e.g., S3, etc.), or is this sufficient?

tests/datasource/fluent/test_spark_filesystem_datasource.py
```diff
@@ -83,6 +83,25 @@ def test_add_csv_asset_to_datasource(
...
+@pytest.mark.unit
+def test_add_parquet_asset_to_datasource(
+    spark_filesystem_datasource: SparkFilesystemDatasource,
+):
+    asset = spark_filesystem_datasource.add_parquet_asset(
+        name="parquet_asset",
+        datetime_rebase_mode="EXCEPTION",
```
alexsherstinsky, line 93: @anthonyburdi To me an `Enum` would appear better, but I am not sure if this would add more work to users.

alexsherstinsky (approved):
LGTM -- @anthonyburdi I added two non-blocking questions/comments.  Thanks!

tests/datasource/fluent/test_spark_filesystem_datasource.py
anthonyburdi: I'm not sure yet, I'll be doing a deeper dive on testing as I address the other assets. My gut says no, that we should test accessing other storage possibilities separately rather than for each asset.

tests/datasource/fluent/test_spark_filesystem_datasource.py
anthonyburdi: That's a good point. I think since this is a pass through to spark and we have pydantic runtime validation for the acceptable values that a string may be good enough here. But I would love to hear team thoughts e.g. @Kilo59

great_expectations/datasource/fluent/spark_file_path_datasource.py
```diff
@@ -36,9 +36,36 @@ def _get_reader_options_include(self) -> set[str] | None:
...
+
+    def _get_reader_method(self) -> str:
+        return self.type
+
+    def _get_reader_options_include(self) -> set[str] | None:
+        """These options are available as of spark v3.4.0
+
+        See https://spark.apache.org/docs/latest/sql-data-sources-parquet.html for more info.
+        """
+        return {"datetimeRebaseMode", "int96RebaseMode", "mergeSchema"}
```
billdirks, line 63: This there a way to introspect the pyspark version and get this? Our min version is 2.3.2. Or maybe this doesn't matter since this is only used internally?

anthonyburdi: It's a good point, probably yes there is a way (would have to investigate). I thought for this near term effort (spark) we were going to manually create these assets since there are relatively few.

great_expectations/datasource/fluent/spark_file_path_datasource.py
```diff
@@ -36,9 +36,36 @@ def _get_reader_options_include(self) -> set[str] | None:
...
+
+
 class _SparkFilePathDatasource(_SparkDatasource):
     # class attributes
-    asset_types: ClassVar[List[Type[DataAsset]]] = [CSVAsset]
+    asset_types: ClassVar[List[Type[DataAsset]]] = [CSVAsset, ParquetAsset]
 
     # instance attributes
     assets: List[_FilePathDataAsset] = []  # type: ignore[assignment]
```
Kilo59:
```suggestion
    assets: List[Union[CSVAsset, ParquetAsset]] = []  # type: ignore[assignment]
```

This will give us a better schemas and runtime validation.
The `assets` can't be any `_FilePathDataAsset` type, they will only ever be a `CSVAsset`, or `ParquetAsset`

anthonyburdi: Good call! I ended up putting it in a variable so I could reference elsewhere: `_SPARK_FILE_PATH_ASSET_TYPES` - happy to rename if there is a better name.

PR #7764, opened 2023-04-28
[DOCS] Temporarily pin ipython for python 3.8 before building api docs

anthonyburdi:
- API docs builds were failing due to an incompatibility with ipython. This was also affecting GX. Since the api docs use the latest `released` version of GX, they will not benefit from the fix. This temporary pin will allow API docs to be built but should be removed in the next release because the change was made in `requirements.txt`

### Definition of Done

### DIFF
--- a/docs/build_docs
+++ b/docs/build_docs
@@ -72,6 +72,11 @@ echo -e "${ORANGE}Installing Great Expectations library dev dependencies.${NC}"
 echo -e "${ORANGE}Installing api docs dependencies.${NC}"
 (cd ../sphinx_api_docs_source; pip install -r requirements-dev-api-docs.txt)
 
+# TODO: Remove the below temporary install after v0.16.10 is released:
+echo -e "${ORANGE}Temporarily install compatible version of Ipython (remove this after v0.16.10 is released).${NC}"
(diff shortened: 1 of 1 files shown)


cdkini (approved):

docs/build_docs
```diff
@@ -72,6 +72,11 @@ echo -e "${ORANGE}Installing Great Expectations library dev dependencies.${NC}"
 echo -e "${ORANGE}Installing api docs dependencies.${NC}"
 (cd ../sphinx_api_docs_source; pip install -r requirements-dev-api-docs.txt)
 
+# TODO: Remove the below temporary install after v0.16.10 is released:
+echo -e "${ORANGE}Temporarily install compatible version of Ipython (remove this after v0.16.10 is released).${NC}"
+# Due to https://github.com/ipython/ipython/issues/14053
+(pip install --upgrade 'ipython<8.13.0; python_version <= "3.8"')
```
Kilo59: Will this work if the installed version is more recent?

```suggestion
(pip install 'ipython==8.12.1; python_version <= "3.8"')
```

anthonyburdi: I think so but it's not necessary. Removed `--upgrade`. Good catch!

PR #7769, opened 2023-04-28
[MAINTENANCE] Cleanup of Fluent `BatchRequest` type and immutability constraints

NathanFarmer:
- DX-471
- Narrow type constraints on `BatchRequest` model
- Create `SliceValidator` for pydantic validation of `slice` built-in
- Add `BatchRequest` round-trip serialization test
- Move Fluent `BatchRequest` and `BatchRequestOptions` from `interfaces.py` to `batch_request.py`
- Create `BatchRequest.json` schema
- Add Fluent `BatchRequest` to `@public_api`
- Add logging to `parse_batch_slice`

### Definition of Done


great_expectations/datasource/fluent/spark_datasource.py
```diff
@@ -134,7 +134,7 @@ def _validate_batch_request(self, batch_request: BatchRequest) -> None:
                 datasource_name=self.datasource.name,
                 data_asset_name=self.name,
                 options={},
-                batch_slice=batch_request.batch_slice,
+                batch_slice=batch_request.batch_slice_input,  # type: ignore[attr-defined]
```
donaldheppner: we should not have "batch_slice_input" as n attribute, and certainly not a public one. What is it required?

donaldheppner (changes requested):

tests/datasource/fluent/integration/test_integration_datasource.py
```diff
@@ -462,3 +463,43 @@ def test_asset_specified_metadata(
     assert len(batches) == 1
     # Update the batch_metadata from the request with the metadata inherited from the asset
     assert batches[0].metadata == {**asset_specified_metadata, "year": 2019, "month": 2}
+
+
+@pytest.mark.integration
+def test_batch_request_error_messages(
```
donaldheppner, line 469: this all looks good to me!

tests/datasource/fluent/data_asset/data_connector/test_google_cloud_storage_data_connector.py
```diff
@@ -9,10 +9,10 @@
 from great_expectations.core import IDDict
 from great_expectations.core.batch import BatchDefinition
 from great_expectations.core.util import GCSUrl
+from great_expectations.datasource.fluent.batch_request import BatchRequest
```
donaldheppner: could the `__init__.py` be updated to avoid having to import BatchRequest from `batch_request`

donaldheppner (changes requested):

great_expectations/datasource/fluent/spark_datasource.py
NathanFarmer: This is how we are reversing the serialization of the input strings, ints, lists, etc. I want to make this (more) private, but pydantic does weird things when you add the preceeding underscore to variables. We are only using this internally and it requires the type ignore you see on this line. This attribute is not included in the stub file, so it will not show up in autocomplete.

tests/datasource/fluent/data_asset/data_connector/test_google_cloud_storage_data_connector.py
NathanFarmer: Yes! Right now we don't have any public API docs that show importing this class, so I opted to keep it narrowly namespaced at first. I will update the `__init__`.

NathanFarmer: ✅

great_expectations/datasource/fluent/spark_datasource.py
NathanFarmer: I was able to figure this out!

great_expectations/datasource/fluent/batch_request.pyi
Kilo59, line 1: Why do we need a stub file for this?

NathanFarmer: Because mypy doesn't pick up on the fact that you can set the `batch_slice` attribute. Pydantic requires us to use the `property_set_methods` attr in `Config` instead of a traditional setter.

great_expectations/datasource/fluent/batch_request.py
```diff
@@ -0,0 +1,225 @@
...
+        allow_mutation=True,
+        description=(
+            "A map that can be used to filter the batch groups associated with the Data Asset. "
+            "The structure and types depends on the asset type."
+        ),
+    )
+    _batch_slice_input: Optional[BatchSlice] = pydantic.Field(
+        default=None,
+        allow_mutation=True,
+    )
```
Kilo59: Can we just combine a pydantic `PrivateAttr` with traditional `@property` to deal with this, and then get rid of the stub and the all the special `Config` options?
https://docs.pydantic.dev/latest/usage/models/#private-model-attributes

great_expectations/datasource/fluent/batch_request.pyi
Kilo59: You should be able to use regular `@property`'s with setters and getters with pydantic, without mypy complaining.

great_expectations/datasource/fluent/batch_request.py
Kilo59: Can we make `batch_slice` a regular pydantic field and "validate_on_assignment"?

NathanFarmer: This is a private attr (`underscore_attrs_are_private=True`). I'm not sure what other config options you are referring to, but we need a way to set the input of `batch_request.batch_slice` on `_batch_slice_input`.

NathanFarmer: We need to save the `_batch_slice_input` to serialize it out in the `batch_slice` json property.

Kilo59: Some of this pydantic configuration seems a bit overly complicated, I think it's probably worth taking another look to see what can be simplified. But I don't want to block on this, so I'm approving.

great_expectations/datasource/fluent/batch_request.py
```diff
@@ -0,0 +1,225 @@
...
+            type_=Optional[BatchSlice],  # type: ignore[arg-type]
+            required=False,
+            default=None,
+            model_config=self.__config__,
+            class_validators=None,
+        )
+        property_set_methods = self.__config__.property_set_methods  # type: ignore[attr-defined]
+        self.__config__.property_set_methods = {}  # type: ignore[attr-defined]
+        self.__setattr__("batch_slice", self._batch_slice_input)
+        self.__config__.property_set_methods = property_set_methods  # type: ignore[attr-defined]
```
Kilo59: Could you add a comment explaining why we are doing this?

NathanFarmer: ✅

great_expectations/datasource/fluent/batch_request.py
Kilo59: >This is a private attr (underscore_attrs_are_private=True).


Elsewhere in the codebase we explicitly use `pydantic.PrivateAttr` to make fields private, so I think doing it via config like this could be confusing.

https://github.com/great-expectations/great_expectations/blob/b18f1af59add5c0479aa6f5dfc04082cbf3e66bf/great_expectations/datasource/fluent/interfaces.py#L265-L267


------------------

>but we need a way to set the input of `batch_request.batch_slice` on `_batch_slice_input`

I don't think we need this extra `_batch_slice_input` attribute at all. Can we just use a custom validator on `batch_slice` and do whatever needed parsing logic there?

NathanFarmer: I reverted the Config `underscore_attrs_are_private` change to a `PrivateAttr` like you suggested. We used to need the alias until that approach no longer worked.

NathanFarmer: The purpose of saving the input is so we can serialize it back out to config. We want the serialization/deserialization to be symmetrical, but if we lose the input upon deserialization we have to serialize something like `slice(start, stop, step)` back out to config.

great_expectations/datasource/data_connector/batch_filter.py
```diff
@@ -15,7 +16,61 @@
 logger = logging.getLogger(__name__)
 
 
-BatchSlice: TypeAlias = Union[slice, str, List[int], Tuple[int, ...], int]
+class SliceValidator:
```
Kilo59, line 19: Would `SerializableSlice` be a better name?
Or `GxSlice` ??? 
🤷 


It is just a slice that we can serialize right?

Kilo59 (approved):
Have some non-blocking thoughts/ideas for followups but I'm okay with this as is.

great_expectations/datasource/fluent/batch_request.pyi
NathanFarmer: I agree that you should be able to, but you can't until pydantic 2.0: https://github.com/pydantic/pydantic/issues/3395

NathanFarmer: It's not just mypy. Pydantic doesn't support setters in <2.0.

great_expectations/datasource/data_connector/batch_filter.py
NathanFarmer: It is primarily so you can validate a `slice`. Pydantic doesn't support built-in `slice` validation. It also defines the json schema, but serialization is handled elsewhere.

donaldheppner (approved):

PR #7771, opened 2023-04-28
[BUGFIX] Patch Expectation registry issue introduced in 0.16.9

cdkini:
- Our call to `register_core_expectations` in our top-level `__init__` solves the issue we were seeing
    - Note that this was introduced after the 0.16.9 release so the fix is already in
- Deleting the extra call in `Validator` as we'll already have the registery populated by this time

### Definition of Done

### DIFF
--- a/great_expectations/__init__.py
+++ b/great_expectations/__init__.py
@@ -24,6 +24,12 @@ from .util import (
     validate,
 )
 
+# By placing this registry function in our top-level __init__,  we ensure that all
+# GX workflows have populated expectation registries before they are used.
+#
+# Both of the following import paths will trigger this file, causing the registration to occur:
+#   import great_expectations as gx
+#   from great_expectations.core import ExpectationSuite, ExpectationConfiguration
 register_core_expectations()
 
 # from great_expectations.expectations.core import *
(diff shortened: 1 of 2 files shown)


great_expectations/validator/validator.py
```diff
@@ -194,9 +193,6 @@ def __init__(
         include_rendered_content: Optional[bool] = None,
         **kwargs,
     ) -> None:
-        # Ensure that Expectations are available for __dir__ and __getattr__
-        register_core_expectations()
```
cdkini, line 198: No longer needed since we register at the top-level `__init__`

cdkini: Confirmed locally

Shinnnyshinshin (approved):
👍🏼

great_expectations/__init__.py
```diff
@@ -24,6 +24,12 @@
     validate,
 )
 
+# By placing this registry function in our top-level __init__,  we ensure that all
+# GX workflows have populated expectation registries before they are used.
+#
+# Both of the following import paths will trigger this file, causing the registration to occur:
+#   import great_expectations as gx
+#   from great_expectations.core import ExpectationSuite, ExpectationConfiguration
```
Shinnnyshinshin, line 32: Thank you for this note 🙇🏼

PR #7776, opened 2023-04-28
[MAINTENANCE] Enable `flake8-bugbear` rules

Kilo59:
https://beta.ruff.rs/docs/rules/#flake8-bugbear-b

Initially only fixing or ignoring the low warning count rules and temporarily disabling rules with +10 violations

- Enable the `BXX` ruff rules.


great_expectations/jupyter_ux/__init__.py
```diff
@@ -136,14 +136,17 @@ def set_data_source(context, data_source_type=None):
     return data_source_name
 
 
+_LOCAL_ZONE = tzlocal.get_localzone()
```
Kilo59, line 139:
```suggestion
_LOCAL_TZ = tzlocal.get_localzone()
```

great_expectations/jupyter_ux/expectation_explorer.py
```diff
@@ -564,7 +564,11 @@ def generate_tag_button_list(self, expectation_state, tag_list, widget_display):
         )
 
     def generate_zero_or_positive_integer_widget(
-        self, value, max=int(9e300), description="", continuous_update=False
+        self,
+        value,
+        max=int(9e300),  # noqa: B008 # function-call-in-default-argument
```
Kilo59, line 569: Why this number?
I want to save this number to a module-level constant with a descriptive name.

pyproject.toml
```diff
@@ -256,6 +258,12 @@ ignore = [
     "TID252",  # Relative imports from parent modules are banned
     # https://beta.ruff.rs/docs/rules/#flake8-use-pathlib-pth
     "PTH123", # pathlib-open - this would force pathlib usage anytime open or with open was used.
+    # https://beta.ruff.rs/docs/rules/#flake8-bugbear-b
+    # TODO: enable these
+    "B904", # raise-without-from-inside-except
+    "B028", # no-explicit-stacklevel - https://beta.ruff.rs/docs/rules/no-explicit-stacklevel/
+    "B007", # unused-loop-control-variable
+    "B009", # get-attr-with-constant
```
Kilo59, line 266: These rules have a significant number of violations.
Too many to properly review in this PR.
```
$ ruff . --statistics
249	B904	[ ] Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling
 82	B028	[ ] No explicit `stacklevel` keyword argument found
 38	B007	[*] Loop control variable `attr` not used within loop body
 15	B009	[*] Do not call `getattr` with a constant attribute value. It is not any safer than normal property access.
```

tests/ruff.toml
```diff
@@ -10,6 +10,8 @@ extend-ignore = [
     "TCH001",
     # https://beta.ruff.rs/docs/rules/#flake8-tidy-imports-tid
     "TID251",  # banned-api - is used to prevent opaque attribute errors caused by missing optional dependencies
+    # https://beta.ruff.rs/docs/rules/#flake8-bugbear-b
+    "B011", # assert-false - common pattern in pytest
```
Kilo59, line 14: Disabling this rule for tests because `assert False` is a common thing to do in tests to force a failure.

tests/performance/taxi_benchmark_util.py
```diff
@@ -314,8 +316,10 @@ def _add_checkpoint(
     datasource_name: str,
     data_connector_name: str,
     checkpoint_name: str,
-    suite_and_asset_names=[],
+    suite_and_asset_names: list | None = None,
 ) -> SimpleCheckpoint:
+    if suite_and_asset_names is None:
+        suite_and_asset_names = []
```
anthonyburdi, line 322: 🎉

tests/render/test_data_documentation_site_builder.py
```diff
@@ -289,7 +289,7 @@ def test_configuration_driven_site_builder(
     # verify that an additional validation result HTML file was generated
     assert len(index_links_dict["validations_links"]) == 2
 
-    site_builder.site_index_builder.target_store.store_backends[
+    _ = site_builder.site_index_builder.target_store.store_backends[
```
anthonyburdi, line 292: Ambivalent about this one. I guess it is meant to make it obvious that this value is not used. If it isn't used in this test we can remove this line (unless there are side effects). Just thinking out loud & trying to understand, I think this is fine for the scope of the PR.

great_expectations/validation_operators/validation_operators.py
```diff
@@ -190,7 +194,7 @@ def __init__(
         data_context,
         action_list,
         name,
-        result_format={"result_format": "SUMMARY"},
+        result_format={"result_format": "SUMMARY"},  # noqa: B006 # mutable default
```
anthonyburdi, line 197: 😬

anthonyburdi (approved):
Looks good just a couple comments / suggestions

great_expectations/core/expectation_diagnostics/expectation_test_data_cases.py
```diff
@@ -78,7 +78,7 @@ def __init__(
         title,
         exact_match_out,
         out,
-        suppress_test_for=[],
+        suppress_test_for=[],  # noqa: B006 # mutable default
         **kwargs,
     ) -> None:
         super().__init__(
```
anthonyburdi:
```suggestion
        suppress_test_for=None,
        **kwargs,
    ) -> None:
        if not suppress_test_for:
            suppress_test_for = []
        super().__init__(
```

tests/render/test_data_documentation_site_builder.py
Kilo59: At a minimum, it ensures that accessing this chain of attributes and keys doesn't cause an Attribute or Key Error.

But I don't know the original intention behind this, so I opted to make the most minimal change possible.

PR #7777, opened 2023-04-28
[FEATURE] Spark read directory of files as a single batch for CSV

anthonyburdi:
- Added functionality to read a directory of files as a single batch.
- This is implemented currently for CSV but will be extended to other types of files that spark is capable of reading (e.g. parquet, orc, etc) in an upcoming PR.

### Definition of Done


great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
```diff
@@ -73,6 +73,7 @@ def __init__(
         # sorters: Optional[list] = None,
         # TODO: <Alex>ALEX</Alex>
         file_path_template_map_fn: Optional[Callable] = None,
+        get_unfiltered_batch_definition_list_fn: None | Callable = None,
```
billdirks: Can you update this to remove the `None` value in the signature? This is the place where it ultimately needs to be set. Also, we also know how this function will be called so it would be good to annotate the `Callable`. That is:

```
get_unfiltered_batch_definition_list_fn: Callable[[FilePathDataConnector, BatchRequest], list[BatchDefinition]] = file_get_unfiltered_batch_definition_list_fn
```

You could then hopefully also replace the other `Callable` annotations with this typed version.

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
```diff
@@ -90,6 +91,16 @@ def __init__(
 
         self._file_path_template_map_fn: Optional[Callable] = file_path_template_map_fn
 
+        if get_unfiltered_batch_definition_list_fn:
```
billdirks: Doing the above should let you remove this conditional.

great_expectations/datasource/fluent/spark_file_path_datasource.py
```diff
@@ -36,9 +36,31 @@ def _get_reader_options_include(self) -> set[str] | None:
...
+
+    class Config:
+        extra = pydantic.Extra.forbid
+        allow_population_by_field_name = True
+
+    def _get_reader_method(self) -> str:
+        # Reader method is still "csv"
+        return self.type.replace("directory_", "")
+
+    def _get_reader_options_include(self) -> set[str] | None:
```
billdirks, line 55: This function always returns a `set` so do you need the `| None`? Is it a requirement from an interface it is implementing?

great_expectations/datasource/fluent/spark_filesystem_datasource.py
```diff
@@ -56,20 +61,33 @@ def test_connection(self, test_assets: bool = True) -> None:
...
+        data_asset: _SPARK_FILE_PATH_ASSET_TYPES,
+        glob_directive: str = "**/*",
+        **kwargs,
     ) -> None:
         """Builds and attaches the `FilesystemDataConnector` to the asset."""
         if kwargs:
             raise TypeError(
                 f"_build_data_connector() got unexpected keyword arguments {list(kwargs.keys())}"
             )
+        if hasattr(data_asset, "data_directory"):
```
billdirks: Could we do an isinstance check here? It makes it clearer that we are matching over the possible types of the `data_asset`.

billdirks (commented):
Thanks for doing this! I've added a few comments inline.

great_expectations/datasource/fluent/spark_file_path_datasource.py
anthonyburdi: It is required based on `_FilePathDataAsset` `def _get_reader_options_include(self) -> Set[str] | None:`
I thought about removing the `None` everywhere, but we do set the return value to None in `_create_pandas_asset_model`. Though I'm not sure that is necessary - any ideas on the implications of using an empty set instead of None in `_create_pandas_asset_model`?

great_expectations/datasource/fluent/spark_filesystem_datasource.py
anthonyburdi: Agreed. I was trying to avoid keeping a tuple of directory related asset classes, but I agree this is clearer.

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
anthonyburdi: @billdirks, just a question for my own understanding - when adding types to the Callable annotation it looks like [you can only use positional arguments](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#callable-types-and-lambdas) with the function otherwise mypy will complain that keyword arguments are supplied. Should we be moving toward positional arguments instead of keyword? Also `directory_get_unfiltered_batch_definition_list_fn` takes an additional parameter `data_directory: str` - is there a way to handle that in the type hint?

great_expectations/datasource/fluent/spark_filesystem_datasource.py
```diff
@@ -56,20 +65,36 @@ def test_connection(self, test_assets: bool = True) -> None:
...
                 f"_build_data_connector() got unexpected keyword arguments {list(kwargs.keys())}"
             )
+        if isinstance(data_asset, _SPARK_DIRECTORY_ASSET_CLASSES):
+            get_unfiltered_batch_definition_list_fn = cast(
+                Callable[[FilePathDataConnector, BatchRequest], List[BatchDefinition]],
+                partial(
+                    directory_get_unfiltered_batch_definition_list_fn,
+                    data_directory=data_asset.data_directory,
+                ),
+            )
```
anthonyburdi: I was also considering using a Protocol here for the function partial instead of the cast (e.g. https://stackoverflow.com/a/66664521/2916375) -- open to ideas.

great_expectations/datasource/fluent/spark_file_path_datasource.py
billdirks: I don't know the implications of using an empty set instead of None. I am fine keeping it as is.

great_expectations/datasource/fluent/spark_filesystem_datasource.py
billdirks: Oh I didn't know `partials` don't return nice `callables` like we want. I think the cleanest solution might be sidestepping the use of partials and instead replace the definition of `directory_get_unfiltered_batch_definition_list_fn` with something like:
```
def make_directory_get_unfiltered_batch_definition_list_fn(data_directory: str) -> Callable[[FilePathDataConnector, BatchRequest], list[BatchDefinition]]:

    def directory_get_unfiltered_batch_definition_list_fn(data_connector: FilePathDataConnector, batch_request: BatchRequest):
        batch_definition_list: List[BatchDefinition] = list()
        batch_definition = BatchDefinition(
            datasource_name=data_connector.datasource_name,
            data_connector_name=_DATA_CONNECTOR_NAME,
            data_asset_name=data_connector.data_asset_name,
            batch_identifiers=IDDict({"path": data_directory}),
        )
        batch_definition_list.append(batch_definition)

        return [batch_definition]

    return directory_get_unfiltered_batch_definition_list_fn
```
Then we could make this line:
```
get_unfiltered_batch_definition_list_fn = make_directory_get_unfiltered_batch_definition_list_fn(data_asset.data_directory)
```

anthonyburdi: Good call, done!

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
billdirks: Oh, I didn't see this comment before. I think we've already answered your second one. For your first question (positional vs keyword arguments), I think how we are doing it now is fine. This has to do with how the function is defined vs called. Calling the function with keyword arguments is clear. `Callable` enforces the order of the inputs/outputs must be fixed vs specifying (name, type) pairs. That doesn't seem like a burden for us so is fine too. Hopefully, this answers your question. We can chat synchronously if that would help.

billdirks (approved):
LGTM, thanks!

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
anthonyburdi: This is great - thank you! I agree that what we now have in the PR works. We can revisit if there are methods that we want to call with keyword arguments.

great_expectations/datasource/fluent/spark_file_path_datasource.py
Kilo59: I think the "interface" should always return a `set` which could be empty. The `None` just makes it harder to work with.

great_expectations/datasource/fluent/data_asset/data_connector/dbfs_data_connector.py
```diff
@@ -3,11 +3,19 @@
...
     FilesystemDataConnector,
 )
+from great_expectations.datasource.fluent.data_asset.data_connector.file_path_data_connector import (
+    FilePathDataConnector,
+    file_get_unfiltered_batch_definition_list_fn,
+)
+
+if TYPE_CHECKING:
+    from great_expectations.core.batch import BatchDefinition
+    from great_expectations.datasource.fluent.interfaces import BatchRequest
```
Kilo59:
```suggestion
    from great_expectations.datasource.fluent import BatchRequest
```

great_expectations/datasource/fluent/data_asset/data_connector/filesystem_data_connector.py
```diff
@@ -14,6 +14,13 @@
 from great_expectations.datasource.fluent.data_asset.data_connector import (
     FilePathDataConnector,
 )
+from great_expectations.datasource.fluent.data_asset.data_connector.file_path_data_connector import (
+    file_get_unfiltered_batch_definition_list_fn,
+)
+
+if TYPE_CHECKING:
+    from great_expectations.core.batch import BatchDefinition
+    from great_expectations.datasource.fluent.interfaces import BatchRequest
```
Kilo59:
```suggestion
    from great_expectations.datasource.fluent import BatchRequest
```

great_expectations/datasource/fluent/schemas/SparkDBFSDatasource.json
```diff
@@ -132,8 +132,8 @@
             ],
             "additionalProperties": false
         },
-        "ParquetAsset": {
-            "title": "ParquetAsset",
+        "DirectoryCSVAsset": {
+            "title": "DirectoryCSVAsset",
             "description": "Base model for most fluent datasource related pydantic models.\n\nAdds yaml dumping and parsing methods.\n\nExtra fields are not allowed.\n\nSerialization methods default to `exclude_unset = True` to prevent serializing\nconfigs full of mostly unset default values.\nAlso prevents passing along unset kwargs to BatchSpec.\nhttps://docs.pydantic.dev/usage/exporting_models/",
```
Kilo59: Why is this replacing the ParquetAsset? Shouldn't it just be adding an additional definition here?
I think you may need to run `invoke schema --sync` again.

great_expectations/datasource/fluent/spark_file_path_datasource.py
```diff
@@ -36,6 +36,25 @@ def _get_reader_options_include(self) -> set[str] | None:
         return {"header", "infer_schema"}
 
 
+class DirectoryCSVAsset(_FilePathDataAsset):
+    # Overridden inherited instance fields
+    type: Literal["directory_csv"] = "directory_csv"
+    data_directory: str
```
Kilo59: Can we make this a `pathlib.Path` or better yet a `DirectoryPath` pydantic type.
This will ensure the value exists and is a directory, and error if it isn't.

https://docs.pydantic.dev/latest/usage/types/#pydantic-types

Kilo59:
@anthonyburdi disabled auto-merge because I think there may be a couple of issues. Feel free to re-enable.

great_expectations/datasource/fluent/spark_file_path_datasource.py
anthonyburdi: Totally agree - happy to change it, just wanted to make sure there were no other implications first.

great_expectations/datasource/fluent/schemas/SparkDBFSDatasource.json
anthonyburdi: I ran both before and after deleting all of the schemas and both times there were no changes locally (relative to the PR - i.e. ParquetAsset is replaced). I looked a bit at `invoke schema` and I haven't been able to figure out what is missing - am I not registering ParquetAsset somewhere? I'll keep looking but might be good to pair here.

anthonyburdi: Nevermind - I think I know what is missing. ParquetAsset wasn't in `_SPARK_FILE_PATH_ASSET_TYPES` (I think I missed that during merge conflict resolution). Yep that was it. Commit incoming.

Kilo59 (approved):

great_expectations/datasource/fluent/spark_file_path_datasource.py
anthonyburdi: Just to follow up, went with pathlib.Path for now since the data_directory is relative to the base_path so the data_directory by itself will fail the existence check. In a future PR we can validate the full path (base_path + data_directory)

great_expectations/datasource/fluent/spark_file_path_datasource.py
anthonyburdi: See PR https://github.com/great-expectations/great_expectations/pull/7797 for this change

PR #7791, opened 2023-05-02
[BUGFIX] Fix FDS Sqlite config round tripping

Kilo59:
- rename the `type` field for sqlite assets to be to `table`, `query` instead of `sqite_table`, `sqlite_query`
- Fixes the `sqlite` asset round config round tripping errors

### DIFF
--- a/great_expectations/datasource/fluent/sqlite_datasource.py
+++ b/great_expectations/datasource/fluent/sqlite_datasource.py
@@ -143,12 +143,12 @@ class _SQLiteAssetMixin:
 
 
 class SqliteTableAsset(_SQLiteAssetMixin, SqlTableAsset):
-    type: Literal["sqlite_table"] = "sqlite_table"  # type: ignore[assignment]  # override superclass value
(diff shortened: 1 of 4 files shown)


great_expectations/datasource/fluent/sqlite_datasource.py
Kilo59, line 1: These changes on their own ended up being enough to fix the issues.

Screenshots below of the behavior before these changes.

image

image

great_expectations/datasource/fluent/sqlite_datasource.py
```diff
@@ -215,9 +215,3 @@ def add_query_asset(
         )
 
     add_query_asset.__doc__ = SQLDatasource.add_query_asset.__doc__
-
-
-# Removed automatically added add_*_asset methods we don't want.
-# TODO: Prevent these from being created.
-delattr(SqliteDatasource, "add_sqlite_table_asset")
```
billdirks, line 222: Oh nice, we can get rid of these now. I wonder if it was this that was causing the issue?

billdirks (approved):
LGTM, thanks!

PR #7809, opened 2023-05-03
[FEATURE] Plumbing of validation_result_url from cloud response

billdirks:
### Definition of Done

### DIFF
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -872,7 +872,28 @@ class StoreValidationResultAction(ValidationAction):
         checkpoint_identifier: Optional[GXCloudIdentifier] = None,
     ):
         logger.debug("StoreValidationResultAction.run")
+        run_return_value = self._basic_run(
+            validation_result_suite,
+            validation_result_suite_identifier,
+            expectation_suite_identifier,
+            checkpoint_identifier,
+        )
+        if self._using_cloud_context and isinstance(
+            run_return_value, GXCloudResourceRef
+        ):
+            return self._run_cloud_post_process_resource_ref(
+                run_return_value, validation_result_suite_identifier
+            )
 
+    def _basic_run(
+        self,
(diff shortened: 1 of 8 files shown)


great_expectations/validation_operators/validation_operators.py
```diff
@@ -235,6 +254,13 @@ def __init__(
                     class_name=config["class_name"],
                 )
             self.actions[action_config["name"]] = new_action
+        if notify_before_store and self._using_cloud_context:
+            logger.warning(
```
Kilo59: Should we raise a real `warning` here?
It will look more or less the same to a user but it would allow us to optional turn this into an exception when debugging or testing.
```python
import warnings

...

warnings.warn(f"The checkpoints action_list configuration has a notification, {notify_before_store}", category=UserWarning)
```
```

File: great_expectations/checkpoint/actions.py [hunk 2]
```diff
@@ -895,16 +897,19 @@ def _run(
...
             checkpoint_id=checkpoint_ge_cloud_id,
             expectation_suite_id=expectation_suite_ge_cloud_id,
         )
         if self._using_cloud_context:
-            return_val: GXCloudResourceRef
-            new_ge_cloud_id = return_val.id
+            store_set_return_value: GXCloudResourceRef
+            new_ge_cloud_id = store_set_return_value.id
             validation_result_suite_identifier.id = new_ge_cloud_id
+            return store_set_return_value
```
anthonyburdi: Would this _run() method be simpler to read if cloud vs non cloud is broken out into two separate private methods? Something like (params excluded for brevity):
```
def _run(...):
    if self._using_cloud_context:
        return self._cloud_run(...)
    else:
        return self._non_cloud_run(...)
```

anthonyburdi (approved):
Approving pending @Kilo59's comment is addressed (mine is a non blocking suggestion, feel free to implement or ignore).

PR #7812, opened 2023-05-04
[MAINTENANCE] Remove runme fixtures/stages and enable docs-integration to run automatically

kenwade4:
- Remove runme from fixtures
- Remove runme ci stage
- Enable docs-integration pipeline to run automatically


ci/azure-pipelines-docs-integration.yml
```diff
@@ -58,14 +56,15 @@ stages:
                 docs/**
                 tests/integration/docusaurus/**
                 tests/integration/fixtures/**
+                tests/integration/test_definitions/**
+                tests/integration/test_script_runner.py
```
anthonyburdi, line 60: Nice catch!

anthonyburdi (approved):
LGTM! In the future we might want to add the manual trigger as an `or` i.e. runs on all PRs and also when we want to manually run it. But not necessary for this PR. Thanks for turning these tests back on!

ci/azure-pipelines-docs-integration.yml
kenwade4: Yea for sure... it's nice to manually trigger when needed

kenwade4: But we are gonna talk about a bunch of potential pipeline changes at the arch meeting tomorrow. I think we should make all/most pipelines and stages manually runnable.

PR #7821, opened 2023-05-04
[BUGFIX] Upper bound `pyathena` due to breaking API in V3

cdkini:
- `from pyathena import sqlalchemy_athena` no longer works in V3
    - Let's set an upper bound so our current code work
    - Should circle back to see what imports we can use to support both

### Definition of Done


billdirks (approved):
Thanks!

Kilo59 (approved):

PR #7837, opened 2023-05-08
[FEATURE] Add scaffolding for gx-agent

joshua-stauffer:
- add cli entrypoint `gx-agent` and framework for agent

Previous Design Review notes:
-
-
-

### Definition of Done

Thank you for submitting!


requirements.txt
```diff
@@ -23,6 +23,7 @@ pandas>=1.1.0,<2.0.0; python_version <= "3.8"
 pandas>=1.1.3,<2.0.0; python_version == "3.9"
 pandas>=1.3.0,<2.0.0; python_version >= "3.10"
 # patch version updates `typing_extensions` to the needed version
+pika==1.3.1
```
Kilo59: We should make this an optional requirement

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,124 @@
...
+class GXAgentConfig:
+    """GXAgent configuration.
+
+    Attributes:
+        organization_id: GX Cloud organization identifier
+        broker_url: address of broker service
+    """
+
+    organization_id: str
+    broker_url: str
```
Kilo59: Can we make this a pydantic dataclass and use a special URL type for broker_url

https://docs.pydantic.dev/latest/usage/dataclasses/
https://docs.pydantic.dev/latest/usage/types/#urls
```python
from pydantic.dataclasses import dataclass
from pydantic import HttpUrl, AnyUrl

```


```suggestion
@dataclass(frozen=True)
class GXAgentConfig:
    """GXAgent configuration.

    Attributes:
        organization_id: GX Cloud organization identifier
        broker_url: address of broker service
    """

    organization_id: str
    broker_url: HttpUrl
```

joshua-stauffer: unfortunately not -- these will be `amqps` protocol, not `http/https`

requirements.txt
Kilo59: There are 3 steps here.
1. Remove this from requirements.txt
2. Create a new requirements file under `/reqs` called `requirements-dev-cloud.txt`
    * https://github.com/great-expectations/great_expectations/tree/develop/reqs
3. Add this `requirements-dev-cloud.txt` file to this [requirements-types.txt](https://github.com/great-expectations/great_expectations/blob/develop/requirements-types.txt). This ensure `pika` is installed for our type-checking step.
```suggestion
```

This will make it an optional dependency and `pika` and any other dependencies installable as "extra" dependencies.
```
pip install 'great_expectations[cloud]'
```

setup.py
```diff
@@ -92,7 +92,10 @@ def get_extras_require():
         exclude=["contrib*", "docs*", "tests*", "examples*", "scripts*"]
     ),
     "entry_points": {
-        "console_scripts": ["great_expectations=great_expectations.cli:main"]
+        "console_scripts": [
+            "great_expectations=great_expectations.cli:main",
+            "gx-agent=great_expectations.agent:run_agent"
+        ]
```
Kilo59: ❤️

great_expectations/agent/agent.py
joshua-stauffer: actually we can! just need to use the `AmqpDsn` type instead

great_expectations/agent/models.py
```diff
@@ -0,0 +1,21 @@
+from typing import Literal, Union
+from typing_extensions import Annotated
```
Kilo59: Until we drop `python 3.7` we are bound to an earlier version of `typing_extentions` due to an `airflow` compatibility problem.
So you might not be able to use `Annotated` at runtime.

But we are dropping `python 3.7` within the next few weeks so It's probably not worth making too much effort with a workaround if it leads to worse code.

great_expectations/agent/agent.py
Kilo59: You can use `AnyUrl`

Kilo59: Oh nice! Didn't notice the `AmqpDsn ` type, even better!

setup.py
joshua-stauffer: thanks for this suggestion!

tests/agent/message_service/test_subscriber.py
```diff
@@ -0,0 +1,84 @@
...
+def test_subscriber_has_client_attribute():
+    client = Mock()
+
+    subscriber = Subscriber(client=client)
+
+    assert subscriber.client is client
+
+
+def test_subscriber_consume_calls_basic_consume():
+    client = Mock()
```
Kilo59: We try to never use a mock without a `spec`.

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock

tests/agent/message_service/test_rabbit_mq_client.py
```diff
@@ -0,0 +1,45 @@
+from unittest.mock import patch
```
Kilo59: Would prefer if we used the `pytest-mock` version here.
https://pytest-mock.readthedocs.io/en/latest/index.html

And/or `pytest` `monkeypatch`.

In which case you don't actually need to import it. You can use use the `mocker` fixture.

And `from pytest_mock import MockerFixture` for the fixture annotation.

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Run GX in any environment from GX Cloud.
+
+    To start the agent, install Python and great_expectations and run `gx-agent`.
+    The agent loads a DataContext configuration from GX Cloud, and listens for
+    user events triggered from the UI.
+    """
+
+    def __init__(self):
+        print("Initializing GX-Agent")
```
Kilo59, line 57: `non-blocking`
As a followup, I'd love to see us use a logger instead of printing for all these messages and make the level configurable when running the agent from the command line.

Kilo59 (changes requested):
Code looks good.

Requesting changes to the requirement files and `Mock` usage.

requirements.txt
Kilo59: Updated the suggestion to call our extra dependencies `cloud` instead of `agent`, this is based on some discussion from Monday's standup.

cc @DrewHoo

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,122 @@
...
+from great_expectations.agent.message_service.subscriber import (
+    Subscriber,
+    OnMessageCallback,
+    SubscriberError,
+)
+from great_expectations.agent.models import Event
+from great_expectations.data_context import CloudDataContext
+
+
+HandlerMap = Dict[str, OnMessageCallback]
```
Kilo59, line 32:
```suggestion
HandlerMap: TypeAlias = Dict[str, OnMessageCallback]
```

tests/agent/message_service/test_subscriber.py
joshua-stauffer: updated in [866d2c0](https://github.com/great-expectations/great_expectations/pull/7837/commits/866d2c0ed473753180af6008a19680e4e48f5bf5)

tests/agent/message_service/test_rabbit_mq_client.py
joshua-stauffer: okay - i'm curious, what's the reasoning behind this?

great_expectations/agent/agent.py
joshua-stauffer: added in [87fa5a5](https://github.com/great-expectations/great_expectations/pull/7837/commits/87fa5a54b9427683e0fde954c3a5051c1516b511)

requirements.txt
joshua-stauffer: added in [6928465](https://github.com/great-expectations/great_expectations/pull/7837/commits/6928465da4e2e63309254d6b0b32598de9f4b411)

great_expectations/agent/agent.py
joshua-stauffer: agree, i had explored that as an initial implementation, but was much more straightforward to just go with print (also, is inline with the gx cli). let's consider it as followup work.

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Run GX in any environment from GX Cloud.
+
+    To start the agent, install Python and great_expectations and run `gx-agent`.
+    The agent loads a DataContext configuration from GX Cloud, and listens for
+    user events triggered from the UI.
+    """
+
+    def __init__(self):
+        print("Initializing GX-Agent")
```
cdkini, line 57: Should these be logger calls?

joshua-stauffer: see https://github.com/great-expectations/great_expectations/pull/7837#discussion_r1187569761

Kilo59: Just FYI we don't have the "resolve all conversations" requirement in OSS. So leaving conversations open can sometimes prevent multiple reviewers from asking the same question.

tests/agent/message_service/test_rabbit_mq_client.py
Kilo59: `pytest-mock` implements a superset of the `unittest.mock` API but has the added safety of always removing itself after a test finishes.

great_expectations/agent/message_service/subscriber.py
```diff
@@ -0,0 +1,104 @@
...
+        Args:
+            channel: the instance of pika.channel.Channel that delivered the message.
+            method_frame: pika object containing context on the delivered message like delivery_tag,
+                consumer_tag, redelivered, exchange, and routing_key.
+            header_frame: pika object containing any header fields, such as correlation_id
+            body: the message body in bytes
+            on_message: the caller-provided callback
+        """
+        try:
+            event = parse_raw_as(Event, body)
```
Kilo59: `non-blocking/nit`
I would prefer if we used the fully qualified `pydantic.parse_raw_as()` style here so that it's evident that `parse_raw_as` is a pydantic thing.
I don't believe we've used this anywhere else in the codebase, so it will be unfamiliar to most developers.

```suggestion
            event = pydantic.parse_raw_as(Event, body)
```

joshua-stauffer: updated in [46b4c22](https://github.com/great-expectations/great_expectations/pull/7837/commits/46b4c225dfb74874b90024564ecd0b8c2b8cf466)

tests/agent/message_service/test_rabbit_mq_client.py
joshua-stauffer: thanks for the context! updated in [d8f6b2b](https://github.com/great-expectations/great_expectations/pull/7837/commits/d8f6b2b42aeee27ed34ec0e2841b2b7043d4e416)

tests/agent/message_service/conftest.py
Kilo59, line 1: Why are we importing all the exceptions into `conftest.py`?

Kilo59:
```
great_expectations/agent/message_service/subscriber.py:70: error: Need type
annotation for "event"  [var-annotated]
                event = pydantic.parse_raw_as(Event, body)
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
great_expectations/agent/message_service/subscriber.py:70: error: Argument 1 to
"parse_raw_as" has incompatible type "<typing special form>"; expected
"Type[<nothing>]"  [arg-type]
                event = pydantic.parse_raw_as(Event, body)
                                              ^~~~~
great_expectations/agent/agent.py:153: error: Argument "broker_url" to
"GXAgentConfig" has incompatible type "str"; expected "AmqpDsn"  [arg-type]
            return GXAgentConfig(organization_id=org_id, broker_url=url)
                                                                    ^~~
Found 3 errors in 2 files (checked 475 source files)
```

great_expectations/agent/message_service/subscriber.py
```diff
@@ -0,0 +1,103 @@
...
+        Args:
+            channel: the instance of pika.channel.Channel that delivered the message.
+            method_frame: pika object containing context on the delivered message like delivery_tag,
+                consumer_tag, redelivered, exchange, and routing_key.
+            header_frame: pika object containing any header fields, such as correlation_id
+            body: the message body in bytes
+            on_message: the caller-provided callback
+        """
+        try:
+            event = pydantic.parse_raw_as(Event, body)
```
Kilo59:
```
great_expectations/agent/message_service/subscriber.py:70: error: Need type
annotation for "event"  [var-annotated]
                event = pydantic.parse_raw_as(Event, body)
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
great_expectations/agent/message_service/subscriber.py:70: error: Argument 1 to
"parse_raw_as" has incompatible type "<typing special form>"; expected
"Type[<nothing>]"  [arg-type]
                event = pydantic.parse_raw_as(Event, body)
```

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,157 @@
...
+        """Construct GXAgentConfig from available environment variables"""
+        url = os.environ.get("BROKER_URL", None)
+        if url is None:
+            raise GXAgentError("Missing required environment variable: BROKER_URL")
+        org_id = os.environ.get("GE_CLOUD_ORGANIZATION_ID", None)
+        if org_id is None:
+            raise GXAgentError(
+                "Missing required environment variable: GE_CLOUD_ORGANIZATION_ID"
+            )
+        return GXAgentConfig(organization_id=org_id, broker_url=url)
```
Kilo59:
```
great_expectations/agent/agent.py:153: error: Argument "broker_url" to
"GXAgentConfig" has incompatible type "str"; expected "AmqpDsn"  [arg-type]
            return GXAgentConfig(organization_id=org_id, broker_url=url)
                                                                    ^~~
```
`pydantic` will attempt to coerce the `str` to a `AmqpDsn` but `mypy` doesn't know that.
I would ignore the mypy warning with a comment explaining that `pydantic` will attempt to coerce the string and error raise an error if it cannot.

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,157 @@
...
+    def _get_config_from_env(cls) -> GXAgentConfig:
+        """Construct GXAgentConfig from available environment variables"""
+        url = os.environ.get("BROKER_URL", None)
+        if url is None:
+            raise GXAgentError("Missing required environment variable: BROKER_URL")
+        org_id = os.environ.get("GE_CLOUD_ORGANIZATION_ID", None)
+        if org_id is None:
+            raise GXAgentError(
+                "Missing required environment variable: GE_CLOUD_ORGANIZATION_ID"
+            )
```
Kilo59, line 152: You can get rid of both of these checks and instead, just catch and re-raise any pydantic errors.

great_expectations/agent/agent.py
Kilo59: You can get rid of both of these checks and instead, just catch and re-raise any pydantic errors. 

```suggestion
        url = os.environ.get("BROKER_URL", None)
        try:
            return GXAgentConfig(organization_id=org_id, broker_url=url)
        except pydantic.ValidationError as validation_err:
            raise GXAgentError(f"Missing or badly formed environment variable\n{validation_err.errors()}") from validation_err
```

great_expectations/agent/message_service/subscriber.py
Kilo59: The first error is just saying `event` needs an annotation.
```python
        try:
            event: Event = pydantic.parse_raw_as(Event, body)
```

great_expectations/agent/agent.py
Kilo59: Did you revert this?

You probably need to import this from `typing_extensions`.
I believe there is a `ruff` autofix rule that will correct this for you.

great_expectations/agent/message_service/subscriber.py
Kilo59: The second error is a bit surprising. Does it actually work at runtime?
Try ignoring the second error for now.

Kilo59:
Looks like the unit-tests aren't getting the needed dependencies installed but I'm not sure why.
https://github.com/great-expectations/great_expectations/blob/1e7a9acd382a346d444c2b327d11a98d2a7f61ae/ci/azure-pipelines-dev.yml#L129-L131

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,159 @@
...
+
+    @classmethod
+    def _get_config_from_env(cls) -> GXAgentConfig:
+        """Construct GXAgentConfig from available environment variables"""
+        url = os.environ.get("BROKER_URL", None)
+        org_id = os.environ.get("GE_CLOUD_ORGANIZATION_ID", None)
+        try:
+            return GXAgentConfig(
+                organization_id=org_id, broker_url=url
+            )  # noqa - mypy doesnt know pydantic handles url type
```
Kilo59: This is the wrong format for ignoring mypy warnings.
https://mypy.readthedocs.io/en/stable/error_codes.html#silencing-errors-based-on-error-codes

tests/agent/message_service/conftest.py
Kilo59, line 1: By doing all these imports inside `conftest.py` I think you are forcing `pika` to be installed for any gx tests, including our unit tests.

Can you refactor these imports to be done from inside a fixture, and use the fixture to expose the `AMQP_CHANNEL_AND_CONNECTION_ERRORS`?

joshua-stauffer: no unfortunately, since i need this list to parametrize tests. i can try to just rename this file, so pytest doesn't automatically pick it up

setup.py
```diff
@@ -92,7 +92,10 @@ def get_extras_require():
         exclude=["contrib*", "docs*", "tests*", "examples*", "scripts*"]
     ),
     "entry_points": {
-        "console_scripts": ["great_expectations=great_expectations.cli:main"]
+        "console_scripts": [
+            "great_expectations=great_expectations.cli:main",
+            "gx-agent=great_expectations.agent:run_agent",
```
Kilo59, line 98: @erik-hencier this is where we define the command for starting the agent.

great_expectations/agent/agent.py
joshua-stauffer: this was causing an error on python 3.8 so i reverted it

reqs/requirements-dev-cloud.txt
```diff
@@ -0,0 +1 @@
+pika==1.3.1
```
wookasz, line 1: Should this be part of `requirements.txt`? Otherwise we need to ensure that cloud users install from this file as well as `requirements.txt`.

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,201 @@
...
+            subscriber.consume(
+                queue=self._config.organization_id,
+                on_message=self._handle_event_as_thread_enter,
+                wait=1,
+            )
+        except KeyboardInterrupt:
+            print("Received request to shutdown.")
+        except (SubscriberError, ClientError) as e:
+            print("Connection to GX Cloud has encountered an error.")
+            print("Please restart the agent and try your action again.")
```
wookasz: If we're acking only after processing the request, all the user has to do is restart the agent and it should automatically process the action again

great_expectations/agent/message_service/rabbit_mq_client.py
```diff
@@ -0,0 +1,52 @@
...
+    channel: BlockingChannel
+
+    def __init__(self, url: str):
+        try:
+            # SSL Context for TLS configuration of Amazon MQ for RabbitMQ
+            ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
+            ssl_context.set_ciphers("ECDHE+AESGCM:!ECDSA")
+
+            parameters = pika.URLParameters(url)
+            parameters.ssl_options = pika.SSLOptions(context=ssl_context)
```
wookasz: I suggest checking the connection string for scheme and only setting SSLOptions only if `amqps`. This will allow running against local amqp instance for development

great_expectations/agent/models.py
```diff
@@ -0,0 +1,23 @@
...
+    type: str
+
+
+class RunDataAssistantEvent(EventBase):
+    type: Literal[
+        "onboarding_data_assistant_request.received"
+    ] = "onboarding_data_assistant_request.received"
+
+
+class ShutdownEvent(EventBase):
```
wookasz: Is this currently used? Do we plan on allowing the agent to be shut down via the UI?

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,201 @@
...
+    def _listen(self) -> None:
+        """Manage connection lifecycle."""
+        subscriber = None
+        try:
+            client = RabbitMQClient(url=self._config.broker_url)
+            subscriber = Subscriber(client=client)
+            print("GX-Agent is ready.")
+            # Open a blocking connection until encountering a shutdown event
+            subscriber.consume(
+                queue=self._config.organization_id,
```
wookasz, line 89:
```suggestion
                # TODO: replace with queue from agent-sessions endpoint
                queue=self._config.organization_id,
```

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,201 @@
...
+        try:
+            subscriber.close()
+        except SubscriberError as e:
+            print("Subscriber encountered an error while closing:")
+            print(e)
+
+    @classmethod
+    def _get_config_from_env(cls) -> GXAgentConfig:
+        """Construct GXAgentConfig from available environment variables"""
+        url = os.environ.get("BROKER_URL", None)
```
wookasz, line 200:
```suggestion
        # TODO: replace with connection string from agent-sessions endpoint
        url = os.environ.get("BROKER_URL", None)
```

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,201 @@
...
+        org_id = os.environ.get("GE_CLOUD_ORGANIZATION_ID", None)
```
wookasz: `GE_` has been deprecated, use `GX_`

```suggestion
        org_id = os.environ.get("GX_CLOUD_ORGANIZATION_ID", None)
```

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,201 @@
...
+            result = future.result()
+            print(result)
+
+        # TODO lakitu-139: record job as complete and send results
+
+        # ack message and cleanup resources
+        if isinstance(error, UnknownEventError):
+            # We might not have the latest Event definitions, so we requeue
+            # this event in the hope that another agent will understand it.
+            retry = True
```
wookasz: Seems like in this case we should re-queue but also exit the agent and tell the user to update to the latest version. Otherwise the agent will indefinitely loop trying to process with no result

great_expectations/agent/agent.py
```diff
@@ -0,0 +1,201 @@
...
+
+        Callback passed to Subscriber.consume which forwards events to
+        the EventHandler for processing.
+
+        Args:
+            event_context: An Event with related properties and actions.
+        """
+        if self._can_accept_new_task() is not True:
+            # signal to Subscriber that we can't process this message right now
+            return event_context.event_processed(retry=True)
```
wookasz: Without any delay, this will cause the agent to spin, trying to execute the event and sending nacks to the broker. Is there a way we could delay attempting to process the next message somehow?

Maybe this is not something to be concerned with for now.

wookasz (commented):
Some initial comments

great_expectations/agent/models.py
joshua-stauffer: looks like 3.7 will get dropped just in time for this to merge 👌

reqs/requirements-dev-cloud.txt
joshua-stauffer: yeah, per review we're asking users to install cloud deps separately with `pip install 'great_expectations[cloud]'` https://github.com/great-expectations/great_expectations/pull/7837#discussion_r1187512851

great_expectations/agent/models.py
joshua-stauffer: no, i left it in because the pydantic union broke if it was a single object

joshua-stauffer: and if we do plan on exposing a way to shutdown an agent, we should do so via consumer tags, so we know which agent we're shutting down

great_expectations/agent/agent.py
joshua-stauffer: thanks 👍

tests/agent/message_service/test_subscriber.py
```diff
@@ -0,0 +1,82 @@
...
+)
+from tests.agent.message_service.amqp_errors import get_amqp_errors
+
+
+def test_subscriber_has_client_attribute():
+    client = Mock(autospec=RabbitMQClient)
+
+    subscriber = Subscriber(client=client)
+
+    assert subscriber.client is client
```
Kilo59: This seems like an unnecessary test.
The type-checker verifies the attribute is present.

great_expectations/agent/agent.py
joshua-stauffer: thanks - in conjunction with the logic to check correlation IDs, we should only see the same message max 10 times before nacking it with requeue=False. This logic has also been updated to delay a couple seconds before sending the requeue request

great_expectations/agent/agent.py
joshua-stauffer: my concern with that is if somehow unexpected messages end up passed to the agent it would cause all agents to disconnect, which seems undesirable. my current solution is to allow the same message to be requeued x amount of times (currently set at 10) before sending it to DLX

great_expectations/agent/agent.py
joshua-stauffer: updated in [3e1c119](https://github.com/great-expectations/great_expectations/pull/7837/commits/3e1c119d4f38e51e18de72a6baf8e930a0aa7b0b)

great_expectations/agent/agent.py
joshua-stauffer: diff was outdated so added in [90c1ee2](https://github.com/great-expectations/great_expectations/pull/7837/commits/90c1ee2c5c9f43c4fe98f958d8814bf16b8e8d30)

great_expectations/agent/agent.py
joshua-stauffer: diff was outdated so added in [90c1ee2](https://github.com/great-expectations/great_expectations/pull/7837/commits/90c1ee2c5c9f43c4fe98f958d8814bf16b8e8d30)

great_expectations/agent/models.py
joshua-stauffer: updated this to `RunCheckpointEvent`, since that's likely to be useful in the near term: [21c804e](https://github.com/great-expectations/great_expectations/pull/7837/commits/21c804ecddac11ff83efc0079ec3fe42f3768de6)

great_expectations/agent/message_service/rabbit_mq_client.py
joshua-stauffer: thanks - updated in [0a69b2d](https://github.com/great-expectations/great_expectations/pull/7837/commits/0a69b2d624c65f6dbb3c33ca86cc2b8a1daf8aae)

great_expectations/agent/agent.py
wookasz: Our current assumption is that the user has precisely 1 agent running. In that scenario, the retries will happen so quickly with the same agent, that the requeueing won't make a difference.

joshua-stauffer: sorry wasn't clear in my last message - these retries are set on an async delay, so there's a chance for another agent to pick it up without blocking the thread

joshua-stauffer: happy to update this behavior going forward, if we decide this isn't best

wookasz (approved):
Looks great. We can iterate on any edge cases or improvements

Kilo59 (approved):
Looks good to me.

I think the test(s) that simply check for things like the presence of an attribute should be removed.

PR #7841, opened 2023-05-08
[MAINTENANCE] Adding docs link checker to invoke

donaldheppner:
- [x] Code is linted


Shinnnyshinshin (changes requested):
Hi @donaldheppner thank you for this. One very small change request :)

tasks.py
```diff
@@ -651,3 +651,25 @@ def _exit_with_error_if_not_run_from_correct_dir(
             exit_message,
             code=1,
         )
+
+
+@invoke.task(
+    name="link_checker",
+    aliases=("links",),
```
Shinnnyshinshin: I think this will have to be `link-checker` to make the invoke task work (`_` is not supported)
```
(supercon_ge) ➜  great_expectations git:(m/_/inv_links) invoke link_checker
No idea what 'link_checker' is!
(supercon_ge) ➜  great_expectations git:(m/_/inv_links) invoke link-checker
No broken links found
```

Shinnnyshinshin: The second command was done after renaming the task name to `link-checker`

tasks.py
```diff
@@ -651,3 +651,25 @@ def _exit_with_error_if_not_run_from_correct_dir(
             exit_message,
             code=1,
         )
+
+
+@invoke.task(
+    name="link_checker",
```
```suggestion
```
This should address @Shinnnyshinshin 's comment

tasks.py
Kilo59: If line `657` is removed it should work. Invoke will automatically rename snake-case method names to use hyphens.

Shinnnyshinshin: oo that's even better. Let's go that direction

donaldheppner: I've applied the suggestion -- however, please note that `invoke --list` (which lists commands) never said that `link_checker` was a valid task name (it listed `link-checker` instead). Other tasks also have underscore in their name

Shinnnyshinshin (approved):
🚀

PR #7844, opened 2023-05-08
[FEATURE] Spark file reader support for fluent datasources

anthonyburdi:
- Support for pyspark file reader methods for the following file types: csv, parquet, orc, json, text


great_expectations/datasource/fluent/spark_file_path_datasource.py
```diff
@@ -1,87 +1,480 @@
...
-    infer_schema: bool = Field(False, alias="InferSchema")
+
+    # vvv spark parameters for pyspark.sql.DataFrameReader.csv() (ordered as in pyspark v3.4.0) appear in comment above
+    # parameter for reference (from https://github.com/apache/spark/blob/v3.4.0/python/pyspark/sql/readwriter.py#L604)
+    # See https://spark.apache.org/docs/latest/sql-data-sources-csv.html for more info.
+    # path: PathOrPaths,
+    # NA - path determined by asset
+    # schema: Optional[Union[StructType, str]] = None,
+    # schema shadows pydantic BaseModel attribute
+    spark_schema: Optional[Union[pyspark_types.StructType, str]] = Field(
```
anthonyburdi: I believe this line is leading to a `E   pydantic.errors.ConfigError: field "spark_schema" not yet prepared so type is still a ForwardRef, you might need to call JSONAsset.update_forward_refs().` error. Trying to determine how best to handle / call update_foward_refs().

anthonyburdi: Currently setting `arbitrary_types_allowed = True` but long term solution is to add a custom validator for this type.

anthonyburdi: Custom validator attempted in https://github.com/great-expectations/great_expectations/pull/7844/commits/db36ec26eaaf29d24be97aa26dcad2df59cb82b7, but it is a bit bigger lift so that won't be included in this PR and instead we will accept arbitrary types.

great_expectations/datasource/fluent/serializable_types/pyspark.py
```diff
@@ -0,0 +1,45 @@
+from __future__ import annotations
+
+from great_expectations.compatibility.pyspark import pyspark
+
+
+class SerializableStructType(dict):
+    """Custom type implementing pydantic validation."""
+
+    struct_type: pyspark.sql.types.StructType
```
alexsherstinsky, line 9: @anthonyburdi This may not work without first some kind of a boolean check on the `pyspark.*` type (but if it works, great).  Thanks.

great_expectations/datasource/fluent/serializable_types/pyspark.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def __init__(
+        self,
+        fields_or_struct_type: pyspark.sql.types.StructType
+        | list[pyspark.sql.types.StructField]
+        | None,
+    ):
+        if isinstance(fields_or_struct_type, pyspark.sql.types.StructType):
```
alexsherstinsky, line 18: @anthonyburdi e.g., here, we might need to say "if pyspark.sql....` (or something like that -- we have the examples elsewhere).  Thanks.

great_expectations/datasource/fluent/serializable_types/pyspark.py
```diff
@@ -0,0 +1,45 @@
...
+            self.struct_type = fields_or_struct_type
+        else:
+            self.struct_type = pyspark.sql.types.StructType(
+                fields=fields_or_struct_type
+            )
+
+        json_value = self.struct_type.jsonValue()
```
alexsherstinsky, line 26: @anthonyburdi Just curious, is `jsonValue()` a PySpark method?  Thanks.

great_expectations/datasource/fluent/spark_filesystem_datasource.pyi
```diff
@@ -41,20 +47,206 @@ class SparkFilesystemDatasource(_SparkFilePathDatasource):
...
+        sep: Optional[str] = None,
+        # encoding: Optional[str] = None,
+        encoding: Optional[str] = None,
+        # quote: Optional[str] = None,
+        quote: Optional[str] = None,
+        # escape: Optional[str] = None,
+        escape: Optional[str] = None,
+        # comment: Optional[str] = None,
+        comment: Optional[str] = None,
+        # header: Optional[Union[bool, str]] = None,
```
alexsherstinsky, line 66: @anthonyburdi Curious -- what is the reason for adding the same line with a comment (and it is not even the same in a number of places)?  Is this something our documentation system looks for?  Thanks.

tests/datasource/fluent/test_spark_filesystem_datasource.py
```diff
@@ -85,23 +86,329 @@ def test_add_csv_asset_to_datasource(
...
-        int_96_rebase_mode="CORRECTED",
-        merge_schema=False,
-    )
-    assert asset.name == "parquet_asset"
-    m1 = asset.batching_regex.match("this_can_be_named_anything.parquet")
-    assert m1 is not None
-    assert asset.datetime_rebase_mode == "EXCEPTION"
-    assert asset.int_96_rebase_mode == "CORRECTED"
-    assert asset.merge_schema is False
+    asset = getattr(spark_filesystem_datasource, add_method_name)(
```
alexsherstinsky, line 331: ❤️

tests/datasource/fluent/test_spark_filesystem_datasource.py
```diff
@@ -85,23 +86,329 @@ def test_add_csv_asset_to_datasource(
...
+@pytest.mark.parametrize(
+    "add_method_name,add_method_params",
+    add_directory_asset_test_params,
+)
+def test_add_directory_asset_with_asset_specific_params(
+    spark_filesystem_datasource: SparkFilesystemDatasource,
+    add_method_name: str,
+    add_method_params: dict[str, str | bool | int | float],
+):
+    asset = getattr(spark_filesystem_datasource, add_method_name)(
```
alexsherstinsky, line 464: ❤️

great_expectations/datasource/fluent/spark_datasource.py
```diff
@@ -57,6 +58,12 @@ class _SparkDatasource(Datasource):
     spark_config: Union[SparkConfig, None] = None
     force_reuse_spark_context: bool = True
 
+    @staticmethod
+    def _update_asset_forward_refs(asset_type: Type[_DataAssetT]) -> None:
+        # Only update forward refs if pyspark types are available.
+        if pyspark:
```
alexsherstinsky, line 64: @anthonyburdi I am not sure whether or not this will work -- it might need to be `pyspark.<something>`.  Thanks.

great_expectations/datasource/fluent/spark_file_path_datasource.py
```diff
@@ -1,87 +1,485 @@
...
 
-    def _get_reader_method(self) -> str:
-        return self.type
+    @classmethod
+    def _get_reader_method(cls) -> str:
+        return "csv"
 
     def _get_reader_options_include(self) -> set[str] | None:
-        return {"header", "infer_schema"}
+        """These options are available as of spark v3.4.0
```
alexsherstinsky, line 182: ❤️

great_expectations/datasource/fluent/directory_data_asset.py
```diff
@@ -62,7 +62,6 @@ def _get_reader_method(self) -> str:
         )
 
     def _get_reader_options_include(self) -> set[str] | None:
-        raise NotImplementedError(
-            """One needs to explicitly provide set(str)-valued reader options for "pydantic.BaseModel.dict()" method \
-to use as its "include" directive for File-Path style DataAsset processing."""
-        )
+        return {
+            "data_directory",
```
alexsherstinsky, line 66: @anthonyburdi Should we not leave this as before (raising an exception), such that the subclasses will return the value they need?  Or is `"data_directory"` used in all subclasses, which will call `super()._get_reader_options_include()`?  Thanks.

great_expectations/datasource/fluent/serializable_types/pyspark.py
```diff
@@ -0,0 +1,45 @@
+from __future__ import annotations
+
+from great_expectations.compatibility.pyspark import pyspark
+
+
+class SerializableStructType(dict):
```
alexsherstinsky, line 6: @anthonyburdi Unless this has already been published (sorry if I missed it), it would be really nice to see a writeup explaining the serializable types, why they are needed, and how they are utilized (and whether or not there will be other ones, in addition to this one in the `pyspark.py` module).  Thank you!

alexsherstinsky (changes requested):
@anthonyburdi This has been very educational.  I asked for some clarification and a few questions.  Happy to re-review later tonight and/or tomorrow after approximately noon Pacific.  Thank you!

great_expectations/datasource/fluent/serializable_types/pyspark.py
anthonyburdi: It is!

great_expectations/datasource/fluent/spark_filesystem_datasource.pyi
anthonyburdi: This is a good question. I copied over the signature from the pyspark source code to check whether what I was adding was consistent. Now that they are exactly the same I think I can remove these and instead keep the comments that this set of params is for spark 3.4.0.

great_expectations/datasource/fluent/directory_data_asset.py
anthonyburdi: Good question - all subclasses that use a directory instead of individual file paths will need `"data_directory"` so they will inherit from `_DirectoryDataAssetMixin` in addition to the file asset type to get this additional field. I don't generally like to use multiple inheritance, but I think this is a good use case to reduce code duplication whereby we are adding standard functionality (data_directory) via the mixin. Take a look at `DirectoryCSVAsset` for an example. In a followup PR I'll be adding `DirectoryParquetAsset`, `DirectoryORCAsset`, etc.

great_expectations/datasource/fluent/serializable_types/pyspark.py
anthonyburdi: Good call, this is a pattern that I worked out with huge help from @Kilo59 and @NathanFarmer just yesterday to serialize non GX types. I'll add more in the README.md (added in this PR) to explain in more detail.

great_expectations/datasource/fluent/serializable_types/pyspark.py
anthonyburdi: Another good question! This does work. My understanding is that the type is not evaluated unless a spark asset model is created, and if spark is not installed we raise the not imported error prompting for spark to be installed.

great_expectations/datasource/fluent/serializable_types/pyspark.py
anthonyburdi: This works for the same reason as the previous comment. I think maybe the main difference between other uses (where we do have to guard with an if) is that those code paths are handling multiple potentially installed backends e.g. sql, spark where the other paths still must work even if some optional installs are not installed. Here we are only handling spark so if spark isn't installed we are OK to fail and ask for spark to be installed. Does that make sense?

great_expectations/datasource/fluent/spark_datasource.py
anthonyburdi: Good catch! I missed the second pyspark. Fixing.

great_expectations/datasource/fluent/spark_filesystem_datasource.pyi
anthonyburdi: Actually I think for now I would like to leave them in since they are not exactly the same in many cases (e.g. spark_schema and the spark params are in camel case). It shows the correspondence between spark and our parameter names. This is also in the asset model as aliases, so I may be convinced at some point to remove these but I think they will be helpful when we add support for other spark versions.

great_expectations/datasource/fluent/spark_datasource.py
anthonyburdi: Fixed!

great_expectations/datasource/fluent/spark_file_path_datasource.py
```diff
@@ -1,87 +1,481 @@
...
+    # enable_date_time_parsing_fallback: bool = Field(
+    #     alias="enableDateTimeParsingFallback"
+    # )
+    # ^^^ Docs <> Source Code mismatch
 
     class Config:
         extra = pydantic.Extra.forbid
         allow_population_by_field_name = True
+        # pyspark types (from pyspark_types) are not currently validated
+        arbitrary_types_allowed = True
```
anthonyburdi: Can this be removed?

great_expectations/datasource/fluent/spark_file_path_datasource.py
```diff
@@ -1,87 +1,481 @@
...
+    # enable_date_time_parsing_fallback: bool = Field(
+    #     alias="enableDateTimeParsingFallback"
+    # )
+    # ^^^ Docs <> Source Code mismatch
+
+    class Config:
+        extra = pydantic.Extra.forbid
+        allow_population_by_field_name = True
+        # pyspark types (from pyspark_types) are not currently validated
+        arbitrary_types_allowed = True
```
anthonyburdi: Can I remove this?

NathanFarmer (approved):
LGTM pending the minor change we discussed in the synchronous review.

great_expectations/datasource/fluent/spark_file_path_datasource.py
alexsherstinsky: @anthonyburdi If the tests pass, then why not?!

great_expectations/datasource/fluent/spark_file_path_datasource.py
alexsherstinsky: @anthonyburdi If the tests pass, then I would say, "Go for it!"

alexsherstinsky (approved):
LGTM (once all tests pass).

PR #7849, opened 2023-05-09
[MAINTENANCE] Pin altair

anthonyburdi:
Altair recently released 5.0.0, we will temporarily pin until we can become compatible. https://pypi.org/project/altair/#history

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-altair>=4.2.1<5.0.0
+altair>=4.2.1,<5.0.0
 Click>=7.1.2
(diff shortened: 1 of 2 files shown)


donaldheppner (approved):

kg005:
Hi @anthonyburdi , just a question, is there a plan to support altair >= 5.0.0 in the near future? Thanks!

Kilo59:
@kg005 as a first step I would suggest opening a PR that removes or loosens that pin and seeing what breaks.

Issue #6393, opened 2022-11-18
SparkDF can't compute mean for decimal column


rogalski:
**Describe the bug**
https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/dataset/sparkdf_dataset.py#L646
Decimal type fails explicitly, yet Spark is fully capable of computing mean of decimal typed column.

**To Reproduce**
1. Create parquet with decimal type
2. Run `expect_column_mean_to_be_between` expectations using Spark
3. Encounter exception

**Expected behavior**
- Exception should not happen, as Spark is capable of computing mean of such column.
- If exception is raised, it should explicitly mean the offending column type so that feedback to user is self-explainatory

**Environment (please complete the following information):**
- N/A, affects code in develop

**Additional context**
N/A

austiezr:
Hey @rogalski ! Thanks for raising this. I've been able to reproduce this behavior, and will discuss with the team in the coming week. If you're interested in making a contribution to resolve this bug, we would be happy to provide review and guidance.

PR #7867, opened 2023-05-11
[BUGFIX] DX-506: Fix sparkDF cannot compute mean for DecimalType

TrangPham:
Addresses: https://github.com/great-expectations/great_expectations/issues/6393

I reproduced the issue within a unit test:

Screenshot 2023-05-12 at 2 09 43 PM

And updated the code such that the unit test passes. 

I also discovered that the pandas std metric does not work with Decimal either. It errs with `MetricResolutionError: unsupported operand type(s) for -: 'float' and 'decimal.Decimal'` I have added a unit test for that but commented it out for now.

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```


TrangPham:
@cla-bot check

great_expectations/expectations/metrics/column_aggregate_metrics/column_mean.py
```diff
@@ -32,6 +32,11 @@ def _sqlalchemy(cls, column, **kwargs):
     def _spark(cls, column, _table, _column_name, **kwargs):
         """Spark Mean Implementation"""
         types = dict(_table.dtypes)
-        if types[_column_name] not in ("int", "float", "double", "bigint"):
-            raise TypeError("Expected numeric column type for function mean()")
+        if (
+            types[_column_name] not in ("int", "float", "double", "bigint")
+            and "decimal" not in types[_column_name]
```
alexsherstinsky: @TrangPham Would it work if we added `"decimal"` to the `tuple` above?  Thank you.

tests/expectations/metrics/test_core.py
```diff
@@ -76,8 +77,18 @@ def test_basic_metric_pd():
...
+@pytest.mark.parametrize(
+    "dataframe,result",
+    [
+        [pd.DataFrame({"a": [1, 2, 3, None]}), 2],
+        [
+            pd.DataFrame({"a": [Decimal(2.0), Decimal(0.18781)]}),
+            1.093905,
+        ],
+    ],
+)
```
alexsherstinsky, line 89: @TrangPham This is a nice parametrization.  How would you feel if we also tried to include a case where a column contains both, integers and decimals in different rows (right now, all types are the same) -- and see if it passes?  Thanks.

tests/expectations/metrics/test_core.py
```diff
@@ -99,11 +110,71 @@ def test_mean_metric_pd():
...
 
 
-def test_stdev_metric_pd():
-    engine = build_pandas_engine(pd.DataFrame({"a": [1, 2, 3, None]}))
+@pytest.mark.parametrize(
+    "dataframe,result",
+    [
+        [pd.DataFrame({"a": [1, 2, 3, None]}), 2],
+        [
+            pd.DataFrame({"a": [Decimal(2.0), Decimal(0.18781)]}),
```
alexsherstinsky, line 121: @TrangPham Same comment -- about trying a heterogenous column in addition to the above homogenous ones.  Thanks.

tests/expectations/metrics/test_core.py
```diff
@@ -99,11 +110,71 @@ def test_mean_metric_pd():
...
+            1.093905,
+        ],
+    ],
+)
+def test_mean_metric_spark(spark_session, dataframe, result):
```
alexsherstinsky: 🚀

tests/expectations/metrics/test_core.py
```diff
@@ -76,8 +77,18 @@ def test_basic_metric_pd():
     assert results == {desired_metric.id: 3}
 
 
-def test_mean_metric_pd():
-    engine = build_pandas_engine(pd.DataFrame({"a": [1, 2, 3, None]}))
+@pytest.mark.parametrize(
+    "dataframe,result",
```
alexsherstinsky: 🚀

tests/expectations/metrics/test_core.py
```diff
@@ -99,11 +110,71 @@ def test_mean_metric_pd():
...
+        [build_pandas_engine, pd.DataFrame({"a": [1, 2, 3, None]}), 1],
+        # Does not work with Decimals now. MetricResolutionError: unsupported operand type(s) for -: 'float' and 'decimal.Decimal'
+        # [
+        #     build_pandas_engine,
+        #     pd.DataFrame({"a": [Decimal(2.0), Decimal(0.18781)]}),
+        #     1.093905,
+        # ],
+    ],
+)
+def test_stdev_metric_pd(build_engine, dataframe, result):
```
alexsherstinsky: @TrangPham If your finding is that in Pandas decimals work with the `mean` but not with the `stddev`, we need to understand this better -- it seems surprising.  We should also remove the comment from the code, but definitely discuss what is going on.  How about next week pairing to troubleshoot and understand why (or perhaps find documentation that states this explicitly with an explanation).  Thank you.

tests/expectations/metrics/test_core.py
```diff
@@ -125,7 +196,10 @@ def test_stdev_metric_pd():
         metrics_to_resolve=(desired_metric,), metrics=metrics
     )
     metrics.update(results)
-    assert results == {desired_metric.id: 1}
+    assert results == {desired_metric.id: result}
+
+
+# Add tests SQLAlchemy Spark for stdev and mean
```
alexsherstinsky: @TrangPham We should remove comments like this (or add them with an annotation with a date).  Thanks.

tests/expectations/metrics/test_core.py
```diff
@@ -125,7 +196,10 @@ def test_stdev_metric_pd():
         metrics_to_resolve=(desired_metric,), metrics=metrics
     )
     metrics.update(results)
-    assert results == {desired_metric.id: 1}
+    assert results == {desired_metric.id: result}
```
alexsherstinsky: @TrangPham The previous assertion was better (in my opinion), because it states explicitly what we expect; if we say `result`, the `result` may be wrong, but the assertion will give `True`.  Thanks.

tests/test_definitions/column_aggregate_expectations/expect_column_mean_to_be_between.json
```diff
@@ -53,6 +58,20 @@
             "observed_value": 3.5
           }
         },
+        {
```
alexsherstinsky: @TrangPham Side note: If we indeed find that decimals are supported by some backends, while not by others, we will have to use `only_for` / `suppress_for` in these `JSON` test cases and/or potentially add `JSON` files for specific backends (we can figure this out next week after we understand why decimals are not supported for certain backends).  Thank you.

alexsherstinsky (changes requested):
@TrangPham Good stuff -- I added some thoughts.  Thank you!

tests/test_definitions/column_aggregate_expectations/expect_column_mean_to_be_between.json
Shinnnyshinshin: Great point, would be happy to pair on this if you need.

great_expectations/expectations/metrics/column_aggregate_metrics/column_mean.py
TrangPham: No as it actually returns “decimal(x, y)” where x and y are ints. I believe they are for scale and precision.

tests/expectations/metrics/test_core.py
TrangPham: There seems to be float specific logic within std dev metric which does not work with the Decimals. Yes let’s discuss next week.

great_expectations/expectations/metrics/column_aggregate_metrics/column_mean.py
alexsherstinsky: but then `and "decimal" not in types[_column_name]` -- seems to do the same?..

alexsherstinsky: I think I see what is going on; seems like we should handle this more robustly than just string inclusion (if possible).

tests/expectations/metrics/test_core.py
alexsherstinsky: @TrangPham On a second thought, columns with values that have different types (heterogeneous) are rare/unlikely.  So if this does not look meaningful after an initial quick experiment, let us abandon this and stick to homogeneous (same-type valued) columns (like you have it already).  Thank you.

tests/expectations/metrics/test_core.py
alexsherstinsky: @TrangPham On a second thought, columns with values that have different types (heterogeneous) are rare/unlikely.  So if this does not look meaningful after an initial quick experiment, let us abandon this and stick to homogeneous (same-type valued) columns (like you have it already).  Thank you.

tests/expectations/metrics/test_core.py
alexsherstinsky: @TrangPham This is very odd indeed; looking forward to learning what is going on in there.  Thanks!

great_expectations/expectations/metrics/column_aggregate_metrics/column_mean.py
alexsherstinsky: @TrangPham I am thinking that if we use `_table.schema[_column_name].dataType` (instead of `_table.dtypes`), then we will get formal `PySpark` type information (`StringType`, `IntegerType`, `DecimalType`, etc.), and then we can do the checks for which of them satisfies the "numeric type" requirements.  Please note that if you take this approach, you would need to modify the `Spark` compatibility import to also include `types` and be able to reference `types.StringType`, `types.IntegerType`, `types.DecimalType`, etc. as needed.  Thank you!

tests/expectations/metrics/test_core.py
anthonyburdi: I think we are good here - the `result` comes from the parametrization so this assertion is checking against the value provided for this test case (`1` in this case).

alexsherstinsky: @anthonyburdi How are you seeing the `result` coming from parametrization?  I am seeing it come from resolving a metric (in line 195); or is it guaranteed thanks to line 167?  Not sure without deeper thinking that it is "necessarily" guaranteed (things can go wrong while computing the metric).  In our style of tests in this module, I believe that we compare explicitly -- "just in case" -- happy to discuss what is best.  Thank you.

tests/expectations/metrics/test_core.py
```diff
@@ -99,11 +110,71 @@ def test_mean_metric_pd():
...
+    )
+
+    assert results == {desired_metric.id: result}
+
+
+@pytest.mark.parametrize(
+    "build_engine,dataframe,result",
+    [
+        [build_pandas_engine, pd.DataFrame({"a": [1, 2, 3, None]}), 1],
+        # Does not work with Decimals now. MetricResolutionError: unsupported operand type(s) for -: 'float' and 'decimal.Decimal'
```
alexsherstinsky: @TrangPham For tidiness purposes, we should clean up before merging (i.e., if we do not need the code currently appearing as commented out, then we should consider deleting it).  Thanks.

great_expectations/expectations/metrics/column_aggregate_metrics/column_mean.py
TrangPham: Will update! Thanks for finding a more robust solution!

tests/expectations/metrics/test_core.py
TrangPham: Should I open a new bug for this case?

tests/test_definitions/column_aggregate_expectations/expect_column_mean_to_be_between.json
TrangPham: For some reason, including these test cases here pass when I don't think they should have. The decimal test case for spark was passing before I added my new code.

tests/expectations/metrics/test_core.py
alexsherstinsky: @TrangPham I am not sure -- happy to discuss to better understand how things run/fail.  Thanks.

tests/test_definitions/column_aggregate_expectations/expect_column_mean_to_be_between.json
alexsherstinsky: @TrangPham Are we sure that they even run?

tests/expectations/metrics/test_core.py
TrangPham: I have updated the variable name to be less confusing and more readable.

tests/expectations/metrics/test_core.py
TrangPham: I've opened https://github.com/great-expectations/great_expectations/issues/7897

TrangPham:
@cla-bot check

tests/test_definitions/column_aggregate_expectations/expect_column_mean_to_be_between.json
TrangPham: Ah I think you are right. I was not passing --spark before and that skipped the new test I was interested in.

great_expectations/expectations/metrics/column_aggregate_metrics/column_mean.py
```diff
@@ -31,7 +32,15 @@ def _sqlalchemy(cls, column, **kwargs):
     @column_aggregate_partial(engine=SparkDFExecutionEngine)
     def _spark(cls, column, _table, _column_name, **kwargs):
         """Spark Mean Implementation"""
-        types = dict(_table.dtypes)
-        if types[_column_name] not in ("int", "float", "double", "bigint"):
-            raise TypeError("Expected numeric column type for function mean()")
+        column_data_type = _table.schema[_column_name].dataType
+        if type(column_data_type) not in (
```
alexsherstinsky, line 36: @TrangPham I believe that we still need to import `types` from pyspark compatibility.  Thanks.

alexsherstinsky: @TrangPham Also, FYI -- I am not sure whether or not `type()` in line 36 is how we would check (could be not needed).

TrangPham: It seems to be required:

Screenshot 2023-05-15 at 4 51 56 PM

alexsherstinsky: @TrangPham I could be wrong -- I am just thinking that there might be something else going on -- it might be easier to troubleshoot via Zoom -- I would be curious to see the types involved in order to figure out if `type()` -- or something else -- is the best fit here.  Thanks!

great_expectations/self_check/util.py
```diff
@@ -805,6 +806,7 @@ def _get_test_validator_with_data_spark(  # noqa: C901 - 19
         "BooleanType": pyspark.types.BooleanType,
         "DataType": pyspark.types.DataType,
         "NullType": pyspark.types.NullType,
+        "DecimalType": partial(pyspark.types.DecimalType, 38, 18),
```
alexsherstinsky, line 811: @TrangPham The use of `partial` is new for us -- so a comment or a docstring explaining what is being done here would be very helpful.  Also, are we sure about the precision having to be `(38, 18)`?  A comment or a docstring would be nice.  Thank you.

great_expectations/self_check/util.py
```diff
@@ -840,25 +842,30 @@ def _get_test_validator_with_data_spark(  # noqa: C901 - 19
                 print(schema)
             for col in schema:
                 type_ = schema[col]
+                # Ints cannot be None...but None can be valid in Spark (as Null)
+                vals: List[Union[str, int, float, None, Decimal]] = []
```
alexsherstinsky, line 848: @TrangPham If we are going to have a type hint for `vals` here, should we also have it for all cases of `type_` below?  Thanks.

tests/expectations/metrics/test_core.py
```diff
@@ -99,11 +110,65 @@ def test_mean_metric_pd():
         metrics_to_resolve=(desired_metric,), metrics=metrics
     )
     metrics.update(results)
-    assert results == {desired_metric.id: 2}
+    assert results == {desired_metric.id: expected_result}
```
alexsherstinsky, line 113: 👍

tests/test_definitions/test_expectations_v3_api.py
```diff
@@ -59,408 +59,414 @@ def pytest_generate_tests(metafunc):  # noqa C901 - 35
         )
         for backend in backends:
             for filename in test_configuration_files:
-                pk_column: bool = False
```
alexsherstinsky: @TrangPham If these improvements (the use of `with`) are not in support of solving the decimals issues, then it would probably be easier to handle in a separate pull request -- or does it belong as part of this one?  Thank you.

TrangPham:
@alexsherstinsky can you take another look at this PR?

great_expectations/expectations/metrics/column_aggregate_metrics/column_mean.py
TrangPham: Without type() here, it would return "Decimal(38,18)". The type of which is "Decimal". What are your suggestions here?

great_expectations/self_check/util.py
TrangPham: This would apply to all of them. I cannot redefine the variable with its type again, doing so causes an error with the static type checking in CI pipeline.

tests/test_definitions/test_expectations_v3_api.py
TrangPham: I can move it to another pull request. This was just something small I found an improvement for.

alexsherstinsky (approved):
LGTM!

PR #7882, opened 2023-05-15
[FEATURE] Add Onboarding Data Assistant agent action

joshua-stauffer:
- refactor the agent's EventHandler to use AgentActions
- implement a RunOnboardingDataAssistantAction

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```


great_expectations/agent/actions/action.py
```diff
@@ -0,0 +1,27 @@
...
+
+class CreatedResource(BaseModel):
+    type: str
+    id: str
+
+
+class ActionResult(BaseModel):
+    id: str
+    type: str
+    created_resources: List[CreatedResource]
```
Kilo59: Should the `created_resources` ever be modified or appended to after creation?

If not I think we should have mypy enforce this by behavior by making this a `Sequence[CreatedResource]`
```python
from typing import Sequence

...

class ActionResult(BaseModel):
    id: str
    type: str
    created_resources: Sequence[CreatedResource]

```

great_expectations/agent/actions/run_onboarding_data_assistant.py
```diff
@@ -0,0 +1,86 @@
...
+from great_expectations.agent.models import RunOnboardingDataAssistantEvent
+from great_expectations.exceptions import StoreBackendError
+
+if TYPE_CHECKING:
+    from great_expectations.core import ExpectationSuite
+
+
+class RunOnboardingDataAssistantAction(AgentAction):
+    def run(self, event: RunOnboardingDataAssistantEvent, id: str) -> ActionResult:
+        expectation_suite_name = f"{event.data_asset_name} onboarding assistant suite"
```
cdkini, line 18: Probably doesn't matter but should we use `_` instead of spaces?

great_expectations/agent/actions/run_onboarding_data_assistant.py
```diff
@@ -0,0 +1,86 @@
...
+            self._context.get_expectation_suite(
+                expectation_suite_name=expectation_suite_name
+            )
+            raise ValueError(
+                f"Onboarding Assistant Expectation Suite `{expectation_suite_name}` already exists. "
+                + "Please rename or delete suite and try again"
+            )
+        except StoreBackendError:
+            # resource is unique
+            pass
```
cdkini, line 32: Could we log here?

great_expectations/agent/actions/run_onboarding_data_assistant.py
```diff
@@ -0,0 +1,86 @@
...
+
+        try:
+            self._context.get_checkpoint(name=checkpoint_name)
+            raise ValueError(
+                f"Onboarding Assistant Checkpoint `{checkpoint_name}` already exists. "
+                + "Please rename or delete Checkpoint and try again"
+            )
+        except StoreBackendError:
+            # resource is unique
+            pass
```
cdkini, line 42: Same here

great_expectations/agent/actions/run_onboarding_data_assistant.py
```diff
@@ -0,0 +1,86 @@
...
+
+        datasource = self._context.get_datasource(datasource_name=event.datasource_name)
+        asset = datasource.get_asset(asset_name=event.data_asset_name)
+        batch_request = asset.build_batch_request()
+
+        data_assistant_result = self._context.assistants.onboarding.run(
+            batch_request=batch_request,
+        )
```
cdkini, line 59: This is where 99% of the effort goes right? There's a lot of complexity here - should we wrap with a `try/except` to cover possible failure cases?

Could be fine as is.

great_expectations/agent/actions/run_onboarding_data_assistant.py
```diff
@@ -0,0 +1,86 @@
...
+        expectation_suite: ExpectationSuite = (
+            data_assistant_result.get_expectation_suite(
+                expectation_suite_name=expectation_suite_name
+            )
+        )
+        self._context.add_or_update_expectation_suite(
```
Kilo59: Why are we using `add_or_update_*()` instead of just `add_*()` for these given we just checked to ensure the `expectation_suite` and `checkpoint` did not exist?

tests/agent/actions/test_run_onboarding_assistant_action.py
```diff
@@ -0,0 +1,64 @@
...
+@pytest.fixture
+def event():
+    return RunOnboardingDataAssistantEvent(
+        type="onboarding_data_assistant_request.received",
+        datasource_name="test-datasource",
+        data_asset_name="test-data-asset",
+    )
+
+
+def test_run_onboarding_data_assistant_event_requires_unique_expectation_suite(
```
cdkini, line 29: Could these tests (or the whole file) be annotated with `@pytest.mark.unit` or `@pytest.mark.integration` as applicable?

great_expectations/agent/event_handler.py
```diff
@@ -26,28 +16,19 @@ class EventHandler:
...
-            return self._handle_run_data_assistant(event_context)
-        elif isinstance(event_context.event, RunCheckpointEvent):
-            return self._handle_run_checkpoint(event_context)
+    def handle_event(self, event: Event, id: str) -> ActionResult:
+        """Transform an Event into an ActionResult."""
+
+        if isinstance(event, RunOnboardingDataAssistantEvent):
+            action = RunOnboardingDataAssistantAction(context=self._context)
+        elif isinstance(event, RunCheckpointEvent):
+            raise NotImplementedError
```
Kilo59, line 25: `non-blocking`
I would seriously consider using `singledispatch` for this.
It's the perfect use-case for when we want to handle something differently according to the type of one of the arguments. And when we expect this list to grow.... but we can't use a match statement because python 3.8 😄 

https://www.educative.io/answers/what-is-singledispatch-descriptor-in-functools-module-in-python

Although this might be better as an isolated followup.

tests/agent/actions/test_run_onboarding_assistant_action.py
```diff
@@ -0,0 +1,64 @@
...
+    context, event
+):
+    action = RunOnboardingDataAssistantAction(context=context)
+    id = "096ce840-7aa8-45d1-9e64-2833948f4ae8"
+    context.get_checkpoint.side_effect = StoreBackendError("test-message")
+
+    with pytest.raises(ValueError):
```
Kilo59:
```suggestion
    with pytest.raises(ValueError, match=r"Expectation Suite"):
```

tests/agent/actions/test_run_onboarding_assistant_action.py
```diff
@@ -0,0 +1,64 @@
...
+        action.run(event, id=id)
+
+
+def test_run_onboarding_data_assistant_event_requires_unique_checkpoint(context, event):
+    action = RunOnboardingDataAssistantAction(context=context)
+    id = "096ce840-7aa8-45d1-9e64-2833948f4ae8"
+    context.get_expectation_suite.side_effect = StoreBackendError("test-message")
+
+    with pytest.raises(ValueError):
```
Kilo59:
```suggestion
    with pytest.raises(ValueError, match=r"Checkpoint"):
```

great_expectations/agent/actions/action.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class AgentAction(ABC):
+    def __init__(self, context: CloudDataContext):
+        self._context = context
+
+    @abstractmethod
+    def run(self, event: Event, id: str) -> ActionResult:
```
Kilo59: `non-blocking`
Could also use a `Protocol` for this.

joshua-stauffer: `Protocol` doesn't work here, since we're defining default behavior in the `__init__` method.

great_expectations/agent/actions/action.py
joshua-stauffer: correct, `created_resources` is expected to be static. updated in [72d707b](https://github.com/great-expectations/great_expectations/pull/7882/commits/72d707b8e3879af37f412ef03fe14b74bd851985)

great_expectations/agent/actions/run_onboarding_data_assistant.py
joshua-stauffer: sure, happy to update, but curious why? i'm not immediately seeing the benefit, since this is the successful path and we'll reach this most times the code runs.

joshua-stauffer: but absolutely need to add logging to the whole agent - will address that in a follow up PR

great_expectations/agent/actions/run_onboarding_data_assistant.py
joshua-stauffer: if there isn't a path to retry, then raising an exception here is appropriate. This code is running in a thread, so if the job fails we need the thread to exit. Reporting the failure is the responsibility of the caller, currently `GXAgent`

great_expectations/agent/actions/run_onboarding_data_assistant.py
joshua-stauffer: sure, this is updated in [bd77fb9](https://github.com/great-expectations/great_expectations/pull/7882/commits/bd77fb9d20972ec8ba1c2dff7eee5d308f6fb4c7). `CloudContext` doesn't have `add_checkpoint` method though [per the docs](https://docs.greatexpectations.io/docs/reference/api/data_context/CloudDataContext_class) so i'll keep using `add_or_update`

tests/agent/actions/test_run_onboarding_assistant_action.py
joshua-stauffer: yeah absolutely, that's a great idea! how do i mark the whole module? i can update all of the agent tests accordingly

great_expectations/agent/event_handler.py
joshua-stauffer: yeah, that's a valid solution, but in my experience people have a more difficult time reading and understanding `singledispatch` methods. Let's go with the less magical solution, and if in the future this gets unwieldy we can implement this in a different way.

great_expectations/agent/actions/run_onboarding_data_assistant.py
joshua-stauffer: since this is a value prominently displayed in the UI i think spaces are cleaner - happy to update if there is more concern around this, but as far as i know we fully support spaces in resource names.

tests/agent/actions/test_run_onboarding_assistant_action.py
joshua-stauffer: thanks, added in [0aed60c](https://github.com/great-expectations/great_expectations/pull/7882/commits/0aed60cfa7720c5f6f5de7dcec156cd0e0afe8d9)

tests/agent/actions/test_run_onboarding_assistant_action.py
joshua-stauffer: thanks, added in [0aed60c](https://github.com/great-expectations/great_expectations/pull/7882/commits/0aed60cfa7720c5f6f5de7dcec156cd0e0afe8d9)

great_expectations/agent/actions/run_onboarding_data_assistant.py
wookasz: Agreed that spaces are more visually appealing

great_expectations/agent/event_handler.py
wookasz: An alternative would be using the gilded rose approach, put a method that takes a context and returns an action on the event object itself

great_expectations/agent/actions/run_onboarding_data_assistant.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class RunOnboardingDataAssistantAction(AgentAction):
+    def run(self, event: RunOnboardingDataAssistantEvent, id: str) -> ActionResult:
+        expectation_suite_name = f"{event.data_asset_name} onboarding assistant suite"
+        checkpoint_name = f"{event.data_asset_name} onboarding assistant checkpoint"
+
+        # ensure resources we create don't already exist before we run the Data Assistant.
+        try:
+            self._context.get_expectation_suite(
```
wookasz, line 23: Consider appending a unique suffix, maybe a timestamp, and having this always succeed. The user can always rename the entity as they wish after the fact

great_expectations/agent/actions/run_onboarding_data_assistant.py
```diff
@@ -0,0 +1,84 @@
...
+from great_expectations.exceptions import StoreBackendError
+
+if TYPE_CHECKING:
+    from great_expectations.core import ExpectationSuite
+
+
+class RunOnboardingDataAssistantAction(AgentAction):
+    def run(self, event: RunOnboardingDataAssistantEvent, id: str) -> ActionResult:
+        expectation_suite_name = f"{event.data_asset_name} onboarding assistant suite"
+        checkpoint_name = f"{event.data_asset_name} onboarding assistant checkpoint"
```
wookasz, line 19: Have the names been run through product/design. There may be an agent branding opportunity here

great_expectations/agent/actions/run_onboarding_data_assistant.py
wookasz: @joshua-stauffer agree, logging here would just be noisy

great_expectations/agent/actions/run_onboarding_data_assistant.py
```diff
@@ -0,0 +1,84 @@
...
+            self._context.get_checkpoint(name=checkpoint_name)
+            raise ValueError(
+                f"Onboarding Assistant Checkpoint `{checkpoint_name}` already exists. "
+                + "Please rename or delete Checkpoint and try again"
+            )
+        except StoreBackendError:
+            # resource is unique
+            pass
+
+        datasource = self._context.get_datasource(datasource_name=event.datasource_name)
```
wookasz, line 44: We really should use IDs instead of names to ensure consistency. What happens when this event sits in the queue for a while, and the datasource is renamed in the meantime?

wookasz (approved):

tests/agent/actions/test_run_onboarding_assistant_action.py
Kilo59: https://docs.pytest.org/en/7.3.x/example/markers.html#marking-whole-classes-or-modules

great_expectations/agent/actions/action.py
Kilo59: I think it's fine not to use a Protocol, but you **can** define default behavior in a Protocol.
https://mypy.readthedocs.io/en/stable/protocols.html?highlight=Protocol#defining-subprotocols-and-subclassing-protocols
>You can also include default implementations of methods in protocols. If you explicitly subclass these protocols you can inherit these default implementations.

great_expectations/agent/actions/run_onboarding_data_assistant.py
joshua-stauffer: hadn't considered that, but it's a good idea - i'll reach out and we can circle back and update this if/when there's direction

Kilo59:
I think you can ignore the docusaurus errors but the mypy issues will need to be addressed.
LMK, if you'd like some help with these.
```
great_expectations/agent/actions/run_onboarding_data_assistant.py:16: error:
Argument 1 of "run" is incompatible with supertype "AgentAction"; supertype
defines the argument type as
"Union[RunOnboardingDataAssistantEvent, RunCheckpointEvent]"  [override]
        def run(self, event: RunOnboardingDataAssistantEvent, id: str) -> ...
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
great_expectations/agent/actions/run_onboarding_data_assistant.py:16: note: This violates the Liskov substitution principle
great_expectations/agent/actions/run_onboarding_data_assistant.py:16: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
great_expectations/agent/actions/run_onboarding_data_assistant.py:44: error:
Item "BaseDatasource" of
"Union[BaseDatasource, Datasource[Any, Any], LegacyDatasource]" has no attribute
"get_asset"  [union-attr]
            asset = datasource.get_asset(asset_name=event.data_asset_name)
                    ^~~~~~~~~~~~~~~~~~~~
great_expectations/agent/actions/run_onboarding_data_assistant.py:44: error:
Item "LegacyDatasource" of
"Union[BaseDatasource, Datasource[Any, Any], LegacyDatasource]" has no attribute
"get_asset"  [union-attr]
            asset = datasource.get_asset(asset_name=event.data_asset_name)
                    ^~~~~~~~~~~~~~~~~~~~
great_expectations/agent/actions/run_onboarding_data_assistant.py:78: error: No
overload variant of "add_or_update_checkpoint" of "AbstractDataContext" matches
argument type "Dict[str, object]"  [call-overload]
            checkpoint = self._context.add_or_update_checkpoint(**checkpoi...
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~...
great_expectations/agent/actions/run_onboarding_data_assistant.py:78: note: Possible overload variants:
great_expectations/agent/actions/run_onboarding_data_assistant.py:78: note:     def add_or_update_checkpoint(self, name: str = ..., id: Optional[str] = ..., config_version: Union[int, float] = ..., template_name: Optional[str] = ..., module_name: str = ..., class_name: str = ..., run_name_template: Optional[str] = ..., expectation_suite_name: Optional[str] = ..., batch_request: Optional[Dict[Any, Any]] = ..., action_list: Optional[Sequence[ActionDict]] = ..., evaluation_parameters: Optional[Dict[Any, Any]] = ..., runtime_configuration: Optional[Dict[Any, Any]] = ..., validations: Optional[List[Dict[Any, Any]]] = ..., profilers: Optional[List[Dict[Any, Any]]] = ..., site_names: Union[str, List[str], None] = ..., slack_webhook: Optional[str] = ..., notify_on: Optional[str] = ..., notify_with: Union[str, List[str], None] = ..., expectation_suite_id: Optional[str] = ..., default_validation_id: Optional[str] = ..., checkpoint: None = ...) -> Checkpoint
great_expectations/agent/actions/run_onboarding_data_assistant.py:78: note:     def add_or_update_checkpoint(self, name: None = ..., id: None = ..., config_version: Union[int, float] = ..., template_name: None = ..., module_name: str = ..., class_name: str = ..., run_name_template: None = ..., expectation_suite_name: None = ..., batch_request: None = ..., action_list: Optional[Sequence[ActionDict]] = ..., evaluation_parameters: None = ..., runtime_configuration: None = ..., validations: None = ..., profilers: None = ..., site_names: None = ..., slack_webhook: None = ..., notify_on: None = ..., notify_with: None = ..., expectation_suite_id: None = ..., default_validation_id: None = ..., checkpoint: Checkpoint = ...) -> Checkpoint
great_expectations/agent/actions/run_onboarding_data_assistant.py:85: error:
Argument "id" to "CreatedResource" has incompatible type "Optional[str]";
expected "str"  [arg-type]
                        id=expectation_suite.ge_cloud_id, type="Expectatio...
                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
great_expectations/agent/agent.py:150: error: Argument "event" to
"handle_event" of "EventHandler" has incompatible type
"Union[RunOnboardingDataAssistantEvent, RunCheckpointEvent, None]"; expected
"Union[RunOnboardingDataAssistantEvent, RunCheckpointEvent]"  [arg-type]
                event=event_context.event, id=event_context.correlation_id
                      ^~~~~~~~~~~~~~~~~~~
Found 6 errors in 2 files (checked 484 source files)
```

great_expectations/agent/actions/run_onboarding_data_assistant.py
Kilo59: It's inherited from `AbstractDataContext`
https://docs.greatexpectations.io/docs/reference/api/data_context/abstractdatacontext_class/#great_expectations.data_context.AbstractDataContext

great_expectations/agent/actions/run_onboarding_data_assistant.py
```diff
@@ -0,0 +1,94 @@
...
+        checkpoint = self._context.add_checkpoint(**checkpoint_config)  # type: ignore[arg-type]
+
+        return ActionResult(
+            id=id,
+            type=event.type,
+            created_resources=[
+                CreatedResource(
+                    id=expectation_suite.ge_cloud_id, type="ExpectationSuite"  # type: ignore[arg-type]
+                ),
+                CreatedResource(id=checkpoint.ge_cloud_id, type="Checkpoint"),  # type: ignore[arg-type]
```
Kilo59: Please add a comment to these explaining what the error mypy is complaining about.
Or just copy and paste the exact warning.

For `add_checkpoint`, & `ActionResult/CreatedResource` that is.

Either it's being used incorrectly or we need to update some function signatures.

great_expectations/agent/actions/agent_action.py
```diff
@@ -0,0 +1,30 @@
...
+    id: str
+
+
+class ActionResult(BaseModel):
+    id: str
+    type: str
+    created_resources: Sequence[CreatedResource]
+
+
+_TEvent = TypeVar("_TEvent", bound=Event)
```
Kilo59: The convention we've been using for type vars is to have a `T` suffix.
```suggestion
_EventT = TypeVar("_EventT", bound=Event)
```

great_expectations/agent/actions/run_onboarding_data_assistant.py
joshua-stauffer: `ge_cloud_id` is typed as an optional string:
```
great_expectations/agent/actions/run_onboarding_data_assistant.py:90: error: Argument "id" to "CreatedResource" has incompatible type "Optional[str]"; expected "str"  [arg-type]
great_expectations/agent/actions/run_onboarding_data_assistant.py:92: error: Argument "id" to "CreatedResource" has incompatible type "Optional[str]"; expected "str"  [arg-type]
```
but since we just received it from cloud, we know it's a valid ID

great_expectations/agent/actions/run_onboarding_data_assistant.py
joshua-stauffer: yeah this edge case is possible, but `get_datasource` only accepts a name, not ID, so we'll live with it until it becomes an issue

joshua-stauffer: `get_checkpoint` and `get_expectation_suite` both do accept IDs, but there we're just verifying that the name is unique, so our usage is appropriate

great_expectations/agent/actions/run_onboarding_data_assistant.py
joshua-stauffer: well, we received it from cloud so we assume its a valid ID.. i'll add a check to make that explicit

great_expectations/agent/actions/run_onboarding_data_assistant.py
wookasz: Agreed about the checkpoint and expectation suite, comment was specific to the datasource. There actually is an issue here outside just the edge case. We are making the statement that data sent over the MQ is not "sensitive." Names may be considered sensitive.

great_expectations/agent/actions/run_onboarding_data_assistant.py
joshua-stauffer: the checkpoint usage is taken from the docs, and creates a number of mypy errors:
```
great_expectations/agent/actions/run_onboarding_data_assistant.py:83: error: Argument 1 to "add_checkpoint" of "CloudDataContext" has incompatible type "**Dict[str, object]"; expected "Optional[str]"  [arg-type]
great_expectations/agent/actions/run_onboarding_data_assistant.py:83: error: Argument 1 to "add_checkpoint" of "CloudDataContext" has incompatible type "**Dict[str, object]"; expected "Union[int, float]"  [arg-type]
great_expectations/agent/actions/run_onboarding_data_assistant.py:83: error: Argument 1 to "add_checkpoint" of "CloudDataContext" has incompatible type "**Dict[str, object]"; expected "str"  [arg-type]
great_expectations/agent/actions/run_onboarding_data_assistant.py:83: error: Argument 1 to "add_checkpoint" of "CloudDataContext" has incompatible type "**Dict[str, object]"; expected "Optional[Dict[Any, Any]]"  [arg-type]
great_expectations/agent/actions/run_onboarding_data_assistant.py:83: error: Argument 1 to "add_checkpoint" of "CloudDataContext" has incompatible type "**Dict[str, object]"; expected "Optional[Sequence[ActionDict]]"  [arg-type]
great_expectations/agent/actions/run_onboarding_data_assistant.py:83: error: Argument 1 to "add_checkpoint" of "CloudDataContext" has incompatible type "**Dict[str, object]"; expected "Optional[List[Dict[Any, Any]]]"  [arg-type]
great_expectations/agent/actions/run_onboarding_data_assistant.py:83: error: Argument 1 to "add_checkpoint" of "CloudDataContext" has incompatible type "**Dict[str, object]"; expected "Union[str, List[str], None]"  [arg-type]
great_expectations/agent/actions/run_onboarding_data_assistant.py:83: error: Argument 1 to "add_checkpoint" of "CloudDataContext" has incompatible type "**Dict[str, object]"; expected "Optional[Validator]"  [arg-type]
great_expectations/agent/actions/run_onboarding_data_assistant.py:83: error: Argument 1 to "add_checkpoint" of "CloudDataContext" has incompatible type "**Dict[str, object]"; expected "Optional[Checkpoint]"  [arg-type]
```
as far as i can tell, this is correct usage and works, but please correct me if i'm wrong!

great_expectations/agent/actions/agent_action.py
joshua-stauffer: addressed in [5a5b74f](https://github.com/great-expectations/great_expectations/pull/7882/commits/5a5b74f70ece32713e37ee247935342ed86606cc)

great_expectations/agent/actions/run_onboarding_data_assistant.py
joshua-stauffer: thanks! my thought is to keep this in mind for the future - these names are going to be read a lot in the UI, so keeping them clean is nice, and running this action multiple times in a row is not something i'm anticipating users wanting to do. In the case that we find out that isn't the case, it's an easy change

Kilo59 (approved):

PR #7885, opened 2023-05-15
[SPIKE] Research for Cloud Quickstart compatibility

cdkini:
- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```

### DIFF
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -912,11 +912,11 @@ class StoreValidationResultAction(ValidationAction):
 
         checkpoint_ge_cloud_id = None
         if self._using_cloud_context and checkpoint_identifier:
-            checkpoint_ge_cloud_id = checkpoint_identifier.id
+            checkpoint_ge_cloud_id = checkpoint_identifier.id or None
 
         expectation_suite_ge_cloud_id = None
         if self._using_cloud_context and expectation_suite_identifier:
-            expectation_suite_ge_cloud_id = expectation_suite_identifier.id
+            expectation_suite_ge_cloud_id = expectation_suite_identifier.id or None
 
         return self.target_store.set(
             validation_result_suite_identifier,
--- a/great_expectations/validation_operators/validation_operators.py
+++ b/great_expectations/validation_operators/validation_operators.py
@@ -415,7 +415,9 @@ class ActionListValidationOperator(ValidationOperator):
                 validation_result = async_batch_validation_result.result()
                 validation_result.meta["validation_id"] = validation_id
                 validation_result.meta["checkpoint_id"] = (
-                    checkpoint_identifier.id if checkpoint_identifier else None
+                    checkpoint_identifier.id
+                    if checkpoint_identifier and checkpoint_identifier.id
+                    else None
                 )
 
                 batch_actions_results = self._run_actions(
--- a/quickstart.py
+++ b/quickstart.py
@@ -1,0 +1,40 @@
+import great_expectations as gx
+
+
+# Set up
+context = gx.get_context()
+
+
+# Connect to data
+validator = context.sources.pandas_default.read_csv(
+    "https://raw.githubusercontent.com/great-expectations/gx_tutorials/main/data/yellow_tripdata_sample_2019-01.csv",
+    asset_name="taxi_asset",
+)
+
+# Create Expectations
+validator.expect_column_values_to_not_be_null("pickup_datetime")
+validator.expect_column_values_to_be_between("passenger_count", auto=True)
+
+validator.expectation_suite_name = "taxi_suite"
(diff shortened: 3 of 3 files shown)


great_expectations/checkpoint/actions.py
```diff
@@ -912,11 +912,11 @@ def _basic_run(
 
         checkpoint_ge_cloud_id = None
         if self._using_cloud_context and checkpoint_identifier:
-            checkpoint_ge_cloud_id = checkpoint_identifier.id
+            checkpoint_ge_cloud_id = checkpoint_identifier.id or None
```
cdkini, line 915: We often interchange `""` and `None` in OSS code but tend to use a bit more explicitness in the Cloud backend. Since our schemas validate this data to be nullable UUIDs, passing forth empty strings results in runtime errors.

The approach here may also be changing `GXCloudIdentifier.id`'s type to a nullable string (instead of just a string).

great_expectations/validation_operators/validation_operators.py
```diff
@@ -415,7 +415,9 @@ def run(
                 validation_result = async_batch_validation_result.result()
                 validation_result.meta["validation_id"] = validation_id
                 validation_result.meta["checkpoint_id"] = (
-                    checkpoint_identifier.id if checkpoint_identifier else None
+                    checkpoint_identifier.id
+                    if checkpoint_identifier and checkpoint_identifier.id
+                    else None
```
cdkini, line 420: Same issue - was resulting in:
```
{
    "validation_id": None,
    "checkpoint_id": "",
    ...
}
```

quickstart.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+# Set up
+context = gx.get_context()
+
+
+# Connect to data
+validator = context.sources.pandas_default.read_csv(
+    "https://raw.githubusercontent.com/great-expectations/gx_tutorials/main/data/yellow_tripdata_sample_2019-01.csv",
+    asset_name="taxi_asset",
```
cdkini, line 11: Need to specify an asset name to get this to persist

quickstart.py
```diff
@@ -0,0 +1,40 @@
...
+)
+
+# Create Expectations
+validator.expect_column_values_to_not_be_null("pickup_datetime")
+validator.expect_column_values_to_be_between("passenger_count", auto=True)
+
+validator.expectation_suite_name = "taxi_suite"
+context.add_expectation_suite(expectation_suite=validator.expectation_suite)
```
cdkini, line 19: Need to persist the suite so we can reference in the validations list below

quickstart.py
```diff
@@ -0,0 +1,40 @@
...
+    data_context=context,
+    validations=[
+        {
+            "expectation_suite_name": "taxi_suite",
+            "batch_request": {
+                "datasource_name": "default_pandas_datasource",
+                "data_asset_name": "taxi_asset",
+            },
+        }
+    ],
```
cdkini, line 34: We can easily wrap this is some sort of abstraction but for the time being, I'm focusing on getting this functional

quickstart.py
```diff
@@ -0,0 +1,40 @@
...
+)
+
+checkpoint_result = checkpoint.run()
+
+# View results
+context.view_validation_result(checkpoint_result)
```
cdkini, line 40: Does not currently exist but I think the primary hurdle around this whole task is getting the above to work - I'm confident the implementation for this method is a small lift so let's just assume it works for now.

quickstart.py
Kilo59: At the beginning of last week we talked about auto-generating an asset name when doing this for cloud and making this change to the quickstart unnecessary. 

Did we decide against that? Or is this just an intermediate step to get the quick-start working while we make the more substantive changes?

cc @DrewHoo

cdkini: Ah yeah I considered that a follow-up step (but thanks for calling it out!). I'm just trying to get a barebones version working so we can iterate on it in the epic.

great_expectations/checkpoint/actions.py
Kilo59: Good catch.
IMO having strings that could be empty is often a confusing pattern.

Kilo59 (approved):

quickstart.py
Kilo59: Also, we could temporaliy make this quickstart prevent asset collisions here by making `asset_name=f"taxi_asset-{uuid.uuid4()}"`, or something similar.

NathanFarmer: Why do we need to persist the quickstart asset?

quickstart.py
NathanFarmer: I thought the quickstart was supposed to use a `Validator` in the `Checkpoint`.

PR #7894, opened 2023-05-15
[DOCS] Less verbose logging during docs build

anthonyburdi:
Default to less verbose logging during docs build. Parameters added to use more verbose output during debugging.

### DIFF
--- a/docs/docusaurus/scripts/remark-named-snippets/snippet.js
+++ b/docs/docusaurus/scripts/remark-named-snippets/snippet.js
@@ -173,9 +173,22 @@ function getDirs() {
   return manualDirs.concat(versionDirs)
 }
 
+function processVerbose() {
+  const args = process.argv.slice(2)
+
+  let verbose = false
+  if (args.includes('verbose') || args.includes('-v') || args.includes('--verbose')) {
+    verbose = true
+    console.log('Using verbose mode. Printing all snippets.')
+  }
+  return verbose
+}
+
 function main(verbose = false) {
   const snippets = parseSourceDirectories(getDirs())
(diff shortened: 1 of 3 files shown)


docs/docusaurus/scripts/remark-named-snippets/snippet.js
```diff
@@ -191,12 +191,13 @@ function main() {
     delete snippet.file // Remove duplicate filename to clean up stdout
     out[file].push(snippet)
   }
-  console.log(out)
+  if (verbose) {
+    console.log(out)
+  }
```
Shinnnyshinshin, line 209: ❤️

docs/docusaurus/scripts/remark-named-snippets/snippet.js
```diff
@@ -173,7 +173,7 @@ function getDirs() {
   return manualDirs.concat(versionDirs)
 }
 
-function main() {
+function main(verbose = false) {
```
Shinnnyshinshin, line 187: Where would we (if ever) run this function as `verbose = true`?

Shinnnyshinshin: If so, how would we do this?

anthonyburdi: Good call, in https://github.com/great-expectations/great_expectations/pull/7894/commits/19b2b1f9a76f64e6182196b5fe293a0cb168c9c6 I changed `yarn snippet-check` to use the verbose mode which you can run locally to check all snippets.

alexsherstinsky (approved):
LGTM -- thank you!

PR #7898, opened 2023-05-15
[MAINTENANCE] Optional dependency tests

anthonyburdi:
- Tests for optional dependencies
- Add compatibility stage to PR CI pipeline
- Add `compatibility` pytest marker
- Move `test_not_imported.py` into `tests/compatibility`
- Update `invoke tests` to not run compatibility tests by default (these should be run in an environment without optional dependencies installed).


tests/compatibility/test_compatibility.py
```diff
@@ -0,0 +1,17 @@
...
+
+# TODO: parametrize test_error_raised_when_optional_import_not_installed over all optional dependencies
+# TODO: Test to make sure we don't have optional dependencies that are not in our set of tests in test_error_raised_when_optional_import_not_installed
+
+
+@pytest.mark.compatibility
+def test_error_raised_when_optional_import_not_installed():
+    from great_expectations.compatibility.pyspark import pyspark
+
+    breakpoint()
```
billdirks: This looks like a stray breakpoint

anthonyburdi: 😅 Removed in https://github.com/great-expectations/great_expectations/pull/7898/commits/42383734c7093ec2e58f3fcb5eb5c04332b8e4c6

great_expectations/compatibility/azure.py
```diff
@@ -6,6 +6,11 @@
     "azure blob storage components are not installed, please 'pip install azure-storage-blob azure-identity azure-keyvault-secrets'"
 )
 
+try:
+    import azure
```
alexsherstinsky: Is this used?  If not, then maybe we do not need to import it?

great_expectations/compatibility/google.py
```diff
@@ -6,6 +6,11 @@
     "google cloud storage components are not installed, please 'pip install google-cloud-storage google-cloud-secret-manager'"
 )
 
+try:
+    import google
```
alexsherstinsky: Same question -- if it is not used for the direct `google.<something>`, then maybe we do not need it at top level?

tests/compatibility/test_compatibility.py
```diff
@@ -0,0 +1,33 @@
...
+libraries = [pyspark, boto3, botocore, azure, docstring_parser, google]
+
+@pytest.mark.unit
+@pytest.mark.parametrize(
+    ["package"],
+    [
+        pytest.param(
+            library,
+        )
+        for library in libraries
```
alexsherstinsky: I see what is being done here; I would propose that since we do not use `google` and `azure` at top level in the logic of GX OSS, then we should not be importing them, and the package tests here would be on one of the entities from each package that is actually being used (and imported in the compatibility set of modules).  Thanks!

alexsherstinsky (changes requested):
Suggested a slight modification.

great_expectations/compatibility/azure.py
anthonyburdi: Good call, removed 🙇

great_expectations/compatibility/google.py
anthonyburdi: Good call, removed 🙇

tests/compatibility/test_compatibility.py
anthonyburdi: Good call, removed 🙇

anthonyburdi:
OBE

PR #7899, opened 2023-05-16
[MAINTENANCE] type-checking `checkpoint/actions.py`

Kilo59:
Type check `checkpoint/actions.py`
There are no runtime changes in this PR.

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```

### DIFF
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -274,9 +274,7 @@ class SlackNotificationAction(ValidationAction):
                     data_docs_pages = payload[action_names]
 
         # Assemble complete GX Cloud URL for a specific validation result
-        data_docs_urls: list[
-            dict[str, str | None]
-        ] = self.data_context.get_docs_sites_urls(
+        data_docs_urls: list[dict[str, str]] = self.data_context.get_docs_sites_urls(
             resource_identifier=validation_result_suite_identifier
         )
 
@@ -853,7 +851,7 @@ class StoreValidationResultAction(ValidationAction):
         self,
         data_context: AbstractDataContext,
         target_store_name: Optional[str] = None,
-    ) -> Optional[GXCloudResourceRef]:
+    ) -> None:
         super().__init__(data_context)
         if target_store_name is None:
             self.target_store = data_context.stores[data_context.validations_store_name]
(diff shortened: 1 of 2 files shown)


great_expectations/checkpoint/actions.py
```diff
@@ -237,7 +237,7 @@ def __init__(
         self.notify_with = notify_with
         self.show_failed_expectations = show_failed_expectations
 
-    def _run(
+    def _run(  # type: ignore[override] # signature does not match parent
```
cdkini, line 240: 😢 how can we remedy this issue? It feels like a violation of sound OOO design and I'm quite confident there are probably some latent bugs around as a result of it.

great_expectations/checkpoint/actions.py
```diff
@@ -1182,7 +1181,7 @@ def _run(
         # <snippet name="great_expectations/checkpoint/actions.py get_docs_sites_urls">
         docs_site_urls_list = self.data_context.get_docs_sites_urls(
             resource_identifier=validation_result_suite_identifier,
-            site_names=self._site_names,
+            site_names=self._site_names,  # type: ignore[arg-type] # ._site_names could be None
```
cdkini, line 1184: Could we instead do `self._site_names or []`? (assuming this is a nullable list or collection)

cdkini (approved):

great_expectations/checkpoint/actions.py
Kilo59: Good call, that sounds right. I'll test that out, and do it as part of this final `checkpoint` typing PR.
* https://github.com/great-expectations/great_expectations/pull/7901

great_expectations/checkpoint/actions.py
Kilo59:
```python
   # parent
    def _run(
        self,
        validation_result_suite: ExpectationSuiteValidationResult,
        validation_result_suite_identifier: Union[
            ValidationResultIdentifier, GXCloudIdentifier
        ],
        data_asset,
        expectation_suite_identifier=None,
        checkpoint_identifier=None,
    ):
```
```python
    # child
    def _run(  # type: ignore[override] # signature does not match parent
        self,
        validation_result_suite: ExpectationSuiteValidationResult,
        validation_result_suite_identifier: Union[
            ValidationResultIdentifier, GXCloudIdentifier
        ],
        data_asset=None,
        payload=None,
        expectation_suite_identifier=None,
        checkpoint_identifier=None,
    ):
```

Kilo59: The difference is `data_asset` not being required and the addition of the optional `payload`.
So we could either add `payload` to the parent signature, and make `data_asset` optional.... or we could replace all the individual parameters with a single `options`-like type. 

We already have an issue with a lot of these methods only use a few of these parameters. This particular method only uses `payload` (pulse the required params).

https://refactoring.guru/smells/long-parameter-list

PR #7905, opened 2023-05-16
[MAINTENANCE] `CloudDataContext` `ExpectationsStore` can only request one Suite at a time

NathanFarmer:
- [] Description of PR changes above includes a link to [an existing GitHub issue](https://github.com/great-expectations/great_expectations/issues)
- [X] Code is linted

    ```
    black .

    ruff . --fix
    ```


great_expectations/data_context/store/store.py
```diff
@@ -98,7 +98,9 @@ def _validate_key(self, key: DataContextKey) -> None:
         # STORE_BACKEND_ID_KEY always validated
         if key == StoreBackend.STORE_BACKEND_ID_KEY:
             return
-        elif not isinstance(key, self.key_class):
+        elif isinstance(key, self.key_class):
```
billdirks, line 101: This is the same logic as before right just a little cleaned up?
You could also do this or just removed the last `else` to remove a branch:
```
if key == StoreBackend.STORE_BACKEND_ID_KEY or isinstance(key, self.key_class):
    return
raise TypeError(
                f"key must be an instance of {self.key_class.__name__}, not {type(key)}"
            )
```
Also, the raised `TypeError` seems a little wrong since `StoreBackend.STORE_BACKEND_ID_KEY` is also valid but maybe the user doesn't care about that?

billdirks (approved):
One small comment/question but lgtm. Thanks!

great_expectations/data_context/store/store.py
NathanFarmer: Yes, the logic is the same. I prefer the `else` because I find it more readable, and I doubt the compiler treats it any differently.

The user doesn't know about `StoreBackend.STORE_BACKEND_ID_KEY`. It references a hidden file (`.ge_store_backend_id`) in the `FileDataContext` `ExpectationsStore`.

tyler-hoffman (approved):

PR #7911, opened 2023-05-17
[BUGFIX] fix `add_or_update_expectation_suite` update path

dctalbot:
- This PR fixes an issue in which sequential calls to `ctx.add_or_update_expectation_suite(expectation_suite_name="foo")` would result in an exception (it should be idempotent).
- The bug was as follows: GX Cloud requires the Expectation Suite's Cloud ID in order to request a PUT. This ID was not being set properly in the `add_or_update_expectation_suite` method.

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2941,6 +2941,7 @@ def add_expectation_suite(
 
         Raises:
             DataContextError: A suite with the same name already exists (and `overwrite_existing` is not enabled).
+            ValueError: The arguments provided are invalid.
```
NathanFarmer, line 2941: 🙇🏻

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3105,30 +3103,47 @@ def add_or_update_expectation_suite(
...
-            execution_engine_type: Name of the execution engine type.
+            data_asset_type: Type of Data Asset to associate with this suite.
+            execution_engine_type: Name of the Execution Engine type.
             meta: Metadata related to the suite.
-            expectation_suite: An existing ExpectationSuite object you wish to persist.
+            expectation_suite: The `ExpectationSuite` object you wish to persist.
 
         Returns:
-            A new ExpectationSuite or an updated once (depending on whether or not it existed before this method call).
+            The persisted `ExpectationSuite`.
```
NathanFarmer, line 3127: ❤️

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -3105,30 +3103,47 @@ def add_or_update_expectation_suite(
...
+                ge_cloud_id=id,
+                expectations=expectations,
+                evaluation_parameters=evaluation_parameters,
+                data_asset_type=data_asset_type,
+                execution_engine_type=execution_engine_type,
+                meta=meta,
+            )
+
+        try:
+            got_expectation_suite = self.get_expectation_suite(
```
NathanFarmer: I would typically use an underscore here to indicate the variable is not used like: `_ = self.get_expectation_suite()`, but I like the variable name 🤣

NathanFarmer (changes requested):
Can you please add a test to `tests/data_context/cloud_data_context/test_expectation_suite_crud.py`, so we can't have a regression on this?

great_expectations/data_context/data_context/abstract_data_context.py
dctalbot: this variable is used!

NathanFarmer (approved):

PR #7935, opened 2023-05-18
[DOCS] Technical tags in Versioned Docs reference correct version

anthonyburdi:
Technical tasks were not adding the version number to the path. This PR adds a function to `prepare_prior_versions.py` to add the appropriate version into the path while building the documentation.


TrangPham (commented):
Should we add a unit test or smoke test for this?

Shinnnyshinshin (approved):
LGTM. Thank you for the refactor, and adding the test :)  Please go ahead and merge after the linting is taken care of

anthonyburdi:
> Should we add a unit test or smoke test for this?

Good call! Added in https://github.com/great-expectations/great_expectations/pull/7935/commits/1fb542e83e4a5fe1f1f2f825cf3444a7ba8ed76b. I think in a follow up we can add tests for the other methods and perhaps encapsulate them better.

PR #7989, opened 2023-05-24
[BUGFIX] Cloud - Fix `context.sources.update_*()` `POST` instead of `PUT` calls

Kilo59:
`context.sources.update_*()` calls were deleting the local datasource objects and the `id` field before making the external HTTP call.
This resulted in a failing`POST` for every call instead of a `PUT`.

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```


great_expectations/datasource/fluent/sources.py
```diff
@@ -512,6 +512,18 @@ def update_datasource(
                 datasource_name,  # type:ignore[arg-type] # datasource_name is expected to be a str from assignment above
                 datasource_type,
             )
+
+            # preserve any pre-existing id for usage with cloud
+            id_: uuid.UUID | None = getattr(
+                self._data_context.datasources.get(datasource_name), "id", None
+            )
```
cdkini, line 519: Do we need to use `getattr` here? Could we split this up by trying to retrieve the datasource first and then conditionally trying to access the `id` attr?

cdkini (approved):

great_expectations/datasource/fluent/sources.py
Kilo59: I'm using `gettattr` because the `.get()` call could return `None` and I don't want to add another conditional.

cdkini: hmm okay I've always avoided `getattr` but I'm good if you are

Kilo59: I don't love `getattr` either. But this also avoids creating another variable.
I'd like to refactor this and the other code block that has the same logic and I can remove the use of `getattr` then without adding a new variable.

PR #7995, opened 2023-05-24
[BUGFIX] `meta_notes` rendered inline can fail schema validation for a valid configuration

NathanFarmer:
We were allowing users to pass a string into the meta notes dictionary, but it would fail schema validation because we weren't adding that string to a list of length 1.

- [X] Code is linted

    ```
    black .

    ruff . --fix
    ```


great_expectations/render/renderer_configuration.py
```diff
@@ -370,15 +370,26 @@ def _validate_for_row_condition(cls, values: dict) -> dict:
...
         ):
-            values["meta_notes"] = values["result"].expectation_config.meta.get("notes")
+            meta_notes = values["result"].expectation_config.meta.get("notes")
         else:
-            values["meta_notes"] = values["configuration"].meta.get("notes")
+            meta_notes = values["configuration"].meta.get("notes")
+
+        meta_notes_content = meta_notes.get("content") if meta_notes else None
+
+        if isinstance(meta_notes_content, (list, tuple)):
```
Kilo59, line 387: https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence
```suggestion
        if isinstance(meta_notes_content, Sequence):
```

great_expectations/render/renderer_configuration.py
```diff
@@ -370,15 +370,26 @@ def _validate_for_row_condition(cls, values: dict) -> dict:
 
     @root_validator()
     def _validate_for_meta_notes(cls, values: dict) -> dict:
-        meta_notes: Optional[MetaNotes]
+        meta_notes: Optional[
+            dict[str, Optional[dict[str, list[str] | tuple[str] | str]]]
```
Kilo59, line 374: `non-blocking`
It seems like there is some expected structure to this dict, can we use a TypedDict here?

Kilo59 (approved):

great_expectations/render/renderer_configuration.py
NathanFarmer: No can do! It could be a `str` which is also a `Sequence`.

great_expectations/render/renderer_configuration.py
NathanFarmer: `MetaNotes` is a `TypedDict`, but we allow for more loose type constraints at the API level and then this validator transforms the dict into a `MetaNotes` dict. I prefer to use the primitives here or else we end up with something like `MetaNotesInput` and `MetaNotes`.

PR #8020, opened 2023-05-27
[FEATURE] Cloud - generate unique asset names for `pandas_default` assets

Kilo59:
Generate unique asset names for `PandasDatasource` `read_*()` assets when no asset name is provided.
This is so that the assets will be persisted to GX Cloud.

Link to gist demonstrating the frequency of name collisions we can expect using shortened `uuid4` ids.
https://gist.github.com/Kilo59/82f227d9dba4e5cce62bc22b245b2638

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```


great_expectations/datasource/fluent/sources.py
```diff
@@ -43,10 +46,6 @@
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_PANDAS_DATASOURCE_NAME: Final[str] = "default_pandas_datasource"
-
-DEFAULT_PANDAS_DATA_ASSET_NAME: Final[str] = "#ephemeral_pandas_asset"
-
```
Kilo59, line 49: moving these to `constants.py`

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -513,8 +523,19 @@ def _add_asset(
 
         asset_names: Set[str] = self.get_asset_names()
 
+        in_cloud_context: bool = False
+        if self._data_context:
+            in_cloud_context = self._data_context._datasource_store.cloud_mode
```
cdkini, line 528: Any pros/cons to use `isinstance(self._data_context, CloudDataContext)` here? I think this works but I don't love the direct access to the private `_datasource_store`

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -513,8 +523,19 @@ def _add_asset(
...
-            if asset_name in asset_names:
+            if in_cloud_context:
+                # In cloud mode, we need to generate a unique name for the asset so that it gets persisted
+                asset_name = f"{asset.type}-{_short_id()}"
+                logger.info(
+                    f"Generating unique name for '{DEFAULT_PANDAS_DATA_ASSET_NAME}' asset '{asset_name}'"
+                )
+                asset.name = asset_name
+            elif asset_name in asset_names:
                 self.delete_asset(asset_name=asset_name)
```
cdkini, line 539: We're just assuming no collisions right? Was thinking this should be an `if` but no way this will ever be a material issue

cdkini (approved):

great_expectations/datasource/fluent/pandas_datasource.py
Kilo59: An instance check makes `PandasDasasource` directly dependent on `CloudDataContext`.
That is its own code smell and leads to an increased chance of a circular import.

I think the bigger problem is the conditional and special casing itself, but I don't see a good solution here.

cdkini: That's true 👍🏽

great_expectations/datasource/fluent/pandas_datasource.py
Kilo59: If there's going to be a name collision, it will happen at the return line when we add the asset.
We can expect a naming collision after several million iterations (of the same asset type).

If that's too frequent we can tweak the implementation of `_short_id()` to add a few characters. Right now it's `11`.
https://gist.github.com/Kilo59/82f227d9dba4e5cce62bc22b245b2638

PR #8031, opened 2023-05-30
[MAINTENANCE] Postgres starter reference environment

anthonyburdi:
An example reference environment to get started quickly with a postgres database pre-loaded with data. Reference environments are still experimental and the API is likely to change so please be aware of this until the design and implementations are more stable.


examples/reference_environments/postgres/docker-compose.yaml
```diff
@@ -0,0 +1,41 @@
+services:
```
alexsherstinsky: Are the file suffixes supposed to be `.yaml` or `.yml`?  In our existing `assets/docker` they are the latter.  Thanks.

great_expectations/cli/example.py
```diff
@@ -0,0 +1,58 @@
+import pathlib
+import subprocess
```
alexsherstinsky, line 3: @anthonyburdi May I please recommend trying to use `great_expectations/cli/python_subprocess.py`?  Thank you!

alexsherstinsky (changes requested):
This is really good!  Only minor suggestions and/or questions so far.  Thank you.

examples/reference_environments/postgres/docker-compose.yaml
anthonyburdi: Good question! `.yaml` is preferred for `docker-compose.yaml` but either is OK: https://docs.docker.com/compose/compose-file/03-compose-file/. But it looks like `compose.yaml` is actually the preferred, so I will probably rename to that.

great_expectations/cli/example.py
anthonyburdi: Thanks for the suggestion! I took a look at `execute_shell_command` and I'm not sure I understand the difference for this use case. One difference I see is passing in only the PATH env var vs the default behavior of [inheriting the current process' environment](https://docs.python.org/3/library/subprocess.html#subprocess.run). In our case we want to inherit the environment so that the user environment variables can contain things like credentials for connecting to AWS for example that are then passed through and used in the docker environment.
Is there something else I'm missing?

examples/reference_environments/postgres/docker-compose.yaml
anthonyburdi: Renamed to the new preferred filename `compose.yaml` 🙇

great_expectations/cli/example.py
alexsherstinsky: @anthonyburdi I am not sure about the environment -- I feel that it would be worth a test (looking at the code it should work as you describe).  In my mind, the biggest benefit of our own wrapper is that we encapsulate the complexity of and add nice capabilities on top of the raw Python call (e.g., `execute_shell_command_with_progress_polling()`) including friendlier error messages.  Just a suggestion -- please let me know what you think; I am ready to approve if you feel that this is out of scope (I would still want to try it at some point across the codebase -- there is at least one more candidate location that I recently found).  Thanks!

anthonyburdi: Thanks for the suggestion @alexsherstinsky! I think though that we should stick to just `subprocess.run()` here for a couple reasons:
1. Docker progress bars and logging are already pretty comprehensive, so no need to add our own.
2. Our use of `subprocess.run()` is quite simple, we call a single command or run a single script from a working directory.
3. When I do try to implement, the progress bar and docker logging conflict somehow and all I see is a bar stuck at 0%:  image

alexsherstinsky (approved):
LGTM!

PR #8032, opened 2023-05-30
[BUGFIX] Ensure That DataFrame for Fluent Datasource Pandas/Spark DataFrame Data Assets is specified only in one API method

alexsherstinsky:
### Scope
The way that Fluent Datasources (FDS) API enables adding Pandas/Spark DataFrame assets today is still not intuitive.  It causes a jarring user/developer experience, because the user has to specify DataFrame in more than one API call.

While there are no errors raised are raised then the existing code is executed, unfortunately, this process is confusing to the users.  On one hand, if the user specified the DataFrame in add_dataframe_asset(), they expect to get the DataFrame back “on read”.  However, we know that this is impossible, because a DataFrame is a runnable object and is thus unpickleable/unserializable.   But given that the user already specified the DataFrame when creating the DataAsset and then having the user to set the same DataFrame on the DataAsset again raises legitimate questions such as “I already specified my DataFrame once; why do I have to do it again?”

The proposed solution is the following:
a) Change the add_dataframe_asset()  API to include only the name and any other metadata, but not accept the DataFrame as an argument.
b) Modify the DataFrameAsset.build_batch_request() method to take in the DataFrame as the argument.

With this change, the serialization/deserialization of the Project Configuration will continue working properly, and the users will know that they have to specify the DataFrame in exactly one place, and this will fit very well with the data pipeline use cases, which is exactly what this ephemeral DataAsset is designed for.

### Remarks
* JIRA: DX-469/DX-553
* 

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```


great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -591,23 +594,28 @@ def _validate_asset_name(asset_name: Optional[str] = None) -> str:
             asset_name = DEFAULT_PANDAS_DATA_ASSET_NAME
         return asset_name
 
-    def _get_validator(self, asset: _PandasDataAsset) -> Validator:
-        batch_request: BatchRequest = asset.build_batch_request()
+    def _get_validator(
+        self, asset: _PandasDataAsset, dataframe: pd.DataFrame | None = None
+    ) -> Validator:
+        batch_request: BatchRequest
+        if isinstance(asset, DataFrameAsset):
```
alexsherstinsky, line 627: @Kilo59 I am uncertain about this `isinstance()` check.  Is this extra precaution needed (i.e., only `read_dataframe ()` will execute the signature containing the `dataframe` argument, and the asset in this case is guaranteed to always be `DataFrameAsset`)?  Thank you.

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -362,12 +360,17 @@ def _get_reader_options_include(self) -> set[str]:
         )
 
     @public_api
-    def build_batch_request(self) -> BatchRequest:  # type: ignore[override]
-        if self.dataframe is None:
-            raise ValueError(
-                "Cannot build batch request for dataframe asset without a dataframe"
-            )
+    def build_batch_request(self, dataframe: pd.DataFrame) -> BatchRequest:  # type: ignore[override]
```
Kilo59: Why should I have to provide a `dataframe` if I've already added it via `add_dataframe_asset()`?

Kilo59 (changes requested):
I don't understand why we have an `add_dataframe_asset()` method and yet require a `dataframe` to be passed to `build_batch_request`.

This seems like an API change, not a bugfix.

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @Kilo59 We are changing the API, because the existing one causes usability problems.  We are not supposed to quote JIRA or Slack links here in GitHub, but the ticket indicated in the description as well as the Slack thread with @NathanFarmer goes into the detail.  TL/DR: Instead of supplying the DataFrame in `add_dataframe_asset()`, we want to do it in `build_batch_request()` so that there is exactly one place where the DataFrame can and must be supplied.  The bug -- which is a usability bug -- is that with the current API, one can supply the DatFrame as part of the `add_dataframe_asset()` call, but then if the DataContext is reloaded, one has to supply it again.  With the proposed change, since `DataFrame` is an ephemeral entity, `add_dataframe_asset()` will only specify the (serializable) metadata, while the non-serializable object, `DataFrame` will go right into the `BatchRequest` via `build_batch_request()`.  Happy to discuss.  Thanks.

alexsherstinsky:
> I don't understand why we have an `add_dataframe_asset()` method and yet require a `dataframe` to be passed to `build_batch_request`.
> 
> This seems like an API change, not a bugfix.

@Kilo59 Indeed, it is an API change to fix a usability problem (so "BUGFIX" was the closest I could find -- as in: better fit than "MAINTENANCE" nor "FEATURE" for the kind of change that is being proposed).  Thanks.

great_expectations/datasource/fluent/pandas_datasource.py
Kilo59: Okay, so this is for when the DataContext is being reloaded and the original `dataframe` is no longer in memory.

But now we are breaking the workflow where the DataContext hasn't been reloaded. Seems like `dataframe` should be optional here, and we should do something like this...
```python
def build_batch_request(self, dataframe: pd.DataFrame | None) -> BatchRequest:
        df = dataframe or self.dataframe
        if not df:
            raise ValueError(
                "Cannot build batch request for dataframe asset without a dataframe"
            )
        self.dataframe = df
        return super().build_batch_request()
```

great_expectations/datasource/fluent/schemas/PandasAzureBlobStorageDatasource.json
Kilo59, line 1: Why are these schemas being deleted?

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -338,9 +338,7 @@ def _short_id() -> str:
 class DataFrameAsset(_PandasDataAsset, Generic[_PandasDataFrameT]):
     # instance attributes
     type: Literal["dataframe"] = "dataframe"
-    dataframe: Optional[_PandasDataFrameT] = pydantic.Field(
-        default=None, exclude=True, repr=False
-    )
+    dataframe: Optional[_PandasDataFrameT] = pydantic.Field(exclude=True, repr=False)
```
NathanFarmer: Why remove the default value?

great_expectations/datasource/fluent/pandas_datasource.py
NathanFarmer: I agree that the `dataframe` should be optional here. What if they used the `dataframe` in `add_dataframe` or assigned to the attribute directly? Why should they have to pass it twice?

great_expectations/datasource/fluent/pandas_datasource.py
NathanFarmer: It shouldn't be needed if `dataframe` is optional in `build_batch_request()`.

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -591,23 +594,28 @@ def _validate_asset_name(asset_name: Optional[str] = None) -> str:
...
+            batch_request = asset.build_batch_request()
+
         # TODO: raise error if `_data_context` not set
         return self._data_context.get_validator(batch_request=batch_request)  # type: ignore[union-attr] # self._data_context must be set
 
     @public_api
     def add_dataframe_asset(
         self,
         name: str,
-        dataframe: pd.DataFrame,
```
NathanFarmer, line 603: Like we stated offline, this should be deprecated, not removed.

great_expectations/datasource/fluent/schemas/PandasAzureBlobStorageDatasource/CSVAsset.json
```diff
@@ -1,515 +0,0 @@
-{
```
NathanFarmer: Why remove all of these schemas?

tests/checkpoint/test_checkpoint_result_format.py
```diff
@@ -3513,21 +3513,30 @@ def test_pandas_result_format_in_checkpoint_one_expectation_complete_output_flue
...
     context.add_expectation_suite(expectation_suite_name=expectation_suite_name)
 
-    context.sources.add_pandas(name="pandas_datasource").add_dataframe_asset(
+    data_frame_asset = context.sources.add_pandas(
+        name="pandas_datasource"
+    ).add_dataframe_asset(
         name="IN_MEMORY_DATA_ASSET",
-        dataframe=pandas_animals_dataframe_for_unexpected_rows_and_index,
+        batch_metadata={
+            "batch_slice": -1,
```
NathanFarmer: Why would a user add this? The `batch_metadata` is associated with the data asset, you don't know until you `build_batch_request()` which slice you are requesting.

tests/checkpoint/test_checkpoint_result_format.py
```diff
@@ -3513,21 +3513,30 @@ def test_pandas_result_format_in_checkpoint_one_expectation_complete_output_flue
...
+        },
+    )
+    runtime_batch_request = data_frame_asset.build_batch_request(
```
NathanFarmer: Why are you calling this a `runtime_batch_request`? It's just a `BatchRequest` like any other.

tests/checkpoint/test_checkpoint_result_format.py
```diff
@@ -3550,12 +3559,19 @@ def test_pandas_result_format_in_checkpoint_one_expectation_complete_output_flue
     result: CheckpointResult = context.run_checkpoint(
         checkpoint_name="my_checkpoint",
         expectation_suite_name=expectation_suite_name,
-    )
-
-    expected_batch_request = {
```
NathanFarmer, line 3555: Why does this assertion need to change?

tests/datasource/fluent/test_pandas_datasource.py
```diff
@@ -439,13 +439,17 @@ def test_default_pandas_datasource_name_conflict(
...
     # validates that a dataframe object is passed
-    with pytest.raises(pydantic.ValidationError) as exc_info:
+    with pytest.raises(ValueError) as exc_info:
         _ = empty_data_context.sources.pandas_default.read_dataframe(dataframe={})
 
-    errors_dict = exc_info.value.errors()[0]
-    assert errors_dict["loc"][0] == "dataframe"
+    assert (
+        "RuntimeDataBatchSpec must provide a Pandas DataFrame or PandasBatchData object"
+        in str(exc_info.value)
```
NathanFarmer: This error message is not meaningful to users. What is a `RuntimeDataBatchSpec`? Why doesn't this error just say that the `dataframe` parameter needs to be a `pd.DataFrame`?

tests/datasource/fluent/test_pandas_datasource.py
```diff
@@ -511,19 +516,3 @@ def test_pandas_data_asset_batch_metadata(
...
-    dataframe_asset = empty_data_context.sources.add_or_update_pandas(
-        name="fluent_pandas_datasource"
-    ).add_dataframe_asset(name="my_df_asset", dataframe=df)
-    dataframe_asset.dataframe = None
-    with pytest.raises(ValueError) as e:
-        dataframe_asset.build_batch_request()
-
-    assert "Cannot build batch request for dataframe asset without a dataframe" in str(
-        e.value
-    )
```
NathanFarmer: Why remove this test? An error should be raised if the `dataframe` is missing.

tests/datasource/fluent/test_spark_datasource.py
```diff
@@ -129,28 +130,3 @@ def test_spark_config_passed_to_execution_engine(
...
-        dataframe=spark_df,
-    )
-    dataframe_asset.dataframe = None
-
-    with pytest.raises(ValueError) as e:
-        dataframe_asset.build_batch_request()
-
-    assert "Cannot build batch request for dataframe asset without a dataframe" in str(
-        e.value
-    )
```
NathanFarmer: I also think this test should still apply.

NathanFarmer (changes requested):

NathanFarmer (changes requested):
I would also change the title of this PR. We can only deprecate the existing behavior.

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @Kilo59 Because when creating `DataFrameAsset` in the new way, we do not supply the `DataFrame`; hence, if there is a default value, we get a type violation in other APIs -- so `DataFrame` here is truly optional: as in, it can be omitted entirely.  Thanks.

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @NathanFarmer Got you -- 1) I will remove "isinstance" and 2) Will deprecate the API (to avoid any problems in the field if it is already use).  I will reach out for the right way to do that.  Thanks.

great_expectations/datasource/fluent/schemas/PandasAzureBlobStorageDatasource/CSVAsset.json
alexsherstinsky: @NathanFarmer It happened automatically by running this: `invoke schema --sync --clean ; invoke schema --sync`. /cc @Kilo59 -- thanks.

tests/checkpoint/test_checkpoint_result_format.py
alexsherstinsky: This was just fixing the broken test.  I will check (not sure right now).  Thanks.

tests/checkpoint/test_checkpoint_result_format.py
alexsherstinsky: @NathanFarmer To indicate that this has to do with a DataFrame as in a "running pipeline" -- was for extra clarification, but may not be necessary; I will change.  Thanks.

tests/checkpoint/test_checkpoint_result_format.py
alexsherstinsky: @NathanFarmer The previous assertion was not useful -- it was not sufficient.  The new one really tests the fact that the checkpoint runs and evaluates 1 expectation with the `success` result being affirmative.  Thanks.

tests/datasource/fluent/test_pandas_datasource.py
alexsherstinsky: @NathanFarmer Good point.  The error comes from the execution engine.  This means to me that it has to be caught earlier and changed to a better message.  I will come up with a way to do it.  Thanks for pointing it out.

tests/datasource/fluent/test_pandas_datasource.py
alexsherstinsky: @NathanFarmer This test is not useful any more.  This is because static type checking takes care of the situation.

tests/datasource/fluent/test_spark_datasource.py
alexsherstinsky: @NathanFarmer This test is not useful any more. This is because static type checking takes care of the situation.

alexsherstinsky:
> I would also change the title of this PR. We can only deprecate the existing behavior.

@NathanFarmer Can we not do both?  As in: deprecate existing and introduce new?  Please suggest a title that would fit this goal better.  Thanks.

great_expectations/datasource/fluent/schemas/PandasAzureBlobStorageDatasource.json
alexsherstinsky: @Kilo59 They got deleted by running `invoke schema --sync --clean ; invoke schema --sync` -- I was surprised myself.  Do you see a problem?  Thanks.

great_expectations/datasource/fluent/schemas/PandasAzureBlobStorageDatasource/CSVAsset.json
Kilo59: Please revert these changes and run the schema sync without the -clean flag

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @NathanFarmer Will fix (in next commit).  Thanks.

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @Kilo59 @NathanFarmer The point is that having `dataframe` optional is confusing.  The goal is to make exactly one location where the `dataframe` is needed.  This opinionated way will reduce cognitive overload, especially for first-time users.  Thus, no `dataframe` in `add_dataframe_asset()` -- only metadata; AND: must include `dataframe` in `DataFrameAsset.build_batch_request()`.  This clarity is the essence of the change.  Happy to discuss.  Thanks.

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @NathanFarmer It turns out that `isinstance()` is needed.  Moreover, for a better error message when the `dataframe` is not supplied correctly, another `isinstance()` was needed.  I committed my changes thus far.  Thanks.

great_expectations/datasource/fluent/pandas_datasource.py
Kilo59: You have to provide a default value if you want it to be optional.

great_expectations/datasource/fluent/pandas_datasource.py
Kilo59: Still, if we are doing a deprecation before removing the `dataframe` arg to `add_dataframe_asset()` then doesn't it need to be optional here as well ( until we've removed it   from `add_dataframe_asset()`)

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @Kilo59 If that is the case, how come it works right now (i.e., without the default value)?  In fact, with the default value, I had run into type checking problems in other methods.  By the way, to me, `Optional` really does mean that the argument can be omitted (as opposed to `Union` of a type and `None`).  Please advise for the best path forward.  Thanks.

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @Kilo59 I now see what you mean -- you are absolutely correct.  Thanks! /cc @NathanFarmer (and thanks to @anthonyburdi as the way we handle deprecated arguments informs leaving `dataframe` as optional).  This is regarding the `add_dataframe_asset()` method in the stub file (`pandas_datasources.pyi`).

alexsherstinsky: This has been committed.

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: The fix has been committed -- along with `depracted_api` and `new_api` decorators with messages.

great_expectations/datasource/fluent/schemas/PandasAzureBlobStorageDatasource/CSVAsset.json
alexsherstinsky: @Kilo59 I do not understand -- if the command I ran performs its function and exits without errors, why would we not accept the results?  What if these old `JSON` files are in error?  Alternatively, what command will bring them back automatically?  I will try `invoke schema --sync` without "clean" and let you know.  Thanks.

alexsherstinsky: @Kilo59 Running `invoke schema --sync` without "clean" did not bring back those JSON files.  What does this mean, and what should I do?  Thank you.

tests/datasource/fluent/test_pandas_datasource.py
alexsherstinsky: @NathanFarmer This has been fixed and committed.  Thanks.

great_expectations/datasource/fluent/pandas_datasource.py
Kilo59: @alexsherstinsky I think you still need to add the logic I proposed otherwise the function does not work as advertised.

great_expectations/datasource/fluent/schemas/PandasAzureBlobStorageDatasource/CSVAsset.json
Kilo59: You need to revert the changes. The pandas schemas have gone out of sync since we dropped python 3.7.
I'm working on a fix for that right now.

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -362,12 +367,25 @@ def _get_reader_options_include(self) -> set[str]:
         )
 
     @public_api
-    def build_batch_request(self) -> BatchRequest:  # type: ignore[override]
-        if self.dataframe is None:
-            raise ValueError(
-                "Cannot build batch request for dataframe asset without a dataframe"
-            )
```
NathanFarmer: Why aren't we raising this error anymore?

tests/checkpoint/test_checkpoint_result_format.py
NathanFarmer: It sure looks like you added this line. It is confusing.

tests/checkpoint/test_checkpoint_result_format.py
NathanFarmer: The assertion was testing that passing a fluent batch request dict results in a fluent batch request in the checkpoint result. Without this we could have a regression where the fluent dict becomes a block config dict.

tests/datasource/fluent/test_pandas_datasource.py
NathanFarmer: This test is meant for the API. I don't understand what type checking has to do with it. The user needs to see this message if they don't pass in a dataframe.

tests/datasource/fluent/test_spark_datasource.py
NathanFarmer: It's not useful for the user to know why they cannot build a batch request?

NathanFarmer (changes requested):

tests/datasource/fluent/integration/test_integration_datasource.py
```diff
@@ -515,35 +517,84 @@ def test_pandas_data_adding_dataframe_in_cloud_context(
...
+def test_pandas_data_adding_dataframe_in_file_reloaded_context(
+    empty_file_context: FileDataContext,
+):
+    df = pd.DataFrame({"column_name": [1, 2, 3, 4, 5]})
+
+    context = empty_file_context
+
+    datasource = context.sources.add_or_update_pandas(name="fluent_pandas_datasource")
+    dataframe_asset = datasource.add_dataframe_asset(name="my_df_asset")
+    context._save_project_config()
```
Kilo59: `non-blocking`
`._save_project_config()` should automatically be called now after the asset is added. So you shouldn't need this in the test.

Kilo59 (approved):

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -616,9 +658,9 @@ def add_dataframe_asset(
         """
         asset: DataFrameAsset = DataFrameAsset(
             name=name,
-            dataframe=dataframe,
             batch_metadata=batch_metadata or {},
         )
+        asset.dataframe = dataframe
```
Kilo59, line 667: Why the change to not provide `dataframe` in the constructor?

tests/datasource/fluent/integration/test_integration_datasource.py
alexsherstinsky: @Kilo59 I deleted it -- thanks.  Probably left over from trying to get the test to work, and thought that it was needed during one of the attempts, and then left it in there once things worked, without checking to see if it will work without that "save".  Thanks.

great_expectations/datasource/fluent/pandas_datasource.py
alexsherstinsky: @Kilo59 We are still doing it now (but deprecating).  The goal is to avoid any possible confusion for DataFrames -- ephemeral data references: the add_asset for ephemeral data references accepts only (serializable) metadata.  The only place to specify the (non-serializable) data reference is in `.read_dataframe()` (Pandas only for now) or `.build_batch_request()`.  This will finally make the API for ephemeral data precise and interpretable.  Thanks.

alexsherstinsky: @Kilo59 We are still doing it now (but deprecating).  The goal is to avoid any possible confusion for DataFrames -- ephemeral data references: the add_asset for ephemeral data references accepts only (serializable) metadata.  The only place to specify the (non-serializable) data reference is in `.read_dataframe()` (Pandas only for now) or `.build_batch_request()`.  This will finally make the API for ephemeral data precise and interpretable.  Thanks.

NathanFarmer (changes requested):
Please see my outstanding comments!

tests/datasource/fluent/test_pandas_datasource.py
alexsherstinsky: @NathanFarmer We do not accept `dataframe` any more in `add_dataframe_asset()`.  The type checking comment was that if we leave this test in, we see a PyCharm violation that the argument is unknown.  But the real reason for this test not being needed is the API change.  Thanks.

tests/datasource/fluent/test_spark_datasource.py
alexsherstinsky: @NathanFarmer We do not accept `dataframe` any more in `add_dataframe_asset()`.

tests/checkpoint/test_checkpoint_result_format.py
alexsherstinsky: @NathanFarmer I see what you mean.  You are absolutely correct.  I tried to include the `Batch` metadata in the `add_dataframe_asset()` method, because that is where the `batch_metadata` argument is allowed, but the specific option about the `slice` is confusing, because, like you said, the slicing is not known till later.  But we have a problem here -- because the Checkpoint specified an ephemeral DataAsset, and the new API change exposes the problem with it.  So we need to be able to specify `batch_metadata` as part of the `build_batch_request()`.  This can be done in the present PR or in the very next one.  Let us discuss the best way to proceed.  Thanks!

tests/checkpoint/test_checkpoint_result_format.py
alexsherstinsky: @NathanFarmer Let us discuss -- I believe that this test in the present form exposes problems previously glossed over.  Please see my previous comment -- let us discuss.  Thank you!

NathanFarmer (approved):

Issue #7628, opened 2023-04-14
Passing a column with space fails with validator object for expectation - 'expect_column_values_to_be_unique' 


PSmayank21:
## Issue

For the expectation 'expect_column_values_to_be_unique' - on trying to pass a column name with space (e.g. 'Order Number') to an expectation using validator object fails even if backticks are used.

## Steps to reproduce the behaviour:

1. Create a spark dataframe having a column name that contains a space. (e.g. 'Order number')
2. Create a validator object using the spark dataframe.
3. Run the expectation `expect_column_values_to_be_unique` using the validator object.

Code to reproducer error- 
```python
from pyspark.sql.types import StructType,StructField, StringType, IntegerType
from great_expectations.core.batch import RuntimeBatchRequest
from great_expectations.util import get_context
from great_expectations.data_context.types.base import (DataContextConfig,FilesystemStoreBackendDefaults,)
from great_expectations.dataset import SparkDFDataset

metastore_dir = "/dbfs/great_expectations_checks/"

data_context_config = DataContextConfig(store_backend_defaults=FilesystemStoreBackendDefaults(
                                        root_directory=metastore_dir
                                        ),
                                       )
context = get_context(project_config=data_context_config)
batch_identifiers = ["customer",]
my_spark_datasource_config = {
    "name": "test_dataset",
    "class_name": "Datasource",
    "execution_engine": {"class_name": "SparkDFExecutionEngine"},
    "data_connectors": {
        "runtimedataconnector_test_dataset": {
            "module_name": "great_expectations.datasource.data_connector",
            "class_name": "RuntimeDataConnector",
            "batch_identifiers": batch_identifiers,
        }
    },
}

context.add_datasource(**my_spark_datasource_config)

data2 = [("James",100),
    ("Michael",101),
    ("Robert",102),
    ("Maria",103),
    ("Jen",104)
  ]

schema = StructType([ \
    StructField("customer",StringType(),True), \
    StructField("order number", IntegerType(), True) \
  ])
 
df = spark.createDataFrame(data=data2,schema=schema)

batch_request = RuntimeBatchRequest(
    datasource_name="test_dataset",
    data_connector_name="runtimedataconnector_test_dataset",
    data_asset_name="test_dataset", 
    batch_identifiers={
        "customer": "test",
    },
    runtime_parameters={"batch_data": df},  # Your dataframe goes here
    )
  
[...]

PR #8038, opened 2023-05-31
[FEATURE] DX 524 column with space fails expect values to be unique

TrangPham:
Addresses: https://github.com/great-expectations/great_expectations/issues/7628

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```

### DIFF
--- a/great_expectations/expectations/metrics/util.py
+++ b/great_expectations/expectations/metrics/util.py
@@ -703,6 +703,23 @@ def _verify_column_names_exist_and_get_normalized_typed_column_names_map(
             ) or (column_name == str(typed_column_name_cursor)):
                 return column_name, typed_column_name_cursor
 
+            # use explicit identifier if passed in by user
+            if (type(typed_column_name_cursor) == str) and (
+                (
+                    column_name.casefold().strip('"')
+                    == typed_column_name_cursor.casefold()
+                )
+                or (
+                    column_name.casefold().strip("[]")
+                    == typed_column_name_cursor.casefold()
+                )
+                or (
+                    column_name.casefold().strip("`")
+                    == typed_column_name_cursor.casefold()
+                )
+            ):
+                return column_name, column_name
+
         return None
 
     normalized_batch_columns_mappings: List[
--- a/tests/execution_engine/test_sparkdf_execution_engine.py
+++ b/tests/execution_engine/test_sparkdf_execution_engine.py
@@ -9,15 +9,21 @@ import pytest
 import great_expectations.exceptions as gx_exceptions
 from great_expectations.compatibility import pyspark
 from great_expectations.compatibility.pyspark import functions as F
+from great_expectations.core.batch import RuntimeBatchRequest
 from great_expectations.core.batch_spec import PathBatchSpec, RuntimeDataBatchSpec
 from great_expectations.core.metric_domain_types import MetricDomainTypes
 from great_expectations.core.metric_function_types import MetricPartialFunctionTypes
+from great_expectations.data_context.types.base import (
+    DataContextConfig,
+    FilesystemStoreBackendDefaults,
+)
 from great_expectations.execution_engine import SparkDFExecutionEngine
 from great_expectations.expectations.row_conditions import (
     RowCondition,
     RowConditionParserType,
 )
 from great_expectations.self_check.util import build_spark_engine
+from great_expectations.util import get_context
 from great_expectations.validator.computed_metric import MetricValue
 from great_expectations.validator.metric_configuration import MetricConfiguration
 from tests.expectations.test_util import get_table_columns_metric
@@ -1192,3 +1198,90 @@ def test_schema_properly_added(spark_session):
     )
     df = engine.dataframe
     assert df.schema == schema
+
+
+@pytest.mark.integration
+def test_explicit_string_identifiers_should_work_with_validator(spark_session):
+    """
+    Integration test taken from: https://github.com/great-expectations/great_expectations/issues/7628
+    Ensures that the expectations can be used with a string identifier, resolves to the correct column
+    in SQL and a result is returned.
+    """
+    metastore_dir = "/tmp/great_expectations_v3"
+
+    data_context_config = DataContextConfig(
+        store_backend_defaults=FilesystemStoreBackendDefaults(
+            root_directory=metastore_dir
+        ),
+    )
+    context = get_context(project_config=data_context_config)
+    batch_identifiers = [
+        "customer",
+    ]
+    my_spark_datasource_config = {
+        "name": "test_dataset",
+        "class_name": "Datasource",
+        "execution_engine": {"class_name": "SparkDFExecutionEngine"},
+        "data_connectors": {
+            "runtimedataconnector_test_dataset": {
+                "module_name": "great_expectations.datasource.data_connector",
+                "class_name": "RuntimeDataConnector",
+                "batch_identifiers": batch_identifiers,
+            }
+        },
+    }
+
+    context.add_datasource(**my_spark_datasource_config)
+
+    data2 = [
+        ("James", 100),
+        ("Michael", 101),
+        ("Robert", 102),
+        ("Maria", 103),
+        ("Jen", 104),
+    ]
+
+    schema = pyspark.types.StructType(
+        [
+            pyspark.types.StructField("customer", pyspark.types.StringType(), True),
+            pyspark.types.StructField(
+                "order number", pyspark.types.IntegerType(), True
+            ),
+        ]
+    )
+    df = spark_session.createDataFrame(data=data2, schema=schema)
+
+    batch_request = RuntimeBatchRequest(
+        datasource_name="test_dataset",
+        data_connector_name="runtimedataconnector_test_dataset",
+        data_asset_name="test_dataset",
+        batch_identifiers={
+            "customer": "test",
+        },
+        runtime_parameters={"batch_data": df},  # Your dataframe goes here
+    )
+
+    context.add_or_update_expectation_suite(
+        expectation_suite_name="test_ge_unique_record"
+    )
+
+    validator = context.get_validator(
+        batch_request=batch_request,
+        expectation_suite_name="test_ge_unique_record",
+    )
+
+    col_name = "`order number`"
+    result = validator.expect_column_values_to_be_unique(column=col_name)
+
+    assert result["success"]
+    assert result["expectation_config"]["kwargs"]["column"] == col_name
+    assert result["result"] == {
+        "element_count": 5,
+        "unexpected_count": 0,
+        "unexpected_percent": 0.0,
+        "partial_unexpected_list": [],
+        "missing_count": 0,
+        "missing_percent": 0.0,
+        "unexpected_percent_total": 0.0,
+        "unexpected_percent_nonmissing": 0.0,
+    }
--- a/tests/expectations/metrics/test_core.py
+++ b/tests/expectations/metrics/test_core.py
@@ -337,6 +337,7 @@ def test_column_quoted_name_type_sa(sa):
             == f'Error: The column "{column_name}" in BatchData does not exist.'
         )
 
+
 @pytest.mark.unit
 def test_column_quoted_name_type_sa_handles_explicit_string_identifiers(sa):
     """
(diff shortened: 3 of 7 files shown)


great_expectations/expectations/metrics/util.py
```diff
@@ -703,6 +703,23 @@ def _get_normalized_column_name_mapping_if_exists(
...
+                    column_name.casefold().strip("[]")
+                    == typed_column_name_cursor.casefold()
+                )
+                or (
+                    column_name.casefold().strip("`")
+                    == typed_column_name_cursor.casefold()
+                )
+            ):
+                return column_name, column_name
+
```
TrangPham, line 722: If there's a better way to get which identifiers were used in a given dialect here, we can update this code to use that.

I also was not sure about the quoted_name values and how exactly that mechanism works.

great_expectations/expectations/metrics/util.py
```diff
@@ -703,6 +703,23 @@ def _get_normalized_column_name_mapping_if_exists(
...
 
+            # use explicit identifier if passed in by user
+            if (type(typed_column_name_cursor) == str) and (
+                (
+                    column_name.casefold().strip('"')
+                    == typed_column_name_cursor.casefold()
+                )
+                or (
+                    column_name.casefold().strip("[]")
+                    == typed_column_name_cursor.casefold()
```
Shinnnyshinshin, line 714: Today I learned about using the `casefold()` method over `lower()`. Thank you @TrangPham

tests/expectations/metrics/test_core.py
```diff
@@ -333,6 +338,66 @@ def test_column_quoted_name_type_sa(sa):
         )
 
 
+def test_column_quoted_name_type_sa_handles_explicit_string_identifiers(sa):
+    """
+    Within SQLite, identifiers can be quoted using one of the following mechanisms:
```
Shinnnyshinshin, line 344: :heart:

tests/expectations/metrics/test_core.py
```diff
@@ -333,6 +338,66 @@ def test_column_quoted_name_type_sa(sa):
...
+
+    table_columns_metric: MetricConfiguration = MetricConfiguration(
+        metric_name="table.columns",
+        metric_domain_kwargs={},
+        metric_value_kwargs=None,
+    )
+    table_columns_metric_id: Tuple[str, str, str] = table_columns_metric.id
+    batch_column_list = metrics[table_columns_metric_id]
+
+    for column_name in [
```
Shinnnyshinshin, line 389: Great test. thank you

tests/expectations/metrics/test_core.py
```diff
@@ -333,6 +338,66 @@ def test_column_quoted_name_type_sa(sa):
         )
```
Shinnnyshinshin, line 401: Would it be worth marking this with `@pytest.mark.unit`?

tests/expectations/metrics/test_core.py
```diff
@@ -6964,3 +7029,84 @@ def test_map_select_column_values_unique_within_record_spark(spark_session):
         {"a": 1.0, "b": 1.0, "c": 2.0},
         {"a": 4.0, "b": 4.0, "c": 4.0},
     ]
+
+
+def test_dx_524_explicit_string_identifiers_should_work_with_validator(spark_session):
```
Shinnnyshinshin: I dont think we use this convention (dx-524) for naming tests, but I see the benefit. Another option could be to include it in a docstring, but I think either way could work.

tests/expectations/metrics/test_core.py
```diff
@@ -6964,3 +7029,84 @@ def test_map_select_column_values_unique_within_record_spark(spark_session):
         {"a": 1.0, "b": 1.0, "c": 2.0},
         {"a": 4.0, "b": 4.0, "c": 4.0},
     ]
+
+
```
Shinnnyshinshin: Would it be worth marking this with `@pytest.mark.integration`?

tests/expectations/metrics/test_core.py
TrangPham: Sounds good - I can move it to a docstring or a comment

alexsherstinsky: How would the external reader be able to interpret "dx_524"?

TrangPham: Ah true! I will mention the external github issue instead

great_expectations/expectations/metrics/util.py
alexsherstinsky: @TrangPham Why would we be looking for `"` and/or "[]"?  Are these delimiters in the specifications?  Interesting problem, by the way.  I know that we are dealing with Spark here, but it is also quite possible that the use of `casefold()` for SQLAlchemy is not a perfect solution either (I heard from a user that they still had problems with Snowflake, but they did not provide an example; other problems reported with Snowflake were fixed as far as I can tell).  Thanks.

tests/expectations/metrics/test_core.py
```diff
@@ -7,9 +7,13 @@
...
 import pandas as pd
 import pytest
+from great_expectations.data_context.types.base import (
+    DataContextConfig,
+    FilesystemStoreBackendDefaults,
+)
 
 import great_expectations.exceptions as gx_exceptions
-from great_expectations.core.batch import Batch
+from great_expectations.core.batch import Batch, RuntimeBatchRequest
```
alexsherstinsky, line 16: @TrangPham @Shinnnyshinshin Here are some thoughts: First, I feel that we should be reproducing this error using Fluent Datasources (FDS) and debugging it with FDS as well.  There is no benefit in supporting `RuntimeBatchRequest` if we do not also solve the issue for the FDS based approach as well.  Second, I believe that tests with `BatchRequest` should not be in the `test_core.py` module on the account of the fact that this module is for the lower level -- metrics -- scenarios.  I am happy to brainstorm both of these points.  Thank you.

tests/expectations/metrics/test_core.py
```diff
@@ -294,8 +299,8 @@ def test_column_quoted_name_type_sa(sa):
 
     for column_name in [
         "non_existent_column",
-        '"NAMES"',
-        '"Names"',
+        "?NAMES?",
```
alexsherstinsky, line 302: @TrangPham I am not familiar with the origin of these markers -- were the question mark and the asterisk mentioned in a specification document (or a ticket)?  Please let me know what use case we are aiming for.  Thank you!

tests/expectations/metrics/test_core.py
```diff
@@ -333,6 +338,66 @@ def test_column_quoted_name_type_sa(sa):
         )
 
 
+def test_column_quoted_name_type_sa_handles_explicit_string_identifiers(sa):
+    """
+    Within SQLite, identifiers can be quoted using one of the following mechanisms:
+    'keyword'		A keyword in single quotes is a string literal.
+    "keyword"		A keyword in double-quotes is an identifier.
```
alexsherstinsky, line 346: @TrangPham : If this is for SQL, then perhaps we need to reproduce how a user with a command-line SQL editor would handle the column names with the spaces in them.  Could we find a definitive recommendation from, say, ANSI SQL documentation?  Thank you.

tests/expectations/metrics/test_core.py
```diff
@@ -6964,3 +7029,84 @@ def test_map_select_column_values_unique_within_record_spark(spark_session):
...
+        [
+            pyspark.types.StructField("customer", pyspark.types.StringType(), True),
+            pyspark.types.StructField(
+                "order number", pyspark.types.IntegerType(), True
+            ),
+        ]
+    )
+    df = spark_session.createDataFrame(data=data2, schema=schema)
+
+    batch_request = RuntimeBatchRequest(
```
alexsherstinsky: @TrangPham @Shinnnyshinshin This looks like an integration test (the present module is only for low-level metrics tests).  Thank you.  P.S.: Also, please see my thoughts regarding ensuring that the FDS case works.  Thanks!

alexsherstinsky (changes requested):
@TrangPham I offered a few questions, thoughts, and suggestions.  Thank you!

great_expectations/expectations/metrics/util.py
TrangPham: I would prefer to be able to deduce the delimiters from the dialect if possible. I was playing around with SQLalchemy and quoted_name but could not figure out an elegant solution. The code is passed through many partials as well which makes it hard to understand where the engine should come from. 

I wasn't sure if using quoted_name to standardize to would result in a correctly delimited identifier.

```
def _verify_column_names_exist_and_get_normalized_typed_column_names_map(
    column_names: List[str] | str,
    batch_columns_list: List[str | sqlalchemy.quoted_name],
    error_message_template: str = 'Error: The column "{column_name:s}" in BatchData does not exist.',
    verify_only: bool = False,
) -> List[Tuple[str, str | sqlalchemy.quoted_name]] | None:
    """
    Insures that column name or column names (supplied as argument using "str" representation) exist in "Batch" object.

    Args:
        column_names: Single string-valued column name or list of string-valued column names
        batch_columns_list: Properly typed column names (output of "table.columns" metric)
        verify_only: Perform verification only (do not return normalized typed column names)
        error_message_template: String template to output error message if any column cannot be found in "Batch" object.

    Returns:
        List of tuples having mapping from string-valued column name to typed column name; None if "verify_only" is set.
    """
```

Would calling this function with `batch_columns_list` as a list of quoted_name result in the correct raw SQL for the dialect be used?

tests/expectations/metrics/test_core.py
TrangPham: Unfortunately I haven't seen that the ANSI is used in most places.

great_expectations/expectations/metrics/util.py
alexsherstinsky: @TrangPham The `quoted_name` is solely for `SQLAlchemy` based backends.  The others were, till now, just strings.  We had also grappled with columns whose names would have their first letter capitalized (that was solved).  However, unless I am forgetting, the space in the column name seems to be first time.  Maybe ANSI SQL and other standards documentation (for Pandas and Spark) can help us?  Thanks!

tests/expectations/metrics/test_core.py
alexsherstinsky: @TrangPham Then I wonder if this should be a reason for us to disallow columns/tables with spaces in their names?  If we do not have anything definitive for ANSI SQL, is there anything for Snowflake?  MS SQL Server?  Oracle? Etc. -- the dialects, like you said.  Also, does Spark API mention these cases?  It seems very painful to handle!

tests/expectations/metrics/test_core.py
TrangPham: Where should we move the test? 

Do you have an example of similar test in FDS?

great_expectations/expectations/metrics/util.py
TrangPham: @alexsherstinsky @Shinnnyshinshin 

I played around with this more

I'm not sure how quoted_name functionality works/would help
Screenshot 2023-06-01 at 5 11 34 PM

But I think we could use `Column` class to force quoting on all columns and normalize to that
Screenshot 2023-06-01 at 4 59 45 PM

With current code line 158 would fail comparing "`other names" with "other names".
If "other names" is passed in, line 158 would not fail but the raw SQL would be incorrect.

I like the idea of either:
1. forcing user to use explicit identifiers if they want it
2. OR allowing user to pass in any string and always forcing quotes for them

tests/expectations/metrics/test_core.py
```diff
@@ -34,7 +39,7 @@
     build_sa_engine,
     build_spark_engine,
 )
-from great_expectations.util import isclose
+from great_expectations.util import get_context, isclose
```
alexsherstinsky, line 42: @TrangPham The context import is indicative of a potential integration style test (this module is for unit tests).  Thank you.

tests/expectations/metrics/test_core.py
alexsherstinsky: @TrangPham My initial thought is that we can probably avoid this integration test if we create unit tests (for every `ExecutionEngine` where the column names having the space character in them can occur, or the capitalization patterns, etc.) such that the data is set up in the way that previously gave us difficulties (and with the fix should not), and then we just run a column-`Domain` metric, such as the one on which `expect_column_values_to_be_unique` relies, and/or perhaps a few others for extra confidence, and assert that there are no errors and that the value of the computed metric is what we expect.  Happy to discuss -- thank you!

tests/expectations/metrics/test_core.py
TrangPham: Oh I was using these as counter examples as in these should not be valid. 

I just had an idea though. We could take in an optional parameter from the user which is `identifier_delimiter` and then check column names and add that for the user to col string.

Usage would then look like: 
``` 
    col_name = "other name"
    result = validator.expect_column_values_to_be_unique(column=col_name, identifier_delimiter="`")
```

great_expectations/expectations/metrics/util.py
TrangPham: Hmmm if we have backends which use string here, then using dialect from the SQLAlchemy engine would not work. Neither would using `Column` from SQLAlchemy. What are some examples of other backends?

tests/expectations/metrics/test_core.py
TrangPham: I moved the test to test_sparkdf_execution_engine

tests/expectations/metrics/test_core.py
TrangPham: As discussed in slack - we should handle column names that require delimited identifiers. I updated the code to specifically handle this for SparkDF instead of using a generic set of delimiters.

TrangPham:
I have updated the solution to be specific to SparkDF. Now if the spark would require the user to delimit their identifiers, then we would recognize the delimited identifier as a correct column.

great_expectations/execution_engine/execution_engine.py
```diff
@@ -112,20 +112,20 @@ class ExecutionEngine(ABC):
...
 
     Crucially, ExecutionEngine serves as focal point for resolving (i.e., computing) metrics.  Wherever opportunities
-    arize to bundle multiple metric computations (e.g., SQLAlchemy, Spark), ExecutionEngine utilizes subclasses in order
+    arise to bundle multiple metric computations (e.g., SQLAlchemy, Spark), ExecutionEngine utilizes subclasses in order
     to provide specific functionality (bundling of computation is available only for "deferred execution" computational
     systems, such as SQLAlchemy and Spark; it is not available for Pandas, because Pandas computations are immediate).
 
     Finally, ExecutionEngine defines interfaces for Batch data sampling and splitting Batch of data along defined axes.
 
-    Constructor builds an ExecutionEngine, using provided configuration options (instatiation is done by child classes).
```
Shinnnyshinshin: 😅

tests/execution_engine/test_sparkdf_execution_engine.py
```diff
@@ -1192,3 +1198,90 @@ def test_schema_properly_added(spark_session):
     )
     df = engine.dataframe
     assert df.schema == schema
+
+
+@pytest.mark.integration
+def test_using_explicit_string_identifiers_should_work_sparkdf(spark_session):
+    """
+    Integration test taken from: https://github.com/great-expectations/great_expectations/issues/7628
```
Shinnnyshinshin, line 1206: ❤️ love the addition of the link

Shinnnyshinshin (approved):
LGTM!

PR #8076, opened 2023-06-07
[MAINTENANCE] Misc/docs integration uncommented

kenwade4:
- Remove comments and test skipping added in #7735
- Fix how_to_create_and_edit_expectations_with_instant_feedback_block_config.py
- Fix bug in _get_regex_config
- Use add_or_update_expectation_suite over deprecated create_expectation_suite in non-cloud_data_context tests
- For ConfiguredAssetGCSDataConnector in tests, specify gcs_options delimiter
- Update ConfiguredAssetGCSDataConnector.__init__ to pop delimiter from gcs_options
- Update load_data_into_test_database to check if mysql should try to drop existing table or not
- Don't pass default_flow_style arg to yaml.dump
- Blob.download_as_string() is deprecated, so use download_as_bytes for gcs/bigquery
- Re-comment out the remaining 6 IntegrationTestFixture objects with failures


great_expectations/data_context/store/tuple_store_backend.py
```diff
@@ -852,7 +852,7 @@ def _get(self, key):
                 f"Unable to retrieve object from TupleGCSStoreBackend with the following Key: {str(key)}"
             )
         else:
-            return gcs_response_object.download_as_string().decode("utf-8")
+            return gcs_response_object.download_as_bytes().decode("utf-8")
```
anthonyburdi, line 855: :rocket:

great_expectations/datasource/data_connector/configured_asset_file_path_data_connector.py
```diff
@@ -184,8 +184,10 @@ def _get_regex_config(self, data_asset_name: Optional[str] = None) -> dict:
             # Override the defaults
             if asset.pattern:
                 regex_config["pattern"] = asset.pattern
-            if asset.group_names:
-                regex_config["group_names"] = asset.group_names
+                if asset.group_names:
+                    regex_config["group_names"] = asset.group_names
+                elif not regex_config.get("group_names"):
+                    regex_config["group_names"] = []
```
anthonyburdi, line 191: Just checking - should this be indented under the `if asset.pattern:` or unindented?

great_expectations/datasource/data_connector/configured_asset_gcs_data_connector.py
```diff
@@ -72,14 +72,17 @@ def __init__(  # noqa: PLR0913
             sorters=sorters,
             batch_spec_passthrough=batch_spec_passthrough,
         )
+        if gcs_options is None:
+            gcs_options = {}
+
+        if "delimiter" in gcs_options:
+            delimiter = gcs_options.pop("delimiter")
```
anthonyburdi: Since this overrides the value passed for the `delimiter` param, what do you think about updating the docstring to add something like "overridden if provided via gcs_options" to the `delimiter` param?

Also I'm wondering if instead of using the untyped `gcs_options` dictionary we could use the `delimiter` param, or is there a need to use `gcs_options` instead? I haven't been able to find an example in our docs.

tests/integration/docusaurus/connecting_to_your_data/cloud/gcs/pandas/configured_python_example.py
```diff
@@ -17,6 +17,9 @@
             "class_name": "ConfiguredAssetGCSDataConnector",
             "bucket_or_name": "<YOUR_GCS_BUCKET_HERE>",
             "prefix": "<BUCKET_PATH_TO_DATA>",
+            "gcs_options": {
+                "delimiter": "/",
+            },
```
anthonyburdi: How about just passing the delimiter directly via the delimiter param?

```suggestion
            "delimiter": "/",
```

tests/integration/docusaurus/connecting_to_your_data/cloud/gcs/pandas/configured_yaml_example.py
```diff
@@ -18,6 +18,8 @@
         class_name: ConfiguredAssetGCSDataConnector
         bucket_or_name: <YOUR_GCS_BUCKET_HERE>
         prefix: <BUCKET_PATH_TO_DATA>
+        gcs_options:
+            delimiter: /
```
anthonyburdi: How about just passing the delimiter directly via the delimiter param?

```suggestion
        delimiter: /
```

anthonyburdi (commented):
Thank you for getting our pipeline back up and running! Just a few comments and suggestions.

great_expectations/datasource/data_connector/configured_asset_file_path_data_connector.py
kenwade4: Yea this is in the correct spot under `if asset.pattern`. It's possible for there to be a "pattern" that is actually just a filename and it has no groups in the "regex"... Something somewhere else will try to access the group_names key and it will throw a KeyError when it doesn't exist, so need to set it to an empty list if there wasn't already some default regex in the config.

anthonyburdi (approved):
Thanks! LGTM!

PR #8084, opened 2023-06-08
[BUGFIX] Fix Update Checkpoint for Cloud

roblim:
Update Checkpoint operations (`add`, `update`, `add_or_update`) were broken for cloud in two ways:
1. `GXCloudStoreBackend` did not support updating Checkpoints by name. If `add_or_update` was called once, a new Checkpoint would be added, and the returned Checkpoint (in memory) would be properly updated with the new `id`. If the same method was called again with the same in-memory objects, the update operation would work, but if the same script was run at a later time, the operation would fail due to absence of `id` (GX would attempt adding a new Checkpoint with the same name, which violates unique constraint).
2. Since update Checkpoint operations use a PUT request, no object is returned with the response, which prevented GX from returning complete/updated Checkpoint configs (since cloud may add default values and/or assign new ids to Validations).
3. Similar errors would also occur in add operations if the returned Checkpoint object was not properly updated.

These bugs were fixed by:
1. Adding update by name support to `GXCloudStoreBackend`. The update flow was already checking for record existence by attempting to GET a resource, but the returned object was not being used. Instead, if the object exists, the update key is updated with the appropriate id if missing. 
2. Upon successful completion of an Update operation, the Checkpoint is re-fetched and returned from Cloud.
3. Since add operations use a POST request, which returns a complete Checkpoint config, logic was updated to simply return this Checkpoint config, rather than manually updating attributes on the existing in-memory Checkpoint.


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1916,7 +1915,16 @@ def update_checkpoint(self, checkpoint: Checkpoint) -> Checkpoint:
...
         """
-        return self.checkpoint_store.update_checkpoint(checkpoint)
+        result: Checkpoint | CheckpointConfig = self.checkpoint_store.update_checkpoint(
+            checkpoint
+        )
+        if isinstance(result, CheckpointConfig):
+            result = Checkpoint.instantiate_from_config_with_runtime_args(
+                checkpoint_config=result,
+                data_context=self,
+                name=Checkpoint.name,
```
cdkini: Should this be `Checkpoint`? The class itself? Or should it be `result.name`?

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -759,16 +760,25 @@ def add_checkpoint(  # noqa: PLR0913
             checkpoint=checkpoint,
         )
 
+        result: Checkpoint | CheckpointConfig
```
cdkini, line 763: Should we maybe always return a `CheckpointConfig`? Feels a little odd that we have this union (but maybe a matter for another time)

great_expectations/data_context/data_context/abstract_data_context.py
roblim: Great catch! Fixing

roblim: Addressed in https://github.com/great-expectations/great_expectations/pull/8084/commits/d21a474ecc16d15000eefc368c05fbee79ec5d63

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -227,18 +231,20 @@ def _persist_checkpoint(
...
-                "attributes"
-            ]["checkpoint_config"].get("validations")
-            checkpoint.config.action_list = checkpoint_ref.response["data"][
-                "attributes"
-            ]["checkpoint_config"].get("action_list")
+            # return CheckpointConfig from cloud POST response to account for any defaults/new ids added in cloud
+            checkpoint_config = checkpoint_ref.response["data"]["attributes"][
+                "checkpoint_config"
+            ]
+            checkpoint_config["ge_cloud_id"] = checkpoint_config.pop("id")
```
cdkini, line 241: Are we guaranteed to have "id" in `checkpoint_config`? Will this `pop()` possible cause a runtime issue?

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -542,6 +542,25 @@ def remove_key(self, key):
                 f"Unable to delete object in GX Cloud Store Backend: {repr(e)}"
             )
 
+    def _update(self, key, value, **kwargs):
+        existing = self._get(key)
+        if key[1] is None:
```
cdkini, line 547: I think we opt for truthy/falsy in OSS: `if not key[1]`

Good either way but we should be consistent if possible

cdkini: Same note below

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -542,6 +542,25 @@ def remove_key(self, key):
...
+
+    def _add_or_update(self, key, value, **kwargs):
+        try:
+            existing = self._get(key)
+        except StoreBackendError as e:
+            logger.info(f"Could not find object associated with key {key}: {e}")
+            existing = None
+        if existing is not None:
+            id = key[1] if key[1] is not None else existing["data"]["id"]
+            key = (key[0], id, key[2])
```
cdkini, line 560: I feel like this logic is pretty difficult to follow if you don't understand stores in depth (not a fault of this PR but more of an architectural thing) - hmm is there any way we can make this a bit clearer?

tests/data_context/cloud_data_context/test_checkpoint_crud.py
```diff
@@ -66,7 +68,7 @@ def _mocked_post_response(*args, **kwargs):
                 "data": {
                     "id": checkpoint_id,
                     "attributes": {
-                        "checkpoint_config": checkpoint_config_with_ids,
+                        "checkpoint_config": copy.deepcopy(checkpoint_config_with_ids),
```
cdkini, line 71: Any reason this is necessary?

great_expectations/data_context/data_context/cloud_data_context.py
Kilo59: Yeah, adding new types that could be returned from functions feels like we are going in the wrong direction.
The return annotation says `-> Checkpoint` so I would expect to see a mypy error here.

@roblim is this needed for the bugfix?

great_expectations/data_context/store/gx_cloud_store_backend.py
Kilo59, line 560:
```suggestion
        if existing:
            id = key[1] if key[1] else existing["data"]["id"]
            key = (key[0], id, key[2])
```

tests/data_context/cloud_data_context/test_checkpoint_crud.py
Kilo59: Looks like it's because this gets modified

Kilo59 (approved):

great_expectations/data_context/store/checkpoint_store.py
roblim: Yeah, this should always be returned by cloud.

great_expectations/data_context/store/gx_cloud_store_backend.py
roblim: Gotcha. Will note for future OSS work.

tests/data_context/cloud_data_context/test_checkpoint_crud.py
roblim: Yeah, since this fixture was getting modified, was running into weird behavior.

great_expectations/data_context/data_context/cloud_data_context.py
roblim: Had synchronous review with Gabe. Gabe was fine with change, since the type narrows by the return, so the context methods are only ever returning Checkpoint type.

great_expectations/data_context/store/gx_cloud_store_backend.py
roblim: Will add commentary in separate PR.

PR #8094, opened 2023-06-10
[FEATURE] Support DataFrame Argument in AbstractDataContext.get_validator method

alexsherstinsky:
### Scope
This change supports `dataframe` argument in `AbstractDataContext.get_validator()`.  Together with passing `validator` (instead of `batch_request`) to `Checkpoint.run()`, this provides support for ephemeral data in `Validator` and `Checkpoint` usage scenarios.

The coding pattern enabled is as follows:

```
batch_request = FluentBatchRequest(
        datasource_name='my_datasource_name',
        data_asset_name='my_asset_name',
)
```

Then:

```
validator = context.get_validator(
        expectation_suite_name="my_suite",
        batch_request=batch_request,
        dataframe=my_pandas_df,  # or my_spark_df
)
```
This then enables commands, such as `validator.expect_*`, to use the `DataFrame` (and the underlying `DataFrameAsset` as the `Batch` of data.

Finally:

```
checkpoint_result = checkpoint.run(validator=validator)
```

Where the `validator` is passed instead of `batch_request`, thereby supplying the ephemeral DataAsset to Checkpoint.

### Remarks
* If the general approach makes sense, then we can strengthen the type checking by passing not `"dataframe=my_pandas_df,  # or my_spark_df"` to `context.get_validator()`, but extending `FluentBatchRequest`, which is serializable, to the special Ephemeral subclass, which would not be serializable, but contain a `DataFrame` (Pandas or Spark -- or we can have two separate extensions, one for `Pandas`, the other for `Spark` -- to be decided).  This way, `context.get_validator()` will still accept a formal `BatchRequest` type, without the separate `dataframe` argument.
* JIRA: DX-469/DX-553

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```

### DIFF
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -2419,7 +2419,7 @@ class AbstractDataContext(ConfigPeer, ABC):
             )
         return keys  # type: ignore[return-value]
 
-    # @public_api
+    @public_api
     def get_validator(  # noqa: PLR0913
         self,
         datasource_name: Optional[str] = None,
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -2841,7 +2841,7 @@ class AbstractDataContext(ConfigPeer, ABC):
             isinstance(batch_request_updated, FluentBatchRequest)
             and dataframe is not None
         ):
-            data_asset = datasource.get_asset(
+            data_asset = datasource.get_asset(  # type: ignore[union-attr, return-value, arg-type]
                 asset_name=batch_request_updated.data_asset_name
             )
             batch_request_updated = data_asset.build_batch_request(dataframe=dataframe)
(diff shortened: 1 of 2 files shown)


NathanFarmer (changes requested):
We should not need to pass the dataframe to the validator.

alexsherstinsky:
> We should not need to pass the dataframe to the validator.

@NathanFarmer I agree -- please see Remark 1 above, where the `dataframe` would be passed as part of a "Runtime" `BatchRequest` (non-serializable Pandas and Spark extensions of the standard FDS `BatchRequest`).

What do you think about this approach?  Thanks.

NathanFarmer (changes requested):
Somehow my previous review was dismissed.

alexsherstinsky:
Indeed @NathanFarmer is correct -- thanks, @NathanFarmer for emphasizing the clean approach.  Indeed, we can cleanly obtain (at least in GX OSS) the `BatchRequest` and run `get_validator()` in a way that the `DataFrame` remains on `DataFrameAsset` without problems. Hence, I am closing this pull request.

PR #8104, opened 2023-06-13
[MAINTENANCE] Remove the mysql-mac-m1 and starburst directories from assets/docker

kenwade4:
- mysql dir works with m1 macs as of #8046
- trino dir is used for testing, not starburst


assets/docker/mysql-mac-m1/docker-compose.yml
```diff
@@ -1,18 +0,0 @@
-version: '3.2'
-services:
-  mysql_db:
-    image: mysql/mysql-server:8.0.20
-    volumes:
-        - ./conf.d:/etc/mysql/conf.d
-    environment:
-      MYSQL_ALLOW_EMPTY_PASSWORD: "true"
-      MYSQL_DATABASE: test_ci
-      MYSQL_ROOT_HOST: "%"
```
anthonyburdi, line 10: @Shinnnyshinshin this MYSQL_ROOT_HOST was a recent addition in one of your PRs, does it need to be added to the non m1 mysql docker compose?

Shinnnyshinshin: This was to address an error where the localhost IP wasn't "allowed" to connected to the newly connected doctor container.  It only appeared to me on my M1 machine, but thinking about it more closely, I don't think it needs to be added to the non-M1 docker compose.

[Here is the StackOverflow that I used to find the solution](https://serverfault.com/questions/793058/can-not-access-mysql-docker)

Shinnnyshinshin (approved):
Thank you @kenwade4 !

PR #8108, opened 2023-06-13
[BUGFIX] Fix conditional for pyspark compatibility

anthonyburdi:
Looks like we were checking the compatibility module instead of the actual pyspark import. The issue is that the truthyness of the module will not change whether pyspark is installed or not, which is what we are trying to test in the `if pyspark:`.


billdirks (approved):
LGTM, thanks!

great_expectations/core/batch.py
```diff
@@ -690,9 +690,9 @@ def head(self, *args, **kwargs) -> pd.DataFrame:
         return pd.DataFrame({})
 
 
-if pyspark:
+if pyspark_DataFrame:  # type: ignore[truthy-function]
```
alexsherstinsky: @anthonyburdi This does not sound right -- per your own idea, I thought that we could actually make `pyspark.DataFrame` the way of passing the `DataFrame` (and I got to appreciate that idea!).  Happy to discuss.  Thanks.

alexsherstinsky (changes requested):
@anthonyburdi I do not yet see this change to be needed.  Happy to discuss.  Thanks!

great_expectations/core/batch.py
anthonyburdi: Hi @alexsherstinsky, the issue with the code as-is was that the `if pyspark:` as written was checking truthyness of the compatibility/pyspark.py module and not whether pyspark was installed. We could have changed it to `if pyspark.pyspark:` but thought that since only the dataframe is imported that this might be cleaner.

alexsherstinsky: @anthonyburdi You got me to believe that using the pattern `if pyspark.pyspark` or `if pyspark.DataFrame` would work -- and it did!  So I utilized it uniformly; the benefit was full scoping (no need to create special variables).  So I would vote for that approach, because if that pyspark.DataFrame does not exist, the `SPARK_NOT_IMPORTED` will produce the relevant error message.  Thanks!

Kilo59: @anthonyburdi can you add a comment to this type ignore explaining why the `truthy-function` is "okay"?

Kilo59 (approved):

great_expectations/core/batch.py
anthonyburdi: That makes sense @alexsherstinsky, the issue though here is that the `if` statement was effectively not checking for `SPARK_NOT_IMPORTED` but instead for the existence of the `pyspark` compatibility module. That is a real bug and what this PR fixes.

anthonyburdi: Thanks for the catch!

anthonyburdi: @alexsherstinsky fixed, thanks for the discussion

alexsherstinsky (approved):
LGTM!  Thank you!

PR #8112, opened 2023-06-13
[MAINTENANCE] Pandas and SqlAlchemy 2.0 follow-up 

Shinnnyshinshin:
* Follow up to #8101 - one `docs-integration` test that needed to be updated.
* `docs-integration` pipeline will use constraints and more specific installation requirements. In the case of non-supported backends (like `redshift` and `bigquery`) we use constraints files and ensure that `pandas < 2.0.0` and `sqlalchemy < 2.0.0` are installed
* Clean up of `docs-integration` tests : requirements like Spark are better organized, and `pyarrow` requirements is applied to more general situations (we want to bring it in even if we aren't using `dremio`)
* Removal of V2 to V3 migration guide tests and fixtures, which are no longer supported

PR #8120, opened 2023-06-14
[MAINTENANCE] Fix unexpected_index_query and result_index_query in result_format tests

kenwade4:


Shinnnyshinshin:
Hi @kenwade4 there were some linting errors in this PR which prevented us from approving. Same tests will be addressed in #8112

kenwade4:
Same fix got merged in from another branch

PR #8136, opened 2023-06-16
[CONTRIB] Add custom expectation column count

data-han:

### DIFF
--- a/contrib/experimental/great_expectations_experimental/expectations/expect_column_count_to_be_between.py
+++ b/contrib/experimental/great_expectations_experimental/expectations/expect_column_count_to_be_between.py
@@ -1,0 +1,213 @@
+"""
+This is a template for creating custom ColumnAggregateExpectations.
+For detailed instructions on how to use it, please see:
+    https://docs.greatexpectations.io/docs/guides/expectations/creating_custom_expectations/how_to_create_custom_column_aggregate_expectations
+"""
+
+from typing import Dict, Optional
+
+from great_expectations.core.expectation_configuration import ExpectationConfiguration
+from great_expectations.exceptions import InvalidExpectationConfigurationError
+from great_expectations.execution_engine import (
+    ExecutionEngine,
+    PandasExecutionEngine,
+    SparkDFExecutionEngine,
+    SqlAlchemyExecutionEngine,
+)
+from great_expectations.expectations.expectation import ColumnAggregateExpectation
+from great_expectations.expectations.metrics import (
+    ColumnAggregateMetricProvider,
+    column_aggregate_partial,
+    column_aggregate_value,
+)
+
+
+# This class defines a Metric to support your Expectation.
+# For most ColumnAggregateExpectations, the main business logic for calculation will live in this class.
+class ColumnCountToBeBetween(ColumnAggregateMetricProvider):
+    # This is the id string that will be used to reference your Metric.
+    metric_name = "column.row_count"
+
+    # This method implements the core logic for the PandasExecutionEngine
+    @column_aggregate_value(engine=PandasExecutionEngine)
+    def _pandas(cls, column, **kwargs):
+        return column.notnull().sum()
+
+    # This method defines the business logic for evaluating your Metric when using a SqlAlchemyExecutionEngine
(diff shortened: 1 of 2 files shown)


data-han:
for the CLA, how is it trying to access his account @mh-iot ? in CLA i m using my email @gmail.com

contrib/experimental/great_expectations_experimental/expectations/expect_column_count_to_be_between.py
```diff
@@ -0,0 +1,213 @@
...
+    column_aggregate_partial,
+    column_aggregate_value,
+)
+
+
+# This class defines a Metric to support your Expectation.
+# For most ColumnAggregateExpectations, the main business logic for calculation will live in this class.
+class ColumnCountToBeBetween(ColumnAggregateMetricProvider):
+    # This is the id string that will be used to reference your Metric.
+    metric_name = "column.row_count"
```
alexsherstinsky: Is the name `column.row_count` for the metric a good fit for the expectation name containing "ColumnCount"?  I am thinking that making it clear what is being counted would help.  Please consider renaming one or the other.

If we are indeed counting columns, then using the metric dependency "table.columns" would help, because the count can then be obtained by `len()`.

Thank you very much!

alexsherstinsky (changes requested):
Comment/suggestion added.  Thank you!

contrib/experimental/great_expectations_experimental/expectations/expect_column_count_to_be_between.py
data-han: Hi there , thank for your feedback. What i was trying to do is to get the row count of a particular column including duplicates. This is a referenced to a problem faced by a user from Slack - so wanted to try creating this custom expectation :) 
image

data-han: In accordance to that, I have renamed all of the objects to row_count instead , please help to take a look :)

alexsherstinsky: Hello, @data-han -- thank you for your changes; now it makes it easier for me to see what is being done.  My question is why would we not be able to use the existing metric "table.row_count" and the expectation "expect_table_row_count_to_be_between"?  I am thinking that the row count would be the same for every column.  Please clarify if I am missing an important detail about what is being requested in Slack and your solution.  Thank you!

data-han: Hi @alexsherstinsky, thanks for the feedback once again. For table row count, it counts the total table row. Some columns include nulls, they will count them nonetheless. This is commonly done in pandas using len.

What I'm doing here would be to count individual column count that's avoiding null values and also including duplicated values.

For instance, if a table has 2 columns, both have 100 rows. Table count is 100. However, if column A has only 100 rows but 22 are null then count is 78 while column B may have 98 rows because 2 are null. So this aspect will output different result.

Would that answer your question?

alexsherstinsky: Hello, @data-han -- and thank you very much for your clarification-- I really appreciate it as it helped me finally understand your requirement.  Here is another idea for how existing expectations might be able to fulfill this requirement.  Please consider the expectation "expect_column_values_to_be_null" with the value of the `mostly` parameter equal to `1.0` (default).  Then, the output of executing this expectation will contain `unexpected_count` in the result object.  That would be exactly the number of rows where the given column value is not `None`.  Please let me know whether or not this approach would work for you.  Thank you!

data-han: @alexsherstinsky  Ah i see, will check it out. If that's working then we can close this. I will try to see if there is any other expectations I can proceed to create and try it out

alexsherstinsky: @data-han Great -- thank you -- please either let me know when I can close, or close it yourself, if the approach works for you and your team.  Thanks!

data-han:
could be resolved using other expectation

PR #8125, opened 2023-06-15
[DOCS] Enable docs versioning for 0.16.16

anthonyburdi:
Enable docs versioning for 0.16.16 as we prepare to release 0.17.0 by doing the following:
- Temporarily set onBrokenLinks to 'warn' instead of 'throw' until the broken links listed below are able to be addressed.
- Handle an import with an absolute path: `'/components/warnings/_cli_removal.md'`
- Set NODE_OPTIONS in both netlify and azure-pipelines-docs.yml to avoid `JavaScript heap out of memory` errors (see also https://github.com/gatsbyjs/gatsby/issues/15190).
- Update versioning instructions to temporarily change onBrokenLinks since we reference earlier versions (which will not be available during docs versioning).
- Increase deprecation warning threshold.

The following will need to be addressed in future PRs:
- [ ] Fix the broken links in the list below from a recent stack trace.
- [ ] Handle all imports with absolute paths (not just `'/components/warnings/_cli_removal.md'`).
- [ ] Remove deprecated code and reduce the threshold to 0.

```
7:03:18 PM: [ERROR] Unable to build website for locale en.
7:03:18 PM: [ERROR] Error: Docusaurus found broken links!
7:03:18 PM: 
7:03:18 PM: Please check the pages of your site in the list below, and make sure you don't reference any path that does not exist.
Note: it's possible to ignore broken links with the 'onBrokenLinks' Docusaurus configuration, and let the build pass.
7:03:18 PM: Exhaustive list of all broken links found:
7:03:18 PM: 
7:03:18 PM: - On source page path = /docs/0.16.16/deployment_patterns/how_to_use_gx_with_aws/how_to_use_gx_with_aws_using_athena:
7:03:18 PM:    -> linking to ../../../../docs/guides/validation/index.md#checkpoints (resolved as: /docs/guides/validation/index.md)
7:03:18 PM:    -> linking to ../../../../docs/guides/validation/index.md#actions (resolved as: /docs/guides/validation/index.md)
7:03:18 PM: 
7:03:18 PM: - On source page path = /docs/0.16.16/deployment_patterns/how_to_use_gx_with_aws/how_to_use_gx_with_aws_using_cloud_storage_and_pandas:
7:03:18 PM:    -> linking to ../../../../docs/guides/validation/index.md#checkpoints (resolved as: /docs/guides/validation/index.md)
7:03:18 PM:    -> linking to ../../../../docs/guides/validation/index.md#actions (resolved as: /docs/guides/validation/index.md)
[...]

PR #8144, opened 2023-06-20
[MAINTENANCE] Add env var to avoid out of memory error when building docs

anthonyburdi:
Our async build pipeline was failing due to an out of memory error while building docs. In https://github.com/great-expectations/great_expectations/pull/8125 we introduced an env variable to avoid this issue and it wasn't also implemented in the async pipeline. This PR rectifies that.


billdirks (approved):
LGTM, thanks!

PR #8154, opened 2023-06-21
[MAINTENANCE] Add --bash command for postgres reference env

anthonyburdi:
Add a `--bash` option to the postgres reference environment as a shortcut to enter into a bash session in the container instead of the jupyter notebook (e.g. to use a python repl or run a script).


alexsherstinsky (approved):
LGTM!

PR #8158, opened 2023-06-21
[MAINTENANCE] Raise error on unsupported versions of python.

billdirks:
```
% python                 
Python 3.11.3 (main, May 15 2023, 10:43:03) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import great_expectations
  File "/Users/bill/311Test/great_expectations/great_expectations/__init__.py", line 9, in <module>
    raise ImportError(
ImportError: Great Expectations is only supported on python 3.8 through 3.10. You are using: 3.11.3 (main, May 15 2023, 10:43:03) [Clang 14.0.6 ]
>>> 
```


great_expectations/__init__.py
```diff
@@ -1,3 +1,17 @@
+import sys
+
+PYTHON_SUPPORTED_VERSION = {"major": 3, "minor": {"lower": 8, "upper": 10}}
```
cdkini: we could use the `version` package here to make this easier.

`version.Version` has built in comparators

great_expectations/__init__.py
```diff
@@ -1,3 +1,17 @@
...
+if (sys.version_info.major != PYTHON_SUPPORTED_VERSION["major"]) or (
+    sys.version_info.major == PYTHON_SUPPORTED_VERSION["major"]
+    and sys.version_info.minor < PYTHON_SUPPORTED_VERSION["minor"]["lower"]
+    or sys.version_info.minor > PYTHON_SUPPORTED_VERSION["minor"]["upper"]
+):
+    raise ImportError(
+        "Great Expectations is only supported on python 3.8 through 3.10. "
```
cdkini: Nit - Could we capitalize Python?

great_expectations/__init__.py
cdkini: `from packaging import version` I believe

great_expectations/__init__.py
billdirks: Done.

great_expectations/__init__.py
billdirks: I would prefer not to take on a dependency. `sys` is builtin and the conditional is not complicated.
I do see comparators builtin if you pass in a string representation to `version.Version` but I don't know how to get this representation of the python runtime without using `sys` and then doing some manipulation of `sys.version` or `sys.version_info` which seems more complicated than this.

Shinnnyshinshin (approved):
I was really looking forward to this :) Thank you @billdirks

great_expectations/__init__.py
```diff
@@ -1,3 +1,17 @@
+import sys
+
+PYTHON_SUPPORTED_VERSION = {"major": 3, "minor": {"lower": 8, "upper": 10}}
+if (sys.version_info.major != PYTHON_SUPPORTED_VERSION["major"]) or (
+    sys.version_info.major == PYTHON_SUPPORTED_VERSION["major"]
```
Kilo59: Why do we care about major version at all?
Python 4 isn't going to be a thing, and users can't install us with Python 2?

billdirks: I guess I like to be explicit. People could install use via a `git clone` and not be gated via `pip`. I don't think python 2 users are going to use us but my preference is to leave the check in.

Kilo59 (approved):

PR #8184, opened 2023-06-26
[FEATURE] DatabricksSQL FDS

joshua-stauffer:
Add support for Databricks SQL through our fluent-datasource API


great_expectations/datasource/fluent/databricks_sql_datasource.py
```diff
@@ -0,0 +1,106 @@
...
+            For example: "databricks+connector://token:<token>@<host>:<port>/<database>"
+        http_path: HTTP path of Databricks SQL endpoint
+        assets: An optional dictionary whose keys are TableAsset or QueryAsset names and whose values
+            are TableAsset or QueryAsset objects.
+    """
+
+    type: Literal["databricks_sql"] = "databricks_sql"  # type: ignore[assignment]
+    connection_string: Union[ConfigStr, DatabricksDsn]
+    # connect args
+    http_path: str
```
cdkini: Not sure how I feel about having this extra field - I'd rather we just accept the connection string and then parse this out under the hood?

cc @Kilo59

Kilo59: @cdkini I agree. For a first pass if when can connect with just the connection string, we should do that.
We can add additional ways of connecting based on users' feedback.

great_expectations/datasource/fluent/databricks_sql_datasource.py
```diff
@@ -0,0 +1,105 @@
...
+
+def _parse_http_path_from_query(query: str) -> str | None:
+    url_components = parse.urlparse(query)
+    path = str(url_components.path)
+    parse_results: dict[str, list[str]] = parse.parse_qs(path)
+    path_results = parse_results.get("http_path", [])
+
+    if path_results and len(path_results) == 1:
+        return path_results[0]
+    return None
```
cdkini: Made this a global util since it is shared between the DSN and the FDS - should I move it as a staticmethod into the DSN? Who is responsible for parsing logic?

great_expectations/datasource/fluent/databricks_sql_datasource.py
```diff
@@ -0,0 +1,105 @@
...
+    code = "url.query"
+    msg_template = "URL query invalid"
+
+
+class _UrlHttpPathError(pydantic.UrlError):
+    """
+    Custom Pydantic error for missing http_path in DatabricksDsn query.
+    """
+
+    code = "url.http_path"
```
cdkini: Should this be `url.query.http_path`?

great_expectations/datasource/fluent/databricks_sql_datasource.py
```diff
@@ -0,0 +1,105 @@
...
+    msg_template = "HTTP path invalid"
+
+
+class DatabricksDsn(AnyUrl):
+    allowed_schemes = {
+        "databricks+connector",
+    }
+
+    @classmethod
+    def validate_parts(cls, parts: Parts, validate_port: bool = True) -> Parts:
```
cdkini, line 61: Only validating query and http_path for now - can come back for more robust error handling but want to keep the scope of this relatively contained

great_expectations/datasource/fluent/databricks_sql_datasource.py
```diff
@@ -0,0 +1,105 @@
...
+    from great_expectations.compatibility import sqlalchemy
+
+
+def _parse_http_path_from_query(query: str) -> str | None:
+    url_components = parse.urlparse(query)
+    path = str(url_components.path)
+    parse_results: dict[str, list[str]] = parse.parse_qs(path)
+    path_results = parse_results.get("http_path", [])
+
+    if path_results and len(path_results) == 1:
```
Kilo59, line 30:
```suggestion
    if path_results and len(path_results) == 1:
```
This will silently fail if the user provides multiple `http_path` query entries.

I would suggest we either warn/fail or just select the first `http_path` entry. Not sure off the top of my head which is the best approach. 🤔

great_expectations/datasource/fluent/databricks_sql_datasource.py
```diff
@@ -0,0 +1,105 @@
...
+
+        http_path = _parse_http_path_from_query(connection_string.query)
+        assert (
+            http_path
+        ), "Presence of http_path query string is guaranteed due to prior validation"
+
+        connect_args = {"http_path": http_path}
+        self._engine = sa.create_engine(
+            connection_string, connect_args=connect_args, **kwargs
+        )
```
cdkini: The only way I could get DatabricksSQL to work is the `connect_args` approach - feels finicky but I think worth it since we prevent users from having to pass `http_path` explicitly

great_expectations/datasource/fluent/databricks_sql_datasource.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def _parse_http_path_from_query(query: str) -> str | None:
+    url_components = parse.urlparse(query)
+    path = str(url_components.path)
+    parse_results: dict[str, list[str]] = parse.parse_qs(path)
+    path_results = parse_results.get("http_path", [])
+
+    if path_results and len(path_results) == 1:
+        return path_results[0]
```
Kilo59:
```suggestion
    if path_results:
            if len(path_results) >1:
               raise ValueError("only one `http_path` query entry is allowed")  # something like this
        return path_results[0]
```

This will silently fail if the user provides multiple `http_path` query entries.

I would suggest we ~either~ fail ~or just always select the first `http_path` entry. Not sure off the top of my head which is the best approach.~ 🤔

cdkini: the DSN has logic to raise an error if this is `None` - I shifted the burden of error handling to the callers of this util

cdkini: I don't think multiple `http_path` queries will be provided in a valid use-case (or maybe I'm wrong?)

Kilo59: That makes sense. If only one `http_path` is allowed then we should add a simple check to fail if the user provides multiple `http_path` entries.
Query params can be repeated in a URI.

Just updated the code suggestion.

great_expectations/datasource/fluent/databricks_sql_datasource.py
Kilo59: Yeah that seems right.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -933,13 +933,7 @@ def execution_engine_type(self) -> Type[SqlAlchemyExecutionEngine]:
...
         if self.connection_string != self._cached_connection_string or not self._engine:
             try:
-                model_dict = self.dict(
-                    exclude=self._get_exec_engine_excludes(),
-                    config_provider=self._config_provider,
-                )
-                connection_string = model_dict.pop("connection_string")
-                kwargs = model_dict.pop("kwargs", {})
-                self._engine = sa.create_engine(connection_string, **kwargs)
+                self._create_engine()
```
Kilo59:
```suggestion
                self._engine = self._create_engine()
```

Thanks for breaking out this method. Much easier to reason about and extend.

Can you make this one small change so we aren't modifying state and returning an `sqlalchemy.Engine`?

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -951,6 +945,15 @@ def get_engine(self) -> sqlalchemy.Engine:
...
         return self._engine
 
+    def _create_engine(self) -> sqlalchemy.Engine:
+        model_dict = self.dict(
+            exclude=self._get_exec_engine_excludes(),
+            config_provider=self._config_provider,
+        )
+        connection_string = model_dict.pop("connection_string")
+        kwargs = model_dict.pop("kwargs", {})
+        self._engine = sa.create_engine(connection_string, **kwargs)
```
Kilo59:
```suggestion
        return sa.create_engine(connection_string, **kwargs)
```

cdkini: yes, great call! done

great_expectations/datasource/fluent/databricks_sql_datasource.py
cdkini: done!

great_expectations/datasource/fluent/databricks_sql_datasource.py
```diff
@@ -0,0 +1,108 @@
...
+        kwargs = model_dict.pop("kwargs", {})
+
+        http_path = _parse_http_path_from_query(connection_string.query)
+        assert (
+            http_path
+        ), "Presence of http_path query string is guaranteed due to prior validation"
+
+        # Databricks connection is a bit finicky - the http_path portion of the connection string needs to be passed in connect_args
+        connect_args = {"http_path": http_path}
+        self._engine = sa.create_engine(
```
Kilo59:
```suggestion
        return sa.create_engine(
```

Need to make this one last change based on the changes to the parent.

tests/datasource/fluent/test_databricks_sql_datasource.py
```diff
@@ -0,0 +1,71 @@
...
+            "databricks+connector://token:my_token>@my_host:1234/my_db?my_query=data",
+            [
+                {
+                    "loc": ("connection_string",),
+                    "msg": "ConfigStr - contains no config template strings in the format '${MY_CONFIG_VAR}' or '$MY_CONFIG_VAR'",
+                    "type": "value_error",
+                },
+                {
+                    "loc": ("connection_string",),
+                    "msg": "HTTP path invalid",
```
Kilo59:
```suggestion
                    "msg": "'http_path' query param is invalid or missing",
```
Would this be a better error message?

Kilo59 (approved):
@cdkini @joshua-stauffer 

Nice work ya'll!!

PR #8194, opened 2023-06-27
[FEATURE] Enable updating active stores and CRUD for data docs sites

anthonyburdi:
This PR contains the following changes:
- Changing the active metadata store for instantiated data contexts (expectations, validations, checkpoints, profiler stores). This also enables the "add store then set store as active" workflow.
- Adding create, update and delete methods for data docs sites (read was already implemented for site names).


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1608,6 +1660,70 @@ def add_store(self, store_name: str, store_config: StoreConfigTypedDict) -> Stor
...
+
+    @public_api
+    @new_method_or_class(version="0.17.2")
+    def delete_data_docs_site(self, site_name: str):
+        """Delete an existing Data Docs Site.
+
+        Args:
+            site_name: Site name to delete.
+        """
+        if site_name not in self.config.data_docs_sites:
```
anthonyburdi: This might need to be idempotent - check consistency with datasources.

tests/data_context/test_loading_and_saving_of_data_context_configs.py
```diff
@@ -25,6 +25,31 @@ def test_add_store_immediately_adds_to_config(empty_data_context):
     assert "my_new_store" in read_config_from_file(config_filename)
 
 
+@pytest.mark.integration
```
anthonyburdi: Instead, just check that the persistence method is called.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -547,6 +550,19 @@ def stores(self) -> dict:
...
+    @expectations_store_name.setter
+    def expectations_store_name(self, value: str) -> None:
+        """Set the name of the expectations store.
+
+        Args:
+            value: New value for the expectations store name.
+        """
+
+        self.variables.expectations_store_name = value
+        self._save_project_config()
```
cdkini, line 564: Does this also work with `self.variables.save_config()`

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1608,6 +1660,70 @@ def add_store(self, store_name: str, store_config: StoreConfigTypedDict) -> Stor
         self._save_project_config()
         return store
 
+    @public_api
+    @new_method_or_class(version="0.17.2")
+    def add_data_docs_site(
```
cdkini, line 1665: Should this error if an existing resource is around?

tests/data_context/test_loading_and_saving_of_data_context_configs.py
anthonyburdi: Done

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: Good question - I was aiming for consistency - let me check a few other add methods. Is there a canonical doc on what our CRUD should do?

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: It sure does, but I thought better to use the method on the data context vs the method on a dependency (e.g. if the dependency changes we only would have to change the implementation of _save_project_config() vs all the callers).

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: It looks like in:
- datasource we overwrite
- checkpoint we overwrite but that is deprecated in favor of raising in `add_checkpoint` and upserting in `add_or_update_checkpoint`
- add_expectation_suite uses a flag to choose between `add` and `add_or_update` on the underlying store.

So it seems like we should error if a site exists but let me know if I'm mistaken!

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: Currently the data context methods do raise an exception if the item to be deleted does not exist, so our implementation is OK. Here is the behavior of each (listed in order they appear in abstract_data_context.py):
- NEW: `delete_data_docs_site()`: Raise InvalidKeyError in body in body of method if key (name) doesn't exist.
- `delete_store()`: Raise StoreConfigurationError in body in body of method if key (name) doesn't exist.
- `delete_datasource()`: Raise ValueError in datasource_store on `get_datasource()` call if key (name) doesn't exist.
- `delete_checkpoint()`: Raise InvalidKeyError in checkpoint_store if key (name, id) doesn't exist.
- `delete_expectation_suite()`: DataContextError raised in body of method if key (name) doesn't exist.
- `delete_profiler()`: InvalidKeyError raised in `delete_profiler` in rule_based_profiler.py if key (name or id) doesn't exist

cc @billdirks since we had discussed.

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: I think so! Does feel a little inconsistent but I prefer erroring

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1608,6 +1660,85 @@ def add_store(self, store_name: str, store_config: StoreConfigTypedDict) -> Stor
...
+            sites = self.config.data_docs_sites
+            sites[site_name] = site_config
+            self.variables.data_docs_sites = sites
+            self._save_project_config()
+
+    @public_api
+    @new_method_or_class(version="0.17.2")
+    def list_data_docs_sites(
+        self,
+    ) -> dict:
```
cdkini: Can we be more specific at all? `dict[str, someObj]`?

tests/data_context/abstract_data_context/test_data_docs_config_crud.py
```diff
@@ -0,0 +1,206 @@
...
+
+    @pytest.mark.unit
+    def test_add_data_docs_site_persists(
+        self,
+        ephemeral_context_with_defaults: EphemeralDataContext,
+        new_site_config: dict,
+    ):
+        new_site_name = "my_new_site"
+        with mock.patch(
+            "great_expectations.data_context.EphemeralDataContext._save_project_config"
```
cdkini, line 48: Do we need to mock this? What happens if it gets called on an in-memory context?

tests/data_context/abstract_data_context/test_store_config_crud.py
```diff
@@ -0,0 +1,49 @@
...
+    SETTER_METHOD_NAMES,
+)
+def test_store_name_setters_persist(
+    store_setter_method_name: str,
+    ephemeral_context_with_defaults: EphemeralDataContext,
+):
+    new_store_name = "new_store_name"
+    with mock.patch(
+        "great_expectations.data_context.EphemeralDataContext._save_project_config"
+    ) as mock_save_project_config:
```
cdkini, line 44: Same note here

tests/data_context/abstract_data_context/test_data_docs_config_crud.py
anthonyburdi: Mostly mocking just to make sure the method to persist gets called (`_save_project_config()`). We test the actual add in `test_add_data_docs_site` above.

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: Good call, changed to `dict[str, DataDocsSiteConfigTypedDict]`

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: this is a great change. I can't believe we were missing this for so long.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1608,6 +1660,85 @@ def add_store(self, store_name: str, store_config: StoreConfigTypedDict) -> Stor
...
+
+        Example site config dicts can be found in our "Host and share Data Docs" guides.
+
+        Args:
+            site_name: New site name to add.
+            site_config: Config dict for the new site.
+        """
+        if self.config.data_docs_sites is not None:
+            if site_name in self.config.data_docs_sites:
+                raise gx_exceptions.InvalidKeyError(
```
Shinnnyshinshin, line 1678: :heart:

great_expectations/data_context/data_context/abstract_data_context.py
Shinnnyshinshin: wanting to double-check that this change went in

tests/data_context/abstract_data_context/test_store_config_crud.py
Shinnnyshinshin: I'm glad we are getting these tests :rocket:

Shinnnyshinshin (approved):
Looks great. I've been really wanting these for DataDocs thank you :bow:

great_expectations/data_context/data_context/abstract_data_context.py
anthonyburdi: I accidentally included in another branch instead of here, good catch! It is included now.

PR #8199, opened 2023-06-27
[BUGFIX] Relax numpy version requirements for python 3.10

jcampbell:
Amazon Managed Airflow (MWAA) pins numpy to version 1.22.4 for airflow 2.4.3 and 2.5.1; it runs on python 3.10.

This PR relaxes the constraint on numpy for python 3.10 to include that version and allow support for the Amazon managed airflow.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,7 +14,7 @@ mistune>=0.8.4
 nbformat>=5.0
 notebook>=6.4.10
 numpy>=1.19.5; python_version == "3.8" or python_version == "3.9"
-numpy>=1.23.0; python_version >= "3.10"
+numpy>=1.21.2; python_version >= "3.10"
 packaging
 pandas>=1.1.0; python_version <= "3.8"
 pandas>=1.1.3; python_version == "3.9"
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -14,7 +14,7 @@ mistune>=0.8.4
 nbformat>=5.0
 notebook>=6.4.10
 numpy>=1.19.5; python_version == "3.8" or python_version == "3.9"
-numpy>=1.23.0; python_version >= "3.10"
+numpy>=1.21.2; python_version >= "3.10"
```
billdirks: You will need to update `ci/constraints-test/py310-min-install.txt` so we run tests against this minimum version.
Also, have you looked at usage stats for say python 3.9 users to see if they are often using this lower version of pandas?

billdirks: Talked to @jcampbell. AWS managed airflow pins numpy to a lower version than we currently support so we want to drop this version to support that.

jcampbell: Updated the PR description to better capture the issue, and the version to focus narrowly on compatibility.

billdirks (approved):
LGTM, thanks!

cdkini (approved):

PR #8204, opened 2023-06-28
[MAINTENANCE] enable typechecking in validator.py

tyler-hoffman:


great_expectations/validator/validator.py
```diff
@@ -2010,7 +2027,7 @@ def __init__(
         if self.expectation_engine is None:
             from great_expectations.compatibility import pyspark
 
-            if pyspark.DataFrame and isinstance(batch.data, pyspark.DataFrame):
```
tyler-hoffman, line 2013: @alexsherstinsky mypy was upset that we were evaluating this and it should always be truthy. Was this actually doing something?

great_expectations/validator/validator.py
```diff
@@ -1582,14 +1592,17 @@ def validate(  # noqa: C901, PLR0912, PLR0913, PLR0915
...
+
+            if not isinstance(expectation_suite, ExpectationSuite):
                 logger.error(
                     "Unable to validate using the provided value for expectation suite; does it need to be "
                     "loaded from a dictionary?"
                 )
-                if getattr(data_context, "_usage_statistics_handler", None):
+                if data_context and hasattr(data_context, "_usage_statistics_handler"):
                     # noinspection PyProtectedMember
                     handler = data_context._usage_statistics_handler
```
tyler-hoffman: Can we just use the `usage_statistics_handler` property on data_context here, rather than using the hasattr? I wasn't confident because `_usage_statistics_handler` does seem to be assigned in outside of init...

cdkini: I think that should be okay! It should be of type `UsageStatisticsHandler | None` based on the init code I'm reading.

great_expectations/validator/validator.py
billdirks: There is a little magic going on here. `pyspark` is imported on the line above from `great_expectations.compatibility`. If users don't have pyspark installed they will get an instance of our `NotImported` which overrides `__bool__` to be `False` (see [here](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/compatibility/not_imported.py#L33)).
Checking `if pyspark.DataFrame` is basically a check whether pyspark is installed or not and we only want to do the `isinstance` check if it is installed.

tyler-hoffman: Gotcha. I had a hunch it was going to be some magic like that. I wonder if there's a better way that we can make what's going on here clear to mypy 🤔 . But obvs that's out of scope for this. I'll revert this part. Thanks!

pyproject.toml
```diff
@@ -186,7 +186,6 @@ exclude = [
     'rule_based_profiler/rule_based_profiler\.py',  # 40
     'validation_operators/types/validation_operator_result\.py', # 35
     'validation_operators/validation_operators\.py',  # 16
-    'validator/validator\.py', # 54
```
Kilo59, line 189: 🎊

great_expectations/validator/validator.py
```diff
@@ -1711,9 +1726,11 @@ def validate(  # noqa: C901, PLR0912, PLR0913, PLR0915
         finally:
             self._active_validation = False
 
-        if getattr(data_context, "_usage_statistics_handler", None):
```
Kilo59, line 1714: @tyler-hoffman why are we changing this?
It's gone from 1 operation to 2.

Is it to help mypy type-narrow?

great_expectations/validator/validator.py
```diff
@@ -40,6 +41,7 @@
 )
 from great_expectations.core.metric_domain_types import MetricDomainTypes
 from great_expectations.core.run_identifier import RunIdentifier
+from great_expectations.core.usage_statistics.usage_statistics import UsageStatisticsHandler
```
Kilo59: Can these 3 imports be moved to the `if TYPE_CHECKING:` block?

Kilo59 (approved):
@tyler-hoffman Thanks for this!!

great_expectations/validator/validator.py
tyler-hoffman: Yeah, this was for type narrowing, but thinking about it again, the walrus operator is going to make this quite nice. I'll move that way.

PR #8221, opened 2023-06-29
[FEATURE] Host data docs for the postgres reference environment

anthonyburdi:
This is an example for the postgres reference environment of using a separate container to host the generated data docs for easy access.


examples/reference_environments/postgres/compose.yaml
```diff
@@ -36,7 +36,28 @@ services:
...
+      dockerfile: data_docs.Dockerfile
+    depends_on:
+      db:
+        condition: service_healthy
+    networks:
+      - dbnet
+    ports:
+      - "3000:3000"
+    volumes:
+      - gx_stores:/gx:ro
```
anthonyburdi: FYI the `:ro` is for read only access to the volume.

examples/reference_environments/postgres/jupyter.Dockerfile
```diff
@@ -5,3 +5,7 @@ WORKDIR /gx
 COPY ./postgres_example.ipynb ./
 
 RUN pip install great_expectations[postgresql]
+
+# Use this line to install GX from the develop branch,
+# or replace develop with the branch name you want to install from:
+RUN pip install git+https://github.com/great-expectations/great_expectations.git@develop
```
anthonyburdi, line 14: I thought it might be more likely that during debugging we would be using develop or another branch rather than the latest release so I included this, and maybe we consider adding this to other ref envs (with instructions to rebuild on change).

TrangPham (approved):
LGTM

Shinnnyshinshin (changes requested):
I ran into some issues trying to reproduce this work locally. Could we debug together before merging?

Shinnnyshinshin (approved):
LGTM. Thanks for pairing on this!

PR #8225, opened 2023-06-30
[MAINTENANCE] Refactor validations in Checkpoint to use `CheckpointValidationConfig`

cdkini:
Instead of black-box dicts, we should use the rich `CheckpointValidationConfig` type. Our APIs should support both but we should convert to the rich type under the hood so we have better type safety throughout checkpoint flows.


great_expectations/checkpoint/checkpoint.py
```diff
@@ -773,12 +781,16 @@ def __init__(  # noqa: PLR0913
         action_list: Optional[Sequence[ActionDict]] = None,
         evaluation_parameters: Optional[dict] = None,
         runtime_configuration: Optional[dict] = None,
-        validations: Optional[List[dict]] = None,
+        validations: Optional[List[dict] | List[CheckpointValidationConfig]] = None,
```
TrangPham: nit (non-blocking): IMO This formate is more readable:
```suggestion
        validations: List[dict] | List[CheckpointValidationConfig] | None = None,
```

great_expectations/checkpoint/checkpoint.py
```diff
@@ -853,7 +865,9 @@ def run_with_runtime_args(  # noqa: PLR0913
         action_list: Optional[Sequence[ActionDict]] = None,
         evaluation_parameters: Optional[dict] = None,
         runtime_configuration: Optional[dict] = None,
-        validations: Optional[List[dict]] = None,
+        validations: Optional[
+            Union[List[CheckpointValidationConfig], List[dict]]
+        ] = None,
```
TrangPham:
```suggestion
        validations: List[dict] | List[CheckpointValidationConfig] | None = None,
```

great_expectations/checkpoint/checkpoint.py
```diff
@@ -1154,7 +1158,9 @@ def __init__(  # noqa: PLR0913
...
```
TrangPham:
```suggestion
        validations: List[dict] | List[CheckpointValidationConfig] | None = None,
```

great_expectations/checkpoint/checkpoint.py
```diff
@@ -365,6 +359,20 @@ def run(  # noqa: C901, PLR0913, PLR0915
...
+    ) -> list[CheckpointValidationConfig]:
+        # We accept both dicts and rich config types but all internal usage should use the latter
+        if not validations:
+            return []
+        return [
+            CheckpointValidationConfig(**validation)
+            if isinstance(validation, dict)
+            else validation
+            for validation in validations
+        ]
```
TrangPham: (non-blocking): There are 3 instances of this block of code in this PR. Is there a way to make this more DRY?

cdkini: Yeah trying to figure out a better way but as long as this is incrementally better, I'd push to make these changes and then continue iterating with subsequent PRs - finding it quite difficult to refactor given how deeply entrenched some of these typing issues are

great_expectations/checkpoint/checkpoint.py
cdkini: I agree! I think `ruff` provides an auto-fix for a lot of these so I'm just making sure the types are correct (and going back later to clean up the actual formatting). I'll make these changes before merging

great_expectations/checkpoint/checkpoint.py
cdkini: moved to a util function - don't love it but it allows things to be shared. This is all a stopgap measure until we create a better Checkpoint (in `experimental`)

tests/checkpoint/test_checkpoint_config.py
```diff
@@ -62,50 +62,19 @@ def checkpoint(
...
-    {
-      "name": "store_validation_result",
-      "action": {
-        "class_name": "StoreValidationResultAction"
-      }
-    },
-    {
-      "name": "store_evaluation_params",
-      "action": {
-        "class_name": "StoreEvaluationParametersAction"
```
cdkini, line 78: The repr has differing outputs due to the variability of the `CheckpointValidationConfig`'s ordering - I figured it was easier to just test some top level keys but lmk if you disagree.

great_expectations/checkpoint/util.py
```diff
@@ -450,39 +450,48 @@ def get_updated_action_list(
 
 
 def does_batch_request_in_validations_contain_batch_data(
-    validations: Optional[List[dict]] = None,
+    validations: List[CheckpointValidationConfig],
```
Kilo59, line 453: I'm guessing `CheckpointValidationConfig` is just a `DictDot`? Or does it actually declare fields and types?

great_expectations/checkpoint/checkpoint.py
```diff
@@ -636,14 +630,14 @@ def config(self) -> CheckpointConfig:
         return self._checkpoint_config
 
     @property
-    def name(self) -> Optional[str]:
+    def name(self) -> str | None:
         try:
             return self.config.name
         except AttributeError:
             return None
```
Kilo59: Is `name` really optional/None-able?

great_expectations/checkpoint/checkpoint.py
```diff
@@ -1143,26 +1125,24 @@ def __init__(  # noqa: PLR0913
...
         # the following four arguments are used by SimpleCheckpointConfigurator
-        site_names: Union[str, List[str]] = "all",
-        slack_webhook: Optional[str] = None,
+        site_names: str | list[str] = "all",
+        slack_webhook: str | None = None,
         notify_on: str = "all",
-        notify_with: Union[str, List[str]] = "all",
-        expectation_suite_ge_cloud_id: Optional[str] = None,
+        notify_with: str | list[str] = "all",
+        expectation_suite_ge_cloud_id: str | None = None,
```
Kilo59, line 1141: `non-blocking`
How many overloads would it take to cover this `__init__` method?
If it's small could you define them in this PR?

great_expectations/core/configuration.py
```diff
@@ -16,6 +17,9 @@ def __init__(self, id: Optional[str] = None, name: Optional[str] = None) -> None
         self.name = name
         super().__init__()
 
+    def __repr__(self) -> str:
+        return json.dumps(self.to_dict(), indent=2)
```
Kilo59: Are we just using `json.dumps()` for the pretty formatting?
If so I would recommend using `pprint.pformat()` instead.
https://docs.python.org/3/library/pprint.html#pprint.pformat
```python

...

    def __repr__(self) -> str:
        return pf(self.to_dict(), indent=2, sort_dicts=True)
```
Can use `sort_dicts` if desired.

Also, you can take advantage of nice formatting features like `depth` which will cause the formatter to only show structures up to `x` depth, and uses `...` for everything else.
```python
>>> pf({"a": {"1": 1}, "b": {"2": 2}, "c": [1, 2, 3]}, depth=1)
... {'a': {...}, 'b': {...}, 'c': [...]}
```

Kilo59 (approved):
Couple non-blocking comments

great_expectations/checkpoint/util.py
cdkini: yup its an `AbstractConfig` (which inherits from `SerializableDictDot`)

great_expectations/checkpoint/checkpoint.py
cdkini: hmm no I don't think so - updating now

great_expectations/checkpoint/checkpoint.py
cdkini: ah I'm afraid it is quite cumbersome - I think it isn't worth the effort (but we can apply this to our updated checkpoint in experimental)

great_expectations/core/configuration.py
cdkini: love it!

PR #8238, opened 2023-07-05
[BUGFIX] Ensure DataAssistantResult.plot_expectations_and_metrics does not raise exceptions when no Metrics or Expectations are available to plot

alexsherstinsky:
### Scope
* Currently, if there are no Metrics or Expectations are available to plot, `DataAssistantResult.plot_expectations_and_metrics()` raises an `Exception`.  The present change safeguards against this situation and includes a test.

### Remarks
* JIRA: https://greatexpectations.atlassian.net/browse/DX-600
* This fix becomes critical for future, narrow-theme-focused `DataAssistant` realizations.

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```

### DIFF
--- a/tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
+++ b/tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
@@ -684,7 +684,10 @@ def test_onboarding_data_assistant_plot_expectations_and_metrics_correctly_handl
         include_column_names=include_column_names
     )
 
-    # This test passes only if absense of any metrics and expectations to plot does not cause exceptions to be raised.
+    """
+    This test passes only if absense of any metrics and expectations to plot does not cause exceptions to be raised.
+    Since column under test is semantically "NUMERIC", 8 (eight) numeric metrics to plot must be only remaining ones.
+    """
     column_domain_charts: List[dict] = [p.to_dict() for p in plot_result.charts[2:]]
     assert len(column_domain_charts) == 8
 
--- a/tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
+++ b/tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
@@ -678,39 +678,6 @@ def test_onboarding_data_assistant_result_plot_expectations_and_metrics_correctl
 
 
 @pytest.mark.integration
-@pytest.mark.slow  # 1.67s
-def test_onboarding_data_assistant_plot_expectations_and_metrics_correctly_handle_empty_plot_data(
-    bobby_columnar_table_multi_batch_probabilistic_data_context: DataContext,
-) -> None:
-    context: DataContext = bobby_columnar_table_multi_batch_probabilistic_data_context
-
-    batch_request: dict = {
-        "datasource_name": "taxi_pandas",
-        "data_connector_name": "monthly",
-        "data_asset_name": "my_reports",
-    }
-
-    include_column_names: List[str] = [
-        "congestion_surcharge",
-    ]
-
-    data_assistant_result: DataAssistantResult = context.assistants.onboarding.run(
-        batch_request=batch_request,
-        include_column_names=include_column_names,
-    )
-    plot_result: PlotResult = data_assistant_result.plot_expectations_and_metrics(
-        include_column_names=include_column_names
-    )
-
-    """
-    This test passes only if absense of any metrics and expectations to plot does not cause exceptions to be raised.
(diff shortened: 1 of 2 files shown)


tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
```diff
@@ -659,6 +659,39 @@ def test_onboarding_data_assistant_plot_include_and_exclude_column_names_raises_
     assert "either use `include_column_names` or `exclude_column_names`" in str(e.value)
 
 
+@pytest.mark.integration
+@pytest.mark.slow  # 1.67s
+def test_onboarding_data_assistant_plot_expectations_and_metrics_correctly_handle_empty_plot_data(
+    bobby_columnar_table_multi_batch_probabilistic_data_context: DataContext,
```
anthonyburdi: Are there any smaller fixtures we can use to reduce the runtime here?

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
```diff
@@ -659,6 +659,39 @@ def test_onboarding_data_assistant_plot_include_and_exclude_column_names_raises_
...
+        include_column_names=include_column_names,
+    )
+    plot_result: PlotResult = data_assistant_result.plot_expectations_and_metrics(
+        include_column_names=include_column_names
+    )
+
+    """
+    This test passes only if absense of any metrics and expectations to plot does not cause exceptions to be raised.
+    Since column under test is semantically "NUMERIC", 8 (eight) numeric metrics to plot must be only remaining ones.
+    """
```
Shinnnyshinshin: Hi @alexsherstinsky just a question about this test: So previously the DataAssistant was giving an error in situations where there was no Metric or Expectations to plot. Am I correct in understanding that this result would have had 8 by default (ie. would still have passed even without the changes to `data_assistant_result.py`?)

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
alexsherstinsky: @anthonyburdi I am sure that we can; I would only prefer to do it as a separate TODO -- and happy to add tickets for it and enact soon.

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
alexsherstinsky: @Shinnnyshinshin The 8 reflects the fact that there are 8 of them in the `OnboardingDataAssistant` for the specific column under test (`"congestion_surcharge"`).  Since per the Numeric Rule, this column is numeric, there will be 8 such entities computed.  But it also happens to undergo the Nullity/Non-Nullity rules, and they do not recommend anything to plot, because those plots were not implemented (at least not yet).  Thanks.

Shinnnyshinshin: Sorry, I'm a little slow getting up to speed on this change: 

My question is pretty simple: 

* up in `data_assistant_result.py`, there are two cases where having no metrics returns an empty lists or lists of charts.
* Here we have a list of 8 charts.

I now understand that it is 8+0=8 is what we are testing, but would there be a way to test just the 0 case?

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
alexsherstinsky: @anthonyburdi Update: Actually, I was able to provide a tight unit test as part of the present pull request.  In fact, without the present fix, this simple unit test fails with an exception.  Please re-review.  Thank you.

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
alexsherstinsky: @Shinnnyshinshin Indeed there is -- I just added a unit test.  In fact, without the present fix, this simple unit test fails with an exception. Please re-review. Thank you.  Please re-review.  Thank you!

Shinnnyshinshin: Ok great. If the unittest tests the change, do we need the integration test?

alexsherstinsky: @Shinnnyshinshin I made the update per our discussion -- the integration tests has been deleted.  Thanks.

Shinnnyshinshin (approved):
LGTM!

PR #8239, opened 2023-07-05
[MAINTENANCE] Reference environment consistency - unpin python version

anthonyburdi:
Instead of pinning python to 3.10, now that 3.11 is supported open the pin and add instructions for manually setting the python version.


Shinnnyshinshin (approved):
LGTM

PR #8240, opened 2023-07-05
[MAINTENANCE] Use data docs container for cloud blob stores

anthonyburdi:
Instead of only writing data docs to a cloud store, also write to local storage in a location that can be served via a different container. 
Also update comments in existing notebooks.


Shinnnyshinshin (approved):
Looks great!

PR #8263, opened 2023-07-10
[FEATURE] add http session manager to agent

joshua-stauffer:
This PR introduces an abstraction to manage agent http requests with cloud.
It also addresses a bug where agent workflows failed due to duplicate resource names, after a recent OSS update changed the exception type being raised.


ghost:
<details open>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=8263&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

####

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=8263&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map legend
</details>

great_expectations/agent/agent_server_session.py
```diff
@@ -0,0 +1,27 @@
+import pydantic
+from requests import Session
+
+from great_expectations.core.http import create_session
+
+
+class GxAgentServerSessionConfig(pydantic.BaseSettings):
+    gx_cloud_base_url: str = "https://api.greatexpectations.io"
```
Kilo59, line 8:
```suggestion
    gx_cloud_base_url: pydantic.HttpUrl = "https://api.greatexpectations.io" # type: ignore[assignment] # str will be coerced
```
https://docs.pydantic.dev/1.10/usage/types/#urls

I would suggest a `HttpUrl` type here. Raise issues early related to bad URL format.

great_expectations/agent/agent_server_session.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class AgentServerSession:
+    def __init__(self, config: GxAgentServerSessionConfig):
+        self._config = config
+
+    def session(self) -> Session:
+        return create_session(access_token=self._config.gx_cloud_access_token)
+
+    def build_url(self, resource_endpoint: str) -> str:
```
Kilo59, line 20: I think we have enums for the resource endpoints.
Then you can get rid of some of this defensive programming.

Kilo59: https://github.com/great-expectations/great_expectations/blob/70198abacaf7f0713400f64b87c8f51a8f04b137/great_expectations/data_context/cloud_constants.py#L19-L31

```suggestion
    def build_url(self, resource_endpoint: GXCloudRESTResource) -> str:
```

tests/agent/test_agent_server_session.py
```diff
@@ -0,0 +1,47 @@
...
+        gx_cloud_base_url="https://greatexpectations.io/test",
+        gx_cloud_organization_id="fd7838b2-e29e-4594-b29f-6bd378284a43",
+        gx_cloud_access_token="NGM3YTlkNGQtMWI3ZC00YTNhLTlhOTktMDQ5Y2I2YzAyNGE5",
+    )
+
+    server_session = AgentServerSession(config=config)
+
+    session = server_session.session()
+
+    assert isinstance(session, RequestsSession)
```
Kilo59, line 21: `non-blocking`
This assert seems unnecessary. The type checker already enforces this behavior.

great_expectations/agent/agent_server_session.py
Kilo59, line 8:
```suggestion
    gx_cloud_base_url: str = CLOUD_DEFAULT_BASE_URL
```
https://github.com/great-expectations/great_expectations/blob/70198abacaf7f0713400f64b87c8f51a8f04b137/great_expectations/data_context/cloud_constants.py#L5

Let's use the constant we define for this.

great_expectations/agent/actions/run_onboarding_data_assistant.py
```diff
@@ -32,15 +32,15 @@ def run(self, event: RunOnboardingDataAssistantEvent, id: str) -> ActionResult:
             )
             # if that didn't error, this name exists, so we add the timestamp
             expectation_suite_name = f"{expectation_suite_name} {timestamp}"
-        except StoreBackendError:
+        except (StoreBackendError, ValueError):
```
Kilo59, line 35: @joshua-stauffer were you able to track down the commit that caused a `ValueError` to be raised here?

cc @cdkini

great_expectations/agent/agent.py
```diff
@@ -212,29 +218,26 @@ def _reject_correlation_id(self, id: str):
...
                 f"Missing or badly formed environment variable\n{validation_err.errors()}"
             ) from validation_err
 
         # obtain the broker url and queue name from Cloud
+        agent_server_session = AgentServerSession(config)
 
-        agent_sessions_url = f"{config.gx_cloud_base_url}/organizations/{config.gx_cloud_organization_id}/agent-sessions"
+        agent_sessions_url = agent_server_session.build_url(
+            resource_endpoint="/agent-sessions"
+        )
```
roblim, line 238: I find this pattern (or maybe just the naming) a bit confusing. Seeing "agent_server_session" makes me want to think of the object as a Session-like object, but it's more like a Session factory with a util method. Why not just call `create_session` on `__init__` and stick the Session object in `self._session`. Then there could be a method like `AgentServerSession.post_to_resource_endpoint(resource_endpoint, ...)`, which internally calls `self._session.post(self._build_url(resource_endpoint=resource_endpoint))`.

great_expectations/agent/agent.py
```diff
@@ -212,29 +218,26 @@ def _reject_correlation_id(self, id: str):
...
-
-        class GxAgentConfigSettings(pydantic.BaseSettings):
-            gx_cloud_base_url: str = "https://api.greatexpectations.io"
-            gx_cloud_organization_id: str
-            gx_cloud_access_token: str
-
         try:
-            config = GxAgentConfigSettings()
+            config = GxAgentServerSessionConfig()
+            self._agent_server_session_config = config
```
roblim, line 227: Feels weird to stick this config on `GXAgent`. Also a bit confusing since there's already a `GXAgentConfig` in `self._config`, which is more concerned with the queue. Maybe we should be more explicit and indicate this somehow?

I think it would make more sense to stick the `GxAgentServerSessionConfig` config inside of `AgentServerSession`. Could make `config` init arg optional and if None, just use the same try block to populate `AgentServerSession._config`.

Also just realizing we're also in a method called `_get_config`... lot's to keep straight in working memory :).

great_expectations/agent/agent.py
```diff
@@ -212,29 +218,26 @@ def _reject_correlation_id(self, id: str):
...
         # obtain the broker url and queue name from Cloud
+        agent_server_session = AgentServerSession(config)
 
-        agent_sessions_url = f"{config.gx_cloud_base_url}/organizations/{config.gx_cloud_organization_id}/agent-sessions"
+        agent_sessions_url = agent_server_session.build_url(
+            resource_endpoint="/agent-sessions"
+        )
 
-        session = create_session(access_token=config.gx_cloud_access_token)
+        session = agent_server_session.session()
```
roblim, line 240: Isn't it expensive to keep re-instantiating these Session objects? Why not do once during __init__ by instantiating a `AgentServerSession` (the non-factory version alluded to above) and sticking in `self.agent_server_session` (or do once, but using `@property` pattern with a `self._agent_server_session` attr). Then when needed, agent methods could do something like `self.agent_server_session.post_to_resource_endpoint(resource_endpoint, ...)`.

roblim (changes requested):
Much cleaner! Just had some thoughts/ideas - LMK what you think.

great_expectations/agent/actions/run_onboarding_data_assistant.py
cdkini: hmm do we know what the `ValueError` was?

great_expectations/agent/actions/run_onboarding_data_assistant.py
```diff
@@ -32,15 +32,15 @@ def run(self, event: RunOnboardingDataAssistantEvent, id: str) -> ActionResult:
...
             # resource is unique
             pass
 
         try:
             self._context.get_checkpoint(name=checkpoint_name)
             # if that didn't error, this name exists, so we add the timestamp
             checkpoint_name = f"{checkpoint_name} {timestamp}"
-        except StoreBackendError:
+        except (StoreBackendError, ValueError):
             # resource is unique
```
cdkini, line 44: Should we be logging in these `except` blocks?

tests/agent/test_agent_server_session.py
cdkini: are we running `mypy` on these tests?

tests/agent/test_agent_server_session.py
```diff
@@ -0,0 +1,47 @@
...
+        ("/test-resource", "https://greatexpectations.io/test/"),
+        ("test-resource", "https://greatexpectations.io/test"),
+        ("test-resource", "https://greatexpectations.io/test/"),
+    ],
+)
+def test_server_session_build_url(resource_endpoint, base_url):
+    config = GxAgentServerSessionConfig(
+        gx_cloud_base_url=base_url,
+        gx_cloud_organization_id="fd7838b2-e29e-4594-b29f-6bd378284a43",
+        gx_cloud_access_token="NGM3YTlkNGQtMWI3ZC00YTNhLTlhOTktMDQ5Y2I2YzAyNGE5",
```
cdkini, line 38: These are both dummy values right?

great_expectations/agent/agent_server_session.py
```diff
@@ -0,0 +1,27 @@
...
+        """Construct a fully qualified URL."""
+        if not resource_endpoint.startswith("/"):
+            resource_endpoint = "/" + resource_endpoint
+        base_url = self._config.gx_cloud_base_url
+        if base_url.endswith("/"):
+            base_url = base_url[:-1]
```
cdkini, line 26: Nit - I feel like `urllib` or another built-in probably has this logic covered?

tests/agent/test_agent_server_session.py
Kilo59: Yes, we are but I meant if the implementation itself wasn't returning the `RequestSession` we would get a `mypy` error.

great_expectations/agent/agent_server_session.py
Kilo59: Can also normalize the lagging `/` with a simple [pydantic validator method](https://docs.pydantic.dev/1.10/usage/validators/).
And then this logic can just be...
```python
def build_url(self, resource_endpoint: GXCloudRESTResource) -> str:
    return f"{base_url}/organizations/{self._config.gx_cloud_organization_id}/{resource_endpoint}"
```
The combination of `mypy`, enum usage and the custom pydantic validator will ensure the constructed url is correct.

Or even better, return a `URL` type instead of `str`
https://docs.pydantic.dev/1.10/usage/models/#parsing-data-into-a-specified-type


```python
def build_url(self, resource_endpoint: GXCloudRESTResource) -> HttpUrl:
    return pydantic.parse_obj_as(
        HttpUrl,
        f"{base_url}/organizations/{self._config.gx_cloud_organization_id}/{resource_endpoint}"
)

```

great_expectations/agent/agent.py
Kilo59: I agree.
Also the fact that we are using a session seems like an implementation detail. `AgentServerSession` might be better named as `GxCloudClient`.

Although (non-blocking) I would also argue against having this abstraction at all, or at least keeping it very light.

Kilo59: Or maybe `AgentServerSession` should implement `__enter__` & `__exit__` so that it can be used as a context manager and the session should exist for the lifetime of the agent job.

```python
with GxCloudSession() as cloud_session_client:
    # job stuff
    method_that_needs_to_talk_to_cloud(cloud_session_client)
```

great_expectations/agent/agent.py
joshua-stauffer: thanks for the feedback @roblim, let's make this less confusing! the main drawback of the approach you propose is that we need to wrap all of the `Session` API - my goal was to make a pre-configured `Session` object available, and to have the code using the session (agent actions) and to be able to define general purpose config (like building the url) in a single place.

great_expectations/agent/agent_server_session.py
joshua-stauffer: The use case for this session wrapper is to allow agent actions to easily query the Cloud backend directly. This would add an unnecessary dependency on OSS code - I would expect that pretty much every time we use this, the endpoint we're hitting is not used by OSS (outside of the agent)

great_expectations/agent/agent_server_session.py
joshua-stauffer: great idea, thanks!

joshua-stauffer: this doesn't work because `HttpUrl` doesn't allow `localhost` + a port, which we need for local dev work.
```python
from pydantic import BaseModel, HttpUrl
class A(BaseModel):
    url: HttpUrl

a = A("http://localhost:5000")
  File "pydantic/main.py", line 332, in pydantic.main.BaseModel.__init__
TypeError: __init__() takes exactly 1 positional argument (2 given)
>>> a = A(url="http://localhost:5000")
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for A
url
  URL host invalid, top level domain required (type=value_error.url.host)
```

great_expectations/agent/agent_server_session.py
Kilo59: I don't understand the point, or maybe I just disagree with it.
The agent already depends on OSS code, and you are doing extra defensive coding to account for the fact that you aren't using something like this enum for `resource_endpoint`.

great_expectations/agent/agent_server_session.py
Kilo59: You can use `AnyUrl`

great_expectations/agent/agent_server_session.py
joshua-stauffer: the set of endpoints the Agent uses has little overlap with the set of endpoints that OSS uses (since, if OSS already is using an endpoint, the agent should just rely on OSS functionality). Given that the agent will be refactored to its own repo, this suggestion would mean we need to make a PR into OSS before we could update the agent to use some new mercury endpoint.

Kilo59: My understanding is that the only endpoints the agents are currently using are defined in this OSS enum. That's why I suggested using it. The parameter itself is even called `resource_endpoint`.
If that's not true, then please create a new enum or `Literal`. I don't care that much about which enum it uses.

great_expectations/agent/agent.py
joshua-stauffer: specifically, would renaming the class to `CloudSession`, and renaming the `session` method to `create_session` make this abstraction easier to grok?

roblim: I think that would be an improvement, maybe `GxCloudSession` to make things more clear? Def like `create_session` more. I also like @Kilo59's suggestion of `GxCloudClient`. 

Revisiting this again, I think more so than just the naming, it's all the disparate parts that feel off/confusing - I feel like there could be more encapsulation. Currently, in order to achieve the end of doing something with some GX Cloud endpoint using `AgentServerSession`, you need to: 
- Materialize a `GxAgentServerSessionConfig` object, then pass to `AgentServerSession` to instantiate an instance of `AgentServerSession`
- Use that `AgentServerSession` instance to materialize both a:
  - Session object and
  - Appropriate cloud endpoint
- Then call `session.[___](url=url...)`
- Handle/raise possible errors

Seems like a lot to keep track of whenever we want to hit GX Cloud endpoint in an Action.

I think `AgentServerSession` can at least handle materializing its own config if one isn't passed in. Given all of the above required steps to do X, it's not clear what we're getting by introducing a separate class, vs. say just having 2 functions called `get_agent_cloud_session` (which internally just figures out the right values, since `GxAgentServerSessionConfig` just seems to do just that - looks for env/global config vars) and `get_agent_cloud_url`, then using like above. Do you foresee `AgentServerSession` getting more functionality over time?

roblim: Er, realizing the first point doesn't need to happen every time. Maybe I'm also confused because the currently available impl example is within `GXAgent` (`_update_status`, etc.), but intention is also to be called within `AgentAction` as well.

great_expectations/agent/agent_server_session.py
joshua-stauffer: ah great, thanks!

PR #8278, opened 2023-07-11
[FEATURE]  single batch missingness data assistant

TrangPham:


great_expectations/rule_based_profiler/parameter_builder/unexpected_count_statistics_multi_batch_parameter_builder.py
```diff
@@ -220,13 +220,11 @@ def _build_parameters(  # PLR0915, PLR0915
...
             if single_batch_mode:
-                # TODO: <Alex>ALEX</Alex>
-                # TODO: <Thu>Here is where we use Tal's flowchart to compute the value of "mostly" (I put a placeholder).</Thu>
-                # TODO: <Thu>This is also where we can use "self.expectation_type" for asymmetric logic.</Thu>
-                mostly: np.float64 = np.float64(6.5e-1)
-                # TODO: <Thu></Thu>
-                # TODO: <Alex>ALEX</Alex>
-                result["mostly"] = mostly
+                unexpected_fraction: np.float64 = unexpected_count_fraction_values[-1]
+                expected_fraction: np.float64 = 1.0 - unexpected_fraction
```
alexsherstinsky: @TrangPham At least according to PyCharm, you might need to cast this result into `np.float64()` to avoid warnings.  Or we can choose to ignore those warnings completely and delete this cast throughout.  Happy to discuss.  Thanks.

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
```diff
@@ -76,7 +76,7 @@ class MetaDataAssistant(ABCMeta):
 
     def __new__(cls, clsname, bases, attrs):
         """
-        Instantiate class as part of descentants calling "__init__()" and register its type in "DataAssistant" registry.
+        Instantiate class as part of descendants calling "__init__()" and register its type in "DataAssistant" registry.
```
alexsherstinsky, line 79: 💙  -- thank you (my mistake)!

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
```diff
@@ -85,12 +85,14 @@ def __new__(cls, clsname, bases, attrs):
...
+            alias: Optional[str] = getattr(newclass, "__alias__", None)
+            if alias and not alias.startswith("_"):
+                from great_expectations.rule_based_profiler.data_assistant.data_assistant_dispatcher import (
+                    DataAssistantDispatcher,
+                )
 
-            # noinspection PyTypeChecker
-            DataAssistantDispatcher._register_data_assistant(data_assistant=newclass)
+                # noinspection PyTypeChecker
+                DataAssistantDispatcher._register(name=alias, data_assistant=newclass)
```
alexsherstinsky, line 95: @TrangPham This is nice! 💙 (hope no IDE errors, right?).  Thanks!

great_expectations/rule_based_profiler/data_assistant/data_assistant_dispatcher.py
```diff
@@ -64,25 +64,6 @@ def __getattr__(self, name: str) -> DataAssistantRunner:
 
         return data_assistant_runner
 
-    @classmethod
```
alexsherstinsky, line 67: 💙

tests/rule_based_profiler/parameter_builder/test_unexpected_count_statistics_multi_batch_parameter_builder.py
```diff
@@ -756,3 +759,31 @@ def test_unexpected_count_statistics_multi_batch_parameter_builder_bobby_check_s
...
+        # Standardizes for expect_column_values_to_be_null
+        ["expect_column_values_to_be_null", 1.0, 1.0],
+        ["expect_column_values_to_be_null", 0.997, 0.99],
+        ["expect_column_values_to_be_null", 0.99, 0.99],
+        ["expect_column_values_to_be_null", 0.98, 0.975],
+        ["expect_column_values_to_be_null", 0.975, 0.975],
+        # returns mostly as-is
+        ["expect_column_values_to_be_null", 0.9612, 0.9612],
+        ["expect_column_values_to_be_null", 0.1010, 0.1010],
+        #####
```
alexsherstinsky, line 770: Very minor: Do we need `###` -- if there is a comment to add here, or a blank line -- perhaps we should do it?  Thanks!

tests/rule_based_profiler/parameter_builder/test_unexpected_count_statistics_multi_batch_parameter_builder.py
```diff
@@ -756,3 +759,31 @@ def test_unexpected_count_statistics_multi_batch_parameter_builder_bobby_check_s
         "expectation_type",
         "evaluation_parameter_builder_configs",
     }
+
+
+@pytest.mark.unit
+@pytest.mark.parametrize(
```
alexsherstinsky, line 758: @TrangPham This looks like it matches the specifications diagrammed by @talagluck -- UAT and/or dogfooding would be great to organize for this capability as well.  Thank you.

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
TrangPham: No problem. I have issues with spelling too. I installed a spell checker extension on VSCode. I'm not sure if pycharm has a similar feature.

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
```diff
@@ -87,3 +90,45 @@ def test_column_value_missing_data_assistant_plot_expectations_and_metrics_corre
...
+    df = pd.DataFrame(
+        {
+            "non-null": [i for i in range(100)],
+            "null": [None for _ in range(100)],
+            "low-null": [None for _ in range(38)] + [i for i in range(62)],
+        }
+    )
+    batch_request = asset.build_batch_request(dataframe=df)
+
+    context.assistants._register("my_data_assistant", ColumnValueMissingDataAssistant)
```
alexsherstinsky, line 52: @TrangPham -- Brilliant! 💙

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
```diff
@@ -87,3 +90,45 @@ def test_column_value_missing_data_assistant_plot_expectations_and_metrics_corre
...
+    result = context.assistants.my_data_assistant.run(batch_request=batch_request)
+
+    assert len(result.expectation_configurations) == 3
+
+    expected_results = {
```
alexsherstinsky, line 57: This is a very cool way to test it! 💙

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
```diff
@@ -87,3 +90,45 @@ def test_column_value_missing_data_assistant_plot_expectations_and_metrics_corre
...
+            "mostly": 1.0,
+            "expectation": "expect_column_values_to_not_be_null",
+        },
+        "low-null": {
+            "mostly": 0.6,
+            "expectation": "expect_column_values_to_not_be_null",
+        },
+    }
+
+    for expectation in result.expectation_configurations:
```
alexsherstinsky, line 69: @TrangPham Do we really need this assertion (as in "what else can `expectation_type` be here?")?  I am fine either way.  Your call.

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
```diff
@@ -87,3 +90,45 @@ def test_column_value_missing_data_assistant_plot_expectations_and_metrics_corre
...
+        assert expectation.expectation_type in [
+            "expect_column_values_to_not_be_null",
+            "expect_column_values_to_be_null",
+        ]
+        column = expectation.kwargs["column"]
```
alexsherstinsky, line 74: @TrangPham We have a practice of adding type hints.  The documented style is to add it when it is hard to tell what it would be (so, `column: str` for this case).  I tend to err on the side of adding them, rather than skipping them (arguing consistency, even when it slightly contravenes the style guide).  Hence, please look over your code in this PR and decide what you will do.  Thanks!

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
```diff
@@ -87,3 +90,45 @@ def test_column_value_missing_data_assistant_plot_expectations_and_metrics_corre
...
+        assert expectation.kwargs["mostly"] == expected_results[column]["mostly"]
```
alexsherstinsky, line 75: @TrangPham Again, this assertion seems to be not needed, because there ought to be lower level tests that do that for `Expectation`, `ExpectationConfiguration`, and ``ExpectationConfigurationBuilder`.

tests/rule_based_profiler/parameter_builder/test_unexpected_count_statistics_multi_batch_parameter_builder.py
TrangPham: Sounds great! Do we merge then allow for UAT? They should be able to register the class as an assistant similar to the integration test I made.

alexsherstinsky (approved):
@TrangPham This is great!  I am pre-approving under the condition that you go over my questions and comments and handle them as you deem appropriate.  Thank you very much!

tests/rule_based_profiler/parameter_builder/test_unexpected_count_statistics_multi_batch_parameter_builder.py
alexsherstinsky: @TrangPham Definitely merge now!  If UAT reveals any issues, we can update then.  Thanks!

great_expectations/rule_based_profiler/parameter_builder/unexpected_count_statistics_multi_batch_parameter_builder.py
```diff
@@ -264,6 +262,35 @@ def _build_parameters(  # PLR0915, PLR0915
         )
 
 
+def _standardize_mostly_for_single_batch(  # noqa: PLR0911
+    expectation_type: str, mostly: np.float64
+) -> np.float64:
+    """
+    Applies business logic to standardize "mostly" value for single-Batch case.
+    """
+    if expectation_type == "expect_column_values_to_be_null":
```
donaldheppner, line 271: I think this would read more easily as `if else` with a single return statement

Also, are we not rounding mostly to the nearest 2.5%?

great_expectations/rule_based_profiler/parameter_builder/unexpected_count_statistics_multi_batch_parameter_builder.py
```diff
@@ -264,6 +262,35 @@ def _build_parameters(  # PLR0915, PLR0915
...
+
+        if mostly >= 0.99:  # noqa:  PLR0915
+            return np.float64(0.99)
+
+        if mostly >= 0.975:  # noqa:  PLR0915
+            return np.float64(0.975)
+
+        return mostly
+
+    if expectation_type == "expect_column_values_to_not_be_null":
```
donaldheppner, line 283: again, a single return statement would help readability here.

Finally, don't these same conditions _determine_ the expectation type? Is there a way to dedupe this logic?

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
```diff
@@ -87,3 +90,45 @@ def test_column_value_missing_data_assistant_plot_expectations_and_metrics_corre
     # This test passes only if absense of any metrics and expectations to plot does not cause exceptions to be raised.
     column_domain_charts: List[dict] = [p.to_dict() for p in plot_result.charts[2:]]
     assert len(column_domain_charts) == 0
+
+
+@pytest.mark.integration
+@pytest.mark.slow
```
donaldheppner, line 38: Is this slow?

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
donaldheppner: Since the code we are writing and testing is focused on the "mostly" parameter, I consider this test relevant .. and it's super cheap/quick

tests/rule_based_profiler/parameter_builder/test_unexpected_count_statistics_multi_batch_parameter_builder.py
donaldheppner: In the 96% case (and certainly the 10% case) the expectation returned should be `expect_column_values_to_not_be_null`

tests/rule_based_profiler/parameter_builder/test_unexpected_count_statistics_multi_batch_parameter_builder.py
```diff
@@ -756,3 +759,31 @@ def test_unexpected_count_statistics_multi_batch_parameter_builder_bobby_check_s
...
+        ["expect_column_values_to_be_null", 0.1010, 0.1010],
+        #####
+        # Standardizes for expect_column_values_to_not_be_null
+        ["expect_column_values_to_not_be_null", 1.0, 1.0],
+        ["expect_column_values_to_not_be_null", 0.997, 0.99],
+        ["expect_column_values_to_not_be_null", 0.99, 0.99],
+        # Rounds down to nearest 0.025
+        ["expect_column_values_to_not_be_null", 0.62, 0.6],
+        ["expect_column_values_to_not_be_null", 0.64, 0.625],
+        ["expect_column_values_to_not_be_null", 0.062, 0.05],
```
donaldheppner, line 778: In the 97.5% cases and up the expectation returned should be `expect_column_values_to_be_null`

donaldheppner (changes requested):

great_expectations/rule_based_profiler/parameter_builder/unexpected_count_statistics_multi_batch_parameter_builder.py
alexsherstinsky: @donaldheppner Regarding the return statement, I feel differently -- in fact, at one memorable customer engagement, they specifically enforced (!) the style where if there are returns, then there should be no "else".  To me, the style @TrangPham has it now reads well (but I emphasize that this is one person's opinion).  What I can say is that wherever I had found in our GX code base the cases of "if/else" where there are returns, I corrected them, because "return" is such that nothing after it is reachable (i.e., arguing for it being "logical").  Again, I emphasize that this is not a big point for me.  Thank you for considering.

great_expectations/rule_based_profiler/parameter_builder/unexpected_count_statistics_multi_batch_parameter_builder.py
alexsherstinsky: @donaldheppner I see that @TrangPham is rounding down to the nearest `2.5%` in this code:
```
        # round down to nearest 0.025
        return np.floor(mostly * 40) / 40
```
Is this not correct?  Thank you.

great_expectations/rule_based_profiler/parameter_builder/unexpected_count_statistics_multi_batch_parameter_builder.py
```diff
@@ -264,6 +262,35 @@ def _build_parameters(  # PLR0915, PLR0915
...
+        if mostly >= 1.0:  # noqa:  PLR0915
+            return np.float64(1.0)
+
+        if mostly >= 0.99:  # noqa:  PLR0915
+            return np.float64(0.99)
+
+        if mostly >= 0.975:  # noqa:  PLR0915
+            return np.float64(0.975)
+
+        return mostly
```
alexsherstinsky, line 281: @TrangPham This may be a bug: The logic says that if mostly is between `0.975` and `0.99` then we emit `expect_column_values_to_not_be_null` with rounding:
```
`expect_column_values_to_not_be_null()` where mostly is the percent missing rounded down to the nearest 2.5% (e.g. 62% missing gets a mostly of .6, 64% gets a mostly of .625)
```
I believe that @donaldheppner is correct in his earlier comment.  Happy to discuss.  Thank you.

great_expectations/rule_based_profiler/parameter_builder/unexpected_count_statistics_multi_batch_parameter_builder.py
alexsherstinsky: > again, a single return statement would help readability here.
> 
> Finally, don't these same conditions _determine_ the expectation type? Is there a way to dedupe this logic?

@donaldheppner This is how it works: The DataAssistant sends to ParameterBuilder the `expectation_type` that it is concerned with (only the two are possible), and the ParameterBuilder computes the `mostly` for it (this is one of the strengths of the Profiler -- its components are computational blocks that implement the plurality of functions, even if the way they are put together does not read like that function itself -- this makes it a "close to machine" layer, meaning that a "friendlier" layer for programmers can be added on top).  I believe that @TrangPham 's code is correct; we must be very thorough with unit tests (as in: pay a lot of attention to the test `test_standardize_mostly_for_single_batch()`) to ensure that I am not wrong here; we also need UAT to test against real-life scenarios.

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
alexsherstinsky: @donaldheppner Yes, agreed only the "mostly" part (and on the fact that it is negligible cost).  Thanks.

tests/rule_based_profiler/parameter_builder/test_unexpected_count_statistics_multi_batch_parameter_builder.py
alexsherstinsky: @donaldheppner I am not following your comments/notes: all values of mostly can exist for either expectation.  We process each expectation separately and compute the mostly for it based on the percentage of missing data (missing means NULLs for one expectation and non-NULL for the other).

@TrangPham Having said that, how about you and I review the test method and the logic together with a fine-tooth comb to make sure that we did not overlook anything (unless you feel that this has already been done).

Thanks.

great_expectations/rule_based_profiler/data_assistant/data_assistant.py
TrangPham: No errors within my IDE or the CI pipeline.

tests/rule_based_profiler/parameter_builder/test_unexpected_count_statistics_multi_batch_parameter_builder.py
TrangPham: Ah yes - this is slightly confusing here.

The flow of the code currently is as follows:
- DA calls out to this Param Builder 4 times (single-batch non-null, single-batch null, multi-batch non-null, multi-batch null)
- These params are passed to Expectation Config Builder - which uses `condition` to decide if it should output the expectation. The condition is currently set to use the mostly value to decide if it should output the expectation.

@alexsherstinsky My assumption is that underneath the rule profiler and metrics are not being recomputed with each of the Param Builders. Is this correct?

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
TrangPham: Currently there cannot be other expectation_types returned here. This is a defensive assertion to ensure that this remains the case in the future.

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
TrangPham: Thanks! I'll update the tests to ensure that they have the same standard of typing as our main code.

great_expectations/rule_based_profiler/parameter_builder/unexpected_count_statistics_multi_batch_parameter_builder.py
TrangPham: Yes that is correct. We only round for one case. Please see diagram attached.
Screenshot 2023-07-12 at 9 57 38 AM

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
TrangPham: I have seen tests marked slow if they take more than 1 second. What is the threshold for marking slow?

tests/rule_based_profiler/parameter_builder/test_unexpected_count_statistics_multi_batch_parameter_builder.py
alexsherstinsky: @TrangPham Thank you for this question as it gives me a chance to highlight some aspects of the architecture.  Indeed, this `ParameterBuilder` does not compute any metrics (unlike its older cousin `MeanUnexpectedMapMetricMultiBatchParameterBuilder`, which optionally can be configured to do so).  The reason for this is that all metrics that could ever be needed are supplied to it by the caller and are reused.  This is one of the nice aspects of our `MissignessDataAssistant` design.  Thank you.

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
alexsherstinsky: 🙇

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
```diff
@@ -27,3 +32,45 @@ def test_column_value_missing_data_assistant_result_plot_expectations_and_metric
...
+def test_single_batch_multiple_columns(empty_data_context):
```
cdkini: what type of context is this? I want to say file - I think we can leverage an ephemeral context so we don't have side effects here

alexsherstinsky: @cdkini Our `empty_data_context` is indeed `FileDataContext` with nothing in it.  @TrangPham please feel free to try `ephemeral_context_with_defaults` and see if it works.  (It is defined in "tests/conftest.py".) Thank you both.

tests/rule_based_profiler/data_assistant/test_column_value_missing_data_assistant.py
donaldheppner: it is 1s ... and I think we have a tool that can annotate slow tests with a comment indicating the number of seconds

donaldheppner (approved):

PR #8288, opened 2023-07-11
[MAINTENANCE] Pytest Mark - `RuleBasedProfiler` and others

Shinnnyshinshin:
# Update Pytest Mark
- RuleBasedProfiler folder
- Some others as well. 
- Used module-level marks where appropriate.

### DIFF
--- a/tests/test_great_expectations.py
+++ b/tests/test_great_expectations.py
@@ -321,8 +321,8 @@ def test_stats_mixed_expectations():
     assert expected == actual
 
 
+@pytest.mark.filesystem
 class TestIO(unittest.TestCase):
-    @pytest.mark.filesystem
     def test_read_csv(self):
         script_path = os.path.dirname(os.path.realpath(__file__))  # noqa: PTH120
         _ = gx.read_csv(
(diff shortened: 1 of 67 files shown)


Kilo59 (approved):

tests/test_great_expectations.py
```diff
@@ -312,12 +322,14 @@ def test_stats_mixed_expectations():
 
 
 class TestIO(unittest.TestCase):
+    @pytest.mark.filesystem
```
Kilo59:
```suggestion

@pytest.mark.filesystem
class TestIO(unittest.TestCase):
```

You can mark the `TestClass` itself to cause all of its methods to become marked as well.

https://docs.pytest.org/en/7.1.x/example/markers.html#marking-whole-classes-or-modules

```python
import pytest


@pytest.mark.webtest
class TestClass:
    def test_startup(self):
        pass

    def test_startup_and_more(self):
        pass
```

Shinnnyshinshin: Thank you for this @Kilo59

Shinnnyshinshin: I'll keep it in mind for upcoming test classes.

PR #8301, opened 2023-07-13
[MAINTENANCE] Pytest Mark - Render folder

Shinnnyshinshin:
- Adds `@pytest.mark` to `/tests/render/` folder. 
- Moved helper methods to own `util.py` file.
- removed `print()` statements. 
- Includes marks in #8299 so the initial diff will include changes from that PR as well.


tests/render/util.py
```diff
@@ -0,0 +1,130 @@
+import copy
+from typing import Dict, List, Optional
+
+import nbformat
+from nbconvert.preprocessors import ExecutePreprocessor
+from nbformat.notebooknode import NotebookNode
```
Shinnnyshinshin, line 6: These were mixed in with actual tests for utils. I moved to its own file, and adjusted imports accordingly

tests/render/test_column_section_renderer.py
```diff
@@ -1029,7 +993,6 @@ def test_ExpectationSuiteColumnSectionRenderer_expectation_with_string_list_meta
     result_json = (
         ExpectationSuiteColumnSectionRenderer().render(expectations).to_json_dict()
     )
-    print(result_json)
```
Kilo59: Why are you removing these print statements as part of this PR?

Shinnnyshinshin: They are outputted to the console when `pytest` is run, and are tested using `assert` statements directly afterwards. It was a part of some clean-up that I was attempting to do as part of this PR.

Do you think it would be better to keep them in?

Kilo59: @Shinnnyshinshin 
pytest captures stdout/stderr by default, and only shows it on a test failure. This is super useful for debugging failing tests.
If you are experiencing something different than something is wrong.

Where did you see this extra console output when running the tests? Locally or in CI?
Were you running `pytest` with the `-s` flag?

https://docs.pytest.org/en/7.1.x/how-to/capture-stdout-stderr.html#default-stdout-stderr-stdin-capturing-behaviour

tests/render/test_column_section_renderer.py
```diff
@@ -237,7 +211,7 @@ def test_ProfilingResultsColumnSectionRenderer_render_header_with_unescaped_doll
         [evr_with_unescaped_dollar_sign],
         column_type=[],
     ).to_json_dict()
-    print(content_block)
+    print(content_block.to_json_dict())
```
Kilo59:
```suggestion
    print(content_block)
```

tests/render/test_column_section_renderer.py
```diff
@@ -140,34 +140,7 @@ def test_render_expectation_suite_column_section_renderer(titanic_expectations):
             )
 
 
-@pytest.mark.smoketest
-def test_ProfilingResultsColumnSectionRenderer_render(
```
Kilo59, line 144: Why is this test being deleted?

tests/render/test_column_section_renderer.py
```diff
@@ -1273,8 +1247,6 @@ def test_ValidationResultsColumnSectionRenderer_render_header_evr_with_unescaped
         validation_results=[evr_with_unescaped_dollar_sign],
     )
 
-    print(content_block.to_json_dict())
```
Kilo59, line 1276: `non-blocking`
You missed adding this one back.

tests/render/test_default_jinja_view.py
```diff
@@ -159,8 +160,8 @@ def test_rendering_components_without_section_loop_index():
             "content_block_loop": {"index": 2},
         }
     )
-    print(rendered_doc)
     rendered_doc = rendered_doc.replace(" ", "").replace("\t", "").replace("\n", "")
+    print(rendered_doc)
```
Kilo59: Looks like you are changing the order of when these prints are being displayed.
I would defer to the original test author regarding when the debugging print statements should invoked.

Kilo59: I think it would probably be easier to do a git checkout from develop but keep your marker changes, rather than manually trying to add back each line.

tests/render/test_column_section_renderer.py
```diff
@@ -1273,8 +1247,6 @@ def test_ValidationResultsColumnSectionRenderer_render_header_evr_with_unescaped
         validation_results=[evr_with_unescaped_dollar_sign],
     )
```
Kilo59, line 1249: print(content_block.to_json_dict())

Kilo59 (approved):

tests/render/test_column_section_renderer.py
Shinnnyshinshin: added back :)

PR #8302, opened 2023-07-13
[MAINTENANCE] Replace many repetitive slow integration tests for ParameterBuilder components with a few fast unit tests

alexsherstinsky:
### Scope
Instead having repetitive tests for the same underlying functionality running for multiple `ParameterBuilder` components, we provide a few unit tests for that low-level functionality, thereby eliminating up to approximately twenty slow integration tests.

### Remarks
* JIRA: DX-613/DX-633

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```


great_expectations/rule_based_profiler/data_assistant/column_value_missing_data_assistant.py
```diff
@@ -228,10 +228,10 @@ def _build_column_value_missing_rule() -> Rule:
         map_metric_name = "column_values.null"
         evaluation_parameter_builder_configs = [
             ParameterBuilderConfig(
-                **total_count_metric_multi_batch_parameter_builder_for_evaluations.to_json_dict()
+                **column_values_null_unexpected_count_metric_multi_batch_parameter_builder_for_evaluations.to_json_dict()
```
alexsherstinsky, line 226: Reordering for consistency/elegance.

great_expectations/rule_based_profiler/data_assistant/column_value_missing_data_assistant.py
```diff
@@ -228,10 +228,10 @@ def _build_column_value_missing_rule() -> Rule:
...
             ),
             ParameterBuilderConfig(
-                **column_values_null_unexpected_count_metric_multi_batch_parameter_builder_for_evaluations.to_json_dict()
+                **total_count_metric_multi_batch_parameter_builder_for_evaluations.to_json_dict()
```
alexsherstinsky, line 229: Reordering for elegance/style -- "total goes last".

tests/rule_based_profiler/test_rule_based_profiler.py
```diff
@@ -946,9 +946,9 @@ def test_reconcile_profiler_rules_existing_rule_full_rule_override_update(
     assert effective_rule_configs_actual == expected_rules
 
 
+@pytest.mark.unit
```
alexsherstinsky, line 949: Reordering for elegance: type of test decorator goes on top.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    my_dependent_parameter_builder.build_parameters(
+        domain=domain,
+        variables=variables,
+        parameters=parameters,
+        batch_request=None,
+        runtime_configuration=None,
+    )
+    assert mock_parameter_builder__build_parameters.call_count == 3
```
donaldheppner: I don't think this test makes sense, it's only testing the dummy

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    my_evaluation_dependency_0_parameter_builder.build_parameters(
+        domain=domain,
+        variables=variables,
+        parameters=parameters,
+        batch_request=None,
+        runtime_configuration=None,
+    )
+    assert mock_parameter_builder__build_parameters.call_count == 1
```
donaldheppner: aren't you invoking your dummy parameter builder and then asserting that it was invoked? I fail to see what is being tested here

donaldheppner (changes requested):

great_expectations/rule_based_profiler/data_assistant/column_value_missing_data_assistant.py
```diff
@@ -52,7 +52,7 @@ class ColumnValueMissingDataAssistant(DataAssistant):
         ColumnValueMissingDataAssistantResult
     """
 
-    __alias__: str = "_missingness"
+    __alias__: str = "missingness"
```
donaldheppner: please consider this change in a different PR, it's unrelated to the description

donaldheppner (changes requested):

great_expectations/core/util.py
```diff
@@ -411,6 +412,11 @@ def convert_to_json_serializable(  # noqa: C901, PLR0911, PLR0912
         # Connection is a module, which is non-serializable. Return module name instead.
         return "sqlalchemy.engine.base.Connection"
 
+    # Provisions for pytest and unittest.mock (to facilitate developing unit tests that utilize Mock objects).
+    if isinstance(data, mock.Mock):
+        return "mock.Mock"
```
cdkini: I don't think we should change production code for the sake of making it work better with a test - instead, we can mock the actual util in the test and scope it accordingly

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,259 @@
...
+            {
+                FULLY_QUALIFIED_PARAMETER_NAME_VALUE_KEY: 1,
+                FULLY_QUALIFIED_PARAMETER_NAME_ATTRIBUTED_VALUE_KEY: {"batch_id_0": 1},
+                FULLY_QUALIFIED_PARAMETER_NAME_METADATA_KEY: {"environment": "test"},
+            }
+        )
+
+
+@pytest.mark.unit
+@mock.patch("test_parameter_builder.DummyParameterBuilder._build_parameters")
```
cdkini: Instead of mocking this private method, can the dummy actual override the definition? If this needs to be different per test, we can parameterize with the constructor

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,259 @@
...
+    assert mock_parameter_builder__build_parameters.call_count == 3
+
+    parameter_node: ParameterNode = get_parameter_value_and_validate_return_type(
+        domain=domain,
+        parameter_reference=my_dependent_parameter_builder.json_serialized_fully_qualified_parameter_name,
+        expected_return_type=None,
+        variables=variables,
+        parameters=parameters,
+    )
+    assert parameter_node == "mock.Mock"
```
cdkini: Circling back from the prior note - I don't think the change to `convert_to_json_serializable` makes sense. This `assert` only tests that we made that change it doesn't actually test user facing behavior

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+@pytest.mark.unit
+@mock.patch("test_parameter_builder.DummyParameterBuilder._build_parameters")
```
cdkini: Same note - why are we mocking if we can override in the dummy definition

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,259 @@
...
+def test_parameter_builder_dependencies_evaluated_in_parameter_builder(
+    mock_parameter_builder__build_parameters: mock.MagicMock,
+):
```
cdkini: These tests still feel quite cumbersome for a "unit" test - if the unit in question in the parameter builder, introducing external dependencies like `Domain` strongly couple this test to other concepts.

I think exceptions can definitely be made (I know the RBP is quite a complex part of our domain) but I'd strongly urge us to simplify our tests and narrow what exactly we're testing if we're deeming them to be "unit"

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    my_evaluation_dependency_0_parameter_builder.build_parameters(
+        domain=domain,
+        variables=variables,
+        parameters=parameters,
+        batch_request=None,
+        runtime_configuration=None,
+    )
+    assert mock_parameter_builder__build_parameters.call_count == 1
```
cdkini: Is this just asserting that the call on L223 was made?

cdkini (changes requested):
I appreciate the cleanup! Left misc comments

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner Actually, if you look at the double underscore -- we call the Dummy but test on the Mock.  In other words, we call the public ParameterBuilder interface method and test on the implementation method in the Mock.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner Please see `https://github.com/great-expectations/great_expectations/pull/8302#discussion_r1262574558` (we call the interface method, but test on the implementation).  Thanks.

great_expectations/rule_based_profiler/data_assistant/column_value_missing_data_assistant.py
alexsherstinsky: @donaldheppner so you want us to enable the new DataAssistant as a separate PR?  Thanks.

great_expectations/core/util.py
alexsherstinsky: @cdkini You are right -- very good point; I disabled automerge and see how this can be solved properly.  Thanks!

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @cdkini We can absolutely do that -- and I had that solution, but the nI could not use the "Mock.call_count" -- can you suggest working around that?  In other words, we have to test that the implementation method was called.  Otherwise, let us leave the use of the Mock.  Please let me know what you think.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @cdkini If we replace the util with a Mock, this will no longer be relevant.  We do not need to test this return value.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @cdkini The only reason is to be able to know that it was called.  Or do you suggest we put the logic for how many times it was called into the Dummy itself?  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @cdkini We should leave it as is, because the Domain here is a dummy -- and is needed to run the ParameterBuilder.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @cdkini Yes -- specifically, it is asserting that the call to the implementation method `_build_parameters()` as made by calling the public interface method `build_parameters()`.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,259 @@
...
+        Returns:
+            Attributes object, containing computed parameter values and parameter computation details metadata.
+        """
+        return Attributes(
+            {
+                FULLY_QUALIFIED_PARAMETER_NAME_VALUE_KEY: 1,
+                FULLY_QUALIFIED_PARAMETER_NAME_ATTRIBUTED_VALUE_KEY: {"batch_id_0": 1},
+                FULLY_QUALIFIED_PARAMETER_NAME_METADATA_KEY: {"environment": "test"},
+            }
+        )
```
cdkini:
```suggestion
    def _build_parameters(
        self,
        domain: Domain,
        variables: Optional[ParameterContainer] = None,
        parameters: Optional[Dict[str, ParameterContainer]] = None,
        runtime_configuration: Optional[dict] = None,
    ) -> Attributes:
        """
        Builds ParameterContainer object that holds ParameterNode objects with attribute name-value pairs and details.

        Returns:
            Attributes object, containing computed parameter values and parameter computation details metadata.
        """
        self.call_count += 1
        return Attributes(
            {
                FULLY_QUALIFIED_PARAMETER_NAME_VALUE_KEY: 1,
                FULLY_QUALIFIED_PARAMETER_NAME_ATTRIBUTED_VALUE_KEY: {"batch_id_0": 1},
                FULLY_QUALIFIED_PARAMETER_NAME_METADATA_KEY: {"environment": "test"},
            }
        )
```

cdkini: set the call count in the constructor and then check in the test (removes the need for a mock)

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
cdkini: I don't really know if testing that a private method is called by a public interface is really providing much value - feels like we're testing an implementation detail instead of the behavior.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
donaldheppner: can you please clarify this test? Testing the mock also isn't a valid test. What non-mock, non-dummy code is being tested?

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
cdkini: The latter!

great_expectations/rule_based_profiler/data_assistant/column_value_missing_data_assistant.py
donaldheppner: yes, I don't see this change related to or dependent on test improvements

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
cdkini: Is the Domain a dummy? It looks like its the actual `Domain` object used by users - therefore, it adds more coupling to this test

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
cdkini: But is that worth all the setup? I'm not sure if just saying something is called is sufficient to gain adequate test coverage here - can we assert against the output and make sure it is of the form/shape we desire?

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner We are testing that an implementation method is called -- we do this against Mocks in various places (please see for example `tests/rule_based_profiler/test_rule_based_profiler.py` line 964.  I will add a docstring in the next commit.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @cdkini We have to test that the implementation method in the ParameterBuilder, listed as dependency, is called.  In the interface method, there is actually logic determining how and when to call the implementation method.  This is located around line 168 in `great_expectations/rule_based_profiler/parameter_builder/parameter_builder.py`.  Thanks.

great_expectations/rule_based_profiler/data_assistant/column_value_missing_data_assistant.py
alexsherstinsky: @donaldheppner Good point.  Will affect at next commit.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @cdkini It is playing the role of a Dummy -- it is the simplest nearly-empty Table Domain Type.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @cdkini What we can do is attempt to create a truly `Dummy` Domain and Parameters objects -- but the real test is that the implementation method `._build_parameters` is called, and in the production code it requires those arguments.  We must get the system to be able to execute this implementation method and us to confirm it.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @cdkini This looks promising, and I will try it (in the next commit).  My main question now is about the utility method, if you have an idea for that.  Thanks!

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @cdkini I created an actual `DummyDomain` class to underscore this point.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @cdkini This worked (I had to add some exclusion provisions to comply with the way fields are managed in `Builder` objects.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    my_evaluation_dependency_0_parameter_builder.build_parameters(
+        domain=domain,
+        variables=None,
+        parameters=parameters,
+        batch_request=None,
+        runtime_configuration=None,
+    )
+    assert my_evaluation_dependency_0_parameter_builder.call_count == 1
```
TrangPham, line 223: Should these assertions all go after the build_parameters() calls, in order to ensure calling `my_dependent_parameter_builder` does not recompute the evaluation param builders?

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,269 @@
...
+        )
+
+
+# noinspection PyUnresolvedReferences,PyUnusedLocal
+@pytest.mark.unit
+@mock.patch(
+    "great_expectations.rule_based_profiler.parameter_builder.parameter_builder.convert_to_json_serializable",
+    return_value="my_json_string",
+)
+def test_parameter_builder_dependencies_evaluated_separately(
```
TrangPham:
```suggestion
def test_parameter_builder_should_not_recompute_evaluation_parameter_builders_if_precomputed(
```

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
cdkini, line 223: I'm still not sure this actually provides much value - asserting that we call a private method doesn't feel like a test that'll prevent errors from sneaking by. What is the actual behavior we want to test? We should be asserting that the built parameters are accurate I think

alexsherstinsky: @cdkini To me, this test just has to insure that the dependencies are processed, and calling that implementation method in those dependency ParameterBuilders is necessary and sufficient.  I added a docstring to that effect.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,269 @@
...
+    assert my_dependent_parameter_builder.call_count == 1
+
+
+# noinspection PyUnresolvedReferences,PyUnusedLocal
+@pytest.mark.unit
+@mock.patch(
+    "great_expectations.rule_based_profiler.parameter_builder.parameter_builder.convert_to_json_serializable",
+    return_value="my_json_string",
+)
+def test_parameter_builder_dependencies_evaluated_in_parameter_builder(
```
TrangPham:
```suggestion
def test_parameter_builder_dependencies_evaluated_in_parameter_builder_if_not_precomuted(
```

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,269 @@
...
+    assert my_dependent_parameter_builder.call_count == 1
+
+
+# noinspection PyUnresolvedReferences,PyUnusedLocal
+@pytest.mark.unit
+@mock.patch(
+    "great_expectations.rule_based_profiler.parameter_builder.parameter_builder.convert_to_json_serializable",
+    return_value="my_json_string",
+)
+def test_parameter_builder_dependencies_evaluated_mixed(
```
TrangPham:
```suggestion
def test_parameter_builder_should_only_evalute_dependencies_that_arent_precomputed(
```

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @TrangPham It is designed exactly for the `my_dependent_parameter_builder` to run -- and that run causes the independent ones (in other words, dependencies) to be executed.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @TrangPham The fact that we assert on exactly 1 call is made -- and after the `my_dependent_parameter_builder` one runs.  Happy to discuss.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @TrangPham I am accepting these method names suggestions.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: I am accepting this suggestion.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
TrangPham: Adding testing specific code to our production code is not a good practice.

@cdkini Could you clarify the benefits of avoiding using mocks?

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,269 @@
...
+    assert my_evaluation_dependency_1_parameter_builder.call_count == 1
+
+    my_dependent_parameter_builder.build_parameters(
+        domain=domain,
+        variables=None,
+        parameters=parameters,
+        batch_request=None,
+        runtime_configuration=None,
+    )
+    assert my_dependent_parameter_builder.call_count == 1
```
TrangPham, line 273: Perhaps I should rephrase my previous comment on this test as:
```suggestion
    assert my_dependent_parameter_builder.call_count == 1
    
    # Assert that precomuted evaluation parameters are not recomputed - call count should remain 1
    assert my_evaluation_dependency_0_parameter_builder.call_count == 1
    assert my_evaluation_dependency_1_parameter_builder.call_count == 1
```

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
TrangPham: Ah @cdkini I see what you mean now - I will make another suggestion on the tests for this to not require the Dummy classes

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,269 @@
...
+        self._call_count += 1
+
+        return Attributes(
+            {
+                FULLY_QUALIFIED_PARAMETER_NAME_VALUE_KEY: 1,
+                FULLY_QUALIFIED_PARAMETER_NAME_ATTRIBUTED_VALUE_KEY: {"batch_id_0": 1},
+                FULLY_QUALIFIED_PARAMETER_NAME_METADATA_KEY: {"environment": "test"},
+            }
+        )
+
```
TrangPham:
```suggestion
```

TrangPham: Instead of this, would patch object on function `__build_parameters` for the specific ParameterBuilder used in test cases work?

ie `with patch.object(ProductionClass, 'method', return_value=Attributes(...)) as mock_method:` 
documentation: https://docs.python.org/3/library/unittest.mock.html

alexsherstinsky: @TrangPham About half a year ago, the team has adopted the approach of actually creating fake and dummy objects.  Also, the "patch.object" approach has a nicer form (without the "object").  I am actually using its method decorator form (they are equivalent, and in the current situation, this "with" would encapsulate most of the logic of the method, so I just put it at the top for cleanliness).  Happy to discuss.  Thank you.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @TrangPham and @cdkini -- just making sure that you saw that we do not modify the production code; instead, I mocked the output of that utility as follows:
```
@mock.patch(
    "great_expectations.rule_based_profiler.parameter_builder.parameter_builder.convert_to_json_serializable",
    return_value="my_json_string",
)
```
and it works.  Thanks.

TrangPham (approved):
Approving given that Alex mentioned that the Dummy classes were added to address your previous comments @cdkini. Chetan please let me know if your comments are sufficiently addressed.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,486 @@
...
+"""
+Tests in this module focus on behavior aspects of "ParameterBuilder.build_parameters()" -- this public method assesses
+dependencies on evaluation "ParameterBuilder" objects (if any specified), and resolves these dependencies, prior to
+calling its own interface method, "ParameterBuilder.build_parameters()".  One key method that is under test is
+"ParameterBuilder.resolve_evaluation_dependencies()" in terms of how it affects contents of shared memory object --
+"parameters: Dict[str, ParameterContainer]", whose "ParameterContainer" stores computation results of every
+"ParameterBuilder", evaluated within scope of given "Domain" object.  When one "ParameterBuilder" specifies dependencies
+on other "ParameterBuilder" objects for its evaluation, then "ParameterBuilder.resolve_evaluation_dependencies()" will
+process these dependencies recursively.  Hence, bebefore/after tests are provided for behavior of this method.  Utility
+method "get_fully_qualified_parameter_names()", whose thoroughly tests are elsewhere, is used to query shared memory.
```
donaldheppner: is there an opportunity now to move tests for other parts of parameter builder into unit tests here? Tests for a single module should not be elsewhere

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,486 @@
...
+    my_evaluation_dependency_1_parameter_builder.build_parameters(
+        domain=domain,
+        variables=None,
+        parameters=parameters,
+        batch_request=None,
+        runtime_configuration=None,
+    )
+    assert my_evaluation_dependency_1_parameter_builder.call_count == 1
+
+    dependencies_fully_qualified_parameter_names: List[
```
donaldheppner, line 239: what can we assert here? Why is this setup step needed?

Specifically, I should know what `get_fully_qualified_parameter_names` is expected to return given this setup, and create a specific unit test for this method.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,486 @@
...
+        str
+    ] = get_fully_qualified_parameter_names(
+        domain=domain,
+        variables=None,
+        parameters=parameters,
+    )
+
+    my_dependent_parameter_builder.resolve_evaluation_dependencies(
```
donaldheppner: what does this do and why is it needed as part of this test's setup? Specifically, there are assertions below and I'm not sure how or if they relate to this method, and there are no assertions here to either explain or validate the behavior of this method

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,486 @@
...
+        domain=domain,
+        variables=None,
+        parameters=parameters,
+    )
+
+    """
+    These assertions show that both "ParameterBuilder" dependencies computed their values, but dependent
+    "ParameterBuilder" has not yet computed its value.
+    """
+    assert (
```
donaldheppner, line 251: I'd consider breaking this into two assertions and avoiding the use of and or or in these statements

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,486 @@
...
+        and my_evaluation_dependency_0_parameter_builder.json_serialized_fully_qualified_parameter_name
+        in dependencies_fully_qualified_parameter_names
+    )
+    assert (
+        my_evaluation_dependency_1_parameter_builder.raw_fully_qualified_parameter_name
+        in dependencies_fully_qualified_parameter_names
+        and my_evaluation_dependency_1_parameter_builder.json_serialized_fully_qualified_parameter_name
+        in dependencies_fully_qualified_parameter_names
+    )
+    assert (
```
donaldheppner: This seems to assert specifically that something did not happen. Consider at least moving this assertion closer to the `build_parameters` invocation on L209, and following with the positive assert. This would more easily as:

1. There is no value
2. I did something
3. There is now a value

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,486 @@
...
+        not in dependencies_fully_qualified_parameter_names
+        or my_dependent_parameter_builder.json_serialized_fully_qualified_parameter_name
+        not in dependencies_fully_qualified_parameter_names
+    )
+    """
+    No "ParameterBuilder" dependencies need to be executed, because None are specified (hence,
+    "my_dependent_parameter_builder.resolve_evaluation_dependencies()" has no effect), but their values exist, because
+    they were executed individually previously.  Thus, set of computed fully_qualified_parameter_names does not change.
+    """
+    assert sorted(all_fully_qualified_parameter_names) == sorted(
```
donaldheppner: what action did we take to make this true? Are we really testing `resolve_evaluation_dependencies` here?

Furthermore, if that _is_ what we are testing here, then I'd consider moving it to a separate test. When a unit test fails I want to know exactly where to look.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,486 @@
...
+        dependencies_fully_qualified_parameter_names
+    )
+
+    my_dependent_parameter_builder.build_parameters(
```
donaldheppner, line 266: I understand this to be the method under test: assertions should immediately follow its invocation to ensure the proper behavior. What can we assert after calling `build_parameters` that doesn't require me to invoke other methods?

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,486 @@
...
+    assert my_evaluation_dependency_0_parameter_builder.call_count == 1
+
+    my_evaluation_dependency_1_parameter_builder.build_parameters(
+        domain=domain,
+        variables=None,
+        parameters=parameters,
+        batch_request=None,
+        runtime_configuration=None,
+    )
+    assert my_evaluation_dependency_1_parameter_builder.call_count == 1
```
donaldheppner, line 232: There must be more of an impact of invoke `build_parameters` than asserting that `build_parameters` was called once. What can we assert here knowing this is the method under test?

donaldheppner (changes requested):

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,486 @@
...
+    )
+
+    """
+    This assertion shows that because namespace of fully_qualified_parameter_names is empty (as previously asserted),
+    and because "ParameterBuilder" dependencies were specified as argument to instantiation of dependent
+    "ParameterBuilder", both "ParameterBuilder" dependencies computed their values (this was accomplished by
+    "my_dependent_parameter_builder.resolve_evaluation_dependencies()" which ran "_build_parameters()" of each
+    dependency "ParameterBuilder" object).  However, dependent "ParameterBuilder" has not yet computed its value (as
+    name "my_dependent_parameter_builder" is not in list), since it has yet been called.
+    """
```
donaldheppner: Could this more plainly be stated as "ensuring that test setup was successful?"

Where are the tests for resolve_evaluation_dependencies? These kinds of explanations, if required, belong there. I'd consider breaking this into its own test and potential local fixture rather than copy/pasting these `resolve_evaluation_dependencies` tests into each of the parameter builder tests

donaldheppner (changes requested):

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @TrangPham I added these assertions as well (thanks @donaldheppner -- sorry I missed it before).  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner As I mentioned in the introductory docstring, `get_fully_qualified_parameter_names()` has been thoroughly unit tested (in "tests/rule_based_profiler/parameter_builder/test_parameter_container.py").  Here, it is used as a fast utility so that the actual contents of the shared memory are tracked truthfully as the ParameterBuilders are executed.  We can assert on what these are:
```
    assert dependencies_fully_qualified_parameter_names == [
        "$parameter.raw.my_evaluation_dependency_parameter_name_1",
        "$parameter.raw.my_evaluation_dependency_parameter_name_0",
        "$parameter.my_evaluation_dependency_parameter_name_1",
        "$parameter.my_evaluation_dependency_parameter_name_0",
    ]
```
but I think that we do not want to put this assertion into this test module, because it has already been done elsewhere, as I mentioned, so it would be redundant; instead, we use this fast utility to track the contents of the shared memory.  

Please notice that when we do have a different situation in the next test (`test_parameter_builder_dependencies_evaluated_in_parameter_builder_if_not_precomputed()`), where we do test exactly for the resolution (lines 303 and 327) -- and then we must assert against the known expected list of names.

Please let me know if this clarifies the situation and/or if you have further suggestions.  Thank you.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner This tests one of the behaviors: resolution of prerequisite ParameterBuilders.  The need for the setup stems from the fact that all ParameterBuilder computations output result into shared memory.  Resolution of prerequisite ParameterBuilders (via `ParameterBuilder.resolve_evaluation_dependencies()`) involves ParameterBuilder computation -- it runs `ParameterBuilder.build_parameters()` for all prerequisite ParameterBuilders, thereby all updating the same shared memory space.  Hence, this set up is for the behavior that is key part of the overall ParameterBuilder behavior.  It is tested as part of the `ParameterBuilder.build_parameters()` because it is part of it, and itself involves execution of `ParameterBuilder.build_parameters()`.  The assertions are as follows: On line 178, we assert that the progress of filling the shared memory from resolving the prerequisite ParameterBuilders is as expected.  In the next test, we do the same (but for a different expected result) on line 293.  I have added another docstring in order to help making this more clear to the reader.  Please let me know if this helps.  Thank you.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner I enacted the change per your feedback, and enhanced the docstring as per your comment.  Thank you.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner Indeed here, the fact that `resolve_evaluation_dependencies()` runs inside `build_parameters()` -- please see earlier explanation -- is tested.  I was reluctant in putting it into separate test, because shared memory would need to be set up -- and it looks like we are trying to reduce setup.  Do you have a recommendation?  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner Got it -- I moved relevant invocation closer to assertions so that it is easier to read.  Thank you.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner After recent updates based on your previous comments, the code for testing "heavier" behaviors than "just being called" appears on line 154 and below.  Please let me know if I am answering your inquiry correctly.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner This is more than just set up.  Please see my earlier response in `https://github.com/great-expectations/great_expectations/pull/8302#discussion_r1263799594` about the place of `resolve_evaluation_dependencies()` -- still, your suggestion of testing it once, with its own fixture is very appealing.  Should we go ahead then and refactor as such (even if it requires its own shared memory setup)?  Thank you.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner Remarkably, I do not see any more such situations -- all other `ParameterBuilder` test modules focuses on their own functionality.  Thank you.

cdkini (approved):

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
donaldheppner: should this comment be updated then?

> Utility method "get_fully_qualified_parameter_names()", whose thoroughly tests are elsewhere, is used to query shared memory.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,485 @@
...
+        return Attributes(
+            {
+                FULLY_QUALIFIED_PARAMETER_NAME_VALUE_KEY: 1,
+                FULLY_QUALIFIED_PARAMETER_NAME_ATTRIBUTED_VALUE_KEY: {"batch_id_0": 1},
+                FULLY_QUALIFIED_PARAMETER_NAME_METADATA_KEY: {"environment": "test"},
+            }
+        )
+
+
+#
```
donaldheppner: this commented out code can be removed

donaldheppner (changes requested):

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,485 @@
...
+    )
+
+    """
+    This assertion shows that because "ParameterBuilder" dependencies were specified as argument to instantiation of
+    dependent "ParameterBuilder", both "ParameterBuilder" dependencies computed their values (this was accomplished by
+    "my_dependent_parameter_builder.resolve_evaluation_dependencies()", which ran "build_parameters()" method of each
+    dependency "ParameterBuilder" object).  Consequently, shared memory namespace became filled with dependency results.
+    """
+    assert all_fully_qualified_parameter_names == [
+        "$parameter.raw.my_evaluation_dependency_parameter_name_1",
```
donaldheppner: Are there referenceable constants we should also reference here, like `parameter`, `raw`, `$`?

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,485 @@
...
+
+    """
+    These assertions show that dependent "ParameterBuilder" computed its values.
+    """
+    assert (
+        my_dependent_parameter_builder.raw_fully_qualified_parameter_name
+        in all_fully_qualified_parameter_names
+        or my_dependent_parameter_builder.json_serialized_fully_qualified_parameter_name
+        in all_fully_qualified_parameter_names
+    )
```
donaldheppner, line 295: Why is there an `or` here? Don't we know where the parameters are going to go?

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,485 @@
...
+
+
+# noinspection PyUnresolvedReferences,PyUnusedLocal
+@pytest.mark.unit
+@mock.patch(
+    "great_expectations.rule_based_profiler.parameter_builder.parameter_builder.convert_to_json_serializable",
+    return_value="my_json_string",
+)
+def test_parameter_builder_dependencies_evaluated_in_parameter_builder_if_not_precomputed(
+    mock_convert_to_json_serializable: mock.MagicMock,
```
donaldheppner, line 305: Why is this mock required? It's not used

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
```diff
@@ -0,0 +1,485 @@
...
+
+
+# noinspection PyUnresolvedReferences,PyUnusedLocal
+@pytest.mark.unit
+@mock.patch(
+    "great_expectations.rule_based_profiler.parameter_builder.parameter_builder.convert_to_json_serializable",
+    return_value="my_json_string",
+)
+def test_parameter_builder_should_only_evaluate_dependencies_that_are_not_precomputed(
+    mock_convert_to_json_serializable: mock.MagicMock,
```
donaldheppner, line 378: why is this mock required if it's not used?

donaldheppner (changes requested):

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner Thank you -- this was removed (in next commit).

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner It is used internally by the methods called.  We have such situations throughout GX OSS tests.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: Please see above `https://github.com/great-expectations/great_expectations/pull/8302#discussion_r1265546905` -- happy to describe and see if this is normal or symptomatic of an issue.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner We could do this, but I do not think that we should.  My reasoning is as follows: in the implementation, we use those constants in order for the code to follow any possible changes in the syntax.  However, in tests, we want "to be literal" so that if something is wrong with the implementation, the tests will catch it.  Having said that, if you have a good argument for generalizing these strings, I would be happy to make the change.  Thanks.

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
alexsherstinsky: @donaldheppner Fantastic catch -- this should be `AND`!  Thank you VERY much (fixed in next commit).

tests/rule_based_profiler/parameter_builder/test_parameter_builder.py
donaldheppner: The value in referencing the constants is that the test becomes less brittle, and I can't think of a scenario where referencing constants here would fix an implementation error.

alexsherstinsky: @donaldheppner Change made (committed) -- please re-review.  Thank you!

donaldheppner (approved):

PR #8322, opened 2023-07-14
[MAINTENANCE] add `invoke ci-tests` and  `invoke service`

Kilo59:
This PR adds 2 new invoke tasks `ci-tests` & `service`.

The remaining work around these tasks involves defining and registering `docker-compose` services & `setup_funcs` that our various markers require.

## `invoke service`

```console
$ invoke --help service
Usage: inv[oke] [--core-opts] service [--options] [other tasks here ...]

Docstring:
  Startup a service, by referencing its name directly or by looking up a pytest marker.

  If a marker is specified, the services listed in `MARKER_DEPENDENDENCY_MAP` will be used.

  Note:
      The main reason this is a separate task is to make it easy to start services
      when running tests locally.

Options:
  -m, --markers   pytest markers to lookup and start services for
  -n, --names     Names of the (docker-compose) services to start
  ```

## `invoke ci-tests`
```console
$ invoke --help ci-tests
Usage: inv[oke] [--core-opts] ci-tests [--options] [other tasks here ...]

Docstring:
  Run tests in CI.

  This method looks up the pytest marker provided and runs the tests for that marker,
  as well as looking up any required services, testing dependencies and extra CLI flags
  that are need and starting them if `up_services` is True.

  Example:
      If the provided marker is 'spark or postgres and not trino', the setup needed
      for both `postgres` and `spark` tokens will be run.

  `up_services` is False by default to avoid starting services which may already be up
  when running tests locally.

  Defined this as a new invoke task to avoid some of the baggage of our old test setup.

Options:
  -m STRING, --marker=STRING   pytest marker string to run tests for.
  -r, --reports
  -u, --up-services
  -v, --verbose
```


reqs/requirements-dev-contrib.txt
```diff
@@ -3,4 +3,5 @@ black[jupyter]==23.3.0
 invoke>=2.0.0
 mypy==1.3.0
 pre-commit>=2.21.0
-ruff==0.0.271
\ No newline at end of file
+ruff==0.0.271
+tomli>=2.0.1
```
Kilo59, line 7: Add `tomli` library as a dev dependency

tasks.py
```diff
@@ -782,14 +785,74 @@ def show_automerges(ctx: Context):
...
+        ("reqs/requirements-dev-postgresql.txt",),
+        # example of using a simple service lookup and a setup function to do the same thing
+        # we can remove one of these methods if one seems obviously better
+        services=("postgres",),
+        setup_funcs=(
+            lambda ctx: print("Dummy setup function"),
+            # lambda ctx: ctx.run("docker-compose up -d postgres"),
+        ),
+        exta_pytest_args=("--postgresql",),
+    ),
```
Kilo59: I believe I've left this PR in a state that only requires changes to the new `docker-compose.yaml` file and this mapping.

Hopefully, the examples make how to add new services and/or `setup_funcs` clear.

No changes are needed for the `invoke task` logic or the github actions CI (apart from adding new markers to the matrix).

As of the time of this posting the `postgres` test fails because the postgres test clients aren't able to connect to the service (I'm sure @nicgrayson or @billdirks  could resolve that without too much trouble). 
However, if we just want to get this merged all that is required is to comment out the `postgres` entry from the test matrix.

.github/workflows/ci.yml
```diff
@@ -65,9 +65,12 @@ jobs:
         markers:
           - big
           - cli
-          - cloud
+          # `not e2e` because 1 test & 1 fixture need access token
+          - cloud and not e2e
           - filesystem
           - openpyxl or pyarrow or project or sqlite
+          # - spark
+          - postgres
```
Kilo59:
```suggestion
          # - postgres
```
Commenting this out will remove the failing potgres tests and allow this PR to merge.

.github/workflows/ci.yml
```diff
@@ -65,9 +65,12 @@ jobs:
         markers:
           - big
           - cli
-          - cloud
+          # `not e2e` because 1 test & 1 fixture need access token
+          - cloud and not e2e
```
Kilo59, line 78: The cloud test step was always passing because everything was always being skipped due to our custom flagging system 🙃 ...
Now that we are using the `invoke ci-tests` command, it ensures the `--cloud` flag is provided and so this now runs and fails unless `not e2e` is added.

tasks.py
```diff
@@ -387,6 +388,8 @@ def tests(  # noqa: PLR0913
     Run tests. Runs unit tests by default.
 
     Use `invoke tests -p=<TARGET_PACKAGE>` to run tests on a particular package and measure coverage (or lack thereof).
+
+    See also, the newer `invoke ci-tests --help`.
```
Kilo59, line 392:
```suggestion
    See also, the newer `invoke --help ci-tests`.
```

Kilo59:
Need to either move the `docker-compose.yaml` or update the `custom_checks check_repo_root_size` azure test.

reqs/requirements-dev-contrib.txt
Shinnnyshinshin: For reference: [tomli](https://pypi.org/project/tomli/) is a TOML parsing package

pyproject.toml
```diff
@@ -547,7 +547,7 @@ markers = [
     "clickhouse: mark a test as Clickhouse-dependent.",
     "athena: mark a test as AWS Athena-dependent.",
     "spark: mark a test as Spark-dependent.",
-    "postgres: mark a test as postgres-dependent.",
+    "postgresql: mark a test as postgresql-dependent.",
```
Shinnnyshinshin, line 550: This is to make the naming consistent between folders and files

.github/workflows/ci.yml
```diff
@@ -87,8 +99,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install $(grep -E '^(invoke)' reqs/requirements-dev-contrib.txt)
+          pip install -c constraints-dev.txt ".[test]"
           invoke deps --gx-install -m '${{ matrix.markers }}' -r test
```
Kilo59:
```suggestion
          pip install $(grep -E '^(invoke)' reqs/requirements-dev-contrib.txt)
          invoke deps --gx-install -m '${{ matrix.markers }}' -r test
```
I would revert this change. It defeats the purpose of `invoke deps`, invoke `deps` uses the constraints file by default when installing dependencies.

https://github.com/great-expectations/great_expectations/blob/f51c946b14403352ee4e2573b4b8906859c2fc50/tasks.py#L848-L849

If you need to add an additional dependency as part of this you can add it the `MARKER_DEPENDENDENCY_MAP` in `tasks.py`.

```python
    "postgresql": TestDependencies(
        ("reqs/requirements-dev-postgresql.txt",
         "reqs/requirements-dev-my_other_req_file.txt"), # 👈 
        setup_funcs=(
            lambda ctx: ctx.run(
                "docker-compose -f assets/docker/postgresql/docker-compose.yml up -d"
            ),
        ),
        exta_pytest_args=("--postgresql",),
```

Shinnnyshinshin: @Kilo59 I was seeing this error before the change

```
##[debug]'
##[debug]Loading env
Run pip install $(grep -E '^(invoke)' reqs/requirements-dev-test.txt)
##[debug]/usr/bin/bash -e /home/runner/work/_temp/c52f66bd-5158-456d-930b-fd0f75576e[40](https://github.com/great-expectations/great_expectations/actions/runs/5602528262/jobs/10247904502#step:5:41).sh
ERROR: You must give at least one requirement to install (see "pip help install")
```

Shinnnyshinshin: updating now. There may still be some errors that need to be resolved with the install dependencies stage

Shinnnyshinshin: it worked! Thank you @Kilo59

tests/agent/message_service/test_subscriber.py
```diff
@@ -11,7 +11,7 @@
 )
 from tests.agent.message_service.amqp_errors import get_amqp_errors
 
-pytestmark = pytest.mark.unit
+pytestmark = [pytest.mark.unit, pytest.mark.cloud]
```
billdirks: These labels should be mutually exclusive so tests will only one in 1 stage. This happens below also.

tests/test_markers.py
```diff
@@ -0,0 +1,69 @@
...
+from tasks import MARKER_DEPENDENDENCY_MAP
+
+from great_expectations.core.yaml_handler import YAMLHandler
+
+pytestmarks = [pytest.mark.project]
+
+LOGGER: Final = logging.getLogger(__name__)
+PROJECT_ROOT: Final = pathlib.Path(__file__).parent.parent
+PYPROJECT_TOML: Final = PROJECT_ROOT / "pyproject.toml"
+DOCKER_COMPOSE_YAML: Final = PROJECT_ROOT / "docker-compose.yaml"
```
billdirks: I'm not sure we want this at the root level. Maybe inside of `tests`?

tests/test_markers.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    print(f"pytest_markers:\n----------\n{pf(pytest_markers)}")
+
+    for marker in MARKER_DEPENDENDENCY_MAP:
+        assert marker in pytest_markers
+
+
+@pytest.fixture(scope="module")
+def docker_compose_services() -> dict[str, dict]:
+    assert DOCKER_COMPOSE_YAML.exists()
```
billdirks: I don't see this yaml file. Can you verify that these tests actually run? I don't see them in this stage:
https://github.com/great-expectations/great_expectations/actions/runs/5603307187/jobs/10249741129?pr=8322

tasks.py
```diff
@@ -782,14 +785,72 @@ def show_automerges(ctx: Context):
...
-    "pyarrow": ("reqs/requirements-dev-arrow.txt",),
-    "external_sqldialect": ("reqs/requirements-dev-sqlalchemy.txt",),
+class TestDependencies(NamedTuple):
+    requirement_files: tuple[str, ...]
+    services: tuple[str, ...] = tuple()
+    setup_funcs: tuple[Callable[[Context], Result | None], ...] = tuple()
+    exta_pytest_args: tuple[  # TODO: remove this once remove the custom flagging system
+        str, ...
+    ] = tuple()
+    # NOTE: We probably don't need both `services` and `setup_funcs`
```
billdirks: Could we get rid of one of these? `services` looks simpler and we can do service specific things in whatever function is associated with that service.

billdirks: I am fine with either for now.

tasks.py
```diff
@@ -782,14 +785,72 @@ def show_automerges(ctx: Context):
...
+    ),
+    "spark": TestDependencies(
+        ("reqs/requirements-dev-spark.txt",),
+        # services=("spark",),
+        exta_pytest_args=("--spark",),
+    ),
 }
 
 
+# TODO: return list & lru_cache this??
```
billdirks: Remove this TODO.

tasks.py
```diff
@@ -782,14 +785,72 @@ def show_automerges(ctx: Context):
...
+def _tokenize_marker_string(marker_string: str) -> Generator[str, None, None]:
+    """
+    Split a marker string on ' or ' and ' and ' and return an iterable of tokens.
+    Exclude parts of the string after ' and not '.
+    """
+    remaining_str, _, _ = marker_string.partition(" and not ")
```
billdirks: This feels fragile to me. We're basically building a parser for pytests marker syntax and we've made assumptions that will bite use later if we think we're using the full parser. Right now this only supports one `not`, it must be at the end, and it must be preceded by another marker. 

This makes it coupled to how we call tests in the `ci.yml` file even though it looks generic.

Alternatives:
* Use the parser from pytest. I looked at this a little bit and, at first glance, it looks like the marker string parsing is private. We could dig a little more.
* Explicitly handle any markers that have more than 1 token in them. That is, have explicit cases for `openpyxl or pyarrow or project or sqlite` and `cloud and not e2e`. With this we can throw an clear error message if we try to run this with a combo we haven't seen before.

tasks.py
```diff
@@ -849,3 +908,88 @@ def deps(  # noqa: PLR0913
...
+        pytest_cmds.append("-vv")
+
+    for test_deps in _get_marker_dependencies(marker):
+        for extra_pytest_arg in test_deps.exta_pytest_args:
+            pytest_cmds.append(extra_pytest_arg)
+
+        for setup_func in test_deps.setup_funcs:
+            setup_func(ctx)
+
+        if test_deps.services and up_services:
```
billdirks: If we remove the `services` flag we can remove this check and the function below. It looks like we don't use this now?

billdirks (commented):
How we parse the `-m` flags in this PR makes me uneasy. See my comment inline.

tests/agent/message_service/test_subscriber.py
Shinnnyshinshin: done

tests/test_markers.py
Shinnnyshinshin: removed unhelpful test

tasks.py
billdirks: The function below is a task which we use.

tasks.py
Shinnnyshinshin: I'm totally with you. I think the pytest parser would be the "correct" way to go, and to "hardcode" for now.

tests/data_context/store/test_database_store_backend.py
```diff
@@ -11,7 +11,7 @@
 pytestmark = pytest.mark.sqlalchemy_version_compatibility
 
 # module level markers
-pytestmark = [pytest.mark.postgres]
+pytestmark = [pytest.mark.postgresql]
```
billdirks: nit: we can remove brackets here. This is not blocking.

tests/test_markers.py
```diff
@@ -0,0 +1,48 @@
...
+    LOGGER.debug(f"marker_details ->\n{pf(marker_details)}")
+    return [m.split(":")[0] for m in marker_details]
+
+
+def test_marker_mappings_are_registered(pytest_markers: list[str]):
+    """
+    Check that all pytest marker mappings are acutallly valid,
+    and have been registered with pytest.
+    """
+    print(f"pytest_markers:\n----------\n{pf(pytest_markers)}")
```
billdirks: Should this be a LOGGER statement?

tasks.py
```diff
@@ -782,14 +785,89 @@ def show_automerges(ctx: Context):
...
+    elif marker_string == "openpyxl or pyarrow or project or sqlite":
+        yield "openpyxl"
+        yield "pyarrow"
+        yield "project"
+        yield "sqlite"
+    else:
+        raise ValueError(f"Unable to tokenize marker string: {marker_string}")
+
+
+def old_tokenize_marker_string(marker_string: str) -> Generator[str, None, None]:
```
billdirks: We can remove this function.

.github/workflows/ci.yml
```diff
@@ -91,4 +103,4 @@ jobs:
...
 
       - name: Run the tests
-        run: pytest -m "${{ matrix.markers }} and not unit" -vv -rEf
+        run: invoke ci-tests '${{ matrix.markers }} and not unit ' --up-services --verbose
```
billdirks: It looks like we should handle the suffix `and not unit` when we parse in the markers. See my comment below.

tasks.py
```diff
@@ -782,14 +785,89 @@ def show_automerges(ctx: Context):
...
+    #
+
+    remaining_str, _, _ = marker_string.partition(" and not ")
+    for partially_tokenized in remaining_str.split(" and "):
+        for token in partially_tokenized.split(" or "):
+            yield token
+
+
+def _get_marker_dependencies(markers: str | Sequence[str]) -> list[TestDependencies]:
+    print(markers)
```
billdirks: You can remove this stray print and the one on line 861.

tasks.py
```diff
@@ -782,14 +785,89 @@ def show_automerges(ctx: Context):
...
+MARKER_DEPENDENDENCY_MAP: Final[Mapping[str, TestDependencies]] = {
+    "athena": TestDependencies(("reqs/requirements-dev-athena.txt",)),
+    "cloud": TestDependencies(
+        ("reqs/requirements-dev-cloud.txt",), exta_pytest_args=("--cloud",)
+    ),
+    "external_sqldialect": TestDependencies(("reqs/requirements-dev-sqlalchemy.txt",)),
+    "pyarrow": TestDependencies(("reqs/requirements-dev-arrow.txt",)),
+    "postgresql": TestDependencies(
+        ("reqs/requirements-dev-postgresql.txt",),
+        setup_funcs=(
```
billdirks: It seems like all this does is up the service and we could have opted to leave that argument in and remove this one since we want to keep the `service` task. We can leave this for now but we might want to cleanup in a subsequent PR.

tasks.py
```diff
@@ -782,14 +785,89 @@ def show_automerges(ctx: Context):
...
+            lambda ctx: ctx.run(
+                "docker-compose -f assets/docker/postgresql/docker-compose.yml up -d"
+            ),
```
billdirks: We should leverage the `service` task here so we don't up the service differently in these 2 places. 

```suggestion
            lambda ctx: service(ctx, names=["postgresql"]),
```

billdirks (commented):
A few cleanup comments. Thanks for doing this!

billdirks (approved):
Please change the lambda and remove the stray prints and function. Otherwise, LGTM.

tests/data_context/store/test_database_store_backend.py
Shinnnyshinshin: done

tasks.py
Shinnnyshinshin: 😅

tests/test_markers.py
Shinnnyshinshin: thanks for catching. fixed

tasks.py
Shinnnyshinshin: I believe this was the other way around @billdirks. We kept `setup_funcs` and took out `services`

Shinnnyshinshin: Done. Thanks for catching this

tests/test_markers.py
Kilo59: I would say no, this should stay a `print()`.
Mixing logs and prints in a test functions makes it harder to separate the normal application logging from the test logging.
For this reason, the only time I tend to use a logger in tests, is in my test doubles (or inside fixtures). In that case I do find it useful to see those logs in the context of normal application logs.
```
teapot.py:INFO:teapot is empty
faucet_test_double.py:INFO:filling teapot
teapot.py:INFO:teapot is full
```

Shinnnyshinshin:
@billdirks seeing some issues with invoking `service` from test. I'll update before merging

PR #8327, opened 2023-07-17
[BUGFIX] Ensure that Fluent Datasources support database table names in lowercase for Oracle, DB2, and Snowflake

alexsherstinsky:
### Scope
For certain database management systems (DBMSs), such as Oracle, DB2, and Snowflake, standard casing is in all capitals.  However, SQLAlchemy standardizes on all lowercase.  This change fixes issues with Fluent Datasources (FDS), whereby quoting of lowercase table names is provided so that SQLAlchemy can pass them to DBMS and access table properly.

### Remarks
* JIRA: DX-307

- [x] Code is linted

    ```
    black .

    ruff . --fix
    ```


great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -11,6 +11,7 @@
     Literal,
     Optional,
     Protocol,
+    Tuple,
```
cdkini: Nit - don't need to import this we can just use `tuple`

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -883,6 +886,41 @@ def _create_batch_spec_kwargs(self) -> dict[str, Any]:
...
+        SQLAlchemy assumes they are case-sensitive and encloses the names with quotes.  This behavior will cause
+        mismatches agaisnt data dictionary data received from such DBMS, so unless identifier names have been created as
+        truly case sensitive using quotes, e.g., "TestDb", all lowercase names should be used on SQLAlchemy side.
+
+        This method starts with assumption that table_name has correct casing.  If this assumption is incorrect, then
+        hypothesis is made that table_name is at least correct within quotation marks.  To validate this assumption, if
+        "self.table_name" exists (i.e., if inspection finds this name, identical, possibly short of quotation marks), in
+        database schema), then quoted version of supplied "self.table_name" is returned, and this property is updated.
+        """
+        table_exists: bool = inspector.has_table(
```
cdkini: You don't need to annotate a type unless mypy requires you to clear up some ambiguity

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -1038,3 +1076,46 @@ def add_query_asset(
             batch_metadata=batch_metadata or {},
         )
         return self._add_asset(asset)
+
+
+def _verify_table_name_exists_and_get_normalized_typed_name_map(
+    name: str,
+    typed_names: List[str | sqlalchemy.quoted_name],
+    error_message_template: str = 'Error: The table "{table_name:s}" does not exist.',
+    verify_only: bool = False,
```
cdkini: are any callers using these parameters? if not, let's remove them - no need to overcomplicate a util if we can always modify it later to allow for more flexibility

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -1038,3 +1076,46 @@ def add_query_asset(
...
+    Verifies that table name exists.
+
+    Args:
+        name: string-valued name
+        typed_names: Properly typed names (e.g., output of "inspector.get_table_names()" SQLAlchemy method call)
+        verify_only: Perform verification only (do not return normalized typed name map)
+        error_message_template: String template to output error message if name cannot be found in typed_names list
+
+    Returns:
+        Single tuple holding mapping from string-valued name to typed name; None if "verify_only" is set.
```
cdkini: I'd rather we avoid union returns - if the caller wants to ignore the return value, they can do so.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -1038,3 +1076,46 @@ def add_query_asset(
...
+    def _get_normalized_table_name_mapping_if_exists(
+        name: str,
+    ) -> Tuple[str, str | sqlalchemy.quoted_name] | None:
+        typed_name_cursor: str | sqlalchemy.quoted_name
+        for typed_name_cursor in typed_names:
+            if (
+                (type(typed_name_cursor) == str)
+                and (name.casefold() == typed_name_cursor.casefold())
+                or (name == str(typed_name_cursor))
+            ):
```
cdkini: why `type` instead of `isinstance`?

cdkini: also what's our stance on closures? should to be moved outside the function?

tests/datasource/fluent/test_sql_datasources.py
```diff
@@ -1,19 +1,23 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Sequence, Tuple
```
cdkini: Same note - you don't need `Tuple` when you can use `tuple`

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: Fix committed -- thanks.

tests/datasource/fluent/test_sql_datasources.py
```diff
@@ -28,5 +32,114 @@ def test_kwargs_are_passed_to_create_engine(mocker: MockerFixture):
...
+        _verify_table_name_exists_and_get_normalized_typed_name_map(
+            name=name,
+            typed_names=quoted_table_names,
+        )
+        for name in table_names_in_dbms_schema
+    ]
+
+    idx: int
+    normalized_table_name_mapping: Tuple[str, str | sqlalchemy.quoted_name] | None
+    for idx, normalized_table_name_mapping in enumerate(normalized_table_name_mappings):
```
cdkini: Why enumerate when you're not using the index? Same for both tests

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: Fix committed -- thanks.

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: The suggestion has been committed -- thanks.

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: @cdkini this has been cleaned up / committed -- thanks.

tests/datasource/fluent/test_sql_datasources.py
alexsherstinsky: @cdkini Suggestion incorporated/committed -- thanks!

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -883,6 +885,41 @@ def _create_batch_spec_kwargs(self) -> dict[str, Any]:
...
+        self,
+        engine: sqlalchemy.Engine,
+        inspector: sqlalchemy.Inspector,
+    ) -> None:
+        """
+        Oracle, DB2, and Snowflake store all case-insensitive object names in uppercase text.  In contrast, SQLAlchemy
+        considers all lowercase object names to be case-insensitive.  Hence, SQLAlchemy converts object name case during
+        schema-level communication, i.e. during table and index reflection.  If uppercase object names are used,
+        SQLAlchemy assumes they are case-sensitive and encloses the names with quotes.  This behavior will cause
+        mismatches agaisnt data dictionary data received from such DBMS, so unless identifier names have been created as
```
cdkini: typo with "agaisnt"

tests/datasource/fluent/test_sql_datasources.py
alexsherstinsky: @cdkini The index is used in the loop.

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: @cdkini Thanks!  Committed.

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: @cdkini Great question: the `isinstance` did not work for me because `sqlalchemy.quoted_name` extends `str`, so I had to be precise.  Thanks.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -1038,3 +1075,44 @@ def add_query_asset(
             batch_metadata=batch_metadata or {},
         )
         return self._add_asset(asset)
+
+
+def _verify_table_name_exists_and_get_normalized_typed_name_map(
+    name: str,
+    typed_names: List[str | sqlalchemy.quoted_name],
+    error_message_template: str = 'Error: The table "{table_name:s}" does not exist.',
```
cdkini: I don't think this is a necessary parameter - if a caller needs to extend it, we can modify later

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -1038,3 +1075,44 @@ def add_query_asset(
...
+                and (name.casefold() == typed_name_cursor.casefold())
+                or (name == str(typed_name_cursor))
+            ):
+                return name, typed_name_cursor
+
+        return None
+
+    normalized_table_name_mapping: tuple[
+        str, str | sqlalchemy.quoted_name
+    ] | None = _get_normalized_table_name_mapping_if_exists(name=name)
```
cdkini: Don't think we even need the closure - why not just move the logic out of the function and run it here?

tests/datasource/fluent/test_sql_datasources.py
cdkini: I don't see `idx` being used anywhere? Same note with the other test

tests/datasource/fluent/test_sql_datasources.py
```diff
@@ -28,5 +32,114 @@ def test_kwargs_are_passed_to_create_engine(mocker: MockerFixture):
...
+        "scientists",
+        "literary_professionals",
+    ]
+
+    name: str
+
+    quoted_table_names: list[sqlalchemy.quoted_name] = [
+        sqlalchemy.quoted_name(value=name, quote=True)
+        for name in table_names_in_dbms_schema
+    ]
```
cdkini: Instead of relying on `sqlalchemy` to do this, can we just hardcode the list? We're adding an external dependency and now coupling that with our test here

cdkini: If our unit in question in your util method, let's try and isolate that

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: Working on it now.  Thanks, @cdkini

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: @cdkini Which parameter would it be?  The error message template?  Thanks.

tests/datasource/fluent/test_sql_datasources.py
alexsherstinsky: @cdkini I made a change to hardcore the `quoted_table_names` -- however, the only way to do this is through the `sqlalchemy.quoted_name()` method.  Thanks.

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: I assumed that it was the error message template -- committed.

cdkini: yes ty

tests/datasource/fluent/test_sql_datasources.py
cdkini: I still see `Tuple` here and used in type hints

tests/datasource/fluent/test_sql_datasources.py
```diff
@@ -28,5 +32,122 @@ def test_kwargs_are_passed_to_create_engine(mocker: MockerFixture):
...
+        "musicians",
+        "scientists",
+        "literary_professionals",
+    ]
+
+    name: str
+
+    quoted_table_names: list[sqlalchemy.quoted_name] = [
+        sqlalchemy.quoted_name(value="actors", quote=True),
+        sqlalchemy.quoted_name(value="artists", quote=True),
```
cdkini, line 135: Since this relies on sqlalchemy, should we mark it some way? Please discuss with testing tiger team - cc @Shinnnyshinshin @billdirks

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -883,6 +885,41 @@ def _create_batch_spec_kwargs(self) -> dict[str, Any]:
...
+                schema=self.schema_name
+            )
+            normalized_table_name_mapping: tuple[
+                str, str | sqlalchemy.quoted_name
+            ] = _verify_table_name_exists_and_get_normalized_typed_name_map(
+                name=self.table_name,
+                typed_names=typed_names,
+            )
+            if normalized_table_name_mapping:
+                self.table_name = normalized_table_name_mapping[1]
```
billdirks: You introduce a side-effect here. This updates the table name only when `test_connection` is called. I believe this is only called by us when interactively creating a new data asset but not when loading it from config. This feels like the wrong place.

2 Questions:
1. Can we sidestep this by always quoting the name and letting the db backend handle it? If it's case insensitive, it won't matter that sqlalchemy is lowercasing it or not. If it's case sensitive, the user needs to use the correct casing anyways.
2. When do we hit the issue of case insensitivity? Your docstring seems to indicate in the data dictionary. That's a different abstraction than this one and it feels strange to need to know about data dictionaries to update sql datasources. Can we instead normalize in the data dictionary abstraction?

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -1038,3 +1075,39 @@ def add_query_asset(
             batch_metadata=batch_metadata or {},
         )
         return self._add_asset(asset)
+
+
+def _verify_table_name_exists_and_get_normalized_typed_name_map(
```
billdirks: The map data structure is synonymous with a dictionary (see [wikipedi](https://en.wikipedia.org/wiki/Associative_array)). Could you update the name since you are returning a tuple?

tests/datasource/fluent/test_sql_datasources.py
billdirks: If you don't require a specific backend, unit is the appropriate marker. Otherwise `sqlite`, `postgres`, etc.

billdirks (changes requested):

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: @billdirks These are good questions.  I like the idea of 1 above -- if I can make it work, that would be best.  I will try it and let you know what happens.  For 2 above, we hit it whenever the user decides to name their table in any way other than all-uppercase for Snowflake and other DBMS backends mentioned in the title.  The problem is the these DBMS backends and SQLAlchemy have different standards (SQLAlchemy standardizes on lowercase).  Let me see how 1 turns out, and then perhaps we can reason together about number 2.  Thank you.

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: @billdirks Good catch -- updated and committed this one for now.  Thanks.

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: @billdirks For 1, it appears that your idea is working!  Thanks for pointing it out.  I would have to make sure, but if it is indeed the case, if we insert

```
table_name = sqlalchemy.quoted_name(value=self.table_name, quote=True)
```
just above the call to `inspector.has_table()` and use `table_name` instead of `self.table_name` (so we would not modify our instance variable), then this would fit well with our abstraction?  Thanks!

billdirks: This is more in line with what I'm thinking of. I'd rather keep `self.table_name` immutable so some code path does/does not work conditionally on calling this function first.
I wonder if, in the constructor, we could do `self.table_name = sqlalchemy.quoted_name(value=passed_in_table_name, quote=True)` and then we wouldn't have to worry about it here at all?

alexsherstinsky: @billdirks I am happy to do that -- as long as we are OK with the passed in `self.table_name` become quoted (as opposed to what the user originally specified).  I can experiment with it.  I will update you once I know more.  Thanks.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -802,7 +802,7 @@ class TableAsset(_SQLAsset):
 
     # Instance fields
     type: Literal["table"] = "table"
-    table_name: str = pydantic.Field(
+    table_name: Union[str, sqlalchemy.quoted_name] = pydantic.Field(
```
billdirks: I wonder if this can always be `sqlalchemy.quoted_name`?

billdirks (approved):
Thanks @alexsherstinsky! LGTM. I added one small question inline.

great_expectations/datasource/fluent/sql_datasource.py
alexsherstinsky: @billdirks I am not sure -- it seems to me that the user first provides a string, and then we -- in the validator -- check it, and if it is not `None`, then we quote it.  So the question is: how many times does the field get set?  Twice or once?  Thanks!

alexsherstinsky: @billdirks The verdict is in!  While `mypy` accepts the main sql_datasource.py module either way, the test module shows a type violation in PyCharm, because we submit a string valued table name to the `TableAsset()` constructor.  So I am leaving the `Union` in.  Thanks!

PR #8336, opened 2023-07-17
[FEATURE] Add Agent support for Missingness Data Assistant

TrangPham:
Generalizes the RunOnboardingDataAssitantAction to allow for running any data assistant using 'assitant_name'.

The `RunOnboardingDataAssistantEvent` is kept in order to allow other systems to switch to using the new event `RunDataAssistantEvent`. `RunOnboardingDataAssistantEvent` should be handled until it is removed from other systems.


great_expectations/agent/actions/run_data_assistant.py
```diff
@@ -1,63 +1,69 @@
...
+                "The RunDataAssistant Action can only be used with a fluent-style datasource."
             )
+
         asset = datasource.get_asset(asset_name=event.data_asset_name)
         try:
             batch_request = asset.build_batch_request()
         except ValueError as e:
             raise ValueError(
-                "The RunOnboardingDataAssistant Action cannot be used with an in-memory dataframe asset."
+                "The RunDataAssistant Action cannot be used with an in-memory dataframe asset."
```
alexsherstinsky: @TrangPham Do we want to include -- as part of this error message -- specifically which DataAssistant could not be run?  Thank you.

great_expectations/agent/actions/run_data_assistant.py
```diff
@@ -1,63 +1,69 @@
...
+            ) from e
+
+        try:
+            data_assistant = getattr(self._context.assistants, assistant_name)
```
alexsherstinsky: 💙

great_expectations/agent/actions/run_data_assistant.py
TrangPham: Good idea I will update this!

TrangPham:
@alexsherstinsky can you review this again?

great_expectations/agent/models.py
```diff
@@ -14,13 +14,21 @@ class EventBase(AgentBaseModel):
...
+    # TODO(https://greatexpectations.atlassian.net/browse/DX-652): Deprecate once event is no longer being emitted
     type: Literal[
         "onboarding_data_assistant_request.received"
     ] = "onboarding_data_assistant_request.received"
     datasource_name: str
     data_asset_name: str
 
 
+class RunDataAssistantEvent(EventBase):
+    type: Literal["data_assistant_request.received"] = "data_assistant_request.received"
```
joshua-stauffer: while we're updating this, could we remove the `.received`? We're not using these events as broadly as we imagined we might when we first spec'd this out, so that extra namespace is unnecessary.

joshua-stauffer:
Thanks @TrangPham, after a quick pass this looks good! could you point me to any docs around the decision to extend a single action, instead of implementing an action per Assistant? I'm not sure that would be better, just would like to stay in sync with why we chose this approach. Last context I had on this was [this slack thread](https://greatexpectationslabs.slack.com/archives/C04SG48J95K/p1688671703125749).

great_expectations/agent/models.py
TrangPham: Thanks will update

TrangPham:
> Thanks @TrangPham, after a quick pass this looks good! could you point me to any docs around the decision to extend a single action, instead of implementing an action per Assistant? I'm not sure that would be better, just would like to stay in sync with why we chose this approach. Last context I had on this was [this slack thread](https://greatexpectationslabs.slack.com/archives/C04SG48J95K/p1688671703125749).

Thanks for the question. We implemented in the manor mentioned in the slack thread:
> i think the simplest implementation would be to add a field to the existing action that indicates which assistant should be run, then add some logic to great_expectations/agent/actions/run_onboarding_data_assistant.py to run the correct assistant.

The assistants all take a batch_request as their main argument and thus can operate just from the assistant name, asset name and datasource name. Using those the batch_request is built and sent to the data assistant's run method.

We would be deprecating the RunOnboardingDataAssistantEvent after it is no longer being emitted by mercury

great_expectations/agent/actions/data_assistants/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    event: RunDataAssistantEvent,
+    data_assistant_result: DataAssistantResult,
+) -> ActionResult:
+    # build tz aware timestamp
+    tz = datetime.now().astimezone().tzinfo  # noqa: DTZ005
+    timestamp = datetime.now(tz=tz)
+
+    # ensure we have unique names for created resources
+    expectation_suite_name = f"{event.type} {event.data_asset_name} suite {timestamp}"
+    checkpoint_name = f"{event.type} {event.data_asset_name} checkpoint {timestamp}"
```
joshua-stauffer: Nonblocking, but i notice we're now always adding a timestamp - design has consistently asked us to only add timestamp if we have to ([context in slack](https://greatexpectationslabs.slack.com/archives/C056ZF0SXQB/p1686088272369179?thread_ts=1686086296.685339&cid=C056ZF0SXQB)), which is why the previous behavior did what it did.

joshua-stauffer: i don't have a strong stance on this, so would suggest to follow up with Trang directly if you feel this is important

great_expectations/agent/actions/data_assistants/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    checkpoint = self._context.add_checkpoint(**checkpoint_config)  # type: ignore[arg-type]
+
+    expectation_suite_id = expectation_suite.ge_cloud_id
+    checkpoint_id = checkpoint.ge_cloud_id
+    if expectation_suite_id is None or checkpoint_id is None:
+        raise ValueError("Cloud backed resources must have an ID.")
+    return ActionResult(
+        id=id,
```
joshua-stauffer, line 102: this looks like a bug - where is this `id` coming from? needs to be the `id` passed into the `run` method

great_expectations/agent/actions/data_assistants/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    from great_expectations.core import ExpectationSuite
+    from great_expectations.core.batch import BatchRequest
+    from great_expectations.rule_based_profiler.data_assistant_result.data_assistant_result import (
+        DataAssistantResult,
+    )
+    from great_expectations.data_context.data_context import CloudDataContext
+
+
+def build_batch_request(
+    context: CloudDataContext,
```
joshua-stauffer, line 24: Are you able to run this successfully on your end? this is raising an exception for me locally - i don't think it can be in a type checking blog.

joshua-stauffer (changes requested):
Thanks for addressing my feedback, @TrangPham, much appreciated! i would suggest that we organize the contents of `data_assistants/utils` in a class instead of just standalone functions in a module, but we could do that in a follow up PR if necessary. Looks like there's a bug or two, so blocking until that's cleaned up.

great_expectations/agent/actions/data_assistants/utils.py
TrangPham: Okay I'll update this to use the previous logic

great_expectations/agent/actions/data_assistants/utils.py
TrangPham: good catch, will update

TrangPham:
> Thanks for addressing my feedback, @TrangPham, much appreciated! i would suggest that we organize the contents of `data_assistants/utils` in a class instead of just standalone functions in a module, but we could do that in a follow up PR if necessary. Looks like there's a bug or two, so blocking until that's cleaned up.

A follow up PR sounds good - the one thing I wasn't sure about was extending a generic with a sub-class that is also a generic

joshua-stauffer (approved):
Thanks @TrangPham! it works and looks great -- the `utils` refactor can come as follow up work later. much appreciated!

PR #8353, opened 2023-07-19
[MAINTENANCE] Fix non-AssertErrors for Expectations during build-gallery process

kenwade4:
- Use the correct import in expect_column_values_to_be_xml_parseable.py to prevent all Expectations in the semantic_types package from failing with `ModuleNotFoundError: No module named 'great_expectations.expectations.metrics.map_metric'` (the semantic_types package imports all semantic_types Expectations)
- Use the same correct import in 2 other experimental Expectations
- Fix the query_result object in the `_validate` method for 2 QueryExpectations that were modified incorrectly in PR #6636
- Rename the new chisquare Expectation in contrib from PR #8314
- Update _check_docstring to also check that short description begins with 'Expect ' and ends with a period
- Update one-line docstrings of Expectations to be compliant


austiezr (approved):
LGTM! 🚀

PR #8362, opened 2023-07-19
[BUGFIX] Scaffolding .gitignore should not require writeable file system

### LINKED ISSUE #8133: GE - AWS MWAA Airflow 2.0.2 integration
**Describe the bug**
We are using GE for validating few reports and trying to move those validations as part of existing data pipeline. We are using 
AWS MWAA 2.0.2 hence it is not possible to use GreatExpectationOperator as it requires Airflow 2.1.0+. Hence using PythonVirtualenvOperator with great-expectations and required python modules as requirements to it. We are getting an error while initiating the data_context by the following command.

data_context: FileDataContext = get_context(context_root_dir=context_root_dir)

The context_root_dir is pointing to the great_expectations project directory in /usr/local/airflow directory. We are getting the below error.

File "/tmp/venvubafuq2v/lib/python3.7/site-packages/great_expectations/data_context/data_context/serializable_data_context.py", line 268, in _scaffold_directories
with open(os.path.join(base_dir, ".gitignore"), "w") as f:  # noqa: PTH118
OSError: [Errno 30] Read-only file system: '/usr/local/airflow/great_expectations/.gitignore'

I verified the great_expectations directory we already have the .gitignore file with uncommitted/ added to it. How can we ignore this step if the .gitignore file already exists in the context_root_dir? Since the MWAA files are mostly read-only how we can integrate GE with it?

**To Reproduce**
Please include your great_expectations.yml config, the code you’re executing that causes the issue, and the full stack trace of any error(s).

**Expected behavior**
While initiating the great_expectations data_context if the .gitignore file already presents then it should not try to write it again.

 - Operating System: Linux 
 - Great Expectations Version: 0.16.13
 - Datasource: Postgres
 - Cloud environment: AWS MWAA

**Additional context**

metzlar:
A small change to make sure scaffolding a new project works on a read-only file system (if the files alreay exist) This should fix https://github.com/great-expectations/great_expectations/issues/8133


metzlar:
@cla-bot check

great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -310,6 +310,19 @@ def _scaffold_directories(cls, base_dir: PathStr) -> None:
...
+    def _scaffold_gitignore(cls, base_dir: PathStr) -> None:
+        """Make sure .gitignore exists and contains uncommitted/"""
+        gitignore = os.path.join(base_dir, ".gitignore")  # noqa: PTH118
+
+        if os.path.isfile(gitignore):  # noqa: PTH113
+            lines = open(gitignore).readlines()
+            if "uncommited/" in lines:
+                return
+
+        with open(gitignore, "a") as f:
```
alexsherstinsky, line 328: Quick question: How can we open the file in the append mode if the filesystem is read-only?  Thank you.

alexsherstinsky (changes requested):
Question inside about ability to append to a file on a non-writeable filesystem.  Thank you.

great_expectations/data_context/data_context/serializable_data_context.py
metzlar: It only does that if it is really necessary, in other words, there is no existing `.gitignore` or there is, but it doesn't contain a line with `uncomitted/`

If it is necessary and the scaffolding is done on a read-only filesystem, the `open()` call will fail with an `OSError` which - I assume - is the correct behaviour.

alexsherstinsky: Hi @ivanstillfront !  Thanks so much for your answer.  It seems to me that if the filesystem is not writeable, then our attempt to append to a file will not succeed, and thus the `OSError` exception you mentioned is guaranteed to be raised.  Hence, in my opinion, we could perhaps try to first detect whether or not the filesystem is writeable, and then proceed based on the results of this check.  If the system is not writeable, then we should raise something like a `ValueError` or one of the appropriate exceptions in the `great_expectations/exceptions` code area so that it is clear to the user what happened.  Please let me know what you think.  Thank you.

alexsherstinsky: @ivanstillfront In addition, I wanted to request that we do not use `noqa` as we are moving away from this pattern.  The appropriate remediations that we are now standardizing the GX OSS codebase on can be found on `https://beta.ruff.rs/docs/rules/#flake8-use-pathlib-pth` -- please search for the relevant code, and the better code pattern will be there.  We have already imported `pathlib` into this module, so this change should not be very difficult.  Please do not hesitate to let us know if there are any questions.  Thank you very much for working on this contribution.

metzlar: @alexsherstinsky there are many more reasons why the scaffolding could fail, for example, one could have no free space left on its device. The only reason we discovered this behaviour is because the original code opens `.gitignore` for writing even if it already exists. IMO, scaffolding should succeed if the required files and folders exist even if the filesystem is read-only.

I thought about proposing to ignore any problems when scaffolding `.gitignore`, basically wrap the original code in a `try..except..pass` because that would make the PR the smallest and simplest.

My change detects if `.gitignore` exists and if it does, it tests for the required `uncomitted/` line. I can intercept all exceptions from `_scaffold_gitignore` and raise a `ValueError` with a custom exception, that would also be testable because I could mock `_scaffold_gitignore` It does feel a little bit over engineered for this very specific case.

metzlar: @alexsherstinsky of course I will remove the `# noqa` statements, I was trying to keep in style with the surrounding code.

alexsherstinsky: @ivanstillfront Is it feasible to adopt your approach in the third paragraph of your message `https://github.com/great-expectations/great_expectations/pull/8362#discussion_r1269524418` -- at least some way of capturing errors and reporting them via a nice message to the user?  I worry that without proper error handling, the issues with the filesystem you described might be difficult to interpret and trace to the cause without a helpful error message.  Thank you.

great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -267,8 +267,13 @@ def _write_project_template_to_disk(
     def _scaffold_directories(cls, base_dir: PathStr) -> None:
         """Safely create GE directories for a new project."""
         os.makedirs(base_dir, exist_ok=True)  # noqa: PTH103
-        with open(os.path.join(base_dir, ".gitignore"), "w") as f:  # noqa: PTH118
-            f.write("uncommitted/")
+
```
metzlar, line 270: @alexsherstinsky here is the custom error being raised

great_expectations/data_context/data_context/serializable_data_context.py
metzlar: @alexsherstinsky I already did ;-) see my other comment

alexsherstinsky (approved):
Your changes are very good -- thank you so much -- just need to get that one failing test working.  Thanks, @ivanstillfront !

alexsherstinsky (changes requested):
Once all the tests pass, I will approve -- thank you!

alexsherstinsky (approved):
LGTM

metzlar:
@alexsherstinsky what do I have to do to get this PR merged? I have been trying to keep the branches synced but the Github Actions take so long, `develop` will have a new change, before they complete. Is there anyone else who needs to approve something?

Shinnnyshinshin:
Hi @ivanstillfront, Will here with the Core OSS team. I'll make sure this PR gets in today. thank you so much for your contribution and your patience :)

PR #8363, opened 2023-07-19
[MAINTENANCE] Add detect-private-key pre-commit hook

anthonyburdi:
- Add the detect-private-key pre commit hook: https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/detect_private_key.py
- This is from the list of supported hooks: https://pre-commit.com/hooks.html

PR #8365, opened 2023-07-19
[MAINTENANCE] Mark datasource top level tests

billdirks:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,6 @@ repos:
       - id: no-commit-to-branch
         args: [--branch, develop, --branch, main]
       - id: detect-private-key
-        exclude: tests/datasource/test_new_datasource_with_sql_data_connector.py
   - repo: https://github.com/psf/black
     rev: 23.3.0
     hooks:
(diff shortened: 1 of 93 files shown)


.pre-commit-config.yaml
```diff
@@ -15,6 +15,7 @@ repos:
       - id: no-commit-to-branch
         args: [--branch, develop, --branch, main]
       - id: detect-private-key
+        exclude: tests/datasource/test_new_datasource_with_sql_data_connector.py
```
billdirks: I've excluded this file from this check since it requires fake private keys to validate. It feels strange to do this. Let me know if there is a better way.

Shinnnyshinshin: From what I understand this is a really recent change #8363.  I dont think we should block on this for this PR, but I think it would help to get @anthonyburdi 's input on how these situations would be best-handled.

Shinnnyshinshin (approved):
LGTM!

.pre-commit-config.yaml
anthonyburdi: This looks good to me! Thanks for adding the exclude. If you want to eliminate the exclude, one route might be to programmatically generate a fake key for the test instead of hardcoding. It looks like maybe `cryptography` (which is already a core dependency) might be able to generate a fake key: https://stackoverflow.com/a/39126754

billdirks: Thanks for the pointer! I modified the stack overflow answer and am generating a fake key.

PR #8378, opened 2023-07-21
[MAINTENANCE] Do not run CI on draft pull requests, enable running manually

anthonyburdi:
- Add ability to run CI manually
- Don't run CI on draft PRs
- Stage with descriptive name runs on draft PRs and describes why CI is not running
- I also tried to look for a method (using the conditional `if:` or the `on:`) to skip all jobs (so we don't have to duplicate the conditional on all jobs) but I was not successful. This issue explores some of these options: https://github.com/orgs/community/discussions/25722

When the PR is a draft PR, note the `ci / ci-does-not-run-on-draft-pull-requests (pull_request)` job and other jobs are skipped:
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
   schedule:
     # https://crontab.guru/every-3-hours
     - cron: "0 */3 * * *"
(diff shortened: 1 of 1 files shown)


alexsherstinsky (approved):
❤️

PR #8381, opened 2023-07-23
[MAINTENANCE] Revert linting changes to format_docstring_to_markdown func

kenwade4:
- Revert linting changes to format_docstring_to_markdown func from PR #8123


kwcanuck (approved):
I took a quick look at the output and we need to find some time to address some of the other issues with this documentation. For example, removing punctuation from headings and verifying that links still point to relevant documentation.

PR #8393, opened 2023-07-24
[FEATURE] Allow a rule to fail within DataAssistant and still run other rules

TrangPham:
DataAssistants rely on RuleBasedProfiler, thus this change is within RuleBasedProfiler itself. This does a simple try catch and is the low hanging fruit of what we could do here. The next steps would be to bubble up errors to the user - that would be coming in a follow-up PR.


great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -324,15 +324,20 @@ def run(  # noqa: PLR0913
...
+                    batch_list=batch_list,
+                    batch_request=batch_request,
+                    runtime_configuration=runtime_configuration,
+                    reconciliation_directives=reconciliation_directives,
+                    rule_state=RuleState(),
+                )
+                self.rule_states.append(rule_state)
+            except Exception as e:
+                logger.error(
+                    f'An exception occurred while running rule "{rule.name}": {e}'
```
Kilo59:
```suggestion
                    f'An exception occurred while running rule "{rule.name}": {repr(e)}'
```
I would also log the exception repr or the exception class name otherwise only the exception message will be logged which may not be enough to determine what the actual exception type was.

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
```diff
@@ -317,6 +319,66 @@ def test_onboarding_data_assistant_result_batch_id_to_batch_identifier_display_n
     )
 
 
+@pytest.mark.integration
```
Kilo59:
```suggestion
@pytest.mark.big
```

We got rid of the `integration` marker as part of the testing overhaul effort.
If you update this branch with develop this integration marker will cause this test to break (or never be run).

The new marker doesn't have to be `big` as long as it's something other than `integration`

Kilo59 (approved):
I don't know the RBP code very well but the change seems reasonable.

Approving pending the 2 suggestions I made are addressed.

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
```diff
@@ -317,6 +319,65 @@ def test_onboarding_data_assistant_result_batch_id_to_batch_identifier_display_n
...
+    batch_request: dict = {
+        "datasource_name": "taxi_pandas",
+        "data_connector_name": "monthly",
+        "data_asset_name": "my_reports",
+    }
+
+    num_rules = 8  # There are 8 rules in the onboarding data assistant.
+    with mock.patch(
+        "great_expectations.rule_based_profiler.rule.rule.Rule.run",
+    ) as mock_run:
```
cdkini, line 339: Do we need to evaluate all 8 rules to test this change? Just thinking about test performance (I know this area is historically quite cumbersome in terms of testing)

TrangPham: Yes I agree about test performance. 

The rules aren't run at all in this test. They're all mocked out. 

The only purpose of this test is to ensure that the call count of rules is equal to the number of rules in the DA, even if one of the rules failed.

PR #8410, opened 2023-07-25
[FEATURE] Add Test Draft Config Workflow to Agent

joshua-stauffer:
This PR adds a new agent workflow, which retrieves a draft datasource config from Cloud, and tests its connection.


ghost:
<details open>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=8410&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

####

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=8410&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map legend
</details>

tests/agent/actions/test_draft_datasource_config.py
```diff
@@ -0,0 +1,169 @@
...
+    return "MTg0NDkyYmYtNTBiOS00ZDc1LTk3MmMtYjQ0M2NhZDA2NjJk"
+
+
+@pytest.fixture(autouse=True)
+def set_required_env_vars(monkeypatch, org_id, token):
+    env_vars = {
+        "GX_CLOUD_ORGANIZATION_ID": org_id,
+        "GX_CLOUD_ACCESS_TOKEN": token,
+    }
+    monkeypatch.setattr(os, "environ", env_vars)
```
Kilo59: Please use the `monkeypatch.setenv` for this.
https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html#monkeypatching-environment-variables

tests/agent/actions/test_draft_datasource_config.py
```diff
@@ -0,0 +1,169 @@
...
+def org_id():
+    return "81f4e105-e37d-4168-85a0-2526943f9956"
+
+
+@pytest.fixture
+def token():
+    return "MTg0NDkyYmYtNTBiOS00ZDc1LTk3MmMtYjQ0M2NhZDA2NjJk"
+
+
+@pytest.fixture(autouse=True)
```
Kilo59:
```suggestion
@pytest.fixture
```
I think we should avoid `autouse` unless absolutely necessary, it makes things harder to reason about.

tests/agent/actions/test_draft_datasource_config.py
joshua-stauffer: thanks! [90923bc](https://github.com/great-expectations/great_expectations/pull/8410/commits/90923bc765c227a1e940d35d7ad9c8052b1e77dd)

tests/agent/actions/test_draft_datasource_config.py
joshua-stauffer: yep, we normally don't use `autouse`, but in this case every single test in this module requires this fixture and has no use for the return value within the test, so this is the correct time to use it.

Kilo59: I still disagree with the use of `autouse`. This test module is easy to reason about now, but it will inevitably grow in size and complexity.

You can use it in your `context` fixture and then you don't have to declare it in every test (and with less magic than autouse).

https://docs.pytest.org/en/6.2.x/fixture.html#fixtures-can-request-other-fixtures

```python

@pytest.fixture(scope="function")
def context(set_required_env_vars):
    return MagicMock(autospec=CloudDataContext)

...

def test_test_draft_datasource_config_success(context, mocker):
   ...
```

great_expectations/agent/actions/test_draft_datasource_config.py
```diff
@@ -0,0 +1,49 @@
...
+        try:
+            datasource_cls = self._context.sources.type_lookup[datasource_type]
+        except KeyError as exc:
+            raise ValueError(
+                "DraftDatasourceConfigAction received an unknown datasource type."
+            ) from exc
+        datasource = datasource_cls(**draft_config)
+        datasource.test_connection(
+            test_assets=True
+        )  # raises exception if config isn't valid
```
Kilo59: I think this comment is misleading.
Exceptions will be raised on `datasource_cls(**draft_config)` if the config is invalid according to pydantic model validation.

`test_connection()` can still raise exceptions even if the config is "valid", there could be a number of reasons why a connection failed. A temporary outage, expired credentials, slow internet leading to a timeout etc.

joshua-stauffer: sure - [493ef06](https://github.com/great-expectations/great_expectations/pull/8410/commits/493ef0663ac7233d36c4b686b114e17540e6eeeb)

tests/agent/actions/test_draft_datasource_config.py
joshua-stauffer: okay, looks like we just have different opinions here. Why would `context` require that fixture? it doesn't need the environment variables, and has no use for the return value. When we write a test that defines its own `context` mock, that test will unexpectedly fail due to missing environment variables, even though its otherwise identical to other tests in the module. That seems more magical and more unexpected.

Kilo59: I agree `context` doesn't need the `set_required_env_vars` fixture.
The example with `context` was to just illustrate how fixtures can use other fixtures.

And I agree it's still a bit of magic. I would prefer if all the tests were explicit, about the fixtures they use. 

👇 This is how I would prefer the fixtures be declared.

```python
def test_test_draft_datasource_config_success(context, mocker, set_required_env_vars):
   ...

def test_test_draft_datasource_config_failure(context, mocker, set_required_env_vars):
   ...
```

cdkini: I agree with avoiding `autouse` if possible! Explicit is always better than implicit and as a future reader of this test, it would be nice to know exactly what setup logic is occurring before I get to the actual core logic of the test. 

I'm of the opinion that `pytest` is already a little too magical so it would be nice to minimize that wherever we can imo.

great_expectations/agent/models.py
```diff
@@ -43,6 +44,11 @@ class RunColumnDescriptiveMetricsEvent(EventBase):
     data_asset_name: str
 
 
+class DraftDatasourceConfig(EventBase):
+    type: Literal["test_datasource_config"] = "test_datasource_config"
```
roblim: A bit confused by the naming. Should class be `TestDraftDatasourceConfig` or `TestDatasourceConfig`? To stick with convention, should we prepend `Run` and append `Event` to the class name? 

Also, all the other event type fields have `.received` at the end so far - should this also follow that?

great_expectations/agent/actions/test_draft_datasource_config.py
```diff
@@ -0,0 +1,49 @@
...
+import pydantic
+
+from great_expectations.agent.actions import ActionResult, AgentAction
+from great_expectations.agent.agent import GxAgentEnvVars
+from great_expectations.agent.models import DraftDatasourceConfig
+from great_expectations.core.http import create_session
+
+
+class DraftDatasourceConfigAction(AgentAction[DraftDatasourceConfig]):
+    def run(self, event: DraftDatasourceConfig, id: str) -> ActionResult:
```
roblim: See other comment on naming confusion.

great_expectations/agent/actions/test_draft_datasource_config.py
```diff
@@ -0,0 +1,49 @@
...
+
+import pydantic
+
+from great_expectations.agent.actions import ActionResult, AgentAction
+from great_expectations.agent.agent import GxAgentEnvVars
+from great_expectations.agent.models import DraftDatasourceConfig
+from great_expectations.core.http import create_session
+
+
+class DraftDatasourceConfigAction(AgentAction[DraftDatasourceConfig]):
```
roblim: Shouldn't this action also be added to `actions/__init__.py`? (I think this is where I ran into circular import thing)

great_expectations/agent/models.py
roblim: Doesn't this Event need to be connected to Action in the Eventhandler?

tests/agent/actions/test_draft_datasource_config.py
roblim: +1 on preferring minimal use of autouse (as someone who is regretful author of several autouse fixtures). As useful as I've found that option in the past, now that we have a lot more tests, I'm learning how much it can bog down test debugging. Particularly with pytest's complicated fixture import/scope hierarchy. And especially when working in a new part of codebase.

roblim (dismissed):
Core logic/tests look good, just had a few questions on naming/conventions and comment on possible missing piece.

Kilo59 (changes requested):
Please remove the use of `autouse` for the `set_required_env_vars` fixtures

great_expectations/agent/actions/test_draft_datasource_config.py
Kilo59: https://github.com/great-expectations/great_expectations/blob/1ca106cea4d96040946427a43a8a757c760146ed/great_expectations/agent/actions/__init__.py#L6-L12

great_expectations/agent/agent.py
```diff
@@ -48,7 +49,13 @@ class GXAgentConfig(AgentBaseModel):
...
     queue: str
     connection_string: AmqpDsn
-    gx_cloud_base_url: str = "https://api.greatexpectations.io"
+    gx_cloud_base_url: AnyUrl = CLOUD_DEFAULT_BASE_URL  # type: ignore[assignment]
+    gx_cloud_organization_id: str
+    gx_cloud_access_token: str
+
+
+class GxAgentEnvVars(pydantic.BaseSettings):
+    gx_cloud_base_url: AnyUrl = CLOUD_DEFAULT_BASE_URL  # type: ignore[assignment]
```
Kilo59: `non-blocking`
I would add a comment here explaining the type ignore.

>`# pydantic will coerce this string to AnyUrl type`

tests/agent/actions/test_draft_datasource_config.py
joshua-stauffer: updated in [62ef383](https://github.com/great-expectations/great_expectations/pull/8410/commits/62ef3838fd686b8bd586649aa9c8483af59bdae2)

great_expectations/agent/actions/test_draft_datasource_config.py
joshua-stauffer: thanks, great catch!

great_expectations/agent/models.py
joshua-stauffer: yes, thanks!

tests/agent/actions/test_draft_datasource_config.py
joshua-stauffer: if this is a blocking req, this needs to be added to our style guide

great_expectations/agent/models.py
joshua-stauffer: thanks for the questions - `TestDraftDatasourceConfig` caused issues with pytest, since it tries to pick it up as a test. `.received` should be removed in new actions - it's not doing anything, and we don't have any plans to use other suffixes. The `Run` prefix isn't a broad convention, it's a part of the action name, i.e. "Run Data Assistant". In this case, we're not running anything, so isn't necessary.

great_expectations/agent/actions/test_draft_datasource_config.py
joshua-stauffer: thanks, updated this as well - not sure what i was thinking when i named these 🙃

great_expectations/agent/agent.py
joshua-stauffer: sure - [df253ec](https://github.com/great-expectations/great_expectations/pull/8410/commits/df253ecd31e09b4e5bdb9410ea01f20e7fe7c7b9)

great_expectations/agent/models.py
cdkini: I think we could get around this if we wanted to prefix with `Test`: https://stackoverflow.com/a/24635769

tests/agent/actions/test_draft_datasource_config.py
cdkini: I think we're adding this as an agenda item for OSS arch review tomorrow! Don't think this should be a blocker but a good conversation to have

great_expectations/agent/models.py
joshua-stauffer: thanks @cdkini - not sure it's worth the effort to try and work around pytest convention, the rename is clear enough 👍

Kilo59 (approved):

roblim (approved):
LGTM - thanks for addressing comments!

PR #8429, opened 2023-07-26
[FEATURE] DataAssistantResult should include exceptions from rules

TrangPham:
Recently we added the ability for the DataAssistant run method to fail on a rule but still continue with the other rules.
This logic adds the ability to inform the user which rule failed and in what way did it fail.


tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
```diff
@@ -378,6 +376,13 @@ def side_effect(*args, **kwargs):
         assert mock_run.call_count == num_rules
         assert call_count == num_rules
 
+        result = data_assistant_result.to_json_dict()
+        assert result["rule_exception_tracebacks"]
+        assert (
+            result["rule_exception_tracebacks"]["table_rule"]["exception_message"]
+            == "This rule failed."
+        )
```
anthonyburdi, line 398: What do you think about separating these asserts into a different test? This one is already marked "big" Something like `test_data_assistant_rule_exception_captured`.

great_expectations/rule_based_profiler/rule/rule_state.py
```diff
@@ -10,22 +10,24 @@
...
     """
 
-    def __init__(
+    def __init__(  # noqa: PLR0913
         self,
         rule: Optional[Rule] = None,
         domains: Optional[List[Domain]] = None,
         variables: Optional[ParameterContainer] = None,
         parameters: Optional[Dict[str, ParameterContainer]] = None,
+        catch_exceptions: Optional[bool] = None,
```
anthonyburdi: Are these marked Optional because of backward compatibility with existing configs or can we use a default of `False` here?

TrangPham: I think default false could work here. I'm not sure about backwards compatibility though as I haven't looked at that many examples of profilers being used outside of data assistants.

I was trying to match the catch_exceptions parameter & functionality we have for metrics.

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
TrangPham: I'm actually not sure if this should be marked "big" previously the notation was "integration". Do the two mean the same thing?

This test should be much faster than the other onboarding data assistant tests since the running of the rule is mocked and thus data would not be evaluated in this test.

TrangPham: I could also see if changing to using an in-memory dataframe asset would speed up this test. I don't think the batch_request data is being loaded until the rules are run, but I may be incorrect about that.

great_expectations/rule_based_profiler/rule/rule_state.py
Kilo59: If there's no different behavior between a `False` and `None` I would just remove the `Optional` and make it default to `False` and therefore removing a potential element of confusion about whether to pass `False` or `None`.

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
anthonyburdi: I agree - the mock for `run` is great, we definitely don't want to run the assistant on data for this test. I was thinking along the lines of splitting the asserts into two tests, mainly to have less asserts per test:
1. checking that the other rules run even if one fails
2. when a rule fails, does it's exception info get captured in the results

But for this particular test it's not a strong conviction so if you disagree I'm ok leaving as-is.

As an aside, I'm not sure if it is possible with the current architecture, but it would be nice to also not have to load a real data context for this test e.g. `bobby_columnar_table_multi_batch_deterministic_data_context`.

TrangPham: I think leaving them together is okay - since the tests would do the same thing but just two different assertions 

I did change to not loading real data. This made the test much faster - thus I have removed the big decorator

anthonyburdi: Thanks! I think we now require a decorator on all tests based on the work being done on testing, but I could be mistaken. I'll approve but please check with the team working on testing, I'm not sure if there is an automated check in place for ensuring there is a decorator.

anthonyburdi (approved):
Thank you! Please just confirm which decorator should go on the test but otherwise LGTM!

tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
TrangPham: Screenshot 2023-07-31 at 3 23 21 PM
Looks like the test for it exists. :D updating the marker

great_expectations/rule_based_profiler/rule_based_profiler_result.py
```diff
@@ -56,6 +59,8 @@ class RuleBasedProfilerResult(SerializableDictDot):
     citation: dict
     rule_domain_builder_execution_time: Dict[str, float]
     rule_execution_time: Dict[str, float]
+    rule_exception_tracebacks: Dict[str, Optional[str]]
+    catch_exceptions: Optional[bool] = field(default=None)
```
Kilo59:
```suggestion
    catch_exceptions: bool = field(default=False)
```

great_expectations/rule_based_profiler/rule_based_profiler.py
```diff
@@ -1602,6 +1623,7 @@ def __init__(  # noqa: PLR0913
         rules: Optional[Dict[str, Dict[str, Any]]] = None,
         data_context: Optional[AbstractDataContext] = None,
         id: Optional[str] = None,
+        catch_exceptions: Optional[bool] = None,
```
Kilo59:
```suggestion
        catch_exceptions: bool = False,
```

anthonyburdi: Good catch! @TrangPham I disabled automerge to address removing Optional in favor of the default (here and anywhere else)

PR #8439, opened 2023-07-27
[BUGFIX] Cleanup aws_postgres reference environment

anthonyburdi:
Don't use airflow in aws_postgres dockerfile


Shinnnyshinshin (approved):
oo I like this.

PR #8440, opened 2023-07-27
[BUGFIX] Use consistent name between compose and cli command

anthonyburdi:
Use consistent name between compose and cli command in the aws postgres reference environment. This was leading to an error when trying to find the url of the jupyter server running in the container.


Shinnnyshinshin (approved):
shoot this was probably me 😅 LGTM

PR #8443, opened 2023-07-27
[MAINTENANCE] Run cloud e2e tests.

billdirks:


Kilo59 (approved):

.github/workflows/ci.yml
```diff
@@ -107,8 +107,7 @@ jobs:
         markers:
           - big
           - cli
-          # `not e2e` because 1 test & 1 fixture need access token
-          - cloud and not e2e
```
billdirks, line 111: I will need to update the branch protection rules before this will merge.
Also, after this merges I will need to add the cloud step as a required test.

.github/workflows/ci.yml
```diff
@@ -137,6 +136,10 @@ jobs:
           invoke deps --gx-install -m '${{ matrix.markers }}' -r test
 
       - name: Run the tests
+        env:
```
billdirks: Maybe I should only set these env variables for the cloud mark stage?

tests/datasource/fluent/test_config.py
```diff
@@ -819,7 +819,8 @@ def test_config_substitution_retains_original_value_on_save(
     from great_expectations import get_context
 
     context = get_context(
-        context_root_dir=file_dc_config_file_with_substitutions.parent
+        context_root_dir=file_dc_config_file_with_substitutions.parent,
+        ge_cloud_mode=False,
```
billdirks: I've set the cloud env variables for all marker stages so had to make sure these didn't get a cloud context.
Maybe I should just set the env variables on the cloud marker test? Or maybe I should do both. These tests should be robust to the env variables.

billdirks: Doing things like this let's me get rid of that darwin check in `__init__.py`. Ultimately, making the tests pass independent of a developer's environment is better. Currently linux users would fail these tests if they have gx cloud env variables set.  I still see tests dying though because I've set the env variables in all marker tests. Hopefully isn't a pain to clean up as we go.

.github/workflows/ci.yml
Kilo59: There may be a way to conditionally inject env vars with gh actions but another simpler solution could be to just break the steps that need different env vars to their own job.

billdirks: Yeah, that sounds like a good idea.

tests/datasource/fluent/test_config.py
billdirks: I will move and use the fixture I introduced and use that instead of setting the `ge_cloud_mode` parameter.

tests/validator/test_validator.py
```diff
@@ -342,6 +343,14 @@ def multi_batch_taxi_validator_ge_cloud_mode(
...
 
 
+@pytest.fixture()
+def _unset_gx_env_variables(monkeypatch: pytest.MonkeyPatch) -> None:
+    for var in GXCloudEnvironmentVariable:
+        monkeypatch.delenv(var, raising=False)
+
+
+# TODO: There is something wrong with this test. It is trying to mock out cloud but if I don't
+#       unset the gx_env_variables (eg if i remove this fixture) this test will fail.
```
Kilo59: I had a similar approach in this abandoned effort from about a month ago.
Except in that case I was injecting dummy Cloud variables for the non `e2e` tests.

Unsetting the `gx_env` variables might be something we want to apply to the test session setup???

- https://github.com/great-expectations/great_expectations/pull/8134
  - https://github.com/great-expectations/great_expectations/pull/8134/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128

.github/workflows/ci.yml
billdirks: I'm going to leave it like this for this PR and move it out later if we want.

tests/validator/test_validator.py
billdirks: We do need to normalize our environment somehow. If I unset these across the board the tests that require these environment variables set will fail. We need a way that is not burdensome to a test writer that gets this correct.

.github/workflows/ci.yml
billdirks: Actually, I will move this out now. Setting the environmental variables cause a lot of issues. I will file a followup ticket to review this further. It only impacts people who are both gx cloud and OSS developers.

PR #8448, opened 2023-07-28
[MAINTENANCE] Default snippet-check to non verbose mode

anthonyburdi:
Verbose mode can be invoked by adding the verbose flag, we don't need to default to verbose for every invocation.


docs/docusaurus/package.json
```diff
@@ -53,4 +53,4 @@
       "last 1 safari version"
     ]
   }
-}
\ No newline at end of file
+}
```
anthonyburdi, line 56: Newline added at end of file by linter

donaldheppner (approved):

PR #8449, opened 2023-07-28
[FEATURE] add run checkpoint action to agent

jshaikGX:

### DIFF
--- a/docs/sphinx_api_docs_source/public_api_excludes.py
+++ b/docs/sphinx_api_docs_source/public_api_excludes.py
@@ -762,6 +762,11 @@ DEFAULT_EXCLUDES: list[IncludeExcludeDefinition] = [
         filepath=pathlib.Path("great_expectations/agent/actions/list_table_names.py"),
     ),
     IncludeExcludeDefinition(
+        reason="The agent is not part of the public API",
+        name="run",
+        filepath=pathlib.Path("great_expectations/agent/actions/run_checkpoint.py"),
+    ),
+    IncludeExcludeDefinition(
         reason="The add method shares a name with a public API method",
         name="add",
         filepath=pathlib.Path(
--- a/great_expectations/agent/event_handler.py
+++ b/great_expectations/agent/event_handler.py
@@ -13,6 +13,7 @@ from great_expectations.agent.actions.data_assistants import (
 from great_expectations.agent.actions.draft_datasource_config_action import (
     DraftDatasourceConfigAction,
 )
+from great_expectations.agent.actions.run_checkpoint import RunCheckpointAction
 from great_expectations.agent.models import (
     DraftDatasourceConfigEvent,
     Event,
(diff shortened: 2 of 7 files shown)


great_expectations/agent/event_handler.py
```diff
@@ -63,7 +64,7 @@ def get_event_action(self, event: Event) -> AgentAction:
             return ListTableNamesAction(context=self._context)
 
         if isinstance(event, RunCheckpointEvent):
-            raise NotImplementedError
+            return RunCheckpointAction(context=self._context)
```
billdirks, line 67: This is ok for this PR since this comment is about `get_event_action` in general but this `isinstance` pattern seems non-ideal:
* It requires this boilerplate code for every new action
* It violates the open-closed principle (the O in SOLID)
* Grows the cyclomatic complexity of this function.
Ultimately, all these branches call an action constructor with the context (which the exception of `RunColumnDescriptiveMetricsEvent` though that looks like all the work in that branch is based on the context so could be modified) either via a protocol definition/base class abstract interface or something similar.

billdirks (approved):
I left one comment that should be addressed but it is perhaps out of scope of this PR.

It also appears an agent action could die and we don't handle that and report it to the gx cloud backend (I believe you brought this up in conversation) so we must not display those currently. That is an effort outside of the PR but something we should discuss at some point.

docs/sphinx_api_docs_source/public_api_excludes.py
```diff
@@ -761,6 +761,11 @@
         name="run",
         filepath=pathlib.Path("great_expectations/agent/actions/list_table_names.py"),
     ),
+    IncludeExcludeDefinition(
+        reason="The run method shares a name with a public API method",
```
joshua-stauffer: this isn't accurate - the agent is simply excluded from the public API

joshua-stauffer (approved):
this is great @jshaikGX, thanks! looks good, and it works 👍 left one non-blocking comment

great_expectations/agent/event_handler.py
Kilo59: `non-blocking`
Could also use `single-dispatch` here 😄 .
The first time I brought this up there was only 1 or 2 actions.
I think it's worth revisiting.

https://peps.python.org/pep-0443/

From the `pep` that added it.

>A generic function is composed of multiple functions implementing the same operation for different types. Which implementation should be used during a call is determined by the dispatch algorithm. When the implementation is chosen based on the type of a single argument, this is known as single dispatch.

>... it is currently a common anti-pattern for Python code to inspect the types of received arguments, in order to decide what to do with the objects.

>For example, code may wish to accept either an object of some type, or a sequence of objects of that type. Currently, the “obvious way” to do this is by type inspection, but this is brittle and closed to extension.


But as @billdirks said, this is not a problem for @jshaikGX to work out on this PR.
I'd be happy to implement this if we agree this is the direction we want to go.

---------

Note @jshaikGX I unresolved the conversation so it's easier to discover.
We don't require conversations to be resolved in the OSS repo.

PR #8457, opened 2023-07-31
[BUGFIX] Remove batching_regex param from directory asset add methods

anthonyburdi:
`batching_regex` is not needed (and was not working) for a directory asset since the files are essentially treated as a single batch (unless you split them later).


Kilo59 (changes requested):
You need to update the implementation as well. Changing the stubs has no impact at runtime. 

I would expect to see changes here.
https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/datasource/fluent/spark_file_path_datasource.py

anthonyburdi:
Closing for now until we can address.

PR #8477, opened 2023-08-02
[MAINTENANCE] Add check that each test only has one required marker.

billdirks:
Most of these changes are updating markers but there is a few other changes that I call out explicitly in comments.

Here is my decision process when updating markers.
**cli vs unit => cli**: cli tests are all in 1 directory and may go away. They should be tested as 1 whole segment
**big vs all_backends => all_backends**: this will ensure the backend variants will get run along with the no backend required variants. We may repeat running the no backend required variant for each backend but that is ok for now.
**cloud vs unit => unit**: we may pull cloud out. These are all 1 logical unit.
**filesystem vs big => big**: big seems to convey more than filesystem. 
**spark vs filesystem => spark**: this is 1 logical unit.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -795,7 +795,7 @@ class TestDependencies(NamedTuple):
     ] = tuple()
 
 
-MARKER_DEPENDENDENCY_MAP: Final[Mapping[str, TestDependencies]] = {
+MARKER_DEPENDENCY_MAP: Final[Mapping[str, TestDependencies]] = {
     "athena": TestDependencies(("reqs/requirements-dev-athena.txt",)),
     "clickhouse": TestDependencies(("reqs/requirements-dev-clickhouse.txt",)),
     "cloud": TestDependencies(
@@ -860,9 +860,9 @@ MARKER_DEPENDENDENCY_MAP: Final[Mapping[str, TestDependencies]] = {
 
 
 def _add_all_backends_marker(marker_string: str) -> bool:
-    # We should generalize this, possibly leveraging MARKER_DEPENDENDENCY_MAP, but for now
+    # We should generalize this, possibly leveraging MARKER_DEPENDENCY_MAP, but for now
     # right I've hardcoded all the containerized backend services we support in testing.
-    return marker_string in ["postgresql", "mssql", "mysql", "trino"]
+    return marker_string in ["postgresql", "mssql", "mysql", "spark", "trino"]
 
 
 def _tokenize_marker_string(marker_string: str) -> Generator[str, None, None]:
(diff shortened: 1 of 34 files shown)


tasks.py
```diff
@@ -795,7 +795,7 @@ class TestDependencies(NamedTuple):
     ] = tuple()
 
 
-MARKER_DEPENDENDENCY_MAP: Final[Mapping[str, TestDependencies]] = {
+MARKER_DEPENDENCY_MAP: Final[Mapping[str, TestDependencies]] = {
```
billdirks, line 798: Fixed spelling.

tasks.py
```diff
@@ -860,9 +860,9 @@ class TestDependencies(NamedTuple):
 
 
 def _add_all_backends_marker(marker_string: str) -> bool:
-    # We should generalize this, possibly leveraging MARKER_DEPENDENDENCY_MAP, but for now
+    # We should generalize this, possibly leveraging MARKER_DEPENDENCY_MAP, but for now
     # right I've hardcoded all the containerized backend services we support in testing.
-    return marker_string in ["postgresql", "mssql", "mysql", "trino"]
+    return marker_string in ["postgresql", "mssql", "mysql", "spark", "trino"]
```
billdirks, line 865: Added spark here because some of the tests run by `test_case_runner_v3_api` in `test_expectations_v3_api.py` specify a spark backend. This test runs against different backends and requires the `all_backends` marker to do this. I didn't want spark to get skipped so I added it here.

tests/conftest.py
```diff
@@ -135,14 +135,16 @@
     "cloud",
     "docs",
     "filesystem",
-    "google_creds",
+    "mssql",
     "mysql",
     "openpyxl",
+    "performance",
```
billdirks, line 141: I updated the `google_creds` marker to `performance`. Since that is the only marker we exclude from our test run, I updated the coverage error message to explicitly state this. I also though `performance` conveyed this better since we performance tests are a different thing than correctness tests.

tests/conftest.py
```diff
@@ -364,25 +366,67 @@ def pytest_generate_tests(metafunc):
...
-    uncovered: list[tuple[str, str, set[str]]] = []
+) -> tuple[list[TestMarkerCoverage], list[TestMarkerCoverage]]:
+    uncovered: list[TestMarkerCoverage] = []
+    multiple_markers: list[TestMarkerCoverage] = []
     for test in session.items:
         markers = {m.name for m in test.iter_markers()}
-        if not REQUIRED_MARKERS.intersection(markers):
-            uncovered.append((str(test.path), test.name, markers))
-    return uncovered
+        required_intersection = markers.intersection(REQUIRED_MARKERS)
```
billdirks, line 386: This is the new logic for testing only 1 marker is present.

tasks.py
Shinnnyshinshin: 😓

tasks.py
Shinnnyshinshin: Thank you. I sounds right to me.

tests/conftest.py
Shinnnyshinshin: This makes a lot of sense. Thank you

tests/conftest.py
Shinnnyshinshin: :rocket:

Shinnnyshinshin (approved):
This looks great. Thank you :)

PR #8509, opened 2023-08-08
[MAINTENANCE] Better testing for FDS quoted identifier logic

Kilo59:
Update the testing scaffolding for SQLDatasource integration testing
Create the needed tables for tests where possible.

Don't include quote strings as part of the `TestConnectionError`

Expect followups to actually fix some of these bugs and add more backends.

----


tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -43,12 +57,85 @@ def context() -> EphemeralDataContext:
...
+) -> Generator[TableFactory, None, None]:
+    """
+    Given a SQLALchemy engine, table_name and schema,
+    create the table if it does not exist and drop it after the test.
+    """
+    all_created_tables: dict[
+        str, list[dict[Literal["table_name", "schema"], str | None]]
+    ] = {}
+    engines: dict[str, engine.Engine] = {}
+
```
Kilo59, line 91: This could be simplified.
Originally this was going to be a module/class-scoped fixture that setup tables for multiple backends, and so needed to save the engine used so that it could be used for cleanup.

I'd like to keep it like this for the next iteration but if we don't end up using it I will simplify this.

tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -68,17 +155,22 @@ def postgres_ds(context: EphemeralDataContext) -> PostgresDatasource:
         param("quoted_lower"),
         param(
             "unquoted_upper",
-            marks=[pytest.mark.xfail(reason="table names should be lowercase")],
+            marks=[pytest.mark.xfail(reason="TODO: fix or remove")],
```
wookasz: definitely fix

tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -68,17 +155,22 @@ def postgres_ds(context: EphemeralDataContext) -> PostgresDatasource:
...
         ),
+        param("quoted_upper"),
         param(
-            "quoted_upper",
-            marks=[pytest.mark.xfail(reason="table names should be lowercase")],
+            "unquoted_mixed",
+            marks=[pytest.mark.xfail(reason="TODO: fix or remove")],
```
wookasz: also a fix

wookasz (commented):
Are we trying to DRY too much here. There are several ways that per-DS behavior is being overridden (`TABLE_NAME_MAPPING`, `DO_NOT_CREATE_TABLES`, the specific table names).

I think it would be easier to reason about these tests if we did a bit of repetition and set up a separate test class for each DS we're putting under test. Each of these would be explicit about the table names, whether tables are being created or not, etc.

tests/datasource/fluent/integration/test_sql_datasources.py
Kilo59: Updated

~So this does pass when I run this locally, but for some reason, it's failing in CI.~
~Possiblly because there's some old `test_table` table sitting around from a previous test run.~ 🤷‍♂️ 

It's because I wasn't committing the drop operations at the end of the tests locally. So a previous run was affecting it.

tests/datasource/fluent/integration/test_sql_datasources.py
Kilo59: Updated the `reason`

Kilo59:
@wookasz 

> Are we trying to DRY too much here. There are several ways that per-DS behavior is being overridden

Yes, my plan is to remove these mappings and simplify things (adding some repetition). Did this out of laziness as I was changing how these tests work.
I agree that this makes the tests confusing.

tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -43,6 +59,80 @@ def context() -> EphemeralDataContext:
...
 
 
+class TableFactory(Protocol):
+    def __call__(
+        self,
+        engine: engine.Engine,
+        table_names: set[str],
+        schema: str | None = None,
+    ) -> None:
+        ...
```
Kilo59, line 69: Doing this instead of annotating the factory as `Callable` which can't deal with keyword arguments well.

wookasz (approved):

PR #8512, opened 2023-08-08
[BUGFIX] Fixed persist to be considered for spark fluent datasource

### LINKED ISSUE #8511: Not able to set persist to False using Spark Execution Engine
Hey Guys, I was trying to migrate GX from 0.16.5 and am not being able to, cause apparently since GX 0.16.12 supposedly there was a fix for the persist parameter to work. The thing is that I wanted it to be False, but it seems the parameter is not taking effect.
Anyone facing similar problems?
I was following this guide [How to connect to in-memory data in a Spark dataframe | Great Expectations](https://docs.greatexpectations.io/docs/0.15.50/guides/connecting_to_your_data/in_memory/spark/), which seems to have been removed. Also tried with fluent approach and still failing.
Is the property persist being considered/passed? Shouldn’t it be one of the parameters in add_or_update_spark ?

More information: https://discourse.greatexpectations.io/t/not-able-to-set-persist-to-false-using-spark-execution-engine/1320?u=jose.correia

jmcorreia:
Fix persist parameter to be considered for Spark Datasources, so that people can pass it as False.
Enhanced test to cover the code changes.

Fixes #8511 .


tests/datasource/fluent/test_spark_datasource.py
```diff
@@ -69,6 +76,13 @@ def test_dataframe_asset(
         for asset in datasource.assets
     )
 
+    datasource_config = yaml.load(empty_data_context.fluent_config.yaml())[
+        "fluent_datasources"
+    ][datasource_name]
+    assert datasource_config["spark_config"] == {}
+    assert datasource_config["force_reuse_spark_context"] is True
+    assert datasource_config["persist"] is False
```
Kilo59: This part of the test shouldn't be necessary. This is just testing the yaml serialization logic which is tested elsewhere.

tests/datasource/fluent/test_spark_datasource.py
```diff
@@ -43,7 +44,13 @@ def test_dataframe_asset(
...
     assert errors_dict["loc"][0] == "dataframe"
 
-    datasource = empty_data_context.sources.add_spark(name="my_spark_datasource")
+    datasource_name = "my_spark_datasource"
+    datasource = empty_data_context.sources.add_spark(
+        name=datasource_name,
+        persist=False,
+        force_reuse_spark_context=True,
+        spark_config={},
+    )
```
Kilo59: I think this might fit better in the `test_spark_config_passed_to_execution_engine` test below.
Or maybe that test should be parametrized to include the `persist` key.

tests/datasource/fluent/test_spark_datasource.py
jmcorreia: The idea was basically to ensure that the persist to False is really there. 
But I will do the suggested changes.

tests/datasource/fluent/test_spark_datasource.py
jmcorreia: Done

Kilo59 (approved):
Thanks for this!

I don't know Spark that well but this seems reasonable.
I'd like to get some reviews from some other engineers before merging.

great_expectations/datasource/fluent/sources.pyi
```diff
@@ -320,20 +320,23 @@ class _SourceFactories:
...
         force_reuse_spark_context: bool = True,
+        persist: bool = True,
     ) -> SparkDatasource: ...
     def add_or_update_spark(
         self,
         name: str,
         *,
         spark_config: SparkConfig | None = None,
         force_reuse_spark_context: bool = True,
+        persist: bool = True,
```
NathanFarmer, line 339: Can you also add this parameter to the `*_spark_filesystem`, `*_spark_dbfs`, `*_spark_s3`, `*_spark_gcs`, and `*_spark_abs` methods in this file?

NathanFarmer (commented):
One request from me. Thank you for this contribution!

great_expectations/datasource/fluent/sources.pyi
jmcorreia: Nice callout, applied.
Anyway, I have also tested without adding persist as a clear parameter and it also works, as we have a * there.

jmcorreia:
@anthonyburdi , @NathanFarmer could you please let me know if you still something missing? Really need this to be able to upgrade great-expectations on my side :)
Btw, this fixes only fluent datasource (spark_datasource), it does not fix for people using the non-fluent version sparkdf_datasource. Should I try to apply the fix there as well? And if I do it, can I change the default of force_reuse_spark_context to True?

Or if you think it is better, I can also propose the rest in a separate PR.

PR #8524, opened 2023-08-09
[MAINTENANCE] Remove datasource cache from context

cdkini:

### DIFF
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -310,9 +310,6 @@ class AbstractDataContext(ConfigPeer, ABC):
             self.project_config_with_variables_substituted.anonymous_usage_statistics
         )
 
-        # Store cached datasources but don't init them
-        self._cached_datasources: dict = {}
-
         # Build the datasources we know about and have access to
         self._init_datasources()
 
@@ -780,7 +777,7 @@ class AbstractDataContext(ConfigPeer, ABC):
             )
 
         # We currently don't allow one to overwrite a datasource with this internal method
-        if datasource_name in self.datasources:
+        if self._has_datasource(datasource_name):
             raise gx_exceptions.DataContextError(
                 f"Can not write the fluent datasource {datasource_name} because a datasource of that "
                 "name already exists in the data context."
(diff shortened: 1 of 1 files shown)


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -4787,21 +4783,6 @@ def include_rendered_content(self) -> IncludeRenderedContentConfig:
...
-        """A single holder for all Datasources in this context"""
-        return self._cached_datasources
-
-    @property
-    def fluent_datasources(self) -> Dict[str, FluentDatasource]:
-        return {
-            name: ds
-            for (name, ds) in self.datasources.items()
-            if isinstance(ds, FluentDatasource)
-        }
```
Kilo59, line 4803: Thinking about it more I do think we should do this by making a `MutableMapping` or custom `UserDict`
whose `__getitem__` method uses `self._datasource_store.retrieve_by_name(name)`.
Then the logic for accessing it doesn't need to change it all.

https://realpython.com/inherit-python-dict/#subclassing-userdict-from-collections
https://treyhunner.com/2019/04/why-you-shouldnt-inherit-from-list-and-dict-in-python/
```python
class StoreDict(UserDict):

   def __item__(self, store)
        self._store = store
        super().__init__()

    def __getitem__(self, key: str):
        return self._store.retrieve_by_name(name)

    ...
```
```python

    @property
    def datasources(
        self,
    ) -> StoreDict[str, Union[LegacyDatasource, BaseDatasource, FluentDatasource]]:
        """A single holder for all Datasources in this context"""
        self._datasource_store_dict
```



Then we can keep these public properties.
And we can use generics to encode the type information of its member items.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -4787,21 +4783,6 @@ def include_rendered_content(self) -> IncludeRenderedContentConfig:
     def notebooks(self) -> NotebookConfig:
         return self.variables.notebooks  # type: ignore[return-value]
 
-    @property
-    def datasources(
```
wookasz, line 4791: I didn't think we'd get rid of this property, but instead, have it fetch the full list of datasources.

There are definitely other places in the code that depend on this property. Some of them are incorrect as they seem to mutate it, meaning that the config won't be persisted

PR #8526, opened 2023-08-10
[BUGFIX] Snowflake column identifiers 

wookasz:
A bug in the Snowflake connector causes our normal identifier wrapping logic to break Snowflake expectations.
This is resolved by converting the column identifier to `str` inside of `expectations.metrics.util.get_sqlalchemy_column_metadata()`.

- https://github.com/snowflakedb/snowflake-sqlalchemy/issues/157

Alternatively, we could conditionally convert to a `str` if the dialect is `snowflake`.

https://github.com/great-expectations/great_expectations/blob/d32c40829ba1976c015f0b07798131e6bae99dad/great_expectations/expectations/metrics/util.py#L326-L346

This fix also addresses problems when running DatabricksSQL checkpoints:

- https://github.com/great-expectations/great_expectations/pull/8523

----------


pyproject.toml
Kilo59, line 1: @wookasz I think your editor is auto-formatting this file and introducing a lot of whitespace changes.

.github/workflows/ci.yml
```diff
@@ -412,26 +415,26 @@ jobs:
...
-        python setup.py bdist_wheel
-    - name: Publish distribution to PyPI
-      uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        skip-existing: true # prevent against pushing duplicate versions
+      - uses: actions/checkout@master
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v3
+        with:
+          python-version: '3.8'
```
Kilo59, line 462: Unintentional whitespace changes?
@wookasz Are you using PyCharm or VSCode? 

To turn this off for VSCode you can add this setting (at the user or workspace level).
```json
{
    "yaml.format.enable": false
}
```

wookasz: it is, but at the same time this file is inconsistently formatted

pyproject.toml
```diff
@@ -58,118 +58,118 @@ exclude = [
...
-    'render/view/view\.py',                                                                                # 11
-    'rule_based_profiler/domain_builder/map_metric_column_domain_builder\.py',                             # 8
-    'rule_based_profiler/estimators/bootstrap_numeric_range_estimator\.py',                                # 8
-    'rule_based_profiler/estimators/kde_numeric_range_estimator\.py',                                      # 7
-    'rule_based_profiler/expectation_configuration_builder',                                               # 13
-    'rule_based_profiler/helpers/util\.py',                                                                # 46
+    'core/usage_statistics/anonymizers/batch_anonymizer\.py', # 10
+    'core/usage_statistics/anonymizers/batch_request_anonymizer\.py', # 16
+    'core/usage_statistics/anonymizers/checkpoint_anonymizer\.py', # 16
+    'core/usage_statistics/anonymizers/data_docs_anonymizer\.py', # 5
```
Kilo59: Can you revert these whitespace changes?

.github/workflows/ci.yml
Kilo59: Mostly a reminder to myself to try out one of these pre-commit hook yaml formatters
https://github.com/macisamuele/language-formatters-pre-commit-hooks

tasks.py
```diff
@@ -869,6 +869,10 @@ class TestDependencies(NamedTuple):
         services=("postgresql",),
         extra_pytest_args=("--postgresql",),
     ),
+    "snowflake": TestDependencies(
+        requirement_files=("reqs/requirements-dev-snowflake.txt",),
+        extra_pytest_args=("--snowflake",),
```
```suggestion
```
This isn't needed for `snowflake`.
These are holdovers from an old custom pytest cli flagging system we had.
It caused a lot of confusion about when tests would run or not.
The goal is to remove it completely, but until then we maintain this mapping, to insert needed flags for certain markers.

TLDR: for new markers we don't need to pass extra args.
But for the old markers, you would have to do things like .... `pytest -m postgresql --postgresql` ... and if you didn't pass the extra flag the tests would be skipped. 😬

Kilo59 (approved):
🚀

.github/workflows/ci.yml
```diff
@@ -264,6 +264,15 @@ jobs:
...
       AWS_SECRET_ACCESS_KEY: ${{secrets.AWS_SECRET_ACCESS_KEY}}
+      SNOWFLAKE_CI_ACCOUNT: ${{secrets.SNOWFLAKE_CI_ACCOUNT}}
+      SNOWFLAKE_CI_USER_PASSWORD: ${{secrets.SNOWFLAKE_CI_USER_PASSWORD}}
+      SNOWFLAKE_ACCOUNT: ${{secrets.SNOWFLAKE_ACCOUNT}}
+      SNOWFLAKE_USER: ${{secrets.SNOWFLAKE_USER}}
+      SNOWFLAKE_PW: ${{secrets.SNOWFLAKE_PW}}
+      SNOWFLAKE_DATABASE: ${{secrets.SNOWFLAKE_DATABASE}}
+      SNOWFLAKE_SCHEMA: ${{secrets.SNOWFLAKE_SCHEMA}}
+      SNOWFLAKE_WAREHOUSE: ${{secrets.SNOWFLAKE_WAREHOUSE}}
+      SNOWFLAKE_ROLE: ${{secrets.SNOWFLAKE_ROLE}}
```
Kilo59, line 286: `followup`
We should talk about the spend this test is going to cause, and ways to reduce it like only running it on the `scheduled` jobs and `release` jobs.

great_expectations/expectations/metrics/util.py
```diff
@@ -343,7 +343,9 @@ def get_sqlalchemy_column_metadata(
                     columns = table_selectable.columns().columns
             else:
                 columns = inspector.get_columns(
-                    table_selectable,
+                    str(
+                        table_selectable
+                    ),  # TODO: remove cast to a string once [this](https://github.com/snowflakedb/snowflake-sqlalchemy/issues/157) issue is resovled
```
Kilo59, line 348: - https://github.com/snowflakedb/snowflake-sqlalchemy/issues/157

Kilo59 (approved):

great_expectations/expectations/metrics/util.py
Kilo59, line 348:
```suggestion
                    # TODO: remove cast to a string once [this](https://github.com/snowflakedb/snowflake-sqlalchemy/issues/157) issue is resolved
                    table_selectable if engine.dialect.name != "snowflake" else str(table_selectable),
```

We can pull out the dialect name from the `engine` and then only do this change for `snowflake`.
That way we limit the blast radius of this change.

.github/workflows/ci.yml
wookasz: Let's not worry about the spend until we see that it is too much

great_expectations/expectations/metrics/util.py
wookasz: I want to avoid adding conditionals unless there's a clear need. Based on the tests we have, the conversion here does not seem to be negatively affecting the test cases

PR #8532, opened 2023-08-10
[FEATURE] Make SQLDatasource.assets field a `tagged-union`

Kilo59:
The Pydantic generated JSON schemas declare our `assets` as an `anyOf` field however `oneOf` is more correct. 
Using `pydantic` [tagged-union](https://docs.pydantic.dev/1.10/usage/types/#discriminated-unions-aka-tagged-unions) to encode this difference into the JSON Schema.

Limiting this PR to only SQLDatasources asset types but we likely want to apply this change to the other Datasources.

------

https://stackoverflow.com/a/34038277

> [anyOf](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-10.2.1.2):

...

>An instance validates successfully against this keyword if it validates successfully against **at least one schema** defined by this keyword's value. Note that when annotations are being collected, all subschemas MUST be examined so that annotations are collected from each subschema that validates successfully.

[oneOf](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-10.2.1.3):

...

>An instance validates successfully against this keyword if it validates successfully against **exactly one schema** defined by this keyword's value.

--------

Normally this would also result in more specific pydantic errors being thrown but we already had some custom asset type lookup logic that gave us this benefit. Using the canonical method of union discrimination may allow us to simplify our custom logic around this. Although it may not be possible to use this method for the Datasource types themselves and/or the dynamic Pandas Asset types (because we won't know them all before runtime/import-time). 

## TODO

- [ ] investigate min Airflow version compatibility with `typing_extensions` usage
- [x] fix error message tests

--------


great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -925,6 +927,9 @@ def _is_bracketed_by_quotes(target: str) -> bool:
         return False
 
 
+SQLAssets = Annotated[Union[TableAsset, QueryAsset], Field(discriminator="type")]
```
Kilo59:
```suggestion
AssetTypes = Annotated[Union[TableAsset, QueryAsset], Field(discriminator="type")]
```

Does anyone have opinions on the name of this type? 🤷

elenajdanova (approved):

Kilo59:
Using typing_extension might break our compatibility with the min airflow version.
Will investigate before merging.

great_expectations/datasource/fluent/sql_datasource.py
donaldheppner: AssetTypes is sufficiently unambiguous to me because of the module name

Kilo59: Done

DrewHoo (approved):

PR #8542, opened 2023-08-14
[MAINTENANCE] Fix conditional in CI

kenwade4:
- Put conditional statement at correct indent level and remove outer 'run'


Shinnnyshinshin (approved):
LGTM!

PR #8549, opened 2023-08-14
[MAINTENANCE] Update Docusaurus to `2.4.1` and ensure Algolia search works

Shinnnyshinshin:
# What is this for? 
* follow up to #8508 and #8546
* Update search footer element which returns a link to the "full results page". This broke the previous upgrade. 
* [Security Vulnerability in our Docusaurus packages](https://github.com/great-expectations/great_expectations/security/dependabot/109)
* Upgraded Docusaurus 2.4.1, which had additional dependencies (`@svgr/webpack": "^5.5.0"`) which had additional security vulnerabilities. 
* `nth-check` is upgraded to `2.0.1` as recommended by dependabot. 
Screenshot 2023-08-10 at 6 38 47 PM
* Deprecates `UserConfigurableProfiler`, which is a V2 feature. 
* Docstrings polish

# Is the search issue solved now? 

Yes! and here is a [Loom](https://www.loom.com/share/ee572f6117514ff2a81c685665297b53?sid=3697d414-11f9-4088-9d04-5f95da5cd045)


docs/docusaurus/package.json
```diff
@@ -16,27 +16,31 @@
...
-    "@docusaurus/plugin-sitemap": "^2.0.0-beta.21",
-    "@docusaurus/preset-classic": "^2.0.0-beta.21",
-    "@docusaurus/theme-search-algolia": "^2.0.0-beta.21",
+    "@docusaurus/core": "2.4.1",
+    "@docusaurus/plugin-google-gtag": "2.4.1",
+    "@docusaurus/plugin-sitemap": "2.4.1",
+    "@docusaurus/preset-classic": "2.4.1",
+    "@docusaurus/theme-search-algolia": "2.4.1",
     "@mdx-js/react": "^1.6.21",
-    "@svgr/webpack": "^5.5.0",
```
Shinnnyshinshin, line 28: This guy was our security issue

docs/docusaurus/src/theme/SearchBar/index.js
```diff
@@ -6,7 +6,7 @@ import { useBaseUrlUtils } from '@docusaurus/useBaseUrl'
 import Link from '@docusaurus/Link'
 import Head from '@docusaurus/Head'
 import { isRegexpStringMatch } from '@docusaurus/theme-common'
-import { useSearchPage } from '@docusaurus/theme-common/internal'
+import { useSearchPage } from '@docusaurus/theme-common'
 import { DocSearchButton, useDocSearchKeyboardEvents } from '@docsearch/react'
 import { useAlgoliaContextualFacetFilters } from '@docusaurus/theme-search-algolia/client'
 import Translate from '@docusaurus/Translate'
```
Shinnnyshinshin, line 12: post to FE community of practice. This needs updating.

docs/docusaurus/src/theme/SearchBar/index.js
```diff
@@ -23,18 +23,18 @@ function Hit ({ hit, children }) {
...
-      <Translate
-        id='theme.SearchBar.seeAll'
-        values={{ count: state.context.nbHits }}
-      >
-        {'See all {count} results'}
-      </Translate>
-    </Link>
-  )
-}
+     <Link to={makeLink()} onClick={onClose}>
```
Shinnnyshinshin: updated function to create search link for all results

docs/docusaurus/package.json
```diff
@@ -16,27 +16,31 @@
     "snippet-check": "node scripts/remark-named-snippets/snippet.js"
   },
   "dependencies": {
+    "@algolia/client-search": ">= 4.9.1 < 6",
```
dctalbot: All dependencies in this file should be pinned. We need to ensure that the documentation can build reliably and in a reproducible way. The lockfile enables this, but it's a lot easier to maintain `package.json`. This way, we only need to dig into the lockfile on rare occasions.

docs/docusaurus/package.json
```diff
@@ -16,27 +16,31 @@
...
-    "@svgr/webpack": "^5.5.0",
     "clsx": "^1.1.1",
     "docusaurus-gtm-plugin": "^0.0.2",
     "docusaurus-plugin-sass": "^0.2.2",
     "plugin-image-zoom": "ataft/plugin-image-zoom",
-    "react": "^17.0.1",
-    "react-dom": "^17.0.1",
+    "react": "^16.8.4",
+    "react-dom": "^16.8.4",
+    "react-loadable": "*",
```
dctalbot: Same as above, we should pin this

docs/docusaurus/package.json
```diff
@@ -16,27 +16,31 @@
...
+    "react-router-dom": "^5.1.2",
     "react-select": "^4.3.0",
     "remark-code-import": "^0.3.0",
     "sass": "^1.56.1",
     "search-insights": "^2.2.3",
-    "standard": "^17.1.0"
+    "standard": "^17.1.0",
+    "typescript": ">= 2.7",
+    "webpack": "^4.36.0 || ^5.0.0"
```
dctalbot: I don't really understand how the docs were being built for this change. Was webpack a transitive dependency of something else? Same as above, we should pin this. If we can use webpack 5 that would be great.

docs/docusaurus/package.json
```diff
@@ -16,27 +16,31 @@
...
+    "react-dom": "^16.8.4",
+    "react-loadable": "*",
+    "react-router-dom": "^5.1.2",
     "react-select": "^4.3.0",
     "remark-code-import": "^0.3.0",
     "sass": "^1.56.1",
     "search-insights": "^2.2.3",
-    "standard": "^17.1.0"
+    "standard": "^17.1.0",
+    "typescript": ">= 2.7",
```
dctalbot: Same as above, we should pin this.

great_expectations/profile/user_configurable_profiler.py
```diff
@@ -38,12 +40,14 @@
     from great_expectations.data_context.data_context import AbstractDataContext
 
 
-@public_api
+@deprecated_method_or_class(
```
dctalbot: Is this change relevant for this PR? It seems like it should be broken into its own PR. We shouldn't bundle extra changes in with the dependency upgrade in case we need to revert it. Same question on the other files.

dctalbot (changes requested):
Nice, thanks for working on this. I think we can prevent future headaches by pinning some of these new dependencies. FWIW it looks like we aren't impacted by that vulnerability, so urgency is not high on getting this merged.

docs/docusaurus/package.json
Shinnnyshinshin: good question:

`svgr/webpack` is a dependency of `@docusaurus/utils`, which we need. 

The security issue is `nth-check`, which is a dependency of `webpack` through `css-select`. 


```
svgr/webpack@5.5.0` requires `nth-check@^1.0.2` via a transitive dependency on `css-select@2.1.0`
```

With the upgrade, `nth-check` now has version `2.0.1` (from `yarn.lock`), which addresses the vulnerability :) 
```
css-select@^4.1.3:
  version "4.3.0"
  resolved "https://registry.yarnpkg.com/css-select/-/css-select-4.3.0.tgz#db7129b2846662fd8628cfc496abb2b59e41529b"
  integrity sha512-wPpOYtnsVontu2mODhA19JrqWxNsfdatRKd64kmpRbQgh1KtItko5sTnEpPdpSaJszTOhEMlF/RPz28qj4HqhQ==
  dependencies:
    boolbase "^1.0.0"
    css-what "^6.0.1"
    domhandler "^4.3.1"
    domutils "^2.8.0"
    nth-check "^2.0.1"
```

docs/docusaurus/package.json
Shinnnyshinshin: done

great_expectations/profile/user_configurable_profiler.py
Shinnnyshinshin: good point. i'll separate this into a follow-up PR

dctalbot (approved):
Love it, thank you for fixing this!

PR #8552, opened 2023-08-15
[MAINTENANCE] ruff 0.0.284

Kilo59:
Update from `ruff 0.0.282` -> `0.0.284`
Fix or temporarily ignore new warnings.
See below 👇 for a summary of the new warnings `ruff 0.0.284` produces.  

```console
$ ruff . --statistics
76	E721  	[ ] Do not compare types, use `isinstance()`
20	PYI041	[ ] Use `float` instead of `int | float`
11	UP032 	[*] Use f-string instead of `format` call
 7	PYI024	[ ] Use `typing.NamedTuple` instead of `collections.namedtuple`
 1	PYI013	[*] Non-empty class body must not contain `...`
 1	PYI019	[ ] Methods like `add_sorters` should return `typing.Self` instead of a custom `TypeVar`
 1	PYI032	[*] Prefer `object` to `Any` for the second parameter to `__eq__`
 1	PYI034	[ ] `__new__` methods in classes like `MetaDatasource` usually return `self` at runtime
 1	PYI046	[ ] Private protocol `_Splitter` is never used
```


docs_rtd/feature_annotation_parser.py
```diff
@@ -34,7 +34,9 @@
     "expectation_completeness",
 ]
 
-AnnotatedNode = namedtuple("AnnotatedNode", ["name", "path", "annotation", "type_"])
+AnnotatedNode = namedtuple(  # noqa: PYI024
+    "AnnotatedNode", ["name", "path", "annotation", "type_"]
+)
```
Kilo59: @Shinnnyshinshin this class doesn't appear to be used anywhere in our codebase.
Can I delete it or should this be used somewhere?

great_expectations/agent/message_service/subscriber.py
```diff
@@ -133,7 +133,10 @@ def _on_message_handler(
         return on_message(event_context)
 
     async def _redeliver_message(
-        self, delivery_tag: int, requeue: bool = True, delay: Union[float, int] = 3
+        self,
+        delivery_tag: int,
+        requeue: bool = True,
+        delay: Union[float, int] = 3,  # noqa: PYI041
```
Kilo59, line 139: >In Python, `int` is a subtype of `float`, and `float` is a subtype of `complex`. As such, a union that includes both `int` and `float` is redundant, as it is equivalent to a union that only includes `float`.

T.I.L. 🤯 


https://beta.ruff.rs/docs/rules/redundant-numeric-union/

There could still be value in keeping the annotations as they are, our users may not know that an `int` is a `float`. I certainly didn't.

docs_rtd/feature_annotation_parser.py
Shinnnyshinshin: I think this should be good to delete :)

Kilo59:
```suggestion
```

NathanFarmer (approved):

PR #8558, opened 2023-08-15
[BUGFIX]Fix Agent error types

elenajdanova:

### DIFF
--- a/great_expectations/agent/actions/data_assistants/utils.py
+++ b/great_expectations/agent/actions/data_assistants/utils.py
@@ -60,7 +60,7 @@ def build_action_result(
         context.get_expectation_suite(expectation_suite_name=expectation_suite_name)
         # if that didn't error, this name exists, so we add the timestamp
         expectation_suite_name = f"{expectation_suite_name} {timestamp}"
-    except (StoreBackendError, ValueError, DataContextError):
+    except (DataContextError):
         # resource is unique
         pass
 
@@ -68,7 +68,7 @@ def build_action_result(
         context.get_checkpoint(name=checkpoint_name)
         # if that didn't error, this name exists, so we add the timestamp
         checkpoint_name = f"{checkpoint_name} {timestamp}"
-    except (StoreBackendError, ValueError, DataContextError):
+    except (DataContextError):
         # resource is unique
         pass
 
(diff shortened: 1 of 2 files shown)


wookasz (approved):

great_expectations/agent/actions/data_assistants/utils.py
```diff
@@ -59,15 +60,15 @@ def build_action_result(
         context.get_expectation_suite(expectation_suite_name=expectation_suite_name)
         # if that didn't error, this name exists, so we add the timestamp
         expectation_suite_name = f"{expectation_suite_name} {timestamp}"
-    except (StoreBackendError, ValueError):
+    except (StoreBackendError, ValueError, DataContextError):
```
cdkini: I'm a little wary of adding more types to this since the more granular we are, the safer this bit of code is.

@joshua-stauffer @wookasz @Kilo59 do you happen to know what situations cause each of these to arise? Should we document this? 

I'm surprised `StoreBackendError` is being propagated up to the context level - it should probably be caught by the logic in the store or context and reraised as something more user friendly (just an observation not something we need to deal with here).

I'm okay with this change just to get the agent up and running but wanted to point it out since `try/except` with a `pass` can make debugging difficult down the line.

joshua-stauffer: thanks @cdkini, i agree - originally, the exception that GX raised here was `StoreBackendError`. Then last month it started raising `ValueError` instead, so we started catching that as well. My understanding was that the change to `ValueError` was a mistake, and that we would be reverting back to `StoreBackendError`. Looks like it's now been updated to the parent class of `StoreBackendError`.

I don't have a strong opinion about which error is the right one for GX to raise, but we should only (need to) catch one.

Kilo59 (changes requested):
Sounds like this is not the change we want to make.

I would suggest closing this PR and adding this work to the backlog of one of our current epics.

Unless we just want to temporarily make this change for tomorrow's release. I'd be okay with that.

Kilo59 (approved):
Changing to approve in order to ensure this isn't still broken on tomorrow's release.
But I agree with everything being said by @cdkini and @joshua-stauffer

cdkini (approved):

PR #8562, opened 2023-08-16
[CONTRIB] fix query template values metric

itaise:
Hi,
The query template values metric (that we contributed) is sometime getting INT values in the values of the dict, which fails the .format() function - becuae INT has no format function.
I casted the value to str for protection.
I checked it with our expectations and it solve the issue.
can you please check?

Thanks!


Shinnnyshinshin (approved):
Looks good @itaise :) Thank you for catching this

PR #8570, opened 2023-08-17
[BUGFIX] Query Template not always String

Shinnnyshinshin:
Follow up to #8562 where we need to consider both `int` and `sa.text()` values to be a query template. Adds conversion in cases where the input is `int`

@itaise could you check if this works for your Expectations? 

### DIFF
--- a/great_expectations/expectations/metrics/query_metrics/query_template_values.py
+++ b/great_expectations/expectations/metrics/query_metrics/query_template_values.py
@@ -28,7 +28,7 @@ class QueryTemplateValues(QueryMetricProvider):
     @classmethod
     def get_query(cls, query, template_dict, selectable):
         template_dict_reformatted = {
-            k: str(v).format(active_batch=selectable)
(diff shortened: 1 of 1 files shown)


great_expectations/expectations/metrics/query_metrics/query_template_values.py
```diff
@@ -28,7 +28,10 @@ class QueryTemplateValues(QueryMetricProvider):
     @classmethod
     def get_query(cls, query, template_dict, selectable):
         template_dict_reformatted = {
-            k: str(v).format(active_batch=selectable) for k, v in template_dict.items()
+            k: sa.text(v).format(active_batch=selectable)
+            if isinstance(v, int)
```
Kilo59, line 32: Do we care about floats?
If so you can just check for float. Int is a subclass of float.

Kilo59 (approved):
Approved with a question

PR #8592, opened 2023-08-19
[MAINTENANCE] Add implementation for CloudDataStore.add

kenwade4:


great_expectations/experimental/metric_repository/cloud_data_store.py
```diff
@@ -11,11 +13,41 @@
 
 
 class CloudDataStore(DataStore[StorableTypes]):
+    def __init__(self, context: AbstractDataContext):
+        super().__init__(context=context)
+        config = self._context._cloud_config
```
anthonyburdi: How about using the public cloud config accessor `ge_cloud_config()`?

great_expectations/experimental/metric_repository/cloud_data_store.py
```diff
@@ -11,11 +13,41 @@
...
+                    "id": str(metric.id),
+                    "batch_id": metric.batch.id,
+                    "data_asset_id": str(metric.batch.data_asset.id),
+                    "metric_name": metric.metric_name,
+                    "exception_message": metric.exception.exception_message,
+                    "exception_type": metric.exception.exception_type,
+                    "value": metric.value,
+                    "metric_type": type(metric).__name__,
+                    "value_type": type(metric.value).__name__,
+                    "value_parameter_names": [],
```
anthonyburdi, line 41: If we keep this field we will need to populate it here

great_expectations/experimental/metric_repository/cloud_data_store.py
```diff
@@ -11,11 +13,41 @@
...
+                }
+            )
+        return {"data": {"attributes": result}}
+
     def add(self, value: T) -> T:
```
anthonyburdi, line 46: Currently the method signature here is meant to take in any type and add it (e.g. not just MetricRun). In that case maybe we look up a serializer based on the thing being serialized (metric run) rather than the cloud data store knowing how to serialize a MetricRun. That might be overkill for now. Maybe instead we can rename the methods in here to `add_metric_run` and `_construct_metric_run_json_payload()` so we can make it more generic without breaking the callers later. What do you think?

anthonyburdi (commented):
Added some thoughts, happy to discuss.

great_expectations/experimental/metric_repository/cloud_data_store.py
anthonyburdi: After thinking about this more, I think we will be removing this field.

great_expectations/experimental/metric_repository/cloud_data_store.py
```diff
@@ -11,11 +13,41 @@
...
 class CloudDataStore(DataStore[StorableTypes]):
+    def __init__(self, context: AbstractDataContext):
+        super().__init__(context=context)
+        config = self._context._cloud_config
+        self._session = create_session(access_token=config.access_token)
+        self._url = (
+            f"{config.base_url}/organizations/{config.organization_id}/metric-runs"
+        )
+
+    def _construct_json_payload(self, metric_run: MetricRun):
```
anthonyburdi: Since metric_run is a pydantic model can we use the `dict()` method to create most of the payload?

great_expectations/experimental/metric_repository/cloud_data_store.py
```diff
@@ -11,11 +13,41 @@
...
+            "id": str(metric_run.id),
+            "metrics": [],
+        }
+        for metric in metric_run.metrics:
+            result["metrics"].append(
+                {
+                    "id": str(metric.id),
+                    "batch_id": metric.batch.id,
+                    "data_asset_id": str(metric.batch.data_asset.id),
+                    "metric_name": metric.metric_name,
```
anthonyburdi, line 35: I've been thinking that maybe we can get rid of metric_name since we have metric_type. If we need the name we can associate it with the type on either side of the API but no need to send it in the payload. What do you think?

great_expectations/experimental/metric_repository/cloud_data_store.py
```diff
@@ -11,11 +13,41 @@
...
+        result = {
+            "id": str(metric_run.id),
+            "metrics": [],
+        }
+        for metric in metric_run.metrics:
+            result["metrics"].append(
+                {
+                    "id": str(metric.id),
+                    "batch_id": metric.batch.id,
+                    "data_asset_id": str(metric.batch.data_asset.id),
```
anthonyburdi, line 34: Let's move `data_asset_id` up out of metric and into metric_run.

great_expectations/experimental/metric_repository/cloud_data_store.py
kenwade4: A lot of the data returned when calling dict() is not needed and some is not serializable. The uuid fields still need to be cast to str.

great_expectations/experimental/metric_repository/cloud_data_store.py
kenwade4: The metric_type of TableMetric is the same for both table.columns and table.row_count. I don't see anything wrong with keeping the metric_name included.

great_expectations/experimental/metric_repository/cloud_data_store.py
kenwade4: Removing `value_parameter_names`?

great_expectations/experimental/metric_repository/cloud_data_store.py
kenwade4: I was thinking about checking the type of T and calling the _construct if it's a MetricRun and just saving the organization url in __init__

kenwade4: I also thought about returning the response.json() of the POST instead of the value object that was passed to it.

great_expectations/experimental/metric_repository/cloud_data_store.py
anthonyburdi: Exactly. There are a few other changes in the confluence doc I sent earlier.

great_expectations/experimental/metric_repository/cloud_data_store.py
anthonyburdi: Good point and good catch 🙇

great_expectations/experimental/metric_repository/cloud_data_store.py
anthonyburdi: If we check the type of T then each time we add a new data type (besides MetricRun) we will have to modify the CloudDataStore.

anthonyburdi: What about using something like model.model_dump_json(...) https://docs.pydantic.dev/latest/usage/serialization/#modelmodel_dump_json to serialize to json? So we can hand off the serialization responsibilities to pydantic and then as long as we get a pydantic model as the value (one of the StorableTypes, mypy will help us check those) then we can just call the model_dump_json() method.

anthonyburdi: My mistake - we are using pydantic <2.0 currently until we can upgrade. But there is a `.json()` method we can use.

anthonyburdi:
Implementation in https://github.com/great-expectations/great_expectations/pull/8656

PR #8593, opened 2023-08-19
[MAINTENANCE] Add mode param to get_context

abegong:
Adds a `mode` parameter to `get_context`, so that users can explicitly set the mode in which they want to initialize a DataContext.


great_expectations/util.py
```diff
@@ -1804,6 +1807,7 @@ def get_context(  # noqa: PLR0913
     cloud_access_token: str | None = None,
     cloud_organization_id: str | None = None,
     cloud_mode: bool | None = None,
+    mode: str | None = None,
```
Kilo59:
```suggestion
    mode: Literal["ephemeral", "file", "cloud"] | None = None,
```
This ensures better auto-complete and IDE help

great_expectations/util.py
```diff
@@ -1746,6 +1746,7 @@ def get_context(  # type: ignore[misc] # overlapping overload false positive?  #
     cloud_access_token: None = ...,
     cloud_organization_id: None = ...,
     cloud_mode: Literal[False] | None = ...,
+    mode: str | None = None,
```
Kilo59:
```suggestion
    mode: Literal["file"] | None = ...,
```

great_expectations/util.py
```diff
@@ -1764,6 +1765,7 @@ def get_context(  # noqa: PLR0913
     cloud_access_token: str | None = ...,
     cloud_organization_id: str | None = ...,
     cloud_mode: Literal[True] = ...,
+    mode: str | None = None,
```
Kilo59:
```suggestion
    mode: Literal["cloud"] | None = ...,
```

great_expectations/util.py
```diff
@@ -1782,6 +1784,7 @@ def get_context(  # noqa: PLR0913
     cloud_access_token: str | None = ...,
     cloud_organization_id: str | None = ...,
     cloud_mode: bool | None = ...,
+    mode: str | None = None,
```
Kilo59:
```suggestion
    mode: None = ...,
```

great_expectations/util.py
abegong: ty!

great_expectations/util.py
abegong: I don't know this convention. Cargo culting for now.

great_expectations/util.py
```diff
@@ -1899,6 +1905,37 @@ def get_context(  # noqa: PLR0913
     """
     project_config = _prepare_project_config(project_config)
 
+    if mode is not None:
```
donaldheppner: let's merge this parameter with the logic below to a) avoid duplication, and b) handle cases when one of the _get_***_context methods fails to return (which should raise a ValueError)

donaldheppner (changes requested):

great_expectations/util.py
```diff
@@ -1899,6 +1905,37 @@ def get_context(  # noqa: PLR0913
...
+                cloud_mode=cloud_mode,
+                ge_cloud_base_url=ge_cloud_base_url,
+                ge_cloud_access_token=ge_cloud_access_token,
+                ge_cloud_organization_id=ge_cloud_organization_id,
+                ge_cloud_mode=ge_cloud_mode,
+            )
+        else:
+            raise ValueError(
+                f"Unknown mode {mode}. Please choose one of: ephemeral, file, cloud."
+            )
```
cdkini: As is, this is a breaking change for users - if we want to make this required, we should warn before making it mandatory

cdkini: jk didn't realize `if mode is not None:`

great_expectations/util.py
```diff
@@ -1899,6 +1905,37 @@ def get_context(  # noqa: PLR0913
     """
     project_config = _prepare_project_config(project_config)
 
+    if mode is not None:
+        if mode == "ephemeral":
+            return _get_ephemeral_context(
```
cdkini: All these `_get` methods have nullable returns - we need to ensure we return that type if the mode is requested

great_expectations/util.py
Kilo59: @donaldheppner just addressed this in [79f2c39](https://github.com/great-expectations/great_expectations/pull/8593/commits/79f2c3903ac8ce6221928b9f80338b37b0198317)

Kilo59 (approved):

tests/data_context/test_get_data_context.py
```diff
@@ -248,6 +248,12 @@ def test_get_context_with_no_arguments_returns_ephemeral_with_sensible_defaults(
     assert context.config.stores == defaults.stores
 
 
+@pytest.mark.unit
+def test_get_context_with_mode_equals_ephemeral_returns_ephemeral_data_context():
```
donaldheppner, line 252: please add tests for the other modes

donaldheppner (changes requested):

great_expectations/util.py
```diff
@@ -1896,7 +1896,7 @@ def get_context(  # noqa: PLR0913
         "file": dict(
             project_config=project_config,
             context_root_dir=context_root_dir,
-            project_root_dir=project_root_dir,
+            project_root_dir=project_root_dir or Path.cwd(),
```
cdkini, line 1899: required if `context = gx.get_context(mode="file")` is to work without input args - we should assume that if a user is requesting a `FileDataContext` and isn't supplying a `project_root_dir`, they're using cwd

Kilo59 (approved):

donaldheppner (approved):

PR #8604, opened 2023-08-22
[MAINTENANCE] Pydantic 2.0 support

### LINKED ISSUE #8259: Pydantic V2 support
Hi,

Are there any plans in the near future to support [Pydantic V2](https://github.com/pydantic/pydantic/releases/tag/v2.0)?

NathanFarmer:
Add pydantic compatibility module and update imports

Migration guide for reference:
[https://docs.pydantic.dev/dev-v2/migration/](https://docs.pydantic.dev/dev-v2/migration/)

- Fixes #8259 

Note: this PR does not aim to fully migrate to pydantic v2 patterns but simply to remove the upper bound on v2 so that downstream dependencies and users can install pydantic v2 along with `great_expectations`.

-----


ghost:
<details open>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=8604&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

####

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=8604&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map legend
</details>

great_expectations/compatibility/pydantic.py
```diff
@@ -0,0 +1,60 @@
+import pydantic
+
+from great_expectations.compatibility.not_imported import (
+    is_version_greater_or_equal,
+)
+
+if is_version_greater_or_equal(version=pydantic.VERSION, compare_version="2.0.0"):
+    # from pydantic.v1 import BaseModel, Field, StrictStr
+    from pydantic.v1 import *  # noqa: F403 # type: ignore[assignment]
```
Kilo59, line 9: TODO: don't use star imports

.github/workflows/ci.yml
```diff
@@ -57,13 +57,12 @@ jobs:
...
       - run: invoke fmt --check
-      - name: Marker-coverage-check
-        run: |
-          invoke marker-coverage
       - name: Type-check
         run: |
           invoke type-check --ci --pretty
           invoke type-check --ci --pretty  --check-stub-sources
+      - name: Marker-coverage-check
+        run: invoke marker-coverage
```
Kilo59, line 65: I swapped the order of these to get faster feedback on typing issues

.github/workflows/ci.yml
```diff
@@ -405,6 +404,24 @@ jobs:
       - name: Run the tests
         run: invoke ci-tests -m unit --xdist --slowest=10  --timeout=2.0
 
+  pydantic-v1:
```
Kilo59, line 407: Added a test for our min version of pydantic now that the other tests will be installing v2

Kilo59 (approved):

ci/constraints-test/pydantic-v1-install.txt
```diff
@@ -0,0 +1,3 @@
+# this is higher than our min version but is the earliest version that actually works
+# TODO: update our min version or fix gx for our current min version
+pydantic==1.10.8
```
Kilo59, line 3: Our min version of pydantic in our requirements file is `1.9` but gx actually breaks at import time on this version, and it looks like we were never testing against this version in our CI.
This does not appear related to moving to pydantic v2.

I will create followup tickets for taking the action here.

great_expectations/compatibility/pydantic.py
```diff
@@ -0,0 +1,56 @@
...
+    "fields",
+    "GenericModel",
+    "generics",
+    "json",
+    "ModelMetaclass",
+    "networks",
+    "schema",
+    "typing",
+    "UrlError",
+]
```
NathanFarmer, line 56: Shouldn't we update the `__init__.py` files instead of doing this?

great_expectations/compatibility/pydantic.py
```diff
@@ -0,0 +1,56 @@
...
+        UrlError,
+        error_wrappers,
+        errors,
+        fields,
+        generics,
+        json,
+        networks,
+        schema,
+        typing,
+    )
```
NathanFarmer, line 21: How does it work if I'm creating a V2 Base Model and I want to use one of these classes? I guess we can cross that bridge when we get there.

great_expectations/compatibility/pydantic.py
Kilo59: @NathanFarmer why?

This controls what gets exported when doing a star import and prevents the `ruff` unused import rule from auto-removing everything.

But unlike Javascript, everything in python modules is "exported" and available for import from other modules.
`__all__` only controls what `import *` does, but doesn't prevent normal imports from accessing things outside of `__all__`.

great_expectations/compatibility/pydantic.py
Kilo59: That won't be possible yet, without some conditional imports.

We would need some kind of `pydantic_v2` compatibility module that works like our other optional dependencies (`sqlalchemy`).
Unless we just decide to drop pydantic v1, but I think we need to let the ecosystem catch up before we can do that.

great_expectations/compatibility/pydantic.py
NathanFarmer: I did not know it was related to import *!

contrib/cli/requirements.txt
```diff
@@ -2,7 +2,7 @@ black[jupyter]==23.3.0 # Linting / code style
 Click>=7.1.2         # CLI tooling
 cookiecutter==2.1.1  # Project templating
 mypy==1.5.1          # Type checker
-pydantic>=1.0,<2.0    # Needed for mypy plugin
+pydantic>=1.0        # Needed for mypy plugin
```
billdirks, line 5: Should make me 1.10.8?

Kilo59: Actually, I don't understand why we have this entry in cli/requirements at all.
It being in our core requirements should be enough.

PR #8621, opened 2023-08-25
[MAINTENANCE] add py311 min versions tests

Kilo59:
- Start testing min versions with python 311
- use xdist for min version tests (for faster test)
- Increase the allowed `timeout` for min version tests (earlier versions of dependencies may not be as performant)
- ~change marker on "slow" `test_onboarding_data_assistant_should_fail_forward` from `unit` to `big`~


tests/rule_based_profiler/data_assistant/test_onboarding_data_assistant.py
```diff
@@ -326,7 +326,7 @@ def test_onboarding_data_assistant_result_batch_id_to_batch_identifier_display_n
     )
 
 
-@pytest.mark.unit
+@pytest.mark.big  # marked as big because this test is regularly taking over 1.5s in CI
```
Kilo59:
```suggestion
@pytest.mark.unit
```

ci/constraints-test/py311-min-install.txt
```diff
@@ -0,0 +1,3 @@
+numpy==1.25.2
```
Kilo59: Going to decrease this until it starts failing.

Will merge once I find the min version.

Kilo59: `1.23` seems to be the oldest version of `numpy` that works on `3.11`

Shinnnyshinshin (approved):
🚀

PR #8647, opened 2023-08-31
[MAINTENANCE] Automatically emit deprecation warnings when decorating deprecated args

cdkini:


great_expectations/core/_docs_decorators.py
```diff
@@ -126,18 +127,22 @@ class MyClass:
...
+                argument_name=argument_name,
+                text=text,
+            )
 
-        return _add_text_below_function_docstring_argument(
-            func=func,
-            argument_name=argument_name,
-            text=text,
-        )
+        return wrapper
```
Kilo59, line 143: I would suggest raising the deprecation warning before the wrapped function is called.

Issue #8387, opened 2023-07-24
Incorrect result file referencing on Windows


mspaulik:
**Describe the bug**

When executing Great Expectations on Windows, index.html incorrectly references the filepath inside results folder. The key '_table_row_link_path' in the file contains link to the path which has **'%5C'** symbol in the filepath instead of regular **'/'.** Thus, the file cannot be found.

Example of code snippet from index.html :

Object.assign(
{
columns: [.....]
data: [ "_table_row_link_path" : 'validations%5CData_Suite%5C6567%5C20230724T080927.1287666Z%5C123.html']
},
toolbar: [....]
)

Thus, when trying to open these results in index.html file, it give error that: "Hmmmmm.... cant reach page."

The output folder structure created is:
```
- great-expectations/runResults/
     - static/
     - validations/
         - Data_Suite /
             - 6567 /
                 - 20230724T080927.1287666Z/
                     - 123.html
     - index.html
```

**To Reproduce**

```
yaml = YAMLHandler()
context = gx.get_context()
datasource_config = {
  "name" :  "customDatasource",
  "class_name" : "Datasouce",
  "execution_engine" : { "class_name": "PandasExecutionEngine"},
  "data_connectors": {
     "DefaultRuntimeDataConnector": {
          "class_name" : "RuntimeDataConnector",
          "batch_identifiers" : ["default_batch"],
      }
  }
}

context.test_yaml_config(yaml.dump(datasource_config))
context.add_datasource(**datasource_config)

batch_request = RuntimeBatchRequest(
  datasource_name = "customDatasource",
  data_connector_name = "DefaultRuntimeDataConnector",
  data_asset_name = "custom_data_asset",
  runtime_parameters = {"path": datafile },
  batch_identifiers = {"default_batch": "default_batch"},
  batch_spec_passthrough = batch_spec_passthrough 
)

validator = context.get_validator{
  batch_request = batch_request,
  expectations_suite_name = rulesfile,
}

checkpoint = gx.checkpoint.SimpleCheckpoint(
  name = "default_checkpoint",
  data_context = context,
  validator = validator,
  run_name_template = custom_id
)

checkpoint_result = checkpoint.run()
```

config_variables.yml

`BASE_DIRECTORY: '..\runResults'`

MarcelBeining:
@HaebichanGX I have a fix for this error, but it seems I cannot push a new branch to this repo to be able to start a Pull Request (error 403 after authorization). Do you need to authorize me for this?

[...]

PR #8663, opened 2023-09-05
[BUGFIX] Fix backslashes in url path under Windows

MarcelBeining:
- [x] Fixes the bug described here [#8387]
- [ ] Appropriate tests and docs have been updated (None found)


MarcelBeining:
@cla-bot check

MarcelBeining:
@cla-bot check

great_expectations/render/renderer/site_builder.py
```diff
@@ -602,18 +602,20 @@ def add_resource_info_to_index_links_dict(  # noqa: PLR0913
...
-                    *expectation_suite_name.split("."),
-                    *run_id.to_tuple(),
-                    batch_identifier,
+                "/".join(
+                    [
+                        "validations",
+                        *expectation_suite_name.split("."),
+                        *run_id.to_tuple(),
+                        batch_identifier,
+                    ]
```
Kilo59: Please use `pathlib.Path` to construct the path here, rather than using raw strings.

Kilo59 (changes requested):
Thanks for this fix.
The canonical way to deal with this issue is to use `pathlib.Path` instead of creating string paths by hand.

great_expectations/render/renderer/site_builder.py
Kilo59:
```suggestion
                str(pathlib.Path(
                    *[
                        "validations",
                        *expectation_suite_name.split("."),
                        *run_id.to_tuple(),
                        batch_identifier,
                    ]
                 ))
```

great_expectations/render/renderer/site_builder.py
```diff
@@ -602,18 +602,20 @@ def add_resource_info_to_index_links_dict(  # noqa: PLR0913
...
                 )
                 + ".html"
             )
         else:
             filepath = (
-                os.path.join(  # noqa: PTH118
-                    "expectations", *expectation_suite_name.split(".")
+                "/".join(
+                    ["expectations", *expectation_suite_name.split(".")]
```
Kilo59:
```suggestion
                    str(pathlib.Path(*["expectations", *expectation_suite_name.split(".")]))
```

great_expectations/render/renderer/site_builder.py
MarcelBeining: @Kilo59 The way you did it, will result in the same error on Windows as the bugfix should fix ;-) This is because `str(Path)` will again introduce backslashes on Windows. If you insist on using pathlib it has to be changed to
```
                pathlib.Path(
                    *[
                        "validations",
                        *expectation_suite_name.split("."),
                        *run_id.to_tuple(),
                        batch_identifier,
                    ]
                 ).as_posix()
```

Kilo59 (approved):

great_expectations/render/renderer/site_builder.py
Kilo59: @MarcelBeining good catch, I just approved.

MarcelBeining:
Seems `@cla-bot check` does not work

NathanFarmer:
@MarcelBeining while we are figuring out the CLA issue, can you look into why this is failing Databricks tests? At first glance it seems to have something to do with the pydantic validation in the `DatabricksDsn` class at `great_expectations/datasource/fluent/databricks_sql_datasource.py`. Possibly something needs to change in `_parse_param_from_query_string`?

Shinnnyshinshin:
> @MarcelBeining while we are figuring out the CLA issue, can you look into why this is failing Databricks tests? At first glance it seems to have something to do with the pydantic validation in the `DatabricksDsn` class at `great_expectations/datasource/fluent/databricks_sql_datasource.py`. Possibly something needs to change in `_parse_param_from_query_string`?

I think this was an issue with the databricks tests being run in contributor forks, which (I think) was addressed in PR #8662

austiezr:
@cla-bot check

PR #8669, opened 2023-09-06
[MAINTENANCE] Column Descriptive Metrics: Add tests and cleanup

anthonyburdi:
- Pass through the data_asset_id
- Add tests for each of the major components of the Column Descriptive Metrics implementation that didn't already have them.
- Add docstrings.
- Remove demo.py
- Note: Passing through exceptions is split into a separate ticket DX-749 and the tests and implementation have not been completed in this PR.


tests/experimental/metric_repository/test_batch_inspector.py
```diff
@@ -0,0 +1,77 @@
...
+
+pytestmark = pytest.mark.unit
+
+
+def test_compute_metric_run_with_no_metric_retrievers():
+    mock_context = Mock(spec=CloudDataContext)
+    batch_inspector = BatchInspector(context=mock_context, metric_retrievers=[])
+    mock_batch_request = Mock(spec=BatchRequest)
+
+    data_asset_id = uuid.uuid4()
```
JennyTee, line 27: In past orgs I've worked at we tried to avoid introducing randomness into unit tests (e.g., generating random ids) in an attempt to make the tests deterministic. So, for example, instead of using `uuid.uuid4()` we would hard-code the value of the `data_asset_id` (or stub out a method like `generateFakeUuid4()` that always returns the same UUID) so that the same ID would be used in subsequent test runs.

In my opinion this is a pretty nitpicky thing and I'm totally fine continuing to use randomly-generated uuids in unit tests--just wanted to bring it up as food for thought/discussion.

JennyTee (approved):

tests/experimental/metric_repository/test_batch_inspector.py
anthonyburdi: Thank you for bringing this up! Good point for discussion. I have used both approaches in the past. In a conversation with @donaldheppner recently he made the point that we aren't testing a specific uuid here but that the same uuid (data_asset_id) ends up as part of the returned `MetricRun`. Its a subtle distinction, that this test should be able to run with any uuid. I don't think this works with all forms of random data, but entity ids seem to fit that rationale. I could also have totally misinterpreted Don's point though, but at least you know my thought process.

anthonyburdi: Also - if we decide on a consistent go forward approach I'm happy to conform :)

JennyTee: Cool, tks for the additional info.  I think it's fine to leave as-is

PR #8682, opened 2023-09-07
[BUGFIX] Missing DataFrame for in-memory Data Assets using GX Cloud

NathanFarmer:
Fixing the ability to get a `Validator` from a `DataFrameAsset` using a `CloudDataContext`. The DataFrame was being dropped from the Data Asset during round trip to Cloud DB.


great_expectations/core/datasource_dict.py
```diff
@@ -77,6 +89,10 @@ def __contains__(self, name: object) -> bool:
     def __setitem__(self, name: str, ds: FluentDatasource | BaseDatasource) -> None:
         config: FluentDatasource | DatasourceConfig
         if isinstance(ds, FluentDatasource):
+            if isinstance(ds, SupportsInMemoryDataAssets):
+                for asset in ds.assets:
```
cdkini, line 104: Could we combine the `if` statements?

great_expectations/core/datasource_dict.py
```diff
@@ -77,6 +89,10 @@ def __contains__(self, name: object) -> bool:
...
+                    if asset.type == "dataframe":
```
cdkini: Do we have an enum for this?

great_expectations/core/datasource_dict.py
```diff
@@ -108,7 +124,16 @@ def __getitem__(self, name: str) -> FluentDatasource | BaseDatasource:
 
         ds = self._datasource_store.retrieve_by_name(name)
         if isinstance(ds, FluentDatasource):
-            return self._init_fluent_datasource(ds)
+            hydrated_ds = self._init_fluent_datasource(ds)
+            if isinstance(hydrated_ds, SupportsInMemoryDataAssets):
+                for asset in hydrated_ds.assets:
+                    if asset.type == "dataframe":
+                        cached_data_asset = self._in_memory_data_assets.get(
+                            f"{name}-{asset.name}"
```
cdkini: Could we encapsulate this f-string name creation into a helper method? Same note about type comparison with enum

tests/datasource/fluent/test_pandas_datasource.py
```diff
@@ -476,6 +476,27 @@ def test_dataframe_asset(
...
+        dataframe_asset = datasource.add_dataframe_asset(name=dataframe_asset_name)
+    except ValueError:
+        dataframe_asset = datasource.get_asset(asset_name=dataframe_asset_name)
+
+    dataframe_asset.build_batch_request(dataframe=test_df_pandas)
+
+    assert dataframe_asset.dataframe.equals(
+        context.datasources[datasource_name].assets[0].dataframe  # type: ignore[union-attr]
+    )
+
```
cdkini: Is this an E2E test? Can we avoid `try/except` logic and instead set up guaranteed prereq state?

NathanFarmer: Yes, it is a Cloud integration test. We can avoid it by just using `get`, but the codepaths are different so I did have to add 2 different changes for each codepath.

great_expectations/core/datasource_dict.py
NathanFarmer: Not that I'm aware of. I'm not sure how we would keep it in sync with the data assets. We probably should defer that work.

great_expectations/core/datasource_dict.py
NathanFarmer: I encapsulated the f-string into a method and created a new protocol for `InMemoryDataAsset`s.

great_expectations/core/datasource_dict.py
NathanFarmer: No because it's possible a datasource `SupportsInMemoryDataAssets`, but doesn't actually have any or only some of them are `InMemoryDataAsset`s. We don't want to iterate through all assets unless it makes sense to.

tests/datasource/fluent/test_pandas_datasource.py
NathanFarmer: Updated to use just the `get` commands.

tests/datasource/fluent/test_pandas_datasource.py
```diff
@@ -476,6 +476,21 @@ def test_dataframe_asset(
     )
 
 
+@pytest.mark.cloud
```
cdkini, line 479: We don't mark E2E tests any differently right? Might be helpful to add a comment or update the test name to indicate that this makes external requests and relies on state in our OSS CICD acct in prod

cdkini (approved):

great_expectations/core/datasource_dict.py
```diff
@@ -17,11 +17,27 @@
...
+    def assets(self) -> list[DataAsset]:
+        ...
+
+    def add_dataframe_asset(self, **kwargs) -> DataAsset:
+        ...
+
+
+@runtime_checkable
+class InMemoryDataAsset(Protocol):
+    type: Literal["dataframe"] = "dataframe"
```
Kilo59, line 38:
```suggestion
    @property
    def type(self) -> Literal["dataframe"]:
       ...
```

NathanFarmer: I can make this change, but do we need it?

Kilo59:
```python
class InMemoryDataAsset(Protocol):
    type: Literal["dataframe"] = "dataframe"
```

The above is saying that this is a class attribute, but it's an instance attr.

I would suggest changing it but it's not urgent.

---

Separate from this issue, we should double check this conditional only returns true when `type` is `"dataframe"` rather than if it returns any string.

PR #8683, opened 2023-09-07
[MAINTENANCE] Azure Blob Stores `TupleStoreBackend` now supports `credential` string

Shinnnyshinshin:
# What does this PR do? 
* Small update to the `TupleStoreBackend` to support `credential` string. 
* Chose this instead of limiting access to just the `connection_string` since the `BlobServiceClient` which we use to connect to Azure, supports all 3. 
* Updated reference Environment notebook.
* Added unittest


tests/data_context/store/test_store_backends.py
```diff
@@ -1204,6 +1204,49 @@ def test_TupleGCSStoreBackend():  # noqa: PLR0915
     )
 
 
+@pytest.mark.big
```
anthonyburdi: Non blocking - but does this test need to be marked big and not unit? It looks like relevant items are mocked.

anthonyburdi (approved):

anthonyburdi (commented):

tests/data_context/store/test_store_backends.py
Shinnnyshinshin: Great catch. Re-marking before merging :)

PR #8702, opened 2023-09-12
[MAINTENANCE] Add `get_all` method to Store and StoreBackend APIs

cdkini:
Retrieval of an entire collection is not currently supported by our stores - we should allow for a `get_all` mechanism to improve performance (especially around datasources).

### DIFF
--- a/great_expectations/data_context/store/_store_backend.py
+++ b/great_expectations/data_context/store/_store_backend.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import urllib
 import uuid
@@ -121,6 +123,9 @@ class StoreBackend(metaclass=ABCMeta):
         value = self._get(key, **kwargs)
         return value
 
+    def get_all(self):
+        return self._get_all()
+
     def set(self, key, value, **kwargs):
         self._validate_key(key)
         self._validate_value(value)
@@ -214,6 +219,10 @@ class StoreBackend(metaclass=ABCMeta):
 
     @abstractmethod
     def _get(self, key) -> None:
+        raise NotImplementedError
+
+    @abstractmethod
+    def _get_all(self) -> list[Any]:
         raise NotImplementedError
 
     @abstractmethod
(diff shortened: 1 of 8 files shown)


great_expectations/data_context/store/_store_backend.py
```diff
@@ -216,6 +221,10 @@ def _validate_value(self, value) -> None:  # noqa: B027 # no abstract decorator
     def _get(self, key) -> None:
         raise NotImplementedError
 
+    @abstractmethod
+    def _get_all(self) -> list[Any]:
+        raise NotImplementedError
```
Kilo59, line 226: `#future-enhancement`

Can we make the StoreBackend Generic and then we don't have to use `Any` for the return types?

But this should be done in a different PR. I don't want to slow down this PR from getting merged.

https://docs.python.org/3/library/typing.html#user-defined-generic-types

```python
from typing import TypeVar, Generic

T = TypeVar("T")

class StoreBackend(Generic[T], metaclass=ABCMeta):

    def get_all(self) -> T:
        return self._get_all()

    @abstractmethod
    def _get_all(self) -> list[T]:
        raise NotImplementedError

````

File: tests/data_context/store/test_gx_cloud_store_backend.py [hunk 2]
```diff
@@ -271,7 +271,24 @@ def test_list_keys(
...
+            params=None,
+        )
+
+
+def test_get_all(
+    construct_ge_cloud_store_backend: Callable[
+        [GXCloudRESTResource], GXCloudStoreBackend
+    ],
+) -> None:
+    store_backend = construct_ge_cloud_store_backend(GXCloudRESTResource.DATASOURCE)
```
Kilo59, line 284: `non-blocking`

Can you use a `responses` mock here?

https://github.com/getsentry/responses#shortcuts

Kilo59 (approved):
Thanks for this!

great_expectations/data_context/store/_store_backend.py
cdkini: Hmm I agree with this but we're not actually returning hydrated objects (just dict configs) - we should be returning rich config types or actual datasources but our stores use primitives

tests/data_context/store/test_gx_cloud_store_backend.py
cdkini: I'll add this in the follow-up PR - thanks!

PR #8719, opened 2023-09-14
[BUGFIX] Snowflake column name case sensitivity

Kilo59:
Due to a known issue in the `snowflake-sqlalchemy` package, column_names are always converted to lowercase when we pull down column metadata. This meant that our column based expectations would fail if anything other than unquoted lowercase columns were used (regardless of the actual casing of the column in question).

### Related Issues
 - https://github.com/snowflakedb/snowflake-sqlalchemy/issues/157

- https://github.com/great-expectations/great_expectations/pull/8526
- #8630 

This change adds conditional logic (snowflake only) to our `metrics.util.get_sqlalchemy_column_metadata()` function that returns a special dictionary object that returns a `CaseInsenitiveString` whenever the column `name` key is accessed. This eliminates the bug in the most important cases, but will be breaking change in others.

### Note
There is another related issue that is still being worked on that causes expectations to fail because of how we are defining our metric keys.

## TODO:
- [x] test [QUOTED_IDENTIFIERS_IGNORE_CASE](https://docs.snowflake.com/en/sql-reference/parameters#label-quoted-identifiers-ignore-case) with this fix
- [ ] Merge `SnowflakeDatasource` update to accept `QUOTED_IDENTIFIERS_IGNORE_CASE`??
- [ ] Metric key lookup fix - https://github.com/great-expectations/great_expectations/pull/8725

## Column casing behavior

This fix supports the most common uses, but it makes using quoted columns unusable with `great_expectations`, unless combined with the `QUOTED_IDENTIFIERS_IGNORE_CASE` session param.

**Table represents the results of an `expect_column_to_exist` checkpoint run.**
✅ Means the validation succeeded
❌ Means the validation failed
`*` Snowflake lowercase, unquoted column names are not possible, these tables will still default to `FOOBAR`

DDL COLUMN_NAME | Expectation `column_name` arg | expected | current | with fix | with fix + <br>`IQUOTED_IDENTIFIERS_IGNORE_CASE` param
-----------------| ---------------------------------|--------|-------|----|---
`foobar`* | `foobar`| ✅ | ✅  | ✅ | ✅
`foobar`* | `FOOBAR`| ✅| ❌ | ✅ | ✅
`foobar`* | `"foobar"`|❌| ❌ | ❌ | ❌ 
`foobar`* | `"FOOBAR"`|✅| ❌  | ❌ | ❌ 
`FOOBAR` | `FOOBAR`|✅| ❌ | ✅ | ✅
`FOOBAR` | `foobar` | ✅ | ✅  | ✅ | ✅ 
`FOOBAR` | `"foobar"` | ❌  | ❌   | ❌ | ❌ 
`FOOBAR` | `"FOOBAR"` | ✅  | ❌   | ❌ | ❌ 
`"foobar"` | `foobar` | ❌  | ✅  | ❌ | ❌ 
`"foobar"` | `FOOBAR` | ❌  | ❌ | ❌| ❌ 
`"foobar"` | `"foobar"`| ✅ | ❌ | ❌ | ❌ 
`"foobar"` | `"FOOBAR"`| ❌  | ❌ | ❌ | ❌ 
`"FOOBAR"` | `foobar` | ✅ |✅ | ❌ | ✅
`"FOOBAR"` | `FOOBAR`| ✅ | ❌ | ✅|✅ 
`"FOOBAR"` | `"foobar"`| ❌ | ❌ | ❌ |❌ 
`"FOOBAR"` | `"FOOBAR"`|✅| ❌ | ❌ | ❌

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -306,19 +306,20 @@ jobs:
       SNOWFLAKE_ROLE: ${{secrets.SNOWFLAKE_ROLE}}
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         markers:
           - athena or clickhouse or openpyxl or pyarrow or project or sqlite or aws_creds
-          - aws_deps
-          - big
-          - cli
+          # - aws_deps
+          # - big
+          # - cli
           - databricks
-          - filesystem
-          - mssql
-          - mysql
+          # - filesystem
+          # - mssql
+          # - mysql
           - postgresql
           - snowflake
-          - spark
+          # - spark
           - trino
         python-version: ["3.8", "3.9", "3.10", "3.11"]
         exclude:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -306,20 +306,19 @@ jobs:
       SNOWFLAKE_ROLE: ${{secrets.SNOWFLAKE_ROLE}}
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
       matrix:
         markers:
           - athena or clickhouse or openpyxl or pyarrow or project or sqlite or aws_creds
-          # - aws_deps
-          # - big
-          # - cli
+          - aws_deps
+          - big
+          - cli
           - databricks
-          # - filesystem
-          # - mssql
-          # - mysql
+          - filesystem
+          - mssql
+          - mysql
           - postgresql
           - snowflake
-          # - spark
+          - spark
           - trino
         python-version: ["3.8", "3.9", "3.10", "3.11"]
         exclude:
--- a/great_expectations/expectations/metrics/util.py
+++ b/great_expectations/expectations/metrics/util.py
@@ -322,12 +322,19 @@ def attempt_allowing_relative_error(dialect):
 
 
 class _CaseInsensitiveString(str):
+    """
+    A string that compares equal to another string regardless of case,
+    unless it is quoted.
+    """
+
     def __init__(self, string: str):
+        # TODO: check if string is already a _CaseInsensitiveString?
         self._original = string
         self._lower = string.lower()
         self._quote_string = '"'
 
-    def __eq__(self, other):
+    @override
+    def __eq__(self, other: _CaseInsensitiveString | str | object):
         if self.is_quoted():
             return self._original == str(other)
         if isinstance(other, _CaseInsensitiveString):
@@ -340,11 +347,17 @@ class _CaseInsensitiveString(str):
     def __hash__(self):
         return hash(self._lower)
 
+    @override
+    def __str__(self) -> str:
+        return self._original
+
     def is_quoted(self):
         return self._original.startswith(self._quote_string)
 
 
-class SmartColumnLookup(UserDict):
+class CaseInsensitiveNameDict(UserDict):
+    """Normal dict except it returns a case-insensitive string for any `name` key values."""
+
     def __init__(self, data: Dict[str, Any]):
         self.data = data
 
@@ -352,7 +365,7 @@ class SmartColumnLookup(UserDict):
     def __getitem__(self, key: Any) -> Any:
         item = self.data[key]
         if key == "name":
-            logger.debug(f"SmartColumnLookup - {key}:{item}")
+            logger.debug(f"CaseInsensitiveNameDict.__getitem__ - {key}:{item}")
             return _CaseInsensitiveString(item)
         return item
 
@@ -418,7 +431,7 @@ def get_sqlalchemy_column_metadata(
         if dialect_name == "snowflake":
             return [
                 # TODO: SmartColumn should know the dialect and do lookups based on that
-                SmartColumnLookup(column)  # type: ignore[misc]
+                CaseInsensitiveNameDict(column)  # type: ignore[misc]
                 for column in columns
             ]
 
--- a/great_expectations/expectations/metrics/util.py
+++ b/great_expectations/expectations/metrics/util.py
@@ -322,23 +322,23 @@ def attempt_allowing_relative_error(dialect):
     return detected_redshift or detected_psycopg2
 
 
-class _CaseInsensitiveString(str):
+class CaseInsensitiveString(str):
     """
     A string that compares equal to another string regardless of case,
     unless it is quoted.
     """
 
     def __init__(self, string: str):
-        # TODO: check if string is already a _CaseInsensitiveString?
+        # TODO: check if string is already a CaseInsensitiveString?
         self._original = string
         self._lower = string.lower()
         self._quote_string = '"'
 
     @override
-    def __eq__(self, other: _CaseInsensitiveString | str | object):
+    def __eq__(self, other: CaseInsensitiveString | str | object):
         if self.is_quoted():
             return self._original == str(other)
-        if isinstance(other, _CaseInsensitiveString):
+        if isinstance(other, CaseInsensitiveString):
             return self._lower == other._lower
         elif isinstance(other, str):
             return self._lower == other.lower()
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -306,19 +306,20 @@ jobs:
...
-          - filesystem
-          - mssql
-          - mysql
+          # - filesystem
+          # - mssql
+          # - mysql
           - postgresql
           - snowflake
-          - spark
+          # - spark
```
Kilo59, line 322: revert before merge

great_expectations/expectations/metrics/util.py
```diff
@@ -308,6 +321,42 @@ def attempt_allowing_relative_error(dialect):
     return detected_redshift or detected_psycopg2
 
 
+class _CaseInsensitiveString(str):
+    def __init__(self, string: str):
+        self._original = string
+        self._lower = string.lower()
+        self._quote_string = '"'
+
+    def __eq__(self, other):
```
billdirks: Please type `other`. I'm guessing it's a `str` or a `_CaseInsensitiveString`?

great_expectations/expectations/metrics/util.py
```diff
@@ -308,6 +321,42 @@ def attempt_allowing_relative_error(dialect):
...
+        if self.is_quoted():
+            return self._original == str(other)
```
billdirks, line 340: Do you have to implement `__str__` for `str(other)` to return `other._original`? Or does `__hash__` get you this?

great_expectations/expectations/metrics/util.py
```diff
@@ -308,6 +321,42 @@ def attempt_allowing_relative_error(dialect):
...
+        if isinstance(other, _CaseInsensitiveString):
+            return self._lower == other._lower
+        elif isinstance(other, str):
+            return self._lower == other.lower()
+        else:
+            return False
+
+    def __hash__(self):
```
billdirks, line 348: Why do you need this?

great_expectations/expectations/metrics/util.py
```diff
@@ -308,6 +321,42 @@ def attempt_allowing_relative_error(dialect):
...
+        return hash(self._lower)
+
+    def is_quoted(self):
+        return self._original.startswith(self._quote_string)
+
+
+class SmartColumnLookup(UserDict):
```
billdirks: Rename this since it only does a special get for the key `name`? Maybe `CaseInsensitiveNameDict` though that's a bit awful too.

Kilo59: That sounds good to me. If it becomes something else we can rename it.

great_expectations/expectations/metrics/util.py
Kilo59: I was thinking this string might be used as a dictionary key and therefore we wouldn't want it to hash differently than a differently cased instance.

But I don't actually know that that's true. Do you think I should remove it?

great_expectations/expectations/metrics/util.py
Kilo59: Let me double check that. I assumed `str` implements `__str__`.
I want the behavior of calling `str()` on a `_CaseInsensitiveString` to cause it to become a normal string.

billdirks: Yes, I think you need to implement `__str__` for `str(_CaseInsensitiveString)` to return a normal string?

great_expectations/expectations/metrics/util.py
billdirks: Oh, that is probably fine then. Thanks for the explanation!

great_expectations/expectations/metrics/util.py
Kilo59: Done

great_expectations/expectations/metrics/util.py
```diff
@@ -308,13 +322,62 @@ def attempt_allowing_relative_error(dialect):
...
+    def __init__(self, string: str):
+        # TODO: check if string is already a _CaseInsensitiveString?
+        self._original = string
+        self._lower = string.lower()
+        self._quote_string = '"'
+
+    @override
+    def __eq__(self, other: _CaseInsensitiveString | str | object):
+        if self.is_quoted():
+            return self._original == str(other)
```
cdkini: Do we have unit tests for `__eq__`? Might be a nice-to-have

great_expectations/expectations/metrics/util.py
```diff
@@ -359,9 +428,17 @@ def get_sqlalchemy_column_metadata(
                 sqlalchemy_engine=engine,
             )
 
+        dialect_name = execution_engine.dialect.name
+        if dialect_name == "snowflake":
```
cdkini: Do we have an enum here? Prob not since this is metrics code

tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -600,74 +658,172 @@ def _is_quote_char_dialect_mismatch(
...
                 quote=None,
             ),
-            id="quoted_name upper quote=None",
+            id="quoted_name quoted_upper_col quote=None",
         ),
     ],
 )
 class TestColumnIdentifiers:
+    # unskip to help debug `test_column_expectation` failures
+    @pytest.mark.skip(reason="sanity checks for helping debug issues")
```
cdkini, line 726: Hmm should we keep this skip?

great_expectations/expectations/metrics/util.py
Kilo59: There is an enum. I'll use it.

great_expectations/expectations/metrics/util.py
Kilo59: No, I'll add those.

tests/datasource/fluent/integration/test_sql_datasources.py
Kilo59: These tests definitely need some attention. They are meant to mirror the checkpoint test below but without the layer of `gx` abstraction which can cause issues for pulling column details depending on the casing, quoting and dialect.

They can be used for debugging where the issue with the column expectation lies. I don't intend for them to stick around once we have solved our column identifier problems (which we also have with postgres and sqlite).

I could just delete them, but that would make sharing them with other developers working on this problem more difficult.

billdirks (approved):
Thanks @Kilo59! 

Remember to turn back on the tests in `ci.yml` and add those tests @cdkini asked for.

great_expectations/expectations/metrics/util.py
Kilo59: Done

great_expectations/expectations/metrics/util.py
Kilo59: Done

tests/expectations/metrics/test_metrics_util.py
```diff
@@ -478,3 +479,40 @@ def test_get_dbms_compatible_metric_domain_column_list_kwargs(
...
+        other_case_insensitive = CaseInsensitiveString(other)
+        input_case_insensitive = CaseInsensitiveString(input_str)
+
+        # if either string is quoted, they must be exact match
+        if input_case_insensitive.is_quoted() or other_case_insensitive.is_quoted():
+            if input == other:
+                assert input_case_insensitive == other_case_insensitive
+            assert input_case_insensitive != CaseInsensitiveString(other.swapcase())
+        elif input_str.lower() == other.lower():
+            assert input_case_insensitive == CaseInsensitiveString(other.swapcase())
```
Kilo59: TIL about `.swapcase()`

great_expectations/expectations/metrics/util.py
```diff
@@ -308,13 +322,62 @@ def attempt_allowing_relative_error(dialect):
...
+            return CaseInsensitiveString(item)
+        return item
+
+
 def get_sqlalchemy_column_metadata(
     execution_engine: SqlAlchemyExecutionEngine,
     table_selectable: sqlalchemy.Select,
     schema_name: Optional[str] = None,
-) -> Optional[List[Dict[str, Any]]]:
+) -> Sequence[Mapping[str, Any]] | None:
```
Kilo59, line 378: The fact that I could change this to an immutable type means that we must not be directly mutating this return value.
Either that or our type-checking coverage is still too bad 😄

PR #8725, opened 2023-09-15
[BUGFIX] Test metric id creation changes

Kilo59:
WIP

Testing out @billdirks discoveries regarding `get_dbms_compatible_metric_domain_kwargs()` changing metric config ids.


great_expectations/expectations/metrics/map_metric_provider/column_condition_partial.py
```diff
@@ -166,10 +166,10 @@ def inner_func(  # noqa: PLR0913
...
                 runtime_configuration: dict,
             ):
-                metric_domain_kwargs = get_dbms_compatible_metric_domain_kwargs(
-                    metric_domain_kwargs=metric_domain_kwargs,
-                    batch_columns_list=metrics["table.columns"],
-                )
+                # metric_domain_kwargs = get_dbms_compatible_metric_domain_kwargs(
+                #     metric_domain_kwargs=metric_domain_kwargs,
+                #     batch_columns_list=metrics["table.columns"],
+                # )
```
Kilo59, line 172: @billdirks If this doesn't cause any of our existing tests to fail can we just merge this?

I will make sure to post to `#engineering` and solicit feedback more widely before taking that step.

billdirks: Yes, if no other tests break. I think if this actually changes `metric_domain_kwargs` then, when running checkpoints, gx would always fail at metric lookup time when evaluating the expectation. Not sure if this is used in any other code path. Changing ids seems like very unexpected behavior.

billdirks: Please delete this instead of commenting out before you merge.

.github/workflows/ci.yml
```diff
@@ -306,19 +306,20 @@ jobs:
...
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         markers:
           - athena or clickhouse or openpyxl or pyarrow or project or sqlite or aws_creds
-          - aws_deps
-          - big
-          - cli
+          # - aws_deps
```
billdirks, line 313: Uncomment these before merging.

billdirks (approved):
Thanks! Left a few comments about deleting/uncommenting but otherwise LGTM.

Issue #7453, opened 2023-03-24
[BUG] expect_multicolumn_sum_to_equal outputs AttributeError / TypeError


data-han:
**Describe the bug**
When using expect_multicolumn_sum_to_equal, it does not outputs the result as intended. Instead, it outputs an error which fails to show up in Data doc anyway.

**To Reproduce**
using expect_multicolumn_sum_to_equal 

**Expected behavior**
This should work as intended as per document, where multi columns summation should result in a value

**Environment (please complete the following information):**
 - Operating System: [e.g. Linux, MacOS, Windows]: MacOS
 - Great Expectations Version: [e.g. 0.13.2]: 0.16.1

**Additional context**

<img width="1042" alt="image" src="https://user-images.githubusercontent.com/51312880/227598423-dc69bebb-2875-4e5e-8ec8-0eecfbacb250.png">

talagluck:
Thanks for raising this @data-han! We will review and be in touch.

data-han:
any update on this?  @talagluck @austiezr

talagluck:
Hi @data-han - thanks for the ping, and apologies for the delay on this. We have reproduced this, and it is in our backlog for the engineering team to fix. We don't yet have a firm date for this, but will let you know once we have a better idea.

suhren:
I am also experiencing this on 0.16.13. Is there some suggested workaround until there is a fix? Implementing a custom expectation with the same functionality?

Shinnnyshinshin:
@data-han and @suhren thank you for bringing up this issue. The problem was caused by the fact that some Expectations were not properly using inheritance to use the renderers in the base Expectation class. The issue itself has been resolved by #8745 and #8739 will add better renderers to `ExpectMulticolumnSumToEqual`

PR #8745, opened 2023-09-20
[BUGFIX] Ensure renderers use parent class renderers instead of using non-implemented override methods

Shinnnyshinshin:
# What does this PR do? 
* Removes non-implemented `_diagnostic_observed_value_renderer` and `_prescriptive_renderer` methods from Expectations so that the base class's methods can be run. With the current implementation, the `pass` statement will return `None` and will cause a [`AttributeError`](https://github.com/great-expectations/great_expectations/issues/7453).
* Ensure that [the base class's `_prescriptive_renderer`](https://github.com/great-expectations/great_expectations/blob/800df5bc904b334d2844f9131e4098119ffcb0c3/great_expectations/expectations/expectation.py#L536C7-L536C7) and [`_diagnostic_observed_value_renderer`](https://github.com/great-expectations/great_expectations/blob/800df5bc904b334d2844f9131e4098119ffcb0c3/great_expectations/expectations/expectation.py#L1013) can be run. 
# Can you give me an example? 
```python 
class Person:
    def __init__(self, fname, lname):
        self.firstname = fname
        self.lastname = lname
    def printname(self):
        print(self.firstname, self.lastname)

class Student(Person):
    def __init__(self, fname, lname):
        super().__init__(fname, lname)
    def printname(self):
        pass
```
* This is the current inheritance pattern we have now, and we expect the following output for `Student`, the child class

```python 
x = Student("Luke", "Skywalker")
x.printname()  # will print nothing
```

*The current PR is proposing the following change for `Student`*

```python
class Student(Person):
    def __init__(self, fname, lname):
        super().__init__(fname, lname)
```

* And now the output will be:

```python 
x = Student("Luke", "Skywalker")
x.printname() 
Luke Skywalker
```

### DIFF
--- a/great_expectations/expectations/core/expect_column_bootstrapped_ks_test_p_value_to_be_greater_than.py
+++ b/great_expectations/expectations/core/expect_column_bootstrapped_ks_test_p_value_to_be_greater_than.py
@@ -1,16 +1,6 @@
-from typing import Optional
-
-from great_expectations.compatibility.typing_extensions import override
-from great_expectations.core import (
-    ExpectationConfiguration,
-    ExpectationValidationResult,
-)
 from great_expectations.expectations.expectation import (
     BatchExpectation,
-    render_evaluation_parameter_string,
 )
-from great_expectations.render import LegacyDiagnosticRendererType, LegacyRendererType
-from great_expectations.render.renderer.renderer import renderer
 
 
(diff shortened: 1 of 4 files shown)


anthonyburdi (approved):
Thank you for the example! LGTM!

PR #8757, opened 2023-09-22
[MAINTENANCE] SQL column identifier testing - ensure expectations fail when expected

Kilo59:
1. Update tests to fail when expected, according to the Dialect DQL.
2. Apply xfail markers properly to allow for possibility of an `xpass`.

- Followup from https://github.com/great-expectations/great_expectations/pull/8719


tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -227,14 +225,14 @@ def __call__(
...
 
-@pytest.fixture(scope="function")
-def table_factory(
-    capture_engine_logs: pytest.LogCaptureFixture,
-    silence_sqla_warnings: None,  # TODO: remove this
-) -> Generator[TableFactory, None, None]:
+@pytest.fixture(
+    scope="class",
+)
+def table_factory() -> Generator[TableFactory, None, None]:
```
Kilo59, line 231: I made this class scoped so we aren't re-creating the tables and columns every test.

tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -604,40 +603,179 @@ def test_checkpoint_run(
...
+    'str "unquoted_upper_col"': ["snowflake"],
+    'str "UNQUOTED_UPPER_COL"': ["postgres"],
+    # -------------------------------------
+    "str quoted_lower_col": ["postgres", "snowflake"],
+    "str QUOTED_LOWER_COL": ["postgres", "snowflake"],
+    'str "QUOTED_LOWER_COL"': ["postgres", "snowflake"],
+    # -------------------------------------
+    "str quoted_upper_col": ["postgres"],
+    'str "quoted_upper_col"': ["postgres", "snowflake"],
+    "str QUOTED_UPPER_COL": ["postgres"],
```
Kilo59, line 751: Could someone check my math on these expected failures?

tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -604,40 +603,180 @@ def test_checkpoint_run(
...
+    'str "unquoted_upper_col"': ["snowflake"],
+    'str "UNQUOTED_UPPER_COL"': ["postgres"],
+    # DDL: "quoted_lower_col"-----------------------------------
+    "str quoted_lower_col": ["postgres", "snowflake"],
+    "str QUOTED_LOWER_COL": ["postgres", "snowflake"],
+    'str "QUOTED_LOWER_COL"': ["postgres", "snowflake"],
+    # DDL: "QUOTED_UPPER_COL" ----------------------------------
+    "str quoted_upper_col": ["postgres"],
+    'str "quoted_upper_col"': ["postgres", "snowflake"],
+    "str QUOTED_UPPER_COL": ["postgres"],
```
Kilo59: Could someone check my math on these expected failures?

tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -653,77 +793,29 @@ def _is_quote_char_dialect_mismatch(
...
-                quote=None,
-            ),
-            id="quoted_name quoted_upper_col quote=None",
-        ),
     ],
 )
 class TestColumnIdentifiers:
-    # unskip to help debug `test_column_expectation` failures
-    @pytest.mark.skip(reason="sanity checks for helping debug issues")
     def test_raw_queries(
```
wookasz: I think we should get rid of this "test." It is not testing our functionality but rather the databases themselves. It might be nice to have a utility we can run against a database to give us these results if we're adding a new data source, but otherwise it can be confusing to the user about what the purpose of the test is. Also adds time and cost to these test runs, without a large amount of benefit as the underlying databases are unlikely to change their behavior.

tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -653,77 +793,29 @@ def _is_quote_char_dialect_mismatch(
 @pytest.mark.parametrize(
     "column_name",
     [
+        # DDL: unquoted_lower_col ----------------------------------
```
wookasz, line 780: I'm having a lot of trouble reconciling the params with the overrides (e.g., expected to fail). It would be much more clear if we did a little bit of copy-paste here and params looked something like

`"datasource,column_name,expect_success", [param("snowflake", '"unquoted_lower_col"', False, id=...), ...]`

This lets me quickly figure out what should pass and what should fail

tests/datasource/fluent/integration/test_sql_datasources.py
wookasz: `str quoted_lower_col` should not fail for `postgres`
`str QUOTED_LOWER_COL` should not fail for `postgres`

tests/datasource/fluent/integration/test_sql_datasources.py
Kilo59: As discussed in our last sync.
The raw queries test, and the `EXPECTED_FAILURE` mapping have been removed.

We now determine whether or not the expectation should fail by doing a `SELECT <COLUMN_PARAM> FROM <TABLE> LIMIT 1;` before the checkpoint run.

If the raw query returns a result, then the column "exists" according to the specific SQL dialect and the subsequent validation result should succeed, if not it should fail.
This eliminates the maintenance burden and human error that could be introduced by maintaining these mappings and overrides.

tests/datasource/fluent/integration/test_sql_datasources.py
Kilo59: https://github.com/great-expectations/great_expectations/pull/8757#discussion_r1341393595

tests/datasource/fluent/integration/test_sql_datasources.py
Kilo59: https://github.com/great-expectations/great_expectations/pull/8757#discussion_r1341393595

tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -650,180 +748,63 @@ def _is_quote_char_dialect_mismatch(
...
+        # an exception will be raised if the column does not exist
+        try:
+            col_exist_check = connection.execute(
+                TextClause(query),
+            )
+            print(f"\nResults:\n  {col_exist_check.keys()}")
+            col_exist_result = col_exist_check.fetchone()
+            print(f"  Values: {col_exist_result}\n{column_name_param} exists!\n")
+            if not col_exist_result:
+                return False
```
Kilo59, line 769:
```suggestion
```

billdirks (approved):

PR #8773, opened 2023-09-29
[BUGFIX] Improve datasource crud

roblim:
- Update logic in `_SourceFactories.create_update_crud_method` and `_SourceFactories.create_add_or_update_crud_method` to call `AbstractDataContext._update_fluent_datasource`, instead of calling `AbstractDataContext._delete_fluent_datasource` + `AbstractDataContext._add_fluent_datasource`. The former approach was causing an issue with Cloud, since Datasources with associated Checkpoints cannot be deleted, preventing users from updating Datasources
- Revert `CloudDataContext._save_project_config` to being no-op. With the recent removal of Datasource caching and the introduction of `DatasourceDict`, this logic is no longer necessary and was causing unnecessary HTTP calls to cloud.


great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -456,7 +456,7 @@ def _add_splitter(self: Self, splitter: Splitter) -> Self:
         # persist the config changes
         context: AbstractDataContext | None
         if context := self._datasource._data_context:
-            context._save_project_config(self._datasource)
+            context.datasources[self._datasource.id] = self._datasource
```
Kilo59: Why are we storing this by the id and not the name?

roblim: Good catch. Addressed in https://github.com/great-expectations/great_expectations/pull/8773/commits/3f228fa3ed30609530ae4d3162336f05d4ad829a

great_expectations/core/datasource_dict.py
```diff
@@ -92,17 +94,27 @@ def data(self) -> dict[str, FluentDatasource | BaseDatasource]:  # type: ignore[
 
         return datasources
 
-    @override
-    def __setitem__(self, name: str, ds: FluentDatasource | BaseDatasource) -> None:
+    def set_datasource(
+        self, name: str, ds: FluentDatasource | BaseDatasource
+    ) -> FluentDatasource | BaseDatasource | None:
```
cdkini: Is this an accurate return? Looks like it should be `Fluent | None`.

Instead of `None`, could we just return `Fluent | Base`? I'd love to avoid nullable returns if possible

great_expectations/core/datasource_dict.py
```diff
@@ -210,12 +226,15 @@ def __contains__(self, name: object) -> bool:
...
         self.data[name] = ds
 
         # FDS do not use stores
         if not isinstance(ds, FluentDatasource):
-            super().__setitem__(name, ds)
+            return super().set_datasource(name, ds)
```
cdkini: Nit - Could we use key-word args

great_expectations/core/datasource_dict.py
```diff
@@ -210,12 +226,15 @@ def __contains__(self, name: object) -> bool:
...
+        return ds
```
cdkini, line 255: Hmm okay the signature makes more sense here since BDS is a possibility

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -379,9 +379,7 @@ def _init_project_config(
     def _init_variables(self) -> DataContextVariables:
         raise NotImplementedError
 
-    def _save_project_config(
-        self, _fds_datasource: FluentDatasource | None = None
-    ) -> FluentDatasource | None:
+    def _save_project_config(self) -> None:
```
cdkini, line 382: yesssss thank you for cleaning this up

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -943,26 +943,13 @@ def _persist_suite_with_store(
...
-        # the different requirements for FDS vs BDS.
-        # At which time `_save_project_config` will revert to being a no-op operation on the CloudDataContext.
-        if _fds_datasource:
-            return self._datasource_store.set(key=None, value=_fds_datasource)
-        else:
-            logger.debug(
-                "CloudDataContext._save_project_config() has no `fds_datasource` to update"
-            )
-            return None
+        return None
```
cdkini: Could we log here instead of completely no-oping?

tests/datasource/fluent/test_contexts.py
```diff
@@ -69,7 +67,7 @@ def test_add_fluent_datasource_are_persisted(
     assert set_spy.call_count == 1
     cloud_api_fake.assert_call_count(
         f"{GX_CLOUD_MOCK_BASE_URL}/organizations/{FAKE_ORG_ID}/datasources",
-        1,
+        2,
```
cdkini, line 70: Why are we increasing the call count here? I thought this PR was reducing duplicate calls?

tests/conftest.py
```diff
@@ -3711,20 +3745,8 @@ def cloud_data_context_with_datasource_pandas_engine(
...
-        return value
-
-    with mock.patch(
-        "great_expectations.data_context.store.gx_cloud_store_backend.GXCloudStoreBackend.list_keys"
-    ), mock.patch(
-        "great_expectations.data_context.store.datasource_store.DatasourceStore.set",
-        side_effect=set_side_effect,
-    ):
-        fds = PandasDatasource(name="my_datasource")
-        context.add_datasource(datasource=fds)
```
cdkini, line 3727: All these `mock.patch` deletions feel great - thank you!

tests/datasource/fluent/test_contexts.py
roblim: I think it's because the fixture was overriding context.datasources with a dict, but since removing that override, accessing context.datasources results in a REST call to list datasources. So I think it's more the test just wasn't accurate before.

great_expectations/core/datasource_dict.py
roblim: Sure. Addressed in https://github.com/great-expectations/great_expectations/pull/8773/commits/864c95207ad75eca04638f3d16fbf80fe1ec2695

great_expectations/core/datasource_dict.py
roblim: Good catch. Addressed in https://github.com/great-expectations/great_expectations/pull/8773/commits/db0d0ad9848fe204700b50c5cf584eda87185b1a

great_expectations/data_context/data_context/cloud_data_context.py
roblim: Sure. Addressed in https://github.com/great-expectations/great_expectations/pull/8773/commits/460c41b8e2a7becf6e97d41b97d1ba61875d0225.

great_expectations/core/datasource_dict.py
```diff
@@ -92,17 +94,27 @@ def data(self) -> dict[str, FluentDatasource | BaseDatasource]:  # type: ignore[
 
         return datasources
 
-    @override
-    def __setitem__(self, name: str, ds: FluentDatasource | BaseDatasource) -> None:
+    def set_datasource(
+        self, name: str, ds: FluentDatasource | BaseDatasource
+    ) -> FluentDatasource | None:
```
Kilo59, line 109: `non-blocking`
overloads for the fluent vs base datasource could help with the typing of the return type.

```python
   @overload
    def set_datasource(
        self, name: str, ds: FluentDatasource
    ) -> FluentDatasource:
        ...

    @overload
    def set_datasource(
        self, name: str, ds: BaseDatasource
    ) -> None:
       ...

    def set_datasource(
        self, name: str, ds: FluentDatasource | BaseDatasource
    ) -> FluentDatasource | None:
```

great_expectations/datasource/fluent/sources.py
```diff
@@ -512,21 +517,14 @@ def update_datasource(
...
 
-            # local delete only, don't update the persisted store entry
-            self._data_context._delete_fluent_datasource(
-                datasource_name=datasource_name, _call_store=False
-            )
-            # Now that the input is validated and the old datasource is deleted we pass the
-            # original arguments to the add method (ie name and not datasource_name).
-            return self.create_add_crud_method(datasource_type)(
-                name_or_datasource, **kwargs
+            updated_datasource.test_connection()
```
Kilo59, line 522: Why are we calling `test_connection()` here?

tests/cli/test_sanitize_yaml_and_save_datasource.py
```diff
@@ -42,24 +42,23 @@ def test_sanitize_yaml_and_save_datasource_works_without_credentials(
...
 name: my_datasource
-class_name: SimpleSqlalchemyDatasource
-introspection:
-  whole_table:
-    data_asset_name_suffix: __whole_table
-connection_string: sqlite://"""
+class_name: Datasource
+execution_engine:
+  class_name: PandasExecutionEngine
+"""
```
Kilo59: `non-blocking`

Why did these BDS examples need to change?

tests/conftest.py
```diff
@@ -3667,9 +3673,37 @@ def empty_cloud_data_context(
         cloud_organization_id=ge_cloud_config.organization_id,
     )
 
-    context._datasources = (
-        {}
-    )  # Basic in-memory mock for DatasourceDict to avoid HTTP calls
```
Kilo59, line 3672: 👏

tests/experimental/metric_repository/test_cloud_data_store.py
```diff
@@ -17,8 +17,8 @@
     MetricRun,
 )
 
-# Import and use fixtures defined in tests/datasource/fluent/conftest.py
-from tests.datasource.fluent.conftest import (
+# Import and use fixtures defined in tests/conftest.py
+from tests.conftest import (
```
Kilo59: `non-blocking`
@roblim 
Now that you've moved some of these fixtures to the top-level `conftest`, they no longer need to be imported.
They should be available via the normal pytest fixture magic.

tests/experimental/metric_repository/test_column_descriptive_metrics_metric_retriever_integration.py
```diff
@@ -16,8 +16,8 @@
     TableMetric,
 )
 
-# Import and use fixtures defined in tests/datasource/fluent/conftest.py
-from tests.datasource.fluent.conftest import (
+# Import and use fixtures defined in tests/conftest.py
+from tests.conftest import (
```
Kilo59: Same as above/below

great_expectations/datasource/fluent/sources.py
roblim, line 522: @Kilo59 @cdkini - during our pairing session, @Kilo59 mentioned that `test_connection` should only be called during `add_datasource` operation, not `update_datasource`. However, when I went back to double check, it looks like adding `.test_connection()` (in `update_datasource` and `add_or_update_datasource`) is actually required to maintain existing behavior, since existing behavior is to delete then re-add Datasources when doing updates. Since re-adding just uses existing `add_datasource` method, `test_connection()` ends up getting called, just not in the main body of the update methods.

Should we maintain this behavior or is the existing impl a bug? LMK if I'm misreading the logic somewhere.

Kilo59 (approved):
Excellent work!
I know this was a pain!

great_expectations/datasource/fluent/sources.py
Kilo59: That makes, sense.
I wouldn't have expected that but given that's our current behavior, I say let's keep it.

roblim: See above comment: https://github.com/great-expectations/great_expectations/pull/8773#discussion_r1353428484

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -789,28 +785,20 @@ def _add_fluent_datasource(
...
+        return_obj._data_context = self
+        self._save_project_config()
 
-        return datasource
+        return return_obj
 
     def _update_fluent_datasource(
         self, datasource: Optional[FluentDatasource] = None, **kwargs
-    ) -> None:
+    ) -> FluentDatasource | BaseDatasource:
```
cdkini: Is this true? Should just be `FluentDatasource` right?

great_expectations/core/datasource_dict.py
Kilo59: Actually, a `TypeVar` would be better. It doesn't just return any fluent datasource. It's returning the same type it was passed.

```python
T = TypeVar('T', bound=FluentDatasource)

   @overload
    def set_datasource(
        self, name: str, ds: T
    ) -> T:
        ...

    @overload
    def set_datasource(
        self, name: str, ds: BaseDatasource
    ) -> None:
       ...

    def set_datasource(
        self, name: str, ds: T | BaseDatasource
    ) -> T | None:


```

great_expectations/data_context/data_context/abstract_data_context.py
roblim: Good catch! Addressed in https://github.com/great-expectations/great_expectations/pull/8773/commits/b760677108e3d57b9d8fdc4319398bf8d4f265c4

tests/cli/test_sanitize_yaml_and_save_datasource.py
roblim: Good catch - meant to revert. Initial approach to `DatasourceDict.set_datasource` attempted to instantiate and return `BaseDatasource` obj from config for BDS, but this broke a bunch of tests (these changes were first few attempts to fix). Since behavior of returning a Datasource from this method is new and only relevant to FDS, decided to just return None for BDS vs. instantiating, instead of wasting time refactoring tests.

Addressed in https://github.com/great-expectations/great_expectations/pull/8773/commits/e51aa04faaf82e6c32e8635e0f9a42f36c54e2a2.

tests/experimental/metric_repository/test_cloud_data_store.py
roblim: Addressed in https://github.com/great-expectations/great_expectations/pull/8773/commits/3e17393f0dc563d9cb3a01b571d5b3f97563eb64

tests/experimental/metric_repository/test_column_descriptive_metrics_metric_retriever_integration.py
roblim: Addressed in https://github.com/great-expectations/great_expectations/pull/8773/commits/3e17393f0dc563d9cb3a01b571d5b3f97563eb64

great_expectations/core/datasource_dict.py
roblim: Thanks for the detailed suggestion! Addressed in https://github.com/great-expectations/great_expectations/pull/8773/commits/7b417a7499e3e6a00075a633205c44b902168cee.

PR #8783, opened 2023-10-02
[MAINTENANCE] Update column identifier tests

Kilo59:
1. Update `TestColumnExpectations` to test for `"mixedCase"` and `"with.dots"` column identifiers.
2. Add a `test_desired_state` which makes a raw query against the dialect in question and uses that result to determine if our expectation should pass or fail.


tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -899,8 +951,7 @@ def test_column_expectation(
         )
         result = checkpoint.run()
 
-        exc_details = _get_exception_details(result, prettyprint=True)
-        assert not exc_details, exc_details[0]["raised_exception"]
+        _ = _get_exception_details(result, prettyprint=True)
```
Kilo59, line 954: Stop caring about the implementation detail of if an expectation failed with exceptions or not.

roblim (approved):
LGTM! Thanks for putting this together!

PR #8803, opened 2023-10-05
[MAINTENANCE] Add cloud api via docker compose for testing

nicgrayson:
This PR:
- adds a new `docker-compose.yml` file for running the mercury API locally
- runs all existing tests marked with `cloud` against a the local API
- adds new end-to-end tests for `CloudDataContext`


tasks.py
```diff
@@ -1115,6 +1133,8 @@ def service(
                 "up",
                 "-d",
                 "--quiet-pull",
+                "--wait",
+                "--wait-timeout 90",
```
NathanFarmer, line 1131: Waits for successful heathchecks or finished containers before proceeding to the next step. Times out and raises an error at 90 seconds for unhealthy containers.

tyler-hoffman (approved):
This looks reasonable to me. Should this be a new step, though, or a new job?

NathanFarmer:
> This looks reasonable to me. Should this be a new step, though, or a new job?

In CI it is grouped under the existing `ci / cloud-tests`.

pyproject.toml
```diff
@@ -540,6 +540,7 @@ markers = [
     "e2e: mark test as an E2E test.",
     "external_sqldialect: mark test as requiring install of an external sql dialect.",
     "filesystem: mark tests using the filesystem as the storage backend.",
+    "mercury: mark tests as using the mercury container locally.",
```
Kilo59: Do we need this new test marker?
Can we just run this as part of the `cloud` tests?

Kilo59: If we re-use the same `cloud` marker then that makes it trivial for me to update existing cloud tests to also run against this containerized version of mercury (instead of against prod or against a mock) without having to write any new tests.

NathanFarmer: I was worried that we would need to re-create a bunch of saved entities in the existing data context, but it was limited to 2 tests. Not sure why I didn't just try it first! All Cloud tests run against a local version of the API now.

PR #8822, opened 2023-10-11
[BUGFIX] Fix missing `config_provider`

Kilo59:
- Followup to #8773 

The `_data_context` was not being attached to the `datasource` before calling `test_connection()` as part of the `context.sources.add_or_update_<DS_TYPE>()` flow.

This causes the `test_connection()` to fail if the user's datasource needs to do a variable config substitution. 
The `config_provider` is pulled from the `data_context`.

### DIFF
--- a/great_expectations/datasource/fluent/sources.py
+++ b/great_expectations/datasource/fluent/sources.py
@@ -519,6 +519,7 @@ class _SourceFactories:
             if id_:
                 updated_datasource.id = id_
 
+            updated_datasource._data_context = self._data_context
             updated_datasource.test_connection()
             return_obj = self._data_context._update_fluent_datasource(
                 datasource=updated_datasource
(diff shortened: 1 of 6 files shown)


great_expectations/datasource/fluent/sources.py
```diff
@@ -519,6 +519,7 @@ def update_datasource(
             if id_:
                 updated_datasource.id = id_
 
+            updated_datasource._data_context = self._data_context
             updated_datasource.test_connection()
```
cdkini, line 523: Let's put a comment here detailing why we need to attach the context right before the test connection call? Feels like such a small nuance that can cause things to break (like it just did).

Or perhaps we encapsulate these two lines in a function to provide more context

cdkini (approved):

great_expectations/datasource/fluent/sources.py
Kilo59: I plan to followup this up and move these to a `attach_context_and_test_connection()` function.

PR #8831, opened 2023-10-13
[MAINTENANCE] Add pact contract testing with GX Cloud

NathanFarmer:
- Add a single pact contract test
- Publish pact to broker
- Verify pact

image


.gitignore
```diff
@@ -148,3 +148,6 @@ metastore_db
 # dependency management
 Pipfile
 Pipfile.lock
+
+# contract testing files
+tests/integration/cloud/rest_contracts/pacts/great_expectations-mercury.json
```
cdkini: should we ignore all JSON files in this directory? Or perhaps the entire directory?

reqs/requirements-dev-lite.txt
```diff
@@ -5,6 +5,7 @@ freezegun>=0.3.15
 ipykernel<=6.17.1 # Newest version (6.19.0 released on 12/7/2022) causes "WARNING  traitlets:client.py:1181 No handler found for comm target 'comm'" by "https://github.com/jupyter/nbclient/blob/main/nbclient/client.py" (version 0.7.2) to be emitted.
 moto>=2.0.0,<3.0.0
 nbconvert>=5
+pact-python>=2.0.1
```
cdkini, line 8: This is a testing dependency right? Should it go in requirements-dev-test or whatever equivalent file we have?

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -0,0 +1,148 @@
+from __future__ import annotations
```
cdkini, line 1: Can we put this file under type checking in our `pyproject.toml`?

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -0,0 +1,148 @@
...
+    pathlib.Path(pathlib.Path(__file__).parent, "pacts").resolve()
+)
+
+
+class RequestMethods(str, enum.Enum):
+    DELETE = "DELETE"
+    GET = "GET"
+    PATCH = "PATCH"
+    POST = "POST"
+    PUT = "PUT"
```
cdkini, line 41: I want to say we get this for free through another library (maybe stdlib) but that could be a later version of Python

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+@pytest.fixture
+def session() -> Session:
+    return create_session(access_token=os.environ.get("GX_CLOUD_ACCESS_TOKEN"))
```
cdkini: Could we add some error handling to panic if we don't have this env var set? Maybe an informative message for devs down the line?

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -0,0 +1,148 @@
...
+
+def get_git_commit_hash() -> str:
+    return subprocess.check_output(["git", "rev-parse", "HEAD"]).decode("ascii").strip()
+
+
+@pytest.fixture
+def pact(request) -> Pact:
+    pact_broker_base_url = "https://greatexpectations.pactflow.io"
+    consumer_name = "great_expectations"
+    provider_name = "mercury"
```
cdkini, line 61: Should these also be constants at the top of the file?

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -0,0 +1,148 @@
...
+            "status": contract_interaction.response_status,
+            "body": contract_interaction.response_body,
+        }
+
+        (
+            pact.given(provider_state=contract_interaction.given)
+            .upon_receiving(scenario=contract_interaction.upon_receiving)
+            .with_request(**request)
+            .will_respond_with(**response)
+        )
```
cdkini, line 175: Do we need these parens?

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -0,0 +1,148 @@
...
+
+        request_url = f"http://{PACT_MOCK_HOST}:{PACT_MOCK_PORT}{path}"
```
cdkini, line 177: for my own underestanding, is there a difference between `http` and `https` here?

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -0,0 +1,148 @@
...
+
+        with pact:
+            session.get(str(request_url))
```
cdkini: Nit - we don't need the str conversion

tests/integration/cloud/rest_contracts/test_data_context_configuration.py
```diff
@@ -0,0 +1,51 @@
...
+import os
+import pathlib
+from typing import Callable, Final
+
+import pytest
+from pact import Like
+
+from tests.integration.cloud.rest_contracts.conftest import ContractInteraction
+
+ORGANIZATION_ID: Final[str] = os.environ.get("GX_CLOUD_ORGANIZATION_ID")
```
cdkini: Same note with error handling here - how do we inform devs if they're missing this env var?

cdkini: Additionally, I thought we had a custom `ruff` rule that disallows `os.environ` usage outside of particular files - @tyler-hoffman might know more?

tests/integration/cloud/rest_contracts/test_data_context_configuration.py
```diff
@@ -0,0 +1,51 @@
...
+                    "expectation_suite": True,
+                },
+            },
+        ),
+    ],
+)
+def test_data_context_configuration(
+    contract_interaction: ContractInteraction,
+    run_pact_test: Callable,
+):
```
cdkini: Since these tests are quite important for making sure we adhere to our contracts, could we be explicit and state what exactly we're testing (maybe in a docstring)?

.gitignore
NathanFarmer: Yeah I meant to do the entire directory and forgot ✅

reqs/requirements-dev-lite.txt
NathanFarmer: `requirements-dev-test.txt` uses this file. Other testing deps are in here like `pytest` and `snapshottest`.

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: Is it not by default?

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: Yeah it seems stupid, but pydantic works nicely with enums, and we can use it in the tests if desired.

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: I had them there initially, but they are only used in this one place. The other constants are used elsewhere. Maybe it's nice to just have them at the top? Not sure that it matters.

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: No but it looks nicer than this IMO, which is what `black` does without them:
```
pact.given(provider_state=contract_interaction.given).upon_receiving(
    scenario=contract_interaction.upon_receiving
).with_request(**request).will_respond_with(**response)
```

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: `https` was requiring me to have an SSL certificate, which added more work.

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: 🤦🏻

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: One thing to note is there is no canonical list of these methods. It can be any string.

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: ✅

tests/integration/cloud/rest_contracts/test_data_context_configuration.py
NathanFarmer: ✅ I added try/except logic to `conftest.py` once I figured out we don't need the real id in the parametrized case.

NathanFarmer: I did that `os.environ` change here: https://github.com/great-expectations/great_expectations/pull/8588

It doesn't seem to be applied to `tests` directory. @tyler-hoffman @Kilo59 we are only concerned with accessing `os.environ` in prod code right now, or should we be trying to banish it from `tests` as well?

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: @Kilo59

tests/integration/cloud/rest_contracts/conftest.py
cdkini: okay no complaints from my end was just curious!

tests/integration/cloud/rest_contracts/test_data_context_configuration.py
NathanFarmer: I added a docstring to `ContractInteraction` and `run_pact_test`, so users can refer to those.

tests/integration/cloud/rest_contracts/test_data_context_configuration.py
Kilo59: I don't think the use of `os.environ` in tests is necessarily a problem.

tests/integration/cloud/rest_contracts/conftest.py
Kilo59: No it isn't. The tests we currently type-check are FDS tests and tests for the `/experimental branch`
I agree it should be added.

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -0,0 +1,179 @@
...
+
+class RequestMethods(str, enum.Enum):
+    DELETE = "DELETE"
+    GET = "GET"
+    PATCH = "PATCH"
+    POST = "POST"
+    PUT = "PUT"
+
+
+@pytest.fixture
```
Kilo59:
```suggestion
@pytest.fixture(scope="module")
```

We shouldn't create a new HTTP session for every test,

tests/integration/cloud/rest_contracts/conftest.py
Kilo59: If they aren't meant to be used outside this test, then I think keeping them in here is cleaner.
If they need to be reused, then we can move them.

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -0,0 +1,179 @@
...
+        response_body: A dictionary or Pact Matcher object representing the response body.
+        request_body (Optional): A dictionary or Pact Matcher object representing the request body.
+        request_headers (Optional): A dictionary representing the request headers.
+
+    Returns:
+        ContractInteraction
+    """
+
+    class Config:
+        arbitrary_types_allowed = True
```
Kilo59, line 115: Is this because of the `Matcher` type?

Kilo59 (approved):
I don't know `Pact` but seems good to me.

I'd like to see us re-use the HTTP session across tests. Can do that by changing the scope of the `session` fixture.

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: Yes

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -0,0 +1,179 @@
...
+class RequestMethods(str, enum.Enum):
+    DELETE = "DELETE"
+    GET = "GET"
+    PATCH = "PATCH"
+    POST = "POST"
+    PUT = "PUT"
+
+
+@pytest.fixture
+def session() -> Session:
```
Kilo59:
```suggestion
def gx_cloud_session() -> Session:
```
Session can be ambiguous.
A DB session? a PACT session?

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: Makes sense to me ✅

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: I like `gx_cloud_session`

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: ✅

PR #8855, opened 2023-10-22
[MAINTENANCE] Use contract test response bodies as unit test mocks

NathanFarmer:
Use Pact test response bodies as unit test mocks


tests/integration/cloud/rest_contracts/unit_test_mocks/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+from tests.integration.cloud.rest_contracts.test_data_context_configuration import (
+    DATA_CONTEXT_CONFIGURATION_MIN_RESPONSE_BODY,
+)
+
+
+def _convert_matcher_to_value(matcher: Matcher) -> JsonData:
+    return matcher.generate()["contents"]
+
+
+def _reify_pact_body(
```
tyler-hoffman, line 21: Is the plan to use this more widely? If not, this seems like a lot of code given how minimal `DATA_CONTEXT_CONFIGURATION_MIN_RESPONSE_BODY`. Also is `reify` the right word?

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -126,8 +131,8 @@ class Config:
     upon_receiving: pydantic.StrictStr
     given: pydantic.StrictStr
     response_status: Annotated[int, pydantic.Field(strict=True, ge=100, lt=600)]
-    response_body: Union[dict, Matcher]
-    request_body: Union[dict, Matcher, None] = None
+    response_body: PactBody
+    request_body: Union[PactBody, None] = None
```
Kilo59, line 136:
```suggestion
    request_body: PactBody | None = None
```

tests/integration/cloud/rest_contracts/unit_test_mocks/test_mock_cloud_data_context.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+from great_expectations.data_context import CloudDataContext
+
+
+@pytest.mark.cloud
+def test_mock_cloud_data_context(mock_cloud_data_context: CloudDataContext):
```
tyler-hoffman, line 7: Why write a test on a fixture? Or is there more that I'm missing?

tests/integration/cloud/rest_contracts/test_data_context_configuration.py
```diff
@@ -1,16 +1,30 @@
...
 
 from tests.integration.cloud.rest_contracts.conftest import (
     ContractInteraction,
 )
 
+DATA_CONTEXT_CONFIGURATION_MIN_RESPONSE_BODY: Final[dict] = {
+    "anonymous_usage_statistics": Like(
+        {
+            "data_context_id": Format().uuid,
+            "enabled": False,
```
Kilo59, line 17: `non-blocking`

Could we use the qualified `pact.Like` and `pact.Format` names?
That way it's more obvious what `Like` and `Format` actually are?
```suggestion
import pact

from tests.integration.cloud.rest_contracts.conftest import (
    ContractInteraction,
)

DATA_CONTEXT_CONFIGURATION_MIN_RESPONSE_BODY: Final[dict] = {
    "anonymous_usage_statistics": pact.Like(
        {
            "data_context_id": pact.Format().uuid,
            "enabled": False,
```

tests/integration/cloud/rest_contracts/unit_test_mocks/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+    elif isinstance(body, Matcher):
+        return _reify_pact_body(body=_convert_matcher_to_value(matcher=body))
+    elif isinstance(body, dict):
+        for key, value in body.items():
+            if isinstance(value, Matcher):
+                body[key] = _convert_matcher_to_value(matcher=value)
+            body[key] = _reify_pact_body(body=body[key])
+        return body
+    else:
+        return body
```
Kilo59, line 39:
```suggestion
    return body
```

Kilo59 (approved):
Non-blocking comments

tests/integration/cloud/rest_contracts/unit_test_mocks/test_mock_cloud_data_context.py
Kilo59: We could do these asserts inside the existing `mock_cloud_data_context` fixture, and or wrap the existing fixture into a new fixture. `mock_cloud_data_context_but_better_this_time` that does those extra checks.

Something like `assert isinstance(mock_cloud_data_context, CloudDataContext)` seem unnecessary if we are type-checking that part of the test code though.

tests/integration/cloud/rest_contracts/unit_test_mocks/conftest.py
NathanFarmer: I like the original way better. Sure this is less code, but it requires my brain to think slightly more.

tests/integration/cloud/rest_contracts/unit_test_mocks/test_mock_cloud_data_context.py
NathanFarmer: This just confirms that the mock response isn't somehow altered when it is hydrated into the object via the `gx.get_context` codepath. We could perform the assertions in the fixture, but there aren't any tests that use the fixture right now.

tests/integration/cloud/rest_contracts/unit_test_mocks/conftest.py
NathanFarmer: Yes, we can use this method for the rest of the contract tests. Using the unit tests in concert with the contract tests allows the writer of the contract test to confirm that the minimal response actually works in GX code.

tests/integration/cloud/rest_contracts/test_data_context_configuration.py
NathanFarmer: Ok I'll do this, but I need to rename a variable and a method that use the work `pact`.

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -132,7 +132,7 @@ class Config:
     given: pydantic.StrictStr
     response_status: Annotated[int, pydantic.Field(strict=True, ge=100, lt=600)]
     response_body: PactBody
-    request_body: Union[PactBody, None] = None
+    request_body: PactBody | None = None
```
Kilo59: Oh my bad, I wasn't paying attention to the fact that this is a pydantic model.
Have to keep using `Union[A, B]` until we are on `3.9` (or maybe `3.10`) 😭

NathanFarmer: Yeah that's what I thought, but I tested it just in case.

tests/integration/cloud/rest_contracts/unit_test_mocks/conftest.py
NathanFarmer: The word `reify` came from an example written in TS by pact: https://github.com/pact-foundation/pact-js/blob/master/src/v3/matchers.ts#L538

I guess it is a software term that means "turning something abstract into something concrete".

tests/integration/cloud/rest_contracts/conftest.py
NathanFarmer: Pydantic doesn't like this

tests/integration/cloud/rest_contracts/unit_test_mocks/test_mock_cloud_data_context.py
NathanFarmer: I will move these asserts into the fixture in a follow-up PR. I am adding one more contract test to help with writing my "How-to" guide and it will make more sense to do that there.

tests/integration/cloud/rest_contracts/unit_test_mocks/conftest.py
NathanFarmer: This method essentially turns an object with pact constructs into a raw json dictionary.

PR #8861, opened 2023-10-24
[MAINTENANCE] Flakey usage-stats test fixes

Kilo59:
Deal with the increasingly frequent flakey test failures of `tests/integration/usage_statistics/test_usage_statistics_messages.py::test_usage_statistics_message`.

1. Double the `request` `timeout` from `2s` to `4s`
2. Adjuct the `requests_session_with_retries` fixture from a static `backoff_factor` of `1` to `> 1.0`, ensuring that the wait time between retries slightly increases each time.

~We should consider running these tests against the mercury container as a followup.~ Tests rely on `aws` lambda.

### DIFF
--- a/tests/integration/usage_statistics/test_usage_statistics_messages.py
+++ b/tests/integration/usage_statistics/test_usage_statistics_messages.py
@@ -2895,7 +2895,7 @@ def test_usage_statistics_message(
 ):
     """known message formats should be valid"""
     res = requests_session_with_retries.post(
-        USAGE_STATISTICS_QA_URL, json=message, timeout=2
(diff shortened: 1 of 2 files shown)


NathanFarmer (approved):

tests/integration/usage_statistics/test_usage_statistics_messages.py
```diff
@@ -2894,8 +2894,9 @@ def test_usage_statistics_message(
     message: dict, requests_session_with_retries: requests.Session
 ):
     """known message formats should be valid"""
+    # TODO: test against the mercury container
```
Kilo59:
```suggestion
```

PR #8892, opened 2023-10-30
[MAINTENANCE] Quieter logging

Kilo59:
1. Reduce the amount of noisy and unhelpful `INFO` level logging messages.
2. Handle possible exceptions when logging HTTP responses
3. 1 log message for `get_config_with_variables_substituted()` with missing config vars.

### DIFF
--- a/great_expectations/core/http.py
+++ b/great_expectations/core/http.py
@@ -19,6 +19,8 @@ def _log_request_method_and_response(r: requests.Response, *args, **kwargs):
         LOGGER.debug(f"{r}\n{pf(r.json(), depth=3)}")
     except json.JSONDecodeError:
         LOGGER.debug(f"{r}\n{r.content.decode()}")
+    except Exception as other_err:
+        LOGGER.info(f"{r} - Error logging response {other_err!r}")
 
 
 class _TimeoutHTTPAdapter(HTTPAdapter):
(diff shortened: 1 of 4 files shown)


great_expectations/core/http.py
```diff
@@ -19,6 +19,8 @@ def _log_request_method_and_response(r: requests.Response, *args, **kwargs):
         LOGGER.debug(f"{r}\n{pf(r.json(), depth=3)}")
     except json.JSONDecodeError:
         LOGGER.debug(f"{r}\n{r.content.decode()}")
+    except Exception as other_err:
+        LOGGER.info(f"{r} - Error logging response {other_err!r}")
```
Kilo59, line 23: We were getting errors when the agent is run on Databricks (likely a dependency issue).
Likely this will just move the error to when we attempt to parse the json response but it's still an improvement.

NathanFarmer (approved):

great_expectations/core/http.py
rachhouse: Nice, also - TIL about `!r`.

rachhouse (approved):

great_expectations/core/http.py
Kilo59: I'm doing `!r` because I want to see the `__repr__` of the exception instead of just the `__str__` (`!s`).
The `__repr__` includes the name of the exception whereas the `__str__` only shows the exception message.

PR #8903, opened 2023-10-31
[MAINTENANCE] Typing render/view 

Kilo59:
Type checking for `render/view/view.py`

This change includes some runtime changes, notable the removal of the `NoOpTemplate` and the default `_template` class attribute for `DefaultJinjaView`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,6 +85,7 @@ exclude = [
     'render/renderer/suite_edit_notebook_renderer\.py',                  # 7 - This is legacy code and will not be typed.
     'render/renderer/suite_scaffold_notebook_renderer\.py',              # 7 - This is legacy code and will not be typed.
     'render/renderer/datasource_new_notebook_renderer\.py',              # 4 - This is legacy code and will not be typed.
+    'render/renderer/checkpoint_new_notebook_renderer\.py',              # 9 - This is legacy code and will not be typed.
     # END ALWAYS EXCLUDE SECTION ######################################################
     #
     # #################################################################################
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -73,18 +73,24 @@ exclude = [
...
+    'core/usage_statistics/anonymizers/checkpoint_anonymizer\.py',       # 16 - This code will be removed in 1.0
+    'core/usage_statistics/anonymizers/data_docs_anonymizer\.py',        # 5 - This code will be removed in 1.0
+    'core/usage_statistics/anonymizers/datasource_anonymizer\.py',       # 9 - This code will be removed in 1.0
+    'core/usage_statistics/anonymizers/expectation_anonymizer\.py',      # 6 - This code will be removed in 1.0
+    'core/usage_statistics/anonymizers/validation_operator_anonymizer\.py',  # 5 - This code will be removed in 1.0
+    'render/renderer/v3/suite_edit_notebook_renderer\.py',               # 11 - This is legacy code and will not be typed.
+    'render/renderer/v3/suite_profile_notebook_renderer\.py',            # 4 - This is legacy code and will not be typed.
+    'render/renderer/suite_edit_notebook_renderer\.py',                  # 7 - This is legacy code and will not be typed.
+    'render/renderer/suite_scaffold_notebook_renderer\.py',              # 7 - This is legacy code and will not be typed.
+    'render/renderer/datasource_new_notebook_renderer\.py',              # 4 - This is legacy code and will not be typed.
```
Kilo59, line 87: @NathanFarmer do these look right? Am I missing anything?

billdirks: `great_expectations/render/renderer/notebook_renderer.py` is now typed but should it be deleted when we remove the cli?

billdirks (approved):
LGTM, ty!

pyproject.toml
```diff
@@ -136,17 +142,10 @@ exclude = [
     'render/renderer/checkpoint_new_notebook_renderer\.py',                                                # 9
```
NathanFarmer: Missed this one

pyproject.toml
NathanFarmer: Just missing: `render/renderer/checkpoint_new_notebook_renderer.py`

Kilo59: Good catch. Will update both.

NathanFarmer: I think `notebook_renderer.py` will also be removed with the CLI removal. It is the base class for the other notebook renderers.

cdkini (approved):

anthonyburdi (approved):
❤️

pyproject.toml
Kilo59: Thanks!

Issue #8619, opened 2023-08-25
Multiple core expectations don't work when upper case column names are provided with a SQL datasource


strictness:
**Describe the bug**
Some of the core expectations raise a KeyError exception whenever upper case column names are used in expectation calls.

**To Reproduce**
```bash
pip install great_expectations==0.17.12
```

```python
import great_expectations as gx
import sqlite3
from sqlite3 import Error


def create_connection(db_file):
    conn = None
    try:
        conn = sqlite3.connect(db_file)
        return conn
    except Error as e:
        print(e)

    return conn


def run_query(conn, query):
    try:
        c = conn.cursor()
        c.execute(query)
    except Error as e:
        print(e)


database = "pythonsqlite.db"
my_table_name = "projects"
column_name = "name"
sql_create_projects_table = f""" CREATE TABLE IF NOT EXISTS {my_table_name} (
                                    id integer PRIMARY KEY,
                                    {column_name} text NOT NULL
                                ); """

sql_insert_data = f"""INSERT INTO {my_table_name} (id, {column_name}) VALUES (1, 'Cool App with SQLite & Python');"""

# create a database connection
conn = create_connection(database)

# create table
if conn is not None:
    # create projects table
    run_query(conn, sql_create_projects_table)

    # fill table
    run_query(conn, sql_insert_data)
else:
    print("Error! cannot create the database connection.")


context = gx.get_context()

connection_string = "sqlite:///pythonsqlite.db"
datasource = context.sources.add_or_update_sqlite(
    name=my_table_name, connection_string=connection_string
)

datasource.add_table_asset(name=my_table_name, table_name=my_table_name)

batch_request = datasource.get_asset(my_table_name).build_batch_request()

expectation_suite_name = f"{my_table_name}_ge_suite"

context.add_or_update_expectation_suite(expectation_suite_name=expectation_suite_name)
validator = context.get_validator(
    batch_request=batch_request,
    expectation_suite_name=expectation_suite_name,
)

validator.expect_column_values_to_not_be_null(column=column_name)
validator.expect_column_values_to_be_null(column=column_name)
validator.expect_table_row_count_to_be_between(min_value=1, max_value=1000000)
validator.expect_column_distinct_values_to_be_in_set(
    column=column_name, value_set=["Cool App with SQLite & Python"]
)
print("Happy path complete!!!")

expectations = {
    "expect_column_values_to_not_be_null": {"column": column_name.upper()},
[...]

PR #8906, opened 2023-11-01
[MAINTENANCE] Helpful Message for Expectations that require Metrics that are not available

Shinnnyshinshin:
Currently, if an Expectation is misconfigured (such as referencing a non-existent column name), we get a cryptic error message.

```python 
    provided_metrics: Dict[str, MetricValue] = {
>       metric_name: metrics[metric_configuration.id]
        for metric_name, metric_configuration in validation_dependencies.metric_configurations.items()
    }
E   KeyError: ('column_values.nonnull.unexpected_count', '9fb963653447edb4ae8538917d6915bc', ())
```

This PR adds a check in `expectation.py`, where we validate dependencies of the current Expectation against available metrics, and outputs a more helpful error message: 

An error message will look something like this: 
```python
great_expectations.exceptions.exceptions.InvalidExpectationConfigurationError: Metric ('table.row_count', '7054e9d72e974fad769b0964f41e7e5b', "result_format={'result_format': 'BASIC', 'partial_unexpected_count': 20, 'include_unexpected_rows': False}") is not available for validation of {
                 "expectation_type": "expect_column_values_to_not_be_null",
                 "kwargs": {
                   "column": "I_dont_exist",
                   "batch_id": "projects-projects"
                 },
                 "meta": {}
              }. Please check your configuration.
```


tests/expectations/test_expectation.py
```diff
@@ -20,10 +21,42 @@ class FakeColumnPairMapExpectation(expectation.ColumnPairMapExpectation):
...
+    return [
+        MetricConfiguration(
+            metric_name=metric_name,
+            metric_domain_kwargs=metric_domain_kwargs,
+            metric_value_kwargs={},
+        )
+    ]
+
+
+def fake_expectation_config(
```
anthonyburdi, line 53: Good names make things so much easier :heart: Thank you for refactoring this.

JennyTee:
I like the extra error message; that seems much more helpful. Is this the only fix needed for https://github.com/great-expectations/great_expectations/issues/8619, then? Does this mean that for databases with case-insensitive column names, GX users will need to define expectations with case-sensitive column names?

anthonyburdi (commented):
Looks great - just wondering about the `pythonsqlite.db-journal` file added. Could you clarify why that was added?

anthonyburdi (approved):
Very helpful error message, thanks for this PR!

PR #8922, opened 2023-11-02
[FEATURE] Update SnowflakeDatasource connection details 

Kilo59:
Updates the `SnowflakeDatasource` model and schema to nest all connection details inside of the `connection_string` field.
This makes it possible to better represent the methods of providing connection details purely in a `jsonschema` which powers our Cloud UI forms.

Previously `connection_string` was either a string or `None`, and distinct connection details such as `user`, `account`, `password`  etc. were stored in root level fields.

#### Current/Old Format

```python
# current dict/json representation of a Snowflake Datasource
{"name":  ..., "user": ..., "account": ..., "password": ..., "connection_string": None}
# or
{"name": ..., "user": None, "account": None, "password": None, "connection_string": "snowflake://..."}
```

#### New Format
With this change the dict/json representation of a Snowflake Datasource is as follows...
```python
{"name":  ..., "connection_string": { 
    "user": ...,
    "account": ...,
    "password": ...,
    }
}
# or
{"name": ..., "connection_string": "snowflake://..."}
```

## Backwards Compatibility
Logic has been added to the `SnowflakeDatasource` deserializer to convert the old style, root level connection detail fields into a mapping/dict under the `connection_str` fields.

This also means that users can continue to use the same parameters when creating snowflake datasources using `context.sources.add_snowflake()` factory method.

```python
import great_expecations as gx

context = gx.get_context()
# this still works
sf_ds1 = context.sources.add_snowflake("my_datasource",
    user="my_name",
    account=12345,
    password="${MY_PASSWORD}"
)

# so does using a connection string
sf_ds2 = context.sources.add_snowflake("my_datasource_3",
    connection_str="snowflake://..."
)

# this is new
sf_ds3 = context.sources.add_snowflake("my_datasource_2",
    connection_str={"user": "my_name", "account": 12345, "password": "${MY_PASSWORD}"}
)
```

### Breaking Changes
`0.18.2` and lower versions of `great_expectations` will break if they attempt to load a `SnowflakeDatasource` that uses the new format for individual connection details.
Snowflake datasources that use a `connection_string` string will continue to function, regardless of `gx` version.


great_expectations/datasource/fluent/snowflake_datasource.py
```diff
@@ -103,20 +112,11 @@ class SnowflakeDatasource(SQLDatasource):
...
-        if not bool(connection_string) ^ bool(account and user and password):
-            raise ValueError(
-                "Must provide either a connection string or a combination of account, user, and password."
-            )
-        return values
+    class Config:
+        @staticmethod
+        def schema_extra(schema: dict, model: type[SnowflakeDatasource]) -> None:
+            """Customize jsonschema for SnowflakeDatasource."""
+            print(schema)
```
Kilo59, line 119: TODO: `OneOf`

elenajdanova (approved):
lgtm

NathanFarmer (changes requested):
Even with the stub file changes...how to does this preserve auto-complete with everything nested inside of `connection_string`?

NathanFarmer (approved):
Nevermind, I see it was already in the stub file.

anthonyburdi:
Force merged after all CI checks passed to enable releasing today.

PR #8958, opened 2023-11-11
[MAINTENANCE] Refactor Cloud end-to-end tests for ease of maintenance

NathanFarmer:
- Move common logic into conftest fixtures
- Parametrize adding Data Assets to the Datasource
- Documentation


tests/integration/cloud/end_to_end/conftest.py
```diff
@@ -37,31 +41,119 @@ def context() -> CloudDataContext:
 
 
 @pytest.fixture(scope="module")
-def tmp_dir(tmpdir_factory) -> py.path:
-    return tmpdir_factory.mktemp("project")
+def datasource_name(
+    context: CloudDataContext,
+) -> Iterator[str]:
```
Kilo59, line 45:
```suggestion
) -> Generator[str, None, None]:
```
`#non-blocking`
You could also type this like this.
https://docs.python.org/3/library/typing.html#typing.Generator

I think I have a slight preference for using `Generator`. I think newer developers might be confused by seeing `Iterator` here. Although I think they'll be confused by pytest magic regardless so 🤷 . 
But as I said, not a blocker.

tests/integration/cloud/end_to_end/conftest.py
```diff
@@ -37,31 +41,119 @@ def context() -> CloudDataContext:
...
+    )
+    with pytest.raises(gx_exceptions.DataContextError):
+        context.get_checkpoint(name=checkpoint_name)
 
 
 @pytest.fixture(scope="module")
-def get_missing_checkpoint_error_type() -> type[Exception]:
-    return gx_exceptions.DataContextError
+def tmp_dir(tmpdir_factory) -> py.path:
+    return tmpdir_factory.mktemp("project")
```
Kilo59: `tmpdir` and `tmpdir_factory` are now "legacy" and pytest recommends using `tmp_path` and `tmp_path_factory`. You can use them the same way, they just return modern `pathlib.Path` objects instead of `py.path`.

https://docs.pytest.org/en/7.4.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures

Kilo59 (approved):
Approved pending updating the `tmp_dir` to use `tmp_path`

tests/integration/cloud/end_to_end/conftest.py
NathanFarmer: I find `Generator[str, None, None]` more confusing. The new developer has to spend the time figuring out what the `None`s represent instead of just using it like an `Iterator`.

PR #8961, opened 2023-11-14
[MAINTENANCE] Make config mandatory in Expectation constructor

cdkini:
Currently, the Expectation constructor takes in an optional configuration - if this is not provided during instantiation, it needs to be provided as a runtime override. This creates situations where an Expectation may not have requisite state to work as intended within downstream callers. Additionally, it defers configuration issues and proper feedback to users; we should always fail fast.


great_expectations/expectations/expectation.py
```diff
@@ -350,12 +350,17 @@ class Expectation(metaclass=MetaExpectation):
     examples: ClassVar[List[dict]] = []
 
     def __init__(
-        self, configuration: Optional[ExpectationConfiguration] = None
```
cdkini, line 353: Breaking change ❗

great_expectations/expectations/expectation.py
```diff
@@ -1292,10 +1288,6 @@ def validate(  # noqa: PLR0913
 
     @property
     def configuration(self) -> ExpectationConfiguration:
-        if self._configuration is None:
```
tyler-hoffman, line 1295: oh man I love seeing stuff like this go away

great_expectations/core/expectation_suite.py
```diff
@@ -754,8 +754,7 @@ def _validate_expectation_configuration_before_adding(
     ):
         try:
             class_ = get_expectation_impl(expectation_configuration.expectation_type)
-            expectation = class_()
-            expectation.validate_configuration(expectation_configuration)
+            _ = class_(expectation_configuration)  # Implicitly validates in constructor
```
billdirks, line 757: A little unrelated but can you fix the docstring for `ExpectationConfiguration`? I think `denies` should be `defines`. [See here](https://github.com/great-expectations/great_expectations/blob/d7ff25553aac1a05736c4ca2979f63d5b17745e7/great_expectations/core/expectation_configuration.py#L135).

billdirks: nit: you can remove the `_ =`. I think this will allocate a variable on the stack that you don't need. Your comment explains why are you doing this.

billdirks (approved):
LGTM, thanks!

Issue #6286, opened 2022-11-04
`expect_column_values_to_be_unique` does not work on MySQL 8.0.19


liorfuks:
When running the command 

```
validator.expect_compound_columns_to_be_unique(column_list=['employee_id', 'first_name'])
```
The following error appears: 

```
MetricResolutionError: An SQL execution Exception occurred: (pymysql.err.OperationalError) (1137, "Can't reopen table: 'ge_temp_7f875c6e'")
[SQL: SELECT `UnexpectedCountSubquery`.unexpected_count 
FROM (SELECT sum(`condition`) AS unexpected_count 
FROM (SELECT CASE WHEN (records_with_grouped_column_counts_subquery._num_rows >= %(num_rows_1)s) THEN CAST(%(param_1)s AS DECIMAL) ELSE CAST(%(param_2)s AS DECIMAL) END AS `condition` 
FROM (SELECT original_table_clause.employee_id AS employee_id, original_table_clause.first_name AS first_name, original_table_clause.last_name AS last_name, original_table_clause.user_id AS user_id, original_table_clause.email_primary_work AS email_primary_work, original_table_clause.email_other_work AS email_other_work, original_table_clause.`year_month` AS `year_month`, group_counts_subquery._num_rows AS _num_rows 
FROM (SELECT employee_id, first_name, last_name, user_id, email_primary_work, email_other_work, `year_month` 
FROM (SELECT * 
FROM ge_temp_7f875c6e 
WHERE NOT (employee_id IS NULL AND first_name IS NULL)) AS anon_2) AS original_table_clause INNER JOIN (SELECT employee_id, first_name, count(*) AS _num_rows 
FROM (SELECT employee_id, first_name, last_name, user_id, email_primary_work, email_other_work, `year_month` 
FROM (SELECT * 
FROM ge_temp_7f875c6e 
WHERE NOT (employee_id IS NULL AND first_name IS NULL)) AS anon_2) AS original_table_clause GROUP BY employee_id, first_name) AS group_counts_subquery ON group_counts_subquery.employee_id = original_table_clause.employee_id AND group_counts_subquery.first_name = original_table_clause.first_name) AS records_with_grouped_column_counts_subquery) AS anon_1) AS `UnexpectedCountSubquery`]
[parameters: {'num_rows_1': 2, 'param_1': 1, 'param_2': 0}]
(Background on this error at: http://sqlalche.me/e/14/e3q8).
```

[...]

PR #8964, opened 2023-11-15
[BUGFIX] Fixed compound_columns_to_be_unique bug caused by the reuse of temporary tables in MySQL #6286

FilipeJesus:
This PR addresses the bug found in #6286, this issue was closed a month ago but my team experienced the same issues on version 0.18.2 of great_expectations.

The bug is related to a limitation of MySQL which does not allow temporary tables to be used twice in the same query. The failing query selected against the temporary table and then joined to a sub query which grouped the temporary table and counted the number of duplicate rows. This was done so that the final output table had the same number of rows as the original table but also had a new column '_num_rows' containing the duplicate column count.

There is another approach which we can take to get the same final table without using any subqueries, instead we can use window functions to partition over the column list and calculate the number of duplicates. As I mainly work with MySQL databases I am not sure if the partition window function is available on all supported databases, therefore I have implemented this alternative approach just for MySQL, similar to the approach taken for #2389.

i have updated one of the tests to add a mysql schema for compound_columns_to_be_unique, not sure if this is sufficient. Please provide some guidance is more testing is required.


FilipeJesus:
@cla-bot check

FilipeJesus:
@cla-bot check

anthonyburdi (approved):
Thank you for this contribution @FilipeJesus!

PR #8965, opened 2023-11-15
[MAINTENANCE] Have `Expectation` inherit from `pydantic.BaseModel`

cdkini:
Preliminary PR to introduce rich typing and validation to the constructors of Expectations - there are a lot of minor tweaks we need to make before this becomes an actuality so just setting up the ground work here. 


great_expectations/expectations/expectation.py
```diff
@@ -338,7 +348,7 @@ class Expectation(metaclass=MetaExpectation):
         "result_format",
     )
     default_kwarg_values: ClassVar[
-        dict[str, bool | str | float | RuleBasedProfilerConfig | None]
+        Dict[str, Union[bool, str, float, RuleBasedProfilerConfig, None]]
```
Kilo59: I think pydantic ignores `ClassVar` items so you should still be able to use future annotations here.

Did something break?

cdkini: Yeah it seemed to break things

great_expectations/expectations/expectation.py
```diff
@@ -329,7 +337,11 @@ class Expectation(metaclass=MetaExpectation):
         2. Data Docs rendering methods decorated with the @renderer decorator. See the
     """
 
-    version: ClassVar = ge_version
+    __metaclass__ = MetaExpectation
```
cdkini: Alleviates the errors I was seeing with metaclass conflicts but no longer registers as intended

great_expectations/expectations/core/expect_column_values_to_not_be_null.py
```diff
@@ -90,7 +90,7 @@ class ExpectColumnValuesToNotBeNull(ColumnMapExpectation):
     }
 
     map_metric: ClassVar[str] = "column_values.nonnull"
-    args_keys: ClassVar[tuple[str, ...]] = ("column",)
+    args_keys: ClassVar[Tuple[str, ...]] = ("column",)
```
cdkini, line 93: All these typing changes are to appease Pydantic

great_expectations/expectations/expectation.py
```diff
@@ -329,7 +331,11 @@ class Expectation(metaclass=MetaExpectation):
         2. Data Docs rendering methods decorated with the @renderer decorator. See the
     """
 
-    version: ClassVar = ge_version
+    class Config:
+        arbitrary_types_allowed = True
+        extra = pydantic.Extra.allow
```
cdkini, line 336: Essentially `**kwargs` but necessary as an intermediate step before we add defined params throughout the hierarchy

great_expectations/expectations/expectation.py
```diff
@@ -349,9 +355,17 @@ class Expectation(metaclass=MetaExpectation):
...
+        super().__init__(**kwargs)
+
+        # Everything below is purely to maintain current validation logic
+        expectation_type = camel_to_snake(self.__class__.__name__)
+        configuration = ExpectationConfiguration(
+            expectation_type=expectation_type,
+            kwargs=kwargs,
+        )
         self.validate_configuration(configuration)
+        self._configuration = configuration
```
cdkini: This validation logic will be moved over to Pydantic validators but needs to sit here for now. Additionally, I'm not sure we can easily remove `self._configuration`; I'm going to keep it around throughout this work and then circle back to remove it once we've met our primary objectives.

great_expectations/expectations/expectation.py
```diff
@@ -1239,8 +1253,9 @@ def validate_configuration(
         except AssertionError as e:
             raise InvalidExpectationConfigurationError(str(e))
 
+    # Renamed from validate due to collision with Pydantic method of the same name
     @public_api
-    def validate(  # noqa: PLR0913
+    def validate_(  # noqa: PLR0913
```
cdkini, line 1269: Not sure if this method will even be around after we finish this work - I'll take name recs though

great_expectations/expectations/core/expect_column_values_to_not_be_null.py
tyler-hoffman: So does pydantic not support `list`, `dict`, `set`, etc?

great_expectations/expectations/expectation.py
```diff
@@ -265,7 +267,7 @@ def wrapper(
 
 
 # noinspection PyMethodParameters
-class MetaExpectation(ABCMeta):
+class MetaExpectation(ModelMetaclass):
```
tyler-hoffman, line 270: Why do we use `ModelMetaClass` and inherit from `BaseModel`? Are we losing out on any ABC functionality here? Googling `ModelMetaClass` leads me to believe it is not public in the most recent release.

great_expectations/expectations/expectation.py
```diff
@@ -338,7 +344,7 @@ class Expectation(metaclass=MetaExpectation):
         "result_format",
     )
     default_kwarg_values: ClassVar[
-        dict[str, bool | str | float | RuleBasedProfilerConfig | None]
+        dict[str, Union[bool, str, float, RuleBasedProfilerConfig, None]]
```
tyler-hoffman, line 349: Does pydantic not allow the other syntax I'm assuming?

great_expectations/expectations/core/expect_column_values_to_not_be_null.py
Kilo59: It does but `tuple`s are immutable and therefore better 😄 

Edit:
Oh I see you are asking about `typing.Tuple` vs the built-in `tuple`.
The issue is our minimum supported version of `python` doesn't support these without using postponed annotations which don't exist at runtime.

great_expectations/expectations/expectation.py
Kilo59: It relies on the types to validate at runtime.
The minimum version in OSS is python 3.8, which doesn't support the alternative union syntax.

You can use the `from __future__ import annotations` to use the alternative union syntax for type-checking, but because pydantic also relies on these types at runtime we have to use the old `Union` syntax.
https://mypy.readthedocs.io/en/stable/kinds_of_types.html?highlight=union#x-y-syntax-for-unions

great_expectations/expectations/expectation.py
joshua-stauffer: +1 that [we probably should not use this](https://github.com/pydantic/pydantic/issues/6381#issuecomment-1619296261) given that we want to upgrade to pydantic v2 some day.

great_expectations/expectations/expectation.py
joshua-stauffer: how about `validate_configuration`?

cdkini: that's already a method haha I'll come up with something more suitable than what I have now once we figure out the Pydantic issue

great_expectations/expectations/expectation.py
cdkini: Great call out - I'll do some research to see how we can alleviate this concern

Kilo59: We already rely on this for FDS datasources.

pydantic v1 is embedded inside pydantic v2, and will be for a while.

If the goal is simply to allow `Expectations` to be a pydantic model (while preserving the Metaclass registration behavior) I don't see an alternative that doesn't add unecessary complexity.

Kilo59 (approved):
If you plan to switch to pydantic v2 for V1 I would just start by making this a v2 model.

If not `ModelMetaclass` is the correct way to do this. Don't overcomplicate it.

great_expectations/expectations/core/expect_column_values_to_not_be_null.py
tyler-hoffman: Sorry, I was referring to `tuple` vs `Tuple` and `list` vs `List`

great_expectations/expectations/expectation.py
tyler-hoffman: Thanks!

tyler-hoffman (approved):

great_expectations/expectations/core/expect_column_values_to_not_be_null.py
Kilo59: Yeah, sorry this is the same reason we have to use `Union[str, int]` instead of `str | int`.
https://github.com/great-expectations/great_expectations/pull/8965#discussion_r1397834916

great_expectations/expectations/expectation.py
```diff
@@ -296,8 +297,15 @@ def __new__(cls, clsname, bases, attrs):
         return newclass
 
 
+class ExpectationBase(pydantic.BaseModel):
+    class Config:
+        arbitrary_types_allowed = True
+
+    pass
```
Kilo59: Sorry, I should have suggested this earlier. If you don't intend this to be extended, I would make this private.
```python
class _ExpectationBase(pydantic.BaseModel)
   ...
```

Alternatively, just get rid of this class...
https://docs.pydantic.dev/1.10/usage/model_config/
>Also, you can specify config options as model class kwargs:
```python
@public_api
class Expectation(pydantic.BaseModel, arbitrary_types_allowed = True)
   ...
```

PR #8967, opened 2023-11-15
[FEATURE] validator 1.0

tyler-hoffman:
Adds the start of a new validator class that only validates expectations and suites against a batch config.

There's more work to do here in terms of batch config options, but I suspect that can come in a subsequent PR.


great_expectations/core/batch_config.py
```diff
@@ -0,0 +1,6 @@
+from typing_extensions import TypeAlias
+
+from great_expectations.datasource.fluent.sql_datasource import QueryAsset
+
+# TODO: fix this type
+BatchConfig: TypeAlias = QueryAsset
```
billdirks: The type here should be [DataAsset](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/datasource/fluent/interfaces.py#L161) which is the superclass of all data assets and has [`build_batch_request` as part of it's interface ](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/datasource/fluent/interfaces.py#L214).

I think we want this as a new type that composes in the DataAsset from which it is derived. I'm worried the API for the config object will be different than that of an asset. For example, splitters and sorters will probably need to be removed from assets and be made part of the `BatchConfig`.

great_expectations/validator/v1_validator.py
```diff
@@ -0,0 +1,92 @@
...
+                "unsuccessful_expectations": statistics.unsuccessful_expectations,
+                "success_percent": statistics.success_percent,
+            },
+        )
+
+    def _validate_expectation_configs(
+        self, expectation_configs: list[ExpectationConfiguration]
+    ) -> list[ExpectationValidationResult]:
+        """Run a list of expectation configurations against the batch config"""
+        batch_request = self.batch_config.build_batch_request()
```
billdirks: You say this the PR description but also wanted to highlight it. This will always create a batch that is the whole data asset. We will need to pass the `batch_asset_options` here to create a batch that is a subset of data.

great_expectations/validator/v1_validator.py
```diff
@@ -0,0 +1,92 @@
...
+    result_format: ResultFormat
+
+    def __init__(
+        self,
+        context: AbstractDataContext,
+        batch_config: BatchConfig,
+        result_format: ResultFormat = ResultFormat.SUMMARY,
+        batch_asset_options: dict | None = None,
+    ) -> None:
+        self.context = context
```
billdirks: I don't think we want people mutating `context`,  `batch_config`, or `batch_asset_options`. Because of this I would prefix the instance variables with a `_`, eg `self._context = context`.

tests/conftest.py
```diff
@@ -3236,21 +3236,40 @@ def generate_ascending_list_of_datetimes(
...
             "test_sets", "test_cases_for_sql_data_connector.db"
         ),
     )
+    return f"sqlite:///{db_file_path}"
+
+
+@pytest.fixture
+def fds_data_context(
+    sa, empty_data_context: AbstractDataContext, sqlite_connection_string: str
+):
```
billdirks: Can you annotate the return type here?

tests/conftest.py
```diff
@@ -3236,21 +3236,40 @@ def generate_ascending_list_of_datetimes(
 
 
 @pytest.fixture
-def data_context_with_simple_sql_datasource_for_testing_get_batch(
-    sa, empty_data_context
-):
-    context = empty_data_context
-
+def sqlite_connection_string():
```
billdirks: It looks like the previous fixture didn't annotate the return type. I'm not too worried about it but it might be good to do.

tests/conftest.py
```diff
@@ -3236,21 +3236,40 @@ def generate_ascending_list_of_datetimes(
...
+        name="trip_asset",
+        query="SELECT * FROM table_partitioned_by_date_column__A",
+    )
+    return context
+
+
+@pytest.fixture
+def data_context_with_simple_sql_datasource_for_testing_get_batch(
+    sa, empty_data_context, sqlite_connection_string
+):
```
billdirks: Return type here also.

tests/validator/test_v1_validator.py
```diff
@@ -0,0 +1,158 @@
...
+)
+from great_expectations.expectations.core.expect_column_values_to_be_in_set import (
+    ExpectColumnValuesToBeInSet,
+)
+from great_expectations.expectations.expectation import Expectation
+from great_expectations.validator.v1_validator import ResultFormat, Validator
+
+
+@pytest.fixture
+def context(
```
billdirks: Do you need this or can you use the `fds_data_context` fixture directly?

billdirks (commented):
Thanks for doing this. I left some comments and am happy to talk sync about them.

tests/conftest.py
Kilo59: We should start type-checking more of our test modules.
The root `tests/conftest.py` would be a good place to start since so much relies on it.

great_expectations/validator/v1_validator.py
tyler-hoffman: That's fair. I originally thought all of these (except context) should function as public attributes, but thinking about it again, I'm with you. Will update.

tests/validator/test_v1_validator.py
tyler-hoffman: Holdover from when I was using a different fixture with a very long name and I wanted to alias it in one place. I'll remove this 😄

great_expectations/validator/v1_validator.py
```diff
@@ -0,0 +1,92 @@
...
+class Validator:
+    """Validator.
+
+    Responsible for running expectations on a batch configuration.
+    """
+
+    context: AbstractDataContext
+    batch_config: BatchConfig
+    batch_asset_options: dict | None
+    result_format: ResultFormat
```
cdkini: Do we need these top level annotations?

great_expectations/validator/v1_validator.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def __init__(
+        self,
+        context: AbstractDataContext,
+        batch_config: BatchConfig,
+        result_format: ResultFormat = ResultFormat.SUMMARY,
+        batch_asset_options: dict | None = None,
```
cdkini: Nit - where should context come in terms of ordering? I think we should be consistent across our API or require mandatory key-word args to improve UX

great_expectations/validator/v1_validator.py
```diff
@@ -0,0 +1,92 @@
...
+        self.batch_asset_options = batch_asset_options
+        self.result_format = result_format
+
+    def validate_expectation(
+        self, expectation: Expectation
+    ) -> ExpectationValidationResult:
+        """Run a single expectation against the batch config"""
+        results = self._validate_expectation_configs([expectation.configuration])
+
+        assert len(results) == 1
```
cdkini, line 56: Could we raise a `ValueError` instead? We can bypass asserts with a Python flag

great_expectations/validator/v1_validator.py
```diff
@@ -0,0 +1,92 @@
...
+        # TODO: This was copy/pasted from Validator, but many fields were removed
+        return ExpectationSuiteValidationResult(
+            results=results,
+            success=statistics.success,
+            statistics={
+                "evaluated_expectations": statistics.evaluated_expectations,
+                "successful_expectations": statistics.successful_expectations,
+                "unsuccessful_expectations": statistics.unsuccessful_expectations,
+                "success_percent": statistics.success_percent,
+            },
```
cdkini, line 75: Not pressing but I'd like to see `statistics` become a `TypedDict`

tests/conftest.py
cdkini: yes please - let's treat test code like production code.

I'd humbly request annotations on all fixture definitions and usages

tests/validator/test_v1_validator.py
```diff
@@ -0,0 +1,158 @@
...
+def validator(context: AbstractDataContext, batch_config: BatchConfig) -> Validator:
+    return Validator(
+        context=context,
+        batch_config=batch_config,
+        batch_asset_options=None,
+        result_format=ResultFormat.SUMMARY,
+    )
+
+
+@pytest.mark.big
```
cdkini: Hmm why are these `big`? Seem pretty lightweight to me.

tests/validator/test_v1_validator.py
```diff
@@ -0,0 +1,158 @@
...
+@pytest.fixture
+def batch_config(context: AbstractDataContext) -> BatchConfig:
+    datasource = context.get_datasource("sqlite_datasource")
+    return datasource.get_asset("trip_asset")
+
+
+@pytest.fixture
+def validator(context: AbstractDataContext, batch_config: BatchConfig) -> Validator:
+    return Validator(
+        context=context,
```
cdkini: Could the context be mocked or is it required here?

great_expectations/validator/v1_validator.py
tyler-hoffman: I don't have super strong opinions here, but here was why I just put an `assert`: AFAIK we know that if we pass in a single configuration, we'll get out exactly one. In the past, I've used asserts when the point is to tell mypy or future devs something we know has to be true that mypy can't infer. Whereas a value error is about saying a bad value was passed into the function.

But I could totally not be following best practices. WDYT?

great_expectations/validator/v1_validator.py
tyler-hoffman: I want it to just be the `ValidationStatistics` object we get back personally, but I figured one thing at a time.

tests/validator/test_v1_validator.py
tyler-hoffman: I don't have a good answer there. TBH I was copying the pytest markers from `test_validator`, and I think it used `big` everywhere. I suspected because we actually perform a validation against a sqlite datasource on disk. Should I change it to something else (and if so what?)

great_expectations/validator/v1_validator.py
tyler-hoffman: Seems reasonable-ish. I don't have strong opinions, but in any code I've written, I've tended to think in terms of putting the most "core" entities first. Nothing makes sense without context, so I put that first. But I'm fine to go with whatever. I love mandating keyword args for internal use. Do we like forcing that on oss users?

cdkini: Just mainly thinking out loud - I also don't care too much here.

I agree with probably not forcing keyword args for users let's leave what you have.

great_expectations/validator/v1_validator.py
cdkini: Ah I didn't realize this was an invariant of sorts - yeah I agree with your decision here. I'd love to codify this distinction in our internal coding standards

great_expectations/validator/v1_validator.py
cdkini: Yup no issues from me here as long as the final V1 implementation is richly typed

tests/validator/test_v1_validator.py
cdkini: I think `unit` is appropriate - you can read about our markers in `pyproject.toml`

```
 "big: mark tests that don't have external dependencies but aren't unit tests",
```

tyler-hoffman: Oh thanks! didn't know we documented the markers there. Perfect! Should I actually be using `external_sqldialect`, as I'm using [this fixture](https://github.com/great-expectations/great_expectations/blob/develop/tests/conftest.py#L516)?

cdkini: hmm I'm actually not too sure here - could go either way I'll defer to you

great_expectations/validator/v1_validator.py
```diff
@@ -0,0 +1,92 @@
...
+class Validator:
+    """Validator.
+
+    Responsible for running expectations on a batch configuration.
+    """
+
+    _context: AbstractDataContext
+    _batch_config: BatchConfig
+    _batch_asset_options: dict | None
+    result_format: ResultFormat
```
cdkini: I don't think we need these annotations here at all right? If it were a dataclass then yes but otherwise I think we're covered by the `__init__`?

great_expectations/validator/v1_validator.py
```diff
@@ -0,0 +1,92 @@
...
+        self,
+        context: AbstractDataContext,
+        batch_config: BatchConfig,
+        result_format: ResultFormat = ResultFormat.SUMMARY,
+        batch_asset_options: dict | None = None,
+    ) -> None:
+        self._context = context
+        self._batch_config = batch_config
+        self._batch_asset_options = batch_asset_options
+        self.result_format = result_format
```
cdkini, line 48: Could we also make this private? If we want to expose it we can do so with a `property`?

tyler-hoffman: I'm fine either way, but what does a property get us that an attribute doesn't?

cdkini: we can make it read-only - essentially just introducing getters/setters to limit access if that's what we desire

great_expectations/core/batch_config.py
```diff
@@ -0,0 +1,14 @@
+from dataclasses import dataclass
+
+from great_expectations.datasource.fluent.interfaces import DataAsset
+
+
+@dataclass(frozen=True)
```
billdirks: We can do this in the future PR.

This should probably be a pydantic model since we will want to serialize/deserialize from config.
When we do this, we will need to think about serialization/deserialization since we likely want to reference the data asset by name in the config. Maybe our config model objects should live in the same directory?

great_expectations/validator/v1_validator.py
```diff
@@ -0,0 +1,90 @@
...
+        """Run a single expectation against the batch config"""
+        results = self._validate_expectation_configs([expectation.configuration])
+
+        assert len(results) == 1
+        return results[0]
+
+    def validate_expectation_suite(
+        self, expectation_suite: ExpectationSuite
+    ) -> ExpectationSuiteValidationResult:
+        """Run an  expectation suite against the batch config"""
```
billdirks: nit: there are 2 spaces in `an  expectation`

great_expectations/validator/v1_validator.py
```diff
@@ -0,0 +1,90 @@
...
+                "unsuccessful_expectations": statistics.unsuccessful_expectations,
+                "success_percent": statistics.success_percent,
+            },
+        )
+
+    def _validate_expectation_configs(
+        self, expectation_configs: list[ExpectationConfiguration]
+    ) -> list[ExpectationValidationResult]:
+        """Run a list of expectation configurations against the batch config"""
+        batch_request = self._batch_config.data_asset.build_batch_request(
```
billdirks: We'll have to think about the API that a batch asset makes for splitting up the data into batches. We do have a lot of DataAssets and want adding more to be formulaic for us or contributors. Right now batching the data is determined by the splitter on the data asset. In the new world, the data asset is the data and the info about how you want to split it into batches lives in the batch config. Maybe this means `data_asset.build_batch_asset` will take in the splitters as well as the `options`? 

It feels like here, the validator should call a method on the batch config to get the `batch_request`, eg `self._batch_config.build_batch_request(...)`. Right now this would call the identical function on the asset vs reaching into the `self._batch_config.data_asset`. I feel this way because I think the splitters will live on `self._batch_config` and the validator code here will not need to explicitly know about splitters, it just wants a `batch_request`. 

The ask for now would be to move `build_batch_request` up to the `batch_config`. We can do that in a followup PR when more functionality gets added if you prefer.

great_expectations/validator/v1_validator.py
```diff
@@ -0,0 +1,90 @@
...
+            options=self._batch_asset_options
+        )
+        wrapped_validator = self._context.get_validator(batch_request=batch_request)
```
billdirks: For a `Validator`, `batch_request` and `wrapped_validator` are invariant on non-mutating initialization variables. We should probably cache and reuse them. One advantage is the validator will store query results. That way, if one uses this validator interactively, it can avoid running expensive queries if the expectation graphs overlap.

great_expectations/validator/v1_validator.py
billdirks: This is something the user may want to change while interactively playing with the Validator so we want it to be mutable. We could leave it public now and change to a property later if we want.

great_expectations/validator/v1_validator.py
tyler-hoffman: Oh cool. I didn't realize it stored query results!

great_expectations/validator/v1_validator.py
tyler-hoffman: Seems reasonable to me. I snuck that change in here.

billdirks (approved):
Thank you!

PR #8977, opened 2023-11-17
[MAINTENANCE] Add wrapper for posthog analytics as well as an initial event

wookasz:
Add basic analytics framework and use it in a single "data_context.initialized" event


great_expectations/analytics/config.py
```diff
@@ -0,0 +1,59 @@
...
+    @property
+    def posthog_project_api_key(self):
+        if self.gx_posthog_project_api_key:
+            return self.gx_posthog_project_api_key
+
+        return self._gx_cloud_dev_posthog_project_api_key
+
+
+class Config(GenericModel):
+    organization_id: UUID = UUID("00000000-0000-0000-0000-000000000000")
```
tyler-hoffman: Is a fake UUID better than `None`? If so, should/could we make it a named constant e.g. DUMMY_UUID

great_expectations/analytics/base_event.py
```diff
@@ -0,0 +1,82 @@
...
+    @property
+    def user_id(self) -> UUID:
+        return get_config().user_id
+
+    _allowed_actions: ClassVar[Optional[List[Action]]] = None
+
+    def __post_init__(self):
+        allowed_actions = self.get_allowed_actions()
+        if (
+            allowed_actions is not None
```
tyler-hoffman, line 53: I feel like this protection isn't really protecting us from much; subclasses have to opt in to using it, and then the subclass defines both its action and that the action is allowed. Maybe there are cases I'm missing though.

great_expectations/analytics/events.py
```diff
@@ -0,0 +1,13 @@
...
+from great_expectations.analytics.base_event import Action, Event
+
+
+class DataContextInitializedEvent(Event):
+    action = DATA_CONTEXT_INITIALIZED
+
+    _allowed_actions: ClassVar[List[Action]] = [DATA_CONTEXT_INITIALIZED]
+
+    def __init__(self):
+        super().__init__(action=self.action)
```
tyler-hoffman: The flow here feels off to me. This and the parent class both define an `action` field, and we read this during init and pass it to the parent, which under the hood assigns it to self.action again. I would remove line 8 and just call `super().__init__(action=DATA_CONTEXT_INITIALIZED)`

tyler-hoffman (approved):
Left some thoughts on the event.action stuff.

I think we should get some basic testing in place on this - does initializing the posthog config actually work? Does submit capture what it looks like we capture? But I won't block the merge on that.

Seems like this should be fairly ergonomic!

great_expectations/analytics/config.py
wookasz: Yes, Tetris preferred dummies in the past

great_expectations/analytics/base_event.py
wookasz: @tyler-hoffman in the case of this specific `Event`, there was only one `Action` that can be associated with it and to make using the event a bit easier, I set it in the ctor. However, there's another, more general pattern here. For example, a `CheckpointEvent` may support multiple actions, e.g. `checkpoint.created`, `checkpoint.updated`, `checkpoint.deleted`.

The use of `allowed_actions` become clearer once we add code to generate an event schema file for use in our analytics pipelines.

PR #8985, opened 2023-11-20
[MAINTENANCE] Start adding shared fields to Expectation model

cdkini:
Start adding fields to Expectation models


great_expectations/expectations/expectation.py
```diff
@@ -2363,6 +2370,11 @@ class BatchExpectation(Expectation, ABC):
             expectation.
     """
 
+    batch_id: Union[str, None] = None
+    table: Union[str, None] = None
+    row_condition: Union[str, None] = None
+    condition_parser: Union[str, None] = None
```
cdkini: Not so sure about these just yet

great_expectations/core/result_format.py
```diff
@@ -0,0 +1,14 @@
...
+
+class ResultFormat(str, enum.Enum):
+    BOOLEAN_ONLY = "BOOLEAN_ONLY"
+    BASIC = "BASIC"
+    COMPLETE = "COMPLETE"
+    SUMMARY = "SUMMARY"
+
+
+class ResultFormatDict(TypedDict):
+    result_format: ResultFormat
```
cdkini: Don't fully know the shape of this but this will provide some basic structure until we refactor/rework result format

great_expectations/expectations/core/expect_column_values_to_be_increasing.py
```diff
@@ -97,25 +96,6 @@ class ExpectColumnValuesToBeIncreasing(ColumnMapExpectation):
...
-
-        Args:
-            configuration: An `ExpectationConfiguration` to validate. If no configuration is provided, it will be pulled
-                           from the configuration attribute of the Expectation instance.
-
-        Raises:
-            InvalidExpectationConfigurationError: The configuration does not contain the values required by the
-                                                  Expectation.
-        """
-        super().validate_configuration(configuration)
```
cdkini: This is done quite often in our codebase - no need to override if you're just going to leverage the parent impl

great_expectations/expectations/expectation.py
```diff
@@ -1245,24 +1252,7 @@ def get_result_format(
...
-        """
-        if not configuration:
-            configuration = self.configuration
-        try:
-            assert (
-                configuration.expectation_type == self.expectation_type
-            ), f"expectation configuration type {configuration.expectation_type} does not match expectation type {self.expectation_type}"
-        except AssertionError as e:
-            raise InvalidExpectationConfigurationError(str(e))
+        pass  # no-op
```
cdkini, line 1261: Would fully delete this but until we 100% shift over to Pydantic validation, we'll require this method on the base class to perform our hand-rolled validation. Child classes still depend on the existence of this method in the hierarchy.

great_expectations/expectations/expectation.py
```diff
@@ -2533,32 +2528,6 @@ def _validate_metric_value_between(  # noqa: C901, PLR0912, PLR0913
         return {"success": success, "result": {"observed_value": metric_value}}
 
 
-@public_api
-class TableExpectation(BatchExpectation, ABC):
```
cdkini, line 2537: Deleted the deprecated classes while I was in here

great_expectations/expectations/expectation.py
```diff
@@ -2693,53 +2660,11 @@ class ColumnAggregateExpectation(BatchExpectation, ABC):
...
 
+    column: str
+
     domain_keys = ("batch_id", "table", "column", "row_condition", "condition_parser")
     domain_type = MetricDomainTypes.COLUMN
 
-    @override
-    def validate_configuration(
-        self, configuration: Optional[ExpectationConfiguration] = None
-    ) -> None:
```
cdkini, line 2702: Starting to chip away at remove this hand-rolled validation funcs within the base classes

great_expectations/expectations/expectation.py
```diff
@@ -3337,31 +3235,14 @@ class MulticolumnMapExpectation(BatchExpectation, ABC):
         "condition_parser": None,  # we expect this to be explicitly set whenever a row_condition is passed
         "mostly": 1,
         "ignore_row_if": "all_values_are_missing",
-        "result_format": "BASIC",
-        "include_config": True,
-        "catch_exceptions": True,
```
cdkini: Both these fields are irrelevant moving forward - I'm going to delete them as I go

great_expectations/warnings.py
```diff
@@ -9,6 +9,7 @@
 def warn_deprecated_parse_strings_as_datetimes() -> None:
     """Warn when using parse_strings_as_datetimes in expectations."""
     # deprecated-v0.13.41
+    # NOTE: Should write ticket to fully delete this support
```
cdkini: Note to self - raise error in this method and delete whatever fails

tests/expectations/core/test_expect_column_max_to_be_between.py
```diff
@@ -1,51 +0,0 @@
...
-import pytest
-
-from great_expectations.core.batch import RuntimeBatchRequest
-from great_expectations.data_context import DataContext
-
-
-@pytest.mark.filesystem
-def test_expect_column_max_to_be_between_warn_parse_strings_as_datetimes(
-    data_context_with_datasource_pandas_engine,
-):
```
cdkini, line 12: All these tests test for a path we no longer support - just deleting them outright

tests/expectations/test_expectation.py
```diff
@@ -168,7 +169,7 @@ def test_expectation_succeeds_with_valid_mostly(fake_expectation_cls, config):
 def test_multicolumn_expectation_validation_errors_with_bad_mostly(
     fake_expectation_cls, config
 ):
-    with pytest.raises(InvalidExpectationConfigurationError):
+    with pytest.raises(pydantic.ValidationError):
```
cdkini, line 172: A breaking change but we should leverage Pydantic's type instead of our own configuration error if we're going to be using models

tests/expectations/test_generate_diagnostic_checklist.py
```diff
@@ -30,9 +30,9 @@ def test_print_diagnostic_checklist__first_iteration():
 
 
 def test_print_diagnostic_checklist__second_iteration():
-    output_message = (
-        ExpectColumnValuesToEqualThree__SecondIteration().print_diagnostic_checklist()
-    )
+    output_message = ExpectColumnValuesToEqualThree__SecondIteration(
+        column="values"
```
cdkini, line 34: Feels iffy but all these diagnostic methods are instance methods - I have a ticket to try and make them classmethods so don't need to pass meaningless params but that's less of a priority atm

great_expectations/expectations/expectation.py
```diff
@@ -2373,6 +2369,12 @@ class BatchExpectation(Expectation, ABC):
     domain_type: ClassVar[MetricDomainTypes] = MetricDomainTypes.TABLE
     args_keys: ClassVar[Tuple[str, ...]] = ()
 
+    @pydantic.validator("mostly")
+    def validate_mostly(mostly: float) -> float:
+        if not (0 <= mostly <= 1):
+            raise ValueError("mostly must be a value between 0 and 1")
+        return mostly
```
cdkini: Originally just used `Field` with `ge` and `le` but that didn't seem to mesh well with inheritance?

This validator works with child classes but I saw test failures with the other pattern - worth investigating further.

great_expectations/expectations/expectation.py
```diff
@@ -2373,6 +2374,13 @@ class BatchExpectation(Expectation, ABC):
     domain_type: ClassVar[MetricDomainTypes] = MetricDomainTypes.TABLE
     args_keys: ClassVar[Tuple[str, ...]] = ()
 
+    @pydantic.validator("mostly")
+    @classmethod
+    def validate_mostly(cls, mostly: float) -> float:
+        if not (0 <= mostly <= 1):
+            raise ValueError("mostly must be a value between 0 and 1")
+        return mostly
```
cdkini: Weirdly enough, `pydantic.Field` and the in-built validation done through `gt` and `lt` didn't mesh well with inheritance

Kilo59: Have you tried using pydantic constrained field types? 
https://docs.pydantic.dev/1.10/usage/types/#constrained-types

great_expectations/expectations/expectation.py
Kilo59: Validators don't run by default for fields that aren't present.
https://docs.pydantic.dev/1.10/usage/validators/#validate-always
There are other scenarios which can cause vaildators not to run on subclasses. But that doesn't look like it applies in this case.
https://docs.pydantic.dev/1.10/usage/validators/#subclass-validators-and-each_item

------

IMO the best thing todo would be to not rely on validators and use a `confloat` or custom type.
https://docs.pydantic.dev/1.10/usage/types/#arguments-to-confloat
Using the `confloat` has the extra benefit of encoding the valid results to any generated jsonschema.

```python
from typing import Final, Type
from pydantic.v1 import BaseModel, confloat

ZeroToOne: Final[Type] = confloat(ge=0, le=1)

class Foo(BaseModel):
  foo_field: ZeroToOne

class Bar(BaseModel):
  bar_field: ZeroToOne
```

tests/expectations/test_expectation.py
Kilo59: `non-blocking`
pydantic validation errors happen on `__init__`. So we could have a custom `Expectation.__init__` that catches and rethrows `pydantic.ValidationError`s as `InvalidExpectationConfigurationError`.

But I also don't think we should be too concerned with breaking changes for the V1 release.

great_expectations/validator/validator.py
```diff
@@ -1122,7 +1122,9 @@ def _generate_metric_dependency_subgraphs_for_each_expectation_configuration(
                 raise InvalidExpectationConfigurationError(str(e))
 
             evaluated_config = copy.deepcopy(configuration)
-            evaluated_config.kwargs.update({"batch_id": self.active_batch_id})
+
```
Kilo59, line 1125: Is `evaluated_config: ExpectationConfiguration` a pydantic model now as well?

Kilo59 (approved):
I don't know the Expectation code that well so I would get a second opinion but LGTM.

great_expectations/validator/validator.py
cdkini: Nope we're going to get rid of that entirely

tests/expectations/test_expectation.py
cdkini: Yeah I think we're okay with `pydantic.ValidationError` and completely moving away from our custom error - good call though!

PR #8986, opened 2023-11-21
[FEATURE] ExpectationSuite 1.0 Add and Delete methods

joshua-stauffer:
This PR introduces the core 1.0 API for Expectation Suites. 

### three new methods:
- `ExpectationSuite.add(expectation: Expectation) -> Expectation`
- `ExpectationSuite.delete(expectation: Expectation) -> Expectation`
- `ExpectationSuite.save()`

### one new property:
- `ExpectationSuite.expectations: list[Expectation]`

## What's changed
The ExpectationSuite object can only be instantiated after a `DataContext` has been created, or a `RuntimeError` will be raised.
```Python
>>> ExpectationSuite(expectation_suite_name="test")
  File "/great_expectations/core/expectation_suite.py", line 133, in __init__
    self._store = project_manager.get_expectations_store()
  File "/great_expectations/data_context/data_context/context_factory.py", line 89, in get_expectations_store
    raise RuntimeError(
RuntimeError: This action requires an active DataContext. Please call `get_context()` first, then try your action again.
```

The `save` method persists the current state of the ExpectationSuite and any Expectations it contains to the context.
```python
>>> import great_expectations as gx
>>> context = gx.get_context()
>>> suite = ExpectationSuite(expectation_suite_name="test")
>>> context.list_expectation_suites()
[]
>>> suite.save()
>>> context.list_expectation_suites()
[ExpectationSuiteIdentifier::test]
```
The `add` and `delete` methods persist the updated state of the suite to the context.


joshua-stauffer:
depends on https://github.com/great-expectations/great_expectations/pull/8973

great_expectations/core/expectation_suite.py
```diff
@@ -133,6 +134,45 @@ def __init__(  # noqa: PLR0913
     def name(self) -> str:
         return self.expectation_suite_name
 
+    @property
+    def expectations(self) -> list[Expectation]:
```
jcampbell: Should uses be thinking of this list as ordered or as a set?

If a set, why does re-adding cause value error?

great_expectations/core/expectation_suite.py
```diff
@@ -133,6 +134,45 @@ def __init__(  # noqa: PLR0913
...
+
+    def add(self, expectation: Expectation) -> Expectation:
+        """Add an Expectation to the collection.
+
+        Raises:
+            ValueError: Suite already has an Expectation with the same configuration.
+        """
+        if any(
+            expectation.configuration == existing_config
+            for existing_config in self.expectation_configurations
```
cdkini, line 158: How do we perform equality checks with expectation configurations? I think it may be by memory address so I'd double check this logic

great_expectations/core/expectation_suite.py
joshua-stauffer: this change was a part of a previous PR, but your right, no reason for this to be ordered.

great_expectations/core/expectation_suite.py
joshua-stauffer: looks like it matches on attributes: https://github.com/great-expectations/great_expectations/blob/64c2a002578192569f12d4e06b2c60845fd3df38/great_expectations/core/expectation_configuration.py#L1311

great_expectations/core/expectation_suite.py
joshua-stauffer: I've updated the docstring to be clear that `ExpectationSuite` is set-like. I also debated renaming `delete` to `remove` to match the set API, but decided consistency with other GX objects is probably clearer.

great_expectations/core/expectation_suite.py
```diff
@@ -130,6 +128,10 @@ def __init__(  # noqa: PLR0913
         ensure_json_serializable(meta)
         self.meta = meta
 
+        from great_expectations import project_manager
+
+        self._store = project_manager.get_expectations_store()
```
cdkini, line 140: Just thinking out loud but are there any pros/cons to this import vs a granular one? I don't think so?

great_expectations/core/expectation_suite.py
```diff
@@ -141,6 +143,56 @@ def expectations(self) -> list[Expectation]:
...
+        ]
+        if len(updated_expectation_configs) != len(self.expectation_configurations) - 1:
+            raise KeyError("No matching expectation was found.")
+        self.expectation_configurations = updated_expectation_configs
+
+        try:
+            self.save()
+        except Exception as exc:
+            # rollback this change
+            # expectation suite is set-like so order of expectations doesn't matter
```
cdkini: If it is set-like, could we have the internal representation actually be a set instead of a list?

great_expectations/core/expectation_suite.py
```diff
@@ -141,6 +143,56 @@ def expectations(self) -> list[Expectation]:
...
+            raise exc
+
+        return expectation
+
+    def save(self) -> None:
+        """Save this ExpectationSuite."""
+        key = ExpectationSuiteIdentifier(
+            expectation_suite_name=self.expectation_suite_name
+        )
+        self._store.set(key=key, value=self)
```
cdkini: Hmm so this will need to be a GXCloudIdentifier if we're Cloud-backed (which is a big violation of encapsulation but just a general problem we seem to have).

I'm okay with this for now but as we're refactoring the stores, perhaps we shift the key generation logic down a layer so it is environment-agnostic

tests/core/test_expectation_suite.py
```diff
@@ -185,6 +193,137 @@ def __dict__(self):
         )
 
 
+class TestCRUDMethods:
```
tyler-hoffman, line 193: I typically see our tests as just top-level functions; why introduce the class? I don't have a preference. just curious.

tests/core/test_expectation_suite.py
```diff
@@ -185,6 +193,137 @@ def __dict__(self):
...
+    """Tests related to the 1.0 CRUD API."""
+
+    expectation_suite_name = "test-suite"
+    store_key = ExpectationSuiteIdentifier(
+        expectation_suite_name=expectation_suite_name
+    )
+
+    @pytest.fixture
+    def expectation_params(self):
+        return {"column": "a", "value_set": [1, 2, 3], "result_format": "BASIC"}
```
tyler-hoffman: why not make the fixture the expectation itself? If we change the signature of `ExpectColumnValuesToBeInSet`, it'd be nice for the type system to catch any missing values before we run the test.

great_expectations/core/expectation_suite.py
```diff
@@ -141,6 +143,56 @@ def expectations(self) -> list[Expectation]:
...
+
+        return expectation
+
+    def delete(self, expectation: Expectation) -> Expectation:
+        """Delete an Expectation from the collection.
+
+        Raises:
+            ValueError: Expectation not found in suite.
+        """
+        updated_expectation_configs = [
```
tyler-hoffman: nit: the name here initially made me think that the expectation configs (rather than the list) were updated. I'd probably go with `remain_expectation_configs`

tyler-hoffman (approved):
Left some minor thoughts (the `expectation_params` fixture being the least minor, though still minor). But this looks great!

great_expectations/core/expectation_suite.py
joshua-stauffer: i decided against this approach since items in a set need to be hashable - we would have to serialize the expectations (or expectation configurations), and it just seems more convenient to store them in the state we want to work with them.

great_expectations/core/expectation_suite.py
joshua-stauffer: oh gross, thanks for raising this, i wasn't aware

tests/core/test_expectation_suite.py
joshua-stauffer: i generally prefer top-level functions, but using the class is a good way to namespace. It lets the test names be shorter, which is useful, but in this case it helps make it clear that all of these tests are 1.0, and any other test in the file is legacy.

tests/core/test_expectation_suite.py
joshua-stauffer: great suggestion, updated in [e064aa7](https://github.com/great-expectations/great_expectations/pull/8986/commits/e064aa749341744d87b48c02cef3f873cc157c53)

great_expectations/core/expectation_suite.py
joshua-stauffer: sure, updated in [0f46abf](https://github.com/great-expectations/great_expectations/pull/8986/commits/0f46abf11088632a96dcf2ad12bef1b1a6d64d66)

great_expectations/core/expectation_suite.py
joshua-stauffer: my thought was that a free-wheeling function called `get_expectations_store` is more likely to be abused than a method namespaced in another object, but i think that's a fairly subjective benefit 😄

great_expectations/core/expectation_suite.py
joshua-stauffer: updated in [6838709](https://github.com/great-expectations/great_expectations/pull/8986/commits/6838709a81b15f7db57b22a8409653f1d8838b0e), thanks!

great_expectations/data_context/store/store.py
```diff
@@ -251,13 +251,17 @@ def _update(self, key: DataContextKey, value: Any, **kwargs) -> None:
...
+    ) -> None | GXCloudIdentifier:
         """
         Conditionally calls `add` or `update` based on the presence of the given key.
         """
         return self._add_or_update(key=key, value=value, **kwargs)
 
-    def _add_or_update(self, key: DataContextKey, value: Any, **kwargs) -> None:
+    def _add_or_update(
+        self, key: DataContextKey, value: Any, **kwargs
+    ) -> None | GXCloudIdentifier:
```
joshua-stauffer, line 264: no functional change here, just updating to accurate type

PR #9018, opened 2023-11-30
[MAINTENANCE] Add batch_configs to context schema

tyler-hoffman:
Adds batch_configs to `DataContextConfigSchema`. This follows the pattern used for `fluent_datasources` - it just takes in the dicts without further marshmallow-ifying it.


great_expectations/data_context/types/base.py
```diff
@@ -2306,6 +2313,7 @@ class DataContextConfig(BaseYamlConfig):
 
     def __init__(  # noqa: C901, PLR0912, PLR0913, PLR0915
         self,
+        batch_configs: Optional[Dict] = None,
```
cdkini, line 2316: Do we want to strictly type the `dict` here?

great_expectations/data_context/types/base.py
```diff
@@ -1517,6 +1517,12 @@ def to_json_dict(self) -> Dict[str, JSONValues]:
 
 
 class DataContextConfigSchema(Schema):
+    batch_configs = fields.Dict(
+        keys=fields.Str(),
+        required=False,
+        allow_none=True,
+        load_only=True,
```
cdkini, line 1524: What is `load_only`?

great_expectations/data_context/types/base.py
tyler-hoffman: I'm on the fence and can be swayed. My thought is that ideally this would be using pydantic for validation (which will end up happening later on when we load the model itself. So this feels like a least-effort-possible intermediary step while we are using marshmallow. This is what we do for fluent datasources, and I was mostly leaning on that as prior art.

great_expectations/data_context/types/base.py
tyler-hoffman: It means that it is used just for deserialization (load), but it doesn't serialize out. This is what we do for fluent datasources as well, and I had assumed there was a good reason for it that we'd uncover when we get into deserialization workflows. Maybe it makes sense to just remove it for now?

cdkini: I'm okay with it I'll defer to you! Wasn't too sure so I thought I'd ask

great_expectations/data_context/types/base.py
cdkini: Yeah I'm okay with this

cdkini (approved):

PR #9034, opened 2023-12-01
[BUGFIX] Databricks shared compute fix

jcampbell:
Calls to `spark.sparkContext.getConf()` fail on a "[high concurrency cluster with credential passthrough enabled](https://community.databricks.com/t5/data-engineering/py4j-security-py4jsecurityexception-constructor-public-org/td-p/27862)" due to security concerns. Users can use a different type of cluster, or set `spark.databricks.pyspark.enablePy4JSecurity` to `false`, but we don't need to call this method anyway unless we are using a local Spark installation.

This PR also deprecates the use of the `force_reuse_spark_context` parameter.


jcampbell:
@NathanFarmer this looks excellent to me. I created the PR so can't be the reviewer of record, but I asked @rachhouse to do that.

great_expectations/core/util.py
```diff
@@ -145,14 +142,14 @@ def in_jupyter_notebook():
...
-def in_databricks() -> bool:
+def in_databricks(spark_session: pyspark.SparkSession) -> bool:
     """
     Tests whether we are in a Databricks environment.
 
     Returns:
         bool
     """
-    return "DATABRICKS_RUNTIME_VERSION" in os.environ  # noqa: TID251
+    return spark_session.sparkContext.appName == "Databricks Shell"
```
rachhouse: Would it be safer to match here on containing `Databricks` instead of an exact match on `Databricks Shell`? Do we have confidence this naming won't change?

great_expectations/core/util.py
```diff
@@ -782,150 +779,63 @@ def sniff_s3_compression(s3_url: S3Url) -> Union[str, None]:
...
+                )
+        else:
+            # in a local pyspark-shell the context config cannot be updated
+            # unless you stop the Spark context and re-recreate it
+            retrieved_spark_config: pyspark.SparkConf = (
+                spark_session.sparkContext.getConf()
+            )
+            stopped = False
+            for key, value in spark_config.items():
+                if retrieved_spark_config.get(key) != value:
```
rachhouse: We check whether the current spark config differs from the input spark config, and I see the current spark session is stopped then restarted if there are differing config values. Are the input `spark_config` values (the one supplied `get_or_create_spark_session`) supposed get set for the restarted session? If so, where does this happen?

NathanFarmer: The input `spark_config` is passed to `builder.config` already, but for local environments it is ignored if there is an existing Spark context. We don't need to pass the config to `builder` again, we just need to stop and restart with the same `builder`.

rachhouse: Oh, wait. I see it now, I was focused on the new code earlier. Please disregard. :)
```
        for k, v in spark_config.items():
            if k != "spark.app.name":
                builder.config(k, v)
```

great_expectations/datasource/fluent/spark_datasource.py
```diff
@@ -70,6 +71,19 @@ class _SparkDatasource(Datasource):
...
+    @pydantic.validator("force_reuse_spark_context")
+    @classmethod
+    def _force_reuse_spark_context_deprecation_warning(cls, v: bool) -> bool:
+        if v is not None:
+            warnings.warn(
+                "force_reuse_spark_context is deprecated and will be removed in version 1.0. "
+                "In environments that allow it, the existing Spark context will be reused, adding the "
+                "spark_config options that have been passed. If the Spark context cannot be updated with "
+                "the spark_config, the context will be stopped and restarted with the new spark_config.",
+                category=RuntimeWarning,
```
Kilo59:
```suggestion
                category=DeprecationWarning,
```
https://docs.python.org/3/library/exceptions.html#DeprecationWarning

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -188,28 +198,28 @@ class SparkDFExecutionEngine(ExecutionEngine):
...
         azure_options: dict = kwargs.pop("azure_options", {})
         self._azure_options = azure_options
 
+        if force_reuse_spark_context is not None:
+            warnings.warn(
+                "force_reuse_spark_context is deprecated and will be removed in version 1.0. "
+                "In environments that allow it, the existing Spark context will be reused, adding the "
+                "spark_config options that have been passed. If the Spark context cannot be updated with "
+                "the spark_config, the context will be stopped and restarted with the new spark_config.",
+                category=RuntimeWarning,
```
Kilo59:
```suggestion
                category=DeprectationWarning,
```
https://docs.python.org/3/library/exceptions.html#DeprecationWarning

Kilo59 (approved):
Some of the `RuntimeWarnings` should be `DeprecationWarning`.
Other than that it looks good.
I don't know spark very well though.

great_expectations/core/util.py
NathanFarmer: Sounds better to me!

great_expectations/core/util.py
```diff
@@ -782,148 +769,124 @@ def sniff_s3_compression(s3_url: S3Url) -> Union[str, None]:
...
+        bool
     """
+    app_name: str
+    try:
+        app_name = spark_session.sparkContext.appName  # type: ignore[assignment]  # handled by try/except
+    except pyspark.PySparkNotImplementedError:
+        app_name = ""
+
+    updatable = "databricks" in app_name.lower()
+    return updatable
```
NathanFarmer: It turns out we can't update a Spark/Databricks Connect session remotely, so only Databricks notebooks have updatable configs without stopping the Spark context.

great_expectations/core/util.py
```diff
@@ -782,148 +769,124 @@ def sniff_s3_compression(s3_url: S3Url) -> Union[str, None]:
     return _SUFFIX_TO_PD_KWARG.get(s3_url.suffix) if s3_url.suffix else None
 
 
-# noinspection PyPep8Naming
-def get_or_create_spark_application(
```
Kilo59, line 786: @NathanFarmer 
Is the utility not user facing? Wouldn't it be a breaking change?

I would suggest keeping the function but having it raise a deprecation warning and call the renamed method.
Although I'm not sure if the usage actually makes sense but you would know. 🤷 
```python
def get_or_create_spark_application():
   warnings.warn("Don't do it", category=DeprecationWarning)
   return get_or_create_spark_session()
```

great_expectations/core/util.py
jcampbell: Couldn't a user have "databricks" in their app name even if it's a connect app?

NathanFarmer: If it's a connect app, the `PySparkNotImplementedError` gets raised and the string is empty. I will add comments explaining that.

great_expectations/core/util.py
NathanFarmer: It shouldn't be user facing, but I guess it probably was in an example notebook or something. I made the updates you suggested.

great_expectations/compatibility/databricks.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import annotations
```
jcampbell: Can we remove this now that we're not testing using this method?

great_expectations/compatibility/pyspark.py
```diff
@@ -67,11 +80,25 @@
     Window = SPARK_NOT_IMPORTED  # type: ignore[assignment,misc]
 
 try:
-    from pyspark.sql.readwriter import DataFrameReader
+    from pyspark.sql.readwriter import DataFrameReader as _DataFrameReader
```
jcampbell: Are these changes exclusively to improve typing?

great_expectations/core/util.py
```diff
@@ -782,148 +771,150 @@ def sniff_s3_compression(s3_url: S3Url) -> Union[str, None]:
...
-    spark_session: Optional[pyspark.SparkSession] = get_or_create_spark_session(
-        spark_config=spark_config
+    spark_session_type: type[pyspark.SparkSession | pyspark.SparkConnectSession]
+    spark_session: pyspark.SparkSession
+    try:
+        spark_session_type = pyspark.SparkSession
+        spark_session = spark_session_type.builder.getOrCreate()
+    except RuntimeError:
+        spark_session_type = pyspark.SparkConnectSession
+        spark_session = spark_session_type.builder.getOrCreate()
```
jcampbell: Can this raise an exception?

great_expectations/core/util.py
```diff
@@ -782,148 +771,150 @@ def sniff_s3_compression(s3_url: S3Url) -> Union[str, None]:
...
+    if allow_restart:
+        try:
+            retrieved_spark_config: pyspark.SparkConf = (
+                spark_session.sparkContext.getConf()
+            )
+            for key, value in spark_config.items():
+                if retrieved_spark_config.get(key) != value:
+                    spark_session.stop()
+                    break
+        except pyspark.PySparkNotImplementedError:
```
jcampbell: What about the security error raised by getConf if we're in a shared compute cluster? Are we sure that we've always caught that with allow_restart? Should we catch the security condition?

great_expectations/compatibility/databricks.py
NathanFarmer: Looks like the answer is yes. I was actually planning to remove this entire module now that we don't reference it anywhere.

great_expectations/compatibility/pyspark.py
NathanFarmer: They were required when I was using `DatabricksConnectSession`, but now it seems they are not so I have reverted this change.

great_expectations/core/util.py
NathanFarmer: It could. If it does I think we should raise the initial `RuntimeError` triggered on line 819. ✅

great_expectations/core/util.py
NathanFarmer: This is outdated now. We are handling the security error in the revisions.

tests/conftest.py
```diff
@@ -2829,7 +2815,7 @@ def data_context_parameterized_expectation_suite(tmp_path_factory):
             )
         ),
     )
-    return get_context(context_root_dir=context_path)
```
NathanFarmer, line 2832: This fails locally unless you turn off cloud mode. Passing `mode="file"` was causing an error in conjunction with `context_root_dir`.

tests/datasource/fluent/integration/conftest.py
```diff
@@ -136,8 +143,12 @@ def sql_data(
 
 
 def spark_filesystem_datasource(
+    test_backends,
     context: AbstractDataContext,
 ) -> SparkFilesystemDatasource:
+    if "SparkDFDataset" not in test_backends:
+        pytest.skip("No spark backend selected.")
```
NathanFarmer, line 151: Moving Spark connection logic into `test_connection` for the datasources meant that we needed to skip the tests earlier when spark wasn't installed.

tests/datasource/fluent/test_spark_datasource.py
```diff
@@ -118,24 +117,22 @@ def test_spark_config_passed_to_execution_engine(
     persist,
     spark_session,
 ):
-    spark_config: SparkConfig | None = {
-        "spark.sql.catalogImplementation": "in-memory",
+    spark_config: SparkConfig = {
         "spark.app.name": "gx_spark_fluent_datasource_test",
         "spark.default.parallelism": 4,
-        "spark.driver.memory": "16g",
-        "spark.executor.memory": 471859200,
```
NathanFarmer, line 126: Warnings are now raised for these config options because they can't be changed unless you restart the Spark Session. Options that raise warnings were either removed from tests or `with pytest.warns` was added.

great_expectations/datasource/fluent/spark_datasource.py
```diff
@@ -99,17 +161,17 @@ def test_connection(self, test_assets: bool = True) -> None:
         Raises:
             TestConnectionError: If the connection test fails.
         """
-        raise NotImplementedError(
-            """One needs to implement "test_connection" on a _SparkDatasource subclass."""
-        )
+        try:
+            self.get_spark()
+        except ConnectionError as e:
+            raise TestConnectionError(e) from e
```
Kilo59: TIL (or maybe Today I Remembered) that we weren't testing spark connections on the standard SparkDatasource.
😬

Kilo59 (approved):
Re-approving

This looks like it was a pain!

PR #9036, opened 2023-12-02
[BUGFIX] validator head query limit

dctalbot:
- There is currently an issue in which a limit is sometimes missing from the `validator.head()` sqlalchemy query.
- An entire table is loaded into a frame, and a limit is applied to the frame via `chunksize`. 
- Instead, we should apply the limit to the underlying query. That is what this PR does.
- example of new query:
```
SELECT *
	FROM (SELECT *
	FROM foo
	WHERE true) AS anon_1
	 LIMIT 5
```


great_expectations/compatibility/sqlalchemy_and_pandas.py
```diff
@@ -82,48 +82,3 @@ def pandas_read_sql(sql, con, **kwargs) -> pd.DataFrame | Iterator[pd.DataFrame]
             sql = sa.select(sa.text("*")).select_from(sql)
         return_value = pd.read_sql(sql=sql, con=con, **kwargs)
     return return_value
-
-
-def pandas_read_sql_query(
```
dctalbot, line 87: redundant, removed in favor of `pandas_read_sql`.

great_expectations/compatibility/sqlalchemy_and_pandas.py
```diff
@@ -82,48 +82,3 @@ def pandas_read_sql(sql, con, **kwargs) -> pd.DataFrame | Iterator[pd.DataFrame]
...
-            between pandas versions.
-    Returns:
-        dataframe
-    """
-    if (
-        sa
-        and is_version_greater_or_equal(sa.__version__, "2.0.0")
-        and is_version_less_than(pd.__version__, "2.0.0")
-    ):
-        warn_pandas_less_than_2_0_and_sqlalchemy_greater_than_or_equal_2_0()
```
NathanFarmer, line 115: The `pip` resolver probably doesn't allow this to happen anymore, but we probably should keep this warning logic somewhere just in case.

NathanFarmer (changes requested):
One requested change to include that warning logic, but other than that this looks like an amazing simplification!

great_expectations/compatibility/sqlalchemy_and_pandas.py
dctalbot: `pandas_read_sql` has the same logic, so I think it's fine to remove here since this function was deleted. `pandas_read_sql` provides equivalent functionality and is available going forward.

NathanFarmer: I did not know that!

NathanFarmer (approved):

PR #9038, opened 2023-12-04
[MAINTENANCE] pydantic fields instead of defaults dict

tyler-hoffman:
Replace use of `default_kwarg_values` with private method to inspect pydantic field defaults.


great_expectations/expectations/core/expect_column_values_to_be_null.py
```diff
@@ -220,9 +220,7 @@ def _validate(
         result_format = self.get_result_format(
             configuration=configuration, runtime_configuration=runtime_configuration
         )
-        mostly = self.get_success_kwargs().get(
-            "mostly", self.default_kwarg_values.get("mostly")
-        )
+        mostly = self.get_success_kwargs().get("mostly")
```
cdkini, line 223: Should this also use `self._get_default_value` to match the pattern of the other changes?

great_expectations/expectations/core/expect_column_values_to_not_be_null.py
```diff
@@ -268,9 +268,7 @@ def _validate(
         result_format = self.get_result_format(
             configuration=configuration, runtime_configuration=runtime_configuration
         )
-        mostly = self.get_success_kwargs().get(
-            "mostly", self.default_kwarg_values.get("mostly")
-        )
+        mostly = self.get_success_kwargs().get("mostly")
```
cdkini, line 271: Same note here

great_expectations/expectations/expectation.py
```diff
@@ -287,14 +286,6 @@ def __new__(cls, clsname, bases, attrs):
...
         # noinspection PyUnresolvedReferences
         newclass._register_renderer_functions()
-        default_kwarg_values = {}
-        for base in reversed(bases):
-            default_kwargs = getattr(base, "default_kwarg_values", {})
-            default_kwarg_values = nested_update(default_kwarg_values, default_kwargs)
-
-        newclass.default_kwarg_values = nested_update(
-            default_kwarg_values, attrs.get("default_kwarg_values", {})
-        )
```
cdkini, line 294: ❤️

great_expectations/expectations/expectation.py
```diff
@@ -1166,11 +1151,22 @@ def get_validation_dependencies(
             metric_configurations={}, result_format=result_format
         )
 
+    def _get_default_value(self, key: str) -> Any:
+        """Get default value for the model if it is not required.
+        NOTE: For reasons I'm unclear on, we can get back either a FieldInfo or a ModelField.
```
cdkini: You're still seeing this? I would this `self.__fields__` would give us back a defined structure

great_expectations/expectations/expectation.py
```diff
@@ -1166,11 +1151,22 @@ def get_validation_dependencies(
...
+        """
+        field = self.__fields__.get(key)
+
+        if isinstance(field, pydantic.fields.ModelField):
+            return field.default if not field.required else None
+        else:
+            return None
```
cdkini: Will we ever call this code with a value that doesn't exist? I'd rather we fail than possibly pass down `None` when we shouldn't

great_expectations/expectations/expectation.py
```diff
@@ -2501,22 +2487,18 @@ class QueryExpectation(BatchExpectation, ABC):
...
-        "catch_exceptions": False,
-        "meta": None,
-        "row_condition": None,
-        "condition_parser": None,
-    }
+    result_format: Union[ResultFormat, ResultFormatDict] = ResultFormat.BASIC
+    catch_exceptions: bool = False
+    meta: Optional[dict] = None
+    row_condition: Optional[str] = None
+    condition_parser: Optional[str] = None
```
cdkini: These are defined on the parent `BatchExpectation` right? I don't think we need to duplicate

great_expectations/expectations/expectation.py
```diff
@@ -2863,9 +2836,7 @@ def _validate(
             success = _mostly_success(
                 nonnull_count,
                 unexpected_count,
-                self.get_success_kwargs().get(
-                    "mostly", self.default_kwarg_values.get("mostly")
-                ),
+                self.get_success_kwargs()["mostly"],
```
cdkini, line 2795: Do we need your helper here?

tests/expectations/core/test_expect_column_values_to_be_in_set.py
```diff
@@ -14,9 +14,6 @@
 # <snippet name="tests/expectations/core/test_expect_column_values_to_be_in_set.py ExpectColumnValuesToBeTwoLetterCountryCode_class_def">
 class ExpectColumnValuesToBeTwoLetterCountryCode(ExpectColumnValuesToBeInSet):
     value_set: List[str] = ["FR", "DE", "CH", "ES", "IT", "BE", "NL", "PL"]
-    default_kwarg_values = {
-        "value_set": ["FR", "DE", "CH", "ES", "IT", "BE", "NL", "PL"],
-    }
```
cdkini, line 19: I believe I fixed one other test like this - we should see it in CI but I'll take a look to try and find it

great_expectations/expectations/core/expect_column_values_to_be_null.py
tyler-hoffman: Nope. AFAIK the previous pattern was redundant; `get_success_kwargs` calls `_get_default_value` (and previously reached into the dict)

great_expectations/expectations/expectation.py
tyler-hoffman: oops! this was old. Removing shortly.

great_expectations/expectations/expectation.py
tyler-hoffman: We definitely shouldn't, but that case was implicitly handled like this before, and I wanted to avoid behavior changes. If you think I should throw a `raise` in here, I'm happy to do it if you think that's better.

great_expectations/expectations/expectation.py
tyler-hoffman: `get_success_kwargs` uses the helper!

tests/expectations/core/test_expect_column_values_to_be_in_set.py
tyler-hoffman: Was it `ExpectColumnValuesAsStringToBePositiveInteger`? I just removed it from there too.

great_expectations/expectations/expectation.py
cdkini: Hmm I agree based on that reasoning but maybe we log before returning `None` just to provide context around the miss if it happens?

cdkini (approved):

PR #9051, opened 2023-12-05
[BUGFIX] re-use sa.engine.Engine object

Kilo59:
Changes the behavior of `SQLDatasource.get_execution_engine()` so that it no longer re-creates the `sqlalchemy.engine.Engine` object and instead re-uses the one used by the Datasource for `.test_connection()`.

This solves 2 problems...
1. Reduces waste and extra connections caused by having 2 different `sqlalchemy.engine.Engine` objects at runtime.
2. Removes the problem of having slightly different configured `sqlalchemy.engine.Engine` objects.
    * This is the source of the bug that requires a new OSS release. https://github.com/great-expectations/great_expectations/pull/9055

## SQLite

~Created an override `_create_engine` method so that we can set a `StaticPool`.
Tests start failing without this.~

SQLite retains the old behavior that creates to separate `sqlalchemy.engine.Engine` objects. This seems like the safer approach (or at least doesn't cause any additional harm).

## Snowflake Datasource

No change was made to Snowflake, but the impetus for this change was because `gx` `SQLAlchemyExecutionEngine` version of the `sqlalchemy.engine.Engine` was erroring on instantiation when the `connection_string` was a dictionary.
Now that the datasource's `sqlalchemy.engine.Engine` is re-used for the `SQLAlchemyExecutionEngine` this is no longer a problem.


great_expectations/datasource/fluent/sqlite_datasource.py
```diff
@@ -209,6 +214,25 @@ class SqliteDatasource(SQLDatasource):
...
+            config_provider=self._config_provider,
+        )
+        _check_config_substitutions_needed(
+            self, model_dict, raise_warning_if_provider_not_present=True
+        )
+        # the connection_string has had config substitutions applied
+        connection_string = model_dict.pop("connection_string")
+        kwargs = model_dict.pop("kwargs", {})
+        # if needed a user could set a different `_poolclass` on the datasource.
+        return sa.create_engine(connection_string, poolclass=self._poolclass, **kwargs)
```
Kilo59: https://github.com/great-expectations/great_expectations/blob/92e6fd2128132e1bb27b114dc0c6231c518ddd6b/great_expectations/execution_engine/sqlalchemy_execution_engine.py#L281-L282

tests/datasource/fluent/test_sql_datasources.py
```diff
@@ -250,5 +254,19 @@ def test_specific_datasource_warnings(
...
+@pytest.mark.parametrize("datasource_type", [SQLDatasource, SqliteDatasource])
+def test_get_engine_is_used_for_sqlalchemy_execution_engine(
+    datasource_type: type[SQLDatasource],
+):
+    """
+    This test ensures that the engine returned by get_engine() is used for the gx `SQLAlchemyExecutionEngine`.
+    """
+    datasource = datasource_type(name="my_datasource", connection_string="sqlite:///")
+
+    assert datasource.get_engine() is datasource.get_execution_engine().engine
```
Kilo59: This test does fail on `develop`

tests/datasource/fluent/test_sqlite_datasource.py
Kilo59, line 1: `test_sqlite_specific_splitter` is now failing 🤔 

```
E       sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "table": syntax error
E       [SQL: SELECT distinct(strftime(?, pickup_datetime)) AS distinct_1 
E       FROM (SELECT  * from table) AS anon_1]
E       [parameters: ('%Y-%m-%d',)]
E       (Background on this error at: https://sqlalche.me/e/14/e3q8)
```

tests/datasource/fluent/test_sqlite_datasource.py
```diff
@@ -119,6 +119,8 @@ def create_sqlite_source() -> (
     return _create_sqlite_source
 
 
+# TODO: figure this out before merging
+@pytest.mark.xfail(reason="why are these tests are now failing???")
```
Kilo59: `test_sqlite_specific_splitter` is now failing 🤔 

No other SQL related tests seem to be failing.

```
E       sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "table": syntax error
E       [SQL: SELECT distinct(strftime(?, pickup_datetime)) AS distinct_1 
E       FROM (SELECT  * from table) AS anon_1]
E       [parameters: ('%Y-%m-%d',)]
E       (Background on this error at: https://sqlalche.me/e/14/e3q8)
```

Kilo59:
Opened a more targeted fix while the SQLite issues are worked out.
- #9055

tests/datasource/fluent/test_sqlite_datasource.py
NathanFarmer: The only thing I can think is that SQLite dbs can be in-memory and maybe it was tossed out somehow? I would think from your changes it would be the opposite - we are preserving in-memory dbs by not closing the engine connection.

great_expectations/datasource/fluent/sqlite_datasource.py
NathanFarmer: TIL that this was an option.

tests/datasource/fluent/test_sqlite_datasource.py
NathanFarmer: Maybe SQL Alchemy is not installed for the sqlite tests (it's included in base python). If not, then it looks like `_poolclass = None`.

NathanFarmer (approved):
Glad somebody got around to fixing this!

tests/datasource/fluent/test_sql_datasources.py
Shinnnyshinshin: Just thinking out-loud: Do we need this to always be true? I think a consequence would have to be with temporary tables. If it is a new engine then we would lose access to the temporary tables created by the old one

Kilo59: The `datasource.get_engine()` was only being used to test the connection, it wasn't ever used to write/read to the database as part of a checkpoint run or metrics calculation.
Now the same `SQLAlchemyEngine` will be used for both.

tests/datasource/fluent/test_sqlite_datasource.py
Kilo59: I'm not sure exactly what was happening.
But I've reverted my changes, and made it so that SQLite retains the old behavior of creating 2 `Engine` objects. 
1 for testing the connection
1 for everything else.

That somehow fixed it.

Kilo59:
@NathanFarmer seeing the same issue with Postgres, so it's not just a SQlite problem.
I'm going to have to come back to this. I don't have the bandwidth to investigate it at the moment. 

```
FAILED tests/datasource/fluent/data_asset/test_sql_asset.py::test_get_batch_list_from_batch_request__sort_descending - sqlalchemy.exc.ProgrammingError:
(psycopg2.errors.UndefinedTable) relation "my_table" does not exist
LINE 2: FROM my_table
             ^

[SQL: SELECT distinct(EXTRACT(year FROM year)) AS concat_distinct_values, CAST(EXTRACT(year FROM year) AS INTEGER) AS year 
FROM my_table]
(Background on this error at: https://sqlalche.me/e/20/f405)
```

NathanFarmer:
> @NathanFarmer seeing the same issue with Postgres, so it's not just a SQlite problem. I'm going to have to come back to this. I don't have the bandwidth to investigate it at the moment.
> 
> ```
> FAILED tests/datasource/fluent/data_asset/test_sql_asset.py::test_get_batch_list_from_batch_request__sort_descending - sqlalchemy.exc.ProgrammingError:
> (psycopg2.errors.UndefinedTable) relation "my_table" does not exist
> LINE 2: FROM my_table
>              ^
> 
> [SQL: SELECT distinct(EXTRACT(year FROM year)) AS concat_distinct_values, CAST(EXTRACT(year FROM year) AS INTEGER) AS year 
> FROM my_table]
> (Background on this error at: https://sqlalche.me/e/20/f405)
> ```

It felt like you had found a hack. Maybe just revert the entire PR.

PR #9053, opened 2023-12-06
[FEATURE] Expectation.save()

joshua-stauffer:


great_expectations/core/expectation_suite.py
```diff
@@ -867,6 +906,15 @@ def _build_expectation(
...
             ) from e
 
+    def _add_save_callback_to_expectation(
+        self,
+        expectation: Expectation,
+        expectation_configuration: ExpectationConfiguration,
+    ) -> Expectation:
+        save_callback = partial(self.update_by_identity, id(expectation_configuration))
+        expectation.register_save_callback(callback=save_callback)
+        return expectation
```
tyler-hoffman, line 916: super nit: the name of this method does not lead me to think I'd get an expectation back. We already have a handle on the expectation; I'd probably just not return anything (unless we are trying for a more fluent-ish flow. Actually looking at the overall flow, I might just include those 2 lines right in `_build_expectation`.

great_expectations/expectations/expectation.py
```diff
@@ -2281,6 +2283,27 @@ def _get_final_maturity_level(
...
+
+    @public_api
+    def save(self):
+        """Save this Expectation."""
+        if not self._save_callback:
+            raise RuntimeError(
+                "Expectation must be added to an ExpectationSuite before it can be saved."
+            )
+        expectation = self._save_callback(self)
+        expectation.transfer_save_callback(expectation=self)
```
tyler-hoffman: I'm not following what transferring the callback to self does

jcampbell: This seems like an odd method to be public. Why not have the transfer happen as part of `_save_callback`?

tests/core/test_expectation_suite.py
```diff
@@ -362,6 +362,210 @@ def test_save_success(self):
...
+        )
+
+        expectation = suite.update_by_identity(
+            identity=id(old_expectation_configuration), expectation=expectation
+        )
+
+        assert len(suite.expectations) == 1, "Method must replace expectation"
+        assert suite.expectations[0].column == expectation.column
+        # expect that the data context is kept in sync
+        context.expectations_store.add_or_update.assert_called_once_with(
```
tyler-hoffman: nit: feels like this check is sufficiently covered by the next test case.

tests/core/test_expectation_suite.py
```diff
@@ -362,6 +362,210 @@ def test_save_success(self):
...
+        # expectation = suite.expectations[0]
+        # expectation.column = "b"
+        expectation = ExpectColumnValuesToBeInSet(
+            column="b",
+            value_set=[1, 2, 3],
+            result_format="BASIC",
+        )
+
+        with pytest.raises(
+            KeyError, match="Cannot update Expectation because it was not found."
```
tyler-hoffman, line 444: Just flagging here for anyone interested that we discussed raising a more descriptive error.

tyler-hoffman (commented):
This is interesting. When I see the stuff about transferring the callback it feels reminiscent of rust's borrowing mechanism.

Not blocking or accepting right now just because I think we're still discussing the approach

great_expectations/core/expectation_suite.py
joshua-stauffer: i initially just had them inline `_build_expectation` then pulled them out for clarity. thanks for the suggestion - will look to incorporate it.

great_expectations/expectations/expectation.py
joshua-stauffer: yeah, this is a bit opaque, i had to think a minute to remind myself why this was necessary, which is never a good sign 😄 we'd like to be able to continue iterating and calling save on this expectation object, but once we call `save` the `ExpectationConfiguration` we were pinned to is no longer present in the suite. The `save_callback` function provided by the suite returns the newly created expectation, which includes a new `save_callback`. This method transfers that new callback to our current object, so that the user can continue editing in place.

joshua-stauffer: you know, not only are you right, but that's already happening 🙃 i was thinking about how to allow a user to continue editing and saving the same expectation, and solved it twice in two places. Better to just allow the suite to register the new callback. Thanks!

joshua-stauffer: ah, after digesting [James's comment here](https://github.com/great-expectations/great_expectations/pull/9053#discussion_r1417953907) I realize that this method is unnecessary.

tests/core/test_expectation_suite.py
joshua-stauffer: thanks! that was just copy pasta, totally is the responsibility of the next test 🙇

great_expectations/core/expectation_suite.py
joshua-stauffer: this is actually called a couple other places, since we need to update the ref on CRUD methods. will leave as is for now, but there might be a clearer pattern in the future

joshua-stauffer:
implementing save via IDs instead

PR #9055, opened 2023-12-06
[BUGFIX] V1 - Fix Snowflake error when using alternative connection details format.

Kilo59:
Changes the behavior of `SnowflakeDatasource.get_execution_engine()` so that it no longer re-creates the `sqlalchemy.engine.Engine` object and instead re-uses the one used by the Datasource for `.test_connection()`.

This solves 2 problems...

1. Reduces waste and extra connections caused by having 2 different `sqlalchemy.engine.Engine` objects at runtime.
2. Removes the problem of having slightly different configured `sqlalchemy.engine.Engine` objects.
   * This is the source of the bug that requires a new OSS release.

This is a more minimal version of the following bugfix. Issues with SQLite are forcing the more minimal PR (9055).
 - #9051 

Follow-up to 
- https://github.com/great-expectations/great_expectations/pull/8922

## Other notes
For Snowflake this also means that when using the `authenticator=externalbrowser` query param, the browser will only need to auth once instead of 2wice for every connection.

### DIFF
--- a/tests/datasource/fluent/test_snowflake_datasource.py
+++ b/tests/datasource/fluent/test_snowflake_datasource.py
@@ -12,6 +12,7 @@ from great_expectations.datasource.fluent.snowflake_datasource import (
     SnowflakeDatasource,
     SnowflakeDsn,
 )
+from great_expectations.execution_engine import SqlAlchemyExecutionEngine
 
 
 @pytest.fixture
@@ -70,7 +71,7 @@ def test_valid_config(
     assert isinstance(sql_engine, sa.engine.Engine)
 
     exec_engine = my_sf_ds_1.get_execution_engine()
-    assert sql_engine is exec_engine.engine
+    assert isinstance(exec_engine, SqlAlchemyExecutionEngine)
 
 
 @pytest.mark.unit
--- a/tests/datasource/fluent/test_snowflake_datasource.py
+++ b/tests/datasource/fluent/test_snowflake_datasource.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import pytest
 import sqlalchemy as sa
+from pytest import param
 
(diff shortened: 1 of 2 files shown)


tests/datasource/fluent/test_snowflake_datasource.py
```diff
@@ -54,6 +70,9 @@ def test_valid_config(
     sql_engine = my_sf_ds_1.get_engine()
     assert isinstance(sql_engine, sa.engine.Engine)
 
+    exec_engine = my_sf_ds_1.get_execution_engine()
+    assert isinstance(exec_engine, SqlAlchemyExecutionEngine)
```
Kilo59, line 74: @DrewHoo 
This added line catches the errors we've seen in `0.18.3`

NathanFarmer (approved):
Why not just do this on `SQLDatasource` like in your other PR and override SQLite with something that works?

Kilo59:
Going to move the `get_execution_engine()` method to `SQLDatasource` and drop it from `Snowflake` in a followup.
Need to finish dealing with SQLite issues and move some `unit-tests` to different makers before I can make this change.

tests/datasource/fluent/test_snowflake_datasource.py
```diff
@@ -23,23 +25,37 @@ def seed_env_vars(monkeypatch: pytest.MonkeyPatch) -> None:
...
+                    "password": "${MY_PASSWORD}",
+                    "account": "my_account",
+                }
+            },
+            id="connection_string dict with password ConfigStr",
+        ),
+        param(
+            {"user": "my_user", "password": "password", "account": "my_account"},
+            id="old config format - top level keys",
+        ),
```
Kilo59, line 58: No params are changing here. Just adding `id` keys.

PR #9095, opened 2023-12-13
[BUGFIX] Run Checkpoints referencing in-memory Assets with Cloud Data Contexts

NathanFarmer:
Previously, we were seeing a 400 when attempting to validate in-memory assets. This PR adds end-to-end tests and fixes buggy behavior for this case (Pandas and Spark).


great_expectations/data_context/store/expectations_store.py
```diff
@@ -135,14 +135,17 @@ def _update(self, key, value, **kwargs):
...
-                    local_suite=value,
-                    cloud_suite=result.response["data"]["attributes"]["suite"],
-                )
+                # if we already had an id, a boolean is returned by _update
+                # otherwise cloud backend has added IDs, and we update our local state to be in sync
+                if isinstance(result, GXCloudResourceRef):
+                    value = self._add_cloud_ids_to_local_suite_and_expectations(
+                        local_suite=value,
+                        cloud_suite=result.response["data"]["attributes"]["suite"],
+                    )
```
NathanFarmer, line 148: This was a bug that was only uncovered because my changes resulted in a `put` when we had the id, instead of a `post`. With a `put` a `bool` is returned and we don't need to update local state.

great_expectations/data_context/store/expectations_store.py
```diff
@@ -135,10 +135,12 @@ def _update(self, key, value, **kwargs):
...
-            result = super()._update(key=key, value=value, **kwargs)
-            if self.cloud_mode:
-                # cloud backend has added IDs, so we update our local state to be in sync
-                result = cast(GXCloudResourceRef, result)
+            result: GXCloudResourceRef | bool | None = super()._update(
+                key=key, value=value, **kwargs
+            )
+            if self.cloud_mode and isinstance(result, GXCloudResourceRef):
+                # if we already had an id, a boolean is returned by _update
+                # otherwise cloud backend has added IDs, and we update our local state to be in sync
```
NathanFarmer: This was a bug that was only uncovered because my changes resulted in a `put` when we had the id, instead of a `post`. With a `put` a `bool` is returned and we don't need to update local state.

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -615,7 +615,7 @@ def _update(
         key: tuple[GXCloudRESTResource, str | None, str | None],
         value: dict,
         **kwargs,
-    ) -> GXCloudResourceRef:
```
NathanFarmer, line 618: I didn't actually change any return types here, this was just wrong.

tests/integration/cloud/end_to_end/conftest.py
```diff
@@ -67,9 +86,6 @@ def expectation_suite(
     )
     assert len(expectation_suite.expectations) == 0
     yield expectation_suite
-    expectation_suite = context.add_or_update_expectation_suite(
-        expectation_suite=expectation_suite
-    )
```
NathanFarmer, line 72: I had to move this into the test modules, because the suite was getting used downstream in tests before it was ever saved.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -549,6 +551,31 @@ def _run_validation(  # noqa: PLR0913
...
+    ) -> list[CheckpointValidationConfig]:
+        """This method ensures that validations passed at runtime do not get added to the
+        complete validations list, if runtime validations are identical to configured validations.
+        """
+        for runtime_validation in runtime_validations:
+            if isinstance(runtime_validation, dict) and (
+                "id" not in runtime_validation.keys()
+            ):
+                for config_validation in self.validations:
+                    if (
```
billdirks, line 566: There is no `else` here so a `runtime_validation` may still not have an `id` after this loop. Is that ok?

great_expectations/checkpoint/util.py
```diff
@@ -403,25 +406,41 @@ def substitute_runtime_config(source_config: dict, runtime_kwargs: dict) -> dict
...
     if runtime_kwargs.get("validations") is not None:
         validations = dest_config.get("validations") or []
-        existing_validations = source_config.get("validations") or []
-        validations.extend(
-            filter(
-                lambda v: v not in existing_validations,
-                runtime_kwargs["validations"],
-            )
-        )
+        runtime_validations = runtime_kwargs.get("validations") or []
```
billdirks, line 409: nit: `.get` allows a default return value
`runtime_validations = runtime_kwargs.get("validations", [])`

billdirks (commented):
I didn't get very far into the review before I got pulled away. I'll look tomorrow.

tests/integration/cloud/end_to_end/test_pandas_datasource.py
```diff
@@ -157,3 +159,95 @@ def test_checkpoint_run(checkpoint: Checkpoint):
...
+    # now that the Data Context is unaware of the Dataframe,
+    # we need to re-associate the DataFrame with the DataFrameAsset
+    # one way to do this is to assign to the attribute directly
+    # and update the datasource
+    # the fixtures came from the old Data Context
+    # we have to get them again, because they exist in a new place in memory
+    datasource = context.get_datasource(datasource_name=datasource.name)  # type: ignore[assignment]
+    in_memory_asset = datasource.get_asset(asset_name=in_memory_asset.name)
+    in_memory_asset.dataframe = pandas_test_df
+    context.update_datasource(datasource=datasource)
```
NathanFarmer: This is weird that we have to update the datasource, but it's not documented to work by assigning to the attribute. It has to do with the logic around caching dataframes during round trips to cloud.

TrangPham:
@NathanFarmer Should this PR be kept open or should it be closed out? Does it still apply? It is a month old now.

NathanFarmer:
> @NathanFarmer Should this PR be kept open or should it be closed out? Does it still apply? It is a month old now.

@TrangPham I am working on this as we speak.

NathanFarmer:
I'm going to put this on top of `0.18.x` and then cherry-pick into `develop`.

PR #9098, opened 2023-12-13
[MAINTENANCE] fix REST API contract tests

dctalbot:
- This PR fixes some issues with the contract tests which were resulting in only the last interaction being shared with the broker.


tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -236,32 +241,16 @@ def _run_pact_test(
...
                 json=contract_interaction.request_body,
                 params=contract_interaction.request_params,
             )
 
-        try:
-            provider_base_url: Final[str] = os.environ["GX_CLOUD_BASE_URL"]
-        except KeyError as e:
-            raise OSError("GX_CLOUD_BASE_URL is not set in this environment.") from e
-
-        verifier = pact.Verifier(
```
dctalbot, line 251: `with pact_test` already includes the verify step

NathanFarmer: Are you sure about this? `with pact_test` runs against a mock service and then produces the json. I think this verify step runs the json against the actual provider.

tests/integration/cloud/rest_contracts/unit_test_mocks/conftest.py
```diff
@@ -1,131 +0,0 @@
-import json
```
NathanFarmer: I know this isn't used right now, but can we leave this in? The idea is to use mocks like this in unit tests once we know how request/response should look.

NathanFarmer (changes requested):

tests/integration/cloud/rest_contracts/conftest.py
dctalbot: The only thing we need to do on the consumer side is produce the contract and back it up with unit test assertions. Then on the provider side, we run the verifier against a real instance of it. That is what distinguishes these tests from end-to-end tests - we do not need to stand up a provider server in this repo.

NathanFarmer: But we have a provider server...

NathanFarmer: If we remove this we should add assertions to the tests, otherwise some tests might pass locally and then fail in CI.

dctalbot: As discussed offline, for now we will have the provider CI take care of the verifier step. If this causes unforeseen pain, we can revisit this and consider doing verification in both repos.

NathanFarmer (approved):

PR #9105, opened 2023-12-14
[FEATURE] Require IDs for ExpectationSuite and Expectations

joshua-stauffer:


great_expectations/data_context/store/expectations_store.py
```diff
@@ -190,20 +112,79 @@ def gx_cloud_response_json_to_object_dict(response_json: Dict) -> Dict:
...
                 f"Could not find an existing ExpectationSuite named {value.expectation_suite_name}."
+            ) from exc
+
+    def _add_ids_to_new_objects(self, suite: ExpectationSuite) -> ExpectationSuite:
+        if self.cloud_mode:
+            # cloud backend creates IDs, and we add them after persistence
+            return suite
+        # ensure suite has an ID
+        if not suite.get("ge_cloud_id"):
+            suite["ge_cloud_id"] = str(uuid.uuid4())
```
tyler-hoffman, line 171: Still looking at the PR, but I think it's a problem to assign `ge_cloud_id` when we are not in cloud mode.

great_expectations/data_context/store/expectations_store.py
```diff
@@ -190,20 +112,79 @@ def gx_cloud_response_json_to_object_dict(response_json: Dict) -> Dict:
...
+        if isinstance(suite, ExpectationSuite):
+            key = "expectation_configurations"
+        else:
+            # this will be true if a serialized suite is provided
```
tyler-hoffman, line 175: How does mypy let us get here given the method signature? Shouldn't require actual domain objects rather than dicts at this point? If we really have a use case for dicts (which I'm assuming we do, given you're handling it), I think we should have a separate method for each (and in the suite one, we should use dot notation for accessing values)

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: do you know where this is a problem? i was wondering if someone has pinned logic to `ge_cloud_id`, but haven't run across an example of it. it'll be renamed to just `id` in a followup PR.

great_expectations/data_context/store/expectations_store.py
```diff
@@ -190,20 +112,79 @@ def gx_cloud_response_json_to_object_dict(response_json: Dict) -> Dict:
...
+            if not expectation_configuration["ge_cloud_id"]:
+                expectation_configuration["ge_cloud_id"] = str(uuid.uuid4())
+        return suite
+
+    def _add_cloud_ids_to_local_suite_and_expectations(
+        self, local_suite: ExpectationSuite, cloud_suite: dict
+    ) -> ExpectationSuite:
+        if not self.cloud_mode:
+            # non-cloud backends must create IDs prior to passing suite to store backend.
+            return local_suite
```
tyler-hoffman: Feels off to even call this method if we are in cloud method. I'd put this check outside of this method.

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -327,8 +327,13 @@ def _put(self, id: str, value: Any) -> bool:
                 response_status_code = response.status_code
 
             response.raise_for_status()
-            return True
-
+            try:
+                response_json = response.json()
+            except JSONDecodeError:
+                return True
```
tyler-hoffman, line 338: wat

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -289,8 +290,7 @@ def _send_get_request_to_api(self, url: str, params: dict | None = None) -> dict
     def _move(self) -> None:  # type: ignore[override]
         pass
 
-    # TODO: GG 20220810 return the `ResponsePayload`
-    def _put(self, id: str, value: Any) -> bool:
+    def _put(self, id: str, value: Any) -> GXCloudResourceRef | bool:
```
tyler-hoffman, line 292: This signature stresses me out. Were we actually using the return value before? Looks like we only ever returned `True`, making the return value useless IIUC. Can we just change it to `GXCloudResourceRef`?

tests/core/test_expectation_suite.py
```diff
@@ -360,10 +359,52 @@ def test_save_success(self):
...
+        # todo: update to new api
+        suite = context.add_expectation_suite(suite_name)
+
+        uuid_to_test = suite.ge_cloud_id
+        assert isinstance(UUID(uuid_to_test), UUID)
+
+        suite.add(expectation)
+        expectation.column = "foo"
+        suite.add(expectation)
+        assert len(suite.expectations) == 2
```
tyler-hoffman, line 383: I think it's probably worthwhile to assert that these don't have uuids before being added to the suite as well so that we actually know that the suite is handling it.

tyler-hoffman (changes requested):
High level seems about right. I have concerns about reusing the ge_cloud_id. We should either rename it `id` (I'm fine if that's in a subsequent PR to avoid blowing up the scope of this), or we should have both an `id` and the separate `ge_cloud_id` (I thought that was a pattern in some of our code?)

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: thanks, reasonable. updated in [4b1a9e6](https://github.com/great-expectations/great_expectations/pull/9105/commits/4b1a9e61cf0ac5d50c5f9552c31bc66de6003c98)

great_expectations/data_context/store/gx_cloud_store_backend.py
joshua-stauffer: yeah. this method is used by all store implementations, and other `update` calls (like Checkpoint) do return a boolean. fixing the store implementation is explicitly out of scope, but let's improve this soon.

great_expectations/data_context/store/gx_cloud_store_backend.py
joshua-stauffer: true - unfortunately other `update` endpoints still return bools

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: the `migrator` code is passing a`dict` here. I think that's probably wrong, and would like to remove that code, but that feels out of scope for this work, and this fork is required to get tests to pass.

great_expectations/data_context/store/expectations_store.py
tyler-hoffman: I think it's just a really misleading name at this point. Let's follow up and rename.

great_expectations/data_context/store/expectations_store.py
tyler-hoffman: I would like the signature to read suite: `dict | ExpectationSuite` before this merges

great_expectations/data_context/store/gx_cloud_store_backend.py
tyler-hoffman: Can you add a "why" comment on this!

tyler-hoffman (approved):
Approved pending convos being resolved.

great_expectations/data_context/store/gx_cloud_store_backend.py
joshua-stauffer: even better - updated this to use the returned HTTP code instead of exceptions. It does require that the backend use the correct code, but is definitely easier to reason about

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: unfortunately that change causes other typing issues. mypy seems convinced that the only thing ever passed to this object is a dict, even though the `CloudMigrator` is passing a dict. For now I'll leave this signature in place, but I'll add a ticket to remove the migrator as a follow up, and change this code to use dot notation. It would need to change at the latest when we migrate the suite object to pydantic, since bracket syntax will no longer work.

PR #9111, opened 2023-12-15
[MAINTENANCE] yaml to batch config

tyler-hoffman:
**Goal**
This gets the yaml -> `BatchConfig` flow working

**Main updates**
* Moves `data_asset` to be a `PrivateAttr` on `BatchConfig`
  * reason: In the yaml flow, this binding must occur after instantiation
  * pydantic requires `PrivateAttr`s be prefixed with `_`. So then I made a property that points to that. Probably better that way anyway, since it makes the referenec pseudo-immutable
  * This had the side effect of massive updates to our schema
* Adds a step after parsing the config to do the wiring up. I would have liked to do this via a pydantic validator, but AFAIK, those don't take in the instance of the object itself. Call me out if I'm missing something!
* Adds some mapping to/from lists/dicts for batch configs. This follows the same pattern that is used for data assets, and I don't like it. But I'd like to save the refactor for later.

**Tips to reviewers**
* To see the files I directly touched in the file viewer, close out `great_expectations/datasource/fluent/schemas/` in the file viewer (but maybe also sanity check a couple of those to make sure I didn't goof anything)
* Am I wrong on validators not taking in a reference to self? Can I make the wiring up of BatchConfig -> DataAsset happen automatically?


great_expectations/datasource/fluent/config.py
```diff
@@ -358,6 +365,17 @@ def _exclude_name_fields_from_fluent_datasources(
                         )
                         for data_asset_config in data_assets
                     }
+                    for data_asset in data_assets_config_as_dict.values():
```
tyler-hoffman, line 370: TODO: refactor this. Probably the whole function

great_expectations/datasource/fluent/config.py
```diff
@@ -330,6 +366,12 @@ def yaml(  # noqa: PLR0913
 
         return stream_or_path.getvalue()
 
+    def _link_batch_configs_to_their_data_assets(self) -> None:
+        for datasource in self.fluent_datasources:
```
billdirks: I'm a little unclear on when different things happen so maybe this isn't relevant.

Are you always calling this when creating a `GXConfig` object? I'm wondering if you can do this in `__init__` to avoid the `parse_*` methods?

You also mention, in your top level comment, about wanting to use a validator but it only takes classes and not instances. There is a `root_validator` which is still a class method but gives you the whole dictionary as input. You can populate/update entries in that and return it to get things passed to the __init__ method.

great_expectations/datasource/fluent/config.py
```diff
@@ -390,17 +443,39 @@ def _convert_fluent_datasources_loaded_from_yaml_to_internal_object_representati
         datasource_config: dict
         for datasource_name, datasource_config in fluent_datasources.items():
             datasource_config[_DATASOURCE_NAME_KEY] = datasource_name
-            if "assets" in datasource_config:
-                data_assets: dict = datasource_config["assets"]
+            if _ASSETS_KEY in datasource_config:
+                data_assets: dict = datasource_config[_ASSETS_KEY]
+                print("a", type(data_assets))
```
billdirks: stray print?

great_expectations/datasource/fluent/config.py
```diff
@@ -390,17 +443,39 @@ def _convert_fluent_datasources_loaded_from_yaml_to_internal_object_representati
...
 
         config[_FLUENT_DATASOURCES_KEY] = list(fluent_datasources.values())
 
     return config
+
+
+def _convert_batch_configs_from_yaml_to_internal_object_representation(
+    batch_configs: Dict[str, Dict]
+) -> List[Dict]:
+    print("ba", type(batch_configs))
```
billdirks: stray print?

billdirks (commented):
Thanks for doing this. Overall, this looks good. I had a comment since I don't know the details of these pydantic helper methods.

great_expectations/datasource/fluent/config.py
tyler-hoffman: Thanks! I overcomplicated things here! I ended up reusing an existing validator to do this.

great_expectations/datasource/fluent/config.py
```diff
@@ -221,6 +228,8 @@ def _load_datasource_subtype(cls, v: List[dict]):
                 # attach the datasource to the nested assets, avoiding recursion errors
                 for asset in datasource.assets:
                     asset._datasource = datasource
+                    for batch_config in asset.batch_configs:
+                        batch_config._data_asset = asset
```
cdkini, line 232: Could we use a setter instead of touching a private attr?

great_expectations/datasource/fluent/config.py
```diff
@@ -390,17 +410,37 @@ def _convert_fluent_datasources_loaded_from_yaml_to_internal_object_representati
...
+    batch_configs: Dict[str, Dict]
+) -> List[Dict]:
+    for (
+        batch_config_name,
+        batch_config,
+    ) in batch_configs.items():
+        if not isinstance(batch_config_name, str):
+            raise ValueError("BatchConfig name must be a string")
+        if not isinstance(batch_config, dict):
+            raise ValueError("BatchConfig must be a dictionary")
```
cdkini: Do we need this runtime validation? Are we not guaranteed this?

great_expectations/datasource/fluent/config.py
tyler-hoffman: I would prefer that too, but my understanding is that there is a bug in pydatic that prevents that: https://github.com/pydantic/pydantic/issues/1577.

great_expectations/datasource/fluent/config.py
tyler-hoffman: Oh good call. I think on a previous iteration they were needed, but no longer! Updated.

cdkini (approved):

PR #9115, opened 2023-12-15
[MAINTENANCE] Improved error message for `expect_column_values_to_be_in_type_list`

dctalbot:
- Improved error message for `expect_column_values_to_be_in_type_list`

Before:
Screenshot 2023-12-15 at 5 02 24 PM

After:
Screenshot 2023-12-15 at 5 02 36 PM


josectobar (changes requested):
This is great, is there a test we can add that reflects this change?

NathanFarmer (commented):
Looks good! I agree with Jose that we could add a regression test. It should be fine to put it in: `tests/expectations/core/test_expect_column_values_to_be_in_type_list.py`.

Issue #8618, opened 2023-08-25
Assets of type query not working for Oracle datasource


hjannasch:
**Describe the bug**
As our database stack if mainly Oracle based. I tried to setup an expectation suite using a fluent oracledb sql datasource, but using assets of type 'query' it not working.
Well I know you are not supporting Oracle officially, but do you have any idea on this?

**To Reproduce**
```
fluent_datasources:
  dm:
    type: sql
    assets:
      anrede:
        type: table
        order_by: []
        batch_metadata: {}
        table_name: anrede
        schema_name:
      person_query:
        type: query
        order_by: []
        batch_metadata: {}
        query: select * from PERSON
    connection_string: oracle+oracledb://<user>:<pwd>@localhost:1521/xe`
```

The table asset works fine, but the query asset generates a none valid SQL.
When the person_query asset is used to create a new validator and suite:
```
import great_expectations as gx

context = gx.get_context()
data_asset = context.get_datasource("dm").get_asset("person_query")
batch_request = data_asset.build_batch_request()

validator = context.get_validator(
    batch_request=batch_request,
    expectation_suite_name="nat_query_expectation_suite",
)
validator.head()
```
This leads to the error:
DatabaseError: (oracledb.exceptions.DatabaseError) ORA-00907: missing right parenthesis
[SQL: CREATE GLOBAL TEMPORARY TABLE gx_temp_5da3476a ON COMMIT PRESERVE ROWS AS SELECT * 
FROM (SELECT * from PERSON FROM DUAL) anon_1 
WHERE 1 = 1]

**Expected behavior**
Generate valid SQL, it looks like ... FROM DUAL is too much here.

**Environment (please complete the following information):**
 - Operating System: Linux
 - Great Expectations Version: 0.17.12
 - Data Source: SQL (Oracle with oracledb driver)

great-expectations==0.17.12
sqlalchemy==2.0.20
oracledb==1.4.0

hjannasch:
Here is the trace.
[traceback.log](https://github.com/great-expectations/great_expectations/files/12436216/traceback.log)

Kilo59:
@hjannasch You can try changing the value of `create_temp_table = False` when defining the Datasource but I suspect you'll still end up with some version of a bad identifier issue when you run a checkpoint.

Please report back if this does or doesn't work.

```yaml
fluent_datasources:
  dm:
    type: sql
    create_temp_table: false
    assets:
      anrede:
        type: table
        order_by: []
[...]

Issue #8325, opened 2023-07-14
Error when running an expectation with `create_temp_table` is set to `False`


moatazelmohtaseb:
**Describe the bug**
In my current environment I want to run validations on an SQL data source (which is a PostgreSQL database), But as the database is a read-replica, I set the datasource config `create_temp_table` to `False` in order to avoid an error the was thrown when great-expectations tried to create a temp table. But I got a different error instead.

Note: the validation worked when it was a read/write instance of the database.

**To Reproduce**
datasource.py:
```
pg_datasource = context.sources.add_or_update_sql(
    name="test_datasource",
    connection_string=my_connection_string,
    create_temp_table=False,
)
query_asset = pg_datasource.add_query_asset(name="test_asset", query="SELECT id  FROM table")
```
Code:
```
data_asset = context.get_datasource("test_datasource").get_asset("test_asset")
expectation_suite_name = "test_suite"
batch_request = data_asset.build_batch_request(options={})
context.add_or_update_expectation_suite(expectation_suite_name)

validator = context.get_validator(
    batch_request=batch_request,
    expectation_suite_name=expectation_suite_name,
)
expectation_validation_result2 = validator.expect_column_values_to_be_between(
    "order_id", 1,80
)
```
great_expectations.yml:
```
# Welcome to Great Expectations! Always know what to expect from your data.
#
# Here you can define datasources, batch kwargs generators, integrations and
# more. This file is intended to be committed to your repo. For help with
# configuration please:
#   - Read our docs: https://docs.greatexpectations.io/docs/guides/connecting_to_your_data/connect_to_data_overview/#2-configure-your-datasource
#   - Join our slack channel: http://greatexpectations.io/slack

# config_version refers to the syntactic version of this config file, and is used in maintaining backwards compatibility
# It is auto-generated and usually does not need to be changed.
config_version: 3.0

# Datasources tell Great Expectations where your data lives and how to get it.
# You can use the CLI command `great_expectations datasource new` to help you
# add a new datasource. Read more at https://docs.greatexpectations.io/docs/guides/connecting_to_your_data/connect_to_data_overview
datasources: {}

# This config file supports variable substitution which enables: 1) keeping
# secrets out of source control & 2) environment-based configuration changes
# such as staging vs prod.
#
# When GX encounters substitution syntax (like `my_key: ${my_value}` or
# `my_key: $my_value`) in the great_expectations.yml file, it will attempt
# to replace the value of `my_key` with the value from an environment
[...]

Issue #8147, opened 2023-06-20
datasource.add_query_asset not respecting create_temp_table equals false


Chr96er:
This issue is analogue to https://github.com/great-expectations/great_expectations/issues/6064.

Describe the bug
When using `datasource.add_query_asset`, the `create_temp_table=false` is not respected.

Reproducible example:
```
import great_expectations as gx

context = gx.get_context()

connection_string = """bigquery://bigquery-public-data/samples"""

datasource = context.sources.add_or_update_sql(
    name="github_timeline", connection_string=connection_string, create_temp_table=False
)

table_asset = datasource.add_query_asset(
        name="github_timeline", 
        query="SELECT * FROM `bigquery-public-data.samples.github_timeline`"
)

my_batch_request = table_asset.build_batch_request()

batches = table_asset.get_batch_list_from_batch_request(my_batch_request)

context.add_or_update_expectation_suite("github_timeline_suite")

validator = context.get_validator(
    batch_request=my_batch_request,
    expectation_suite_name="github_timeline_suite",
)

validator.expect_column_values_to_not_be_null(column = "repository_url")


```

```
NotFound: 404 GET https://bigquery.googleapis.com/bigquery/v2/projects/bigquery-public-data/datasets/samples/tables/great_expectations_sub_selection?prettyPrint=false: Not found: Table bigquery-public-data:samples.great_expectations_sub_selection

During handling of the above exception, another exception occurred:

NoSuchTableError                          Traceback (most recent call last)
```

Environment (please complete the following information):

Operating System: Ubuntu
Great Expectations Version: 0.17.0
.

HaebichanGX:
Hey @Chr96er thanks again for raising this issue! I've put it in our backlog for review.

Kilo59:
@Chr96er, I believe this was fixed a few version back. But I don't recall exactly which one.

molliemarie:
Hello @Chr96er. With the upcoming launch of Great Expectations Core (GX 1.0), we are closing old issues posted regarding previous versions. Moving forward, we will focus our resources on supporting and improving GX Core (version 1.0 and beyond). If you find that an issue you previously reported still exists in GX Core, we encourage you to resubmit it against the new version. With more resources dedicated to community support, we aim to tackle new issues swiftly. For specific details on what is GX-supported vs community-supported, you can reference our [integration and support policy](https://docs.greatexpectations.io/docs/application_integration_support/).

[...]

PR #9116, opened 2023-12-16
[BUGFIX] Change default for create_temp_table to False

roblim:

### DIFF
--- a/great_expectations/execution_engine/sqlalchemy_execution_engine.py
+++ b/great_expectations/execution_engine/sqlalchemy_execution_engine.py
@@ -305,7 +305,7 @@ class SqlAlchemyExecutionEngine(ExecutionEngine):
         connection_string: Optional[str] = None,
         url: Optional[str] = None,
         batch_data_dict: Optional[dict] = None,
-        create_temp_table: bool = False,
+        create_temp_table: bool = True,
         concurrency: Optional[ConcurrencyConfig] = None,
         # kwargs will be passed as optional parameters to the SQLAlchemy engine, **not** the ExecutionEngine
         **kwargs,
--- a/great_expectations/execution_engine/sqlalchemy_execution_engine.py
+++ b/great_expectations/execution_engine/sqlalchemy_execution_engine.py
@@ -305,7 +305,7 @@ class SqlAlchemyExecutionEngine(ExecutionEngine):
         connection_string: Optional[str] = None,
         url: Optional[str] = None,
         batch_data_dict: Optional[dict] = None,
-        create_temp_table: bool = True,
+        create_temp_table: bool = False,
         concurrency: Optional[ConcurrencyConfig] = None,
         # kwargs will be passed as optional parameters to the SQLAlchemy engine, **not** the ExecutionEngine
         **kwargs,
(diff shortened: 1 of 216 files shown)


billdirks (commented):
Thanks for doing this! I have a few comments:

1. It looks like some tests are failing. Likely these depended on the old default value.
2. Have you manually verified that this fixes the observed issue? If you expect you will need other PRs to fix this issue, we can delay this check until later but we should manually verify the fix before releasing.
3. I'm not sure if the `_SQLAsset.get_batch_list_from_batch_request` is actually using this `create_temp_table` when it is populating `batch_spec_kwargs`. We should investigate though that can happen in a separate PR.
4. We should get this into `develop`. If we want to release in a `0.18.x` build, after we are happy with the fix we will need to make another PR into that branch.

roblim:
> Thanks for doing this! I have a few comments:
> 
> 1. It looks like some tests are failing. Likely these depended on the old default value.
> 2. Have you manually verified that this fixes the observed issue? If you expect you will need other PRs to fix this issue, we can delay this check until later but we should manually verify the fix before releasing.
> 3. I'm not sure if the `_SQLAsset.get_batch_list_from_batch_request` is actually using this `create_temp_table` when it is populating `batch_spec_kwargs`. We should investigate though that can happen in a separate PR.
> 4. We should get this into `develop`. If we want to release in a `0.18.x` build, after we are happy with the fix we will need to make another PR into that branch.

1. Took longer than expected, but fixed all the tests and everything is passing now.
2. Yep - confirmed that changes fix observed issue.
3. Was also able to confirm that it appears `_SQLAsset.get_batch_list_from_batch_request` is actually using this `create_temp_table`. Already had this call in notebook I've been using to test and that call only takes a few seconds now instead of 20 minutes.
4. Def want to release in a `0.18.x` build. Can discuss on our Monday call.

Thanks!

NathanFarmer (approved):

billdirks (approved):
Thanks for following up on all my questions!

Kilo59 (approved):

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -305,7 +305,7 @@ def __init__(  # noqa: C901, PLR0912, PLR0913, PLR0915
         connection_string: Optional[str] = None,
         url: Optional[str] = None,
         batch_data_dict: Optional[dict] = None,
-        create_temp_table: bool = True,
+        create_temp_table: bool = False,
```
Kilo59: This shouldn't matter or need to change to change at all.
I'll look into this. I have a guess about what's going on, might be related to default values not being serialized 

Probably also related to these issues
 1. https://github.com/great-expectations/great_expectations/issues/8147
 2. https://github.com/great-expectations/great_expectations/issues/8618
 3. https://github.com/great-expectations/great_expectations/issues/8325

Kilo59: That seems to be the issue.
We should revert this change once this merges 👇 
- https://github.com/great-expectations/great_expectations/pull/9118

Kilo59 (changes requested):
I don't think we need to change anything about the execution engine itself.
This 👇   should prevent the need for changes there.
- https://github.com/great-expectations/great_expectations/pull/9118

Kilo59 (approved):

PR #9125, opened 2023-12-18
[FEATURE] Add Expectation.save()

joshua-stauffer:


great_expectations/core/expectation_suite.py
```diff
@@ -144,31 +144,40 @@ def name(self) -> str:
...
 
     @public_api
     def add(self, expectation: Expectation) -> Expectation:
         """Add an Expectation to the collection."""
-        if not any(
-            expectation.configuration == existing_config
+        if expectation.id:
+            raise RuntimeError(
+                "Cannot add Expectation because it already belongs to an ExpectationSuite. "
+                + "Please make a copy of your Expectation with `copy(expectation)` and try again."
```
joshua-stauffer: this requires implementing `Expectation.__copy__`, but i think this is a good way to allow users to reuse expectations in multiple suites. ex:

billdirks: Since not all expectations implement copy I'd remove the "with \`copy(expectation)\`". Also, they might want to update instead. I would either drop the second sentence or have it say something like `If you want to update your expectation, please call update. If you are copying it to a new expectation suite, please copy it first (the core expectations and some others support copy(expectation)) before adding it.` We could also check if this expectation already lives in this suite.

nit: You can remove the `+`. Python will concatenate strings on separate lines automatically.

great_expectations/core/expectation_suite.py
```diff
@@ -144,31 +144,40 @@ def name(self) -> str:
...
-
-        if self._has_been_saved():
-            # only persist on add if the suite has already been saved
-            try:
-                self.save()
-            except Exception as exc:
-                # rollback this change
-                self.expectation_configurations.pop()
-                raise exc
+            pass  # suite is a set-like collection, so don't add if it not unique
```
billdirks: I would change this to either:

*  Remove the `else` and add the comment after the `if` to be `suite is a set-like collection so we only add an expectation if it is unique`
* Keep the `else` and log here. I'd do this if we think this code path is being invoked in a "bad" way.

It seems like we should go with the first option?

great_expectations/data_context/store/expectations_store.py
```diff
@@ -111,6 +114,38 @@ def gx_cloud_response_json_to_object_dict(response_json: Dict) -> Dict:
...
         return suite_dict
 
+    def add_expectation(
+        self, suite: ExpectationSuite, expectation: Expectation
+    ) -> Expectation:
+        suite_identifier = self.get_key(suite=suite)
+        suite_dict = self.get(key=suite_identifier)
+        suite = ExpectationSuite(**suite_dict)
+
+        expectation.id = str(uuid.uuid4())
```
billdirks: Is it ok if we determine the ids that cloud uses for expectations?

great_expectations/data_context/store/expectations_store.py
```diff
@@ -111,6 +114,38 @@ def gx_cloud_response_json_to_object_dict(response_json: Dict) -> Dict:
...
+
+        self.update(key=suite_identifier, value=suite)
+        return expectation
+
+    def update_expectation(
+        self, suite: ExpectationSuite, expectation: Expectation
+    ) -> Expectation:
+        suite_identifier = self.get_key(suite=suite)
+        suite_dict = self.get(key=suite_identifier)
+        suite = ExpectationSuite(**suite_dict)
```
billdirks: Could you put this in a method called something like `refresh_suite`? We do this for the other crud operations also and encapsulating into something with a name would make this more readable for me. I'm surprised `delete` isn't here?

great_expectations/data_context/store/expectations_store.py
```diff
@@ -190,7 +225,7 @@ def _add_ids_on_update(self, suite: ExpectationSuite) -> ExpectationSuite:
     def _add_cloud_ids_to_local_suite_and_expectations(
         self, local_suite: ExpectationSuite, cloud_suite: dict
     ) -> ExpectationSuite:
-        if not local_suite.ge_cloud_id:
+        if not local_suite["ge_cloud_id"]:
```
billdirks: Why change this from dotted syntax to a dictionary like key lookup?

tests/data_context/store/test_expectations_store.py
```diff
@@ -310,8 +310,177 @@ def test_get_key_in_non_cloud_mode(empty_data_context):
...
+        column="a",
+        value_set=[1, 2, 3],
+        result_format="BASIC",
+    )
+    suite = context.add_expectation_suite(
+        suite_name, expectations=[expectation.configuration]
+    )
+    key = store.get_key(suite)
+    expectation = suite.expectations[0]
+    updated_column_name = "foo"
```
billdirks: Can you add an assert that the current name is `a` before you change it to foo?

tests/data_context/store/test_expectations_store.py
```diff
@@ -310,8 +310,177 @@ def test_get_key_in_non_cloud_mode(empty_data_context):
...
+    expectation = ExpectColumnValuesToBeInSet(
+        column="a",
+        value_set=[1, 2, 3],
+        result_format="BASIC",
+    )
+    suite = context.add_expectation_suite(
+        suite_name, expectations=[expectation.configuration]
+    )
+    key = store.get_key(suite)
+    expectation = suite.expectations[0]
```
billdirks: I've move these lines to the act section. I think the arrange is setting up the expectation suite with fixed data. The action is retrieving the expectation and updating it.

tests/data_context/store/test_expectations_store.py
```diff
@@ -310,8 +310,177 @@ def test_get_key_in_non_cloud_mode(empty_data_context):
...
+    suite_name = "test-suite"
+    expectation = ExpectColumnValuesToBeInSet(
+        column="a",
+        value_set=[1, 2, 3],
+        result_format="BASIC",
+    )
+    suite = context.add_expectation_suite(
+        suite_name, expectations=[expectation.configuration]
+    )
+    key = store.get_key(suite)
```
billdirks: I'd inline this call into the `store.get` call. Or, if you don't want to do that, I'd move it right before you use the key.

tests/data_context/store/test_expectations_store.py
```diff
@@ -310,8 +310,177 @@ def test_get_key_in_non_cloud_mode(empty_data_context):
...
+    context = empty_data_context
+    store = context.expectations_store
+    suite_name = "test-suite"
+    expectation = ExpectColumnValuesToBeInSet(
+        id="2b284004-0e0e-455d-a7f4-11e162fd06c9",
+        column="a",
+        value_set=[1, 2, 3],
+        result_format="BASIC",
+    )
+    suite = context.add_expectation_suite(
```
billdirks, line 487: This seems to be a strange test because it's basically demonstrating `add_expectation_suite` behaves in an non-intuitive way. That is, adding the `expectation.configuration` doesn't add the expectation. We don't want to preserve this behavior and we want to get rid of `add_expectation_suite` where we pass in configuration instead of Expectation.

While this documents current behavior, I would delete this test since this isn't guarding behavior we want and we plant to remove this before we ship. Another option would be to add a comment explaining what this is testing and why.

If I'm refactoring and see I test, I usually assume that it's behavior I want to preserve.

billdirks (changes requested):
Thanks for doing this. I've added some comments inline.

tests/data_context/store/test_expectations_store.py
joshua-stauffer: hm good catch - this test wasn't supposed to pass an expectation in here, so i'm surprised that it passes.

joshua-stauffer: ah right - the issue here is that the ID which was passed in with the expectation gets overwritten, since the backend defines IDs. So when we try to update that expectation, no IDs match and it raises an error. I'll remove the expectation from the suite's init to make that clearer.

I do see why this behavior was unexpected as a reader of the code, though - should we raise an exception instead of overwriting IDs?

tests/data_context/store/test_expectations_store.py
joshua-stauffer: okay - i see this as part of the setup, so had it up top. if its clearer inline, that's fine too.

tests/data_context/store/test_expectations_store.py
joshua-stauffer: updated in [b261708](https://github.com/great-expectations/great_expectations/pull/9125/commits/b26170837df1d346a0633c33950306eaa7ef47b9)

tests/data_context/store/test_expectations_store.py
joshua-stauffer: [b5f13be](https://github.com/great-expectations/great_expectations/pull/9125/commits/b5f13bea3246ec811d5b0f5e59e1f730bd4cb9a1)

tests/data_context/store/test_expectations_store.py
joshua-stauffer: [b5f13be](https://github.com/great-expectations/great_expectations/pull/9125/commits/b5f13bea3246ec811d5b0f5e59e1f730bd4cb9a1)

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: thanks - `delete` is required, i'll add it

joshua-stauffer: `delete_expectation` is added in [60a7207](https://github.com/great-expectations/great_expectations/pull/9125/commits/60a72071bca34e62eed2fc7365a64f571fda88e7), `_refresh_suite` added in [c71e322](https://github.com/great-expectations/great_expectations/pull/9125/commits/c71e322ef4eb0919817517a7da17239d5adb1608)

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: i think this was because the `CloudMigrator` tests are passing serialized suites to the store instead of deserialized suites - will confirm

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: nope - we need to set the ID to `None` here so the backend knows its new [bdf6073](https://github.com/great-expectations/great_expectations/pull/9125/commits/bdf6073ba3da7de1e2836bf8480f3d8f24e6ffff)

great_expectations/core/expectation_suite.py
joshua-stauffer: thanks - updated in [77be13d](https://github.com/great-expectations/great_expectations/pull/9125/commits/77be13d67a3a0c3cd7987690a0a011584b11da4c)

great_expectations/core/expectation_suite.py
joshua-stauffer: sure - updated in [c23eb1b](https://github.com/great-expectations/great_expectations/pull/9125/commits/c23eb1b449510e790b7adaf9dc9fd6b366c3f8a4)

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
```diff
@@ -577,32 +574,6 @@ def test_save_expectation_suite_no_overwrite_id_collision_raises_error(
     )
 
 
-@pytest.mark.cloud
-def test_add_or_update_expectation_suite_adds_new_obj(
```
joshua-stauffer, line 581: Removing because `ExpectationsStore` now uses the returned expectation suite to obtain IDs, these tests don't correctly mock it, and this API will be removed soon.

tests/data_context/cloud_data_context/test_expectation_suite_crud.py
```diff
@@ -625,31 +596,6 @@ def test_expectation_suite_gx_cloud_identifier_requires_id_or_resource_name(
         context.expectations_store._validate_key(key=key)
 
 
-@pytest.mark.cloud
-def test_add_or_update_expectation_suite_updates_existing_obj(
```
joshua-stauffer, line 629: Removing because `ExpectationsStore` now uses the returned expectation suite to obtain IDs, these tests don't correctly mock it, and this API will be removed soon.

great_expectations/expectations/expectation.py
```diff
@@ -297,6 +297,10 @@ class Config:
     expectation_type: ClassVar[str]
     examples: ClassVar[List[dict]] = []
 
+    _save_callback: Union[Callable[[Expectation], Expectation]] = pydantic.PrivateAttr(
```
billdirks: This type looks wrong since the default value is `None` and the types not `Optional`. Maybe some pydantic magic is confusing mypy (the real object is wrapped in a pydantic one). It's also a Union with currently only 1 element.  We do null checks below. Could you add `None` to the Union?

tests/core/test_expectation_suite.py
```diff
@@ -409,24 +414,93 @@ def test_cloud_context_update_suite_adds_ids(
...
+
+    @pytest.mark.cloud
+    def test_cloud_expectation_can_be_saved_after_added(
+        self, empty_cloud_context_fluent, expectation
+    ):
+        context = empty_cloud_context_fluent
+        suite_name = "test-suite"
+        # todo: update to new api
+        suite = context.add_expectation_suite(suite_name)
+        suite.add(expectation)
```
billdirks: You might want to assert the expectation.column != "foo" before you update it.

tests/core/test_expectation_suite.py
```diff
@@ -409,24 +414,93 @@ def test_cloud_context_update_suite_adds_ids(
...
+        expectation.id = provided_id
+
+        with pytest.raises(
+            RuntimeError,
+            match="Cannot add Expectation because it already belongs to an ExpectationSuite.",
+        ):
+            suite.add(expectation)
+
+    @pytest.mark.cloud
+    def test_cloud_expectation_can_be_saved_after_added(
```
billdirks, line 417: This test and the following are identical except for the context fixture. To avoid repetition you could parameterize it, or, I think easier, make a helper function:

```
def _test_can_save_after_add(self, context, expectation):
    # test code you have here

@pytest.mark.cloud
def test_cloud_expectation_can_be_saved_after_added(
    self, empty_cloud_context_fluent, expectation
):
    self._test_can_save_after_add(empty_cloud_context_fluent, expectation)

@pytest.mark.filesystem
def test_cloud_expectation_can_be_saved_after_added(
    self, empty_data_context, expectation
):
    self._test_can_save_after_add(empty_cloud_context_fluent, expectation)

```

tests/core/test_expectation_suite.py
```diff
@@ -409,24 +414,93 @@ def test_cloud_context_update_suite_adds_ids(
...
+        updated_column_name = "foo"
+        expectation.column = updated_column_name
+        expectation.save()
+
+        suite = context.get_expectation_suite(suite_name)
+        assert len(suite.expectations) == 1
+        assert suite.expectations[0].column == updated_column_name
+
+    @pytest.mark.cloud
+    def test_cloud_expectation_can_be_saved_after_update(
```
billdirks, line 444: Similar comment about making a helper function here.

tests/data_context/store/test_expectations_store.py
```diff
@@ -310,8 +310,262 @@ def test_get_key_in_non_cloud_mode(empty_data_context):
...
-    project_manager.set_project(cloud_data_context)
     name = "test-name"
     suite = ExpectationSuite(expectation_suite_name=name)
     key = cloud_data_context.expectations_store.get_key(suite)
     assert isinstance(key, GXCloudIdentifier)
+
+
+@pytest.mark.cloud
+def test_add_expectation_success_cloud_backend(empty_cloud_data_context):
+    # Arrange
```
billdirks, line 321: Similar helper function comment here.

tests/data_context/store/test_expectations_store.py
```diff
@@ -310,8 +310,262 @@ def test_get_key_in_non_cloud_mode(empty_data_context):
...
+    updated_suite = ExpectationSuite(**updated_suite_dict)
+    added_expectation = updated_suite.expectations[0]
+    assert UUID(added_expectation.id)
+    assert expectation.column == added_expectation.column
+    assert expectation.value_set == added_expectation.value_set
+    assert expectation.result_format == added_expectation.result_format
+
+
+@pytest.mark.filesystem
+def test_add_expectation_disregards_provided_id(empty_data_context):
```
billdirks: I wonder if this is the behavior we want or if we want to raise when an id is already present. I guess the good with this is we can `copy` expectations easily?

tests/data_context/store/test_expectations_store.py
```diff
@@ -310,8 +310,262 @@ def test_get_key_in_non_cloud_mode(empty_data_context):
...
+    # Assert
+    updated_suite_dict = store.get(key=key)
+    updated_suite = ExpectationSuite(**updated_suite_dict)
+    added_expectation = updated_suite.expectations[0]
+    assert UUID(added_expectation.id)
+    assert added_expectation.id != provided_id
+
+
+@pytest.mark.cloud
+def test_update_expectation_success_cloud_backend(empty_cloud_data_context):
```
billdirks, line 389: You can use a helper function for these 2 tests also.

tests/data_context/store/test_expectations_store.py
```diff
@@ -310,8 +310,262 @@ def test_get_key_in_non_cloud_mode(empty_data_context):
...
+        store.update_expectation(suite=suite, expectation=expectation)
+
+    # Assert
+    updated_suite_dict = store.get(key=key)
+    updated_suite = ExpectationSuite(**updated_suite_dict)
+    assert suite == updated_suite
+
+
+@pytest.mark.cloud
+def test_delete_expectation_success_cloud_backend(empty_cloud_data_context):
```
billdirks, line 466: Helper function to combine tests here also.

billdirks (commented):
Thanks! I've added a few small comments, mostly around reusing code in test.

tests/data_context/store/test_expectations_store.py
```diff
@@ -310,8 +310,262 @@ def test_get_key_in_non_cloud_mode(empty_data_context):
...
+    provided_id = "e86bb8a8-b75f-4efb-a3bb-210b6440661e"
+    expectation = ExpectColumnValuesToBeInSet(
+        id=provided_id,
+        column="a",
+        value_set=[1, 2, 3],
+        result_format="BASIC",
+    )
+
+    # Act
+    store.add_expectation(suite=suite, expectation=expectation)
```
billdirks, line 379: What is expectation's id after this? Is it the new id or the old one? I think this could get confusing if we are changing the ids. I think if we require no ids on add, it is ok to leave this unchanged and ask the user to retrieve it to get the new id. With this model, the `id` is always meaningful and `None` indicates an suiteless expectation.

tests/data_context/store/test_expectations_store.py
billdirks: I wrote a long blurb in slack. My current thought is to `raise` here. I think `copy` should likely not copy `ids` since the use case is to move the expectation to a new suite? We could parameterize `copy` so one can copy ids if they want to by explicit asking for it.

great_expectations/expectations/expectation.py
joshua-stauffer: thanks!

tests/data_context/store/test_expectations_store.py
joshua-stauffer: `ExpectationSuite.add()` raises if an `Expectation` has an ID, and the `ExpectationsStore` just overwrites. If the inconsistent behavior is weird I can update. My reasoning was that the user should always get an error from the `ExpectationSuite`, and I didn't want to need to try to catch an error from the store as well.

joshua-stauffer: i've implemented `copy` in #9155, and it doesn't copy IDs - if its okay, it would be easiest to leave that change for that PR

tests/data_context/store/test_expectations_store.py
joshua-stauffer: the store is unconcerned about IDs on add - it will always either set them to `None` (cloud mode) or overwrite them with a new `UUID`. The `ExpectationSuite` uses this method, and raises an error if an `Expectation` is passed to `add` with an ID.

tests/core/test_expectation_suite.py
joshua-stauffer: https://greatexpectationslabs.slack.com/archives/C04ST5TSR71/p1703260461204929?thread_ts=1703259519.730379&cid=C04ST5TSR71

joshua-stauffer: thanks - I've updated across the board

billdirks (approved):
LGTM, thanks!

PR #9130, opened 2023-12-19
[MAINTENANCE] persist batch configs

tyler-hoffman:
Adds functionality to add a batch config via the datasource store.

This will just be used when adding new batch configs. Still needed: update and delete (TBD in future PRs)


great_expectations/data_context/store/datasource_store.py
```diff
@@ -320,3 +322,30 @@ def _determine_datasource_key(self, datasource_name: str) -> DataContextVariable
...
+
+        loaded_asset = loaded_datasource.get_asset(data_asset.name)
+        batch_config_names = {bc.name for bc in loaded_asset.batch_configs}
+
+        if batch_config.name in batch_config_names:
+            raise ValueError(
+                f'"{batch_config.name}" already exists (all existing batch_config names are {", ".join(batch_config_names)})'
+            )
+
+        # This MUST be reassigned for it to be picked up during serialization
```
tyler-hoffman: The alternative here is to use `_update__fields_set__on_truthyness(self, _BATCH_CONFIGS_KEY)` during serialization, but that feels like more of a smell to me.

Kilo59: I think you should be able to do this 👇  in order for this field to show up in the serialization with our default of `exclude_unset=True`
```python
self.__fields_set__.add("batch_configs")
```
https://github.com/great-expectations/great_expectations/blob/cb05dbb71f2a968550bed84f1adb2a31956d45bf/great_expectations/datasource/fluent/fluent_base_model.py#L57-L60

tyler-hoffman: Thanks! For anyone who stumbles across this, both options work for this PR, but when we get to the corresponding delete method, we'll probably want to be modifying `__fields_set__` directly, so I'm moving this to follow that pattern.

Kilo59 (approved):

PR #9137, opened 2023-12-19
[DOCS] Fix doc test to use `create_temp_table=True` with QueryAsset

Kilo59:
Needed to unblock the `0.18.6` release


rreinoldsc:
unrelated to the bugfix for BB-16

rreinoldsc (approved):

PR #9139, opened 2023-12-19
[FEATURE] SuiteFactory

joshua-stauffer:


tests/core/test_suite_factory.py
```diff
@@ -0,0 +1,196 @@
...
+    ):
+        factory.delete(suite=suite)
+
+    # Assert
+    store.remove_key.assert_not_called()
+
+
+@pytest.mark.filesystem
+def test_suite_factory_is_initialized_with_context_filesystem(empty_data_context):
+    assert isinstance(empty_data_context.suites, SuiteFactory)
```
tyler-hoffman, line 142: tests where the assert basically just confirms what mypy should confirm feel off. Is this just ensuring that no exception is raised?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -309,6 +310,15 @@ def __init__(self, runtime_environment: Optional[dict] = None) -> None:
...
         self._sources: _SourceFactories = _SourceFactories(self)
 
+        self._suites: Union[SuiteFactory, None]
+        if self.stores.get(self.expectations_store_name):
+            self._suites = SuiteFactory(
+                store=self.expectations_store,
+                include_rendered_content=self._determine_if_expectation_suite_include_rendered_content(),
+            )
+        else:
+            self._suites = None
```
tyler-hoffman, line 320: Is this not an error state?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -543,6 +553,14 @@ def stores(self) -> dict:
     def datasource_store(self) -> DatasourceStore:
         return self._datasource_store
 
+    @property
+    def suites(self) -> SuiteFactory:
+        if not self._suites:
```
tyler-hoffman, line 558: Unless there's a use case I don't know about, i think this confirms my suspicion that we should just raise from the constructor, or just ensure we can always instantiate the factory.

great_expectations/data_context/store/expectations_store.py
```diff
@@ -244,17 +244,28 @@ def get_key(
         self, suite: ExpectationSuite
     ) -> GXCloudIdentifier | ExpectationSuiteIdentifier:
         """Given an ExpectationSuite, build the correct key for use in the ExpectationsStore."""
+        return self._get_key(name=suite.name, id=suite.ge_cloud_id)
+
+    def get_key_by_name(
+        self, name: str
+    ) -> GXCloudIdentifier | ExpectationSuiteIdentifier:
+        """Given a name, obtain the correct key."""
+        # Workaround to allow callers to obtain a key from name without needing to be
```
tyler-hoffman: I think this `workaround` comment should be in the docstring if the method itself is a workaround, but I think a bit more explanation would help. How does `get_key` require needing to aware of the StoreBackend impl?

great_expectations/data_context/store/expectations_store.py
```diff
@@ -244,17 +244,28 @@ def get_key(
         self, suite: ExpectationSuite
     ) -> GXCloudIdentifier | ExpectationSuiteIdentifier:
         """Given an ExpectationSuite, build the correct key for use in the ExpectationsStore."""
+        return self._get_key(name=suite.name, id=suite.ge_cloud_id)
+
+    def get_key_by_name(
```
tyler-hoffman: It's not immediately clear to me the use case of this vs `get_key`. What does it mean if we are in cloud mode and you call each of these methods, getting back one key with an id and one without?

tyler-hoffman (changes requested):
The tests are really nice here!

Two things that I think could be improved:
* Raise early if we can't make a suite factory (or preferably just ensure we can always create it)
* I don't think the `get_key` vs `get_key_by_name` methods are totally clear in terms of which to use or why to prefer one. Do we really need both? If so, can we can a bit more description in the docstrings?

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: thanks for pointing this out - i've updated the signature of `get_key` in order to remove `get_key_by_name` in [d9fa2cb](https://github.com/great-expectations/great_expectations/pull/9139/commits/d9fa2cbbcca96ec608db33cafea99dc3291f8c82).

tests/core/test_suite_factory.py
joshua-stauffer: there is some initialization logic for the SuiteFactory, so I think this is worth having under test.

great_expectations/data_context/data_context/abstract_data_context.py
joshua-stauffer: Unfortunately there's a set of tests which use a context without an `ExpectationsStore` - I'm not sure that's intended behavior, but would suggest scoping to another ticket if we decide to change it.

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: thanks! i removed this method in https://github.com/great-expectations/great_expectations/pull/9139/commits/d9fa2cbbcca96ec608db33cafea99dc3291f8c82

tests/core/test_suite_factory.py
tyler-hoffman: Yeah, that's fair. My comment there was contingent on us ensuring that `_suites` was never `None`

great_expectations/data_context/data_context/abstract_data_context.py
tyler-hoffman: I added https://greatexpectations.atlassian.net/browse/V1-120 to track it. If you have more context, do you mind updating the ticket? My biggest question is if it's a lot of tests. I think in v1, we're likely safe saying there's no good use case for that

joshua-stauffer: thanks!

great_expectations/data_context/store/expectations_store.py
tyler-hoffman: Thanks!

tyler-hoffman (approved):
Looks great!

PR #9140, opened 2023-12-19
[DOCS] `how_to_connect_to_postgresql_data` uses `create_temp_table=True`

Kilo59:
Needed to unblock the release pipfline.

`QueryAsset` does not work without `create_temp_table=True` (the default was changed as part of
- #9116


wookasz (changes requested):
Can you provide context in the PR description on why this change is necessary?

rreinoldsc (approved):

wookasz (approved):

PR #9145, opened 2023-12-20
[CONTRIB] ExpectColumnValuesToBePresentInOtherTable - Referential Integrity (0.18 branch)

Shinnnyshinshin:
# What does this PR do? 
* update to #9117 so that we are merging into the `0.18.x` branch for release

* Adds Core Logic for `ExpectColumnValuesToBePresentInOtherTable` which tests referential integrity. 
* Added Renderers: 

### Success
Screenshot 2023-12-19 at 5 22 43 PM

### Failure
Screenshot 2023-12-19 at 5 22 06 PM

### DIFF
--- a/contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
+++ b/contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
@@ -111,7 +111,7 @@ class ExpectColumnValuesToBePresentInAnotherTable(QueryExpectation):
 
         template_dict = configuration.kwargs.get("template_dict")
 
-        template_str = "$foreign_key_column values exist in table $foreign_table 's $primary_key_column_in_foreign_table column."
+        template_str = "All values in column $foreign_key_column are present in column $primary_key_column_in_foreign_table of  table $foreign_table. (Observed Value is the number of missing values)."
 
         params = {
             "foreign_key_column": template_dict["foreign_key_column"],
--- a/contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
+++ b/contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
@@ -1,0 +1,188 @@
+from typing import Optional, Union
+
+from great_expectations.execution_engine import ExecutionEngine
+from great_expectations.expectations.expectation import (
+    ExpectationValidationResult,
+    QueryExpectation,
+)
+from great_expectations.expectations.expectation_configuration import (
+    ExpectationConfiguration,
+)
+
+
+class ExpectColumnValuesToBePresentInAnotherTable(QueryExpectation):
+    """Expect the values in a column to be present in another table.
+
+     This is an Expectation that allows for the validation of referential integrity, that a foreign key exists in another table.
+
+     In the following example, order table has a foreign key to customer table, and referential integrity is preserved, because all of
+     the values of CUSTOMER_ID in order_table_1 are present in the CUSTOMER_ID column of customer_table.
+
+    "order_table_1": {
+         "ORDER_ID": ["aaa", "bbb", "ccc"],
+         "CUSTOMER_ID": [1, 1, 3],
+     }
+     "customer_table": {
+         "CUSTOMER_ID": [1, 2, 3],
+
+     }
+
+     However, in the second example, referential integrity is not preserved, because there are two values (4 and 5) in the CUSTOMER_ID
+     column of order_table_2 that are not present in the CUSTOMER_ID column of customer_table.
+
+     "order_table_2": {
+         "ORDER_ID": ["ddd", "eee", "fff"],
+         "CUSTOMER_ID": [1, 4, 5],
+     }
+     "customer_table": {
+         "CUSTOMER_ID": [1, 2, 3],
+
+     }
+     ExpectColumnValuesToBePresentInAnotherTable will PASS for example 1 and FAIL for example 2.
+
+     Args:
+         template_dict: dict containing the following keys:
+             foreign_key_column: foreign key-column of current table that we want to validate.
+             foreign_table: foreign table.
+             primary_key_column_in_foreign_table: key column for primary key in foreign table.
+    """
+
+    library_metadata = {
+        "maturity": "experimental",
+        "tags": ["table expectation", "multi-table expectation", "query-based"],
+        "contributors": [
+            "@great_expectations",
+        ],
+        "requirements": [
+            "sqlalchemy",
+            "snowflake-sqlalchemy",
+            "snowflake-connector-python",
+        ],
+        "has_full_test_suite": False,
+        "manually_reviewed_code": True,
+    }
+
+    metric_dependencies = ("query.template_values",)
+    template_dict: dict
+    query = """
+        SELECT count(1) FROM (
+        SELECT a.{foreign_key_column}
+        FROM {active_batch} a
+        LEFT JOIN {foreign_table} b
+            ON a.{foreign_key_column} = b.{primary_key_column_in_foreign_table}
+        WHERE b.{primary_key_column_in_foreign_table} is NULL)
+        """
+    success_keys = (
+        "template_dict",
+        "query",
+    )
+    domain_keys = ("query", "batch_id", "row_condition", "condition_parser")
+
+    default_kwarg_values = {
+        "catch_exceptions": False,
+        "meta": None,
+        "query": query,
+    }
+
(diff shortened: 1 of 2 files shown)


contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,227 @@
...
+        result: Optional[ExpectationValidationResult] = None,
+        runtime_configuration: Optional[dict] = None,
+    ) -> List[RenderedStringTemplateContent]:
+        runtime_configuration = runtime_configuration or {}
+        styling = runtime_configuration.get("styling")
+
+        template_dict = configuration.kwargs.get("template_dict")
+
+        template_str = "No values in the column $foreign_key_column are expected to be missing in table $foreign_table 's $primary_key_column_in_foreign_table column."
+
```
Shinnnyshinshin: This is the current template string. Would appreciate feedback here.

JennyTee: Just throwing an option out: `All values in column $foreign_key_column are present in column $primary_key_column_in_foreign_table of  table $foreign_table.`

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,227 @@
...
+         "CUSTOMER_ID": [1, 2, 3],
+
+     }
+     ExpectColumnValuesToBePresentInAnotherTable will PASS for example 1 and FAIL for example 2.
+
+     Args:
+         template_dict: dict containing the following keys:
+             foreign_key_column: foreign key-column of current table that we want to validate.
+             foreign_table: foreign table.
+             primary_key_column_in_foreign_table: key column for primary key in foreign table.
```
anthonyburdi, line 60: Love this docstring, but took a second to map from the example to the params. Would it make sense to add an example of the `template_dict` for each of the two examples?

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,227 @@
...
+            "snowflake-connector-python",
+        ],
+        "has_full_test_suite": False,
+        "manually_reviewed_code": True,
+    }
+
+    metric_dependencies = ("query.template_values",)
+    template_dict: dict
+    query = """
+        SELECT count(1) FROM (
```
JennyTee: supertinynit (feel free to ignore): if we are doing all caps for keywords in this statement, technically `COUNT` (in this line) and `IS` (line 76) would also be capitalized

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,227 @@
...
+        configuration: Optional[ExpectationConfiguration] = None,
+        result: Optional[ExpectationValidationResult] = None,
+        runtime_configuration: Optional[dict] = None,
+    ) -> List[RenderedStringTemplateContent]:
+        runtime_configuration = runtime_configuration or {}
+        styling = runtime_configuration.get("styling")
+
+        template_dict = configuration.kwargs.get("template_dict")
+
+        template_str = "No values in the column $foreign_key_column are expected to be missing in table $foreign_table 's $primary_key_column_in_foreign_table column."
```
anthonyburdi: What do you think about flipping so that it is asserting a positive? Something like:
```suggestion
        template_str = "$foreign_key_column values exist in table $foreign_table 's $primary_key_column_in_foreign_table column."
```
Also, what do you think about using the column first? Looks like it might be more consistent with some of the other core expectations, here is a sample:
image

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
anthonyburdi: Posted a suggestion on the line, let me know what you think!

anthonyburdi (changes requested):
This is really looking great! Just a few small suggestions. Let me know what you think!

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,227 @@
...
+            raise TypeError("template_dict must be supplied as a dict")
+        if not all(
+            [
+                "foreign_key_column" in template_dict,
+                "foreign_table" in template_dict,
+                "primary_key_column_in_foreign_table" in template_dict,
+            ]
+        ):
+            raise KeyError(
+                "The following keys have to be in the template dict: id, foreign_table, foreign_key_column"
```
JennyTee: Should `id` in this error message be `primary_key_column_in_foreign_table` instead?

Also, nit - as a user, I would find it more helpful to know what is missing...e.g.,


```
        if not all(
            [
                "foreign_key_column" in template_dict,
                "foreign_table" in template_dict,
                "primary_key_column_in_foreign_table" in template_dict,
            ]
        ):
            raise KeyError(
                f"The following keys are missing from the template dict: 
{'foreign_key_column ' if  not ('foreign_key_column' in template_dict) else ''}
{'foreign_table ' if  not ('foreign_table' in template_dict) else ''}
{'primary_key_column_in_foreign_table ' if  not ('primary_key_column_in_foreign_table' in template_dict) else ''}"
```

(there is prob a better way to write that and make it readable but that is the first thing that came to mind)

JennyTee (approved):

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: Thank you so much for catching this 🙇🏼‍♂️

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: The `template_dict` would be the same for the two examples, since the Expectation itself would be run on either `order_table_1` or `order_table_2` (our DataAssets).  I do agree that an example would be very helpful in reducing cognitive load. l add an example to the end of the docstring.

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: This is a great suggestion. Thank you so much @JennyTee

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: Screenshot 2023-12-20 at 12 12 13 PM

I'm not sure how this will fit in with "Observed Value" as `0` when the Expectation passes and non-`0` when the Expectation fails. This is where the original "no values" comes into play, but maybe the results should be self explanatory? @anthonyburdi

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: I tried to display the column value, but unfortunately the `QueryExpectation` is a category of its own, so I think it gets grouped with the Table Expectations

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
anthonyburdi: Good point. How about adding a parenthetical afterward, something like "(observed value is the number of missing items)"?

Also I like Jenny's suggested text better. Mine was trying to match other expectations that show column name first but it's clearer how @JennyTee wrote it.

Shinnnyshinshin: Thank you @anthonyburdi. I'll update and ask the # product channel for some feedback before merging :)

anthonyburdi (approved):
Thanks for making those updates!

PR #9147, opened 2023-12-20
[MAINTENANCE] Persist BatchConfigs when adding them

tyler-hoffman:
**Main work**
* Adds the necessary persistence steps when calling `DataAsset::add_batch_config`
* Backfills tests around that method.


great_expectations/datasource/fluent/interfaces.py
```diff
@@ -226,14 +227,36 @@ def batch_request_options(self) -> tuple[str, ...]:
...
+
         batch_config = BatchConfig(name=name)
+        batch_config._data_asset = self
+        self.__fields_set__.add("batch_configs")
+        if self.datasource.data_context:
+            store = self.datasource.data_context._datasource_store
+            key = DataContextVariableKey(resource_name=self.datasource.name)
+            if store.has_key(key=key):
+                batch_config = store.add_batch_config(batch_config)
+                batch_config._data_asset = self
```
joshua-stauffer: this doesn't feel quite right on a couple levels.  my current thinking is that this logic live in a `Datasource` method, which is passed to the asset when its instantiated. If the asset doesn't have a `Datasource`, it won't have a callback and we can error here. The `Datasource` has a store and can check to see if its been persisted yet.

tyler-hoffman: Coming back to this - my previous comment was a misreading of what you said, but I think it's the direction I want to propose we go with.

A couple thoughts on injecting the callback:
* If we inject in the callback, we'll end up doing that for each crud method, so it's a bit of wiring
* The use of PrivateAttrs feels like a place where things can easily go wrong (i.e. it's easy to instantiate a datasource and not inject everything down to the asset)
* It feels like overhead to me when we already know about and are coupled to the parent model

Given the above, it seems like since we've done the work of wiring up the reference back to Datasource, we should just lean on that. But I do agree that stuff like writing `self.datasource.data_context._datasource_store` is terrible. I think what we should do is call `self.datasource.add_batch_config`. This seems simple, and reasonable decoupled.

Playing out what that would look like:
* we'd have an `add_batch_config`, `update_batch_config`, and `delete_batch_config`,  method on the datasource.
* We'd probably add an `is_persisted` method to datasources as well, since it isn't using the callback pattern expectations are using.
* And then within the asset, we'd just say something like:
```
if self.datasource.is_persisted():
   batch_config = self.datasource.add_batch_config(batch_config)
```

Thoughts?

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -488,6 +510,19 @@ def _execution_engine_type(self) -> Type[_ExecutionEngineT]:
...
         return self.execution_engine_override or self.execution_engine_type
 
+    def add_batch_config(self, batch_config: BatchConfig) -> BatchConfig:
+        if not self.data_context:
+            raise DataContextError("Datasource is not attached to a DataContext")
+        return self.data_context.datasource_store.add_batch_config(batch_config)
+
+    def is_persisted(self) -> bool:
+        if self._data_context:
+            store = self._data_context._datasource_store
```
billdirks, line 520: It would be preferable if the datasouce knew it's store and didn't have to reach into the context. That might mean hooking it up at instantiation time like we do to associated the asset with the batch config. However, this work can be done in a subsequent PR.

billdirks (approved):
LGTM, thanks!
Left a comment which we've already discussed in slack.

great_expectations/datasource/fluent/interfaces.py
tyler-hoffman: Strong agree. Added https://greatexpectations.atlassian.net/browse/V1-115 to track this.

PR #9148, opened 2023-12-20
[BUGFIX] Fix QueryAsset with `create_temp_table=False`

Kilo59:
Fix the metric calculation and lookup error that occurs when `create_temp_table=False` with a fluent `QueryAsset`.

~Fix is applied inside `SQLAlchemyExecutionEngine.get_batch_data_and_markers()`.~
Fix is applied by building a `RuntimeQueryBatchSpec` inside of `QueryAsset.get_batch_list_from_batch_request()`
QueryAsset batch spec needs to follow the same path as a `RuntimeQueryBatchSpec` instead of a `SqlAlchemyDatasourceBatchSpec`.

`RuntimeQueryBatchSpec` is `IDDict` & `BatchSpec`.
It's a dictionary with a `query` and `batch_data` convenience properties.

https://github.com/great-expectations/great_expectations/blob/93ffeaf96f26928ec351821e8c5496e8c3d8e9f6/great_expectations/core/batch_spec.py#L202-L215

`SqlAlchemyDatasourceBatchSpec` has `limt`, `schema` and `batch_data`convenience properties.

### DIFF
--- a/great_expectations/execution_engine/sqlalchemy_execution_engine.py
+++ b/great_expectations/execution_engine/sqlalchemy_execution_engine.py
@@ -1360,11 +1360,15 @@ class SqlAlchemyExecutionEngine(ExecutionEngine):
         create_temp_table: bool = batch_spec.get(
             "create_temp_table", self._create_temp_table
         )
-        query: Optional[str] = batch_spec.get("query")
+        # TODO: clean this up
+        # fix this when building the batch spec or stop doing instance checks at all
+        # and check for the presence of the query key
+        query: Optional[str] = batch_spec.get(
+            "query", getattr(batch_spec, "query", None)
+        )
         if isinstance(batch_spec, RuntimeQueryBatchSpec) or query:
             # query != None is already checked when RuntimeQueryBatchSpec is instantiated
-            query: str = query or batch_spec.query
-
+            assert query
             batch_data = SqlAlchemyBatchData(
                 execution_engine=self,
                 query=query,
--- a/great_expectations/execution_engine/sqlalchemy_execution_engine.py
+++ b/great_expectations/execution_engine/sqlalchemy_execution_engine.py
@@ -1363,9 +1363,7 @@ class SqlAlchemyExecutionEngine(ExecutionEngine):
         # TODO: clean this up
         # fix this when building the batch spec or stop doing instance checks at all
         # and check for the presence of the query key
-        query: Optional[str] = batch_spec.get(
-            "query", getattr(batch_spec, "query", None)
-        )
+        query: Optional[str] = getattr(batch_spec, "query", batch_spec.get("query"))
         if isinstance(batch_spec, RuntimeQueryBatchSpec) or query:
             # query != None is already checked when RuntimeQueryBatchSpec is instantiated
             assert query
--- a/great_expectations/execution_engine/sqlalchemy_execution_engine.py
+++ b/great_expectations/execution_engine/sqlalchemy_execution_engine.py
@@ -1361,7 +1361,7 @@ class SqlAlchemyExecutionEngine(ExecutionEngine):
             "create_temp_table", self._create_temp_table
         )
         query: Optional[str] = batch_spec.get("query")
-        if isinstance(batch_spec, RuntimeQueryBatchSpec) or batch_spec.get("query"):
+        if isinstance(batch_spec, RuntimeQueryBatchSpec) or query:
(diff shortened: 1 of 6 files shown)


tests/integration/test_script_runner.py
```diff
@@ -545,14 +544,12 @@ def pytest_parsed_arguments(request):
 
 @flaky(rerun_filter=delay_rerun, max_runs=3, min_passes=1)
 @pytest.mark.parametrize("integration_test_fixture", docs_test_matrix, ids=idfn)
-@pytest.mark.skipif(sys.version_info < (3, 7), reason="requires Python3.7")
```
Kilo59, line 548: This is a useless skip.
Our min python is now `3.8`.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1360,10 +1360,15 @@ def get_batch_data_and_markers(
...
             "create_temp_table", self._create_temp_table
         )
-        if isinstance(batch_spec, RuntimeQueryBatchSpec):
+        # TODO: clean this up
+        # fix this when building the batch spec or stop doing instance checks at all
+        # and check for the presence of the query key
+        query: Optional[str] = batch_spec.get(
+            "query", getattr(batch_spec, "query", None)
+        )
+        if isinstance(batch_spec, RuntimeQueryBatchSpec) or query:
```
Kilo59: I wouldn't be surprised if this was leading to other problems with `QueryAsset` even with `temp_table=True` because of how the `SqlAlchemyBatchData` was being built (not using `query`).

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1360,10 +1360,15 @@ def get_batch_data_and_markers(
         create_temp_table: bool = batch_spec.get(
             "create_temp_table", self._create_temp_table
         )
-        if isinstance(batch_spec, RuntimeQueryBatchSpec):
+        # TODO: clean this up
+        # fix this when building the batch spec or stop doing instance checks at all
```
tyler-hoffman: TL;DR: I think it's worth some minor cleanup now.

It took me a minute to track down what's going on and why. IIUC:
* `RuntimeQueryBatchSpec` has a property called `query` (hence `getattr(batch_spec, "query", None)`).
* There can be other BatchSpecs such that `batch_spec["query"] exists (hence `batch_spec.get("query")` would exist

If that understanding is right, I think this should be restructured to better reflect that. e.g. branch for RuntimeQueryBatchSpec, else branch if we access query using dict syntax.

Also the current way this is written, `if isinstance(batch_spec, RuntimeQueryBatchSpec) or query:` is equivalent to `if query:`, which adds to the difficulty parsing this.

tyler-hoffman (changes requested):
The code around the actual change took me a minute to get my head around. Both the getattr nested in a get (and why the difference), and also the if condition. I think it's worth making this a bit more self-evident.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
Kilo59: I agree with everything you said.
I have no plans to merge this as is. I would have marked this as a draft except I need the docs tests to run.

This is a draft showing the minimum change needed to fix the issue.
I plan on making the changes described in the TODO comments and talking with @billdirks before merging.

Kilo59:
## WIP

~This is a Work-In-Progress.
The current fix works, but the current implementation will be updated for clarity and ease of maintenance.~

tests/integration/docusaurus/connecting_to_your_data/fluent_datasources/how_to_connect_to_postgresql_data.py
```diff
@@ -65,7 +65,6 @@
 datasource = context.sources.add_postgres(
     name=datasource_name,
     connection_string=my_connection_string,
-    create_temp_table=True,
```
Kilo59, line 68: Reverting these test to what they were before

billdirks (approved):
LGTM.

You may want to check where other `isinstance` checks on BatchSpecs are being done just to make sure there isn't any other consequences of this change.

Kilo59:
@billdirks just confirmed, there are no additional `RuntimeQueryBatchSpec` or `SqlAlchemyDatasourceBatchSpec` instance checks.

tyler-hoffman (approved):

PR #9150, opened 2023-12-20
[BUGFIX] v1 - `QueryAsset` works with `create_temp_table=False`

Kilo59:
Applying the `0.18.x` fix to `develop.
- https://github.com/great-expectations/great_expectations/pull/9148

Discovered that the `0.18.x`/`0.18.6` temp table QueryAsset fix caused splitter logic not to be applied to the queries.
Fixed this by re-compiling the query after the splitter logic was applied in the `selectable`.

### DIFF
--- a/great_expectations/execution_engine/sqlalchemy_batch_data.py
+++ b/great_expectations/execution_engine/sqlalchemy_batch_data.py
@@ -172,7 +172,7 @@ class SqlAlchemyBatchData(BatchData):
                 )
             )
         elif query:
-            self._selectable = self._generate_selectable_from_query(
+            self._selectable = self._generate_selectable_from_query(  # type: ignore[call-overload] # https://github.com/python/mypy/issues/14764
                 query, dialect, create_temp_table, temp_table_schema_name
             )
         else:
--- a/great_expectations/execution_engine/sqlalchemy_batch_data.py
+++ b/great_expectations/execution_engine/sqlalchemy_batch_data.py
@@ -357,7 +357,7 @@ class SqlAlchemyBatchData(BatchData):
     def _generate_selectable_from_query(
         self,
         query: str,
-        dialect: GXSqlDialect | None,
+        dialect: GXSqlDialect,
         create_temp_table: Literal[False],
         temp_table_schema_name: Optional[str] = ...,
     ) -> sqlalchemy.TextClause:
@@ -366,7 +366,7 @@ class SqlAlchemyBatchData(BatchData):
     def _generate_selectable_from_query(
         self,
         query: str,
-        dialect: GXSqlDialect | None,
+        dialect: GXSqlDialect,
         create_temp_table: bool,
         temp_table_schema_name: Optional[str] = None,
     ) -> sqlalchemy.Table | sqlalchemy.TextClause:
--- a/great_expectations/execution_engine/sqlalchemy_batch_data.py
+++ b/great_expectations/execution_engine/sqlalchemy_batch_data.py
@@ -1,18 +1,16 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, Literal, Optional, Tuple, overload
+from typing import Literal, Optional, Tuple, overload
 
 from great_expectations.compatibility import sqlalchemy
+from great_expectations.compatibility.sqlalchemy import Selectable
 from great_expectations.compatibility.sqlalchemy import (
     sqlalchemy as sa,
 )
 from great_expectations.core.batch import BatchData
 from great_expectations.execution_engine.sqlalchemy_dialect import GXSqlDialect
 from great_expectations.util import generate_temporary_table_name
-
-if TYPE_CHECKING:
-    from great_expectations.compatibility.sqlalchemy import Selectable
 
 logger = logging.getLogger(__name__)
 
@@ -86,7 +84,7 @@ class SqlAlchemyBatchData(BatchData):
         schema_name: Optional[str] = None,
         table_name: Optional[str] = None,
         # Option 2
-        query: Optional[str] = None,
+        query: str | Selectable | None = None,
         # Option 3
         selectable: Optional[Selectable] = None,
         create_temp_table: bool = True,
@@ -146,11 +144,19 @@ class SqlAlchemyBatchData(BatchData):
         self._source_table_name = source_table_name
         self._source_schema_name = source_schema_name
 
-        if sum(bool(x) for x in [table_name, query, selectable is not None]) != 1:
+        # Selectable does not implement __bool__ so we need to check for None explicitly
+        query_is_present: bool = query is not None
+
+        if (
+            sum(bool(x) for x in [table_name, query_is_present, selectable is not None])
+            != 1
+        ):
             raise ValueError(
                 "Exactly one of table_name, query, or selectable must be specified"
             )
-        elif (query and schema_name) or (selectable is not None and schema_name):
+        elif (query_is_present and schema_name) or (
+            selectable is not None and schema_name
+        ):
             raise ValueError(
                 "schema_name can only be used with table_name. Use temp_table_schema_name to provide a target schema for creating a temporary table."
             )
@@ -173,7 +179,7 @@ class SqlAlchemyBatchData(BatchData):
                     schema_name=schema_name,
                 )
             )
-        elif query:
+        elif query_is_present:
             # NOTE: we could go down this path for query but there's logic in here that relies on `Selectable`
             # implementing __bool__ which it doesn't.
             # This should be easy to deal with though if it fixes the other query issues.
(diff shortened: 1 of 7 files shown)


Kilo59:
Seeing this failure

```
tests/validator/test_v1_validator.py::test_validate_expectation_with_batch_asset_options[test_backends0]
```
```
>       assert result.success
E       assert False
E        +  where False = {\n  "success": false,\n  "expectation_config": {\n    "expectation_type": "expect_column_values_to_be_in_set",\n    "kwar..."exception_info": {\n    "raised_exception": false,\n    "exception_traceback": null,\n    "exception_message": null\n  }\n}.success
```

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1360,20 +1360,21 @@ def get_batch_data_and_markers(
...
+        # NOTE: what's being checked here is the presence of a `query` attribute, we could check this directly
+        # instead of doing an instance check
         if isinstance(batch_spec, RuntimeQueryBatchSpec):
             # query != None is already checked when RuntimeQueryBatchSpec is instantiated
-            query: str = batch_spec.query
 
             batch_data = SqlAlchemyBatchData(
                 execution_engine=self,
-                query=query,
+                selectable=selectable,
```
Kilo59: This is what's new (compared to the `0.18.x` branch).
`_build_selectable_from_batch_spec` is what adds the splitter logic to the query.

great_expectations/execution_engine/sqlalchemy_batch_data.py
```diff
@@ -173,7 +179,10 @@ def __init__(  # noqa: PLR0913
                     schema_name=schema_name,
                 )
             )
-        elif query:
+        elif query_is_present:
+            # NOTE: we could go down this path for query but there's logic in here that relies on `Selectable`
+            # implementing __bool__ which it doesn't.
+            # This should be easy to deal with though if it fixes the other query issues.
```
Kilo59, line 185:
```suggestion
```

great_expectations/execution_engine/sqlalchemy_batch_data.py
```diff
@@ -342,25 +348,52 @@ def _generate_selectable_from_schema_name_and_table_name(
...
+            sqlalchemy.Table: SqlAlchemy Table that is Selectable or a TextClause.
         """
         if not create_temp_table:
+            if isinstance(query, Selectable):
+                compiled_query = query.compile(
+                    dialect=self.sql_engine_dialect,
+                    compile_kwargs={"literal_binds": True},
+                )
+                text_query = str(compiled_query)
+                return sa.text(text_query)
```
Kilo59: I think this should be done in `get_batch_data_and_markers()` before creating the `SqlAlchemyBatchData()` object.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1360,20 +1363,27 @@ def get_batch_data_and_markers(
...
         if isinstance(batch_spec, RuntimeQueryBatchSpec):
             # query != None is already checked when RuntimeQueryBatchSpec is instantiated
-            query: str = batch_spec.query
-
+            # re-compile the query to include any new parameters
+            compiled_query = selectable.compile(
+                dialect=self.engine.dialect,
+                compile_kwargs={"literal_binds": True},
+            )
+            query_str = str(compiled_query)
```
Kilo59, line 1378: @billdirks re-compiling the query.

Can this be avoided or done better?

great_expectations/execution_engine/sqlalchemy_batch_data.py
Kilo59: Did it

great_expectations/execution_engine/sqlalchemy_execution_engine.py
Kilo59: reverted this because it caused another problem

Kilo59:
## TODO

- [x] turn @billdirks error repro script into a new test

```python
import pathlib

db_path = pathlib.Path.cwd() / "tests" / "test_sets" / "test_cases_for_sql_data_connector.db"
assert db_path.exists()

context = gx.get_context(mode="ephemeral")
datasource = context.sources.add_sqlite(
    name='sqlite_datasource',
    # This is a file in the repo. You will need to update this path
    connection_string=f"sqlite:////{db_path.absolute()}",
    create_temp_table=True # try this with True and False
)
asset = datasource.add_query_asset(name="trip_asset_split_by_event_type",query="SELECT * FROM table_partitioned_by_date_column__A",).add_splitter_column_value("event_type")
batch_request = asset.build_batch_request({"event_type": "start"})
validator = context.get_validator(batch_request=batch_request)
# All rows returned by head have the start event_type.
result = validator.execution_engine.batch_manager.active_batch.head(n_rows=50)
unique_event_types = set(result.data["event_type"].unique())
print(f"{unique_event_types=}")
assert unique_event_types == {"start"}
```

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1315,7 +1315,10 @@ def _subselectable(self, batch_spec: BatchSpec) -> sqlalchemy.Selectable:
             if not isinstance(query, str):
                 raise ValueError(f"SQL query should be a str but got {query}")
             # Query is a valid SELECT query that begins with r"\w+select\w"
-            selectable = sa.select(sa.text(query.lstrip()[6:].lstrip())).subquery()
+            # TODO:? strip `;` from the end of the query??
```
Kilo59, line 1318:
```suggestion
```

great_expectations/execution_engine/sqlalchemy_batch_data.py
```diff
@@ -174,7 +172,7 @@ def __init__(  # noqa: PLR0913
                 )
             )
         elif query:
-            self._selectable = self._generate_selectable_from_query(
+            self._selectable = self._generate_selectable_from_query(  # type: ignore[call-overload] # https://github.com/python/mypy/issues/14764
```
Kilo59, line 175: The code in these `_generate_selectable_from_*` methods are almost identical.
They could be simplified for v1.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
billdirks: Yes, this is ugly but I am ok with this at the moment. I think it would be better to understand and unify these 2 branches but that could be a much bigger change and it sounds like we need to get this fixed ASAP. Could you file a ticket to unify these branches? You can put it in [go/v1-jira](https://www.golinks.io/v1-jira?trackSource=github)

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -1360,20 +1362,27 @@ def get_batch_data_and_markers(
         create_temp_table: bool = batch_spec.get(
             "create_temp_table", self._create_temp_table
         )
+        # this is where splitter components are added to the selectable
```
billdirks, line 1365: I think the return type of this function can be updated to: `Tuple[SqlAlchemyBatchData | None, BatchMarkers]`

tests/datasource/fluent/integration/conftest.py
```diff
@@ -100,22 +100,28 @@ def pandas_data(
...
-        "..",
-        "..",
-        "test_sets",
-        "taxi_yellow_tripdata_samples",
-        "sqlite",
-        db_filename,
-    )
-    db_file = pathlib.Path(__file__).parent.joinpath(relative_path).resolve(strict=True)
+    db_file_path = pathlib.Path(db_filename)
+    if db_file_path.is_absolute():
```
billdirks: Why do we need this new branch?

tests/datasource/fluent/integration/test_integration_datasource.py
```diff
@@ -100,27 +100,47 @@ def test_batch_head(
...
 @pytest.mark.sqlite
-def test_sql_query_data_asset(empty_data_context):
-    context = empty_data_context
-    datasource = sqlite_datasource(context, "yellow_tripdata.db")
-    passenger_count_value = 5
-    asset = (
-        datasource.add_query_asset(
-            name="query_asset",
-            query=f"   SELECT * from yellow_tripdata_sample_2019_02 WHERE passenger_count = {passenger_count_value}",
+class TestQueryAssets:
```
Kilo59, line 103: We need to test more than just `SQLite`

billdirks (approved):
Thanks for digging in and resolving this!

tests/datasource/fluent/integration/conftest.py
Kilo59: I don't think we do. It was because the dB I wanted was at a different level. But I think I can just include  `../..` in my path without changing this.

PR #9152, opened 2023-12-21
[MAINTENANCE] BatchConfig: use public setter instead of private attr

tyler-hoffman:
**Changes**
* Adds public setter method (not the `@property.setter` way) to set `BatchConfig::data_asset`
* Removes outside assignment to `BatchConfig::_data_asset`

**Why not a normal python setter??**
Looks like pydantic interferes with that. https://github.com/pydantic/pydantic/issues/1577


great_expectations/core/batch_config.py
```diff
@@ -28,6 +28,10 @@ class BatchConfig(pydantic.BaseModel):
     def data_asset(self) -> DataAsset:
         return self._data_asset
 
+    def set_data_asset(self, data_asset: DataAsset) -> None:
```
billdirks, line 31: If we want this to look like a property we could do something like what is done [here](https://github.com/great-expectations/great_expectations/pull/9099/files/beff8dacc4c6ec8851a98eb4ecd74e447ee39d32#diff-6eeeb07deb1d49a0a5f86ddc7becf7e3c7b6ce62420124c020c7489e00ae7ac2R878-R896). I ended up making that class not a pydantic model so didn't need to do this.

billdirks (approved):
LGTM! I left a comment if you want to make `data_asset` look like a settable property.

PR #9177, opened 2024-01-03
[MAINTENANCE] Ensure that `Expectation.save()` impacts local suite state

cdkini:
Calling `Expectation.save()` should ensure that the parent suite is updated (both in-memory and in the persistence layer). We should only touch the expectation that has been saved and leave any other in-memory change alone.


great_expectations/data_context/store/expectations_store.py
```diff
@@ -171,6 +171,9 @@ def update_expectation(
                 break
 
         self.update(key=suite_identifier, value=fetched_suite)
+
+        # Ensure that state of local suite is in sync with cloud backend
+        suite.expectation_configurations = fetched_suite.expectation_configurations
```
cdkini: Need to add this to get `expectation.save()` to work as expected

tyler-hoffman (approved):
the change in the store seems reasonable to me, but curious if @joshua-stauffer has other thoughts

great_expectations/data_context/store/expectations_store.py
billdirks: What issue were you seeing before/what wasn't expected? 
It may be strange that saving on expectation syncs the whole suite. For example, if you've modified other expectations in the suite but haven't saved them it, those changes will get overwritten here.

cdkini: `Expectation.save()` did not update the suite at all - I can modify this logic so we only update the specific expectation within the suite instead of everything

tests/core/test_expectation_suite.py
```diff
@@ -436,6 +436,7 @@ def _test_expectation_can_be_saved_after_added(self, context, expectation):
         assert expectation.column != updated_column_name
         expectation.column = updated_column_name
         expectation.save()
+        assert suite.expectations[0].column == updated_column_name
```
cdkini, line 439: Want to check state of in-memory suite BEFORE the subsequent call to `get_expectation_suite` here

tests/core/test_expectation_suite.py
```diff
@@ -464,10 +465,46 @@ def _test_expectation_can_be_saved_after_update(self, context, expectation):
...
+        expectation = suite.expectations[0]
+        other_expectation = suite.expectations[1]
+        expectation.column = "c"
+        other_expectation.column = "d"
+
+        # Only persist change made to first expectation
+        expectation.save()
+
+        assert suite.expectations[0].column == "c"
+        assert suite.expectations[1].column == "b"
```
joshua-stauffer, line 506: Thanks for adding this test! just wanted to flag that it is actually not testing behavior as desired. `other_expectation.column` should still be `d` here, since the user made that change locally. Better would be to fetch the suite from the backend, and assert that the expectation in the remote suite hasn't been changed. I'll update in a followup PR.

PR #9181, opened 2024-01-03
[BUGFIX] Fix batch config persistence

tyler-hoffman:
**Context**
I previously assumed that the DatasourceStore was the correct place to implement persistence of fluent datasources. That ended up not being the case because FDS does not use the store for FileDataContext. This PR moves away from that pattern and toward the use of `context.update_datasource` on the datasource itself. This is **not** where persistence logic should live, but without rewriting FDS to always leverage stores, I think this is about the best we can do. I think after we refactor stores, we should move this logic to the store.


great_expectations/datasource/fluent/interfaces.py
```diff
@@ -227,6 +226,45 @@ def batch_request_options(self) -> tuple[str, ...]:
             """One needs to implement "batch_request_options" on a DataAsset subclass."""
         )
 
+    def build_batch_request(
```
tyler-hoffman, line 229: These methods are NOT new or modified. I updated the below `add_batch_config` and `delete_batch_config` methods and moved them below the abstract method section that it was mistakenly within.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -250,14 +288,17 @@ def add_batch_config(self, name: str) -> BatchConfig:
         assert isinstance(self.datasource, Datasource)
 
         batch_config = BatchConfig(name=name)
-        batch_config.set_data_asset(self)
+        batch_config._data_asset = self
```
billdirks: Why not use the setter any longer?

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -250,14 +288,17 @@ def add_batch_config(self, name: str) -> BatchConfig:
...
         self.__fields_set__.add("batch_configs")
-        if self.datasource.is_persisted():
-            batch_config = self.datasource.add_batch_config(batch_config)
-            batch_config.set_data_asset(self)
-
-        batch_config.set_data_asset(self)
         self.batch_configs.append(batch_config)
+        if self.datasource.data_context:
+            updated_datasource = self.datasource.save()
```
billdirks: If there are other locally modified data assets will this save them also? Will it also save any local changes to the datasource itself?

If you you might need to fetch the datasource, update this one data asset in the fetched datasource, and save that. 

It might be ok taking the datasource changes because presumably they are necessary for the data asset. I guess there is some danger for disrupting the other assets but this should be all shared state like the connection string.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -276,54 +317,15 @@ def delete_batch_config(self, batch_config: BatchConfig) -> None:
         # Let mypy know that self.datasource is a Datasource (it is currently bound to MetaDatasource)
         assert isinstance(self.datasource, Datasource)
 
-        if self.datasource.is_persisted():
-            self.datasource.delete_batch_config(batch_config)
-
         self.batch_configs.remove(batch_config)
+        if self.datasource.data_context:
+            self.datasource.save()
```
billdirks: If this save fails I think you want to add the `batch_config` back into `self.batch_configs` and re-raise, in case a caller tries to handle the exception.

billdirks (commented):
I'm running off to a meeting but wanted to publish my comments. I will look at the rest in a bit.

great_expectations/datasource/fluent/interfaces.py
billdirks: A similar change might be necessary for delete?

great_expectations/datasource/fluent/interfaces.py
tyler-hoffman: oops. didn't mean to change that. I'll update.

great_expectations/datasource/fluent/interfaces.py
tyler-hoffman: Good call

tests/datasource/fluent/data_asset/test_data_asset.py
```diff
@@ -1,91 +1,80 @@
...
 )
-from great_expectations.data_context.store.datasource_store import DatasourceStore
+from great_expectations.data_context.data_context.cloud_data_context import (
+    CloudDataContext,
+)
 from great_expectations.datasource.fluent.interfaces import DataAsset, Datasource
+from great_expectations.datasource.fluent.pandas_datasource import PandasDatasource
 
-
-@pytest.fixture
```
tyler-hoffman, line 12: Converted all these string fixtures to just be constants for better readability.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -276,53 +317,36 @@ def delete_batch_config(self, batch_config: BatchConfig) -> None:
...
-        raise NotImplementedError(
-            """One must implement "_validate_batch_request" on a DataAsset subclass."""
-        )
+    def update_batch_config_field_set(self) -> None:
+        """Ensure that we have __fields_set__ set correctly for batch_configs to ensure we serialize IFF needed."""
 
-    # End Abstract Methods
+        has_batch_configs = len(self.batch_configs) > 0
+        if "batch_configs" in self.__fields_set__ and not has_batch_configs:
+            self.__fields_set__.remove("batch_configs")
```
cdkini, line 335: Is there a way we can get around having to directly update this attr? `__fields_set__` is an implementation detail within Pydantic right I feel a bit wary touching it if we have an alternative.

Totally get it if that's the path we have to take

tyler-hoffman: I really dislike having to touch that as well, but I wasn't able to come up with another solution when I was looking at the same thing with the store approach. And we're already using it in some other places, e.g. `_update__fields_set__on_truthyness`.

great_expectations/datasource/fluent/interfaces.py
tyler-hoffman: Updated!

tests/datasource/fluent/data_asset/test_data_asset.py
cdkini: Nit - could we throw a `Final` annotation on them? I think we do that elsewhere with global constants but not a big deal since the all caps is an understood convention

cdkini (approved):

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -250,14 +288,17 @@ def add_batch_config(self, name: str) -> BatchConfig:
...
         batch_config.set_data_asset(self)
         self.batch_configs.append(batch_config)
+        if self.datasource.data_context:
+            try:
+                batch_config = self.datasource.add_batch_config(batch_config)
+            except DataContextError:
+                self.batch_configs.remove(batch_config)
+                raise
+        else:
+            self.batch_configs.append(batch_config)
```
billdirks, line 300: Does this add it twice because of line 292?

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -250,14 +288,17 @@ def add_batch_config(self, name: str) -> BatchConfig:
...
-        if self.datasource.is_persisted():
-            batch_config = self.datasource.add_batch_config(batch_config)
-            batch_config.set_data_asset(self)
-
         batch_config.set_data_asset(self)
         self.batch_configs.append(batch_config)
+        if self.datasource.data_context:
+            try:
+                batch_config = self.datasource.add_batch_config(batch_config)
+            except DataContextError:
```
billdirks, line 296: I'm wondering if we should do this for any exception since we don't know what the caller might catch and what might make the app crash? I guess it depends what exceptions are possible before/after writing to the backend.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -276,53 +317,36 @@ def delete_batch_config(self, batch_config: BatchConfig) -> None:
...
-        )
-
-    def get_batch_list_from_batch_request(
-        self, batch_request: BatchRequest
-    ) -> List[Batch]:
-        raise NotImplementedError
+        if self.datasource.data_context:
+            try:
+                self.datasource.delete_batch_config(batch_config)
+            except DataContextError:
```
billdirks, line 324: Similar thought here about which exceptions are relevant.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -536,22 +560,40 @@ def _execution_engine_type(self) -> Type[_ExecutionEngineT]:
...
+            raise DataContextError("Cannot save datasource without a data context.")
+
+        loaded_datasource = self.data_context.get_datasource(self.name)
+        if loaded_datasource is not self:
+            # CachedDatasourceDict will return self; only add batch config if this is a remote copy
+            assert isinstance(loaded_datasource, Datasource)
+            loaded_asset = loaded_datasource.get_asset(asset_name)
+            loaded_asset.batch_configs.append(batch_config)
+            loaded_asset.update_batch_config_field_set()
+        updated_datasource = self.data_context.update_datasource(loaded_datasource)
```
billdirks, line 574: I need to read more about the `CachedDatasourceDict` and how that works. If there are changes to the datasource locally other than adding this batch_config do those get persisted. Or does this never happen?

tyler-hoffman: I actually wrote and then didn't commit a test around this. datasources persisted to the `CachedDatasourceDict` are the the same (as in not just `==`, but `is`). See https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/datasource/datasource_dict.py#L250C14-L250C14 and https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/datasource/datasource_dict.py#L268C18-L268C18

great_expectations/datasource/fluent/interfaces.py
tyler-hoffman: Yeah looks like it did. Removed this line in https://github.com/great-expectations/great_expectations/pull/9201/files

billdirks: I hadn't noticed you did that. Thanks!

PR #9188, opened 2024-01-04
[CONTRIB] ExpectColumnValuesToBePresentInOtherTable - GX 1.0

Shinnnyshinshin:
* Updated Expectation to be compatible with GX 1.0
* Moved `example` with rest of the variables
* Now using `pydantic.root_validator` to validate `template_dict`
* Removed `__init__` method. 


contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,299 @@
+from typing import List, Optional, Union
```
cdkini: Could we add `from __future__ import annotations` and `if TYPE_CHECKING`?

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,299 @@
...
+
+    template_dict: dict = {}
+
+    query = """
+        SELECT a.{foreign_key_column}
+        FROM {active_batch} a
+        LEFT JOIN {foreign_table} b
+            ON a.{foreign_key_column} = b.{foreign_table_key_column}
+        WHERE b.{foreign_table_key_column} IS NULL
+        """
```
cdkini: Nit - could we move these to the top?

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,299 @@
...
+
+    @pydantic.root_validator
+    def _validate_template_dict(cls, values):
+        template_dict: dict = {
+            "foreign_key_column": values["foreign_key_column"],
+            "foreign_table": values["foreign_table"],
+            "foreign_table_key_column": values["foreign_table_key_column"],
+        }
+        values["template_dict"] = template_dict
+        return values
```
cdkini: Does this class have tests? How do we know these changes haven't broken things?

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,299 @@
...
+
+    @classmethod
+    @override
+    @renderer(renderer_type="renderer.prescriptive")
+    def _prescriptive_renderer(
```
cdkini, line 200: V1 work hasn't touched renderers - this all LGTM

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,299 @@
...
+            success=(final_value == 0),
+            result={
+                "observed_value": f"{final_value} missing value{'s' if final_value != 1 else ''}.",
+                "unexpected_index_list": unexpected_values,
+            },
+        )
+
+
+if __name__ == "__main__":
+    ExpectColumnValuesToBePresentInAnotherTable().print_diagnostic_checklist()
```
cdkini: This will break because you need explicit args to be passed in to the constructor now (no optional configuration) - I'd like to make `print_diagnostic_checklist` a classmethod but I don't have the bandwidth to do that refactor right now

cdkini: I'd lean towards deleting this whole `if __name__` block or just commenting it out

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: done!

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: Ok to remove.

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: Good call. I will add

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,156 @@
...
+import great_expectations.exceptions as gx_exceptions
+from great_expectations.compatibility.sqlalchemy_compatibility_wrappers import (
+    add_dataframe_to_db,
+)
+
+
+@pytest.fixture
+def referential_integrity_db(sa):
+    sqlite_engine = sa.create_engine(
+        "sqlite:///../../test_sets/referential_integrity_dataset.db"
```
cdkini: Since we duplicate this, can we make it a top level constant or fixture?

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,156 @@
...
+            foreign_table_key_column="CUSTOMER_ID",
+        )
+
+    with pytest.raises(KeyError):
+        # invalid parameter
+        validator.expect_column_values_to_be_present_in_another_table(
+            i_dont_exist="CUSTOMER_ID",
+            foreign_table="customer_table",
+            foreign_table_key_column="CUSTOMER_ID",
+        )
```
cdkini: Nit - not a requirement but could optionally parameterize the test or create three separate ones to keep things tightly scoped. Not a blocker

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,156 @@
...
+        df=customer_table,
+        name="customer_table",
+        con=sqlite_engine,
+        index=False,
+        if_exists="replace",
+    )
+    return sqlite_engine
+
+
+def test_successful_expectation_run(referential_integrity_db):
```
cdkini: Could we add a marker? Seems like an integration test coupled with an SA dep? Same below.

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,156 @@
...
+    batch_request = asset.build_batch_request()
+    expectation_suite_name = "my_suite"
+    context.add_or_update_expectation_suite(
+        expectation_suite_name=expectation_suite_name
+    )
+    validator = context.get_validator(
+        batch_request=batch_request,
+        expectation_suite_name=expectation_suite_name,
+    )
+    res = validator.expect_column_values_to_be_present_in_another_table(
```
cdkini: Instead of this dynamic validator method, could we use `validator.validate` and pass in the expectation? This is our new pattern moving forward.

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
cdkini: Same note about validator's dynamic method

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,311 @@
...
+
+    @root_validator
+    def _validate_template_dict(cls, values):
+        if not all(
+            [
+                "foreign_key_column" in values,
+                "foreign_table" in values,
+                "foreign_table_key_column" in values,
+            ]
+        ):
```
cdkini: I believe these are guaranteed due to how Pydantic works - we require all three params.

Could we remove this extra validation?

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,311 @@
...
+
+    query = """
+            SELECT a.{foreign_key_column}
+            FROM {active_batch} a
+            LEFT JOIN {foreign_table} b
+                ON a.{foreign_key_column} = b.{foreign_table_key_column}
+            WHERE b.{foreign_table_key_column} IS NULL
+            """
+
+    library_metadata = {
```
cdkini: Could we annotate any class vars as `ClassVar`? Sorry should've asked earlier

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,156 @@
...
+
+import great_expectations as gx
+import great_expectations.exceptions as gx_exceptions
+from great_expectations.compatibility.sqlalchemy_compatibility_wrappers import (
+    add_dataframe_to_db,
+)
+
+
+@pytest.fixture
+def referential_integrity_db(sa):
```
cdkini, line 21: These tests look great! I'd like to request one unit test:
- One where we pass in input args and assert that the generated `template_dict` looks like how we expect

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
cdkini: Actually the more I think about this, I don't think we need to even test invalid params - all Pydantic models fail in the same manner when this occurs

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,156 @@
...
+    assert res.success is False
+    assert res["result"]["observed_value"] == "2 missing values."
+    assert res["result"]["unexpected_index_list"] == [
+        {"CUSTOMER_ID": 5},
+        {"CUSTOMER_ID": 6},
+    ]
+
+
+def test_invalid_configurations(referential_integrity_db):
+    context = gx.get_context(cloud_mode=False)
```
cdkini: I think we could use an Ephemeral Context? We should have a fixture around

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,156 @@
+import pandas as pd
```
cdkini, line 3: Could we put this test file under coverage in `pyproject.toml`?

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,156 @@
...
+    assert res.success is True
+
+
+def test_failed_expectation_run(referential_integrity_db):
+    context = gx.get_context(cloud_mode=False)
+    datasource_name = "my_snowflake_datasource"
+    datasource = context.sources.add_sqlite(
+        datasource_name,
+        connection_string="sqlite:///../../test_sets/referential_integrity_dataset.db",
+    )
```
cdkini: Nit - if these lines are the same for each test, we could use a fixture

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: done

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: Done.

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: Removed. Thanks for letting me know about this :)

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: Done. Added `mypy` exception to list.

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: found the pattern using `batch.validate()` updating PR :)

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: I elected to split it up. The unittest you suggested actually tests two of the cases in a simpler way, and the one where the missing `I_DONT_EXIST` column is given, the `batch.validate()` actually swallows the Exception and doesn't raise. 

This is what I was imagining would happen: 
```python
checkpoint_result = checkpoint.run(
    runtime_configuration={
        "catch_exceptions": False,
    }
)
```

But, I was able to do the same test by checking the result's `exception_message` which the same test.

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
Shinnnyshinshin: done. Added a fixture that handles the building of the DataContext and Datasource :)

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,299 @@
...
+        runtime_configuration: Optional[dict] = None,
+    ) -> List[RenderedStringTemplateContent]:
+        runtime_configuration = runtime_configuration or {}
+        styling = runtime_configuration.get("styling")
+
+        foreign_key_column: str = configuration.kwargs.get("foreign_key_column")
+        foreign_table: str = configuration.kwargs.get("foreign_table")
+        foreign_table_key_column: str = configuration.kwargs.get(
+            "foreign_table_key_column"
+        )
```
cdkini, line 213: Small request - we're guaranteed these values due to prior validation so can we use `[]` instead of `.get`?

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,299 @@
...
+
+        unexpected_table_content_block = RenderedTableContent(
+            **{  # type: ignore[arg-type]
+                "content_block_type": "table",
+                "table": row_list,
+                "header_row": header_row,
+                "styling": {
+                    "body": {"classes": ["table-bordered", "table-sm", "mt-3"]}
+                },
+            }
```
cdkini: Instead of unpacking could we use keyword args? We won't need this type ignore then (I think)

tests/expectations/core/test_expect_column_values_to_be_present_in_other_table.py
```diff
@@ -0,0 +1,142 @@
...
+    assert res["result"]["unexpected_index_list"] == [
+        {"CUSTOMER_ID": 5},
+        {"CUSTOMER_ID": 6},
+    ]
+
+
+@pytest.mark.sqlite
+def test_configuration_invalid_column_name(sqlite_datasource):
+    # this is testing default behavior of `batch.validate()` which catches Exception information
+    # and places it in `exception_info`. Here we check that the exception message contains the text we expect
```
cdkini: Nit - this could go in a docstring so it is visible through pytest utilities

cdkini (approved):

PR #9195, opened 2024-01-06
[MAINTENANCE] Recursion fix draft proposal

billdirks:
This isn't quite right. We override pydantics `copy` method the signatures don't match. We could use a different method name but we probably do want to actually use `copy` and make the signatures match and actually use the input parameters. It just occurred to me I should probably use the `override` decorator.


great_expectations/expectations/expectation.py
```diff
@@ -1238,6 +1239,19 @@ def configuration(self) -> ExpectationConfiguration:
...
 
+    def copy(self, deep: bool = True) -> Self:
+        new_expectation = self.__class__(
+            id=None,
+            meta=self.meta,
+            notes=self.notes,
+            result_format=copy.deepcopy(self.result_format),
+            catch_exceptions=self.catch_exceptions,
+            rendered_content=copy.deepcopy(self.rendered_content),
+            **self.configuration.kwargs,
```
billdirks, line 1250: It just occurred to me the Expectation is a pydantic model and there might be non-kwarg additional arguments. Maybe we can leverage `__fields_set__` or some other pydantic metadata to set the arguments correctly. I can followup on Monday.

Kilo59: pydantic models don't accept positional arguments unless we have defined a custom `__init__`.

Do you mean we need to deal with `extra` kwargs that don't have a defined field?
If so, then yes, I think `__fields_set__` would include any extra fields that have been provided.

PR #9198, opened 2024-01-08
[FEATURE] `TableAsset.test_connection()` should fail if table is not queryable.

Kilo59:
This PR changes `TableAsset.test_connection()` to attempt to query the `table_name` in question rather than using the `sqlalchemy` `Inspector`. 
Without this, a user could successfully add a `TableAsset` whose table exists and passes the `.test_connection()` check, but it isn't actually usable because the user/datasource config cannot query the table.

## TODO

- [x] Add tests to ensure unquerable assets raise `TestConnectionError` when adding a `TableAsset` to a `SnowflakeDatasource`
- [x] Create and add snowflake CI role for additional testing

### DIFF
--- a/great_expectations/datasource/fluent/sql_datasource.py
+++ b/great_expectations/datasource/fluent/sql_datasource.py
@@ -913,6 +913,7 @@ class TableAsset(_SQLAsset):
         """
         datasource: SQLDatasource = self.datasource
         engine: sqlalchemy.Engine = datasource.get_engine()
+        # TODO: replace this with a select query
         inspector: sqlalchemy.Inspector = sa.inspect(engine)
 
         if self.schema_name and self.schema_name not in inspector.get_schema_names():
--- a/great_expectations/datasource/fluent/sql_datasource.py
+++ b/great_expectations/datasource/fluent/sql_datasource.py
@@ -913,7 +913,6 @@ class TableAsset(_SQLAsset):
         """
         datasource: SQLDatasource = self.datasource
         engine: sqlalchemy.Engine = datasource.get_engine()
-        # TODO: replace this with a select query
         inspector: sqlalchemy.Inspector = sa.inspect(engine)
 
         if self.schema_name and self.schema_name not in inspector.get_schema_names():
@@ -922,13 +921,24 @@ class TableAsset(_SQLAsset):
                 f'"{self.schema_name}" does not exist.'
             )
 
+        try:
+            engine.execute(f"SELECT * FROM {self.qualified_name} LIMIT 1;")
+        except sa.exc.ProgrammingError as query_error:
+            LOGGER.debug(
+                f"{self.name} `.test_connection()` query failed: {query_error!r}"
+            )
+            raise TestConnectionError(
+                f"Attempt to connect to table: {self.qualified_name} failed because the test query "
+                f"failed. Ensure the table exists and the user has access: {query_error}"
+            ) from query_error
+        # TODO: remove this inspector check
         if not inspector.has_table(
             table_name=self.table_name,
             schema=self.schema_name,
         ):
             raise TestConnectionError(
                 f"Attempt to connect to table: {self.qualified_name} failed because the table"
-                f" {self.table_name} does not exist."
+                f" {self.table_name} does not exist or the user does not have access to it."
             )
 
     @override
--- a/great_expectations/datasource/fluent/sql_datasource.py
+++ b/great_expectations/datasource/fluent/sql_datasource.py
@@ -923,11 +923,9 @@ class TableAsset(_SQLAsset):
 
         try:
             with engine.connect() as connection:
-                connection.execute(
-                    sa.select([sa.text("test_connection")])
-                    .select_from(self.as_selectable())
-                    .limit(1)
-                )
+                table = sa.table(self.qualified_name)
+                # don't need to fetch any data, just want to make sure the table is accessible
+                connection.execute(sa.select(1, table).limit(1))
         except Exception as query_error:
             LOGGER.info(
                 f"{self.name} `.test_connection()` query failed: {query_error!r}"
--- a/great_expectations/datasource/fluent/sql_datasource.py
+++ b/great_expectations/datasource/fluent/sql_datasource.py
@@ -923,7 +923,9 @@ class TableAsset(_SQLAsset):
 
         try:
             with engine.connect() as connection:
-                connection.execute(f"SELECT * FROM {self.qualified_name} LIMIT 1;")
+                connection.execute(
+                    sa.text(f"SELECT * FROM {self.qualified_name} LIMIT 1;")
+                )
         except Exception as query_error:
             LOGGER.info(
                 f"{self.name} `.test_connection()` query failed: {query_error!r}"
--- a/great_expectations/datasource/fluent/sql_datasource.py
+++ b/great_expectations/datasource/fluent/sql_datasource.py
@@ -922,7 +922,8 @@ class TableAsset(_SQLAsset):
             )
 
         try:
-            engine.execute(f"SELECT * FROM {self.qualified_name} LIMIT 1;")
+            with engine.connect() as connection:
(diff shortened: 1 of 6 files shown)


tests/datasource/fluent/integration/test_connections.py
```diff
@@ -0,0 +1,119 @@
...
+            param(
+                "snowflake://ci:${SNOWFLAKE_CI_USER_PASSWORD}@${SNOWFLAKE_CI_ACCOUNT}?warehouse=ci&role=ci",
+                id="missing database + schema",
+            ),
+            param(
+                "snowflake://ci:${SNOWFLAKE_CI_USER_PASSWORD}@${SNOWFLAKE_CI_ACCOUNT}/ci?warehouse=ci&role=ci",
+                id="missing schema",
+            ),
+        ],
+    )
```
Kilo59: @anthonyburdi I haven't actually changed the implementation of `TableAsset.test_connection()` yet so I was expecting these all to fail.
However, none of them do.

I'm using the sqlalchemy inspector to select a `table`, then I ensure that table cannot be queried (it raises a sqlalchemy programming error), but in each of the cases, the `.test_connection()` actually does fail.

Do you have any suggestions for how to reproduce the issue of the asset being un-selectable but `test_connection()` passing?

anthonyburdi: I'm guessing it is because the `ci` user has [select](https://docs.snowflake.com/en/user-guide/security-access-control-privileges#table-privileges) permissions. How about creating another user `ci_without_select_permissions` and using that in this test? I think the specific issue we (me and @josectobar) were seeing was more about the select (or [usage](https://docs.snowflake.com/en/user-guide/security-access-control-privileges#schema-privileges)) permissions rather than the connection string parameters.

tests/datasource/fluent/integration/test_connections.py
```diff
@@ -0,0 +1,123 @@
...
+    @pytest.mark.parametrize(
+        "connection_string",
+        [
+            param(
+                "snowflake://ci:${SNOWFLAKE_CI_USER_PASSWORD}@${SNOWFLAKE_CI_ACCOUNT}/ci/public?warehouse=ci",
+                id="missing role",
+            ),
+            param(
+                "snowflake://ci:${SNOWFLAKE_CI_USER_PASSWORD}@${SNOWFLAKE_CI_ACCOUNT}/ci/public?warehouse=ci&role=no_select",
+                id="role wo select",
```
Kilo59: @allensallinger to create this role.

tests/datasource/fluent/integration/test_connections.py
```diff
@@ -0,0 +1,123 @@
...
+        "connection_string",
+        [
+            param(
+                "snowflake://ci:${SNOWFLAKE_CI_USER_PASSWORD}@${SNOWFLAKE_CI_ACCOUNT}/ci/public?warehouse=ci",
+                id="missing role",
+            ),
+            param(
+                "snowflake://ci:${SNOWFLAKE_CI_USER_PASSWORD}@${SNOWFLAKE_CI_ACCOUNT}/ci/public?warehouse=ci&role=ci_no_select",
+                id="role wo select",
+            ),
```
Kilo59, line 33: @allensallinger 
```
Attempt to connect to datasource failed with the following error message:
(snowflake.connector.errors.DatabaseError) 250001 (08001):
None: Failed to connect to DB: oca29081.us-east-1.snowflakecomputing.com:443. 
Role 'CI_NO_SELECT' specified in the connect string is not granted to this user. 
Contact your local system administrator, or attempt to login with another role, e.g. PUBLIC.
```

NathanFarmer (changes requested):
What if they expect the table to be empty (I'm thinking maybe this is a setup step at the beginning of a pipeline)? I think this logic should only apply to `QueryAsset`s, not `TableAsset`s.

You can use some expectations on an empty table (expect row count to be zero, expect columns to be list, expect types to be in list, etc).

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -921,14 +921,19 @@ def test_connection(self) -> None:
...
-            table_name=self.table_name,
-            schema=self.schema_name,
-        ):
-            raise TestConnectionError(
-                f"Attempt to connect to table: {self.qualified_name} failed because the table"
-                f" {self.table_name} does not exist."
+        try:
+            with engine.connect() as connection:
+                connection.execute(
+                    sa.text(f"SELECT * FROM {self.qualified_name} LIMIT 1;")
```
allensallinger: Will we be able to perform a select here since a datasource doesn't know about a given table like a data asset would?

Kilo59:
@NathanFarmer an empty table will pass this check. We aren't actually checking the result, simply that it executes without error.

I could update this query so that it's more obvious like...
```sql
SELECT 1 from foobar LIMIT 1;
```

great_expectations/datasource/fluent/sql_datasource.py
Kilo59: This is the `TableAsset.test_connection()` method.

NathanFarmer: Can we use `sa.select` here?

tests/datasource/fluent/integration/test_connections.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # the table is queryable so the `add_table_asset()` should pass the test_connection step
+        asset = snowflake_ds.add_table_asset(
+            name="reachable asset", table_name=table_name
+        )
+        print(f"\n  Yay, asset was created!\n{asset!r}")
+
+
+if __name__ == "__main__":
+    pytest.main()
```
NathanFarmer: Why is this needed?

NathanFarmer (approved):

tests/datasource/fluent/integration/test_connections.py
Kilo59: It isn't needed but it makes it easier to run and debug a particular test module from an IDE.
It should be `pytest.min([__file__, "-vv"])`.

So if I hit the shortcut for running the current file the IDE will usually pass the current file to `python <FILE_NAME>`, which won't actually run `pytest`, but with this `if __name__ == "__main__"` it will. 

Also, as a side benefit, it helps anchor the end of the file in the git diff.

great_expectations/datasource/fluent/sql_datasource.py
Kilo59: Done

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -921,14 +921,19 @@ def test_connection(self) -> None:
...
-            table_name=self.table_name,
-            schema=self.schema_name,
-        ):
-            raise TestConnectionError(
-                f"Attempt to connect to table: {self.qualified_name} failed because the table"
-                f" {self.table_name} does not exist."
+        try:
+            with engine.connect() as connection:
+                table = sa.table(self.table_name, schema=self.schema_name)
+                # don't need to fetch any data, just want to make sure the table is accessible
```
anthonyburdi, line 927: Nice

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -921,14 +921,19 @@ def test_connection(self) -> None:
...
+                connection.execute(sa.select(1, table).limit(1))
+        except Exception as query_error:
+            LOGGER.info(
+                f"{self.name} `.test_connection()` query failed: {query_error!r}"
             )
+            raise TestConnectionError(
+                f"Attempt to connect to table: {self.qualified_name} failed because the test query "
+                f"failed. Ensure the table exists and the user has access: {query_error}"
```
anthonyburdi: What about something like this?
```suggestion
                f"Attempt to connect to table: {self.qualified_name} failed because the test query "
                f"failed. Ensure the table exists and the user has access to select data from the table: {query_error}"
```

anthonyburdi (approved):
Nice!! Just an optional suggestion on wording.

PR #9200, opened 2024-01-08
[BUGFIX] Fix Expectation intellisense

tyler-hoffman:
VS Code, using pylance ([which is recommended by pydantic](https://docs.pydantic.dev/latest/integrations/visual_studio_code/)), was unable to infer either class docstrings or constructor signatures for our Expectations. This was due to us subclassing `ModelMetaclass`, which hid some of pydantic's magic from pylance.

This PR adds the [dataclass transform decorator](https://peps.python.org/pep-0681/#the-dataclass-transform-decorator) to our `MetaExpectation` class. This is the same decorator call as used internally by pydantic on `ModelMetaclass`.


great_expectations/expectations/expectation.py
```diff
@@ -104,6 +104,17 @@
 P = ParamSpec("P")
 T = TypeVar("T", List[RenderedStringTemplateContent], RenderedAtomicContent)
 
+try:
```
tyler-hoffman: Not putting this in the compat layer because it's non-functioning (but that's okay for our needs here)

Kilo59 (approved):

great_expectations/expectations/expectation.py
```diff
@@ -104,6 +104,17 @@
...
+    from typing_extensions import dataclass_transform
```
billdirks: This is interesting. I didn't know about this decorator. I'm going to read more about this shortly. Thanks!

great_expectations/expectations/expectation.py
billdirks: Do we need to wrap this in a try/except block? It looks like this appears in typing-extensions 3.11 and we require pydantic>=3.9.2 and [3.9.2 pins typing-extensions 4.1.0](https://github.com/pydantic/pydantic/blob/v1.9.2/requirements.txt)

Does that mean no autocomplete for people with the old typing-extensions version? If so, we should make sure that gets in our docs somewhere.

billdirks (approved):
Thanks for doing this. I'm surprised we can install our package in the context of this airflow constraints file since we require `pydantic>=1.9.2` which pins typing extensions at something different than that file.

The main comment, is if code completion doesn't work without a minimum version of typing-extensions, that we get that documented.

great_expectations/expectations/expectation.py
Kilo59: It's new as of python 3.11
https://peps.python.org/pep-0681/

great_expectations/expectations/expectation.py
Kilo59: `non-blocking`
@tyler-hoffman I don't follow why we wouldn't put this in the `typing-extensions` compat layer.
The existing `@override` decorator that's in there isn't functional (at runtime) either.

PR #9212, opened 2024-01-09
[MAINTENANCE] mypy `follow_import='normal'`

Kilo59:
Update the default behavior of `mypy` to always `follow_imports`.

>follows all imports normally and type checks all top level code (as well as the bodies of all functions and methods with at least one type annotation in the signature).

https://mypy.readthedocs.io/en/stable/running_mypy.html#following-imports


anthonyburdi (approved):

billdirks (approved):
LGTM, thank you!

PR #9213, opened 2024-01-09
[DOCS] Expectation Management Script

joshua-stauffer:


tests/integration/docusaurus/expectations/expectation_management.py
```diff
@@ -0,0 +1,56 @@
...
+
+context = gx.get_context()
+suite_name = "test-suite"
+column_a = "a"
+column_b = "b"
+column_c = "c"
+
+# create a suite
+suite = ExpectationSuite(name=suite_name)
+suite = context.suites.add(suite=suite)
```
tyler-hoffman: This reassignment feels misleading, given that `suites.add` returns the same object that was passed in. I would change this to either of these:
```
suite = context.suites.add(suite=ExpectationSuite(name=suite_name))
```
or (probably better):
```
suite = ExpectationSuite(name=suite_name)
context.suites.add(suite=suite)
```

tests/integration/docusaurus/expectations/expectation_management.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+# add expectations
+expectation_a = ExpectColumnValuesToBeInSet(
```
tyler-hoffman: Non-blocking / very nit: I think this may be 2% easier to read if it were just `suite.add_expectation(ExpectColumnValuesToBeInSet(column=column_a, value_set=["foo", "bar", "baz"]))`. My thinking here is:
* the assignment makes me wonder if this will be referenced again below
* I think 1 line is just a bit less to look at, assuming it's not too terse.

tests/integration/docusaurus/expectations/expectation_management.py
```diff
@@ -0,0 +1,56 @@
...
+    column=column_a, value_set=["foo", "bar", "baz"]
+)
+suite.add_expectation(expectation=expectation_a)
+expectation_b = ExpectColumnValuesToNotBeNull(column=column_a)
+suite.add_expectation(expectation=expectation_b)
+
+# find and update a single expectation
+expectation_type = ExpectColumnValuesToNotBeNull
```
tyler-hoffman: same basic thought as above - this assignment feels like just an empty line that isn't doing anything to improve readability. `isinstance(expectation, ExpectColumnValuesToNotBeNull)` feels totally readable

tests/integration/docusaurus/expectations/expectation_management.py
```diff
@@ -0,0 +1,56 @@
...
+expectation = next(
+    expectation
+    for expectation in suite.expectations
+    if isinstance(expectation, expectation_type)
+)
+expectation.column = column_b
+expectation.save()
+
+# batch update expectations
```
tyler-hoffman: Do we need this example? Feels like you know this out from the previous example. Maybe if we just made 2 comments on the above example like "find a single expectation" and "update expectation"

tests/integration/docusaurus/expectations/expectation_management.py
```diff
@@ -0,0 +1,56 @@
...
+for expectation in suite.expectations:
+    expectation.column = column_c
+suite.save()
+
+# delete an expectation
+expectation_type = ExpectColumnValuesToNotBeNull
```
tyler-hoffman: In the interest of minimizing the amount of code to absorb, I would just delete the first expectation we got (the one we update).

tyler-hoffman (changes requested):
Overall this looks great. I'm being a bit more picky than usual here because this will be public facing and I think we need to prioritize keeping it short and *suite*. A couple thoughts beyond what I left inline:

* IMO we should use more realistic values. I see us elsewhere doing things like using `pickup_datetime` for a column_name. Updating that and not using foo, bar, baz, etc for value names is likely going to make this click a bit more.
* There are a lot of variables assigned before they are used, e.g. `column_a`. I'd want that in a unit test if we used it multiple times, but I think that obscures things if someone is looking at a doc.
* Similarly, given that we don't think most users like using named args, I'd probably just go with positional args in cases where we only pass in one arg.

tests/integration/docusaurus/expectations/expectation_management.py
cdkini: I prefer the former so we can demonstrate that `context.suites.add` returns an object

tests/integration/docusaurus/expectations/expectation_management.py
cdkini: I think we should use `gxe.Expect_` if this is going to be referenced in docs (I believe that is our canonical pattern?)

tests/integration/docusaurus/expectations/expectation_management.py
joshua-stauffer: hm, yes i think we do need an example for how to update multiple expectations as a batch, since batch updating expectations works at the suite level, whereas updating a single expectation works at the expectation level. i've renamed to `update multiple expectations as a batch`, lmk if that helps clarify this!

tyler-hoffman: Oh I missed that we're saving at the suite level. I'm with you now!

tests/integration/docusaurus/expectations/expectation_management.py
joshua-stauffer: thanks both! updated in [018c3a3](https://github.com/great-expectations/great_expectations/pull/9213/commits/018c3a343dbbefc7dbc5c2359be124e73e82bc84)

tests/integration/docusaurus/expectations/expectation_management.py
joshua-stauffer: Passing the object into the `SuiteFactory` mutates it, which I find clearer to understand when the variable is reassigned. I've updated to your first suggestion in [018c3a3](https://github.com/great-expectations/great_expectations/pull/9213/commits/018c3a343dbbefc7dbc5c2359be124e73e82bc84)

tyler-hoffman (approved):
Awesome!

tests/integration/docusaurus/expectations/expectation_management.py
```diff
@@ -0,0 +1,41 @@
...
+suite = context.suites.get(name=suite_name)
+
+# add expectations
+suite.add_expectation(
+    gxe.ExpectColumnValuesToBeInSet(column="passenger_count", value_set=[1, 2, 3, 4, 5])
+)
+suite.add_expectation(gxe.ExpectColumnValuesToNotBeNull(column="pickup_datetime"))
+
+# find a single expectation
+expectation = next(
```
jcampbell, line 22: Do we have to call next here? That doesn't feel very basic to me. Is the idea to wrap up a list comprehension plus index?

tests/integration/docusaurus/expectations/expectation_management.py
```diff
@@ -0,0 +1,41 @@
...
+expectation.column = "pickup_location_id"
+expectation.save()
+
+# update multiple expectations as a batch
+for expectation in suite.expectations:
+    expectation.notes = "This Expectation was generated as part of GX Documentation."
+suite.save()
+
+# delete an expectation
+suite.delete_expectation(expectation=expectation)
```
jcampbell, line 49: Can we move this up to the part about finding an expectation so it's not relying on the left over iteration from above?

joshua-stauffer: okay - the reason its at the bottom is that the two calls above are also relying on the `expectation` variable retrieved above. Should we include `get` code with each example?

tests/integration/docusaurus/expectations/expectation_management.py
joshua-stauffer: yep - would you prefer to explicitly index into the filtered list? this just reads cleaner to me, so I would suggest this as the easiest way to obtain a single expectation, and users can add more filter conditions as needed.

Rachel-Reverie: Is there a reason we don't just put this in a helper function so that the user is running something like:
```
expectation = suite.get_expectation(gxe.ExpectColumnValuesToNotBeNull)
```
or
```
expectation = suite.get_expectation(gxe.SomeExpectationWithDomainArgs, domain_arg_value1, domain_arg_value2)
```
?
I think that would be much simpler for most users than anything involving a list comprehension/indexing method.

tests/integration/docusaurus/expectations/expectation_management.py
Rachel-Reverie: It would be best if each example is as self-contained as possible.  So I would vote for including the `get` code each time, unless the examples are grouped very closely and are related to the same process.

tests/integration/docusaurus/expectations/expectation_management.py
Rachel-Reverie: I agree that the reassignment makes it more intuitive that something has changed.  However, if this method is mutating an object in place, as well as returning that same object, please make sure to include in the docstring for the method that the in-place mutation takes place so that people will be aware that their previous object and "new" object are the same if they reference the API docs.

In general, I think it's considered a more pythonic practice to avoid this potential confusion by having methods mutate _or_ return, but not both.  Just looking at another common mutation, a list's `append` method:
```
>> my_list = []
>> my_list = my_list.append("hi")
>> print(my_list)
>> None
```

tests/integration/docusaurus/expectations/expectation_management.py
joshua-stauffer: @Rachel-Reverie thanks for the suggestion - we decided to defer implementing a helper method like that for the time being. There are several approaches we considered, but since Expectations can be unique on any aspect of their configuration, the API will be relatively complex. I expect we'll add something like this in the future.

tyler-hoffman: @jcampbell I like what @joshua-stauffer has here. This seems idiomatic to me.

Rachel-Reverie: If we aren't going to include a helper function to manage this, can we add an example of getting a specific Expectation based on its type and domain kwargs, for those Expectations that can be in an Expectation Suite multiple times for different fields/values?

jcampbell: I'm comfortable leaving it as is -- thanks @tyler-hoffman and @joshua-stauffer .

jcampbell (approved):

PR #9216, opened 2024-01-10
[MAINTENANCE] mypy - `ignore_missing_import=False`

Kilo59:
Update mypy config for `ignore_missing_imports=False`
Without this setting any imported modules that do not export types or have a stubfile will be treated as `Any`

https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports

Many of our third-party dependencies do not export types or have stubs available and so these packages have been excluded from this rule with overrides in `pyproject.toml`.


tyler-hoffman (approved):
Nice. And thanks for the link in the dsecription!

PR #9235, opened 2024-01-16
[MAINTENANCE] Another docs build fix

tyler-hoffman:
This reverts commit 6bb0ea83b277094760432ee81511cddfcb313df5. It also updates `prepare_prior_versions.py` to be a bit more optimized so it doesn't time out in netlify. I'd rather do that as 2 PRs, but doing it together proves we get things totally green in CI.

this PR introduced the errors seen here https://github.com/great-expectations/great_expectations/actions/runs/7543744906/job/20535756842 and I'd like to isolate the various build issues.


cdkini (approved):

docs/prepare_prior_versions.py
```diff
@@ -590,6 +590,7 @@ def prepare_prior_versions() -> None:
     update_tag_references_for_correct_version()
     use_relative_path_for_imports()
     prepend_version_info_to_name_for_md_relative_links_to_index_files()
+    prepend_version_info_to_name_for_md_relative_links()
```
tyler-hoffman, line 564: Missed this in https://github.com/great-expectations/great_expectations/pull/9234 and to see the docs build go green, we need this change and all the removed public_api decorators.

docs/prepare_prior_versions.py
```diff
@@ -55,7 +55,22 @@ def _paths_to_versioned_docs_after_v0_15_50() -> list[pathlib.Path]:
     paths = [
         f
         for f in data_path.iterdir()
-        if f.is_dir() and ("0.14.13" not in str(f) or "0.15.50" not in str(f))
```
tyler-hoffman, line 58: the or here looks like it was just making this return True for every version AFAICT

cdkini: What is the significance of these strings? Could we make them constants to give them some more context?

docs/prepare_prior_versions.py
cdkini: What's the plan to revert the reversion?

docs/prepare_prior_versions.py
tyler-hoffman: They're past versions. We could make them constants, but I think it'd still be fragile. What if we just plan to fix this in https://greatexpectations.atlassian.net/browse/V1-131?

docs/prepare_prior_versions.py
tyler-hoffman: The diff makes it hard to read, but `prepend_version_info_to_name_for_md_relative_links` and `prepend_version_info_to_name_for_md_relative_links_to_index_files` were merged together and now both are handled in `prepend_version_info_to_name_for_md_relative_links`.

PR #9239, opened 2024-01-17
[MAINTENANCE] Update Expectation Gallery for 1.0

joshua-stauffer:
This PR updates the Expectation Gallery build script to be compatible with 1.0 changes.


output_data.json
```diff
@@ -0,0 +1 @@
...l}, "error_diagnostics": null}, {"test_title": "negative_case_with_75percent_null_values_no_mostly", "backend": "trino", "test_passed": true, "include_in_gallery": false, "validation_result": {"success": false, "exception_info": {"raised_exception": false, "exception_traceback": null, "exception_message": null}, "expectation_config": {"success_on_last_run": null, "kwargs": {"result_format": {"result_format": "COMPLETE"}, "include_config": false, "column": "three_null", "batch_id": "b1ae3225b6aa3d6c44dce474d8b4e741"}, "ge_cloud_id": null, "expectation_context": null, "meta": {}, "raw_kwargs": {"result_format": {"result_format": "COMPLETE"}, "include_config": false, "column": "three_null"}, "expectation_type": "expect_column_values_to_not_be_null", "rendered_content": null}, "result": {"element_count": 4, "unexpected_count": 3, "unexpected_percent": 75.0, "partial_unexpected_list": [null, null, null], "partial_unexpected_counts": [{"value": null, "count": 3}], "unexpected_list": [null, null, null], "unexpected_index_query": "SELECT three_null \nFROM expect_column_values_to_not_be_null_1 \nWHERE three_null IS NULL;"}, "meta": {}, "rendered_content": null}, "error_diagnostics": null}]}
```
tyler-hoffman: What's this?

joshua-stauffer: ah, debug output - thanks!

assets/scripts/build_gallery.py
```diff
@@ -321,9 +329,38 @@ def get_expectation_instances(expectations_info):
             )
             expectation_tracebacks.write(traceback.format_exc())
             continue
+        except pydantic.ValidationError as exc:
+            logger.error(exc)
+            continue
```
tyler-hoffman: I don't think this continue is doing anything. And should the other `except`s reraise?

assets/scripts/build_gallery.py
```diff
@@ -306,11 +311,14 @@ def get_expectation_instances(expectations_info):
...
-                expectation_name
-            ] = great_expectations.expectations.registry.get_expectation_impl(
-                expectation_name
-            )()
+            expectation_class = (
+                great_expectations.expectations.registry.get_expectation_impl(
+                    expectation_name
+                )
+            )
+            # instantiate expectation
```
tyler-hoffman: nit: I don't think this comment is adding any value. If anything I'd just rename `s/exp/expectation` to over communicate it's an instance of an expectation.

assets/scripts/build_gallery.py
joshua-stauffer: we don't want to raise an exception here, since it'll kill the script. if an expectation breaks when we try to build its docs, we just skip it.

you're right about the `continue`, will remove.

assets/scripts/build_gallery.py
```diff
@@ -321,9 +329,38 @@ def get_expectation_instances(expectations_info):
...
+            logger.error(exc)
+            continue
+        except IndexError:
+            logger.error(f"Expectation {expectation_name} has invalid test case.")
+        except Exception as exc:
+            logger.error(f"Unexpected error occurred: {exc}")
     return expectation_instances
 
 
+def create_expectation_instance(expectation_class: type[Expectation]) -> Expectation:
```
tyler-hoffman, line 343: Wow nice job coming up with a solution here. What percent of expectations are getting handled by this? Are we losing anything we had before?

Also I think we should add some explanatory comments around this bad boy and likely make (and link here) a task around coming up with a more robust solution.

tyler-hoffman (approved):
Approved, pending a JIRA to track diffing this against an 0.18 build to see what if anything we're losing.

assets/scripts/build_gallery.py
joshua-stauffer: i'll update the docstring - i actually think this will be a good long-term solution, but we could establish stronger guarantees around `Expectations` always having these example parameters. And we should only have a single way to write `Expectation` tests, preferably in the `tests` dir and not on the `Expectation` itself.

tyler-hoffman: I know we close but not 100% aligned on this part when we last talked, but just to document what I'd like to see in the future:
* Not checking 2 places and potentially raising
* Not reaching into dicts for the data or reading from disk
* Strong guarantees and fast feedback when we fail to implement one of these. If we can do that with static typing, that seems ideal, but I suspect that requires adding the example itself to the expectation.
* I'd also like to see this not get pulled from test data. If anything, I think the example instance should live outside of both tests and business logic, and then get pulled into tests and the gallery build.
* If we can't enforce this with static typing, we should have a test with very nice failure output
But none of that's blocking here of course.

assets/scripts/build_gallery.py
```diff
@@ -320,10 +327,46 @@ def get_expectation_instances(expectations_info):
...
+    """Accepts an Expectation class and attempts to instantiate it with valid parameters
+    obtained from the Expectation's test cases.
+    """
+    expectation_params: dict
+    if expectation_class.examples:
+        # take the first test case available:
+        expectation_params = expectation_class.examples[0]["tests"][0]["in"]
```
cdkini, line 350: Are we confident these always work? We have some test cases that are failure cases (at least in the JSON files).

assets/scripts/build_gallery.py
```diff
@@ -320,10 +327,46 @@ def get_expectation_instances(expectations_info):
...
+            _TEST_DEFS_DIR.rglob(f"**/{expectation_class.expectation_type}.json"), None
+        )
+        if not found:
+            raise Exception(
+                f"No JSON test case found for Expectation {expectation_class.expectation_type}, cannot build gallery example."
+            )
+        with open(found) as fp:
+            test_cases = json.load(fp)
+            # take the first test case available:
+            expectation_params = test_cases["datasets"][0]["tests"][0]["in"]
```
cdkini, line 366: Same note as above - perhaps we check if "catch_exceptions" or another similar key is around

assets/scripts/build_gallery.py
joshua-stauffer: We don't need these to be success cases - as long as they have the correct arguments of the correct type, it works enough to instantiate the expectation. As far as I know these test cases aren't supposed to pass in unexpected parameters. There might be a couple wayward examples, and i think we can clean them up.

PR #9241, opened 2024-01-17
[MAINTENANCE] Add instrumentation for V1 Expectation management APIs

cdkini:
Use new Posthog instrumentation with Expectation and ExpectationSuite workflows


tests/analytics/test_events.py
```diff
@@ -0,0 +1,105 @@
...
+    ExpectationSuiteExpectationCreatedEvent,
+    ExpectationSuiteExpectationDeletedEvent,
+    ExpectationSuiteExpectationUpdatedEvent,
+)
+
+
+@pytest.mark.parametrize(
+    "event_cls, input_kwargs, expected_properties",
+    [
+        pytest.param(
```
cdkini: Pretty silly tests but I want to be extremely careful that we're not changing payload shapes down the line - these are defensive and cost very little

great_expectations/analytics/anonymizer.py
```diff
@@ -0,0 +1,13 @@
...
+from hashlib import md5
+
+
+class Anonymizer:
+    def __init__(self, salt: str | None = None) -> None:
+        self._salt = salt or secrets.token_hex(8)
+
+    def anonymize(self, string_: str) -> str:
+        salted = self._salt + string_
+        return md5(salted.encode("utf-8")).hexdigest()
```
cdkini: Ported over whatever logic from the current OSS analytics I thought made sense here

great_expectations/analytics/events.py
```diff
@@ -1,13 +1,115 @@
...
+class _ExpectationSuiteExpectationEvent(Event):
+    expectation_id: str | None = None
+    expectation_suite_id: str | None = None
+
+    @override
+    def _properties(self) -> dict:
+        return {
+            "expectation_id": self.expectation_id,
+            "expectation_suite_id": self.expectation_suite_id,
+        }
```
cdkini, line 36: Feels like this is always just the properties on the object? We could make this a lot simpler by iterating through `self.__dict__` or something?

great_expectations/core/expectation_suite.py
```diff
@@ -174,8 +184,30 @@ def add_expectation(self, expectation: Expectation) -> Expectation:
...
+
+        submit_event(
+            event=ExpectationSuiteExpectationCreatedEvent(
+                action=EXPECTATION_SUITE_EXPECTATION_CREATED,
+                expectation_id=expectation.id,
+                expectation_suite_id=self.ge_cloud_id,
+                expectation_type=expectation_type,
+                custom_exp_type=custom_exp_type,
+            )
+        )
```
cdkini: Who owns this custom expectation identification logic and the related anonymization? Perhaps a separate Anonymizer?

great_expectations/analytics/actions.py
```diff
@@ -3,3 +3,9 @@
 from great_expectations.analytics.base_event import Action
 
 DATA_CONTEXT_INITIALIZED = Action("data_context.initialized")
+EXPECTATION_SUITE_CREATED = Action("expectation_suite.created")
+EXPECTATION_SUITE_DELETED = Action("expectation_suite.deleted")
+EXPECTATION_SUITE_UPDATED = Action("expectation_suite.updated")
+EXPECTATION_SUITE_EXPECTATION_CREATED = Action("expectation_suite.expectation_created")
+EXPECTATION_SUITE_EXPECTATION_DELETED = Action("expectation_suite.expectation_deleted")
+EXPECTATION_SUITE_EXPECTATION_UPDATED = Action("expectation_suite.expectation_updated")
```
cdkini: Not sure if all these will be used but simply going off of initial discussions and the analytics doc we have in Confluence

tyler-hoffman: Should the latter 3 be just `Action("expectation.<CRUD>")`?

great_expectations/analytics/anonymizer.py
```diff
@@ -0,0 +1,13 @@
...
+
+import secrets
+from hashlib import md5
+
+
+class Anonymizer:
+    def __init__(self, salt: str | None = None) -> None:
+        self._salt = salt or secrets.token_hex(8)
+
+    def anonymize(self, string_: str) -> str:
```
tyler-hoffman: is `string` a reserved word?

great_expectations/analytics/anonymizer.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import annotations
+
+import secrets
+from hashlib import md5
+
+
+class Anonymizer:
+    def __init__(self, salt: str | None = None) -> None:
+        self._salt = salt or secrets.token_hex(8)
```
tyler-hoffman, line 9: why hash at all if it's functionally non-deterministic (when we don't pass in a salt)? Don't we lose all value if every time we initialize an Anonymizer? Like if we always passed in a salt, and then could correlate values without having to know what they were, but I would imagine just removing the value entirely from the payload has as much value as hashing it with a random salt. But I could totally be missing something

great_expectations/analytics/events.py
```diff
@@ -1,13 +1,115 @@
...
+
     _allowed_actions: ClassVar[List[Action]] = [DATA_CONTEXT_INITIALIZED]
 
     def __init__(self):
         super().__init__(action=DATA_CONTEXT_INITIALIZED)
+
+
+@dataclass
+class _ExpectationSuiteExpectationEvent(Event):
+    expectation_id: str | None = None
```
tyler-hoffman, line 28: IMO these defaults make it easy to do the wrong thing. Can we just leave it defaultless?

tests/analytics/test_events.py
```diff
@@ -0,0 +1,105 @@
...
+            {
+                "expectation_suite_id": "fbb7ada0-600d-458d-a4f7-c6c30cb759b4",
+            },
+            id="ExpectationSuiteDeletedEvent",
+        ),
+    ],
+)
+@pytest.mark.unit
+def test_event_properties(event_cls, input_kwargs, expected_properties):
+    event = event_cls(**input_kwargs)
```
tyler-hoffman: why not just instantiate the events in the pytest.params? Seems like representing the args as dictionaries and then building the event here is a tiny bit more work and a tiny bit more indirection.

great_expectations/core/expectation_suite.py
```diff
@@ -174,8 +184,30 @@ def add_expectation(self, expectation: Expectation) -> Expectation:
...
+            custom_exp_type = False
+            expectation_type = expectation.expectation_type
+        else:
+            custom_exp_type = True
+            anonymizer = Anonymizer()
+            expectation_type = anonymizer.anonymize(expectation.expectation_type)
+
+        submit_event(
+            event=ExpectationSuiteExpectationCreatedEvent(
+                action=EXPECTATION_SUITE_EXPECTATION_CREATED,
```
tyler-hoffman: It feels like a smell to me that we have a class just for this action, and we still have to pass the action in. IMO we should either have a generic ExpectationSuiteExpectationEvent that takes in the action, or the event should know its own action and instantiate itself with that.

great_expectations/core/expectation_suite.py
```diff
@@ -174,8 +184,30 @@ def add_expectation(self, expectation: Expectation) -> Expectation:
...
         return expectation
 
+    def _submit_expectation_created_event(self, expectation: Expectation) -> None:
+        if expectation.__module__.startswith("great_expectations."):
+            custom_exp_type = False
+            expectation_type = expectation.expectation_type
+        else:
+            custom_exp_type = True
+            anonymizer = Anonymizer()
+            expectation_type = anonymizer.anonymize(expectation.expectation_type)
```
tyler-hoffman, line 204: we're expecting expectation type to contain sensitive data?

tyler-hoffman (changes requested):
This looks pretty solid to me. The two requests I have though:
* minor, and IIRC I'm complaining about something that's already an established pattern, but can we make our events a bit more ergonomic, e.g. so we don't have to pass `EXPECTATION_SUITE_EXPECTATION_CREATED` to `ExpectationSuiteExpectationCreatedEvent`?
* Looks like most of the dataclasses have defaults provided, but IMO that just makes it easier to do the wrong thing. e.g. if the user wants to not pass in the expectation_type, let them pass in None, rather than defaulting to `"UNKNOWN"`
* I'm not sure I understand the value of hashing expectation_type, but this could likely be me missing context.

great_expectations/analytics/events.py
```diff
@@ -1,13 +1,115 @@
...
+    expectation_suite_id: str | None = None
+
+    @override
+    def _properties(self) -> dict:
+        return {
+            "expectation_id": self.expectation_id,
```
joshua-stauffer, line 34:
```suggestion
    def _properties(self) -> dict:
        from dataclasses import asdict
        return asdict(self)
```

great_expectations/analytics/events.py
joshua-stauffer: i would use the builtin dataclass function, unless there's a strong reason not to

great_expectations/analytics/actions.py
cdkini: Hmm based on the Confluence doc I think we're not doing `expectation.<CRUD>` but rather `expectation_suite.expectation_<CRUD>`

great_expectations/analytics/anonymizer.py
cdkini: Hmm no that's a fair point - I'm basing this off of the prior implementation but that's something to look into for sure

great_expectations/analytics/anonymizer.py
cdkini: nope I can change this

great_expectations/analytics/events.py
cdkini: I'd like that but dataclasses can be a little weird - I'm basing this off of Mercury events but I can try to make things required

great_expectations/analytics/events.py
cdkini: I like that!

great_expectations/core/expectation_suite.py
cdkini: If its a custom name, we don't want to send it over the wire

great_expectations/core/expectation_suite.py
cdkini: I agree but the shape of the data is slightly different between create vs delete/update

tests/analytics/test_events.py
cdkini: I think I might just remove this entirely based on Josh's `asdict` suggestion

great_expectations/analytics/events.py
tyler-hoffman: ah gotcha. I think the main workaround to that that I've seen in the past is to use the `kw_only` argument to the dataclass (I forget the nuance though), but that's only python 3.10 IIRC, so irrelevant for us here.

great_expectations/analytics/anonymizer.py
cdkini: I believe we use the `data_context_id` as salt

great_expectations/core/expectation_suite.py
cdkini: Having all this logic here feels a little bad though?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -487,6 +489,10 @@ def _save_expectation_suite(
         return self.expectations_store.set(key, expectation_suite, **kwargs)
 
     # Properties
+    @property
+    def anonymizer(self) -> Anonymizer:
+        return self._anonymizer
```
cdkini, line 494: Don't love adding more properties but if the anonymizer relies directly on the `data_context_id` for hashing, we need to make it publicly available

great_expectations/analytics/anonymizer.py
tyler-hoffman: Then should we make the salt arg non-optional? Maybe a comment on this would be good. I just feel like if I were to see this in the wild I wouldn't be quite sure how to use it.

great_expectations/core/expectation_suite.py
tyler-hoffman: It does seem like non-business logic in an area I'd want business logic only, but I'm not sure we have architecture that allows for anything better decoupled atm. I wouldn't block this PR on that being there.

tyler-hoffman: I'm wondering if we should wait to send these events until we are connected to a data context. Not a blocker.

great_expectations/analytics/anonymizer.py
tyler-hoffman: Okay, I see it being used below without passing anything in for cases when we don't have a data context. I think it might make sense to consider only sending events when we get a data context, but I think that's a question for later. I'm on board with this now.

tyler-hoffman (approved):

PR #9243, opened 2024-01-17
[DOCS] Customization of Header

deborahniesz:
# Description

Customization of the search bar component [(still using docsearch algolia)](https://docsearch.algolia.com/docs/what-is-docsearch/), changing the Documentation Logo and adding a github badge to match the new mockup.

[Jira ticket](https://greatexpectations.atlassian.net/jira/software/c/projects/DSB/boards/79?issueParent=26175&selectedIssue=DSB-591)

[Deploy Preview](https://deploy-preview-9243.docs.greatexpectations.io/docs/home/)

#### Previous version: 
Screenshot from 2024-01-23 16-18-15

#### New version: 
Screenshot from 2024-01-23 16-01-23

#### Previous version: 
Screenshot from 2024-01-17 12-27-04

#### New version:
Screenshot from 2024-01-17 12-24-51

#### On hover:
Screenshot from 2024-01-17 12-25-58

#### Github badge:
You can see the github badge shown at the top right corner, when clicked, it take us to the repository.

https://github.com/great-expectations/great_expectations/assets/38264052/e448d05d-3eac-4cf2-8766-144db359d17f

#### Github badge if fetch throws an error, exceeds number of available requests or github API is down.
In the video i'm stoping the request from the github API and reloading the page so you can see how the component will render if it can't access the github API.

https://github.com/great-expectations/great_expectations/assets/38264052/1e843a0d-45db-46c9-bb92-da0eafd956d2


deborahniesz:
@cla-bot check

kwcanuck (approved):

docs/docusaurus/src/components/NavbarItems/GithubNavbarItem/index.js
josectobar, line 1: Why is the Github logo not including the release version like it does in the designs? 

image

docs/docusaurus/src/components/NavbarItems/GithubNavbarItem/index.js
```diff
@@ -0,0 +1,52 @@
...
+            <img src={githubMarkImg} className={styles.githubMark}
+                 alt="Github Invertocat Logo"/>
+            { showGithubBadgeInfo && (<div className={styles.githubBadgeInfo}>
+                <img src={githubLogoImg} className={styles.githubLogo}
+                     alt="Github Logo"/>
+                <div className={styles.githubStats}>
+                    <div>
+                        <img src={starIcon} alt="Github Stargazers Count"/>
+                        {starsCount}
+                    </div>
```
josectobar: nit pick: We should use spans here instead.
```suggestion
                    <span>
                        <img src={starIcon} alt="Github Stargazers Count"/>
                        {starsCount}
                    </span>
```

docs/docusaurus/src/components/NavbarItems/GithubNavbarItem/index.js
```diff
@@ -0,0 +1,52 @@
...
+                    <div>
+                        <img src={forkIcon} alt="Github Forks Count"/>
+                        {forksCount}
+                    </div>
```
josectobar:
```suggestion
                    <span>
                        <img src={forkIcon} alt="Github Forks Count"/>
                        {forksCount}
                    </span>
```

docs/docusaurus/src/components/NavbarItems/GithubNavbarItem/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+    function formatCompactNumber(number) {
+        const formatter = Intl.NumberFormat("en", { notation: "compact" });
+        return formatter.format(number).toLowerCase();
+    }
+
+    let githubMarkImg = useBaseUrl(`img/github-mark.svg`);
+    let githubLogoImg = useBaseUrl(`img/github.svg`);
+    let starIcon = useBaseUrl(`img/star.svg`);
+    let forkIcon = useBaseUrl(`img/code-branch.svg`);
```
josectobar: We should avoid using `let` if we're not re-assigning the variable.
```suggestion
    const githubMarkImg = useBaseUrl(`img/github-mark.svg`);
    const githubLogoImg = useBaseUrl(`img/github.svg`);
    const starIcon = useBaseUrl(`img/star.svg`);
    const forkIcon = useBaseUrl(`img/code-branch.svg`);
```

docs/docusaurus/src/components/NavbarItems/GithubNavbarItem/styles.module.css
```diff
@@ -0,0 +1,42 @@
+.githubBadge {
+    display: flex;
+    gap: 0.25rem;
+    align-items: flex-start;
+    font-size: 0.75rem;
```
josectobar: I think we should use a typography variable here.

docs/docusaurus/src/css/custom.scss
josectobar, line 1: I'm not super familiar with Sass. is it possible that instead of putting all custom classes in this one single file, we add them into files and sub-folders?

e.g.:
```
└──custom/
    ├──githubBadge/
    │  └── github-badge.scss
    └──navbar/
       └── navbar.scss
```

docs/docusaurus/src/theme/NavbarItem/ComponentTypes.js
```diff
@@ -0,0 +1,10 @@
+// Temporary workaround to creating custom navbar items according to the documentation when
+// running 'yarn swixxle --list'
+// See https://github.com/facebook/docusaurus/issues/7227.
```
josectobar, line 3: 👍🏽

josectobar (changes requested):
This looks great! Just left a few comments. thanks!

docs/docusaurus/src/components/NavbarItems/GithubNavbarItem/index.js
josectobar: Also, I'm not a fan of how the image flickers when you refresh the page or at first load. Maybe if we add a  min width to the container that shows the star/merge branch image and the count and we make the `showGithubBadgeInfo` true by default and only set it to false it if there's an error.

JessSaavedra: We agreed to take that out during a meeting, as it was not such a relevant info and it's not something that's retrieved in the same endpoint that we're using now, so we had do two requests to github api instead of one. Also, if I recall correctly, the final amount of releases was not something that was already calculated in the API but we had to do some extra processing, maybe @deborahniesz can elaborate on the details

deborahniesz: Yes, there is this workaround to avoid having to request all branches or tags since there was no endpoint that returned the total amount, but it seemed to be a waste to request all branches only to count the elements. So the solution was to request them all but with 1 element per page, so the endopoint only returns one element but the header of the response tells you how many pages there are in total, which would be the total amount of elements since you are requesting one per page
`function getElementCountFromGithubAPI(element) {

    fetch(`https://api.github.com/repos/great-expectations/great_expectations/${element}?per_page=1`)
        .then(response => {
            const headers = response.headers;
            const paginationLinks = headers.get('Link');
            return paginationLinks.split('per_page=1&page=')[2].split('>')[0] // get total pages
        })
}`
Still, since it was decided to only show stars and forks, this code was deprecated

docs/docusaurus/src/css/custom.scss
deborahniesz: Yes you are right! 
I would leave the github badge css inside the github-badge component folder but the css for the searchbar I moved it from the generic custom.scss to its own file!

SuperYeehee (commented):
Having some thoughts around the Hover effect on the Search. I think its a bit inconsistent with how our other 'Icon only' buttons are styled. Looking at the Feedback section, styling is a bit different. Will review the Button styling with Becca.  
Overall work looks great! Thank you!

docs/docusaurus/src/components/NavbarItems/GithubNavbarItem/index.js
josectobar, line 1: Can we add a test for the logic on this component?

josectobar (approved):
👍🏽

josectobar:
I approved this but there's a step failing in CI was it doing this before?

JessSaavedra:
> I approved this but there's a step failing in CI was it doing this before?

I think it's because it was outdated related to develop, I updated the branch and now is passing!

PR #9255, opened 2024-01-19
[MAINTENANCE] Remove Click constraint in 0.18.x

cdkini:
Remove constraint now that we have additional minor versions to grab from Click

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 altair>=4.2.1,<5.0.0
-Click>=7.1.2
+Click>=7.1.2 # 8.1.4 shouldn't cause any runtime issues it will die at type checking
 colorama>=0.4.3
 cryptography>=3.2
 Ipython>=7.16.3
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 altair>=4.2.1,<5.0.0
-Click>=7.1.2,!=8.1.4 # 8.1.4 has a mypy typing bug (https://github.com/pallets/click/issues/2558)
+Click>=7.1.2
```
billdirks: Type checking fails with Click 8.1.4 and our static analysis checks fail. CI would die if Click==8.1.4 is installed. I'm not sure we want this change.

billdirks (changes requested):

requirements.txt
billdirks: It looks like Click 8.1.5 only has type changes (including addition of tests for this):
https://click.palletsprojects.com/en/8.1.x/changes/

There shouldn't be any runtime issues so we can remove this but I'd add a comment saying it dies but only at type checking time.

billdirks (approved):

PR #9258, opened 2024-01-21
[MAINTENANCE] Run docs build file processing during versionsing, rather than build

tyler-hoffman:
This adds methods for the new flow, but does not enable it. The high level is that there's a new command `invoke docs version=<VERSION>`. Our preprocessing step (`prepare_prior_versions.py` is now run only when versioning (previously we ran that script on all versions as part of the build process). After versioning, we upload all files relevant to the build process to s3. Before, this was versions.json, versioned_docs, and versioned_sidebars, but we now include versioned_code (which I want to strip down later if we can). The build process will now just be pulling the s3 zip file, and running `docusaurus build` (which is still slow because it runs for all versions, but at least we aren't processing all versions ourselves during build).

I'll put up a second PR to actually cut over to using this process.


docs/docs_build.py
```diff
@@ -50,9 +70,60 @@ def build_docs_locally(self) -> None:
...
+        self._context.run(f"git checkout {version}")
+        old_version_file = self._read_prior_release_version_file()
+        self._write_release_version(
+            "\n".join(
+                [
+                    "// this file is autogenerated",
+                    "export default {",
+                    f"  release_version: 'great_expectations, version {version}',",
+                    "   min_python: '3.8',",
+                    "   max_python: '3.11',",
```
cdkini: Nit - Should these be constants?

docs/prepare_prior_versions.py
```diff
@@ -474,6 +468,46 @@ def _prepend_version_info_for_md_absolute_links(contents: str, version: Version)
     return contents
 
 
+def remove_extension_from_absolute_md_links(
+    version: Version,
+    verbose: bool = False,
+) -> None:
+    """Needed for a bad link in 0.18 docs"""
```
cdkini, line 475: Do we have a plan for resolving this?

cdkini (approved):

docs/prepare_prior_versions.py
tyler-hoffman: Resolved in https://github.com/great-expectations/great_expectations/pull/9012, but our process checks out old commits where it's not fixed. I propose that when we version 0.18, we just version it to 0.18.9 (which should be the next version to come out)

docs/docs_build.py
tyler-hoffman: updated!

PR #9260, opened 2024-01-22
[MAINTENANCE] Add Cloud user id to analytics calls

cdkini:
Override init analytics to ensure that new user_id passed down from Cloud call is part of emitted Posthog events.


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -337,13 +337,16 @@ def __init__(self, runtime_environment: Optional[dict] = None) -> None:
         self._attach_fluent_config_datasources_and_build_data_connectors(
             self.fluent_config
         )
+        self._init_analytics()
+        submit_event(event=DataContextInitializedEvent())
+
+    def _init_analytics(self) -> None:
         oss_id = self._get_oss_id()
         init_analytics(
             user_id=oss_id,
```
jcampbell: Why are we passing `user_id` in this case instead of leaving it as None?

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -141,6 +143,15 @@ def __init__(  # noqa: PLR0913
             runtime_environment=runtime_environment,
         )
 
+    @override
+    def _init_analytics(self) -> None:
+        init_analytics(
+            user_id=self._user_id,
+            data_context_id=uuid.UUID(self._data_context_id),
+            oss_id=self._get_oss_id(),
```
jcampbell, line 158: What is oss_id in the context of a CloudDataContext? Is it the same client identifier (i.e. machine info)?

jcampbell (dismissed):

great_expectations/data_context/data_context/cloud_data_context.py
cdkini: Yes the same as with file-backed

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: I'm happy to modify that if that makes sense - cc @wookasz

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -145,13 +149,28 @@ def _check_if_latest_version(self) -> None:
...
+        user_id = data["user_id"]
+        return uuid.UUID(user_id)
+
     @override
     def _init_project_config(
         self, project_config: Optional[Union[DataContextConfig, Mapping]]
     ) -> DataContextConfig:
         if project_config is None:
             project_config = self.retrieve_data_context_config_from_cloud(
-                cloud_config=self._cloud_config,
```
cdkini, line 154: We should always use the property when available

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -247,22 +266,30 @@ def retrieve_data_context_config_from_cloud(
...
 
         session = create_session(access_token=cloud_config.access_token)
-
         response = session.get(cloud_url)
-        if response.status_code != 200:  # noqa: PLR2004
+
+        try:
+            response.raise_for_status()
+        except HTTPError:
             raise gx_exceptions.GXCloudError(
```
cdkini, line 294: Better error handling around requests made from the context

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -402,9 +429,9 @@ def _init_datasource_store(self) -> DatasourceStore:
         store_backend: dict = {"class_name": GXCloudStoreBackend.__name__}
         runtime_environment: dict = {
             "root_directory": self.root_directory,
-            "ge_cloud_credentials": self.ge_cloud_config.to_dict(),  # type: ignore[union-attr]
+            "ge_cloud_credentials": self.cloud_config.to_dict(),
```
cdkini: Removed the `ge_` prefix

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -337,13 +337,15 @@ def __init__(self, runtime_environment: Optional[dict] = None) -> None:
...
             self.fluent_config
         )
-        oss_id = self._get_oss_id()
+        self._init_analytics()
+        submit_event(event=DataContextInitializedEvent())
+
+    def _init_analytics(self) -> None:
         init_analytics(
-            user_id=oss_id,
+            user_id=None,
```
cdkini: @jcampbell changed this per the discussion in Tetris

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -338,13 +338,18 @@ def __init__(self, runtime_environment: Optional[dict] = None) -> None:
         self._attach_fluent_config_datasources_and_build_data_connectors(
             self.fluent_config
         )
-        oss_id = self._get_oss_id()
         self._anonymizer = Anonymizer(salt=self._data_context_id)
+
+        # TODO: Need to determine V1 mechanism for usage stats opt-out
```
billdirks: V1 analytics is controlled by env variable `GX_ANALYTICS_ENABLED`. Or does this comment mean something else? For the current usage stats, we should always have that disabled and work on removing it completely.

great_expectations/data_context/data_context/cloud_data_context.py
tyler-hoffman: I lack context her. What's the difference?

tyler-hoffman: Oh I see. I missed that this is just a property wrapping the private attribute.

tests/analytics/test_analytics.py
```diff
@@ -69,3 +72,49 @@ def test_event_identifiers(analytics_config):
         assert "user_id" not in properties
     if "organization_id" not in base_properties:
         assert "organization_id" not in properties
+
+
+@pytest.mark.unit
+def test_local_context_init(monkeypatch):
```
tyler-hoffman: nit: I'd probably `s/local/ephemeral`

tests/analytics/test_analytics.py
```diff
@@ -69,3 +72,49 @@ def test_event_identifiers(analytics_config):
...
+    monkeypatch.setenv("GE_USAGE_STATS", "True")  # Enable usage stats
+
+    with mock.patch(
+        "great_expectations.data_context.data_context.abstract_data_context.init_analytics"
+    ) as mock_init, mock.patch("posthog.capture") as mock_submit:
+        _ = gx.get_context(mode="ephemeral")
```
tyler-hoffman, line 85: Nit: no need to even do an assignment here, right?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -338,13 +338,18 @@ def __init__(self, runtime_environment: Optional[dict] = None) -> None:
...
+        if self._is_usage_stats_enabled():
```
billdirks: You shouldn't do this check here as we no longer want to be gated by this environment variable. The call to `_init_analytics` initializes the posthog client. If `GX_ANALYTICS_ENABLED` is set to False this will disable the client. I don't know the behavior of `submit_event` if posthog.enabled == `false` but presumable, it is a no-op. @wookasz  can you verify? 
I would experiment to understand the behavior here.

If not, you should change the init_analytics call to return a value that let's us know if analytics is enable or not.

tyler-hoffman (approved):
This looks good to me

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -145,13 +149,30 @@ def _check_if_latest_version(self) -> None:
...
+    @override
+    def _init_analytics(self) -> None:
+        init_analytics(
+            user_id=self._get_cloud_user_id(),
+            data_context_id=uuid.UUID(self._data_context_id),
+            oss_id=self._get_oss_id(),
+            cloud_mode=True,
+        )
+
+    def _get_cloud_user_id(self) -> uuid.UUID:
```
billdirks: Should we cache this result? This can be [done with a decorator](https://docs.python.org/3/library/functools.html#functools.cache). It is fine not to but I don't think this should change for a context.

billdirks (changes requested):
Thanks for doing this. We should be using the posthog analytics env variable instead of the old usage stats one.

tests/analytics/test_analytics.py
cdkini: Will change!

tests/analytics/test_analytics.py
cdkini: Nope I tend to do it to be explicit that a function returns something but we're ignoring the return value - good to remove

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: Ah okay I didn't know about this env var

great_expectations/data_context/data_context/cloud_data_context.py
cdkini: We'll only ever call this on init once - since GX workflows aren't long-lived processes, does it make sense to cache?

great_expectations/data_context/data_context/abstract_data_context.py
wookasz: Correct, if that env variable is disabled, `submit` calls will be a no-op. See [this](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/analytics/client.py#L59).

Currently, it's just the environment variable, but @jcampbell did raise a great point that we should probably make this env var OR config. Most other OSS packages we looked at do allow for disabling via config.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -341,9 +341,8 @@ def __init__(self, runtime_environment: Optional[dict] = None) -> None:
         self._anonymizer = Anonymizer(salt=self._data_context_id)
 
         # TODO: Need to determine V1 mechanism for usage stats opt-out
```
billdirks: I think you can remove this comment now. We can file a follow up ticket for making this config in addition to the env variable control.

great_expectations/data_context/data_context/cloud_data_context.py
billdirks: We can leave it as is and, if for some reason, we start hitting it multiple times we can update.

billdirks (approved):
LGTM, thanks!

great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -145,13 +150,33 @@ def _check_if_latest_version(self) -> None:
...
+        init_analytics(
+            user_id=self._get_cloud_user_id(),
+            data_context_id=uuid.UUID(self._data_context_id),
+            oss_id=self._get_oss_id(),
+            cloud_mode=True,
+        )
+
+    def _get_cloud_user_id(self) -> uuid.UUID | None:
+        if not ENV_CONFIG.gx_analytics_enabled:
+            return None
```
cdkini, line 164: Short-circuiting here allows us to skip an unnecessary request to the backend - the main win is the simplicity of testing but that's probably not a great reason to have this.

Can iterate as we productionalize usage stats

PR #9266, opened 2024-01-23
[FEATURE] Handle distinct_id inside analytics config

jcampbell:

### DIFF
--- a/great_expectations/analytics/base_event.py
+++ b/great_expectations/analytics/base_event.py
@@ -49,6 +49,10 @@ class Event:
     
     @property
     def distinct_id(self) -> UUID:
+        """The distinct_id is the primary key for identifying
+        anlaytics events. It is the user_id if it is set
+        (e.g. in a Cloud context), otherwise the oss_id.
+        """
         return self.user_id or self.oss_id
 
     _allowed_actions: ClassVar[Optional[List[Action]]] = None
--- a/great_expectations/analytics/base_event.py
+++ b/great_expectations/analytics/base_event.py
@@ -46,7 +46,7 @@ class Event:
     @property
     def user_id(self) -> Optional[UUID]:
         return get_config().user_id
-    
+
     @property
     def distinct_id(self) -> UUID:
         """The distinct_id is the primary key for identifying
--- a/great_expectations/analytics/base_event.py
+++ b/great_expectations/analytics/base_event.py
@@ -36,7 +36,7 @@ class Event:
         return get_config().data_context_id
 
     @property
-    def organization_id(self) -> UUID:
+    def organization_id(self) -> Optional[UUID]:
         return get_config().organization_id
 
     @property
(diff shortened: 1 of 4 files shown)


great_expectations/analytics/base_event.py
```diff
@@ -36,17 +36,21 @@ def data_context_id(self) -> UUID:
...
         return get_config().oss_id
 
     @property
-    def user_id(self) -> UUID:
+    def user_id(self) -> Optional[UUID]:
         return get_config().user_id
 
+    @property
+    def distinct_id(self) -> UUID:
+        return self.user_id or self.oss_id
```
cdkini: Could we add a docstring here just to clarify what this entails? Why do we prioritize `user_id` over `oss_id`?

great_expectations/analytics/base_event.py
```diff
@@ -66,10 +70,14 @@ def get_allowed_actions(cls):
     def properties(self) -> dict:
         props = {
             "data_context_id": self.data_context_id,
-            "organization_id": self.organization_id,
             "oss_id": self.oss_id,
-            "service": "python-client",
+            "service": "gx-core",
```
cdkini, line 78: This won't break anything right? What's motivating this change?

tests/analytics/test_analytics.py
```diff
@@ -0,0 +1,70 @@
...
+    ],
+)
+def analytics_config(request):
+    base_config = get_config()
+    update_config(request.param[0])
+    yield request.param[1], request.param[2]
+    update_config(base_config)
+
+
+def test_event_identifiers(analytics_config):
```
cdkini, line 51: Could we annotate with `@pytest.mark.unit`?

cdkini (changes requested):
Some clarifying questions

wookasz (approved):
This looks good form my end, pending @cdkini's comments. Thanks for making the changes.

great_expectations/analytics/base_event.py
jcampbell: The motivation was just to be more specific, since I could see there being python clients (potentially even including our agent). gx-core seemed most analogous to other service names currently in use (whereas `python-client` seemed too close to the `library` field auto-added by posthog (`posthog-python`)

tests/analytics/test_analytics.py
jcampbell: I need a primer on the pytest markings -- I don't think there were any other tests for analytics, and I'm not sure. So, I'll add that annotation, but please to tell me why!

great_expectations/analytics/base_event.py
jcampbell: I added a comment that I hope addresses this fully, but happy to revise again. Basically, the user_id is more likely to be correct than a generic (non user specific) OSS ID

tests/analytics/test_analytics.py
jcampbell: Oh, nice, we even had an automated check for this!

great_expectations/analytics/base_event.py
cdkini: I'm okay with this as long as Tetris is!

tests/analytics/test_analytics.py
cdkini: I think the `pyproject.toml` lists all the available ones but maybe we need an analytics one if we start going E2E

cdkini (approved):

PR #9282, opened 2024-01-25
[MAINTENANCE] allow absolute file links

tyler-hoffman:
Allow absolute file links. This is a prereq for https://github.com/great-expectations/great_expectations/pull/9281. I didn't make this as robust as the other checks because I want to work on removing our custom checks in favor of relying on docusurus's build-in checks. If anything, I might want to lint to prevent any links that start with `/docs/`, since they'll break with versioning. But that needs to happen later.

See https://docusaurus.io/docs/markdown-features/links 's info on "absolute file paths"

### DIFF
--- a/docs/checks/docs_link_checker.py
+++ b/docs/checks/docs_link_checker.py
@@ -251,6 +251,7 @@ class LinkChecker:
         - Relative links that begin with either . or .. and have a .md suffix
         - Relative images with an image suffix
         - Docroot links that begin with a character (neither . or /) are relative to the doc root (ex: /docs) and have a .md suffix
+        - Absolute file paths for markdown files
 
         Args:
             match: A positive match of a markdown link (ex: [...](...)) or image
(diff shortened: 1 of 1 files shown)


docs/checks/docs_link_checker.py
```diff
@@ -284,6 +288,10 @@ def _check_link(self, match: re.Match, file: str) -> Optional[LinkReport]:
                     result = self._check_absolute_link(
                         link, file, match.group("path"), match.group("version")
                     )
+                elif match := self._absolute_file_pattern.match(link):  # type: ignore[assignment]
```
billdirks, line 291: nit: You don't need to set `match` in this `elif` since you don't use it. I do think using this walrus pattern elsewhere in this crazy conditional instead of would make things clearer though since that would prevent us from repeatedly shadowing it.

billdirks (approved):
LGTM

PR #9298, opened 2024-01-29
[MAINTENANCE] Remove deprecated fixture mark usage

Kilo59:
Removed mark that was applied to a fixture.
Marking fixtures this way had no effect, but now it's an error that was breaking our CI with the release of `pytest` 8.
https://docs.pytest.org/en/stable/deprecations.html#applying-a-mark-to-a-fixture-function

- Deduplicate s3 related mocks and move them to `tests/datasources/fluent/conftest.py`
- Update the `aws_credentials` to use `monkeypatch.setenv` instead of directly setting them with `os.environ`

Seems like there might be a bug or undocumented change related to `pytest.warns`.
We have a failing spark test that uses `pytest.warns()` but not in a way that has deprecated according to these docs 👇 
https://docs.pytest.org/en/stable/deprecations.html#using-pytest-warns-none

Rather than debug this error, I'm pinning our version of `pytest` to `<8.0`


tests/datasource/fluent/test_spark_s3_datasource.py
```diff
@@ -29,34 +26,11 @@
...
-def aws_credentials() -> None:
-    """Mocked AWS Credentials for moto."""
-    os.environ["AWS_ACCESS_KEY_ID"] = "testing"
-    os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
-    os.environ["AWS_SECURITY_TOKEN"] = "testing"
-    os.environ["AWS_SESSION_TOKEN"] = "testing"
-    os.environ["AWS_DEFAULT_REGION"] = "testing"
-
-
-@pytest.mark.skipif(not aws.boto3)
```
Kilo59, line 52: These marks were causing the CI failure.
I've moved these to `conftest.py` and removed the marks.

tests/datasource/fluent/test_pandas_s3_datasource.py
```diff
@@ -46,34 +44,11 @@
...
 
 
-@pytest.fixture(scope="function")
-def aws_credentials() -> None:
-    """Mocked AWS Credentials for moto."""
-    os.environ["AWS_ACCESS_KEY_ID"] = "testing"
-    os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
-    os.environ["AWS_SECURITY_TOKEN"] = "testing"
-    os.environ["AWS_SESSION_TOKEN"] = "testing"
-    os.environ["AWS_DEFAULT_REGION"] = "testing"
```
Kilo59, line 66: This is a dangerous fixture as these persist beyond the lifetime of the test.
I've moved this to a `conftest.py` file and used `pyest.monkeypatch.setenv()` so that this is unset at the end of the test.

It's possible we've had some tests that were relying on this state.

tests/datasource/fluent/conftest.py
```diff
@@ -235,6 +237,30 @@ def fluent_gx_config_yml_str(fluent_gx_config_yml: pathlib.Path) -> str:
     return fluent_gx_config_yml.read_text()
 
 
+@pytest.fixture()
```
Kilo59, line 240:
```suggestion
@pytest.fixture
```

tyler-hoffman (approved):

reqs/requirements-dev-lite.txt
```diff
@@ -8,7 +8,7 @@ nbconvert>=5
 nbformat>=5.0
 pact-python>=2.0.1
 pyfakefs>=4.5.1
-pytest>=6.2.0
+pytest>=7.2,<8.0
```
cdkini: Could we add a comment around why we're pinning?

cdkini (approved):

reqs/requirements-dev-lite.txt
Kilo59: Added

PR #9308, opened 2024-01-29
[MAINTENANCE] Add `description` to `Expectation` to enable simpler renderered content

cdkini:


great_expectations/expectations/core/expect_multicolumn_sum_to_equal.py
```diff
@@ -82,7 +82,6 @@ class ExpectMulticolumnSumToEqual(MulticolumnMapExpectation):
 
     @classmethod
     @renderer(renderer_type=LegacyRendererType.PRESCRIPTIVE)
-    @renderer(renderer_type="renderer.prescriptive")
```
cdkini, line 85: Let's remove string usage and any duplicate decorators

great_expectations/expectations/expectation.py
```diff
@@ -284,6 +284,7 @@ class Config:
     meta: Union[dict, None] = None
     notes: Union[str, List[str], None] = None
     result_format: Union[ResultFormat, dict] = ResultFormat.BASIC
+    description: ClassVar[Union[str, None]] = None
```
cdkini, line 287: Explicitly making a `ClassVar` because I want it to be defined within the custom class (not overridden during instantiation)

great_expectations/render/renderer/content_block/content_block.py
```diff
@@ -94,8 +94,15 @@ def _render_list(  # noqa: PLR0913, PLR0912
...
+            expectation_config = (
+                obj_.expectation_config
+                if isinstance(obj_, ExpectationValidationResult)
+                else obj_
+            )
 
-            content_block_fn = cls._get_content_block_fn(expectation_type)
+            content_block_fn = cls._get_content_block_fn(
+                expectation_type=expectation_type, expectation_config=expectation_config
+            )
```
cdkini: This logic is a bit convoluted to begin with - we either have an `ExpectationConfiguration` or an `ExpectationValidationResult`.

In order to find out if we have a `description`, we need to parse out the config, convert it to an `Expectation`, access the `description`, and then use it in a custom `content_block_fn`

great_expectations/render/renderer/content_block/content_block.py
```diff
@@ -225,11 +232,22 @@ def _render_other(  # noqa: PLR0912, PLR0913
...
+        content_block_fn = cls._get_content_block_fn(
+            expectation_type=expectation_type, expectation_config=expectation_config
         )
-        content_block_fn = content_block_fn[1] if content_block_fn else None
+        if not content_block_fn:
+            renderer_impl = get_renderer_impl(
+                object_name=expectation_type,
+                renderer_type=LegacyRendererType.PRESCRIPTIVE,
+            )
+            content_block_fn = renderer_impl[1] if renderer_impl else None
```
cdkini: Quick rename since `content_block_fn` should be a `Callable` rather than a `tuple`

great_expectations/render/renderer/content_block/content_block.py
```diff
@@ -294,6 +312,28 @@ def _render_other(  # noqa: PLR0912, PLR0913
...
+                content_block_type="string_template",
+                string_template={
+                    "template": description,
+                    # If we want to support $VAR substitution, params should be `expectation.configuration`
+                    # Keeping this simple to limit scope of this feature for the time being
+                    "params": {},
+                    "styling": runtime_configuration.get("styling", {}),
+                },
+            )
+        ]
```
cdkini: If we have a `description`, this gets invoked and we simply plug into the existing rendering ecosystem

tests/render/test_column_section_renderer.py
```diff
@@ -74,6 +80,27 @@ def titanic_validation_results():
...
+        @render_evaluation_parameter_string
+        def _prescriptive_renderer(
+            cls,
+            **_,
+        ):
+            assert (
+                False
+            ), "Expectation `description` should take precedence over `prescriptive_renderer`"
+
+    return ExpectColumnAgesToBeLegalAdult()
```
cdkini: I know we've been trying to avoid fake classes but `description` needs to be set on something and I don't want to modify source code

tests/render/test_column_section_renderer.py
```diff
@@ -1194,6 +1221,23 @@ def test_ExpectationSuiteColumnSectionRenderer_expectation_with_single_string_me
     assert result_json == expected_result_json
 
 
+@pytest.mark.unit
+def test_ExpectationSuiteColumnSectionRenderer_render_expectation_with_description(
```
cdkini, line 1212: I believe I've encapsulated the primary usecases that would hit this logic within these tests but I'm not 100% confident - some manual testing seems to confirm these beliefs but let me know if you think I'm missing something.

great_expectations/expectations/core/expect_multicolumn_sum_to_equal.py
tyler-hoffman: I dig getting rid of magic strings and duplicate renderers, but I have to imagine we should be moving away from `LegacyRendererType`, right?

cdkini: Yup! This has no functional change just cleaning up a few inconsistencies

billdirks: `LegacyRendererType.PRESCRIPTIVE == "renderer.prescriptive"` which is why this is duplicative?

great_expectations/render/renderer/content_block/content_block.py
```diff
@@ -225,11 +232,22 @@ def _render_other(  # noqa: PLR0912, PLR0913
...
         expectation_type = cls._get_expectation_type(render_object)
+        expectation_config = (
+            render_object.expectation_config
+            if isinstance(render_object, ExpectationValidationResult)
+            else render_object
+        )
 
-        content_block_fn = get_renderer_impl(
-            object_name=expectation_type, renderer_type=LegacyRendererType.PRESCRIPTIVE
+        content_block_fn = cls._get_content_block_fn(
```
billdirks: I think the types are correct in the function because `expectation_config` may be a `render_object` which is of type `Any`.

great_expectations/render/renderer/content_block/content_block.py
billdirks: Can you make this whole block a private function called `get_content_block_fn_from_renderer` or something. This logic is the same here and the other place you modify in this file.

great_expectations/render/renderer/content_block/content_block.py
```diff
@@ -377,12 +417,34 @@ def _process_content_block(
...
+    @classmethod
+    def _get_content_block_fn_from_expectation_description(
+        cls, expectation_config: ExpectationConfiguration | None
+    ) -> Callable | None:
+        if expectation_config:
+            expectation = expectation_config.to_domain_obj()
+            description = expectation.description
+            if description:
+                return cls._render_expectation_description
+        return None
```
billdirks: Do we need to make the expectation_config here or could we do something like:
`return expectation_config.get("description", None) if expectation_config else None`

great_expectations/render/renderer/content_block/content_block.py
billdirks: Multiple types in this list does make this convoluted.

great_expectations/render/renderer/content_block/content_block.py
```diff
@@ -294,6 +312,28 @@ def _render_other(  # noqa: PLR0912, PLR0913
...
 
+    @classmethod
+    def _render_expectation_description(
+        cls,
+        configuration: ExpectationConfiguration,
+        runtime_configuration: dict,
+        **kwargs,
+    ) -> list[RenderedStringTemplateContent]:
+        expectation = configuration.to_domain_obj()
+        description = expectation.description
```
billdirks, line 313: Will passing `None` to `RenderedStringTemplateContent` be ok or should this be `description = expectation.description or ""`?

I know we are only using this in 1 code path right now which has a non-None description.

great_expectations/render/renderer/content_block/validation_results_table_content_block.py
```diff
@@ -92,20 +92,29 @@ def _process_content_block(
...
     @classmethod
-    def _get_content_block_fn(cls, expectation_type):
-        expectation_string_fn = get_renderer_impl(
-            object_name=expectation_type, renderer_type=LegacyRendererType.PRESCRIPTIVE
+    def _get_content_block_fn(  # noqa: PLR0915
+        cls,
+        expectation_type: str,
+        expectation_config: ExpectationConfiguration | None = None,
+    ) -> Callable | None:
+        content_block_fn = cls._get_content_block_fn_from_expectation_description(
```
billdirks: This method is identical to the one it overrides with some extra stuff at the bottom. I know it has an early return in it that makes it harder but it would be nice to call the super version and then do the other stuff if possible.

billdirks (commented):
I totally agree that this code is a bit odd. I left some drive by comments. I haven't looked at tests yet.

great_expectations/render/renderer/content_block/content_block.py
cdkini: the list will be consistent in the types it contains

cdkini: done!

great_expectations/render/renderer/content_block/content_block.py
cdkini: We should be guaranteed a non-null value - I've added a small check

great_expectations/render/renderer/content_block/content_block.py
cdkini: Since `description` is a `ClassVar`, we need the `Expectation` to retrieve it - we want the `Expectation` to be the source of truth moving forward I just don't know if/when we'll have the bandwidth to remove config

tests/render/test_column_section_renderer.py
billdirks: I think the last assert that the result is rendered as the description string is the main use case we want to test.

tests/render/test_render_BulletListContentBlock.py
```diff
@@ -76,15 +76,27 @@ def test_all_expectations_using_test_definitions():
         len(test_files) == 61
     ), "Something went wrong when collecting JSON Expectation test fixtures"
 
+    # Not ported over to V1 so ignore for purposes of this test
```
billdirks, line 79: Should we delete this expectations for V1?

tests/render/test_render_BulletListContentBlock.py
```diff
@@ -76,15 +76,27 @@ def test_all_expectations_using_test_definitions():
...
     for filename in test_files:
         test_definitions = json.load(open(filename))
-
+        if test_definitions["expectation_type"] in UNSUPPORTED_EXPECTATIONS:
+            continue
         for dataset in test_definitions["datasets"]:
             for test in dataset["tests"]:
                 # Construct an expectation from the test.
                 if type(test["in"]) == dict:  # noqa: E721
+                    # Skip tests with invalid configurations
```
billdirks, line 96: Why do we need this new `continue` conditional? This seems like a very lenient short circuit of this test.

billdirks (commented):
This looks significantly better, thank you! I have a couple questions about tests, namely that last short continue question.

tests/render/test_render_BulletListContentBlock.py
cdkini: Probably? Not sure what the plan was here to be honest

tests/render/test_render_BulletListContentBlock.py
cdkini: Because of our new call to `ExpectationConfiguration.to_domain_obj()`, we require a valid configuration for these tests. In production, a user will always have a valid configuration due to Pydantic validation, resulting in renderers always utilizing complete objects

tests/render/test_render_BulletListContentBlock.py
billdirks: I looked at [this spreadsheet](https://docs.google.com/spreadsheets/d/1aIIelGPVdHbijKsqNsTlheg9vAAayi3t_OYHwT9og_o/edit#gid=0) to get a list of the expectations we are going to support. Oh these 4, I see `expect_column_pair_values_A_to_be_greater_than_B` to be on that list. I also see it on your [spreadsheet in row 14](https://docs.google.com/spreadsheets/d/1roV3afWMOzpU52p_dQyab_t64js9aic0ev4eNdzBwEI/edit#gid=0).

I think we should delete the top 3 (I'll file a ticket). The last one should be ported over so should be passing?

tests/render/test_render_BulletListContentBlock.py
billdirks: So this is an artifact of bad configuration in tests?
How many are there, can we fix them? Or at least get a count? Maybe we should just test some expectations vs all. I feel with this `continue` we could skip all the expectations and not be aware that we aren't testing anything.

billdirks: I misunderstood. Having `catch_exceptions` in the configuration mean we expect this test case to fail. Thanks for explaining!

billdirks (approved):
Thanks!

PR #9345, opened 2024-01-31
[MAINTENANCE] Only use @site/src for CardLink and CardLinkGrid

tyler-hoffman:
This removes the versioned copies of this under `docs/` and `versioned_docs/`. And to make that works, it updates all imports to use `@site/src`


docs/docusaurus/docs/components/LinkCard/index.js
```diff
@@ -1,28 +0,0 @@
-import React from 'react';
```
billdirks, line 1: Are these deleted files all identical between versions. Is there a shared copy they are now using?
My only concern would be introducing a dependency on the versions with each other and I'm curious if this would introduce any problems.

tyler-hoffman: yup! so within `docusaurus/docs/` we have `src/components/LinkCard` (and the grid). `diff` shows no difference between either the versioned directories or the src one.

Since this is more global styling than anything version, I think it makes sense to live external to any versions (src). (https://docusaurus.io/docs/2.x/versioning#use-absolute-import-within-the-docs provides an example of this).

billdirks: Sounds good to me.

billdirks (approved):
Thanks!

PR #9364, opened 2024-02-02
[FEATURE] ExpectationSuite accepts Expectations on init

joshua-stauffer:


great_expectations/core/expectation_suite.py
```diff
@@ -122,15 +128,34 @@ def __init__(  # noqa: PLR0913
...
-            ExpectationConfiguration(**expectation)
-            if isinstance(expectation, dict)
-            else expectation
-            for expectation in expectations
-        ]
-        self.expectations: list[Expectation] = [
-            self._build_expectation(config) for config in expectation_configurations
-        ]
+
+        if all(isinstance(exp, Expectation) for exp in expectations):
```
tyler-hoffman: This is a lot to grok for a constructor. Can this logic be moved out into a separate method/function?

great_expectations/core/expectation_suite.py
```diff
@@ -122,15 +128,34 @@ def __init__(  # noqa: PLR0913
...
+                    raise ValueError(
+                        "Expectations in parameter `expectations` must not belong to another ExpectationSuite. "
+                        "Instead, please use copies of Expectations, by calling `copy.copy(expectation)`."
+                    )
+                expectation.register_save_callback(save_callback=self._save_expectation)
+            self.expectations = expectations
+        else:
+            # legacy code path, remove once param `expectations` type is narrowed to just Expectation
+            for expectation in expectations:
+                if isinstance(expectation, Expectation):
```
tyler-hoffman: The need for this check is a side-effect of the type.

I'm guessing we want the runtime check as well, but if we don't allow mixing types, we should fix the type of `expectations`. It's more verbose, but I think we want

```
expectations: Optional[Union[
  Sequence[Expectation],
  Sequence[dict],
  Sequence[ExpectationConfiguration], # maybe these last two can get rolled in together since they are used interchangeably
]] = None,
```

great_expectations/core/expectation_suite.py
joshua-stauffer: okay - that might help with some of the mypy issues as well

great_expectations/core/expectation_suite.py
tyler-hoffman: Alternatively, we could just handle mixed sequences. That actually might make some of the logic here simpler.

tests/core/test_expectation_suite.py
```diff
@@ -215,6 +215,36 @@ def expectation(self) -> gxe.ExpectColumnValuesToBeInSet:
...
+            for column in columns
+        ]
+        with pytest.raises(
+            ValueError,
+            match="Expectations in parameter `expectations` must not belong to another ExpectationSuite.",
+        ):
+            ExpectationSuite(
+                name=self.expectation_suite_name, expectations=expectations
+            )
+
```
tyler-hoffman, line 247: If we need the runtime check about mixed types, we should have a test. But do we actually want to enforce that, or can we just handle mixed types.

tyler-hoffman (changes requested):
Requesting changes around the type of `expectations` and how we handle/don't handle sequences of mixed types. Let's either gracefully handle them, or update the signature.

great_expectations/core/expectation_suite.py
```diff
@@ -122,15 +128,34 @@ def __init__(  # noqa: PLR0913
...
+        if all(isinstance(exp, Expectation) for exp in expectations):
+            for expectation in expectations:
+                cast(Expectation, expectation)
+                if expectation.id:
+                    raise ValueError(
+                        "Expectations in parameter `expectations` must not belong to another ExpectationSuite. "
+                        "Instead, please use copies of Expectations, by calling `copy.copy(expectation)`."
+                    )
+                expectation.register_save_callback(save_callback=self._save_expectation)
+            self.expectations = expectations
```
tyler-hoffman: Alright, final comment. Sorry for so much feedback about something so minor! If mypy is sad about this, then a simple solution might be to replace all of this logic in the constructor with `self.expectations = ExpectationSuite._ensure_expectation([exp for exp in expectations or []])`. Implementing `_ensure_expectation` is left as an exercise to the reader.

tests/core/test_expectation_suite.py
joshua-stauffer: mixed types are weird conceptually but totally fine in practice. updated to reflect

tyler-hoffman (approved):

PR #9394, opened 2024-02-07
[MAINTENANCE] Update docs create version script to omit patch

tyler-hoffman:
See title. We only want to cut major.minor versions of docs, so that we don't end up with e.g. 0.18.0, 0.18.1, 0.18.2, etc in versioned_code. We will manage the display of the patch version in the docusaurus config.

This also required updating some tests. I deleted one test, rather than updating it, because the function under test (like many in this file) will be deleted in the very near future.


docs/prepare_prior_versions.py
```diff
@@ -534,20 +534,15 @@ def prepare_prior_version(version: Version) -> None:
     if version >= Version(0, 16):
         prepend_version_info_to_name_for_md_images(version)
 
-    # TODO: delete this after we cut 0.18.9
```
tyler-hoffman, line 537: We've already done this processing, so this is unneeded. (Really most of this file is, but we can address that separately)

docs/docs_build.py
```diff
@@ -86,6 +86,7 @@ def create_version(self, version: Version) -> None:
         os.chdir("..")  # TODO: none of this messing with current directory stuff
         prepend_version_info_to_name_for_snippet_by_name_references(version)
         os.chdir("docusaurus")
+        self._write_release_version(old_version_file)
```
tyler-hoffman, line 83: Stray fix - sorry for the scope creep. See above for where we store the old version, then temporarily overwrite it.

docs/prepare_prior_versions.py
```diff
@@ -534,20 +534,15 @@ def prepare_prior_version(version: Version) -> None:
...
 @dataclass(frozen=True)
 class Version:
     major: int = 0
     minor: int = 0
-    patch: int = 0
 
-    def as_tuple(self) -> tuple[int, int, int]:
-        return self.major, self.minor, self.patch
+    def as_tuple(self) -> tuple[int, int]:
+        return self.major, self.minor
```
cdkini, line 545: Sorry just seeing this code now - could we instead just leverage `version.Version`? Feels like a place where the stdlib has what we need

cdkini (approved):

PR #9413, opened 2024-02-08
[FEATURE] `CheckpointFactory`

cdkini:


ghost:
<details open>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=9413&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

####

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=9413&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map legend
</details>

great_expectations/core/factory/factory.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    Responsible for basic CRUD operations on collections of GX domain objects.
+    """
+
+    @abstractmethod
+    def add(self, obj: T) -> T:
+        pass
+
+    @abstractmethod
+    def delete(self, obj: T) -> T:
```
cdkini, line 17: Why does this return an object?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -323,6 +314,22 @@ def __init__(self, runtime_environment: Optional[dict] = None) -> None:
...
+
+        self._checkpoints: CheckpointFactory | None = None
+        if checkpoint_store := self.stores.get(self.checkpoint_store_name):
+            self._checkpoints = CheckpointFactory(
+                store=checkpoint_store,
+            )
+
     def _init_analytics(self) -> None:
         init_analytics(
             data_context_id=uuid.UUID(self._data_context_id),
```
cdkini: I actually agree with this comment - we should fail fast here.

If a user doesn't configure their suite store, they'll only see a failure when they run `context.suites` (which could happen much later than this init)

great_expectations/data_context/store/configuration_store.py
```diff
@@ -157,7 +157,7 @@ def self_check(self, pretty_print: bool = True) -> dict:  # type: ignore[overrid
     def serialization_self_check(self, pretty_print: bool) -> None:
         raise NotImplementedError
 
-    def _determine_key(
+    def get_key(
```
cdkini, line 129: Rename to be consistent

great_expectations/core/factory/checkpoint_factory.py
```diff
@@ -0,0 +1,82 @@
...
+        AbstractDataContext,
+    )
+    from great_expectations.data_context.store import CheckpointStore
+
+
+# TODO: Add analytics as needed
+class CheckpointFactory(Factory[Checkpoint]):
+    def __init__(self, store: CheckpointStore, context: AbstractDataContext):
+        self._store = store
+        self._context = context
```
cdkini, line 27: Checkpoints currently have a reliance on contexts - I'd really like to decouple this but for the time being, the factory needs a pointer here so it can create checkpoints

great_expectations/core/factory/checkpoint_factory.py
```diff
@@ -0,0 +1,82 @@
...
+from great_expectations.exceptions import DataContextError
+
+if TYPE_CHECKING:
+    from great_expectations.data_context.data_context.abstract_data_context import (
+        AbstractDataContext,
+    )
+    from great_expectations.data_context.store import CheckpointStore
+
+
+# TODO: Add analytics as needed
```
cdkini, line 23: Not sure if we have a ticket around this but will have to investigate

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -541,6 +548,14 @@ def suites(self) -> SuiteFactory:
             )
         return self._suites
 
+    @property
+    def checkpoints(self) -> CheckpointFactory:
+        if not self._checkpoints:
+            raise gx_exceptions.DataContextError(
+                "DataContext requires a configured CheckpointStore to persist Checkpoints."
+            )
+        return self._checkpoints
```
cdkini, line 556: This is consistent with the suite factory but do we want to just fail fast? Can a user use GX without checkpoints? How about suites?

The only time this would actually pop up if is a user was passing their own context config into the context constructor (and purposely omitted store information).

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -592,43 +607,7 @@ def validations_store(self) -> ValidationsStore:
...
-                error_message: str = (
-                    f"Attempted to access the 'checkpoint_store_name' field "
-                    f"with no `checkpoints` directory.\n "
-                    f"Please create the following directory: {checkpoint_store_directory}.\n "
-                    f"To use the new 'Checkpoint Store' feature, please update your configuration "
-                    f"to the new version number {float(CURRENT_GX_CONFIG_VERSION)}.\n  "
-                    f"Visit {AbstractDataContext.MIGRATION_WEBSITE} "
-                    f"to learn more about the upgrade process."
-                )
-            else:
```
cdkini: I believe this is a legacy remnant from when checkpoints were a new feature.

TLDR: We treat checkpoints as an optional store and ask users to migrate if they're missing it. This object seems pretty integral to how we do validations now so let's remove this cruft and treat it the same as other stores.

great_expectations/expectations/core/expect_column_kl_divergence_to_be_less_than.py
```diff
@@ -226,11 +222,14 @@ def get_validation_dependencies(
                 # Note: 20201116 - JPC - the execution engine doesn't provide capability to evaluate
                 # validation_dependencies, so we use a validator
                 #
+                from great_expectations.validator.metrics_calculator import (
+                    MetricsCalculator,
+                )
+
```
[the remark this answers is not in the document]
cdkini: There are some cyclic import issues with checkpoints so this was necessary - hoping we get rid of things like this as we delete more code

tests/core/factory/test_checkpoint_factory.py
```diff
@@ -0,0 +1,238 @@
...
+        ],
+    }
+
+
+def _assert_checkpoint_equality(actual: Checkpoint, expected: Checkpoint):
+    # Checkpoint equality is currently defined as equality of the config
+    # TODO: We should change this to a more robust comparison (instead of memory addresses)
+    actual_config = actual.config.to_json_dict()
+    expected_config = expected.config.to_json_dict()
+    assert actual_config == expected_config
```
cdkini, line 41: Not sure how often a user will need to use `__eq__` but this feels bad right now

tests/data_context/test_data_context.py
```diff
@@ -963,38 +923,6 @@ def test_scaffold_directories(tmp_path_factory):
     }
 
 
-@pytest.mark.filesystem
-def test_build_batch_kwargs(titanic_multibatch_data_context):
```
cdkini, line 964: This test and its related fixture are no longer relevant - just deleting them rather than updating

tests/test_fixtures/great_expectations_basic.yml
```diff
@@ -1,6 +1,6 @@
 # This is a basic configuration for testing.
 # It has comments that should be preserved.
-config_version: 2
+config_version: 3
```
cdkini, line 3: Something panics if we reference checkpoint stores before this version

great_expectations/core/factory/factory.py
```diff
@@ -0,0 +1,22 @@
+from abc import ABC, abstractmethod
+from typing import Generic, TypeVar
+
+T = TypeVar("T")
+
+
+class Factory(ABC, Generic[T]):
```
tyler-hoffman, line 7: I'm not strongly opinionated on this, but is this class YAGNI?

Given that it's basically just an interface at this point, the main use cases would be either a) if we had class/methods that took in Factories without knowing the concrete implementation, or b) to help hold ourselves to the same interface when implementing new Factories (assuming we decide to opt in to inheriting from Factory).

I think my biggest (though still minor) concern is that we define an interface here and then still end up needing workarounds for things, like how our interface for stores didn't hold up so well.

tests/core/factory/test_checkpoint_factory.py
```diff
@@ -0,0 +1,238 @@
...
+        "action_list": [
+            {
+                "action": {"class_name": "StoreValidationResultAction"},
+                "name": "store_validation_result",
+            },
+        ],
+    }
+
+
+def _assert_checkpoint_equality(actual: Checkpoint, expected: Checkpoint):
```
tyler-hoffman, line 36: Should this just be defined as `__eq__` on `Checkpoint?

tests/core/factory/test_checkpoint_factory.py
tyler-hoffman: Hah just left a comment above about __eq__. Why does it feel bad?

tests/core/factory/test_checkpoint_factory.py
```diff
@@ -0,0 +1,238 @@
...
+        DataContextError, match=f"Checkpoint with name {name} was not found."
+    ):
+        factory.get(name=name)
+
+    # Assert
+    store.get.assert_not_called()
+
+
+@pytest.mark.unit
+def test_checkpoint_factory_add_uses_store_add():
```
tyler-hoffman, line 88: `add` returns the checkpoint, right? We should probably have a test asserting that.

tests/profile/fixtures/great_expectations_titanic_0.13.yml
```diff
@@ -8,7 +8,7 @@
 
 # config_version refers to the syntactic version of this config file, and is used in maintaining backwards compatibility
 # It is auto-generated and usually does not need to be changed.
-config_version: 2.0
```
tyler-hoffman, line 11: Do we still support config version 2? This change seems reasonable, but hoping we have other tests around 2.0

tests/core/factory/test_checkpoint_factory.py
cdkini: Having to go into the config attr and then having to convert that to a dict to get equality working seems like a lot of steps - I'd imagine I could just do:
```python
assert Checkpoint("a") == Checkpoint("a") # Fails right now
```

great_expectations/core/factory/factory.py
cdkini: Hmm yeah I was torn or whether or not to use it but I added it as a way to hopefully disincentivize bloating these factories and ensuring consistency between them.

If we need workarounds maybe we can revisit? I can go either way here

tests/core/factory/test_checkpoint_factory.py
cdkini: hmm maybe but I'm not sure how the structure of that object will change. I can follow up with a PR if we want

tests/profile/fixtures/great_expectations_titanic_0.13.yml
cdkini: I think not; moving forward we might just want to axe the config version entirely? Might be worth a larger discussion

tests/core/factory/test_checkpoint_factory.py
cdkini: Will write one!

great_expectations/core/factory/checkpoint_factory.py
```diff
@@ -0,0 +1,93 @@
...
+        key = self._store.get_key(name=checkpoint.name, id=None)
+        if self._store.has_key(key=key):
+            raise DataContextError(
+                f"Cannot add Checkpoint with name {checkpoint.name} because it already exists."
+            )
+
+        self._store.add(key=key, value=checkpoint.get_config())
+
+        # TODO: Add id adding logic to CheckpointStore to prevent round trip
+        return self._get(key=key)
```
cdkini, line 49: Currently round tripping to get ID.

Ideally, we have the store add the ID to the input object but this is additional work. I can fast follow with a PR to update store logic but thought this was already a hefty PR.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -591,43 +606,20 @@ def validations_store(self) -> ValidationsStore:
...
-                    f"with no `checkpoints` directory.\n  "
-                    f"Please create a `checkpoints` directory in your Great Expectations directory."
-                    f"To use the new 'Checkpoint Store' feature, please update your configuration "
-                    f"to the new version number {float(CURRENT_GX_CONFIG_VERSION)}.\n  "
-                    f"Visit {AbstractDataContext.MIGRATION_WEBSITE} "
-                    f"to learn more about the upgrade process."
-                )
+        name = self.variables.checkpoint_store_name
+        if not name and CheckpointStore.default_checkpoints_exist(directory_path=self.root_directory):  # type: ignore[arg-type]
+            name = DataContextConfigDefaults.DEFAULT_CHECKPOINT_STORE_NAME.value
```
cdkini: Some more cleanup will probably happen here if we clean up these store name keys

great_expectations/data_context/store/configuration_store.py
cdkini: The hope is that all stores are consistent in their APIs eventually (we have a lot of wrapper convenience methods that create inconsistency and hurt polymorphism)

tests/core/factory/test_checkpoint_factory.py
cdkini: Actually think we have what we need with the tests below - let me know if you disagree

great_expectations/core/factory/checkpoint_factory.py
tyler-hoffman: Can you make sure we have a ticket before closing this one out?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -474,7 +481,7 @@ def instance_id(self) -> str:
             if self._in_memory_instance_id is not None:
                 return self._in_memory_instance_id
             instance_id = str(uuid.uuid4())
-            self._in_memory_instance_id = instance_id
+            self._in_memory_instance_id = instance_id  # type: ignore[assignment]
```
tyler-hoffman: I think we can fix this by defining this field as `str | None` on line 269. I would do that rather than introduce a type ignore.

Side-rant: we should not be assigning anything in property, but let's leave that for another PR.

tyler-hoffman: Actually rethinking this... we can just do that fallback to the uuid within the constructor, right? I think we should do that and save ourselves a bunch of hoop jumping.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -643,33 +635,7 @@ def checkpoint_store_name(self, value: str) -> None:
...
-                return self._build_store_from_config(  # type: ignore[return-value]
-                    checkpoint_store_name,
-                    DataContextConfigDefaults.DEFAULT_STORES.value[  # type: ignore[arg-type]
-                        checkpoint_store_name
-                    ],
-                )
-            raise gx_exceptions.StoreConfigurationError(
-                f'Attempted to access the Checkpoint store: "{checkpoint_store_name}". It is not a configured store.'
-            )
+        return self.stores[self.checkpoint_store_name]
```
tyler-hoffman, line 635: Should this be wrapped with a nice error if the store doesn't exist?

tests/datasource/fluent/_fake_cloud_api.py
```diff
@@ -719,14 +719,49 @@ def post_checkpoints_cb(request: PreparedRequest) -> CallbackResult:
...
+            break
+
+    if not checkpoint_id:
+        errors = ErrorPayloadSchema(
+            errors=[{"code": "mock 404", "detail": None, "source": None}]
+        )
+        return CallbackResult(404, headers=DEFAULT_HEADERS, body=errors.json())
+
+    deleted_cp = checkpoints.pop(checkpoint_id)
+    print(pf(deleted_cp, depth=5))
```
tyler-hoffman, line 758: `print`

tyler-hoffman (commented):
Left a couple smaller issues. I think we should fix that type around `instance_id` and remove the `print` before merging.

tests/datasource/fluent/_fake_cloud_api.py
cdkini: this seems to be the pattern with other callbacks so I kept it to be consistent but let me know if you disagree

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: Hmm so we should always have a checkpoint store name moving forward. This property has the same structure as all other store properties now

tyler-hoffman (approved):
Approved pending the discussed changes

PR #3377, opened 2021-09-09
[BUGFIX] Addresses issue #2993 (#3054) by using configuration when it is available instead of discovering keys (listing keys) in existing sources.

donaldheppner:
Changes proposed in this pull request:
- Use configuration to identify where key files should be in an existing store rather than querying the store for a list of potential keys and choosing those that match configuration.

Previous Design Review notes:
- Analysis of potential options can be found in issue comments: https://github.com/great-expectations/great_expectations/issues/3054

### Definition of Done
- [x] My code follows the Great Expectations [style guide](https://docs.greatexpectations.io/en/latest/contributing/style_guide.html?highlight=style%20guide)
- [x] I have performed a [self-review](https://docs.greatexpectations.io/en/latest/contributing/contribution_checklist.html?highlight=checklist) of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added [unit tests](https://docs.greatexpectations.io/en/latest/contributing/testing.html#contributing-testing-writing-unit-tests) where applicable and made sure that new and existing tests are passing.
- [x] I have run any local integration tests and made sure that nothing is broken.

PR #9431, opened 2024-02-09
[BUGFIX] Misconfigured Expectations affecting unassociated Checkpoints

NathanFarmer:
If an Expectation Suite is misconfigured, it will break all Checkpoints in the Data Context, even if they aren't associated with the Checkpoint. The reason this happens is that a user can reference metric values from other Expectation Suite in their evaluation parameters. This causes use to parse all Expectation Suites in the context. This PR introduces try/except logic as a way to address the issue of misconfigured Suites causing other Checkpoints to error.

We attempted to address this [here](https://github.com/great-expectations/great_expectations/pull/3377), but this broke the ability to use evaluation parameters in other Expectation Suites, so we reverted it.


tests/test_ge_utils.py
```diff
@@ -78,15 +78,11 @@ def test_validate_non_dataset(file_data_asset, empty_expectation_suite):
     with pytest.raises(
         ValueError, match=r"The validate util method only supports dataset validations"
     ):
-        with pytest.warns(
-            Warning,
-            match="No great_expectations version found in configuration object.",
-        ):
```
NathanFarmer, line 84: I don't think this has anything to do with my PR. It seems to be due to an inconsistency between `develop` and `0.18.x`. Regardless, I don't think we need the warning if we are raising an error.

tests/integration/cloud/end_to_end/test_pandas_filesystem_datasource.py
```diff
@@ -201,7 +201,7 @@ def test_interactive_validator(
         expectation_suite_name=expectation_suite_name
     )
 
-    assert len(expectation_suite.expectations) == expectation_count
```
NathanFarmer, line 204: This was just wrong and inconsistent with the other end-to-end tests. Must be a side-effect of merges between `0.18.x` and `develop`. It is not related to my PR.

tests/test_ge_utils.py
NathanFarmer: Confirmed this is failing on `0.18.x` without my changes.

tests/integration/cloud/end_to_end/test_pandas_filesystem_datasource.py
NathanFarmer: Confirmed this is failing on `0.18.x` without my changes.

cdkini (approved):

PR #9434, opened 2024-02-12
[BUGFIX] Fix sphinx linx

tyler-hoffman:
This PR addresses a bug resulting from the combo of sphinx, docusaurus limitations around url fragment support, and recent docs build changes.

## The problem
We had a step in the sphinx build that created absolute links, and it defaulted the host to `localhost:3000`. This was to get around docusaurus's lack of support for url fragments. With the new docs build process, we have removed on-the-fly file processing, so these docs were build with localhost links that would error for users when deployed.

## The fix
* Use relative links
* Don't use url fragments in these links. This ends up being fine because all of the links' fragments were to the signature of the class that the link was to. So the functional change is that we link to the top of the given page, rather than scrolled a tiny bit down to the class constructor's signature.

## Notes
* This does not fix links for previous versions of docs. We'll have to update those in a followup PR.
* This affects files generated under `docs/docusaurus/docs/reference/api`, which are gitignored, so the changes don't show up in this PR, but will locally if you check the files. I want to change the gitignore, but it's out of scope for this PR.

## Summary by CodeRabbit

- **Documentation**
	- Improved internal link handling in API documentation for enhanced compatibility with Docusaurus.


anthonyburdi (approved):
Thank you for this change! 
Also if you ever get the urge, the `_parse_and_process_html_to_mdx` method is ripe for refactor, it ended up being much longer than initially planned.

docs/sphinx_api_docs_source/build_sphinx_api_docs.py
```diff
@@ -253,36 +253,35 @@ def _parse_and_process_html_to_mdx(
...
         # Process internal links
-        # Note: Currently the docusaurus link checker does not work well with
-        # anchor links, so we need to make these links absolute.
-        # We also need to use the shortened path version to match
-        # what is used for the page url.
+        # There are a couple things we need to do here:
+        #  - Use relative links. This solves two problems
+        #     - Ensure links continue to work even after cutting new docs versions
+        #     - Ensure links continue to work regardless of host url
+        #  - Remove fragments from links; docusaurus currently errors on these.
```
anthonyburdi, line 260: 🙏

PR #9439, opened 2024-02-12
[BUGFIX] Add 0.17 to docs links for 0.17

tyler-hoffman:
[This is an approach to ](https://github.com/great-expectations/great_expectations/pull/9429) updated sphinx-generated links to reference prod's docs directory, but that could break for older versions, namely 0.17, as a class from 0.17 would have a link that looks into 0.18 docs. This PR attempts to fix that in a quick/semi-dirty way by just updating links to look in the docs/0.17.23 path.

I *think* this is safe because we should be ensuring we don't break backward compatibility with old links (e.g. if we moved 0.17 docs, we'd provide a redirect from the old path to the new one)


billdirks (approved):
LGTM. Thanks for doing this. mdx files with long lines are hard for me to review so I wasn't thorough.

PR #9456, opened 2024-02-15
[BUGFIX] - Prevent duplicate Expectations in Validation Results when Exceptions are triggered

josectobar:
Fix bug where duplicate Expectation in a Expectation Validation Result were getting added when error exceptions were captured. This was being caused when the `aborted_exception_info` was getting duplicated on retry (3x for each retry). 
Changes:
- Only one error is capture on every retry, overwriting the previous one.
- A dictionary with the metric id as string key and an exception is added into an Expectation Validation Result instead of adding an EVR for each exception.


NathanFarmer (commented):
Looks good so far. I will wait for tests to pass, because there are a lot more to run.

tests/validator/test_validator.py
```diff
@@ -1594,3 +1595,69 @@ def test_graph_validate_with_two_expectations_and_first_expectation_with_result_
             exception_info=None,
         ),
     ]
+
+
+@pytest.mark.big
```
NathanFarmer: `big` is for Big Query, I believe. This test doesn't actually rely on any services does it? I think this can be marked `@pytest.mark.unit` and then it will run against all supported versions of python.

NathanFarmer (approved):
Looks good! It should be pretty trivial to update the mysql tests.

tests/validator/test_validator.py
josectobar: haha I thought big meant something else.

PR #9459, opened 2024-02-15
[BUGFIX] Patch Pandas/SQLAlchemy Snowflake issue

cdkini:
equivalent to v0.18 fix: https://github.com/great-expectations/great_expectations/pull/9458

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -467,9 +467,6 @@ filterwarnings = [
     'ignore: distutils Version classes are deprecated. Use packaging.version instead.:DeprecationWarning',
 
     # pandas
-    # Example Actual Warning: UserWarning: pandas only supports SQLAlchemy connectable (engine/connection) or database string URI or sqlite3 DBAPI2 connection. Other DBAPI2 objects are not tested. Please consider using SQLAlchemy.
-    'ignore: pandas only supports SQLAlchemy connectable (engine/connection) or database string URI or sqlite3 DBAPI2 connection:UserWarning',
-    # Found when runing pytest tests/test_definitions/test_expectations_v3_api.py
     # Example Actual Warning: FutureWarning: In a future version of pandas, parsing datetimes with mixed time zones will raise an error unless `utc=True`. Please specify `utc=True` to opt in to the new behaviour and silence this warning. To create a `Series` with mixed offsets and `object` dtype, please use `apply` and `datetime.datetime.strptime`
     # Found when running pytest tests/test_definitions/test_expectations_v3_api.py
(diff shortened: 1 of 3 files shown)


Kilo59 (approved):

NathanFarmer (dismissed):
This is pinned for all dialects. We decided that was fine for `0.18`, but I'm not sure if we want to do this for `1.0` or just fix the code.

tests/test_packaging.py
```diff
@@ -184,8 +184,8 @@ def test_polish_and_ratchet_pins_and_upper_bounds():
     )
 
     # Polish and ratchet this number down as low as possible
-    assert len(sorted_packages_with_pins_or_upper_bounds) == 81
-    assert sorted_packages_with_pins_or_upper_bounds == [
+    assert len(sorted_packages_with_pins_or_upper_bounds) == 85
+    assert set(sorted_packages_with_pins_or_upper_bounds) == {
```
Kilo59, line 188: @anthonyburdi might have something to say about this, but we can deal with that later.

pyproject.toml
```diff
@@ -467,9 +467,6 @@ filterwarnings = [
     'ignore: distutils Version classes are deprecated. Use packaging.version instead.:DeprecationWarning',
 
     # pandas
-    # Example Actual Warning: UserWarning: pandas only supports SQLAlchemy connectable (engine/connection) or database string URI or sqlite3 DBAPI2 connection. Other DBAPI2 objects are not tested. Please consider using SQLAlchemy.
-    'ignore: pandas only supports SQLAlchemy connectable (engine/connection) or database string URI or sqlite3 DBAPI2 connection:UserWarning',
-    # Found when runing pytest tests/test_definitions/test_expectations_v3_api.py
```
Kilo59, line 472: This should be fine to delete but you'll need to add this...

```toml
    # Example Actual Warning: FutureWarning: In a future version of pandas, parsing datetimes with mixed time zones will raise an error unless `utc=True`. Please specify `utc=True` to opt in to the new behaviour and silence this warning. To create a `Series` with mixed offsets and `object` dtype, please use `apply` and `datetime.datetime.strptime`
    'ignore: In a future version of pandas, parsing datetimes with mixed time zones will raise an error:FutureWarning',
```

tests/test_packaging.py
anthonyburdi: I think pinning pandas is fine as a temporary measure. If it is only that specific version you can use the `!=` instead of `<`, the benefit being we would not have to remember to remove the upper bound after fixing the underlying issue if it is only present in one specific version.

The list -> set change I think is fine. It was only a list to make it a little more human readable.

If we move to poetry I think this test goes away.

PR #9460, opened 2024-02-15
[FEATURE] Add backend-agnostic partitioners

joshua-stauffer:
This PR adds a set of backend-agnostic partitioners to be used within the context of the BatchConfig. Assets will then translate these generic partitioners into specific partitioners known to the asset.


great_expectations/core/partitioners.py
```diff
@@ -0,0 +1,64 @@
+from typing import List, Literal, Union
+
+from great_expectations.compatibility import pydantic
+
+
+class PartitionerYear(pydantic.BaseModel):
+    column_name: str
+    method_name: Literal["partition_on_year"] = "partition_on_year"
```
tyler-hoffman, line 8: Non-blocking, but a couplethings on these method names:
* Can we remove them from the constructor's signature altogether? Do pydantic v1 fields support `init=False`?
* Or should it be a ClassVar?
* Even though it's a literal, it's still a magic string, in a followup, should we connect these with the backend via an enum or something along those lines?

great_expectations/core/partitioners.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class PartitionerMultiColumnValue(pydantic.BaseModel):
+    column_names: List[str]
+    method_name: Literal[
+        "partition_on_multi_column_values"
+    ] = "partition_on_multi_column_values"
+
+
+Partitioner = Union[
```
tyler-hoffman, line 55: Very non-blocking, and I think we discussed this synchronously, but is there a benefit of building the union vs inheriting from a `Partitioner` ABC? If we went the ABC route, and moved to using enums, we could potentially increase our type safety with something like.

```
_T = TypeVar("_T", bound=MethodName)

class Partitioner(Generic[_T], ABC, pydantic.BaseModel):
  method_name: _T          # though IIRC, we'd still have to set the default in the child class.
```

tyler-hoffman (approved):

great_expectations/core/partitioners.py
joshua-stauffer: thanks for these suggestions - looks like both could potentially be implemented without changing behavior outside of the `partitioners` module, so i'll look into them as a followup PR 👍

PR #9482, opened 2024-02-16
[FEATURE] Add Partitioner to BatchRequest

joshua-stauffer:
This PR allows Partitioners to be passed from a BatchConfig to a DataAsset via the BatchRequest.


great_expectations/datasource/fluent/batch_request.py
```diff
@@ -83,6 +86,7 @@ class BatchRequest(pydantic.BaseModel):
     _batch_slice_input: Optional[BatchSlice] = pydantic.PrivateAttr(
         default=None,
     )
+    partitioner: Optional[Partitioner] = None
```
tyler-hoffman: I'd move this up above the private attr and update the docstring's args.

great_expectations/datasource/fluent/file_path_data_asset.py
```diff
@@ -59,6 +59,7 @@
 
     from great_expectations.core.batch import BatchDefinition, BatchMarkers
     from great_expectations.core.id_dict import BatchSpec
+    from great_expectations.core.partitioners import Partitioner as AbstractPartitioner
```
tyler-hoffman: `Abstract` seems like the wrong name. What if when we import the spark partitioner, we alias it as SparkPartitioner, and leave this one as just `Partitioner`?

great_expectations/datasource/fluent/file_path_data_asset.py
```diff
@@ -172,6 +173,7 @@ def build_batch_request(
         self,
         options: Optional[BatchRequestOptions] = None,
         batch_slice: Optional[BatchSlice] = None,
+        partitioner: Optional[AbstractPartitioner] = None,
```
tyler-hoffman: The DataAsset this is on only uses spark partitioners. Should this method only accept spark partitioners? I assume that might mean making the base class a generic

cdkini: How are aliases parsed with `@public_api`?

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -198,6 +202,11 @@ def build_batch_request(
                 "batch_slice is not currently supported and must be None for this DataAsset."
             )
 
+        if partitioner is not None:
+            raise ValueError(
+                "partitioner is not currently supported and must be None for this DataAsset."
```
cdkini, line 207: Nit - could we capitalize "partitioner"

cdkini: Or actually if you're referring to the arg I think its fine - good either way

cdkini (approved):

great_expectations/datasource/fluent/pandas_datasource.py
joshua-stauffer: thanks - yes, was following the style in comments above, which refer to the arg names.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -65,6 +65,7 @@
     from typing_extensions import Self
 
     from great_expectations.compatibility import sqlalchemy
+    from great_expectations.core.partitioners import Partitioner as AbstractPartitioner
```
tyler-hoffman: Same as above on `Abstract`

great_expectations/datasource/fluent/file_path_data_asset.py
joshua-stauffer: that's an interesting question - I'm not sure

tyler-hoffman (changes requested):
Requesting changes on not using the alias `AbstractPartitioner`, since that implies an ABC.

I'm also strongly interested in making DataAsset generic wrt the partitioner it allows. If we made that type bound to `Partitioner | None`, I think that would even solve the pandas situation gracefully without the need to raise.

great_expectations/datasource/fluent/file_path_data_asset.py
joshua-stauffer: okay, i like that - will update

great_expectations/datasource/fluent/file_path_data_asset.py
joshua-stauffer: i took a slightly different approach - users will only ever interact with the generic partitioner classes, and the data asset translates it into an asset-specific implementation. If an asset doesn't have an implementation for that particular partitioner, we raise an informative error.

tyler-hoffman: Ah right! Carry on!

tyler-hoffman (approved):

PR #9486, opened 2024-02-16
[FEATURE] Don't break context for invalid datasource configs

Kilo59:
Datasources that fail to deserialize due to a `pydantic.ValidationError` or `TypeLookup` `LookupError` are now deserialized as an `InvalidDatasource`.

The `InvalidDatasource` has an attribute called `config_error` which retains the original pydantic ValidationError.
This error is exposed to the user when methods such as `test_connection()` are invoked.

Users are also warned if there are `InvalidDatasource`s present in their config, but these no longer prevent the users from working with other "valid" datasources.

## TODO
- [x] Only save `name`, `type` (and same for `assets`) to prevent leaking secrets


great_expectations/data_context/store/datasource_store.py
```diff
@@ -118,7 +126,14 @@ def deserialize(
                 datasource_model = _SourceFactories.type_lookup.get(type_)
                 if not datasource_model:
                     raise LookupError(f"Unknown Datasource 'type': '{type_}'")
-                return datasource_model(**value)
+                try:
+                    return datasource_model(**value)
+                except PydanticValidationError as config_error:
+                    warnings.warn(
+                        f"Datasource {value.get('name', '')} configuration is invalid. Check `.config_error` attribute for more details.",
```
Kilo59, line 133:
```suggestion
                        f"Datasource {value.get('name', '')} configuration is invalid. Check `my_datsource.config_error` attribute for more details.",
```

tests/datasource/fluent/test_invalid_datasource.py
```diff
@@ -0,0 +1,145 @@
...
+    TestConnectionError,
+)
+from great_expectations.datasource.fluent.sources import _SourceFactories
+
+_EXCLUDE_METHODS: Final[set[str]] = {
+    # we don't care about overriding these methods for InvalidDatasource
+    "copy",
+    "delete_asset",
+    "delete_batch_config",
+    "dict",
```
Kilo59, line 28: We might need to override `.dict()`, `.yaml()`, `.json()` etc. in order to mask potential secrets from an invalid config.

I will create a followup ticket for this. This PR is targeted against `develop` not `0.18.x` so it won't cause any issues right away.

great_expectations/datasource/fluent/invalid_datasource.py
```diff
@@ -0,0 +1,112 @@
...
+class GxInvalidDatasourceWarning(GxDatasourceWarning):
+    """
+    A warning that the Datasource configuration is invalid and will must be updated before it can used.
+    """
+
+
+class InvalidAsset(DataAsset):
+    """A DataAsset that is invalid."""
+
+    type: str = "invalid"  # TODO: ensure this isn't registered as a real data asset
```
Kilo59, line 31:
```suggestion
    type: str = "invalid"
```

great_expectations/datasource/fluent/invalid_datasource.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class GxInvalidDatasourceWarning(GxDatasourceWarning):
+    """
+    A warning that the Datasource configuration is invalid and will must be updated before it can used.
+    """
+
+
+class InvalidAsset(DataAsset):
+    """A DataAsset that is invalid."""
```
Kilo59, line 29:
```suggestion
    """
    A DataAsset that is invalid.
    The DataAsset itself may be valid, but it is classified as invalid because its parent Datasource or sibling assets are invalid.
    """
```

great_expectations/datasource/fluent/invalid_datasource.py
```diff
@@ -0,0 +1,115 @@
...
+        ) from self.config_error
+
+    @override
+    def get_asset(self, asset_name: str) -> InvalidAsset:
+        """
+        Always raise a warning and return an InvalidAsset.
+        Don't raise an error because the users may want to inspect the asset config.
+        """
+        warnings.warn(
+            f"The {self.name} Datasource configuration for this asset is invalid and cannot be used. Please fix the error and try again",
```
Kilo59, line 102:
```suggestion
            f"The {self.name} Datasource configuration is invalid and cannot be used. Please fix the error and try again",
```

great_expectations/datasource/fluent/invalid_datasource.py
```diff
@@ -0,0 +1,115 @@
...
+    The overridden methods should indicate to the user that the Datasource configuration is invalid and provide details about
+    why it was considered invalid.
+
+    Any errors raised should raise `from self.config_error`.
+    """
+
+    # class var definitions
+    asset_types: ClassVar[List[Type[DataAsset]]] = [InvalidAsset]
+    _type_lookup: ClassVar[TypeLookup] = InvalidAssetTypeLookup()
+    type: str = "invalid"
```
Kilo59, line 76:
```suggestion
    _type_lookup: ClassVar[TypeLookup] = InvalidAssetTypeLookup()
   
    type: str = "invalid"
```

tests/datasource/fluent/test_invalid_datasource.py
```diff
@@ -0,0 +1,145 @@
...
+
+@pytest.mark.unit
+@pytest.mark.parametrize(
+    "invalid_ds_cfg",
+    [
+        pytest.param(
+            {
+                "name": "my pg",
+                "type": "postgres",
+                "connection_string": "postmalone+psycopg2://postgres:@localhost/test_database",
```
NathanFarmer, line 125: 😆

great_expectations/datasource/fluent/invalid_datasource.py
```diff
@@ -0,0 +1,116 @@
...
+    @overload
+    def __getitem__(self, key: str) -> Type:
+        ...
+
+    @overload
+    def __getitem__(self, key: Type) -> str:
+        ...
+
+    @override
+    def __getitem__(self, key: ValidTypes) -> ValidTypes:
```
billdirks, line 123: Why don't we always return an `InvalidAsset`?

great_expectations/datasource/fluent/invalid_datasource.py
```diff
@@ -0,0 +1,116 @@
...
+        if isinstance(key, str):
+            return InvalidAsset
+        return "invalid"
```
billdirks: Does this have to match the type string in InvalidDatasource? If so, could we make this a module level constant?

billdirks (commented):
This looks good. I haven't looked at tests yet. Thanks!

great_expectations/datasource/fluent/invalid_datasource.py
Kilo59: Was trying to satisfy the interface of `TypeLookup` to return an actual `Type`/`class` if a string was provided and the `type: str` value if a `Type / class` was provided.
But in this case, we should never actually need to lookup the `invalid` str for `InvalidType` so I can just raise a `NotImplementedError` in that case because it almost certainly would be being used incorrectly.

great_expectations/datasource/fluent/invalid_datasource.py
Kilo59: This usage doesn't make sense so (as per this comment) I'm going to raise an error here https://github.com/great-expectations/great_expectations/pull/9486/files#r1495895380

great_expectations/datasource/fluent/invalid_datasource.py
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+    @override
+    def test_connection(self, test_assets: bool = True) -> None:
+        raise TestConnectionError(
+            "This Datasource configuration is invalid and cannot be used. Please fix the error and try again"
+        ) from self.config_error
+
+    @override
+    def get_asset(self, asset_name: str) -> InvalidAsset:
```
billdirks, line 171: I don't think we have any tests for this?  Because of the call to `super.get_asset`, do we really get an `InvalidAsset` or it is always going to throw in the call to `super`?

billdirks: Always throwing may be the right action in this case. Maybe add a comment and a test to validate we throw?

Kilo59: `get_asset()` is just iterating over the list of assets and matching on a name. That part will always succeed; the thing that is potentially missing here is `overriding` the methods on `InvalidAsset` to raise errors when called.

I've just added a test for `InvalidDatasource.get_asset()` and implemented `test_connection()` for `InvalidAsset`.

great_expectations/datasource/fluent/metadatasource.py
```diff
@@ -31,7 +31,7 @@ def __new__(  # noqa: PYI034 # Self cannot be used with Metaclass
 
         cls = super().__new__(meta_cls, cls_name, bases, cls_dict)
 
-        if cls_name == "Datasource" or cls_name.startswith("_"):
+        if cls_name in ["Datasource", "InvalidDatasource"] or cls_name.startswith("_"):
```
Kilo59:
```suggestion
        if cls_name in ("Datasource", "InvalidDatasource") or cls_name.startswith("_"):
```

tests/datasource/fluent/test_invalid_datasource.py
Kilo59: Dealt with this by ignoring extra fields.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -526,6 +526,7 @@ def _load_asset_subtype(
         """
         logger.debug(f"Loading '{data_asset.name}' asset ->\n{pf(data_asset, depth=4)}")
         asset_type_name: str = data_asset.type
+        # TODO: 2024-2-22 - (GG:kilo59) - raise a more informative error if the asset type is not found
```
Kilo59, line 529:
```suggestion
```

great_expectations/data_context/store/datasource_store.py
```diff
@@ -113,12 +121,18 @@ def deserialize(
...
             if type_:
-                datasource_model = _SourceFactories.type_lookup.get(type_)
-                if not datasource_model:
-                    raise LookupError(f"Unknown Datasource 'type': '{type_}'")
-                return datasource_model(**value)
+                try:
+                    datasource_model = _SourceFactories.type_lookup[type_]
+                    return datasource_model(**value)
+                except (PydanticValidationError, LookupError) as config_error:
+                    warnings.warn(
```
Kilo59, line 128: @billdirks Since you last reviewed, I added handling for an error due to a `TypeLookup` error. When we talked last, I said I would do this as follow-up work, but it ended up being simpler than I expected.

The 2 scenarios that this covers are.
a. The datasource has a `type` that is not recognized.
b. An asset for a dynamically generated Datasource (pandas) has a type that is not recognized.

great_expectations/datasource/fluent/invalid_datasource.py
```diff
@@ -0,0 +1,176 @@
...
+        return super().get_asset(asset_name)
+
+    @override
+    def get_batch_list_from_batch_request(
+        self, batch_request: BatchRequest
+    ) -> NoReturn:
+        raise TypeError(f"{self.name} Datasource is invalid") from self.config_error
+
+    @override
+    def add_batch_config(self, batch_config: BatchConfig) -> NoReturn:
```
billdirks: You could implement `__getattr__` to catch any other calls that are made on the datasource.

billdirks (approved):
Thanks, LGTM. I left one small comment.

great_expectations/datasource/fluent/invalid_datasource.py
Kilo59: Yeah great idea, I'll do this. We did something similar for block-style datasources when users try to call a Fluent method.

Also much less tedious to maintain.

PR #9493, opened 2024-02-19
[DOCS] Fix interactions with versioning dropdown

JessSaavedra:


ghost:
<details open>
<summary><b>👇 <a href="https://app.codesee.io/r/reviews?pr=9493&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations">Click on the image</a> for a new way to code review</b>
</summary>

####

[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=9493&src=https%3A%2F%2Fgithub.com%2Fgreat-expectations%2Fgreat_expectations)

#### Legend
CodeSee Map legend
</details>

docs/docusaurus/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.js
```diff
@@ -0,0 +1,91 @@
...
+          description:
+            'The label for the navbar versions dropdown on mobile view',
+        })
+      : dropdownVersion.label;
+  const dropdownTo =
+    mobile && items.length > 1
+      ? undefined
+      : getVersionMainDoc(dropdownVersion).path;
+
+  // If section corresponds to homepage or GX Cloud, version dropdown should not be shown
```
JessSaavedra, line 61: These lines are the only real additions, the rest is just the original code of the component because of it being swizzled

docs/docusaurus/src/theme/NavbarItem/DropdownNavbarItem.js
```diff
@@ -0,0 +1,142 @@
...
+      ref={dropdownRef}
+      className={clsx('navbar__item', 'dropdown', 'dropdown--hoverable', {
+        'dropdown--right': position === 'right',
+        'dropdown--show': showDropdown,
+      })}>
+      <NavbarNavLink
+        aria-haspopup="true"
+        aria-expanded={showDropdown}
+        role="button"
+        href='#'
```
JessSaavedra, line 62: This line is the only change, the rest is just the original code of the component because of it being swizzled

docs/docusaurus/src/theme/NavbarItem/DropdownNavbarItem.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import DropdownNavbarItem from '@theme-original/NavbarItem/DropdownNavbarItem';
+
+export default function DropdownNavbarItemWrapper(props) {
+  return (
+    <DropdownNavbarItem {...props} href={'#'} />
```
JessSaavedra, line 8: This href is added to prevent dropdown to redirect to inner link when clicked (it only should make a redirection when a version inside the dropdown options is clicked)

docs/docusaurus/src/css/navbar.scss
```diff
@@ -86,6 +93,11 @@
 
 .navbar__link--active {
   border-bottom: 2px solid var(--ifm-color-primary);
+
+  &[href*="0.17"] {
```
JessSaavedra, line 96: This is added to prevent all three tabs (GX OSS, learn and api) from being selected at the samte in version 0.17. The source problem relies on a retrocompatibility problem due to extensive changes in docs structure from version 0.17 to 0.18 which will be solved by cutting this version out of the current project (similar to what has been done with version 0.16)

docs/docusaurus/versioned_sidebars/version-0.17-sidebars.json
```diff
@@ -1,227 +1,6 @@
 {
   "docs": [
     "intro",
-    {
```
tyler-hoffman: Am I correct in assuming we'll have to make this same change to the versioned sidebars every time we cut a new release? I'm concerned about the maintainability of that approach.

docs/docusaurus/src/theme/NavbarItem/DropdownNavbarItem.js
tyler-hoffman: I'm not sure I follow. What is the behavior with and without this change? And if we need this, do you mind throwing some comments in the code about this? I think this might be a source of confusion to future readers.

docs/docusaurus/src/css/navbar.scss
```diff
@@ -25,6 +25,13 @@
   box-shadow: unset;
   height: unset;
   padding: 0;
+
+  &:has(.home-section.navbar__link--active),
+  &:has(.gx-cloud-section.navbar__link--active) {
+    .navbar__items .dropdown {
+      display: none;
```
tyler-hoffman: So if if I follow, we are adding classnames in the docusaurus config for the cloud and home section, then using the presence of children with those classnames that are active to tell determine if we should hide the dropdown.

I'm nervous about discoverability here. Could / should this type of logic be in js (But maybe with url inspection rather than classnames)? This just feels like more logic than I like to see in CSS. If we do end up needing to go with the css approach, maybe changing those classnames to something that describes their purpose (e.g. `non-versioned`, or something like that)?

tyler-hoffman: There's some jitter when switching between tabs that I think would get resolved by using `display: hidden` instead of `display: none`

docs/docusaurus/src/css/navbar.scss
tyler-hoffman: I don't think I understand. Looks like the links are still clickable, they just aren't styled like they are.

> from being selected at the samte in version 0.17.
Was that supposed to say "at the same time"? Was there just something visually wrong?

I think this chunk of css might need to be fixed up or replaced with a different solution.

docs/docusaurus/src/css/searchBar.scss
```diff
@@ -1,15 +1,19 @@
-.custom-search-bar .DocSearch-Button {
-  background: none;
+.custom-search-bar {
```
tyler-hoffman: This styling looks unrelated to the rest of the PR (same with the change in custom.scss). Would you mind splitting these changes out into a separate PR so this one can be focused only on the versioning concnerns?

tyler-hoffman (changes requested):
Thanks for looking at this. This is a hard problem.

I'm putting a blocker on this for a couple reasons, but if we exhaust other options, I'm open to having my mind changed.

The main concerns here for me are maintainability (specifically around logic in css) and the manual steps this would introduce when cutting new versions of docs. I don't have a great solution to any of this off the top of my head, but I'm happy to chat if you want!

docs/docusaurus/src/css/navbar.scss
JessSaavedra: Yes, they should be clickable, but the problem is that with the fix they're all highlighted at the same time because of the structure of the docs for version 0.17, which is all under a big folder of docs and has no subfolders for the different sections. What you say is correct this is just a visual workaround so it doesn't look weird to habe three tabs with an underline at once, the purpose of this change is to get rid of the underline only for this particular version and we should take this out if we go with the approach of having that version on its own link, as version 0.16 has now

docs/docusaurus/src/theme/NavbarItem/DropdownNavbarItem.js
JessSaavedra: You can see the current behavior in [https://docs.greatexpectations.io/docs/home/]() by clicking the label of the dropdown (not one of the options but the label itself). It makes a weird redirect to why gx cloud page, so here I'm just overriding that behavior so that it stays on the current page. I think this it the simplest solution without having to change the actual implementation of the dropdown from docusaurus, as we're trying to keep that need as a minimum because it's not recommended by the documentation

docs/docusaurus/src/css/searchBar.scss
JessSaavedra: Actually these are not needed anymore by using the approach of `visibility: hidden` instead of `display: none`, so I'm taking them out!

docs/docusaurus/src/css/navbar.scss
JessSaavedra: I know the CSS solution is not the ideal, but this is mainly coming from the need we have to fix this ASAP and we're really limited by the tools that docusaurus allows us to use in terms o versioning or the components, as we're trying to avoid the need of ejecting the dropdown version component which is the ideal place to do the changes but it can bring a lot of problems when updating to newer versions of docusaurus. Maybe we can bring @joshzzheng to discussion here but yesterday we talked and doing this logic in CSS seemed like the best option in this moment

docs/docusaurus/versioned_sidebars/version-0.17-sidebars.json
josectobar: Correct me if I'm wrong but is this change is intended to fix an incompatibility with this earlier version only, right? So going forward we shouldn't need to do this for newer versions is this correct?

docs/docusaurus/src/theme/NavbarItem/DropdownNavbarItem.js
josectobar: I agree with @tyler-hoffman , what is the intent behavior if this `href={"#"}`  I'm not familiarized with this approach.

josectobar (commented):
I left a few comments but overall I agree with @tyler-hoffman.

docs/docusaurus/src/theme/NavbarItem/DropdownNavbarItem.js
tyler-hoffman: @josectobar an href of just `#` means it's a url fragment. When it's not followed by an element id, it's a reference to the top of the page. So this basically hijacks the href to point to the current page, rather than the "why gx cloud" page that it otherwise defaults to. I did a tiny bit of looking to figure out why docusaurus defaults to that, and reported some intermediate findings here: https://greatexpectationslabs.slack.com/archives/C03B8DZCJ07/p1708537374187499.

docs/docusaurus/versioned_sidebars/version-0.17-sidebars.json
JessSaavedra: No, this in particular is because we don't want to version GX cloud documentation, so the idea is to always remove it from all the versions that are not the current one (in this case i'm deleting the one in 0.17 so when you click on GX Cloud tab it always show the one from version 0.18 but we should keep removing them when releasing newer versions, so defintely is not a very maintainable solution in the long term

docs/docusaurus/src/theme/NavbarItem/DropdownNavbarItem.js
JessSaavedra: I'm not able to see the message as I'm not part of the org :eyes: could you paste that info here or send me an invite to join?

docs/docusaurus/src/css/navbar.scss
tyler-hoffman: Note to anyone else looking at this: I'm fine with this approach because it's an improvement, and archiving 0.17 later this week will make fully resolve the issue.

docs/docusaurus/src/theme/NavbarItem/DropdownNavbarItem.js
tyler-hoffman: I'd like to see us figure out a better solution, but I'm okay with this in the short term!

tyler-hoffman (approved):

PR #9499, opened 2024-02-20
[FEATURE] DataAsset uses partitioner from BatchConfig

joshua-stauffer:
This PR updates the DataAsset to use Partitioners defined in the BatchRequest rather than the Partitioner defined on the DataAsset. This change is required since a single asset can now have multiple Partitioners, as defined by multiple BatchConfigs.

## changes
- deprecates `DataAsset.batch_request_options`. Functionality is now available `BatchConfig.get_batch_request_options()`, which invokes `DataAsset.get_batch_request_options_keys(...)`. This has been changed from a property to a method because it potentially triggers a query of the data.
- `PartitionOnConvertedDatetime` is moved alongside the rest of sql partitioners to allow it to join the `SqlPartitioner` union, with the expectation that it will be implemented for other sql backends in the future. This is a workaround to pydantic not allowing child classes to override types from parent classes. Specifically, `SqliteTableAsset` and `SqliteQueryAsset` need to be able to add their implementation of `PartitionerConvertedDateTime` to the `_partitioner_implementation_map`.


great_expectations/core/partitioners.py
```diff
@@ -34,24 +40,36 @@ class PartitionerDividedInteger(pydantic.BaseModel):
...
     ] = "partition_on_multi_column_values"
 
 
+@public_api
+class PartitionerConvertedDatetime(pydantic.BaseModel):
+    column_name: str
+    method_name: Literal[
+        "partition_on_converted_datetime"
+    ] = "partition_on_converted_datetime"
+    date_format_string: str
```
cdkini, line 70: Should we do any validation on this string?

great_expectations/datasource/fluent/file_path_data_asset.py
```diff
@@ -140,6 +164,19 @@ def __init__(self, **data):
...
+        self, abstract_partitioner: Partitioner
+    ) -> SparkPartitioner:
+        PartitionerClass = self._partitioner_implementation_map.get(
+            type(abstract_partitioner)
+        )
+        if PartitionerClass is None:
+            raise ValueError(
+                f"Requested Partitioner `{abstract_partitioner.method_name}` is not implemented for this DataAsset. "
+            )
+        assert PartitionerClass is not None
```
cdkini: Do we need this assert?

great_expectations/core/batch_config.py
```diff
@@ -45,5 +45,10 @@ def build_batch_request(
             options=batch_request_options, partitioner=self.partitioner
         )
 
+    def batch_request_option_keys(self) -> tuple[str, ...]:
```
cdkini: Nit - should this be prefixed with `get_` for consistency?

great_expectations/core/partitioners.py
joshua-stauffer: [current behavior](https://github.com/great-expectations/great_expectations/blob/2bb6081895c1979dfcc74687f88a3b2ff0944978/great_expectations/datasource/fluent/sqlite_datasource.py#L51) is to not validate the string, so i think it's likely okay to not. I'm not 100% how we would validate it, either, since its correctness depends on the shape of the source data.

tests/conftest.py
```diff
@@ -3121,8 +3121,8 @@ def fds_data_context(
     datasource.add_query_asset(
         name="trip_asset_partition_by_event_type",
         query="SELECT * FROM table_partitioned_by_date_column__A",
-    ).add_partitioner_column_value("event_type")
-
+    )
+    # add_partitioner_column_value("event_type")
```
cdkini: Should this be updated to add an actual partitioner?

great_expectations/datasource/fluent/file_path_data_asset.py
joshua-stauffer: i originally added this because mypy wasn't correctly catching that the exception above narrows the type to non null, but i'll double check 🙇

tests/conftest.py
joshua-stauffer: we can no longer add partitioners at the asset level, so it needs to be added where this fixture is used. i can remove this 👍

great_expectations/core/batch_config.py
joshua-stauffer: i'll remove this method entirely. it was meant as a user-facing replacement for the asset property `batch_request_options`, but let's evaluate if we actually want this, and then add it as a followup PR.

tests/conftest.py
joshua-stauffer: removed in [43888ed](https://github.com/great-expectations/great_expectations/pull/9499/commits/43888edfeef513cbc3abbcb2d26ab1cb34fc9fa5)

great_expectations/core/batch_config.py
joshua-stauffer: removed in [9d42d57](https://github.com/great-expectations/great_expectations/pull/9499/commits/9d42d57dd3b7befff07071f70cdcdef8ef0de24d)

great_expectations/datasource/fluent/file_path_data_asset.py
joshua-stauffer: removed in [4b0651e](https://github.com/great-expectations/great_expectations/pull/9499/commits/4b0651e08f11cb67d0d71491e9bacabab8d693b6)

cdkini (approved):

PR #9508, opened 2024-02-21
[FEATURE] V1 Validation scaffolding

cdkini:

### DIFF
--- a/great_expectations/core/__init__.py
+++ b/great_expectations/core/__init__.py
@@ -18,23 +18,25 @@ from .expectation_validation_result import (
 from .id_dict import IDDict
 from .run_identifier import RunIdentifier, RunIdentifierSchema
 from .urn import ge_urn
+from .validation import Validation
 
 __all__ = [
     "Domain",
     "ExpectationSuite",
     "ExpectationSuiteSchema",
-    "expectationSuiteSchema",
     "ExpectationSuiteValidationResult",
     "ExpectationSuiteValidationResultSchema",
     "ExpectationValidationResult",
     "ExpectationValidationResultSchema",
-    "expectationSuiteValidationResultSchema",
-    "expectationValidationResultSchema",
-    "get_metric_kwargs_id",
     "IDDict",
     "RunIdentifier",
     "RunIdentifierSchema",
+    "Validation",
+    "expectationSuiteSchema",
+    "expectationSuiteValidationResultSchema",
+    "expectationValidationResultSchema",
     "ge_urn",
+    "get_metric_kwargs_id",
 ]
 
 logger = logging.getLogger(__name__)
--- a/great_expectations/core/__init__.py
+++ b/great_expectations/core/__init__.py
@@ -18,7 +18,7 @@ from .expectation_validation_result import (
 from .id_dict import IDDict
 from .run_identifier import RunIdentifier, RunIdentifierSchema
 from .urn import ge_urn
-from .validation import Validation
+from .validation_config import ValidationConfig
 
 __all__ = [
     "Domain",
--- a/great_expectations/core/__init__.py
+++ b/great_expectations/core/__init__.py
@@ -31,7 +31,7 @@ __all__ = [
     "IDDict",
     "RunIdentifier",
     "RunIdentifierSchema",
-    "Validation",
+    "ValidationConfig",
     "expectationSuiteSchema",
     "expectationSuiteValidationResultSchema",
     "expectationValidationResultSchema",
--- a/great_expectations/core/factory/validation_factory.py
+++ b/great_expectations/core/factory/validation_factory.py
@@ -1,13 +1,9 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
-
 from great_expectations._docs_decorators import public_api
 from great_expectations.compatibility.typing_extensions import override
 from great_expectations.core.factory.factory import Factory
-
-if TYPE_CHECKING:
-    from great_expectations.core.validation import Validation
+from great_expectations.core.validation import Validation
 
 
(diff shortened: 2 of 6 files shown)


great_expectations/core/factory/validation_factory.py
```diff
@@ -0,0 +1,56 @@
...
+from great_expectations.core.factory.factory import Factory
+
+if TYPE_CHECKING:
+    from great_expectations.core.validation import Validation
+    from great_expectations.data_context.store.validations_store import ValidationsStore
+
+
+# TODO: Add analytics as needed
+class ValidationFactory(Factory[Validation]):
+    def __init__(self, store: ValidationsStore) -> None:
```
cdkini: Current store won't work so we'll need to update accordingly

great_expectations/core/validation.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class Validation:
+    """
+    TBD
+    """
+
+    def __init__(
+        self,
+        name: str,
```
cdkini: Not part of the spec but we need an identifier if we want persistence

great_expectations/core/factory/validation_factory.py
```diff
@@ -0,0 +1,56 @@
...
+
+if TYPE_CHECKING:
+    from great_expectations.core.validation import Validation
+
+
+# TODO: Add analytics as needed
+class ValidationFactory(Factory[Validation]):
+    def __init__(self, store) -> None:
+        # TODO: Update type hints when new ValidationStore is implemented
+        self._store = store
```
cdkini: Think I'm blocked on this. We have an existing `ValidationsStore` but those are for results?

I think we need to do the following:
- Rename the old store to `ValidationResultsStore`
- Create a new `ValidationsStore` (or `ValidationStore` our inconsistent plurality should be resolved)
- Plug in the new one here and have CRUD start working

great_expectations/core/validation.py
```diff
@@ -0,0 +1,45 @@
...
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from great_expectations.core.batch_config import BatchConfig
+    from great_expectations.core.expectation_suite import ExpectationSuite
+    from great_expectations.datasource.fluent.interfaces import DataAsset
+
+
+class Validation:
```
cdkini: Do we want this to be a Pydantic model? Went back and forth on it but figured its a pretty lightweight object so we can just add it if we need (probably when persistence comes into question?)

great_expectations/core/validation.py
```diff
@@ -0,0 +1,31 @@
...
+    Args:
+        name: The name of the validation.
+        data: An asset or batch config to validate.
+        suite: A grouping of expectations to validate against the data.
+
+    """
+
+    name: str
+    data: DataAsset | BatchConfig
+    suite: ExpectationSuite
```
cdkini: If I make a change to a suite, how can I ensure that changes cascade through our persistence layer?

great_expectations/core/factory/validation_factory.py
cdkini: Working on the store now: https://github.com/great-expectations/great_expectations/pull/9515

great_expectations/core/validation.py
```diff
@@ -0,0 +1,31 @@
...
+
+    Args:
+        name: The name of the validation.
+        data: An asset or batch config to validate.
+        suite: A grouping of expectations to validate against the data.
+
+    """
+
+    name: str
+    data: DataAsset | BatchConfig
```
tyler-hoffman: I know we've talked about this some, but did we land on this being a union with DataAsset? I know it can be done with discriminated unions, but do we currently have a good story around serialization there? I'm a bit concerned about the ergonomics of accessing `validation.data` and having to do isinstance checks. Did we rule out a BatchConfig that represents the whole DataAsset?

billdirks: We could make this a BatchConfig and use pydantic coercion to allow a user to instantiate one with an asset.

great_expectations/core/validation.py
billdirks: Do you mean a user has access to both a validation and a suite. They update the suite, how do we guarantee they see that in the validation?

Would be make use of a property and have it do an external call each time someone gets this property?

cdkini: I'm concerned that a deletion or material change to a suite would result in downstream errors if the validations store didn't stay in sync. Would we need to check every time a suite or batch config was updated/deleted?

great_expectations/core/validation.py
billdirks: We've decided to define a protocol and allow this to be a union.

great_expectations/core/validation_config.py
```diff
@@ -0,0 +1,32 @@
...
+
+    Args:
+        name: The name of the validation.
+        data: A batch config to validate.
+        suite: A grouping of expectations to validate against the data.
+
+    """
+
+    name: str
+    data: BatchConfig  # TODO: Should support a union of Asset | BatchConfig
```
cdkini, line 27: Only supporting batch config for the initial implementation - will follow up with asset support but keeping scope limited to get something working

great_expectations/core/__init__.py
```diff
@@ -18,23 +18,25 @@
 from .id_dict import IDDict
 from .run_identifier import RunIdentifier, RunIdentifierSchema
 from .urn import ge_urn
+from .validation_config import ValidationConfig
```
cdkini, line 21: Name is temporary and will be updated based on team consensus in the near future

tyler-hoffman (approved):

PR #9517, opened 2024-02-23
[MAINTENANCE] Remove fluent partitioner methods from DataAssets

joshua-stauffer:


great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -1056,23 +890,6 @@ def test_connection(self) -> None:
                 f"failed. Ensure the table exists and the user has access to select data from the table: {query_error}"
             ) from query_error
 
-    @override
-    def test_partitioner_connection(self) -> None:
```
tyler-hoffman, line 1060: Do we want to move this to the batch config (but probably not call it automatically)?

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -304,11 +284,13 @@ def add_batch_config(
         batch_config = BatchConfig(name=name, partitioner=partitioner)
         batch_config.set_data_asset(self)
         self.batch_configs.append(batch_config)
+        self.update_batch_config_field_set()
```
cdkini, line 287: We now need to always remember to call this after each update to `self.batch_configs`? Is there a better way to do this without putting the burden on future devs?

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -114,13 +114,8 @@ def _get_reader_method(self) -> str:
...
         ...
 
-    @property
-    @override
-    def batch_request_options(self) -> tuple[str, ...]:
-        return tuple()
-
     @override
-    def get_batch_request_options_keys(self, partitioner):
+    def get_batch_request_options_keys(self, partitioner: Optional[Partitioner] = None):
```
cdkini: Can we annotate the return?

great_expectations/datasource/fluent/file_path_data_asset.py
```diff
@@ -176,36 +174,10 @@ def get_partitioner_implementation(
             )
         return PartitionerClass(**abstract_partitioner.dict())
 
-    @property
-    @override
-    def batch_request_options(
```
tyler-hoffman, line 181: Do we have a ticket to add this property to BatchConfigs and have them call this method on their asset?

tyler-hoffman (approved):
Had some questions about future improvements, but this looks solid to me.

great_expectations/datasource/fluent/interfaces.py
joshua-stauffer: cc @tyler-hoffman  - this was a bugfix he snuck in, since the context was in this PR

great_expectations/datasource/fluent/pandas_datasource.py
joshua-stauffer: I was under the impression that we got this typing from the parent class, so adding types here is redundant. But if i'm wrong about that, I'm happy to update!

joshua-stauffer: i gave this some more thought, and i think you're right, even though the type is provided by the parent class, it's easiest to read if its also here. will update, thanks! 👍

great_expectations/datasource/fluent/sql_datasource.py
joshua-stauffer: this is only implemented for SQL partitioners, so to me it felt like a half baked feature. It would make more sense to me if we just had a `BatchConfig.test_connection()` method which checks the configuration, including the partitioner.

tyler-hoffman: Yeah, makes sense to me.

great_expectations/datasource/fluent/file_path_data_asset.py
joshua-stauffer: [v1-205](https://greatexpectations.atlassian.net/browse/V1-205) 👍

great_expectations/datasource/fluent/interfaces.py
joshua-stauffer: i've added [a ticket](https://greatexpectations.atlassian.net/browse/V1-206), and will have a solution as a followup 🙇

PR #9523, opened 2024-02-23
[FEATURE] `ValidationConfigStore`

cdkini:


great_expectations/core/validation_config.py
```diff
@@ -23,10 +20,25 @@ class ValidationConfig(BaseModel):
...
+        if isinstance(v, dict):
+            return ExpectationSuite(**expectationSuiteSchema.load(v))
+        return v
+
     @public_api
     def run(self):
         raise NotImplementedError
+
+
+ValidationConfig.update_forward_refs()
```
cdkini: Not sure why this is necessary 🤔

great_expectations/data_context/data_context/serializable_data_context.py
```diff
@@ -48,6 +48,7 @@ class SerializableDataContext(AbstractDataContext):
         DataContextConfigDefaults.EXPECTATIONS_BASE_DIRECTORY.value,
         DataContextConfigDefaults.PLUGINS_BASE_DIRECTORY.value,
         DataContextConfigDefaults.PROFILERS_BASE_DIRECTORY.value,
+        DataContextConfigDefaults.VALIDATION_CONFIGS_BASE_DIRECTORY.value,
```
cdkini, line 51: Updates to enable proper filesystem scaffolding

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -133,6 +133,7 @@ class GXCloudStoreBackend(StoreBackend, metaclass=ABCMeta):
         GXCloudRESTResource.PROFILER: "profiler",
         GXCloudRESTResource.RENDERED_DATA_DOC: "rendered_data_doc",
         GXCloudRESTResource.VALIDATION_RESULT: "result",
+        GXCloudRESTResource.VALIDATION_CONFIG: "validation_config",
```
cdkini, line 136: Updates to enable proper Cloud integration (purely mocking for now)

great_expectations/data_context/store/validation_config_store.py
```diff
@@ -0,0 +1,46 @@
...
+        store_backend: dict | None = None,
+        runtime_environment: dict | None = None,
+    ) -> None:
+        super().__init__(
+            store_name=store_name,
+            store_backend=store_backend,
+            runtime_environment=runtime_environment,
+        )
+
+    def get_key(
```
cdkini, line 16: Implementing the bare minimum required in the factory. Tests should reflect this as well

great_expectations/core/validation_config.py
```diff
@@ -23,10 +20,25 @@ class ValidationConfig(BaseModel):
 
     """
 
+    class Config:
+        arbitrary_types_allowed = True
+        json_encoders = {
+            ExpectationSuite: lambda v: expectationSuiteSchema.dump(v),
+        }
```
cdkini, line 29: Concurrently using Marshmallow and Pydantic results in something like this. We need custom encoders/decoders but this should work as intended

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -617,6 +625,13 @@ def validations_store_name(self, value: str) -> None:
     def validations_store(self) -> ValidationsStore:
         return self.stores[self.validations_store_name]
 
+    @property
+    def validation_config_store_name(self) -> str:
+        return DataContextConfigDefaults.DEFAULT_VALIDATION_CONFIG_STORE_NAME.value
+
+    def validaiton_config_store(self) -> ValidationConfigStore:
+        return self.stores[self.validation_config_store_name]
```
cdkini: Wasn't sure about this but deciding to not make `validation_config_store_name` a context variable. I've wavered on it but I'm now confident that store naming just causes more harm than benefit.

An argument against this is that it is a deviation from the existing pattern and perhaps consistency is more important?

great_expectations/data_context/store/validation_config_store.py
```diff
@@ -0,0 +1,51 @@
...
+                id=id,
+                resource_name=name,
+            )
+        return StringKey(key=name)
+
+    @override
+    def serialize(self, value):
+        if self.cloud_mode:
+            data = value.dict()
+            data["suite"] = data["suite"].to_json_dict()
```
cdkini, line 32: Another necessity due to Marshmallow

great_expectations/core/validation_config.py
tyler-hoffman: hmm.. just to double check - any chance this was needed before they type checking block was removed, but is no longer required?

great_expectations/data_context/data_context/abstract_data_context.py
tyler-hoffman: I'm usually not pro-deviation, but I think in this case it's probably reasonable. It'd require minimal changes if we changed our mind later.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -617,6 +625,13 @@ def validations_store_name(self, value: str) -> None:
     def validations_store(self) -> ValidationsStore:
         return self.stores[self.validations_store_name]
 
+    @property
+    def validation_config_store_name(self) -> str:
+        return DataContextConfigDefaults.DEFAULT_VALIDATION_CONFIG_STORE_NAME.value
+
+    def validaiton_config_store(self) -> ValidationConfigStore:
```
billdirks:
```suggestion
    def validation_config_store(self) -> ValidationConfigStore:
```

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -331,8 +334,13 @@ def _init_factories(self) -> None:
                 context=self,
             )
 
-        # TODO: Update to follow existing pattern once new ValidationConfigStore is implemented
         self._validations: ValidationFactory | None = None
+        if validation_config_store := self.stores.get(
```
billdirks: Is there no default store, eg an ephemeral store?

great_expectations/data_context/data_context/abstract_data_context.py
billdirks: Do we know this key will always be present?
What is the harm you see from making this a context variable? By context variable, do you mean a top level property?

great_expectations/data_context/store/validation_config_store.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class ValidationConfigStore(Store):
+    _key_class = StringKey
+
+    def get_key(
+        self, name: str, id: str | None = None
+    ) -> GXCloudIdentifier | StringKey:
+        """Given a name and optional ID, build the correct key for use in the ValidationConfigStore."""
+        if self.cloud_mode:
```
billdirks, line 20: Not for this PR but we need to get away from these cloud mode checks.

billdirks (commented):
Thanks for doing this. I don't have the best context with respect to stores. I've asked some questions inline.

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: By adding it as a context variable, its another field that a user can modify.

These top level store names are dangerous in that they can enable the following:
```python
context.checkpoint_store_name = "my_nonexistent_store"
...

context.checkpoint_store # Raises an error
context.checkpoints.add(...) # Raises an error
...
```

cdkini: We allow customization of the names but don't automatically update the `stores` dict, resulting in the possibility of orphaning a store by losing its name.

Seems relatively unlikely since we don't advertise this much in docs/public API.

We should always have a `ValidationConfigStore` but the name may vary.

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: Currently no - we allow optional configuration but won't raise an error until it is used in CRUD.

We should change this but prioritizing feature development first. I am ensuring that all new contexts create this store though

great_expectations/data_context/store/validation_config_store.py
cdkini: I agree. It's poorly encapsulated but I couldn't think of a better way to get both file and cloud working

great_expectations/core/validation_config.py
cdkini: I believe you're right!

great_expectations/data_context/data_context/abstract_data_context.py
cdkini:
```suggestion
        validation_config_store = self.validation_config_store # or something like that
```

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -617,6 +625,13 @@ def validations_store_name(self, value: str) -> None:
     def validations_store(self) -> ValidationsStore:
         return self.stores[self.validations_store_name]
 
+    @property
+    def validation_config_store_name(self) -> str:
+        return DataContextConfigDefaults.DEFAULT_VALIDATION_CONFIG_STORE_NAME.value
+
+    def validation_config_store(self) -> ValidationConfigStore:
```
cdkini, line 625: Should be a property

cdkini: Combine name prop here - limit surface area

great_expectations/core/expectation_suite.py
```diff
@@ -1207,4 +1207,4 @@ def _convert_uuids_to_str(self, data, **kwargs):
         return data
 
 
-expectationSuiteSchema = ExpectationSuiteSchema()
+expectationSuiteSchema: ExpectationSuiteSchema = ExpectationSuiteSchema()
```
billdirks, line 1210: nit: This type can be inferred and is the same as the constructor so you don't need this annotation.

great_expectations/core/validation_config.py
```diff
@@ -23,9 +22,22 @@ class ValidationConfig(BaseModel):
...
     name: str
     data: BatchConfig  # TODO: Should support a union of Asset | BatchConfig
     suite: ExpectationSuite
+    id: Union[str, None] = None
+
+    @validator("suite", pre=True)
+    def _validate_suite(cls, v):
+        if isinstance(v, dict):
+            return ExpectationSuite(**expectationSuiteSchema.load(v))
+        return v
```
billdirks: Should we check if v is an expectationSuiteSchema and raise if it's not in the fall through case? Or does pydantic do this for us since it won't be the right type?

great_expectations/data_context/store/gx_cloud_store_backend.py
billdirks: It seems like the enum can be modified/restructured if we have to list all these enums and define a new value for them. This makes it appear that our enum/abstraction is wrong.

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -157,6 +158,7 @@ class GXCloudStoreBackend(StoreBackend, metaclass=ABCMeta):
             GXCloudRESTResource.PROFILER: "profilers",
             GXCloudRESTResource.RENDERED_DATA_DOC: "rendered_data_docs",
             GXCloudRESTResource.VALIDATION_RESULT: "validation_results",
+            GXCloudRESTResource.VALIDATION_CONFIG: "validation_configs",
```
billdirks, line 161: Similarly here.

billdirks (commented):
Mostly looks good. I've asked 1 question and added a couple of nits that don't need to be solved now. There is also the outstanding property comment you've made.

great_expectations/core/expectation_suite.py
cdkini: weirdly enough I got a mypy error without this (when using it in the custom Pydantic encoder below)

great_expectations/core/validation_config.py
cdkini: Pydantic validation will happen after this return - I've updated this to be a bit more explicit but it should work as-is

great_expectations/core/validation_config.py
```diff
@@ -23,9 +22,26 @@ class ValidationConfig(BaseModel):
 
     """
 
+    class Config:
+        arbitrary_types_allowed = True
+        json_encoders = {
+            ExpectationSuite: lambda v: expectationSuiteSchema.dump(v),
```
tyler-hoffman, line 28: nit: I think we should add a comment on the why here in the code.

great_expectations/core/validation_config.py
```diff
@@ -23,9 +22,26 @@ class ValidationConfig(BaseModel):
...
+        }
+
     name: str
     data: BatchConfig  # TODO: Should support a union of Asset | BatchConfig
     suite: ExpectationSuite
+    id: Union[str, None] = None
+
+    @validator("suite", pre=True)
+    def _validate_suite(cls, v):
```
tyler-hoffman, line 37: same as above - I think it's worth a docstring explaining this is just because the suite isn't pydantic.

tests/data_context/store/test_validation_config_store.py
```diff
@@ -0,0 +1,134 @@
...
+@pytest.fixture
+def validation_config() -> ValidationConfig:
+    return ValidationConfig(
+        name="my_validation",
+        data=BatchConfig(name="my_batch_config"),
+        suite=ExpectationSuite(name="my_suite"),
+    )
+
+
+@pytest.mark.parametrize("store_fixture", ["ephemeral_store", "file_backed_store"])
```
tyler-hoffman, line 61: why use strings for the stores and then call getfixturevalue? Can't you just reference the fixture directly?

tyler-hoffman (approved):
Left some nits, but this LGTM

billdirks (approved):
LGTM!

great_expectations/data_context/types/base.py
```diff
@@ -2320,6 +2389,24 @@ def __init__(  # noqa: PLR0912, PLR0913
...
+        # If missing individual stores, just add what is missing
+        configured_stores = {config["class_name"] for config in store_configs.values()}
+        for name, config in DataContextConfigDefaults.DEFAULT_STORES.value.items():
+            if not isinstance(config, dict):
+                raise ValueError(
+                    "Store defaults must be a mapping of default names to default dictionary configurations."
+                )
+            if config["class_name"] not in configured_stores:
+                # Create ephemeral store config
+                store_configs[name] = {"class_name": config["class_name"]}
```
billdirks, line 2406: Please write a warning if we use the default store.

tests/data_context/store/test_validation_config_store.py
Kilo59: No, the fixture won't be properly instantiated unless you do it this way.
Fixtures only get loaded by pytest if you include them in the test constructor.
Although this is a common complaint because it prevents you from using fixtures as part of `parametrize`.

There are packages out there like `pytest-lazy-fixture` that essentially do this work for you... but you still have to reference it by the string.
https://github.com/tvorog/pytest-lazy-fixture

tests/data_context/store/test_validation_config_store.py
```diff
@@ -0,0 +1,134 @@
...
+@pytest.mark.unit
+def test_add(request, store_fixture: str, validation_config: ValidationConfig):
```
Kilo59, line 63:
```suggestion
def test_add(request: pytest.FixtureRequest, store_fixture: str, validation_config: ValidationConfig):
```
Annoting `request` will developers understand what's going on.

https://docs.pytest.org/en/stable/reference/reference.html#request

PR #9525, opened 2024-02-23
[DOCS] Feedback Modal

deborahniesz:
# Description

Pop-up when you click "No" in the "Was this Helpful?" section. You can dismiss the form clicking outside the form area or closing it. Whether you click "yes", "no", dismiss the form or answer it, all this information will be stored in the [Posthog Dashboard](https://us.posthog.com/project/4219/dashboard/134373)

[Deploy Preview](https://deploy-preview-9525.docs.greatexpectations.io)
[mock-up](https://xd.adobe.com/view/9ef7232c-80ce-4ad7-9c3f-1a96126f84b4-cba4/screen/513bb1be-f143-4f02-94c7-834308ff7bd6/variables/) (attaching files functionality is temporary removed) 

https://github.com/great-expectations/great_expectations/assets/38264052/3b12ff1e-8bb7-4521-bda0-b6f20f5938bf


docs/docusaurus/src/components/WasThisHelpful/index.js
```diff
@@ -14,14 +30,99 @@ export default function WasThisHelpful(){
         }
     };
 
+    const handleNegativeFeedbackReaction = () => {
+        setIsOpen(true)
```
josectobar, line 34: This makes it so the button always trigger the modal even after it turns gray. I don't think we want this. Once the buttons turn gray they should be disabled.

docs/docusaurus/src/components/WasThisHelpful/index.js
```diff
@@ -1,11 +1,27 @@
 import React, {useState} from 'react';
```
josectobar, line 1: We're going to need to add a test for this puppy.

docs/docusaurus/src/components/WasThisHelpful/index.js
```diff
@@ -14,14 +30,99 @@ export default function WasThisHelpful(){
...
+                            <textarea
+                                name="description"
+                                value={formData.description}
+                                className={styles.modalTextInput + ' ' + styles.modalTextareaInput}
+                                onChange={handleChange}
+                                required
+                                placeholder="Provide as much detail as possible about the issue you
+                                experienced or where improvement is needed. Detailed feedback helps
+                                us better identify the problem and determine a solution."
+                            />
```
josectobar, line 121: We need set a min and max width so we can't resize it horizontally, only vertically.

docs/docusaurus/src/components/WasThisHelpful/index.js
```diff
@@ -14,14 +30,99 @@ export default function WasThisHelpful(){
...
                 <div className={styles.feedbackCardActions}>
                     <button className={feedbackSent ? styles.inactiveFeedbackButton : styles.feedbackButton} onClick={() => handleFeedbackReaction('docs_feedback.yes')}>Yes</button>
-                    <button className={feedbackSent ? styles.inactiveFeedbackButton : styles.feedbackButton} onClick={() => handleFeedbackReaction('docs_feedback.no')}>No</button>
+                    <button className={feedbackSent ? styles.inactiveFeedbackButton : styles.feedbackButton} onClick={handleNegativeFeedbackReaction}>No</button>
                 </div>
             </section>
-        </>
+
+            {isOpen && <>
+                <div className={styles.overlay} onClick={dismissFeedbackModal}/>
```
josectobar, line 72: I see we've only used divs, Lets replace some with semantic tags when applicable for this modal.

docs/docusaurus/src/components/WasThisHelpful/styles.module.scss
```diff
@@ -0,0 +1,203 @@
...
+
+.inactiveSubmitButton {
+    border-radius: 20px;
+    background-color: var(--gray-400);
+    font-weight: bold;
+    color: white;
+    border: none;
+    padding: 0.75rem;
+    margin: 1rem 0 2rem;
+}
```
josectobar: can we merge these three in one block?

josectobar (changes requested):
@deborahniesz This looks great. I do think we need a test for this form. I also left a few minor suggestions. Thanks!

docs/docusaurus/___test___/wasThisHelpful.test.js
```diff
@@ -0,0 +1,82 @@
...
+        await selectingNoInWasThisHelpful();
+
+        expect(screen.getByText('Yes')).toBeDisabled();
+        expect(screen.getByText('No')).toBeDisabled();
+    });
+
+    test("Feedback Modal should pop-up when 'No' button has been clicked", async () => {
+        await selectingNoInWasThisHelpful();
+
+        expect(screen.getByTestId('feedback-modal')).toBeInTheDocument();
```
josectobar: Can we assert that the modal is present by finding the title of the modal instead?

docs/docusaurus/___test___/wasThisHelpful.test.js
```diff
@@ -0,0 +1,82 @@
...
+        await selectingNoInWasThisHelpful();
+
+        const descriptionBox = screen.getByTestId('description-textbox');
+
+        await userEvent.type(descriptionBox,'I have a problem');
+
+        expect(screen.getByText("Submit")).not.toBeDisabled();
+    });
+
+    test("", async () => {
```
josectobar: We're missing a title for this test

docs/docusaurus/___test___/wasThisHelpful.test.js
```diff
@@ -0,0 +1,82 @@
...
+        await selectingNoInWasThisHelpful();
+
+        const descriptionBox = screen.getByTestId('description-textbox');
+
+        await userEvent.type(descriptionBox,'I have a problem');
+
+        expect(screen.getByText("Submit")).not.toBeDisabled();
```
josectobar: Can we instead assert that the modal disappear after you hit submit?

docs/docusaurus/___test___/wasThisHelpful.test.js
```diff
@@ -0,0 +1,82 @@
...
+    test("Submit button in Feedback Modal is disabled when description is blank", async () => {
+        await selectingNoInWasThisHelpful();
+
+        expect(screen.getByText("Submit")).toBeDisabled();
+    });
+
+    test("Submit button in Feedback Modal is enabled when description input is filled", async () => {
+        await selectingNoInWasThisHelpful();
+
+        const descriptionBox = screen.getByTestId('description-textbox');
```
josectobar: Can we find the text box by the name instead? I prefer if we stay away from test ids and instead we use attributes that are visible to the user to make the test resemble a use case.

docs/docusaurus/package.json
```diff
@@ -28,6 +28,8 @@
     "@docusaurus/theme-mermaid": "2.4.1",
     "@docusaurus/theme-search-algolia": "2.4.1",
     "@mdx-js/react": "1.6.22",
+    "@testing-library/jest-dom": "^6.4.2",
+    "@testing-library/user-event": "^14.5.2",
```
josectobar: I don't think we need to install these independently but either way we should also install them as a dev dependency. https://testing-library.com/docs/react-testing-library/intro/

docs/docusaurus/src/components/WasThisHelpful/index.js
```diff
@@ -14,14 +30,100 @@ export default function WasThisHelpful(){
...
-                    <button className={feedbackSent ? styles.inactiveFeedbackButton : styles.feedbackButton} onClick={() => handleFeedbackReaction('docs_feedback.no')}>No</button>
+                    <button disabled={feedbackSent} className={feedbackSent ? styles.inactiveFeedbackButton : styles.feedbackButton} onClick={() => handleFeedbackReaction('docs_feedback.yes')}>Yes</button>
+                    <button disabled={feedbackSent} className={feedbackSent ? styles.inactiveFeedbackButton : styles.feedbackButton} onClick={handleNegativeFeedbackReaction}>No</button>
                 </div>
             </section>
-        </>
+
+            {isOpen && <>
+                <div className={styles.overlay} onClick={dismissFeedbackModal}/>
+                <section className={styles.modal} data-testid="feedback-modal">
```
josectobar: I think `dialog` is a better tag for the modal container.

josectobar (changes requested):
Thanks for working on the test, it looks great. I have just a few more minor changes.

docs/docusaurus/___test___/wasThisHelpful.test.js
deborahniesz: Yes, I agree. I tried to find the text box by name but it said that name value was empty. So I couldn't access the textbox by using screen.getByRole('textbox', { name: 'description'}), do you have any idea how I could access it?

docs/docusaurus/package.json
josectobar: actually, I think this is correct, you just need to move it to the dev deps list

docs/docusaurus/___test___/wasThisHelpful.test.js
josectobar: try ` getByPlaceholderText`

josectobar: getByPlaceholderText looks for an aria-label not the `name` attr. https://testing-library.com/docs/queries/about/#priority

deborahniesz: I was not sure about placeholder text because it was a long text, but it's true that compared to having getByTestId that needs test-related code in the react component, the placeholder is a much better solution :place_of_worship:

josectobar (approved):
Thanks for working on these changes, this modal looks great!

PR #9536, opened 2024-02-26
[MAINTENANCE] Replace `black` formatter with `ruff format`

Kilo59:
## Replacing the `black` formatter with `ruff format`
https://docs.astral.sh/ruff/formatter/

>The initial goal of the Ruff formatter is not to innovate on code style, but rather, to innovate on performance, and provide a unified toolchain across Ruff's linter, formatter, and any and all future tools.

>As such, the formatter is designed as a drop-in replacement for [Black](https://github.com/psf/black), but with an excessive focus on performance and direct integration with Ruff.

### Note:
Some of the changes in this PR are because `ruff format` is copying a more recent version of [black `24.2.0`](https://github.com/psf/black/releases/tag/24.2.0), whereas we had been stuck on [`23.11.0`](https://github.com/psf/black/releases/tag/23.11.0)

## Update Ruff `0.2.2` -> `0.3.0`
https://github.com/astral-sh/ruff/releases/tag/v0.3.0

## TODO
If we decide to adopt this formatter I will...
* Before merge
  - [x] address the lining errors caused by formatting changes that move `noqa` comments.
* After merge  
  - [x] add `.gitblame-ignore-revs` entry 

## See also
- https://github.com/great-expectations/cloud/pull/46


pyproject.toml
```diff
@@ -4,21 +4,6 @@ requires = ["setuptools", "wheel"]
...
 # build-backend = "setuptools.build_meta"
 
-[tool.black]
-target-version = ["py38", "py39", "py310"]
-extend_excludes = '''
-(
-     docs/.*
-    | tests/.*.fixture
-    | .*.ge_store_backend_id
-)
```
Kilo59, line 14: These were previously excluded from formatting.
Do we still need to exclude them?

tests/datasource/fluent/_fake_cloud_api.py
```diff
@@ -860,16 +860,20 @@ def gx_cloud_api_fake_ctx(
...
+                post_expectation_suites_cb,
+            ),
+        )
+        (
+            resp_mocker.add_callback(
+                responses.PUT,
+                f"{org_url_base}/expectation-suites/{FAKE_EXPECTATION_SUITE_ID}",
+                put_expectation_suites_cb,
+            ),
+        )
```
Kilo59, line 876:
```suggestion
            resp_mocker.add_callback(
                responses.POST,
                f"{org_url_base}/expectation-suites",
                post_expectation_suites_cb,
            )
            resp_mocker.add_callback(
                responses.PUT,
                f"{org_url_base}/expectation-suites/{FAKE_EXPECTATION_SUITE_ID}",
                put_expectation_suites_cb,
            )
```

tests/data_context/test_get_data_context.py
```diff
@@ -184,7 +184,7 @@ def test_cloud_missing_env_throws_exception(
     clear_env_vars, empty_ge_cloud_data_context_config
 ):
     with pytest.raises(GXCloudConfigurationError):
-        gx.get_context(cloud_mode=True),
+        (gx.get_context(cloud_mode=True),)
```
Kilo59, line 187:
```suggestion
        gx.get_context(cloud_mode=True)
```

tests/rule_based_profiler/conftest.py
Kilo59, line 1: Ruff should be respecting these fmt off comments

## See
https://github.com/great-expectations/great_expectations/pull/9536#discussion_r1511264598

pyproject.toml
billdirks: I think we do since these all look like specially formatted files. For example, changing `docs/.*` may change how things get rendered in our docs.

pyproject.toml
```diff
@@ -264,10 +249,16 @@ lint.select = [
     "NPY201", # numpy 2.0 compatibility (see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#ruff-plugin) -- Explicitly listed to view preview rule errors
 ]
 lint.ignore = [
+    # formatting related ignores
+    # https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules
+    "W191", # tab-identation
```
billdirks, line 249: I think we do want to enforce no tabs in python files.

pyproject.toml
```diff
@@ -264,10 +249,16 @@ lint.select = [
...
+    "E111", # indentation-with-invalid-multiple
+    "E114", # indentation-with-invalid-multiple-comment
+    "E117", # over-idented
```
billdirks, line 252: I think we should turn this one on too.

pyproject.toml
```diff
@@ -264,10 +249,16 @@ lint.select = [
...
+    "E501", # line-too-long
```
billdirks, line 253: I see we have this rule turned off in black. It seems like the lines are much longer here though. What happens if we turn this on? Maybe there will be too many long strings to easily ignore? If rough can add ignores for us, I'd like to turn this on.

billdirks (commented):
In general, I'm happy with this change. I think we need to do a few things before merging:
* Add file ignores
* Add #noqa back
* Update PR autopopulated message to remove mention of `black`
* I also added comments to the toml file for rules I'd like to be turned on if possible

pyproject.toml
Kilo59: @billdirks 
Ruff recommends disabling these rules when using the formatter.

https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules

>Ruff's formatter is designed to be used alongside the linter. However, the linter includes some rules that, when enabled, can cause conflicts with the formatter, leading to unexpected behavior. When configured appropriately, the goal of Ruff's formatter-linter compatibility is such that running the formatter should never introduce new lint errors.

pyproject.toml
Kilo59:
```console
ruff --select E501 --statistics
13482   E501    Line too long (100 > 88)
```

pyproject.toml
Kilo59: Actually it looks like most of these where already being ignored by `ruff lint` so they are already being ignored by `ruff format`.
There doesn't appear to be a way to have a separate `format` ignore list.

pyproject.toml
billdirks: Oh, I misread this as ruff formatter configuration but these are turning off linter rules. I'm good with this.

tests/rule_based_profiler/conftest.py
Kilo59, line 1: `# fmt: off` outside of the top-level/statement level is a known issue with the `ruff` formatter
- https://github.com/astral-sh/ruff/issues/9588
- https://github.com/astral-sh/ruff/issues/9875

Can fix this by de-denting the `# fmt: off` comment or by adding a trailing `# fmt: skip`, but these will disable formatting for more than just the parameter values.

Edit:
Addressed with `# fmt: skip`

contrib/experimental/great_expectations_experimental/expectations/expect_column_values_to_match_thai.py
```diff
@@ -21,7 +20,7 @@ class ExpectColumnValuesToMatchThai(RegexBasedColumnMapExpectation):
 
     # These values will be used to configure the metric created by your expectation
     regex_camel_name = "RegexName"
-    regex = "[\u0E00-\u0E7F]+"
+    regex = "[\u0e00-\u0e7f]+"
```
Kilo59, line 23: 👀

contrib/great_expectations_semantic_types_expectations/great_expectations_semantic_types_expectations/expectations/expect_column_values_to_be_valid_ipv4.py
```diff
@@ -59,8 +59,8 @@ class ExpectColumnValuesToBeValidIpv4(ColumnMapExpectation):
                     4294967295,
                     # integer packed into a bytes object of length 4
                     b"\x00\x00\x00\x00",
-                    b"\xC0\xA8\x00\x01",
-                    b"\xFF\xFF\xFF\xFF",
+                    b"\xc0\xa8\x00\x01",
+                    b"\xff\xff\xff\xff",
```
Kilo59, line 63: 👀

great_expectations/util.py
```diff
@@ -406,29 +400,6 @@ def gen_directory_tree_str(startpath: PathStr):
     return output_str
 
 
-# NOTE: Can delete once CLI is removed
-def lint_code(code: str) -> str:
-    """Lint strings of code passed in.  Optional dependency "black" must be installed."""
```
Kilo59, line 411: @cdkini @billdirks 

I'm not sure why we ever needed this.
Maybe a long time ago we weren't enforcing the `black` format style in our CI, so we had to rely on this?

Maybe this used to be called directly from the CLI?

billdirks (approved):
Thanks, LGTM!

PR #9552, opened 2024-02-29
[FEATURE] Add evaluation parameter support to v1 validator

tyler-hoffman:
Adds evaluation parameter support to V1Validator. Note: we'll rename evaluation parameters at some point soon.

This PR also moves out a chunk of logic from `Validator::validate` responsible for both processing evaluation parameters and grouping by column name so that we can reuse that in the V1Validator. This was needed because V1Validator calls into `Validator::graph_validate`, rather than `Validator::validate` (the latter doesn't support individual expectations).


great_expectations/validator/v1_validator.py
```diff
@@ -48,7 +48,9 @@ def __init__(
         self.result_format = result_format
 
     def validate_expectation(
-        self, expectation: Expectation
+        self,
+        expectation: Expectation,
+        evaluation_parameters: Optional[dict[str, Any]] = None,
```
cdkini, line 54: Nit - I think just `dict` is sufficient here instead of `[str, Any]`

great_expectations/validator/validator.py
```diff
@@ -1511,6 +1492,37 @@ def validate(  # noqa: C901, PLR0912, PLR0913, PLR0915
...
+    ) -> list[ExpectationConfiguration]:
+        """Substitute evaluation parameters into the provided expectations and sort by column."""
+        columns: dict[Any, list[ExpectationConfiguration]] = {}
+
+        for expectation in expectation_configurations:
+            expectation.process_evaluation_parameters(
+                evaluation_parameters=evaluation_parameters,
+                interactive_evaluation=self.interactive_evaluation,
+                data_context=self._data_context,
+            )
```
cdkini, line 1510: Could we do all this work on the actual expectation instead of the config? Perhaps we need to shift the logic from the config to the actual expectation?

cdkini: Also what is the context dependency?

great_expectations/validator/validator.py
```diff
@@ -1511,6 +1492,37 @@ def validate(  # noqa: C901, PLR0912, PLR0913, PLR0915
...
+            if "column" in expectation.kwargs and isinstance(
+                expectation.kwargs["column"], Hashable
+            ):
+                column = expectation.kwargs["column"]
+            else:
+                # noinspection SpellCheckingInspection
+                column = "_nocolumn"
```
cdkini: Not any fault of this PR but I don't love this magic string. Can we see where else we use it and perhaps make it a constant/enum?

great_expectations/validator/validator.py
tyler-hoffman: Most of the heavy lifting here is still on the expectation, since we are just calling `expectation.process_evaluation_parameters`. The sorting below likely needs to live outside of the expectation (unless it were a class method). But overall my goal here was just move this logic from the above `validate` method so that we can reuse it without doing a more heavy refactor.

Right now context is only used (more deep in the call stack in `parse_evaluation_parameter`) to reach into the metrics store. That can come out, but will require some bigger changes that I think we should tackle in our upcoming work on evaluation params.

I'd like to argue that this change in this PR should stay as is because a) this is in the old Validator that we have a shim over top of, and it's not degrading the quality of any of our abstractions (since it's just cut/paste within the same class). LMK if you disagree though!

great_expectations/validator/validator.py
tyler-hoffman: Yeah, I was confused when I first saw that. I originally kept it so that this was just a strict cut/paste, but the lift is low to update this. Will do.

great_expectations/validator/v1_validator.py
billdirks: I like being explicit about the key type.

great_expectations/validator/validator.py
```diff
@@ -1511,6 +1492,39 @@ def validate(  # noqa: C901, PLR0912, PLR0913, PLR0915
...
+    def process_expectations_for_validation(
+        self,
+        expectation_configurations: list[ExpectationConfiguration],
+        evaluation_parameters: Optional[dict[str, Any]] = None,
+    ) -> list[ExpectationConfiguration]:
+        """Substitute evaluation parameters into the provided expectations and sort by column."""
+        NO_COLUMN = (
+            "_nocolumn"  # just used to group expectations that don't specify a column
+        )
+        columns: dict[Any, list[ExpectationConfiguration]] = {}
```
billdirks: I know this is just moved from it's previous location but can you try changing the key from `Any` to `str`. It looks like this should be column names.

great_expectations/validator/v1_validator.py
tyler-hoffman: I figure at least giving the key a type gives us a bit more to go on here, but I'm fine to remove it. I could have sworn I saw a PR in one of our repos to require adding the type params.

great_expectations/validator/validator.py
```diff
@@ -1511,6 +1492,39 @@ def validate(  # noqa: C901, PLR0912, PLR0913, PLR0915
...
         return result
 
+    def process_expectations_for_validation(
+        self,
+        expectation_configurations: list[ExpectationConfiguration],
+        evaluation_parameters: Optional[dict[str, Any]] = None,
+    ) -> list[ExpectationConfiguration]:
+        """Substitute evaluation parameters into the provided expectations and sort by column."""
+        NO_COLUMN = (
+            "_nocolumn"  # just used to group expectations that don't specify a column
```
billdirks, line 1501: Thanks for making this a constant. It appears this magic string is defined and used in a similar way in multiple places in the codebase. Seems like it should be a defined in 1 place for all consumers of it.
This isn't necessary for this PR, just noting.

billdirks (commented):
I'll let you and @cdkini resolve your current thread. LGTM.

great_expectations/validator/validator.py
tyler-hoffman: Yup, I saw that too. Going to make a ticket about it, though I think it's pretty low priority.

cdkini (approved):

PR #9558, opened 2024-03-01
[MAINTENANCE] Ensure proper `ValidationConfig` serialization

cdkini:


great_expectations/core/validation_config.py
```diff
@@ -4,11 +4,39 @@
...
+        },
+        "asset": {
+            "name": parent.name,
+            "id": parent.id,
+        },
+        "batch_config": {
+            "name": data.name,
+            "id": data.id,
+        },
+    }
```
cdkini: Should probably move these to their respective class files

tests/core/test_validation_config.py
```diff
@@ -0,0 +1,346 @@
...
+                "batch_config": {
+                    "name": batch_config.name,
+                    "id": batch_config_id,
+                },
+            },
+            "suite": {
+                "name": validation_config_suite.name,
+                "id": suite_id,
+            },
+            "id": validation_id,  # TODO: Test to ensure ValidationConfigStore adds a top-level id as well
```
cdkini, line 123: Subsequent PR

tests/data_context/store/test_validation_config_store.py
```diff
@@ -50,10 +52,18 @@ def cloud_backed_store(cloud_details: CloudDetails):
...
+    context = in_memory_runtime_context
+    batch_config = (
+        context.sources.add_pandas("my_datasource")
+        .add_csv_asset("my_asset", "data.csv")
+        .add_batch_config("my_batch_config")
+    )
     return ValidationConfig(
         name="my_validation",
-        data=BatchConfig(name="my_batch_config"),
+        data=batch_config,
```
cdkini, line 66: Need the parent/grandparent relationship now that serialization is happening properly. Using idiomatic instantiation pattern here

great_expectations/core/validation_config.py
```diff
@@ -2,13 +2,61 @@
...
+
+class _IdentifierBundle(BaseModel):
+    name: str
+    id: Union[str, None]
+
+
+class _EncodedValidationData(BaseModel):
+    datasource: _IdentifierBundle
+    asset: _IdentifierBundle
+    batch_config: _IdentifierBundle
```
cdkini, line 31: Rich types to ensure strict serialization format (see example below)

great_expectations/core/validation_config.py
```diff
@@ -19,30 +67,139 @@ class ValidationConfig(BaseModel):
...
-            ExpectationSuite: lambda v: expectationSuiteSchema.dump(v),
+            ExpectationSuite: lambda e: _encode_suite(e),
+            BatchConfig: lambda b: _encode_data(b),
         }
 
     name: str
     data: BatchConfig  # TODO: Should support a union of Asset | BatchConfig
     suite: ExpectationSuite
-    id: Union[str, None] = None
+    id: Union[str, None] = None  # TODO: Should be updated when persisted with a store.
```
cdkini, line 114: This will be a follow-up PR

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2497,9 +2497,7 @@ def get_expectation_suite(
             )
 
         if self.expectations_store.has_key(key):
-            expectations_schema_dict: dict = cast(
-                dict, self.expectations_store.get(key)
-            )
+            expectations_schema_dict: dict = self.expectations_store.get(key)
```
cdkini, line 2500: This really tripped me up but we previously claimed that this returned a suite when it actually returns a dict

great_expectations/data_context/data_context/context_factory.py
```diff
@@ -116,6 +117,14 @@ def get_evaluation_parameters_store(self) -> EvaluationParameterStore:
             )
         return self._project.evaluation_parameter_store
 
+    def get_datasources(self) -> DatasourceDict:
+        if not self._project:
+            raise RuntimeError(
+                "This action requires an active DataContext. "
+                + "Please call `get_context()` first, then try your action again."
+            )
+        return self._project.datasources
```
cdkini: We should combine this `datasources` attr with `sources` in some capacity. I need this dependency to grab the grandparent datasource when working with the validation config

great_expectations/data_context/store/tuple_store_backend.py
```diff
@@ -244,7 +244,7 @@ def __init__(  # noqa: PLR0913
         base_directory,
         filepath_template=None,
         filepath_prefix=None,
-        filepath_suffix=None,
+        filepath_suffix=".json",
```
cdkini: Manual testing with a file-backed context showed that we didn't create a JSON file by default. Feels like all file-backed stores should do this so let's update the default

great_expectations/data_context/store/validation_config_store.py
```diff
@@ -32,7 +32,9 @@ def serialize(self, value):
             data["suite"] = data["suite"].to_json_dict()
             return data
 
-        return value.json()
+        # In order to enable the custom json_encoders in ValidationConfig, we need to set `models_as_dict` off
+        # Ref: https://docs.pydantic.dev/1.10/usage/exporting_models/#serialising-self-reference-or-other-models
+        return value.json(models_as_dict=False, indent=2, sort_keys=True)
```
cdkini, line 37: `models_as_dict` is explained above but the other two keys are to ensure consistent formatting when compared to something like suites

tests/core/test_validation_config.py
```diff
@@ -0,0 +1,358 @@
...
+        "suite_id",
+        ["9b35aa4d-7f01-420d-9d45-b45658e60afd", None],
+        ids=["with_suite_id", "without_suite_id"],
+    )
+    @pytest.mark.parametrize(
+        "validation_id",
+        ["708bd8b9-1ae4-43e6-8dfc-42ec320aa3db", None],
+        ids=["with_validation_id", "without_validation_id"],
+    )
+    def test_validation_config_serialization(
```
cdkini, line 78: For all of these tests, mocking felt like quite a burden (especially with JSON serialization issues). I opted for using real objects and relationships but if there are improvements to be made here to better isolate the unit under test, let's talk 😄

great_expectations/data_context/store/tuple_store_backend.py
```diff
@@ -254,6 +254,9 @@ def __init__(  # noqa: PLR0913
         base_public_path=None,
         store_name=None,
     ) -> None:
+        if not filepath_suffix and not filepath_template:
+            filepath_suffix = ".json"
```
cdkini: The `TupleFilesystemStoreBackend` should default to `.json` (currently we have some weird logic in store constructors to set this value). I'd modify the default in the signature but some things break and this felt like the path of least resistance

great_expectations/core/validation_config.py
```diff
@@ -2,13 +2,61 @@
...
+class _EncodedValidationData(BaseModel):
+    datasource: _IdentifierBundle
+    asset: _IdentifierBundle
+    batch_config: _IdentifierBundle
+
+
+def _encode_suite(suite: ExpectationSuite) -> _IdentifierBundle:
+    if not suite.id:
+        expectation_store = project_manager.get_expectations_store()
+        key = expectation_store.get_key(name=suite.name, id=suite.id)
```
billdirks: nit: It is strange to me to reference `suite.id` in a `not suite.id` block. I initially thought it would always be `None` but it looks like it could also be `""` in this branch. 
* Is that distinction important? 
* Could you add a comment calling it out as an explanation of why we can't just pass `None` here.

great_expectations/core/validation_config.py
```diff
@@ -19,30 +67,139 @@ class ValidationConfig(BaseModel):
...
+            "data": {
+                "datasource": {
+                    "name": "my_datasource",
+                    "id": "a758816-64c8-46cb-8f7e-03c12cea1d67"
+                },
+                "asset": {
+                    "name": "my_asset",
+                    "id": "b5s8816-64c8-46cb-8f7e-03c12cea1d67"
+                },
+                "batch_config": {
```
billdirks, line 94: Do we need to serialize out all three layers because we can't hydrate the asset and datasource from the batch_config alone?

great_expectations/core/validation_config.py
```diff
@@ -19,30 +67,139 @@ class ValidationConfig(BaseModel):
...
+    def _decode_data(cls, data_dict: dict) -> BatchConfig:
+        # Take in raw JSON, ensure it contains appropriate identifiers, and use them to retrieve the actual data.
+        try:
+            data_identifiers = _EncodedValidationData.parse_obj(data_dict)
+        except ValidationError as e:
+            raise ValueError(
+                "Serialized data did not contain expected identifiers"
+            ) from e
+
+        ds_name = data_identifiers.datasource.name
```
billdirks, line 173: We serialize with names and ids but only use the ids to decode. Is there a reason for serializing ids?

great_expectations/data_context/data_context/abstract_data_context.py
billdirks: ouch.

great_expectations/data_context/data_context/context_factory.py
```diff
@@ -118,6 +119,14 @@ def get_evaluation_parameters_store(self) -> EvaluationParameterStore:
             )
         return self._project.evaluation_parameter_store
 
+    def get_datasources(self) -> DatasourceDict:
+        if not self._project:
+            raise RuntimeError(
+                "This action requires an active DataContext. "
+                + "Please call `get_context()` first, then try your action again."
```
billdirks: nit: Show the full path to call: `great_expectations.get_context()`

tests/core/test_validation_config.py
```diff
@@ -0,0 +1,358 @@
...
+    @pytest.mark.unit
+    @pytest.mark.parametrize(
+        "ds_id, asset_id, batch_config_id",
+        [
+            (
+                "9a88975e-6426-481e-8248-7ce90fad51c4",
+                "9b35aa4d-7f01-420d-9d45-b45658e60afd",
+                "782c4aaf-8d56-4d8f-9982-49821f4c86c2",
+            ),
+            (None, None, None),
```
billdirks, line 64: add a hybrid case where some but not all are none?

billdirks (commented):
Thanks for doing this. I added a few small questions.

great_expectations/core/validation_config.py
cdkini: Nope you're right it will be `None` - I've updated the code to reflect that

great_expectations/core/validation_config.py
cdkini: Yes because the datasource is the top-level domain object that has an applicable store, we need to use it as our entry point in order to know which asset and batch config to grab

great_expectations/core/validation_config.py
cdkini: Hmm this is a good question. I did some research into how we're currently using identifiers and there appears to be a bit of inconsistency in how we use names vs ids. I decided to always write both to disk with the hope of eventually making ids the source of truth but perhaps YAGNI?

I think its a small price to pay and probably provides more value as-is but I'd be happy to discuss the merits of both sides

billdirks (approved):
LGTM!

PR #9562, opened 2024-03-01
[MAINTENANCE] Prettier yaml formatting

Kilo59:
Use a `prettier` pre-commit hook for consistent yaml formatting.

https://prettier.io/docs/en/precommit.html#option-2-pre-commithttpsgithubcompre-commitpre-commit

### Why

Consistent YAML formatting.
Eliminate formatting and whitespace changes to YAML files.
Makes reviewing changes and debugging easier.

This will be followed up with a `.git-blame-ignore-revs` entry.

See also...
- #9536 

Vote on `"` vs `'` in our Slack.


.pre-commit-config.yaml
```diff
@@ -22,11 +22,18 @@ repos:
...
-        args: ["--fix"]
+        args: ['--fix']
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v4.0.0-alpha.8
+    hooks:
+      - id: prettier
+        args: ['--single-quote']
+        types_or: [yaml]
+        exclude: tests/data_context/fixtures/bad_yml
   # Github Actions Linter
```
Shinnnyshinshin: 🚀

Shinnnyshinshin (approved):
so pretty

PR #9570, opened 2024-03-04
[DOCS] Hide duplicate tabs

JessSaavedra:
# Description
If `className="hidden"` is added to a Tabs component, it will have its tabs hidden (useful for nested tabs inside the same document that will always have the same value).

You can check this does not break the logic of switching tabs, there's an example in this page: [https://deploy-preview-9570.docs.greatexpectations.io/docs/oss/guides/setup/installation/install_gx](). If you select `pip` or `conda` tabs inside the `local` tab, you will see the content is updated properly, even though the second set of tabs that previously showed options `pip` and `conda` again is hidden.

# Screenshots
## Before
screencapture-docs-greatexpectations-io-docs-oss-guides-setup-installation-install-gx-2024-03-04-15_31_27

## After
screencapture-localhost-3000-docs-oss-guides-setup-installation-install-gx-2024-03-04-15_31_56

### DIFF
--- a/docs/docusaurus/src/css/tabs.scss
+++ b/docs/docusaurus/src/css/tabs.scss
@@ -8,6 +8,10 @@ div[role="tabpanel"]{
   font-weight: var(--ifm-font-weight-normal);
   line-height: var(--p-line-height);
 
+  &.hidden {
+    display: none;
+  }
+
   & .tabs__item--active {
     border-bottom: 0;
     font-weight: var(--ifm-font-weight-bold);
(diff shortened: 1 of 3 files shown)


docs/docusaurus/src/css/tabs.scss
```diff
@@ -8,6 +8,10 @@ div[role="tabpanel"]{
   font-weight: var(--ifm-font-weight-normal);
   line-height: var(--p-line-height);
 
+  &.hidden {
```
josectobar, line 11: Are we going to need to add this class to each screen in order to get this effect?

Rachel-Reverie: My understanding is that we'll add the class name to any Tab group that we want to hide the tab selection portion of.

JessSaavedra: Rachel is right, any tab that has the hidden class in the markdown will be hidden!

josectobar (approved):
👍🏽

PR #9579, opened 2024-03-06
[FEATURE] `ValidationConfig.save()`

cdkini:

### DIFF
--- a/great_expectations/core/validation_config.py
+++ b/great_expectations/core/validation_config.py
@@ -223,6 +223,7 @@ class ValidationConfig(BaseModel):
         )
         return validator.validate_expectation_suite(self.suite, evaluation_parameters)
 
+    @public_api
     def save(self) -> None:
         store = project_manager.get_validation_config_store()
         key = store.get_key(name=self.name, id=self.id)
--- a/great_expectations/core/validation_config.py
+++ b/great_expectations/core/validation_config.py
@@ -222,3 +222,12 @@ class ValidationConfig(BaseModel):
             result_format=result_format,
         )
         return validator.validate_expectation_suite(self.suite, evaluation_parameters)
+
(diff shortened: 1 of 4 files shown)


great_expectations/core/validation_config.py
```diff
@@ -222,3 +222,13 @@ def run(
             result_format=result_format,
         )
         return validator.validate_expectation_suite(self.suite, evaluation_parameters)
+
+    @public_api
+    def save(self) -> None:
+        store = project_manager.get_validation_config_store()
```
tyler-hoffman: Didn't we say we want the pattern here to be that we get the store upon initialization? (e.g. that's what we do for expectation suites). Both for raising earlier, and for predictability in case someone switches contexts. I think we could hijack a root_validator for this, even though I don't love it.

tests/core/test_validation_config.py
```diff
@@ -491,3 +497,28 @@ def test_validation_config_deserialization_non_existant_resource(
...
+
+
+@pytest.mark.unit
+def test_validation_config_save_success(
+    ephemeral_context: EphemeralDataContext, validation_config: ValidationConfig
+):
+    context = ephemeral_context
+    vc = validation_config
+
+    vc = context.validations.add(vc)
```
tyler-hoffman, line 509: nit: to make the intention of the test here more clear, having the initial `context.validations.add(vc)` being done in a fixture might make sense, e.g. like a `saved_validation_config` fixture.

tyler-hoffman (approved):

great_expectations/core/validation_config.py
cdkini: Ah okay yeah wasn’t sure the pattern to follow when using Pydantic but you’re right

PR #9582, opened 2024-03-06
[DOCS] Removes remaining OSS docs from the prerelease version

Rachel-Reverie:
## Description
Removes remaining OSS docs from the prerelease folder

## Definition of done


kwcanuck (approved):
In my opinion we should not publish the 1.0 docs until they are ready, or more complete.  I don't think incomplete documentation with under construction icons is helpful or presents the right image. However, it's not my call.

PR #9590, opened 2024-03-07
[FEATURE] V1 Checkpoint

cdkini:
The V1 Checkpoint is a significant simplification of the v0.18 model. It now only takes two primary input params (validation definitions and actions) and utilizes Pydantic.

The complexity within this PR stems from the need for custom serialization logic; due to the shared usage of models like batch definitions and expectation suites, we need to ensure we denormalize our data to obtain a single source of truth about an object's state. We do this utilizing our new pattern of serializing directly into identifiers (which can be used to index into the appropriate store to retrieve the object in its accurate state).


great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -0,0 +1,105 @@
...
+        key = validation_config_store.get_key(name=validation.name, id=None)
+        validation_config_store.add(key=key, value=validation)
+
+    return _IdentifierBundle(name=validation.name, id=validation.id)
+
+
+def _encode_action(action: ValidationAction) -> dict:
+    data = action.__dict__
+    data.pop("data_context")
+    return data
```
cdkini: Currently is not JSON-serializable

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Args:
+        validations: List of validation configs to be run.
+        actions: List of actions to be taken after the validations are run.
+
+    """
+
+    name: str
+    validations: List[ValidationConfig]
+    actions: List[ValidationAction]
```
cdkini, line 36: Must have union of all types

great_expectations/core/expectation_suite.py
```diff
@@ -1110,6 +1111,16 @@ def render(self) -> None:
                 )
             )
 
+    def serialize(self) -> _IdentifierBundle:
```
cdkini: Moved any encoding logic to the actual class being serialized (move from `_encode_suite` in validation config to `serialize` as instance method on suite).

I don't love exposing these publicly but how do we get around this? Do we want custom serializer classes to encapsulate this logic?

great_expectations/core/serdes.py
```diff
@@ -0,0 +1,16 @@
...
+
+class _IdentifierBundle(BaseModel):
+    name: str
+    id: Union[str, None]
+
+
+class _EncodedValidationData(BaseModel):
+    datasource: _IdentifierBundle
+    asset: _IdentifierBundle
+    batch_definition: _IdentifierBundle
```
cdkini, line 16: Shared space with intermediate serialization models

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -0,0 +1,122 @@
...
+    id: Union[str, None] = None
+
+    class Config:
+        arbitrary_types_allowed = (
+            True  # Necessary for compatibility with ValidationAction's Marshmallow dep
+        )
+        """
+        When serialized, the validations field should be encoded as a set of identifiers.
```
billdirks: s/should/will/

billdirks: I don't understand the position of this comment between 2 attributes of `Config` (`arbitrary_types_allowed` and `json_encoders`). It sees like ti should either above `arbitrary_types_allowed` as a top level comment in `Config` or as part of the class comment below `Args`. It's not really a user facing comment though since they don't need t worry about this so maybe at the top of Config?

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -0,0 +1,122 @@
...
+        These will be used as foreign keys to retrieve the actual objects from the appropriate stores.
+
+        Example:
```
billdirks, line 44: Thank you for the example!

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -0,0 +1,122 @@
...
+
+            if validation_config:
+                validations.append(validation_config)
+
+        return validations
+
+    @public_api
+    def run(
+        self,
+        batch_params: Dict[str, Any] | None = None,
```
billdirks: This should be batch_parameters

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -0,0 +1,122 @@
...
+        suite_params: Dict[str, Any] | None = None,
```
billdirks: We've decided to rename this `expectation_parameters`.

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -0,0 +1,122 @@
...
+    to be taken after the validation step.
+
+    Args:
+        validations: List of validation configs to be run.
+        actions: List of actions to be taken after the validations are run.
+
+    """
+
+    name: str
+    validations: List[ValidationConfig]
```
billdirks: Why not a list of `ValidationDefinitions` instead of config objects? That's what this shows: https://greatexpectations.atlassian.net/wiki/spaces/SUP/pages/917471267/Validation+Workflows

Or does the ValidationConfig need to be renamed?

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -0,0 +1,122 @@
...
+            key = store.get_key(name=id_bundle.name, id=id_bundle.id)
+
+            try:
+                validation_config = store.get(key=key)
+            except (KeyError, gx_exceptions.InvalidKeyError):
+                raise ValueError(
+                    f"Unable to retrieve validation config {id_bundle} from store"
+                )
+
+            if validation_config:
```
billdirks: Do we need this `if`? With the try/except above, it looks like this will need to be defined.

great_expectations/core/batch_config.py
```diff
@@ -47,3 +48,22 @@ def build_batch_request(
 
     def save(self) -> None:
         self.data_asset._save_batch_config(self)
+
+    def serialize(self) -> _EncodedValidationData:
```
billdirks: I would expect a method called `serialize` on an object to serialize the object but this looks like it serializes the validation definition? Could you update the name?

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -0,0 +1,122 @@
...
+
+class Checkpoint(BaseModel):
+    """
+    A Checkpoint is the primary means for validating data in a production deployment of Great Expectations.
+
+    Checkpoints provide a convenient abstraction for running a number of validations and triggering a set of actions
+    to be taken after the validation step.
+
+    Args:
+        validations: List of validation configs to be run.
```
billdirks: We should move away from using `validations` as a raw name. We should use validation_definition or validation_result, is the plain validation has led to confusion. Could you update this here and throughout the PR?

billdirks (commented):
Thanks for doing this!
I've only looked at the first 2 files. I need to run an errand and I'll look more when I return.
I've left a few comments, mostly around naming.

great_expectations/checkpoint/v1_checkpoint.py
billdirks: Oh, I think ValidationConfig does need to be renamed. That can be done in a separate PR.

cdkini: Needs to be renamed!

great_expectations/checkpoint/v1_checkpoint.py
cdkini: did it to appease `mypy` but updated to be a bit clearer

great_expectations/core/batch_config.py
cdkini: Updated to `serialize_validation_definition`!

great_expectations/core/expectation_suite.py
billdirks: This only serializes this object when it is used in another object right? I would call this `serialize_identifier_bundle()` or just `identifier_bundler()`. This name should be uniform across any object that needs to do this. I think this public since whatever object serializes this expects this method to exist. We could formalize it as an internal protocol.

great_expectations/core/validation_config.py
```diff
@@ -302,3 +264,16 @@ def _get_expectation_suite_and_validation_result_ids(
                 run_id=run_id,
             )
             return expectation_suite_identifier, validation_result_id
+
+    def serialize(self) -> _IdentifierBundle:
```
billdirks: Same comment as before. I would rename to `identifer_bundle()`.

billdirks (commented):
I will look at tests after my next meeting. Another couple name comments. Thanks!

great_expectations/core/expectation_suite.py
cdkini: I agree for suites and validation definitions but renderers and batch definitions have different outputs (`_EncodedValidationData` and `class_name/module_name`, respectively).

tests/checkpoint/test_v1_checkpoint.py
```diff
@@ -0,0 +1,335 @@
...
+                id="invalid_validation",
+            ),
+        ],
+    )
+    @pytest.mark.unit
+    def test_checkpoint_deserialization_failure(
+        self, serialized_checkpoint: str, expected_error: str
+    ):
+        with pytest.raises(ValidationError) as e:
+            Checkpoint.parse_obj(serialized_checkpoint)
```
billdirks, line 333: What's the difference between `parse_raw` used above and `parse_obj` used here? I was expected the same deserialization call in both places.

cdkini: `parse_raw` is on a string and `parse_obj` is on a dict

billdirks (approved):
LGTM!

PR #9599, opened 2024-03-11
[FEATURE] Save validation results on ValidationDefinition run

tyler-hoffman:
See title.

Also a bit of scope creep: I updated the signature of `store::set` to reflect that it returns something so that I didn't have to introduce more type ignores. I set it to `Any`, which is terrible, but I think we need a broader push to bring the return type to something consistent.


great_expectations/checkpoint/actions.py
```diff
@@ -882,71 +883,51 @@ def _run(  # type: ignore[override] # signature does not match parent  # noqa: P
...
             checkpoint_identifier,
         )
-        if self._using_cloud_context and isinstance(
-            run_return_value, GXCloudResourceRef
-        ):
-            return self._run_cloud_post_process_resource_ref(
-                run_return_value, validation_result_suite_identifier
-            )
 
-    def _basic_run(
```
tyler-hoffman, line 885: I'll remove this whole class in a followup, but leaving it now to let existing tests catch anything I might have messed up

great_expectations/core/validation_config.py
```diff
@@ -68,6 +79,7 @@ def _encode_data(data: BatchConfig) -> _EncodedValidationData:
     )
 
 
+# @dataclass_transform
```
cdkini: Could we remove this?

great_expectations/validator/v1_validator.py
```diff
@@ -83,6 +83,11 @@ def validate_expectation_suite(
             },
         )
 
+    @property
+    def active_batch_id(self) -> Optional[str]:
+        # See validation_operators line 343
```
cdkini: This feels like it might be outdated pretty easily (especially since I'm hoping to remove validation operators). Is there a less specific comment we can use?

great_expectations/checkpoint/actions.py
```diff
@@ -869,71 +870,51 @@ def _run(  # type: ignore[override] # signature does not match parent  # noqa: P
         checkpoint_identifier: Optional[GXCloudIdentifier] = None,
     ):
         logger.debug("StoreValidationResultAction.run")
-        run_return_value = self._basic_run(
+        return store_validation_results(
+            self.target_store,
```
cdkini: I think I'm leaning towards killing off this action but this feels good for now

great_expectations/core/validation_config.py
```diff
@@ -218,9 +237,43 @@ def run(
...
+                batch_identifier=validator.active_batch_id,
+                expectation_suite_identifier=expectation_suite_identifier,
+                run_id=run_id,
+            )
+
+        store_validation_results(
+            validation_result_store=self._validation_results_store,
+            suite_validation_result=results,
+            suite_validation_result_identifier=validation_result_id,
+            expectation_suite_identifier=ExpectationSuiteIdentifier(self.suite.name),
```
cdkini: Maybe we wrap this in a helper? Feels a bit hard to read

cdkini (approved):

great_expectations/checkpoint/actions.py
```diff
@@ -869,71 +870,51 @@ def _run(  # type: ignore[override] # signature does not match parent  # noqa: P
...
+        and isinstance(run_return_value, GXCloudResourceRef)
+        and isinstance(suite_validation_result_identifier, GXCloudIdentifier)
     ):
-        store_set_return_value: GXCloudResourceRef
-        new_id = gx_cloud_resource_ref.id
-        # ValidationResultIdentifier has no `.id`
-        validation_result_suite_identifier.id = new_id  # type: ignore[union-attr]
-        return gx_cloud_resource_ref
+        new_id = suite_validation_result_identifier.id
+        assert new_id is not None
```
billdirks: Instead of a raw assert could we raise an Exception with a descriptive error message?

great_expectations/core/validation_config.py
```diff
@@ -122,6 +134,13 @@ class Config:
     data: BatchConfig = Field(..., allow_mutation=False)
     suite: ExpectationSuite = Field(..., allow_mutation=False)
     id: Union[str, None] = None
+    _validation_results_store: ValidationsStore = PrivateAttr()
+
+    def __init__(self, **data: Any):
+        super().__init__(**data)
+
+        # TODO: Migrate this to model_post_init when we get to pydantic 2
+        self._validation_results_store = project_manager.get_validations_store()
```
billdirks, line 142: Can we move this to a read only property (possible cached)?

tyler-hoffman: Yeah, I considered that. The reason I put it here is because for other models (at least suites), we instantiate it in the constructor, so there'd be a slight difference in behavior by making it a property. Maybe I'll make it a cached property for now and we can iterate if needed.

cdkini: I think we have a slightly inconsistent pattern here - good either way but maybe let's discuss the idiomatic project manager pattern

great_expectations/data_context/store/store.py
```diff
@@ -218,7 +218,7 @@ def get_all(self) -> list[Any]:
 
         return list(map(self.deserialize, objs))
 
-    def set(self, key: DataContextKey, value: Any, **kwargs) -> None:
+    def set(self, key: DataContextKey, value: Any, **kwargs) -> Any:
```
cdkini, line 221: Do we know what this might be? More for my own edification I think we're gonna probably redo stores at some point so doubt we need to get super granular

tests/core/test_validation_config.py
```diff
@@ -141,7 +176,7 @@ def test_passes_complex_data_to_validator(
     @pytest.mark.unit
     def test_returns_expected_data(
         self,
-        mock_validator: mock.MagicMock,  # noqa: TID251
+        mock_validator: MagicMock,
```
cdkini, line 179: Thanks for these annotations!

tests/core/test_validation_config.py
```diff
@@ -160,6 +195,65 @@ def test_returns_expected_data(
...
+            gxe.ExpectColumnMaxToBeBetween(column="foo", max_value=1)
+        )
+        mock_validator.graph_validate.return_value = [
+            ExpectationValidationResult(success=True)
+        ]
+
+        cloud_validation_config.run()
+
+        # validate we are calling set on the store with data that's roughly the right shape
+        [(_, kwargs)] = mock_validations_store_set.call_args_list
```
cdkini, line 251: ooh I've never seen this syntax before - I guess just an alternative way of unpacking huh (given the shape of the call args list)

tests/core/test_validation_config.py
```diff
@@ -160,6 +195,65 @@ def test_returns_expected_data(
...
+        key = kwargs["key"]
+        value = kwargs["value"]
+        assert isinstance(key, ValidationResultIdentifier)
+        assert key.batch_identifier == BATCH_ID
+        assert value.success is True
+
+    @mock.patch.object(ValidationsStore, "set")
+    @pytest.mark.unit
```
cdkini, line 234: Could we mark with Cloud instead? Same for any cloud tests

cdkini (approved):

tests/core/test_validation_config.py
tyler-hoffman: yup. I kinda prefer this because it implicitly asserts that there was only one call made. Otherwise, I see the use of `call_args_list` getting really verbose with length checking, and then pulling out args/kwargs on their own lines.

great_expectations/data_context/store/store.py
tyler-hoffman: I think it varies a lot. Sometimes `None`, sometimes `bool`, sometimes a cloud identifier, etc. I hate using `Any`, but I think this is the most honest return type at this point

PR #9605, opened 2024-03-11
[MAINTENANCE] Remove context dependency from Validation Actions

cdkini:
The current data context dependency in these actions is preventing serialization; we should decouple this dependency by utilizing our project manager pattern.


great_expectations/data_context/data_context/context_factory.py
```diff
@@ -140,6 +141,57 @@ def get_validator(self, batch_request: BatchRequest) -> Validator:
...
+            site_names=site_names,
+            resource_identifiers=resource_identifiers,
+            dry_run=dry_run,
+            build_index=build_index,
+        )
+
+    def get_docs_sites_urls(
+        self,
+        resource_identifier: Any | None = None,
+        site_name: str | None = None,
```
cdkini: I haaaaaate this but what's the alternative? I could encapsulate all data docs CRUD into a `DataDocsManager` class so we can inject that but no matter what, we're going to have some messiness here.

I'd vote to get this in as-is and follow up with that docs encapsulation

tests/datasource/fluent/data_asset/data_connector/test_s3_data_connector.py
```diff
@@ -1090,6 +1090,4 @@ def test_s3_checkpoint_run_using_same_store_prefixes_errors(
         # Happens when an ExpectationIdentifier is being evaluated as ValidationResultIdentifier
         context.build_data_docs()
 
-    with pytest.raises(IndexError):
-        # Happens when an ValidationResultIdentifier is being evaluated as ExpectationIdentifier
-        checkpoint.run()
+    checkpoint.run()
```
cdkini: Honestly not sure why this is no longer failing. I've changed actions (which are likely the culprit) but they should function the exact same given the changes I've made here.

We're going to be deleting these legacy checkpoints quite soon so I'm less inclined to dig deep into this

great_expectations/checkpoint/actions.py
```diff
@@ -48,25 +48,7 @@
...
-    @property
-    def _using_cloud_context(self) -> bool:
-        # Chetan - 20221216 - This is a temporary property to encapsulate any Cloud leakage
-        # Upon refactoring this class to decouple Cloud-specific branches, this should be removed
-        from great_expectations.data_context.data_context.cloud_data_context import (
-            CloudDataContext,
-        )
-
-        return isinstance(self.data_context, CloudDataContext)
+    """Base class for all Actions that act on Validation Results and are aware of a Data Context namespace structure."""
```
billdirks: I know this is the current docstring but I don't really know what `Data Context namespace structure` means. It would be nice to update.

great_expectations/checkpoint/actions.py
```diff
@@ -217,6 +196,10 @@ def __init__(  # noqa: PLR0913
         self.notify_with = notify_with
         self.show_failed_expectations = show_failed_expectations
 
+        from great_expectations import project_manager
```
billdirks: Can we move this import to the top of the file? If not, could you add a comment.

Similar comment at other places in this file.

great_expectations/checkpoint/actions.py
```diff
@@ -973,22 +947,24 @@ class UpdateDataDocsAction(ValidationAction):
...
+        site_names: list[str] | None = None,
     ) -> None:
         """
-        :param data_context: Data Context
         :param site_names: *optional* List of site names for building data docs
         """
-        super().__init__(data_context)
         self._site_names = site_names
 
+        from great_expectations import project_manager
```
billdirks: I may consider moving this to the base class and provide one context API for all actions. I'm not sure what is better.
If the `project_manager` is internal to GX but we expect people to contribute `actions` then this would be a way to provide actions an API for these functions without making the project_manager public. If however, we don't expect anyone to contribute actions or if we expect contributed actions to not care about these things, we could do it this way.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2820,9 +2820,9 @@ def _open_url_in_browser(url: str) -> None:
 
     def get_docs_sites_urls(
         self,
-        resource_identifier=None,
+        resource_identifier: Any | None = None,
```
billdirks, line 2823: nit: You don't need `| None` here since that is part of `Any`.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -2820,9 +2820,9 @@ def _open_url_in_browser(url: str) -> None:
...
         site_name: Optional[str] = None,
-        only_if_exists=True,
+        only_if_exists: bool = True,
```
billdirks, line 2825: Thanks for adding annotations.

great_expectations/data_context/data_context/context_factory.py
```diff
@@ -140,6 +141,57 @@ def get_validator(self, batch_request: BatchRequest) -> Validator:
...
         return self._project.get_validator(batch_request=batch_request)
 
+    def is_using_cloud(self) -> bool:
+        if not self._project:
+            raise RuntimeError(
+                "This action requires an active DataContext. "
+                + "Please call `great_expectations.get_context()` first, then try your action again."
+            )
+
+        from great_expectations.data_context import CloudDataContext
```
billdirks, line 119: Can we move this import up?

great_expectations/data_context/data_context/context_factory.py
```diff
@@ -140,6 +141,57 @@ def get_validator(self, batch_request: BatchRequest) -> Validator:
             )
         return self._project.get_validator(batch_request=batch_request)
 
+    def is_using_cloud(self) -> bool:
+        if not self._project:
```
billdirks: To DRY this up, can you make this if check a private method and call it here and elsewhere?

tests/datasource/fluent/data_asset/data_connector/test_s3_data_connector.py
billdirks: I would look into this and determine if we want to preserve this functionality with the new checkpoint or not. If we don't want to preserve it, we should delete the whole test. If we do want it, we should figure out what's going on. 

The deleted comment does look fishy.

billdirks (commented):
Thanks for doing this. I've added some comments inline.

great_expectations/checkpoint/actions.py
anthonyburdi: I'd vote for moving to the base class. Actions seem like a nice way for users to contribute.

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: this should be `SomeSpecificRichType | None` but I haven't been able to fully narrow it down. I'd vote to keep it this just to give it a bit more value than a bare `Any`

great_expectations/data_context/data_context/context_factory.py
cdkini: Needs to be inline but can move to top of the func

great_expectations/checkpoint/actions.py
```diff
@@ -48,25 +48,19 @@
...
-        # Chetan - 20221216 - This is a temporary property to encapsulate any Cloud leakage
-        # Upon refactoring this class to decouple Cloud-specific branches, this should be removed
-        from great_expectations.data_context.data_context.cloud_data_context import (
-            CloudDataContext,
-        )
-
-        return isinstance(self.data_context, CloudDataContext)
+        self._using_cloud_context = project_manager.is_using_cloud()
+        self._build_data_docs = project_manager.build_data_docs
+        self._get_docs_sites_urls = project_manager.get_docs_sites_urls
```
cdkini: Moved these all to the base class but perhaps it is better to move them to a `DataDocsAction` base class?

billdirks: That does sound better! Yes, please move them there instead.

billdirks (approved):
I like your suggestion about where to put the data doc methods so please do that. Otherwise LGTM, thank you!

PR #9617, opened 2024-03-13
[MAINTENANCE] Refactor `ValidationAction` to use Pydantic

cdkini:
Since our new V1 checkpoints utilize Pydantic and want to support a rich-type instantiation pattern, we should make any nested params also leverage Pydantic.
```python
checkpoint = Checkpoint(
    validations=[ValidationDefinition(...)],
    actions=[SlackAction(...)],
)
```
We currently utilize dictionary configs so there is some complexity in removing the old `instantiate_class_from_config` pattern. Noted that nested within actions are renderers, which have a similar issue; these need to eventually be migrated to Pydantic as well (deeming this out of scope as well).


great_expectations/checkpoint/actions.py
```diff
@@ -58,6 +63,7 @@ class ValidationAction:
     def __init__(self) -> None:
         from great_expectations import project_manager
 
+        super().__init__()
```
cdkini: How can I introduce project manager injection without overriding `__init__`?

great_expectations/checkpoint/actions.py
```diff
@@ -128,6 +134,21 @@ class (for open source Great Expectations) or the GeCloudIdentifier (from Great
...
+            runtime_environment={},
+            config_defaults={},
+        )
+        if not renderer:
+            raise ClassInstantiationError(
+                module_name=config.get("module_name"),
+                package_name=None,
+                class_name=config.get("class_name"),
+            )
+        return renderer
```
cdkini: I'd like to accept the `Renderer` rich type but that introduces more complexity (need to update that class hierarchy to also use Pydantic). For now, let's continue to accept a dictionary config but build the rich type just-in-time when it needs to be used.

great_expectations/checkpoint/actions.py
```diff
@@ -179,42 +200,35 @@ class SlackNotificationAction(DataDocsAction):
...
+    slack_webhook: Optional[str] = None
+    slack_token: Optional[str] = None
+    slack_channel: Optional[str] = None
+    notify_on: Literal["all", "failure", "success"] = "all"
+    notify_with: Optional[list[str]] = None
+    show_failed_expectations: bool = False
+    renderer: dict = {
+        "class_name": "SlackRenderer",
+        "module_name": "great_expectations.render.renderer.slack_renderer",
+    }
```
cdkini: Note that I've added default values for renderers. I'd imagine this covers 99% of usecases

great_expectations/checkpoint/actions.py
```diff
@@ -179,42 +200,35 @@ class SlackNotificationAction(DataDocsAction):
...
+
+        try:
+            if not slack_token and slack_channel:
+                assert slack_webhook
+            if not slack_webhook:
+                assert slack_token and slack_channel
+            assert not (slack_webhook and slack_channel and slack_token)
+        except AssertionError:
+            raise ValueError(
+                "Please provide either slack_webhook or slack_token and slack_channel"
```
cdkini: Shifted any validation from `__init__` to validators

great_expectations/checkpoint/actions.py
cdkini: I'm providing a very sane default for all renderers so the chance a user is going to override it is very slim. We can follow up with cleanup to accept rich type renderers (but I don't think we should prioritize it)

great_expectations/checkpoint/actions.py
```diff
@@ -128,12 +160,30 @@ class (for open source Great Expectations) or the GeCloudIdentifier (from Great
...
+    _build_data_docs: Callable = PrivateAttr()
+    _get_docs_sites_urls: Callable = PrivateAttr()
+
+    def __init__(self, **data: Any) -> None:
         from great_expectations import project_manager
 
-        super().__init__()
+        super().__init__(**data)
         self._build_data_docs = project_manager.build_data_docs
         self._get_docs_sites_urls = project_manager.get_docs_sites_urls
```
cdkini: Is this the pattern for Pydantic private attrs? Feels a bit bad to override init?

great_expectations/checkpoint/actions.py
```diff
@@ -165,11 +215,9 @@ class SlackNotificationAction(DataDocsAction):
...
     Args:
         renderer: Specifies the Renderer used to generate a query consumable by Slack API, e.g.:
-            ```python
-            {
-               "module_name": "great_expectations.render.renderer.slack_renderer",
-               "class_name": "SlackRenderer",
-           }
+           ```python
+            "module_name": "great_expectations.render.renderer.slack_renderer",
+            "class_name": "SlackRenderer",
```
cdkini: Netlify issues and I don't know why 😠

great_expectations/checkpoint/actions.py
```diff
@@ -179,42 +227,41 @@ class SlackNotificationAction(DataDocsAction):
...
+    renderer: dict = {
+        "class_name": "SlackRenderer",
+        "module_name": "great_expectations.render.renderer.slack_renderer",
+    }
+
+    _renderer: SlackRenderer = PrivateAttr()
+
+    def __init__(self, **data: Any) -> None:
+        super().__init__(**data)
+        self._renderer = cast(SlackRenderer, self._build_renderer(config=self.renderer))
```
cdkini: Need to cast for all of these - is there a better solution?

docs/docusaurus/docs/snippets/checkpoints_and_actions.py
```diff
@@ -438,7 +438,6 @@
       class_name: SlackNotificationAction
       slack_webhook: <YOUR SLACK WEBHOOK URL>
       notify_on: failure
-      notify_with: all
```
cdkini, line 441: Should be an empty list if we want all - simplified the constructor below

tests/actions/test_core_actions.py
```diff
@@ -648,3 +671,121 @@ def test_cloud_sns_notification_action(
...
+        ),
+    ],
+)
+def test_action_serialization(
+    mock_context,
+    action_class: Type[ValidationAction],
+    init_params: dict,
+    expected: dict,
+):
+    action = action_class(**init_params)
```
cdkini: Need to instantiate the action within the test since we need the mock context setup first

great_expectations/checkpoint/actions.py
cdkini: Note this also ruins autocomplete in editors

tests/actions/test_core_actions.py
```diff
@@ -648,3 +671,138 @@ def test_cloud_sns_notification_action(
...
+            MicrosoftTeamsNotificationAction,
+            OpsgenieAlertAction,
+            EmailAction,
+            UpdateDataDocsAction,
+            SNSNotificationAction,
+            APINotificationAction,
+        ],
+        actual,
+    )
+    assert isinstance(parsed_obj, action_class)
```
cdkini: Trying to keep this as simple as possible but maybe not the best approach

great_expectations/checkpoint/actions.py
cdkini: Both of these classes are parent classes and won't be exposed to the end user during normal workflows

great_expectations/checkpoint/actions.py
```diff
@@ -179,42 +236,44 @@ class SlackNotificationAction(DataDocsAction):
...
-        )
-        module_name = renderer["module_name"]
-        if not self.renderer:
-            raise ClassInstantiationError(
-                module_name=module_name,
-                package_name=None,
-                class_name=renderer["class_name"],
+    class Config:
+        arbitrary_types_allowed = True
+        json_encoders = {SlackRenderer: lambda s: _encode_renderer(s)}
```
cdkini: Decided to use rich types in the `__init__` but write class_name/module_name to disk when serializing

great_expectations/checkpoint/actions.py
```diff
@@ -8,12 +8,41 @@
...
+)
+from great_expectations.render.renderer.microsoft_teams_renderer import (
+    MicrosoftTeamsRenderer,
+)
+from great_expectations.render.renderer.opsgenie_renderer import (
+    OpsgenieRenderer,
+)
+from great_expectations.render.renderer.slack_renderer import (
+    SlackRenderer,
+)
```
billdirks: Can you move these all to `great_expectations/render/renderer/__init__.py` so we can import them from 1 place?

great_expectations/checkpoint/actions.py
billdirks: You mean the reason you had to remove `{` and `}`? Do you have a link to the error?

great_expectations/checkpoint/actions.py
billdirks: What do you mean it breaks autocomplete? 

Pydantic private attrs mean you have to manage it yourself so you need to explicitly set them. 
You could make them properties and then you wouldn't have to override `__init__()`. Would this let you avoid breaking autocomplete?

great_expectations/checkpoint/actions.py
```diff
@@ -130,10 +187,13 @@ class (for open source Great Expectations) or the GeCloudIdentifier (from Great
 
 
 class DataDocsAction(ValidationAction):
-    def __init__(self) -> None:
+    _build_data_docs: Callable = PrivateAttr()
+    _get_docs_sites_urls: Callable = PrivateAttr()
+
+    def __init__(self, **data: Any) -> None:
```
billdirks: What is data here and why do you need it? Is it really an `Any`? It looks like it is defined in `ValidationAction` so all subclasses should have it in their `__init__` methods so this was previously a Liskov violation.

great_expectations/checkpoint/actions.py
```diff
@@ -179,42 +237,45 @@ class SlackNotificationAction(DataDocsAction):
...
+    slack_channel: Optional[str] = None
+    notify_on: Literal["all", "failure", "success"] = "all"
+    notify_with: Optional[List[str]] = None
+    show_failed_expectations: bool = False
+    renderer: SlackRenderer = SlackRenderer()
+
+    @validator("renderer", pre=True)
+    def _validate_renderer(cls, renderer: dict | SlackRenderer) -> SlackRenderer:
+        if isinstance(renderer, dict):
+            renderer = cast(SlackRenderer, _build_renderer(config=renderer))
```
billdirks: Can you validate the cast?

```
if renderer.__class__ != "SlackRenderer":
    raise ...
```

great_expectations/checkpoint/actions.py
```diff
@@ -179,42 +237,45 @@ class SlackNotificationAction(DataDocsAction):
...
+        return renderer
+
+    @root_validator
+    def _root_validate_slack_params(cls, values: dict) -> dict:
+        slack_token = values["slack_token"]
+        slack_channel = values["slack_channel"]
+        slack_webhook = values["slack_webhook"]
+
+        try:
+            if not slack_token and slack_channel:
```
billdirks: I think this needs to be `if not (slack_token and slack_channel)` since:
```
>>> not True and False
False
>>> not (True and False)
True
```

billdirks: I think all these `if` and `asserts` can be simplified to:
```
if webhook:
  assert not slack_token and not slack_channel
else:
  assert slack_token and slack_channel
```

great_expectations/checkpoint/actions.py
```diff
@@ -349,22 +410,18 @@ class PagerdutyAlertAction(ValidationAction):
...
-        """Create a PagerdutyAlertAction"""
+    type: Literal["pagerduty"] = "pagerduty"
+
+    api_key: str
+    routing_key: str
+    notify_on: Literal["all", "failure", "success"] = "failure"
+    severity: Literal["critical", "error", "warning", "info"] = "critical"
+
+    @root_validator
+    def _root_validate_deps(cls, values: dict) -> dict:
```
billdirks, line 418: I would update the pypd error message to say something like `"ModuleNotFoundError: No module named 'pypd'". "pypd" is required for PageDuty notification actions.`

I would also consider moving the `import` to here since I don't like the pattern of setting imports to `None` when they don't exist. This is the place we do this check.

great_expectations/checkpoint/actions.py
```diff
@@ -457,30 +514,24 @@ class MicrosoftTeamsNotificationAction(ValidationAction):
...
+    teams_webhook: str
+    notify_on: Literal["all", "failure", "success"] = "all"
+    renderer: MicrosoftTeamsRenderer = MicrosoftTeamsRenderer()
+
+    @validator("renderer", pre=True)
+    def _validate_renderer(
+        cls, renderer: dict | MicrosoftTeamsRenderer
+    ) -> MicrosoftTeamsRenderer:
+        if isinstance(renderer, dict):
+            renderer = cast(MicrosoftTeamsRenderer, _build_renderer(config=renderer))
```
billdirks: Similar comment here and below about validating the cast.

great_expectations/checkpoint/actions.py
```diff
@@ -845,16 +876,25 @@ class StoreValidationResultAction(ValidationAction):
         TypeError: validation_result_id must be of type ValidationResultIdentifier or GeCloudIdentifier, not {}.
     """
 
+    type: Literal["store_validation_result"] = "store_validation_result"
+
+    class Config:
+        arbitrary_types_allowed = True
+
+    _target_store: Store = PrivateAttr()
+
```
billdirks, line 887: This action still takes a `data_context` as an `__init__` arg. It seems like it should take a `data` argument. This should this be updated. It seems it was like this before and can be updated in a separate PR. If not as a fast follow PR to this one, it should be ticketed and added to this sprint.

tests/actions/test_core_actions.py
```diff
@@ -648,3 +671,138 @@ def test_cloud_sns_notification_action(
...
+):
+    action = action_class(**init_params)
+
+    # Test serialization
+    json_dict = action.json()
+    actual = json.loads(json_dict)
+    assert actual == expected
+
+    # Test deserialization
+    parsed_obj = parse_obj_as(
```
billdirks: This tests seems to verify it is possible to parse these json strings. But it is not invoking gx code to do it. It seems like we should wrap this `parse_obj_as` call into a gx function and then call it here. Later when we write the code to deserialize it, we will have this function. This `Union` is ugly but maybe ok for now though it should be refactored so we don't have to manually maintain a `Union` if possible.

tests/actions/test_core_actions.py
billdirks: It might be nice to also go through the `init_params`, do a `getattr` on the `parsed_obj` using the key and making sure the returned value matches.

billdirks (commented):
Thanks for doing this. I am out tomorrow so you might need to ping someone else to review.

great_expectations/checkpoint/actions.py
cdkini: In order to set the private attr, we override `BaseModel.__init__`. This will be any params from any of the subclasses so it needs to be a `**kwargs`.

I'd rather we not do this but I'm not sure how to set private attrs without some kind of post init behavior. You can't set with a root validator and we don't have an explicit post init hook like we would with dataclasses.

great_expectations/checkpoint/actions.py
cdkini: I'll be deleting this action once we get validation definitions to store their results. These changes are just to ensure tests pass for now

great_expectations/checkpoint/actions.py
```diff
@@ -130,12 +172,27 @@ class (for open source Great Expectations) or the GeCloudIdentifier (from Great
 
 
 class DataDocsAction(ValidationAction):
-    def __init__(self) -> None:
+    def _build_data_docs(
```
tyler-hoffman, line 174: Is this called?

great_expectations/checkpoint/actions.py
tyler-hoffman: To render curly brackets, you need to escape them, since they have meaning in markdown.

great_expectations/checkpoint/actions.py
```diff
@@ -915,7 +976,7 @@ def _basic_run(
         if self._using_cloud_context and expectation_suite_identifier:
             expectation_suite_id = expectation_suite_identifier.id
 
-        return self.target_store.set(
+        return self._target_store.set(  # type: ignore[func-returns-value]
```
tyler-hoffman: fyi I'm changing the return type of `set` in a PR I'm working on.

tyler-hoffman (approved):
Looks reasonable to me

great_expectations/checkpoint/actions.py
cdkini: Yup L1063 in `UpdateDataDocs` action

PR #9618, opened 2024-03-13
[DOCS] Fix markdown syntax errors

JessSaavedra:


docs/sphinx_api_docs_source/build_sphinx_api_docs.py
```diff
@@ -218,6 +218,39 @@ def _parse_and_process_html_to_mdx(
         title_str = title.get_text(strip=True)
         title_str = title_str.replace("#", "")
 
+        for item in soup.find_all("dt"):
```
josectobar: This looks great, could you throw some high level comments explaining what this is doing?

josectobar: Let's also throw a few unit tests for this fn asserting that it's doing what's expected.  There's a test file you can use `docs/sphinx_api_docs_source/test_build_sphinx_api_docs.py`. Let me know if you need to pair on this.

josectobar (changes requested):
@JessSaavedra, thanks for working on this,  it looks great. I just left a few requests.

docs/sphinx_api_docs_source/test_build_sphinx_api_docs.py
```diff
@@ -62,3 +63,26 @@ def test_convert_code_blocks():
...
+    ("<section><pre></pre></section>", "<section><pre>\r\n</pre></section>", ""),
+    ("<section><cite></cite></section>", "<section><cite>\r\n</cite></section>", "some_path/ConfiguredAssetFilesystemDataConnector.html"),
+    ("<section><span>re.console(*)</span></section>", "<section><span>re.console(&amp;#42;)</span></section>", ""),
+    ("<section><p><span>Child</span></p></section>", "<section>\r\n<p>\r\n<span>Child</span>\r\n</p></section>", "")
+])
+
+def test_apply_markdown_adjustments(input_string, output_string, html_file_path):
+    from bs4 import (
+        BeautifulSoup,
+    )
```
josectobar: Let's move this to the top.

docs/sphinx_api_docs_source/build_sphinx_api_docs.py
```diff
@@ -218,6 +219,8 @@ def _parse_and_process_html_to_mdx(
         title_str = title.get_text(strip=True)
         title_str = title_str.replace("#", "")
 
+        apply_markdown_adjustments(soup, html_file_path, html_file_contents)
```
josectobar, line 222: Can we return a copy of the values we're modifying instead of mutating the originals?

josectobar (changes requested):
Thanks, this is looking great, there are just a few more changes I'm suggesting.

docs/sphinx_api_docs_source/build_sphinx_api_docs.py
JessSaavedra: I could do that change but I don't see a big difference there as the original files we're mutating are just temporary copies for generating the markdown, do you think it's worth it? Besides, I think it might feel a little bit inconsistent with the fact that the rest of the method is going to mutate them anyway and soup gets instantiated with the html content before the changes, so I could override the param but the instance would still be using the original one :thinking:

josectobar: I think you're right. normally you'd want to always copy a parameter and make the changes to that copy instead of the original but in this case it should be fine.

josectobar (approved):
Thanks for adding the changes!

PR #9622, opened 2024-03-14
[MAINTENANCE] Reduce max-complexity from `10->8`

Kilo59:
Reduce the max allowed cyclomatic complexity allowed from `10 -> 8`
https://docs.astral.sh/ruff/rules/complex-structure/


roblim (approved):
LGTM

PR #9636, opened 2024-03-18
[MAINTENANCE] Add invoke docs --clear command.

billdirks:

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -55,7 +55,7 @@ _PTY_HELP_DESC = "Whether or not to use a pseudo terminal"
         "pty": _PTY_HELP_DESC,
     }
 )
-def sort(  # noqa: PLR0913
+def sort(
     ctx: Context,
     path: str = ".",
     check: bool = False,
@@ -95,7 +95,7 @@ def sort(  # noqa: PLR0913
         "pty": _PTY_HELP_DESC,
     }
 )
-def fmt(  # noqa: PLR0913
+def fmt(
     ctx: Context,
     path: str = ".",
     sort_: bool = True,
@@ -126,7 +126,7 @@ def fmt(  # noqa: PLR0913
         "pty": _PTY_HELP_DESC,
     }
 )
-def lint(  # noqa: PLR0913
+def lint(
     ctx: Context,
     path: str = ".",
     fmt_: bool = True,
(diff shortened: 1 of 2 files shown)


tasks.py
```diff
@@ -606,12 +606,13 @@ def api_docs(ctx: Context):
         "lint": "Run the linter",
     },
 )
-def docs(
+def docs(  # noqa: PLR0913
```
billdirks: We now have 6 arguments because there is now a `--clear` flag. I could create a new command to do this but this felt better. Let me know if you have an alternative.

tasks.py
```diff
@@ -634,20 +635,21 @@ def docs(
     elif version:
         docs_builder = DocsBuilder(ctx, docusaurus_dir)
         docs_builder.create_version(version=parse_version(version))
-    else:  # noqa: PLR5501
-        if start:
-            ctx.run(" ".join(["yarn start"]), echo=True)
+    elif start:
+        ctx.run(" ".join(["yarn start"]), echo=True)
+    elif clear:
```
billdirks, line 641: This is the new condition.

The rest of the change here is pulling the start and embedded else branches to the top level to remove PLR5501:
```
else:
  if start:
     ...
  else:
```

tasks.py
Kilo59: I think we can turn off this rule for `tasks.py`
`tool.ruff.lint.per-file-ignores`

Kilo59 (approved):

tasks.py
Kilo59: I would add a brief description for this param.
```python
@invoke.task(
    name="docs",
    help={
        "build": "Build docs via yarn build instead of serve via yarn start. Default False.",
        "start": "Only run yarn start, do not process versions. For example if you have already run invoke docs and just want to serve docs locally for editing.",
        "lint": "Run the linter",
        "clear": "delete stuff 🤷 "
    },
)
def docs(
```

```console
$ invoke docs --help

Usage: inv[oke] [--core-opts] docs [--options] [other tasks here ...]

Docstring:
  Build documentation site, including api documentation and earlier doc versions. Note: Internet access required to download earlier versions.

Options:
  -b, --build                   Build docs via yarn build instead of serve via yarn start. Default False.
  -l, --lint                    Run the linter
  -s, --start                   Only run yarn start, do not process versions. For example if you have already run invoke docs and just want to serve docs locally for editing.
  -v STRING, --version=STRING
```

billdirks: Done.

PR #9647, opened 2024-03-21
[MAINTENANCE] `Checkpoint.run()`

cdkini:
First pass for V1's `Checkpoint.run()`. Will need to follow up for proper action support and plumbing of validation URL


great_expectations/core/validation_config.py
```diff
@@ -194,6 +194,22 @@ def _decode_data(cls, data_dict: dict) -> BatchConfig:
 
         return batch_definition
 
+    @property
+    def active_batch_id(self) -> str | None:
+        validator = self._build_validator()
+        return validator.active_batch_id
```
cdkini: This feels a bit weird but I need a way to grab the `active_batch_id` during the `run` call. Perhaps this property doesn't make sense outside of that context (since "active" denotes something in progress).

An alternative would be grabbing this ID during the run call and updating a private attr?

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -120,7 +129,54 @@ def run(
...
+            run_id=run_id,
+        )
+        self._run_actions(run_results=run_results)
+
+        return CheckpointResult(
+            run_id=run_id,
+            run_results=run_results,
+            checkpoint_config=self,
+            validation_result_url=None,  # Need to plumb from actions
+        )
```
cdkini: Does this overall result get persisted anywhere? I don't think so?

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -120,8 +129,97 @@ def run(
...
         expectation_parameters: Dict[str, Any] | None = None,
     ) -> CheckpointResult:
-        raise NotImplementedError
+        run_id = RunIdentifier(run_time=dt.datetime.now(dt.timezone.utc))
+        run_results = self._run_validation_definitions(
+            batch_parameters=batch_parameters,
+            expectation_parameters=expectation_parameters,
+            run_id=run_id,
+        )
+        self._run_actions(run_results=run_results)
```
cdkini, line 147: no-op for now - will follow up shortly

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -120,8 +129,97 @@ def run(
...
+
+        return CheckpointResult(
+            run_id=run_id,
+            run_results=run_results,
+            checkpoint_config=self,
+            validation_result_url=None,  # TODO: Need to plumb from actions
```
cdkini, line 153: `None` for now - will follow up shortly

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -120,8 +129,97 @@ def run(
...
+        )
+
+    def _run_validation_definitions(
+        self,
+        batch_parameters: Dict[str, Any] | None,
+        expectation_parameters: Dict[str, Any] | None,
+        run_id: RunIdentifier,
+    ) -> Dict[ValidationResultIdentifier, ExpectationSuiteValidationResult]:
+        run_results: Dict[ValidationResultIdentifier, ExpectationSuiteValidationResult] = {}
```
cdkini, line 163: Follows the same structure as the legacy checkpoint

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -120,8 +129,97 @@ def run(
...
+    success: Optional[bool] = None
+
+    class Config:
+        arbitrary_types_allowed = True
+
+    @root_validator
+    def _root_validate_result(cls, values: dict) -> dict:
+        run_results = values["run_results"]
+        if len(run_results) == 0:
+            raise ValueError("CheckpointResult must contain at least one run result")
```
cdkini, line 223: Don't envision this ever happening but still testing for it to be thorough

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -120,8 +129,97 @@ def run(
...
     def save(self) -> None:
         raise NotImplementedError
+
+
+class CheckpointResult(BaseModel):
+    run_id: RunIdentifier
+    run_results: Dict[ValidationResultIdentifier, ExpectationSuiteValidationResult]
+    checkpoint_config: Checkpoint
+    validation_result_url: Optional[str] = None
+    success: Optional[bool] = None
```
cdkini, line 214: Should always be set by the root validator rather than a user (but so should this entire object)

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -120,8 +129,97 @@ def run(
...
+        return {
+            "success": success_count == len(run_result_descriptions),
+            "success_percent": success_count / len(run_result_descriptions) * 100,
+            "validation_results": run_result_descriptions,
+        }
+
+    @public_api
+    def describe(self) -> str:
+        """JSON string description of this CheckpointResult"""
+        return json.dumps(self.describe_dict(), indent=4)
```
cdkini: Same model as the other `describe()` methods we have - took a stab at returning important information but I'm happy to adjust as we see fit

cdkini: It becomes quite difficult to munge so perhaps those helper methods on the legacy checkpoint result are worth investigating

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -120,8 +129,97 @@ def run(
...
+
+
+CheckpointResult.update_forward_refs()
```
cdkini, line 255: Sadly necessary but I'm not sure why

tests/checkpoint/test_v1_checkpoint.py
```diff
@@ -329,3 +339,275 @@ def test_checkpoint_deserialization_failure(
...
+            actual = result.describe_dict()
+
+        validation_results = actual.pop("validation_results")
+        assert len(validation_results) == 1
+
+        expected = {"success": True, "success_percent": 100.0}
+        assert actual == expected
+
+    @pytest.mark.filesystem
+    def test_checkpoint_run_filesytem_e2e(self, tmp_path: pathlib.Path):
```
cdkini, line 548: Since we're getting into precarious territory with upcoming deletions of legacy code, I wanted extra coverage here. The unit tests should also cover these code paths but I wanted to showcase how a user might actually run this domain object in production

great_expectations/core/validation_config.py
billdirks: This is global state that needs to exist but only when we are inside a `run` call?
That seems like we need to redesign something but likely don't have to the time to do it now.
This seems strange because we need create a new validator each time and we should only need the validator in `run` and we should be using the same validator there as here.

For now, I would use your alternative suggestion and set this to `None` when not in a run call. That is, wrap run in a try/finally block and set the private attr after creating the validator there and reset it to None in the `finally` block. I would also add a comment explaining what this global state is.

great_expectations/checkpoint/v1_checkpoint.py
billdirks: Is the validation result url only appropriate for cloud checkpoint runs? I would remove this attribute for now until we plumb it in.

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -120,8 +129,115 @@ def run(
...
+    @property
+    def name(self) -> str:
+        return self.checkpoint_config.name
+
+    def describe_dict(self) -> CheckpointDescriptionDict:
+        success_count = sum(1 for r in self.run_results.values() if r.success)
+        run_result_descriptions = [r.describe_dict() for r in self.run_results.values()]
+
+        return {
+            "success": success_count == len(run_result_descriptions),
```
billdirks: nit: make `len(run_result_descriptions)` a variable with a nice name like `num_results` and use in this dict.

great_expectations/checkpoint/v1_checkpoint.py
billdirks: `Checkpoint` and `CheckpointResult` both refer to each other (the `run` method of checkpoint returns a `CheckpointResult`). Maybe this is the issue?

tests/checkpoint/test_v1_checkpoint.py
```diff
@@ -329,3 +339,288 @@ def test_checkpoint_deserialization_failure(
...
+        _ = checkpoint.run(
+            batch_parameters=batch_parameters, expectation_parameters=expectation_parameters
+        )
+
+        validation_definition.run.assert_called_with(
+            batch_definition_options=batch_parameters, evaluation_parameters=expectation_parameters
+        )
+
+    @pytest.mark.unit
+    def test_checkpoint_run_with_actions(self):
```
billdirks: nit: I'd delete this for now.

billdirks (commented):
Thanks for doing this. My main comment is about the run_id and I have a question about the `validation_result_url`.

great_expectations/checkpoint/v1_checkpoint.py
cdkini: Leaving a comment about circular references (will timebox an effort to determine the cause)

great_expectations/checkpoint/v1_checkpoint.py
cdkini: I believe it is but it is a top-level param (just being explicit about the default here). I'd vote to keep this as-is but let's discuss

tests/checkpoint/test_v1_checkpoint.py
```diff
@@ -329,3 +339,278 @@ def test_checkpoint_deserialization_failure(
...
+
+        mock_run_result = ExpectationSuiteValidationResult(
+            success=True,
+            results=[
+                ExpectationValidationResult(
+                    success=True,
+                    expectation_config=ExpectationConfiguration(
+                        expectation_type="expect_column_values_to_be_between",
+                        kwargs={
+                            "batch_id": f"{self.datasource_name}-{self.asset_name}",
```
billdirks, line 386: I'm surprised the `batch_id` is in the `kwargs`. It looks like this was true before your plumbing of the `batch_id` into the validation result so this is a different thing? I'm curious why we don't see the `batch_id` derived from the `active_batch_id` in this result object?

cdkini: Hmm yeah I'm not too sure 🤔 

This EVR is derived from the E2E test below. I've tested separately and it looks like this particular expectation call somehow grabs that batch_id:
```python
gxe.ExpectColumnValuesToBeBetween(
    column=self.column_name, min_value=0, max_value=10
)
```

cdkini: we're specifically missing it in this object because I just forgot to add it 😅 it should be there

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -34,6 +44,7 @@ class Checkpoint(BaseModel):
     name: str
     validation_definitions: List[ValidationConfig]
     actions: List[ValidationAction]
+    result_format: ResultFormat = ResultFormat.SUMMARY
```
billdirks, line 50: You should add this to the class docstring.

great_expectations/core/expectation_validation_result.py
```diff
@@ -513,6 +513,7 @@ def __init__(  # noqa: PLR0913
         evaluation_parameters: Optional[dict] = None,
         statistics: Optional[dict] = None,
         meta: Optional[ExpectationSuiteValidationResult | dict] = None,
+        batch_id: Optional[str] = None,
```
billdirks, line 516: You likely need to update the example result in the class level docstring.

tests/checkpoint/test_v1_checkpoint.py
billdirks: Does this need a batch_id added to it?

billdirks (approved):
LGTM, thanks!

PR #9651, opened 2024-03-22
[FEATURE] Add support for V1 Cloud Backend endpoints

joshua-stauffer:


great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -38,62 +36,9 @@ class ErrorPayload(TypedDict):
...
-                "organization_id": organization_id,
-                attributes_key: attributes_value,
-                **kwargs,
-            },
-        }
-    }
-    return data
+class EndpointVersion(str, Enum):
+    V0 = "V0"
+    V1 = "V1"
```
Kilo59, line 41: Why does the client need to care about versions other than V1?

joshua-stauffer: in order to support incremental migration - this store needs to know whether it should construct a V1 shape or a V0 shape.

Kilo59: @joshua-stauffer 
Is there a doc or something that explains why we want to support this?
It seems like it would add a fair amount of complexity for marginal value.

joshua-stauffer: Thanks for the comments @Kilo59! there isn't a specific doc on the OSS side of the changes, but this approach allows us to continue testing against cloud while the integration work is underway. For instance, if we were not to take this incremental approach, we couldn't load a V1 DataContext or Datasources. I'd rather keep exercising as many of these code paths as possible while we gradually migrate things over.

great_expectations/data_context/store/expectations_store.py
```diff
@@ -30,6 +31,31 @@
     _TExpectation = TypeVar("_TExpectation", bound=Expectation)
 
 
+class ExpectationConfigurationDTO(pydantic.BaseModel):
+    class Config:
+        extra = pydantic.Extra.ignore
+
+    id: str
+    expectation_type: str
+    rendered_content: Optional[List[dict]]
```
cdkini: Isn't this just an optional dict? Where does the list come in?

great_expectations/data_context/store/gx_cloud_store_backend.py
Kilo59: I see, so this doesn't necessarily mean we will release with both code paths on the OSS side (although we might).

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: it actually [is a list of dicts](https://github.com/great-expectations/great_expectations/blob/4ecacb9c2eeba6d2ee16978da1f97ec4161426ae/great_expectations/expectations/expectation.py#L293)!

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -664,3 +624,107 @@ def _validate_key(self, key) -> None:
...
+        are deprecated in GX V1, and are only required for resources still using V0 endpoints.
+        """
+        version = cls._ENDPOINT_VERSION_LOOKUP.get(resource_type, EndpointVersion.V0)
+        if version == EndpointVersion.V1:
+            if isinstance(attributes_value, dict):
+                payload = {**attributes_value, **kwargs}
+            elif attributes_value is None:
+                payload = kwargs
+            else:
+                raise TypeError(
```
joshua-stauffer: this coverage annotation is incorrect; see `test_construct_json_payload_raises_error_with_V1_resource_and_wrong_attributes_value_type`

great_expectations/data_context/store/expectations_store.py
```diff
@@ -103,14 +128,14 @@ def gx_cloud_response_json_to_object_dict(response_json: Dict) -> Dict:
...
-        ge_cloud_suite_id: str = suite_data["id"]
-        suite_dict: Dict = suite_data["attributes"]["suite"]
-        suite_dict["id"] = ge_cloud_suite_id
 
-        # Temporary fork to account for pre-V1 configs
-        suite_dict.pop("ge_cloud_id", None)
+        # Cloud backend adds a default result format type of None, so ensure we remove it:
+        for expectation in suite_data.get("expectations", []):
+            if expectation.get("kwargs", {}).get("result_format", False) is None:
+                expectation["kwargs"].pop("result_format")
```
joshua-stauffer: this coverage check is incorrect; see `test_expectation_store.test_gx_cloud_response_json_to_object_dict['null_result_format']`

great_expectations/data_context/store/expectations_store.py
```diff
@@ -197,40 +222,34 @@ def _add(self, key, value, **kwargs):
...
             if self.cloud_mode:
                 # cloud backend has added IDs, so we update our local state to be in sync
-                # todo: add back this logic when `update` returns the updated object
-                # result = cast(GXCloudResourceRef, result)
-                # cloud_suite = ExpectationSuite(
-                #     **result.response["data"]["attributes"]["suite"]
-                # )
-                _suite_identifier, cloud_suite = self._refresh_suite(value)
+                result = cast(GXCloudResourceRef, result)
+                suite_kwargs = self.deserialize(
```
joshua-stauffer: this coverage check is incorrect; see `test_expectations_store.test_update_cloud_suite`.

great_expectations/data_context/store/expectations_store.py
Kilo59: @joshua-stauffer, it's possible I missed adding a coverage report upload for this test (or it isn't actually being run).

What test step is this running on? `could-tests` I would assume.

great_expectations/data_context/store/gx_cloud_store_backend.py
cdkini: Seems like this coverage report is creating a lot of unnecessary noise 🤔

great_expectations/data_context/store/expectations_store.py
```diff
@@ -103,14 +128,14 @@ def gx_cloud_response_json_to_object_dict(response_json: Dict) -> Dict:
...
             suite_data = response_json["data"]
-        ge_cloud_suite_id: str = suite_data["id"]
-        suite_dict: Dict = suite_data["attributes"]["suite"]
-        suite_dict["id"] = ge_cloud_suite_id
 
-        # Temporary fork to account for pre-V1 configs
-        suite_dict.pop("ge_cloud_id", None)
+        # Cloud backend adds a default result format type of None, so ensure we remove it:
+        for expectation in suite_data.get("expectations", []):
+            if expectation.get("kwargs", {}).get("result_format", False) is None:
```
tyler-hoffman: non-blocking: This line took me a second to get my head around, mostly around the default return of `False`. I think pulling out kwargs and then `if "result_format" in kwargs and kwargs["result_format"] is None:` might better communicate the intention. But maybe this is just preference 🤷

tests/integration/cloud/rest_contracts/test_expectation_suites.py
```diff
@@ -366,15 +339,15 @@ def test_put_non_existent_expectation_suite(
 
 
 @pytest.mark.cloud
-@pytest.mark.skip(reason="unexpected 401 instead of 204 in CI only")
+# @pytest.mark.skip(reason="unexpected 401 instead of 204 in CI only")
```
cdkini: Should we delete this instead?

great_expectations/data_context/store/gx_cloud_store_backend.py
Kilo59: That should be fixable.

tests/integration/cloud/rest_contracts/test_expectation_suites.py
joshua-stauffer: ah yeah definitely! i was curious to see if it passed, and i guess it does 😄

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -111,7 +56,10 @@ def get_user_friendly_error_message(
 
     try:
         error_json: ErrorPayload = http_exc.response.json()
-        errors = error_json.get("errors")
+        if isinstance(error_json, list):
```
tyler-hoffman, line 59: Haven't made it through the whole PR yet, but what changed so that the json can be either a list of a dict with a nested list?

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -664,3 +624,107 @@ def _validate_key(self, key) -> None:
...
+                url = f"{url}/{id}"
+        return url
+
+    @classmethod
+    def construct_versioned_payload(  # noqa: PLR0913
+        cls,
+        resource_type: str,
+        organization_id: str,
+        attributes_key: str,
+        attributes_value: Union[dict, Any],
```
tyler-hoffman, line 661: `Union[dict, Any]` is the same as `Any`. Is this the intended type?

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -664,3 +624,107 @@ def _validate_key(self, key) -> None:
...
+        if resource_id:
+            data["data"]["id"] = resource_id
+        return data
+
+    @classmethod
+    def _construct_json_payload_v1(
+        cls,
+        payload: dict,
+    ) -> dict:
+        return {
```
tyler-hoffman, line 722: Love the simplified structure.

great_expectations/data_context/store/expectations_store.py
```diff
@@ -197,40 +222,34 @@ def _add(self, key, value, **kwargs):
...
 
             if self.cloud_mode:
                 # cloud backend has added IDs, so we update our local state to be in sync
-                # todo: add back this logic when `update` returns the updated object
-                # result = cast(GXCloudResourceRef, result)
-                # cloud_suite = ExpectationSuite(
-                #     **result.response["data"]["attributes"]["suite"]
-                # )
-                _suite_identifier, cloud_suite = self._refresh_suite(value)
+                result = cast(GXCloudResourceRef, result)
```
cdkini: `cast` vs `if not isinstance raise`? I don't have issues either way but I think we've leaned towards the latter just to catch issues earlier on.

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -664,3 +624,107 @@ def _validate_key(self, key) -> None:
...
+            )
+            if id:
+                url = f"{url}/{id}"
+        else:  # default to EndpointVersion.V0
+            url = urljoin(
+                base_url,
+                f"organizations/{organization_id}/{hyphen(resource_name)}",
+            )
+            if id:
+                url = f"{url}/{id}"
```
cdkini: Nit
```suggestion
        if version == EndpointVersion.V1:
            url = urljoin(
                base_url,
                f"api/v1/organizations/{organization_id}/{hyphen(resource_name)}",
            )
        else:  # default to EndpointVersion.V0
            url = urljoin(
                base_url,
                f"organizations/{organization_id}/{hyphen(resource_name)}",
            )
        if id:
            url = f"{url}/{id}"
```

great_expectations/data_context/store/expectations_store.py
Kilo59: I think `cast` should be an absolute last resort. It's essentially a `type: ignore` but without the benefit of seeing the error code that indicates what the issue was.

tyler-hoffman (approved):

great_expectations/data_context/store/expectations_store.py
cdkini: Hmm fair enough. Then I'd vote to change this to the latter to raise some kind of runtime error

tyler-hoffman: FWIW I've typically used an assert isinstance when we are in a place int the code where we are telling mypy we know what the type must be, and done isinstance checks + raising when we are actually trying to verify the type. Agreed on avoiding `cast`

great_expectations/data_context/store/gx_cloud_store_backend.py
joshua-stauffer: yes - `Any` is the V0 type, and `dict` is the V1 type. So as we migrate things over we'll be able to drop the `Any`. Deprecated status should be indicated in the docstring.

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: haha yeah fair, i'll update per your suggestion, definitely clearer. Thanks!

great_expectations/data_context/store/gx_cloud_store_backend.py
joshua-stauffer: pydantic validation errors by default are formatted as a list

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: [7fff1cd](https://github.com/great-expectations/great_expectations/pull/9651/commits/7fff1cd0113ef492adffd7a28ac7526d87d54a24)

great_expectations/data_context/store/gx_cloud_store_backend.py
joshua-stauffer: definitely clearer, thanks! updated in [15e4b04](https://github.com/great-expectations/great_expectations/pull/9651/commits/15e4b04fc2dbfdc0cc6cf271e171261abb8d9cf4)

great_expectations/data_context/store/expectations_store.py
joshua-stauffer: yeah, this issue is due to our overly complex use of inheritance in the hierarchy of store and backend implementations. The result will always be `GXCloudResourceRef` when using the `GXCloudStoreBackend`, so I'll update to `assert isinstance`.

tests/integration/cloud/rest_contracts/test_expectation_suites.py
```diff
@@ -363,68 +332,3 @@ def test_put_non_existent_expectation_suite(
...
-            query=query,
-        )
-        .will_respond_with(
-            status=status,
-        )
-    )
-
-    with pact_test:
-        with pytest.raises(DataContextError):
-            cloud_data_context.delete_expectation_suite(expectation_suite_name=query["name"])
```
joshua-stauffer, line 430: these tests do not work with the V1 API (suites factory), which requires the suite object in order to delete. There's conversation about changing that to name/ID. If that change occurs, these tests can be added back.

PR #9656, opened 2024-03-24
[MAINTENANCE] Add validation result URL support within V1 Checkpoint

cdkini:
Ensure that Cloud URLs are passed down through to both validation definition and checkpoint results

### DIFF
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -887,13 +887,11 @@ class StoreValidationResultAction(ValidationAction):
         checkpoint_identifier: Optional[GXCloudIdentifier] = None,
     ):
         logger.debug("StoreValidationResultAction.run")
-        output = store_validation_results(
-            self._target_store,
+        output = self._target_store.store_validation_results(
             validation_result_suite,
             validation_result_suite_identifier,
             expectation_suite_identifier,
             checkpoint_identifier,
-            self._using_cloud_context,
         )
 
         if isinstance(output, GXCloudResourceRef) and isinstance(
@@ -903,33 +901,6 @@ class StoreValidationResultAction(ValidationAction):
 
         if self._using_cloud_context and isinstance(output, GXCloudResourceRef):
             return output
-
-
-def store_validation_results(  # noqa: PLR0913
-    validation_result_store: ValidationsStore,
-    suite_validation_result: ExpectationSuiteValidationResult,
-    suite_validation_result_identifier: ValidationResultIdentifier | GXCloudIdentifier,
-    expectation_suite_identifier: Optional[ExpectationSuiteIdentifier | GXCloudIdentifier] = None,
-    checkpoint_identifier: Optional[GXCloudIdentifier] = None,
-    cloud_mode: bool = False,
-) -> bool | GXCloudResourceRef:
-    """Helper function to do the heavy lifting for StoreValidationResultAction and ValidationConfigs.
-    This is broken from the ValidationAction (for now) so we don't need to pass the data_context around.
-    """  # noqa: E501
-    checkpoint_id = None
-    if cloud_mode and checkpoint_identifier:
-        checkpoint_id = checkpoint_identifier.id
-
-    expectation_suite_id = None
-    if isinstance(expectation_suite_identifier, GXCloudIdentifier):
-        expectation_suite_id = expectation_suite_identifier.id
-
-    return validation_result_store.set(
-        key=suite_validation_result_identifier,
-        value=suite_validation_result,
-        checkpoint_id=checkpoint_id,
-        expectation_suite_id=expectation_suite_id,
-    )
 
 
 @public_api
--- a/great_expectations/checkpoint/v1_checkpoint.py
+++ b/great_expectations/checkpoint/v1_checkpoint.py
@@ -8,7 +8,7 @@ import great_expectations.exceptions as gx_exceptions
 from great_expectations import project_manager
 from great_expectations._docs_decorators import public_api
 from great_expectations.checkpoint.actions import ValidationAction  # noqa: TCH001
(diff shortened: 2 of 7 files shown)


great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -150,7 +150,7 @@ def run(
             run_id=run_id,
             run_results=run_results,
             checkpoint_config=self,
-            validation_result_url=None,  # TODO: Need to plumb from actions
+            validation_result_urls=[result.result_url for result in run_results.values()],
```
cdkini: Rather than a single URL like we support in the v0.18 checkpoint, I'm opting to have a collection. These are ordered in the same way that the actual results are but let me know if that isn't super clear or poses usability concerns

great_expectations/core/expectation_validation_result.py
```diff
@@ -514,6 +514,7 @@ def __init__(  # noqa: PLR0913
         statistics: Optional[dict] = None,
         meta: Optional[ExpectationSuiteValidationResult | dict] = None,
         batch_id: Optional[str] = None,
+        result_url: Optional[str] = None,
```
cdkini, line 517: I don't love expanding the definition of what an EVR is but this seems like the path of least resistance to pass the result URL down the call stack

great_expectations/core/validation_config.py
```diff
@@ -215,14 +216,19 @@ def run(
...
             suite_validation_result=results,
             suite_validation_result_identifier=validation_result_id,
             expectation_suite_identifier=expectation_suite_identifier,
             cloud_mode=self._validation_results_store.cloud_mode,
         )
 
+        # TODO(cdkini): Don't love the placement but this gets the URL from Cloud
+        if isinstance(ref, GXCloudResourceRef):
+            result_url = ref.response["data"]["attributes"]["validation_result"]["display_url"]
+            results.result_url = result_url
```
cdkini: This feels quite bad - where does this logic go? Probably something Cloud-specific but our encapsulation leaves a lot to be desired.

I think cleaning up the actions and other code from v0.18 will allow an opportunity to place this in a more relevant file but I'd like to discuss what makes the most sense

great_expectations/checkpoint/v1_checkpoint.py
joshua-stauffer: is this field meant to be used directly by the end user? I wonder if a more idiomatic place to put the url would be inside the `validation_results` key in `CheckpointResult.describe()`.

great_expectations/core/validation_config.py
joshua-stauffer: Yeah I agree, this isn't great for our abstractions. It might make more sense for the store to own this logic, and the ValidationDefinition can retrieve it via `Store.get_result_url(ValidationResult)` or something. It's still not perfect, but at least cloud specific abstractions aren't leaking here.

joshua-stauffer (changes requested):
left two comments on the general approach

great_expectations/checkpoint/v1_checkpoint.py
cdkini: Hmm I quite like that actually!

great_expectations/checkpoint/actions.py
```diff
@@ -905,33 +903,6 @@ def _run(  # type: ignore[override] # signature does not match parent  # noqa: P
...
 
 
-def store_validation_results(  # noqa: PLR0913
-    validation_result_store: ValidationsStore,
-    suite_validation_result: ExpectationSuiteValidationResult,
-    suite_validation_result_identifier: ValidationResultIdentifier | GXCloudIdentifier,
-    expectation_suite_identifier: Optional[ExpectationSuiteIdentifier | GXCloudIdentifier] = None,
-    checkpoint_identifier: Optional[GXCloudIdentifier] = None,
-    cloud_mode: bool = False,
-) -> bool | GXCloudResourceRef:
```
cdkini, line 915: Made this a function of the validations store; will remove this action here entirely soon enough

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -210,7 +209,6 @@ class CheckpointResult(BaseModel):
     run_id: RunIdentifier
     run_results: Dict[ValidationResultIdentifier, ExpectationSuiteValidationResult]
     checkpoint_config: Checkpoint
-    validation_result_url: Optional[str] = None
```
cdkini, line 213: result url can be found in `describe()` but will not be a top level attr

great_expectations/core/expectation_validation_result.py
```diff
@@ -636,11 +638,15 @@ def get_failed_validation_results(
...
 
     def describe_dict(self) -> dict:
-        return {
+        data = {
             "success": self.success,
             "statistics": self.statistics,
             "expectations": [expectation.describe_dict() for expectation in self.results],
         }
+        if self.result_url:
+            data["result_url"] = self.result_url
```
cdkini: Conditionally adding it to the describe payload since OSS users will always see nulls. Is that worth an inconsistent payload shape?

great_expectations/core/validation_config.py
```diff
@@ -215,14 +215,15 @@ def run(
...
+        ref = self._validation_results_store.store_validation_results(
             suite_validation_result=results,
             suite_validation_result_identifier=validation_result_id,
             expectation_suite_identifier=expectation_suite_identifier,
-            cloud_mode=self._validation_results_store.cloud_mode,
         )
 
+        if isinstance(ref, GXCloudResourceRef):
+            results.result_url = self._validation_results_store.get_result_url(ref)
+
```
cdkini: All of these can probably be brought together but that can be done after the store validation result action is removed

joshua-stauffer (approved):
thanks for addressing comments, looks good!

PR #9659, opened 2024-03-25
[MAINTENANCE] V1 Checkpoint Store

cdkini:
Due to utilizing a new model, we need to rewrite the store for our V1 checkpoint. This will be promoted to the "CheckpointStore" namespace once more work is done around cleaning up legacy code.


great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -224,3 +228,65 @@ def create(self, checkpoint_config: CheckpointConfig) -> Optional[DataContextKey
         config = self.get(key=key)
 
         return config
+
+
+# NOTE: Will eventually be promoted to 'CheckpointStore' once the legacy Checkpoint
+#       and its related classes are removed
+class V1CheckpointStore(Store):
+    _key_class = StringKey
```
cdkini, line 237: Follows the exact same model as the `ValidationConfigStore` I wrote recently.

Once this is in, I will take the following steps:
- Delete the legacy checkpoint, checkpoint config, checkpoint store, and store validation result action
- Wire up this new store with the checkpoint factory

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -224,3 +229,65 @@ def create(self, checkpoint_config: CheckpointConfig) -> Optional[DataContextKey
...
+
+    def get_key(self, name: str, id: str | None = None) -> GXCloudIdentifier | StringKey:
+        """Given a name and optional ID, build the correct key for use in the CheckpointStore."""
+        if self.cloud_mode:
+            return GXCloudIdentifier(
+                resource_type=GXCloudRESTResource.CHECKPOINT,
+                id=id,
+                resource_name=name,
+            )
+        return StringKey(key=name)
```
billdirks: It seems like you be using `self._key_class` instead of StringKey directly here or get rid of `_key_class`. Do we actually need that class variable?
The name of it is also a bit confusing since it's only the key class in the non-cloud setting.

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -224,3 +229,65 @@ def create(self, checkpoint_config: CheckpointConfig) -> Optional[DataContextKey
...
+    @override
+    def deserialize(self, value):
+        from great_expectations.checkpoint.v1_checkpoint import Checkpoint as V1Checkpoint
+
+        return V1Checkpoint.parse_raw(value)
+
+    @override
+    def _add(self, key: DataContextKey, value: V1Checkpoint, **kwargs):
+        if not self.cloud_mode:
+            # this logic should move to the store backend, but is implemented here for now
```
billdirks: I don't know if you need this comment. It seems like a lot of this should get moved out of this class if/when we redo stores.

tests/data_context/store/test_v1_checkpoint_store.py
```diff
@@ -0,0 +1,290 @@
...
+            "ge_cloud_credentials": {
+                "access_token": cloud_details.access_token,
+                "organization_id": cloud_details.org_id,
+            },
+        },
+    )
+
+
+@pytest.fixture
+def mock_cp_json() -> dict:
```
billdirks: Can you spell out checkpoint here and below?

tests/data_context/store/test_v1_checkpoint_store.py
```diff
@@ -0,0 +1,290 @@
...
+    cp.json.return_value = json.dumps(mock_cp_json)
+    cp.dict.return_value = mock_cp_dict
+    return cp
+
+
+@pytest.mark.parametrize("store_fixture", ["ephemeral_store", "file_backed_store"])
+@pytest.mark.unit
+def test_add(request, store_fixture: str, checkpoint: Checkpoint):
+    store: ValidationConfigStore = request.getfixturevalue(store_fixture)
+    key = StringKey(key="my_checkpoint")
```
billdirks: Do users need to know that they should be using `StringKey` when using the stores? I guess these users are Store developers? I seems like we should we call `get_key` here instead or `_key_class` should be public.
I also wonder, if we always use `StringKey`, if methods such as `add` should take a union of a `str` or a `StringKey` and convert it for the caller. That avoids the whole question of someone needing to know about `StringKey` or `_key_class`.

billdirks (commented):
Thanks for doing this. I have a few small comments, mostly around `_key_class`.

great_expectations/data_context/store/checkpoint_store.py
cdkini: hmm I don't love it but I'd rather be explicit in this method here - there's other underlying logic that uses the key class property in the actual parent store that I want to make sure we have covered by being explicit about the type.

tests/data_context/store/test_v1_checkpoint_store.py
cdkini: These are just for store developers but I didn't want to also put the `get_key` logic under test here

cdkini: I agree with some of the critiques around ergonomics but I think this adheres to the current pattern pretty consistently - let's discuss

billdirks (approved):
LGTM!

PR #9661, opened 2024-03-25
[MAINTENANCE] Delete `StoreValidationResultAction`

cdkini:


great_expectations/validation_operators/validation_operators.py
```diff
@@ -109,10 +108,6 @@ class (or its descendants). See the actions included in Great Expectations and h
         # you can remove or add actions to this list. See the details in the actions
         # reference
         action_list:
-          - name: store_validation_result
```
cdkini, line 112: This entire module can be deleted once the legacy checkpoint is gone.

contrib/experimental/great_expectations_experimental/tests/rule_based_profiler/data_assistant/test_growth_numeric_data_assistant.py
```diff
@@ -364,12 +364,6 @@ def test_pandas_happy_path_growth_numeric_data_assistant(empty_data_context) ->
             }
         ],
         "action_list": [
-            {
-                "name": "store_validation_result",
-                "action": {
-                    "class_name": "StoreValidationResultAction",
-                },
-            },
```
cdkini, line 372: By removing this action, the behavior of the v0.18 checkpoint fundamentally changes. We no longer persist validations with a store and therefore need to update any tests that rely on stored values or validation result CRUD.

great_expectations/checkpoint/util.py
```diff
@@ -410,9 +410,6 @@ def validate_validation_dict(validation_dict: dict, batch_request_required: bool
     if not validation_dict.get("expectation_suite_name"):
         raise gx_exceptions.CheckpointError("validation expectation_suite_name must be specified")
 
-    if not validation_dict.get("action_list"):
-        raise gx_exceptions.CheckpointError("validation action_list cannot be empty")
-
```
cdkini, line 415: Removing this frees up a lot of errors within the test suite.

Question for V1 - does this hold? Is a checkpoint without an action valid? I say yes you don't need to configure one.

contrib/experimental/great_expectations_experimental/tests/rule_based_profiler/data_assistant/test_growth_numeric_data_assistant.py
billdirks: I thought we were removing this action because we now always persist validations.

great_expectations/checkpoint/util.py
billdirks: Yes, we don't need to configure actions. Checkpoints would still have value in group validations and running them. These results would be persisted without an action.

tests/checkpoint/test_checkpoint.py
```diff
@@ -2316,19 +2228,6 @@ def fake_cloud_context_with_slack(_fake_cloud_context_setup, monkeypatch):
     yield context, slack_counter
 
 
-@pytest.mark.cloud
-def test_use_validation_url_from_cloud(fake_cloud_context_basic):
```
billdirks, line 2320: Even if we aren't using this action, we are still saving the data to cloud. This test seems to be asserting we can construct the URL we save to correctly. Do we need a test to replace this?

tests/checkpoint/test_checkpoint_with_fluent_datasources.py
```diff
@@ -535,7 +477,6 @@ def test_newstyle_checkpoint_instantiates_and_produces_a_validation_result_when_
         ]
     )
 
-    assert len(context.validations_store.list_keys()) == 4
```
billdirks, line 538: Are these `list_keys` asserts testing to make sure that we have stored the appropriate number of results? Even though we aren't using the action, do we still want to assert somehow that the results are being stored?

tests/checkpoint/test_checkpoint_with_fluent_datasources.py
```diff
@@ -815,40 +748,6 @@ def test_newstyle_checkpoint_instantiates_and_produces_a_validation_result_when_
         validations=[{"batch_request": batch_request}],
     )
 
-    assert len(context.validations_store.list_keys()) == 1
-    assert result["success"]
-
-
-@pytest.mark.spark
-def test_newstyle_checkpoint_instantiates_and_produces_a_validation_result_when_run_validations_batch_request_dataframe_in_context_run_checkpoint_sparkdf(  # noqa: E501
```
billdirks, line 823: Why delete this whole test?

tests/data_context/test_configuration_storage.py
```diff
@@ -71,142 +71,3 @@ def data_context_parameterized_expectation_suite_with_usage_statistics_enabled(
         ),
     )
     return gx.get_context(context_root_dir=context_path)
-
-
-def test_preserve_comments_in_yml_after_adding_datasource(
-    data_context_parameterized_expectation_suite_with_usage_statistics_enabled,
-):
-    # Skipping this test for now, because the order of the contents of the returned CommentedMap is inconsistent.  # noqa: E501
-    pytest.skip("KNOWN ISSUE")
```
billdirks, line 80: This seems unrelated to removing actions. Is that right? Are we deleting this because it is always skipped now? This is reasonable, just want to make sure I understand.

tests/datasource/fluent/integration/integration_test_utils.py
```diff
@@ -32,131 +26,6 @@
 logger = logging.getLogger(__name__)
 
 
-def run_checkpoint_and_data_doc(
```
billdirks, line 35: I think this function is used by a lot of fluent tests to verify checkpoints are running successfully. I'm surprised it still works because it uses the old validator workflow, but maybe we haven't ripped that out yet.

This tests a lot more than saving validations. For example, it tests datadocs are being generated.

I'm wondering if we can do something more precise than completely deleting this.

billdirks (commented):
Thanks for doing this!

Most of my comments revolve around:
1. verifying that we know automatically save the validation results and if we can/should assert that the validations are still being saved
2. making sure we still have integration test coverage datadocs for fluent datasources

contrib/experimental/great_expectations_experimental/tests/rule_based_profiler/data_assistant/test_growth_numeric_data_assistant.py
cdkini: Yup we're okay on that front but in the intermediate state we have where we need v0.18 checkpoints to pass, we need to make sure we're removing any asserts around persistence.

cdkini: The actual rationale behind removing this action is exactly as you've stated

great_expectations/checkpoint/util.py
cdkini: I believe you still need to pass an empty list in the V1 checkpoint. Should we allow users to omit that? I'd think yes (i.e. `Field(default_factory=list)`)

tests/checkpoint/test_checkpoint.py
cdkini: I think we have decent coverage over this in our new V1 checkpoint tests but I'll double check and write tests as needed.

tests/checkpoint/test_checkpoint_with_fluent_datasources.py
cdkini: Yup. These are legacy checkpoints so the removal of the action means they aren't persisting anything at all.

The behavior of validation definitions writing directly to stores is unique to V1 checkpoints (v0.18 checkpoints haven't been rewritten to leverage this logic).

tests/checkpoint/test_checkpoint_with_fluent_datasources.py
cdkini: Honestly think it was accidental - should also bring back the `result["success"]` assertion above

tests/data_context/test_configuration_storage.py
cdkini: Yup its been skipped for years so thought I'd just remove it entirely - I can bring it back if we don't want to muddy up the diff though

tests/datasource/fluent/integration/integration_test_utils.py
cdkini: Hmm so due to the fact that I've removed actions and therefore removed the ability to persist validation results with v0.18 checkpoints, data docs don't really work within our test suite. Things are in a bit of a weird intermediate state I do agree that we should probably refrain from a large deletion like this.

great_expectations/checkpoint/util.py
billdirks: With pydantic you can also use `[]`. It will make a deep copy.

PR #9664, opened 2024-03-26
[MAINTENANCE] Use Codecov for test coverage reports

Kilo59:
https://app.codecov.io/gh/great-expectations/great_expectations


codecov.yml
```diff
@@ -0,0 +1,17 @@
+flag_management:
+  default_rules: # the rules that will be followed for any flag added, generally
+    carryforward: false
+    statuses:
+      - type: project
+        target: auto # the default target for the project
+        threshold: 60% # the default minimum threshold for the project
+      - type: patch
+        target: 90%
```
billdirks, line 9: I wonder if this is too high? We can start here and see.

billdirks (approved):
Thanks for hooking this up! When I click on the preview I don't actually see the line by line coverage. The UI seems to return the error:
```
There was a problem getting the source code from your provider. Unable to show line by line coverage.
```
For example: https://app.codecov.io/gh/great-expectations/great_expectations/blob/develop/great_expectations%2Ftypes%2Fbase.py

PR #9670, opened 2024-03-27
[MAINTENANCE] Wire up V1 Checkpoint with factory

cdkini:


great_expectations/core/batch_definition.py
```diff
@@ -52,18 +52,23 @@ def save(self) -> None:
...
-            ),
-            asset=_IdentifierBundle(
-                name=asset.name,
-                id=str(asset.id) if asset.id else None,
-            ),
-            batch_definition=_IdentifierBundle(
-                name=self.name,
-                id=self.id,
-            ),
+            datasource=data_source_bundle,
```
cdkini, line 71: Follow-up PR should rename this to `data_source`

great_expectations/core/factory/checkpoint_factory.py
```diff
@@ -85,9 +84,5 @@ def get(self, name: str) -> Checkpoint:
...
         return self._get(key=key)
 
-    def _get(self, key: GXCloudIdentifier | ConfigurationIdentifier) -> Checkpoint:
-        config: dict | CheckpointConfig = self._store.get(key=key)  # type: ignore[assignment]
-        if isinstance(config, CheckpointConfig):
-            config = config.to_json_dict()
-
-        return Checkpoint(**config, data_context=self._context)
+    def _get(self, key: GXCloudIdentifier | StringKey) -> Checkpoint:
+        return cast(Checkpoint, self._store.get(key=key))
```
cdkini: Until we can make store a generic of sorts, I think we need to cast? Or perhaps `if isinstance` and raise an error?

great_expectations/core/factory/checkpoint_factory.py
```diff
@@ -1,23 +1,22 @@
...
 from great_expectations._docs_decorators import public_api
-from great_expectations.checkpoint.checkpoint import Checkpoint
+from great_expectations.checkpoint.v1_checkpoint import Checkpoint
 from great_expectations.compatibility.typing_extensions import override
 from great_expectations.core.factory.factory import Factory
-from great_expectations.data_context.types.base import CheckpointConfig
 from great_expectations.exceptions import DataContextError
 
 if TYPE_CHECKING:
+    from great_expectations.core.data_context_key import StringKey
```
joshua-stauffer, line 12: codcov is thorough 🙈

great_expectations/core/factory/checkpoint_factory.py
joshua-stauffer: just to double check, are you certain the store will return an instantiated Checkpoint in every situation? I've run into a number of cases where a dict is returned instead (but not in the checkpoint store)

joshua-stauffer: given the state of stores and typing, maybe an isinstance and raise is a cleaner solution

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -325,8 +326,14 @@ def _init_factories(self) -> None:
 
         self._checkpoints: CheckpointFactory | None = None
         if checkpoint_store := self.stores.get(self.checkpoint_store_name):
+            # NOTE: Currently in an intermediate state where both the legacy and V1 stores exist.
+            #       Upon the deletion of the old checkpoint store, the new one will be promoted.
+            v1_checkpoint_store = V1CheckpointStore()
+            v1_checkpoint_store._store_backend = (
+                checkpoint_store.store_backend
+            )  # Leverage same backend as what was configured for the old store
```
joshua-stauffer, line 334: this is probably the best/only way to do this right now, but just noting that this is an example of how our design is broken.

joshua-stauffer (changes requested):
i think your suggestion to check isinstance/raise is probably better, so blocking on that - otherwise looks good!

great_expectations/data_context/data_context/abstract_data_context.py
cdkini: Yeah this feels gross. The lift to actually completely remove old checkpoints was far too cumbersome so we're going to be in a weird intermediate state where both exist (with the legacy ones only being around for test coverage).

great_expectations/core/factory/checkpoint_factory.py
```diff
@@ -85,9 +84,9 @@ def get(self, name: str) -> Checkpoint:
 
         return self._get(key=key)
 
-    def _get(self, key: GXCloudIdentifier | ConfigurationIdentifier) -> Checkpoint:
-        config: dict | CheckpointConfig = self._store.get(key=key)  # type: ignore[assignment]
-        if isinstance(config, CheckpointConfig):
-            config = config.to_json_dict()
+    def _get(self, key: GXCloudIdentifier | StringKey) -> Checkpoint:
+        checkpoint = self._store.get(key=key)
```
cdkini, line 88: Due to Pydantic model we're guaranteed to have `Checkpoint` but we need to do some type guarding to appease `mypy` (highlights a flaw in our store design).

joshua-stauffer: i see - well, looks like this approach makes testing more cumbersome, so I'll leave it up to you which approach you want to go with.

joshua-stauffer (approved):
thanks!

PR #9676, opened 2024-03-28
[FEATURE] `Checkpoint.save()`

cdkini:


great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -204,7 +204,15 @@ def _run_actions(
...
-        raise NotImplementedError
+        from great_expectations import project_manager
+
+        store = project_manager.get_checkpoints_store()
+        key = store.get_key(name=self.name, id=self.id)
+
+        try:
+            store.update(key=key, value=self)
+        except gx_exceptions.StoreBackendError as e:
+            raise ValueError(f"Could not find existing Checkpoint '{self.name}' to update") from e
```
joshua-stauffer: I think this error is too specific - for instance, if the cloud handler returns a 400 due to a field of an unexpected type.

great_expectations/core/factory/checkpoint_factory.py
```diff
@@ -21,9 +18,8 @@
 
 # TODO: Add analytics as needed
 class CheckpointFactory(Factory[Checkpoint]):
-    def __init__(self, store: CheckpointStore, context: AbstractDataContext):
```
joshua-stauffer, line 24: 🥳

joshua-stauffer (changes requested):
left one comment around error handling

great_expectations/checkpoint/v1_checkpoint.py
cdkini: Hmm what error do we want to catch then? The other domain objects don't do error handling here but I don't want to raise a store backend error to the user (feels like that is more internal?)

great_expectations/data_context/store/checkpoint_store.py
```diff
@@ -293,3 +293,11 @@ def _add(self, key: DataContextKey, value: V1Checkpoint, **kwargs):
...
             value.id = str(uuid.uuid4())
         return super()._add(key=key, value=value, **kwargs)
+
+    @override
+    def _update(self, key: DataContextKey, value: V1Checkpoint, **kwargs):
+        try:
+            super()._update(key=key, value=value, **kwargs)
+        except gx_exceptions.StoreBackendError as e:
+            name = key.to_tuple()[0]
+            raise ValueError(f"Could not find existing Checkpoint '{name}' to update") from e
```
joshua-stauffer: isn't this the same issue as before? There are other valid errors -- such as validation --  Cloud backend can throw, so this error message is too specific.

joshua-stauffer (changes requested):
blocking because i believe error handling is still misleading

joshua-stauffer (approved):
thanks for addressing changes!

PR #9678, opened 2024-03-29
[MAINTENANCE] `mypy 1.9` + begin wider tests type-checking

Kilo59:
This PR does a couple different type-checking related things.

1. Update `mypy` to `1.9` and removes a couple of now unused `type: ignore` comments because of better type inference.
2. Begin to type-check some more of our `tests`. Previously, the only tests being checked were `tests/datasource/fluent`. In particular, I tried to include tests under active development for the `v1` release. Also new test files should now be type-checked by default unless they are excluded.
3. Update `test_packaging` to make it easier to update pins/versions for things like `ruff`, & `mypy` and other development only packages that **SHOULD be pinned**.

I'm happy to split this PR up if desired.


tests/test_fixtures/custom_pandas_dataset.py
Kilo59, line 1: This does not appear to be used

tests/test_fixtures/custom_sparkdf_dataset.py
Kilo59, line 1: This does not appear to be used

tests/test_packaging.py
```diff
@@ -2,11 +2,14 @@
...
 import os.path
 import pathlib
-from typing import List
+from pprint import pformat as pf
+from typing import Final, List
 
 import pytest
 import requirements as rp
 
+IGNORE_PINS: Final[set[str]] = {"mypy", "ruff", "pytest"}
```
Kilo59, line 11: It doesn't make sense to check some of these development only packages. These packages SHOULD be pinned, but these checks end up punishing pinning and make it very tedious and error-prone to update the versions/pins.

I've added this `IGNORE_PINS` constant so that the packaging step just ignores anything here.

I've also had it completely ignore the `requirement-types` package since most of those types are either stubs or packages that are pulled from other requirement files which are already checked here.

pyproject.toml
```diff
@@ -128,14 +123,60 @@ exclude = [
     'validation_operators/types/validation_operator_result\.py',                                           # 35
     'validation_operators/validation_operators\.py',                                                       # 16
     # tests
+    'tests/actions/test_core_actions\.py',
+    'tests/checkpoint/test_checkpoint_result_format\.py',
```
Kilo59, line 127: All these new exclude lines were incredibly tedious to add.
I should have just written a script to do it, but I decided I was already in too deep doing it manually. 😆

tests/conftest.py
```diff
@@ -3355,7 +3351,7 @@ def empty_base_data_context_in_cloud_mode_custom_base_url(
 ) -> CloudDataContext:
     project_path = tmp_path / "empty_data_context"
     project_path.mkdir()
-    project_path = str(project_path)
+    project_path = str(project_path)  # type: ignore[assignment]
```
tyler-hoffman, line 3354: Non-blocking since this isn't actively changing anything, but changing the type of `project_path` feels bad. I think the best solution here would be just to update line 3362 to be `context_root_dir=str(project_path)`, since it looks like the string version is only used there.

tests/conftest.py
```diff
@@ -7556,21 +7552,22 @@ def _multibatch_generic_csv_generator(
             5: "category5",
             6: "category6",
         }
-        for batch_num in range(num_event_batches):
+        for batch_num in range(num_event_batches):  # type: ignore[arg-type]
```
tyler-hoffman, line 7555: I'm guessing reording the params and making the `num_` params non-optional is a larger lift? Might be cleanest to just assert instance on them at the top of the method. But that's totally non-blocking.

tyler-hoffman (approved):
Left some small comments, but non-blocking. Thanks for doing this!

tests/conftest.py
Kilo59: I agree.

I want to followup with a PR to actually fix the typing issues in `conftest.py` since it's a core file that almost every other test relies on.

I'm trying to get this one in quickly since new v1 related PR's that merge into develop will almost certainly require changes to the exclude list or the existing type-ignores.

PR #9681, opened 2024-04-01
[MAINTENANCE] Typing improvements in `test_metadatasource`

NathanFarmer:
Fix some typing errors in `test_metadatasource.py`
Add alias to `invoke fmt` so users can use `invoke format` as well
Remove unused `exclude`s in `pyproject.toml`

### DIFF
--- a/tests/datasource/fluent/test_metadatasource.py
+++ b/tests/datasource/fluent/test_metadatasource.py
@@ -10,9 +10,12 @@ from typing import TYPE_CHECKING, ClassVar, Dict, List, Optional, Tuple, Type, U
 import pytest
 
 from great_expectations.compatibility.pydantic import DirectoryPath, validate_arguments
+from great_expectations.compatibility.typing_extensions import override
+from great_expectations.core.partitioners import Partitioner
 from great_expectations.core.yaml_handler import YAMLHandler
 from great_expectations.data_context import AbstractDataContext, FileDataContext
 from great_expectations.data_context import get_context as get_gx_context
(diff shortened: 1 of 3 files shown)


tests/datasource/fluent/test_metadatasource.py
```diff
@@ -66,21 +69,12 @@ def get_context(
...
-                cls._context._add_fluent_datasource(datasource)
-                # TODO: add assets?
-
         return cls._context
 
     @validate_arguments
     def __init__(self, context_root_dir: Optional[DirectoryPath] = None) -> None:
         self.root_directory = context_root_dir
-        self._sources: _SourceFactories = _SourceFactories(self)
+        self._sources: _SourceFactories = _SourceFactories(self)  # type: ignore[arg-type]
```
Kilo59, line 77: That's a weird one

NathanFarmer: It complains that we are using this mock `DataContext` instead of `AbstractDataContext`.

anthonyburdi (approved):
Thanks for driving @NathanFarmer!

tests/datasource/fluent/test_metadatasource.py
billdirks: This is because `DataContext` is a fake and doesn't inherit from `AbstractDataContext` or implement the same interface. To fix this for real, we should change how we write this fake, eg inherit from `AbstractDataContext` or implement a standard interface.

PR #9682, opened 2024-04-01
[FEATURE] Factory iterators

tyler-hoffman:
Creates factory `all` iterators.

### Additional scope creep
I needed to fix our fake api a bit for this PR. Fixes include:
* using dynamic ids in urls for checkpoints and suites
* not nesting 2 layers of `data` in the suites endpoint
* not overwriting checkpoints when posting (by using dynamic ids)


great_expectations/core/factory/checkpoint_factory.py
```diff
@@ -84,6 +84,12 @@ def get(self, name: str) -> Checkpoint:
 
         return self._get(key=key)
 
+    @public_api
+    @override
+    def all(self) -> List[Checkpoint]:
```
Kilo59: I would strongly suggest we make the initial interface for these `all` method more generic.
`typing.Iterable[Checkpoint]` or `typing.Sequence[Checkpoint]`.

These will free us up to do things like implement some of these `.all()` methods as [lazy loading generator functions](https://realpython.com/introduction-to-python-generators/#using-generators).
```suggestion
    def all(self) -> Iterable[Checkpoint]:
```

tyler-hoffman: Glad your brought this up! My initial plan with this was make it an `Iterable` (that's what the ticket even specifies), but I went with a `List` because I figure users might want to iterate through it multiple times, etc. Agreed that `Sequence` is preferable to `List`. I'll at least change it to that, but I'm open to arguments about preferring `Iterable`.

Kilo59: The implementation doesn't need to change at all with `Iterable[Checkpoint]`, you can still return a list.
It just gives us more flexibility going forward.

Kilo59: @tyler-hoffman 
```python
from collections.abc import Iterable
from string import ascii_letters
from typing import Protocol, runtime_checkable


@runtime_checkable
class SupportsAll(Protocol):
    def all(self) -> Iterable: ...


class ListAll:
    def all(self) -> Iterable[str]:
        return list(ascii_letters)


class GeneratorAll:
    def all(self) -> Iterable[str]:
        for c in ascii_letters:
            yield c        
        # ☝️ can also be written as -> yield from ascii_letters


class StrAll:
    def all(self) -> Iterable[str]:
        return ascii_letters


l = ListAll()
g = GeneratorAll()
s = StrAll()

assert isinstance(l, SupportsAll)
assert isinstance(g, SupportsAll)
assert isinstance(s, SupportsAll)

l_all = l.all()
g_all = g.all()
s_all = s.all()

print(f"{type(l_all)} {l_all=}")
print(f"{type(g_all)} {g_all=}")
print(f"{type(s_all)} {s_all=}")

assert isinstance(l_all, Iterable)
assert isinstance(l_all, Iterable)
assert isinstance(l_all, Iterable)

assert l_all == list(g_all) == list(s_all) == list(ascii_letters)
```

great_expectations/core/factory/factory.py
```diff
@@ -22,5 +22,5 @@ def get(self, name: str) -> T:
         pass
 
     @abstractmethod
-    def all(self) -> List[T]:
+    def all(self) -> Iterable[T]:
```
Kilo59, line 25: Not blocking but have we considered making this a protocol?
It can still be inherited, work like a base class, and have abstract methods, but it doesn't HAVE to be inherited.

Kilo59 (approved):
❤️

tests/core/factory/test_checkpoint_factory.py
```diff
@@ -237,6 +237,56 @@ def _test_checkpoint_factory_delete_success(context):
...
+
+
+@pytest.mark.filesystem
+def test_checkpoint_factory_get_all_filesystem(empty_data_context):
+    _test_checkpoint_factory_get_all(empty_data_context)
+
+
+@pytest.mark.cloud
+def test_checkpoint_factory_get_all_cloud(empty_cloud_context_fluent):
+    _test_checkpoint_factory_get_all(empty_cloud_context_fluent)
```
Kilo59: `non-blocking`
You can apply mark different marks to test parameters. So you could have 1 parametrized test here.

https://docs.pytest.org/en/7.1.x/example/markers.html#marking-individual-tests-when-using-parametrize

or do it dynamically.
https://stackoverflow.com/a/73556599/6304433

tyler-hoffman: Oh good idea. Thanks!

tyler-hoffman: I'm going to be working on a couple other PRs that are prereqs for this, but feel free to throw a blocker on this PR over that if you want!

great_expectations/core/factory/suite_factory.py
```diff
@@ -94,3 +94,9 @@ def get(self, name: str) -> ExpectationSuite:
         if self._include_rendered_content:
             suite.render()
         return suite
+
+    @public_api
+    @override
+    def all(self) -> Iterable[ExpectationSuite]:
+        """Get all ExpectationSuites."""
+        return []
```
billdirks: Should this be `return self._store.get_all()`?

great_expectations/core/factory/validation_definition_factory.py
```diff
@@ -75,3 +75,9 @@ def get(self, name: str) -> ValidationDefinition:
             raise DataContextError(f"ValidationDefinition with name {name} was not found.")
 
         return cast(ValidationDefinition, self._store.get(key=key))
+
+    @public_api
+    @override
+    def all(self) -> Iterable[ValidationDefinition]:
+        """Get all ValidationDefinitions."""
+        return []
```
billdirks: Similarly, should this call into the store?

great_expectations/core/factory/suite_factory.py
tyler-hoffman: Hah sorry, this should have been marked as a draft! It certainly should be, but at the time I put this up those weren't yet implemented everywhere!

great_expectations/core/factory/factory.py
tyler-hoffman: I don't think we discussed it, and could go either way. The reason I lean ever-so-slightly toward team ABC is that requiring you to subclass pushes the error slightly earlier in the dev process. If I mess up the signature of a method, I see as I'm defining my class, rather than when I go to use it. I'm guessing there's a good counter argument.

great_expectations/core/factory/validation_definition_factory.py
```diff
@@ -75,3 +75,9 @@ def get(self, name: str) -> ValidationDefinition:
             raise DataContextError(f"ValidationDefinition with name {name} was not found.")  # noqa: TRY003
 
         return cast(ValidationDefinition, self._store.get(key=key))
+
+    @public_api
+    @override
+    def all(self) -> Iterable[ValidationDefinition]:
+        """Get all ValidationDefinitions."""
+        return self._store.get_all()
```
joshua-stauffer, line 83: does this have test coverage?

great_expectations/core/factory/suite_factory.py
```diff
@@ -94,3 +94,9 @@ def get(self, name: str) -> ExpectationSuite:
         if self._include_rendered_content:
             suite.render()
         return suite
+
+    @public_api
+    @override
+    def all(self) -> Iterable[ExpectationSuite]:
+        """Get all ExpectationSuites."""
+        return self._store.get_all()
```
joshua-stauffer: does this have test coverage?

joshua-stauffer (changes requested):
blocking for test coverage on the new methods for suite and validation definition factories

great_expectations/core/factory/suite_factory.py
tyler-hoffman: oh good call on both of these @joshua-stauffer ! After https://github.com/great-expectations/great_expectations/pull/9718 merges, I'll backfill those test cases!

tests/conftest.py
```diff
@@ -3326,6 +3326,11 @@ def cloud_api_fake(cloud_details: CloudDetails):
         yield requests_mock
 
 
+@pytest.fixture
+def ephemeral_data_context() -> EphemeralDataContext:
+    return get_context(mode="ephemeral")
```
cdkini: I think have a `in_memory_runtime_context` fixture that does the same thing

tests/core/factory/test_checkpoint_factory.py
```diff
@@ -221,6 +222,51 @@ def _test_checkpoint_factory_delete_success(context):
         context.checkpoints.get(name)
 
 
+@pytest.mark.parametrize(
+    "context_fixture_name",
+    [
+        pytest.param("empty_cloud_context_fluent", id="cloud", marks=pytest.mark.cloud),
+        pytest.param("ephemeral_data_context", id="ephemeral", marks=pytest.mark.unit),
```
cdkini: Is this a unit test? Maybe `big`?

tyler-hoffman: good call! will update

tyler-hoffman: fwiw I'm updating the corresponding test in each of these files. It's not technically a unit test, but I think it's also a pretty quick test, so I don't have strong opinions on the distinction 🤷

tests/datasource/fluent/_fake_cloud_api.py
```diff
@@ -470,9 +468,9 @@ def get_expectation_suites_cb(request: PreparedRequest) -> CallbackResult:
     queried_names: Sequence[str] = query_params.get("name", [])  # type: ignore[assignment]
 
     exp_suites: dict[str, dict] = _CLOUD_API_FAKE_DB["expectation_suites"]
-    exp_suite_list: list[dict] = list(exp_suites.values())
+    exp_suite_list: list[dict] = [d["data"] for d in exp_suites.values()]
```
tyler-hoffman, line 471: we shouldn't be putting `data` in our fake db, but I don't want to do the more holistic update now. This change just prevents us from getting a `data` array with nested `data` objects.

joshua-stauffer (approved):
thanks for adding the tests!

PR #9689, opened 2024-04-02
[MAINTENANCE] Migrate misc actions to V1 pattern

cdkini:
`UpdateDataDocsAction.v1_run`, `PagerdutyAlertAction.v1_run`, `SNSNotificationAction.v1_run` `APINotificationAction.v1_run` implementations that support an input checkpoint result object


great_expectations/checkpoint/actions.py
```diff
@@ -410,6 +412,18 @@ def _root_validate_deps(cls, values: dict) -> dict:
 
         return values
 
+    @override
+    def v1_run(self, checkpoint_result: CheckpointResult) -> dict:  # type: ignore[override] # signature does not match parent
```
billdirks: Do you need this type ignore?

great_expectations/checkpoint/actions.py
```diff
@@ -439,32 +450,42 @@ def _run(  # type: ignore[override] # signature does not match parent  # noqa: P
...
+            "expectation_suite_name", "__no_expectation_suite_name__"
+        )
+
+        return self._run_pypd_alert(
+            name=expectation_suite_name,
+            template="Great Expectations suite check {name} has failed",
+            success=validation_success,
+        )
+
+    def _run_pypd_alert(self, name: str, template: str, success: bool):
```
billdirks: Can you change the name parameter to suite_name or something more explicit?

great_expectations/checkpoint/actions.py
```diff
@@ -439,32 +450,42 @@ def _run(  # type: ignore[override] # signature does not match parent  # noqa: P
             )
 
         validation_success = validation_result_suite.success
+        expectation_suite_name = validation_result_suite.meta.get(
+            "expectation_suite_name", "__no_expectation_suite_name__"
+        )
+
+        return self._run_pypd_alert(
+            name=expectation_suite_name,
+            template="Great Expectations suite check {name} has failed",
```
billdirks: I would remove the template parameter and hard code it in the method. This is a private method and we never override so I don't know what parameterizing it gives us.

great_expectations/checkpoint/actions.py
```diff
@@ -929,6 +950,29 @@ class UpdateDataDocsAction(DataDocsAction):
 
     site_names: List[str] = []
 
+    @override
+    def v1_run(self, checkpoint_result: CheckpointResult) -> dict[ValidationResultIdentifier, dict]:  # type: ignore[override] # signature does not match parent
```
billdirks: Similar, I don't think you need the ignore here.

great_expectations/checkpoint/actions.py
```diff
@@ -1045,6 +1101,24 @@ class APINotificationAction(ValidationAction):
...
+
+            payload = self.create_payload(
+                data_asset_name=batch_identifier,
+                suite_name=suite_name,
+                validation_results_serializable=serializable_results,
+            )
+            aggregate_payload.append(payload)
+
+        response = self.send_results(aggregate_payload)
+        return f"Successfully Posted results to API, status code - {response.status_code}"
```
billdirks: I would remove `Successfully` here since the status code may indicate a failure.

tests/actions/test_core_actions.py
```diff
@@ -865,13 +913,71 @@ def test_OpsgenieAlertAction_run(self, checkpoint_result: CheckpointResult):
...
+                            "source": "Great Expectations",
+                            "summary": f"Great Expectations Checkpoint {checkpoint_name} has failed",  # noqa: E501
+                        },
+                        "routing_key": "test",
+                    }
+                ),
+            ]
+        )
+
+    @pytest.mark.skipif(
```
billdirks, line 945: The codecov comments made me think this test might not be running so I looked into it and it is being skipped.

In general, we don't want to use these `skipif` markers anymore. Using the correct `pytest.mark.*` should ensure the correct dependencies are installed since each test is only run in 1 job dictated by that marker.

It looks like the `pytest.mark.unit` marker loads in `reqs/requirements-dev-test.txt` (see [ci.yml](https://github.com/great-expectations/great_expectations/blob/develop/.github/workflows/ci.yml#L194)). However, `pypd` is found in `reqs/requirements-dev-pagerduty.txt` which is only pulled in via the top level `requirements-dev.txt`.

I don't think any of our test markers currently pull in `reqs/requirements-dev-pagerduty.txt`. We will need to update a marker, or add a new one, to get this to run.

great_expectations/checkpoint/actions.py
```diff
@@ -439,32 +450,42 @@ def _run(  # type: ignore[override] # signature does not match parent  # noqa: P
...
+        )
+
+        return self._run_pypd_alert(
+            name=expectation_suite_name,
+            template="Great Expectations suite check {name} has failed",
+            success=validation_success,
+        )
+
+    def _run_pypd_alert(self, name: str, template: str, success: bool):
+        import pypd
```
billdirks: Move the `pypd` import to the top level and wrap in `NotImported` so we get a good error message if they try to run this without `pypd` installed.

billdirks (commented):
Thanks for doing this. The overall structure looks good. I've left some comments inline.

great_expectations/checkpoint/actions.py
cdkini: This is to maintain existing behavior until we rip out single validation result support. Maybe not the cleanest approach but it does let us leverage a lot of the same code for both paths

great_expectations/checkpoint/actions.py
cdkini: Well its either a suite name or a checkpoint name (depending on what code path we're serving here). I'd like to keep this as `name` as that is the name of the template param?

This will all go once we remove legacy checkpoints

great_expectations/checkpoint/actions.py
```diff
@@ -161,7 +163,7 @@ class (for open source Great Expectations) or the GeCloudIdentifier (from Great
         """  # noqa: E501
 
     # NOTE: To be promoted to 'run' after V1 development (JIRA: V1-271)
-    def v1_run(self, checkpoint_result: CheckpointResult) -> None:
+    def v1_run(self, checkpoint_result: CheckpointResult) -> str | dict:
```
billdirks, line 164: Not directly related to this PR but I noticed the `_run` method is annotated as public which is strange since it's a `_` method. Could you remove that either in this or a followup PR?

great_expectations/checkpoint/actions.py
billdirks: `_run_pypd_alert` is a new method you've added. It seems strange to pass in `name` and then have that used in a `template`. It begs the question, why not pass in the actual message instead of a template. I see `name` is used as the `dedup_key` as well. I'd change the signature of this new method so `name` -> `dedup_key` or just `key` and change template to `summary`.

great_expectations/checkpoint/actions.py
billdirks: See comment above.

tests/actions/test_core_actions.py
```diff
@@ -865,13 +913,67 @@ def test_OpsgenieAlertAction_run(self, checkpoint_result: CheckpointResult):
         not is_library_loadable(library_name="pypd"),
         reason="pypd is not installed",
     )
+    @mock.patch("pypd.EventV2.create")
```
billdirks: Can you remove `@pytest.mark.skipif` on this test?

billdirks (commented):
Thanks for updating. I've left a few very minor comments. I did notice there are some sql compatibility tests dying. I haven't look into them and I'm wondering if it's related to this PR?

great_expectations/checkpoint/actions.py
```diff
@@ -400,6 +401,18 @@ class PagerdutyAlertAction(ValidationAction):
     notify_on: Literal["all", "failure", "success"] = "failure"
     severity: Literal["critical", "error", "warning", "info"] = "critical"
 
+    @override
+    def v1_run(self, checkpoint_result: CheckpointResult) -> dict:
+        success = checkpoint_result.success or False
+        checkpoint_name = checkpoint_result.checkpoint_config.name
+        summary = f"Great Expectations Checkpoint {checkpoint_name} has "
+        if success:
```
tyler-hoffman, line 409: nit: using an f-string rather than reassigning the variable is probably simpler

tests/actions/test_core_actions.py
```diff
@@ -764,29 +789,81 @@ def test_action_deserialization(
...
+                    batch_identifier=self.batch_id_a,
+                ): ExpectationSuiteValidationResult(
+                    success=True,
+                    statistics={},
+                    results=[],
+                    suite_name=self.suite_a,
                 ),
-                mocker.Mock(spec=ValidationResultIdentifier): mocker.Mock(
-                    spec=ExpectationSuiteValidationResult, success=False
+                ValidationResultIdentifier(
```
tyler-hoffman, line 827: 👍

tests/actions/test_core_actions.py
```diff
@@ -835,17 +969,149 @@ def test_SlackNotificationAction_run(self, checkpoint_result: CheckpointResult):
...
+                            resource_type=GXCloudRESTResource.EXPECTATION_SUITE,
+                            resource_name=self.suite_b,
+                        ),
+                    ],
+                    site_names=site_names,
+                ),
+            ]
+        )
+        assert res == {
+            validation_identifier_a: {},
```
tyler-hoffman, line 1115: I'm assuming this is functionally equivalent to before, but these empty object ids seem pretty darn worthless, right?

tests/actions/test_core_actions.py
```diff
@@ -835,17 +969,149 @@ def test_SlackNotificationAction_run(self, checkpoint_result: CheckpointResult):
...
+                site_names[0]: site_urls[0],
+                site_names[1]: site_urls[1],
+            },
+            validation_identifier_b: {
+                site_names[0]: site_urls[0],
+                site_names[1]: site_urls[1],
+            },
+        }
+
+    @pytest.mark.cloud
```
tyler-hoffman, line 1051: These 2 tests are a bit to take in. Anything we can do to make it a bit simpler? Maybe just test fixtures/constants for the shared data? (If that doesn't help with readability, ignore me!)

tyler-hoffman (approved):
Left a thought on test structure, but otherwise things look good to me

PR #9703, opened 2024-04-03
[FEATURE] TupleGCSStoreBackend::get_all

tyler-hoffman:
Implements `TupleGCSStoreBackend::get_all`. Note that we could improve the performance if/when we bump google-cloud-storage to 2.12, using the `transfer_manager`.

The test coverage here uses fakes to mimic what GCS does, inferred from the preexisting code, but I also did a bit of manual testing. The steps if you want to repro:
* get set up with GCS (we have it in a sandbox)
* add a bucket
* follow google's docs on getting authorization working
* follow our docs to get a validation_result store hosted in that bucket
* run a checkpoint
* run the following code and make sure things look reasonable:
```
store = context.stores["validations_store"].store_backend
all_of_them = store.get_all() # check that this looks good.

# get an individual validation result to make sure it looks like the first in the above list if you want to check
keys = store.list_keys()
first = store.get(keys[0])
```

### DIFF
--- a/great_expectations/data_context/store/tuple_store_backend.py
+++ b/great_expectations/data_context/store/tuple_store_backend.py
@@ -828,12 +828,26 @@ class TupleGCSStoreBackend(TupleStoreBackend):
         return gcs_object_key
 
     def _get(self, key):
-        gcs_object_key = self._build_gcs_object_key(key)
-
         from great_expectations.compatibility import google
 
         gcs = google.storage.Client(project=self.project)
         bucket = gcs.bucket(self.bucket)
+        return self._get_by_gcs_object_key(bucket, key)
+
+    @override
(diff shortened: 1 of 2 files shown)


joshua-stauffer (approved):
given the current pattern, this looks good 👍 I would prefer to see the google client injected as a dependency to the store backend so we could just pass in a mock and assert against it, but definitely out of scope for this PR.

tyler-hoffman:
> given the current pattern, this looks good 👍 I would prefer to see the google client injected as a dependency to the store backend so we could just pass in a mock and assert against it, but definitely out of scope for this PR.

@joshua-stauffer yeah, totally agree. Unfortunately, I think we're going to be in the same place with azure blob stores

PR #9704, opened 2024-04-04
[FEATURE] Refactor FilePathDataConnector

joshua-stauffer:
Previously, the FilePathDataConnector required a `batching_regex` on instantiation, which it uses throughout its lifecycle. The goal here is to rework the code to support passing `batching_regex` as a parameter when transforming a `BatchRequest` into `BatchDefinitions`.

This PR should be a functionally equivalent refactor; the class's `batching_regex` attribute is still the source of truth.

### changes
- `FilePathDataConnector._data_references_cache` changed types from `Dict[str, List[LegacyBatchDefinition] | None]` to `dict[re.Pattern, dict[str, list[LegacyBatchDefinition] | None]]` to allow caching results from multiple `batching_regex`.

### DIFF
--- a/great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
+++ b/great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
@@ -316,7 +316,7 @@ batch identifiers {batch_definition.batch_identifiers} from batch definition {ba
         return {FilePathDataConnector.FILE_PATH_BATCH_SPEC_KEY: path}
 
     def _preprocess_batching_regex(self, regex: re.Pattern) -> re.Pattern:
-        """Add the FILE_PATH_BATCH_SPEC_KEY group to regex in case it is not already present."""
+        """Add the FILE_PATH_BATCH_SPEC_KEY group to regex if not already present."""
         regex_parser = RegExParser(
             regex_pattern=regex,
             unnamed_regex_group_prefix=self._unnamed_regex_group_prefix,
--- a/great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
+++ b/great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
@@ -126,9 +126,7 @@ class FilePathDataConnector(DataConnector):
         )
 
         self._unnamed_regex_group_prefix: str = unnamed_regex_group_prefix
-        self._batching_regex: re.Pattern = self._ensure_regex_groups_include_data_reference_key(
-            regex=batching_regex
-        )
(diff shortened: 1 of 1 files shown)


great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
```diff
@@ -317,15 +315,8 @@ def _generate_batch_spec_parameters_from_batch_definition(
...
-        """
-        Args:
-            regex: regex pattern for filtering data references; if reserved group name "path" (FILE_PATH_BATCH_SPEC_KEY)
-            is absent, then it is added to enclose original regex pattern, supplied on input.
-
-        Returns:
-            Potentially modified Regular Expression pattern (with enclosing FILE_PATH_BATCH_SPEC_KEY reserved group)
-        """  # noqa: E501
+    def _preprocess_batching_regex(self, regex: re.Pattern) -> re.Pattern:
+        """Add the FILE_PATH_BATCH_SPEC_KEY group to regex in case it is not already present."""
```
tyler-hoffman: `s/in case/if/`

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
```diff
@@ -339,74 +330,73 @@ def _ensure_regex_groups_include_data_reference_key(self, regex: re.Pattern) ->
...
+        regex_parser = RegExParser(
+            regex_pattern=batching_regex,
+            unnamed_regex_group_prefix=self._unnamed_regex_group_prefix,
+        )
+        for data_reference in self.get_data_references():
+            batch_definition = self._build_batch_definition(
+                data_reference=data_reference, regex_parser=regex_parser
+            )
+            if batch_definition:
+                # storing these as a list seems unnecessary; in this implementation
```
tyler-hoffman, line 351: Yeah, unnecessary and error-prone. Looks like the change to just make this None or a single instance would be low effort. Maybe as a follow up PR?

tyler-hoffman: Also out of scope here, but why are we mutating `batch_definitions` if this is just a get method?

tyler-hoffman (approved):
Looks correct to me! I like the idea of replacing those lists of one item with just an optional legacy batch def if you/we have time, but might as well make that its own PR.

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
joshua-stauffer: this method populates the cache on get, so shouldn't be mutating after the first calculation

Issue #6701, opened 2023-01-05
Portal Page Failure to Load (GE 0.15.41)


diveyseek:
**Describe the bug**
After upgrading to 0.15.41 from 0.13.45 the GE Portal Page falls into an endless loop
This appears to related to a change in the code below

**To Reproduce**
With the latest version of GE (0.15.41), and the following configuration, the data doc is generating a dark version.
```
    class_name: SiteBuilder
    show_how_to_buttons: false
    store_backend:
      class_name: TupleS3StoreBackend
      bucket: ${DATA_DOCS_S3_BUCKET}  # UPDATE the bucket name here to match the bucket you configured above.
    site_index_builder:
      class_name: DefaultSiteIndexBuilder
      show_cta_footer: false
      validation_results_limit: 1600
```

With 1600 validation results the newer dark version page just keeps loading (the page hangs and I can’t click anywhere)  while the older white version used to load within seconds (we were using GE version 0.13.45 before). 

If I replace the following scripts in the index page generated by v 0.15.41 with older versions of these script, the page loads within seconds.

```
    <link rel="stylesheet" href="https://unpkg.com/bootstrap-table@1.19.1/dist/bootstrap-table.min.css">
    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css"/>

    <link rel="stylesheet" type="text/css" href="https://unpkg.com/bootstrap-table@1.19.0/dist/extensions/filter-control/bootstrap-table-filter-control.min.css">
    <link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-datepicker/1.9.0/css/bootstrap-datepicker.min.css">
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@forevolve/bootstrap-dark@1.1.0/dist/css/bootstrap-prefers-dark.css" />

<script src="https://unpkg.com/bootstrap-table@1.19.1/dist/bootstrap-table.min.js"></script>
<script src="https://unpkg.com/bootstrap-table@1.19.1/dist/extensions/filter-control/bootstrap-table-filter-control.min.js"></script>
```

I replaced them with the following script link to make the page work without an issue

```
  <link rel="stylesheet" href="https://unpkg.com/bootstrap-table@1.16.0/dist/bootstrap-table.min.css">
[...]

PR #9705, opened 2024-04-04
[MAINTENANCE] Speed up index.html by increasing bootstrap-table version

SebastianFrankS17:
This PR updates the "bootstrap-table" dependency of the index.html page to speed up loading times in the browser.

To provide some numbers:
- loading index.html with ~450 validations/table rows takes >60s until the page is useable, i.e. I can interact with and filter the table [all required contents previously downloaded from S3 Store to local machine to rule out any infrastructural problems]

I could verify that the call to `bootstrapTable` causes the undesirable delay by either...
- reducing the html to a single entry in the table manually OR
- leaving the table as is, but switching to a different version of the loaded "bootstrap-table" script

... which in both cases reduced the loading times significantly [~1s]

There exists a closed issue in the GX repo from last year, which didn't get resolved:
https://github.com/great-expectations/great_expectations/issues/6701

- The increased loading times seem to be existing only in specific "bootstrap-table" versions, i.e. 1.19.x
- Versions of the filter-control >1.20.2 did not turn out to work properly when testing
- I was just not sure if picking matching versions for the base integration and the filtering extension is a better idea, instead of using the latest working versions for both


SebastianFrankS17:
closed due to paused contributions as part of the preparation vor OSS 1.0 release. let me know if I should re-open.

PR #9706, opened 2024-04-04
[MAINTENANCE] Refactor FilePathDataConnector

joshua-stauffer:
Some key logic belonging to the `FilePathDataConnector` was previously declared as top level functions, then injected into the class on init, in order to allow callers to choose the functionality they need. This PR refactors that logic to be internal to the DataConnector, and instead exposes a flag to allow callers to select the functionality they require.

This PR should be a functionally equivalent refactor.

## the details
The default behavior of `FilePathDataConnector` is to use a `batching_regex` to generate a `LegacyBatchDefinition` for each data reference found. Some Assets -- specifically those inheriting from `_DirectoryDataAssetMixin` -- want to override this behavior, and instead receive a single `LegacyBatchDefinition` that treats the whole directory as a single Asset. The solution in this PR is to introduce a new parameter to the DataConnector init logic called `whole_directory_path_override: PathStr | None`. Callers who need a single asset can provide the directory as an override flag, and callers who want the standard functionality can pass `None`.

## why bother?
This PR is part of work which refactors the FilePathDataConnector to support passing batching_regex as a parameter when transforming a BatchRequest into BatchDefinitions. At minimum, the functions refactored in this PR will require an updated signature, and it may make sense to reorganize the code within the class. It's easier and safer to do that work when the logic governing to the DataConnector is fully encapsulated within it. Also, the callers of FilePathDataConnector don't actually want to define their own custom logic, so allowing them to pass in a task-critical function breaks the class abstraction unnecessarily.

## previous work
- https://github.com/great-expectations/great_expectations/pull/9704

### DIFF
--- a/great_expectations/datasource/fluent/data_asset/data_connector/dbfs_data_connector.py
+++ b/great_expectations/datasource/fluent/data_asset/data_connector/dbfs_data_connector.py
@@ -9,14 +9,9 @@ from great_expectations.compatibility.typing_extensions import override
 from great_expectations.datasource.fluent.data_asset.data_connector import (
     FilesystemDataConnector,
 )
-from great_expectations.datasource.fluent.data_asset.data_connector.file_path_data_connector import (  # noqa: E501
-    FilePathDataConnector,
-    file_get_unfiltered_batch_definition_list_fn,
-)
 
 if TYPE_CHECKING:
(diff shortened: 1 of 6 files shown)


anthonyburdi (approved):
LGTM! Thank you for laying out the rationale. Never felt great about that implementation.

PR #9716, opened 2024-04-05
[MAINTENANCE] Migrate `OpsgenieNotificationAction`

cdkini:
Opsgenie renderer and action need to support V1's checkpoint result API


great_expectations/checkpoint/types/checkpoint_result.py
```diff
@@ -288,10 +288,10 @@ def _list_data_assets_validated_by_batch_id(self) -> dict:
                         "expectation_suite_names": [expectation_suite_name],
                     }
                 else:
-                    assets_validated_by_batch_id[batch_id]["validation_results"].append(
+                    assets_validated_by_batch_id[batch_id]["validation_results"].append(  # type: ignore[union-attr]
                         validation_result
                     )
-                    assets_validated_by_batch_id[batch_id]["expectation_suite_names"].append(
+                    assets_validated_by_batch_id[batch_id]["expectation_suite_names"].append(  # type: ignore[union-attr]
```
cdkini, line 294: This is the old checkpoint result (not sure why mypy is mad about it now). We're using a new Pydantic type so I think we can ignore these?

great_expectations/core/expectation_validation_result.py
```diff
@@ -514,7 +514,7 @@ def __init__(  # noqa: PLR0913
         suite_name: str,
         evaluation_parameters: Optional[dict] = None,
         statistics: Optional[dict] = None,
-        meta: Optional[ExpectationSuiteValidationResult | dict] = None,
+        meta: Optional[ExpectationSuiteValidationResultMeta | dict] = None,
```
cdkini, line 517: This has been wrong for a while (thank you @tyler-hoffman for pointing it out). Some minor changes in this PR to adhere to the proper type

great_expectations/core/expectation_validation_result.py
```diff
@@ -532,6 +532,12 @@ def __init__(  # noqa: PLR0913
         self.id = id
         self._metrics: dict = {}
 
+    @property
+    def asset_name(self) -> str | None:
+        if "active_batch_definition" in self.meta:
+            return self.meta["active_batch_definition"].data_asset_name
+        return None
```
cdkini, line 539: Would like to make this a top-level property (might write a ticket around it) but not pressing

great_expectations/checkpoint/actions.py
```diff
@@ -606,6 +606,41 @@ def _validate_renderer(cls, renderer: dict | OpsgenieRenderer) -> OpsgenieRender
             renderer = _renderer
         return renderer
 
+    @override
+    def v1_run(self, checkpoint_result: CheckpointResult) -> dict:
+        validation_success = checkpoint_result.success
+        checkpoint_name = checkpoint_result.checkpoint_config.name
+
+        if (
+            self.notify_on == "all"
```
tyler-hoffman: I was thinking about this in the other PR but didn't mention it because it was just reusing existing code: maybe it's just me but I find these longer lines of `or` / `and` hard to parse in python. What about making intermediate variables for the `and` parts, and then just having this condition do the `or`s if that makes sense.

great_expectations/checkpoint/actions.py
```diff
@@ -606,6 +606,41 @@ def _validate_renderer(cls, renderer: dict | OpsgenieRenderer) -> OpsgenieRender
             renderer = _renderer
         return renderer
 
+    @override
+    def v1_run(self, checkpoint_result: CheckpointResult) -> dict:
```
tyler-hoffman, line 621: Lots of overlap with `_run`. I might DRY this up, but if we can remove `_run` in the near term, I'm less stressed about it.

great_expectations/checkpoint/actions.py
```diff
@@ -1044,19 +1080,8 @@ def _run(  # type: ignore[override] # signature does not match parent
         checkpoint_identifier=None,
         **kwargs,
     ):
-        suite_name: str = validation_result_suite.meta["expectation_suite_name"]
```
tyler-hoffman, line 1119: I'm confused on this change. This is a separate action, right?

great_expectations/render/renderer/opsgenie_renderer.py
```diff
@@ -1,14 +1,63 @@
...
+        text_blocks: list[str] = []
+        for run_result in checkpoint_result.run_results.values():
+            text_block = self._v1_render(run_result=run_result)
+            text_blocks.append(text_block)
+
+        return self._concatenate_text_blocks(
+            checkpoint_result=checkpoint_result, text_blocks=text_blocks
+        )
+
+    def _v1_render(self, run_result: ExpectationSuiteValidationResult) -> str:
```
tyler-hoffman: This should be named to indicate it's just about a single result; first look at this method and the above made me question why one was public and one was private, and why they would be different.

tyler-hoffman (approved):
LGTM pending that one private method rename

great_expectations/checkpoint/actions.py
cdkini: Legacy `_run` will be removed shortly!

great_expectations/checkpoint/actions.py
cdkini: Yup but `mypy` kept flagging it as an issue

cdkini: Probably because of the change to `meta` below

PR #9718, opened 2024-04-05
[FEATURE] Update expectations and checkpoints v1 stores to implement gx_cloud_response_json_to_object_collection

tyler-hoffman:
## Goals
* Unlock the ability for our existing factories to implement `all`, which depends on their stores' `get_all`, which in turn relies on the stores' `gx_cloud_response_json_to_object_collection`
* Somewhat formalize a pattern that we can adopt more widely about how to do this

## Implementation
* Move `gx_cloud_response_json_to_object_collection` into the parent store
* Put a not implemented `_convert_raw_json_to_object_dict` in the parent store
* Override `_convert_raw_json_to_object_dict` in the 2 new stores we care about

## What this does not do
* We have a factory for `ValidationDefinition`, but we cannot yet implement `gx_cloud_response_json_to_object_collection` for that store because we don't yet have a REST endpoint
* This also leaves most stores in a state where they don't have the functionality implemented. They will get a `NotImplemented` error when calling `get_all`, but that behavior is the same as before (except now it raises from one step deeper in the call stack, in `_convert_raw_json_to_object_dict`
* Totally fix our inconsistencies. We should eventually move this logic into the cloud store backend, but we aren't there today.

### DIFF
--- a/great_expectations/data_context/store/checkpoint_store.py
+++ b/great_expectations/data_context/store/checkpoint_store.py
@@ -247,8 +247,8 @@ class V1CheckpointStore(Store):
         return self._key_class(key=name)
 
     @override
-    @staticmethod
-    def gx_cloud_response_json_to_object_dict(response_json: dict) -> dict:
+    @classmethod
+    def gx_cloud_response_json_to_object_dict(cls, response_json: dict) -> dict:
         response_data = response_json["data"]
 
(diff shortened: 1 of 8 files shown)


joshua-stauffer (approved):
looks good, thanks!

PR #9720, opened 2024-04-08
[FEATURE] Batch definition sorting

tyler-hoffman:
## Overview
Adds `sort_batches_ascending` property to `Partitioner`s, including both the (soon to be removed) generic ones, as well as the asset-specific ones.

Most of the changes here are around schemas because of ^.

## Important things to look for
* `DataAsset::sort_batches` has been updated to just look at the partitioner tuples. I might circle back on this to make the logic a bit simpler, but for now left it with minimal changes.
* I moved / updated some tests to separate test files specific to sql and file paths around `get_batch_list_from_batch_request`

## What this does not do
This PR does not touch `DataAsset::add_sorters` or `DataAsset::order_by` that it sets. I plan to remove that next, but this PR is already huge enough.


great_expectations/core/partitioners.py
```diff
@@ -7,18 +7,21 @@
 @public_api
 class PartitionerYear(pydantic.BaseModel):
     column_name: str
+    sort_batches_ascending: bool = True
```
tyler-hoffman: Considering DRYing this up with a layer of inheritance. Reviewer: WDYT?

great_expectations/datasource/fluent/file_path_data_asset.py
```diff
@@ -334,9 +335,8 @@ def _get_batch_definition_list(
                 if batch_parameters_counts[param_name] == 1:
                     batch_request_copy_without_partitioner_kwargs.options.pop(param_name)
                 else:
-                    warnings.warn(
-                        f"The same option name is applied for your batch regex and partitioner config: {param_name}"  # noqa: E501
-                    )
+                    # TODO: figure out what to do here!
```
tyler-hoffman, line 338: This checking was a result of issues between sorters for batching regex and partitioners. This issue should be resolved in a followup PR where I remove `add_sorters`

great_expectations/core/partitioners.py
billdirks: I'm ok either way. I guess, slight preference to leaving it as is for now and extracting a parent class if we feel like this is a burden. Maybe change the name to `sort_ascending`.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -95,6 +96,13 @@
     )
 
 
+class _PartitionerProtocol(Protocol):
```
billdirks: There is a `_Partitioner` protocol in both `sql_datasource.py` and `spark_generic_partitioners.py` which are almost identical, the sql one has 1 extra method. It seems like both those protocols should extend this one. Could you do that in this PR?
It also seems like we should pull more up to this parent protocol (eg everything but `param_defaults`). Looking at the code and docstrings, it looks to have been a historical copy/paste.

great_expectations/datasource/fluent/spark_generic_partitioners.py
```diff
@@ -100,11 +102,13 @@ def batch_parameters_to_batch_spec_kwarg_identifiers(
         """Validates all the datetime parameters for this partitioner exist in `options`."""
         identifiers: Dict = {}
         for part in self.param_names:
-            if part not in options:
-                raise ValueError(f"'{part}' must be specified in the batch parameters")  # noqa: TRY003
-            identifiers[part] = options[part]
+            if part in options:
```
billdirks, line 46: Why can we remove the `ValueError` here?

tests/datasource/fluent/conftest.py
```diff
@@ -60,9 +67,11 @@
 
     from great_expectations.data_context import CloudDataContext
 
+CreateSourceFixture: TypeAlias = Callable[..., ContextManager[PostgresDatasource]]
 FLUENT_DATASOURCE_TEST_DIR: Final = pathlib.Path(__file__).parent
 PG_CONFIG_YAML_FILE: Final = FLUENT_DATASOURCE_TEST_DIR / FileDataContext.GX_YML
-
+_DEFAULT_TEST_YEARS = list(range(2021, 2022 + 1))
+_DEFAULT_TEST_MONTHS = list(range(1, 13))
```
billdirks, line 74: nit: Make this `12 + 1` to match `2022 + 1`, or perhaps better, make above `2023`.

tests/datasource/fluent/conftest.py
```diff
@@ -409,3 +418,61 @@ def seeded_contexts(
...
+            for month in _DEFAULT_TEST_MONTHS
+        ]
+    )
+
+    execution_eng_cls = sqlachemy_execution_engine_mock_cls(
+        validate_batch_spec=validate_batch_spec,
+        dialect=dialect,
+        partitioner_query_response=partitioner_response,
+    )
+    original_override = PostgresDatasource.execution_engine_override  # type: ignore[misc]
```
billdirks, line 459: Not for this PR but it seems like `execution_engine_override` should not be a `ClassVar`. It would be great if we could instantiate the PostgresDatasource without having to manage state with try/finally.

great_expectations/datasource/fluent/spark_generic_partitioners.py
tyler-hoffman: Partitioners know about sorting now, so users are free to not pass in any of the params, and just rely on sorting.

This made sense when we required users to pass in batch parameters to select the specific batch, but now that we let sorters do the heavy lifting if users don't pass in batch params, this was raising in cases when we didn't want it to.

tests/datasource/fluent/conftest.py
tyler-hoffman: Yeah, seems reasonable. And just in case it got lost in all the noise of this PR: this was just moved from a deeper-nested test file.

great_expectations/datasource/fluent/interfaces.py
tyler-hoffman: Done! Good call

great_expectations/datasource/fluent/invalid_datasource.py
```diff
@@ -101,7 +101,7 @@ def get_batch_list_from_batch_request(self, batch_request: BatchRequest) -> NoRe
         self._raise_type_error()
 
     @override
-    def sort_batches(self, batch_list: List[Batch]) -> None:
+    def sort_batches(self, batch_list: List[Batch], partitioner) -> None:
```
billdirks: Please add the type to `partitioner` here.

great_expectations/datasource/fluent/spark_generic_partitioners.py
```diff
@@ -15,80 +14,23 @@
...
-                 This is used for all other partitioners and the batch_spec_kwargs["batch_identifiers"]
-                 look like:
-                   {column_name_1: value, column_name_2: value, ...}
-                 where value is the value of the column after being processed by the partitioner.
-                 For example, for the PartitionerModInteger where mod = 3,
-                 {"passenger_count": 2}, means the raw passenger count value is in the set:
-                 {2, 5, 8, ...} = {2*n + 1 | n is a nonnegative integer }
-                 This category was only 1 parameter per column.
-        """  # noqa: E501
+class _Partitioner(PartitionerProtocol): ...
```
billdirks: When you inherit from `Protocol` in python and you want it to still be a `Protocol` you need to explicitly put `Protocol` in the class definition.
```
class _Partitioner(PartitionerProtocol, Protocol): ...
```

[See the docs here](https://mypy.readthedocs.io/en/stable/protocols.html#defining-subprotocols-and-subclassing-protocols).

billdirks (approved):
Thanks, LGTM! A couple very minor comments.

great_expectations/datasource/fluent/spark_generic_partitioners.py
tyler-hoffman: Hah I found that on the other one and forgot to copy it over here. Thanks!

PR #9717, opened 2024-04-05
[FEATURE] Implement LegacyBatchDefinition.batching_regex

joshua-stauffer:
This PR adds a `batching_regex` parameter to `LegacyBatchDefinition`. This is necessary because the DataConnector requires the `batching_regex` in order to produce `BatchSpec`. Previously, the DataConnector used a single `batching_regex` throughout its lifecycle, but in V1 we want to support multiple regexes per asset. 

## previous work
- https://github.com/great-expectations/great_expectations/pull/9704
- https://github.com/great-expectations/great_expectations/pull/9706
- https://github.com/great-expectations/great_expectations/pull/9710

PR #9710, opened 2024-04-04
[FEATURE] Add BatchRequest.batching_regex

joshua-stauffer:
This PR adds a `batching_regex` parameter to `BatchRequest`. If its present, the `FilePathDataConnector` will now use it for its calculations. If it is not present, the DataConnector continues to use the `batching_regex` provided on instantiation.

## previous work

- https://github.com/great-expectations/great_expectations/pull/9704
- https://github.com/great-expectations/great_expectations/pull/9706

PR #9721, opened 2024-04-08
[FEATURE] Add BatchDefinition.batching_regex

joshua-stauffer:
This PR adds `batching_regex` as an attribute of `BatchDefinition`, and ensure its correctly passed to a `BatchRequest`.

 
Previous work to support batching_regex at the DataConnector level:
- https://github.com/great-expectations/great_expectations/pull/9704
- https://github.com/great-expectations/great_expectations/pull/9706
- https://github.com/great-expectations/great_expectations/pull/9710
- https://github.com/great-expectations/great_expectations/pull/9717

### DIFF
--- a/great_expectations/core/batch_definition.py
+++ b/great_expectations/core/batch_definition.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 from typing import TYPE_CHECKING, Optional
 
 from great_expectations.compatibility import pydantic
@@ -26,6 +27,7 @@ class BatchDefinition(pydantic.BaseModel):
     id: Optional[str] = None
     name: str
     partitioner: Optional[Partitioner] = None
+    batching_regex: Optional[re.Pattern] = None
 
     # private attributes that must be set immediately after instantiation
     _data_asset: DataAsset = pydantic.PrivateAttr()
@@ -43,7 +45,9 @@ class BatchDefinition(pydantic.BaseModel):
     ) -> BatchRequest:
         """Build a BatchRequest from the asset and batch request options."""
         return self.data_asset.build_batch_request(
-            options=batch_request_options, partitioner=self.partitioner
+            options=batch_request_options,
+            partitioner=self.partitioner,
+            batching_regex=self.batching_regex,
         )
 
     def save(self) -> None:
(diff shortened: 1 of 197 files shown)


great_expectations/datasource/fluent/interfaces.py
```diff
@@ -228,6 +230,7 @@ def build_batch_request(
             batch_slice: A python slice that can be used to limit the sorted batches by index.
                 e.g. `batch_slice = "[-5:]"` will request only the last 5 batches after the options filter is applied.
             partitioner: A Partitioner used to narrow the data returned from the asset.
+            batching_regex: A Regular Expression used to build batches in some types of Assets.
```
billdirks: s/in some of/in path based/

great_expectations/core/batch_definition.py
```diff
@@ -26,6 +27,7 @@ class BatchDefinition(pydantic.BaseModel):
     id: Optional[str] = None
     name: str
     partitioner: Optional[Partitioner] = None
+    batching_regex: Optional[re.Pattern] = None
```
billdirks, line 30: I'm wondering since we don't use `batching_regex` is some assets if we should really have multiple classes. That is we could have a baseclass or protocol without `batching_regex` and a concrete implementation with and without it. Then we wouldn't need to do the runtime checks for invalid fields for particular assets. This isn't necessary for this PR.

billdirks (approved):
LGTM, thanks!

great_expectations/core/batch_definition.py
joshua-stauffer: yeah, that sounds reasonable. The fact that we're always constructing BatchDefinition for users does make this simpler.

tyler-hoffman: I think I lean toward multiple classes as well. But before doing that, I think we need to think through some of the implications. e.g. does an asset have a `batch_definition: BatchDefinitionWithBatchingRegex`, or does it have a `batch_definition: BatchDefinition` and then do runtime checks to ensure it has a batching regex.

great_expectations/datasource/fluent/interfaces.py
joshua-stauffer: [66023b6](https://github.com/great-expectations/great_expectations/pull/9721/commits/66023b690e233a582597322e30a0164f716ccf80)

tyler-hoffman (approved):

PR #9727, opened 2024-04-09
[MAINTENANCE] Remove renderers from action constructors

cdkini:
We allow a user to pass in a renderer but require a very specific type and corresponding API; we shouldn't allow for this kind of flexibility if it won't be used.

### DIFF
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -30,10 +30,8 @@ from great_expectations.checkpoint.util import (
 )
 from great_expectations.compatibility.pydantic import (
     BaseModel,
-    Field,
     PrivateAttr,
     root_validator,
-    validator,
 )
 from great_expectations.compatibility.pypd import pypd
 from great_expectations.compatibility.typing_extensions import override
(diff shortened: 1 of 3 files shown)


great_expectations/checkpoint/actions.py
```diff
@@ -66,21 +61,6 @@
 logger = logging.getLogger(__name__)
 
 
-def _build_renderer(config: dict) -> Renderer:
-    renderer = instantiate_class_from_config(
-        config=config,
-        runtime_environment={},
-        config_defaults={"module_name": "great_expectations.render.renderer"},
-    )
-    if not renderer:
```
cdkini, line 75: Let's get rid of `instantiate_class_from_config` instances if we can

great_expectations/checkpoint/actions.py
```diff
@@ -248,18 +213,6 @@ class SlackNotificationAction(DataDocsAction):
     notify_on: Literal["all", "failure", "success"] = "all"
     notify_with: Optional[List[str]] = None
     show_failed_expectations: bool = False
-    renderer: SlackRenderer = Field(default_factory=SlackRenderer)
```
cdkini, line 251: If I'm a user, am I really writing my own `SlackRenderer` and passing it in here? We allow users this flexibility but I feel like it isn't actually used.

I feel like this param is a legacy remnant of our overly customizable pre-V1 API and we should be a bit more rigid moving forward?

PR #9732, opened 2024-04-10
[FEATURE] BatchDefinition fluent API for SQL Assets

joshua-stauffer:
This PR updates `DataAsset.add_batch_definition` and implements a fluent API to add BatchDefinitions from `_SqlAsset`.


great_expectations/datasource/fluent/interfaces.py
```diff
@@ -257,7 +257,10 @@ def _validate_batch_request(self, batch_request: BatchRequest) -> None:
 
     @public_api
     def add_batch_definition(
-        self, name: str, partitioner: Optional[Partitioner] = None
+        self,
+        name: str,
+        partitioner: Optional[Partitioner] = None,
+        batching_regex: Optional[re.Pattern] = None,
```
billdirks, line 263: This seems too high in the hierarchy since SQL DataAssets will also inherit from here and don't use the `batching_regex`.

Do we need a generic `add_batch_definition` now that we have specific ones like `add_batch_definition_daily`? Or maybe we remove the public_decorator api if we want some general functionality here. It would still feels odd to have the `batching_regex` argument since that doesn't apply to all assets. I know we only have 1 BatchDefinition class now so it might be ok if this wasn't public but that punts the work to putting it where it should be if we introduce multiple of these.

joshua-stauffer: yeah, that's a good point - i think the best path forward is to use this as the canonical way to add a batch def, but remove it from the public api

tyler-hoffman: Definitely agree on removing `public_api`. But I agree with @billdirks  that the best long term solution is not to have any reference to batch regex at this layer.

My understanding from earlier this week was that just like day/month/year + column name defines how to partition sql, batching_regex (with groups for day/month/year) defines how to partition file-based, which I thought meant that batching_regex gets wrapped in pandas/spark partitioner. Maybe this is an implementation detail I'm off on though.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -688,6 +689,44 @@ def build_batch_request(
             batching_regex=None,
         )
 
+    @public_api
+    def add_batch_definition_whole_table(self, name: str) -> BatchDefinition:
+        return self.add_batch_definition(
+            name=name,
+            partitioner=None,
```
tyler-hoffman, line 696: nit, so ignore me: we have defaults on the parent class - I think we're good to trust them.

tests/datasource/fluent/data_asset/test_sql_asset.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+@pytest.mark.unit
+def test_add_batch_definition_fluent_sql__add_batch_definition_whole_table():
+    # arrange
+    mock_datasource = create_autospec(SQLDatasource)
+    asset = _SQLAsset(
+        name="test_asset",
+    )
+    asset._datasource = mock_datasource  # same pattern Datasource uses to init Asset
```
tyler-hoffman: Wondering if it's worth it to make a fixture that returns the already set up (datasource, asset) tuple. It might not be. Just when I see comments like this where we need to explain something, I think it makes the benefit of the abstraction a bit higher.

tyler-hoffman (changes requested):
Blocking because I'd really like for us to not put the regex pattern at that level of inheritance. I can be convinced if needed, but I want to discuss other options.

But otherwise, this looks lovely.

great_expectations/datasource/fluent/interfaces.py
joshua-stauffer: thanks @tyler-hoffman! `batching_regex` doesn't get wrapped into a partitioner, although it has a similar function. If we have Asset-specific BatchDefinition objects -- say `SqlBatchDefinition` -- then it makes sense to just have a protocol for `add_batch_definition` top level and, and each branch of the inheritance chain needs to implement it. But since we have a single type of BatchDefinition it feels a bit strange to me to exclude a property from the signature, even if some inheriting classes won't use it. 

To put it another way, the responsibility of this function is to construct a BatchDefinition, which has a `batching_regex` property, so it makes sense to me that this base method would have it as well.

tyler-hoffman: > doesn't get wrapped into a partitioner
Is there a reason not to do this, though? We've already branched on on partitioners, so wouldn't it make sense to put them there?

My concerns are:
* A nullable field that only gets used by some implementations feels like a (slight) smell. It's like the parent class knowing about specific child classes, which shouldn't be the case. Why would the base class even know about batching regex? Similar to why would they know about column_name that only gets used for SQL
* A nullable field requires more runtime checking, and the types don't tell engineers what to expect
* It seems more clear to colocate the mechanism for determining year/month/day/etc with type of sorting. i.e. then it parallels column_name + partitioner type. I think a point backing this up is that a nullable partitioner + nullable batching regex means we have to account for cases where engineers only pass in one - in this world we expect both or none, right?

Probably it's higher priority to get this stuff working than for me to argue about this, so maybe I'm in favor of merging and then revisiting this around the time we tackle https://greatexpectations.atlassian.net/browse/V1-281

tyler-hoffman (approved):
Discussed synchronously - let's ship this and revisit later

tests/datasource/fluent/data_asset/test_sql_asset.py
joshua-stauffer: sure, added in [aced278](https://github.com/great-expectations/great_expectations/pull/9732/commits/aced2789ad0a4452365c764257facaa450c18412). I typically prefer tests to be all inline if possible for legibility, but i think it's pretty equivalent in this case.

tests/datasource/fluent/data_asset/test_sql_asset.py
```diff
@@ -12,92 +12,86 @@
 from great_expectations.datasource.fluent.sql_datasource import _SQLAsset
 
 
-@pytest.mark.unit
-def test_add_batch_definition_fluent_sql__add_batch_definition_whole_table():
-    # arrange
-    mock_datasource = create_autospec(SQLDatasource)
+@pytest.fixture
+def datasource() -> Mock:
```
tyler-hoffman: Non-blocking, but I think the best practice here is to use mocker.

joshua-stauffer: yep! 😄 was replaced in [166c692](https://github.com/great-expectations/great_expectations/pull/9732/commits/166c692eb6202d62f0382467c8f9ff4e9c53f330)

PR #9733, opened 2024-04-10
[MAINTENANCE] Make actions sortable

cdkini:
`UpdateDataDocs` action should always be placed first in a sort 


great_expectations/checkpoint/actions.py
```diff
@@ -119,6 +122,9 @@ class Config:
     type: str
     notify_on: Literal["all", "failure", "success"] = "all"
 
+    def __lt__(self, other: ValidationAction) -> bool:
+        return self._sorting_priority < other._sorting_priority
```
cdkini: YAGNI? I preferred this to an explicit "hey data docs action is the very specific case we have to call out"

billdirks: Yeah, I'd just pulled out the special data docs actions now and put it at the top, leaving the rest unchanged. I don't know if we need/want general sorting.

billdirks (changes requested):
I think we should keep it simple now and just move the datadocs action to the top if it exists leaving the rest unchanged.

great_expectations/checkpoint/v1_checkpoint.py
```diff
@@ -221,13 +221,31 @@ def _run_actions(
...
+        """
+        priority_actions: List[CheckpointAction] = []
+        secondary_actions: List[CheckpointAction] = []
+        for action in self.actions:
+            if isinstance(action, UpdateDataDocsAction):
+                priority_actions.append(action)
+            else:
+                secondary_actions.append(action)
+
+        return priority_actions + secondary_actions
```
Kilo59, line 247: `non-blocking`
We should be able to embedd this into the `CheckpointActions` themselves.
And then you can just call `sorted(my_list)` or `my_list.sort()`

Implement `UpdateDataDocsAction.__gt__` and have it return true if the item being compared is not a `UpdateDataDocsAction.`

https://docs.python.org/3/reference/datamodel.html#object.__gt__
https://docs.python.org/3/reference/datamodel.html#object.__lt__

https://docs.python.org/3/howto/sorting.html#sorting-basics

## Edit:
Wrote this before I saw this discussion. I still think embedding this into the action themselves can be useful. And is a good way to prevent someone who doesn't have the context from sorting them incorrectly.
https://github.com/great-expectations/great_expectations/pull/9733#discussion_r1559834008

billdirks (approved):
LGTM

great_expectations/checkpoint/v1_checkpoint.py
cdkini: @Kilo59 in discussing this will @billdirks further, I think we've decided to defer such efforts until they are actually required. In the case we need to deal with dependency graphs when it comes to these actions, we should determine a more robust solution than the naive sorting solution I previously wrote.

Appreciate you chiming in here though!

billdirks: We are keeping sorting internal for now. There is only 1 action that needs to run before everything else. It's unclear if we really want to sort in general or what the requirements are (eg do we really need a dependency graph and is it really better to state your dependencies explicitly vs a sort priority). So I want to punt on proper sorting for now and anything that touches our public api (we expect people to write more actions).

PR #9734, opened 2024-04-10
[MAINTENANCE] Migrate `SlackNotificationAction` to V1 pattern

cdkini:


great_expectations/checkpoint/actions.py
```diff
@@ -300,6 +300,73 @@ def _root_validate_slack_params(cls, values: dict) -> dict:
...
+        )
+        blocks = query.get("blocks")
+        if blocks:
+            if len(blocks) >= 1:
+                if blocks[0].get("text"):
+                    result = self._send_notifications_in_batches(
+                        blocks=blocks, query=query, result=result
+                    )
+                else:
+                    result = self._get_slack_result(query=query)
```
cdkini: There's a decent bit of duplication from the existing run method but I found it easier to do this than refactor old code that we're going to soon delete?

tests/checkpoint/test_v1_checkpoint.py
```diff
@@ -494,20 +494,23 @@ def test_checkpoint_run_no_actions(self, validation_definition: ValidationDefini
...
 
-        with mock.patch.object(ValidationAction, "v1_run") as mock_run:
-            result = checkpoint.run()
+        # Act
+        result = checkpoint.run()
 
-        assert mock_run.call_count == len(actions)
-        mock_run.assert_called_with(checkpoint_result=result, action_context=mock.ANY)
+        # Assert
+        action._copy_and_set_values().v1_run.assert_called_once_with(
```
tyler-hoffman, line 513: `action._copy_and_set_values()` <- not loving reaching into private mock methods, but probably not the end of the world. This is something having to do with pydantic cloning the mock?

great_expectations/checkpoint/actions.py
tyler-hoffman: yeah, seems legit to me.

great_expectations/checkpoint/util.py
```diff
@@ -44,23 +44,17 @@ def send_slack_notification(query, slack_webhook=None, slack_channel=None, slack
 
     try:
         response = session.post(url=url, headers=headers, json=query)
-        if slack_webhook:
-            ok_status = response.text == "ok"
```
tyler-hoffman, line 48: I'm assuming these ok checks were just redundant because we always got okay for 200 responses?

great_expectations/render/renderer/slack_renderer.py
```diff
@@ -1,14 +1,169 @@
...
+                docs_link = data_docs_pages[docs_link_key]
+                report_element = self._get_report_element(docs_link)
+                return report_element
+
+        return None
+
+    def _build_divider_block(self) -> dict:
+        return {"type": "divider"}
+
+    def build_query(self, checkpoint_result: CheckpointResult, text_blocks: list[dict]) -> dict:
```
tyler-hoffman: I see the term query used in the old code as well. Reading the code, it looks like this should be `footer`.

Also, the behavior here of this method seems like it could be confusing. mutating text_blocks and then returning another object. I think the most readable solution would just be to have this method just responsible for building the footer section and returning that, then letting `actions.py` do what it needs to with it.

great_expectations/checkpoint/actions.py
```diff
@@ -300,6 +300,73 @@ def _root_validate_slack_params(cls, values: dict) -> dict:
...
+                result_identifier=validation_result_suite_identifier,
+                result=validation_result_suite,
+                action_context=action_context,
+            )
+            checkpoint_text_blocks.extend(validation_text_blocks)
+
+        query = self.renderer.build_query(
+            checkpoint_result=checkpoint_result, text_blocks=checkpoint_text_blocks
+        )
+        blocks = query.get("blocks")
```
tyler-hoffman: the flow here with `query` and `blocks` doesn't feel intuitive to me. I also left comments on this below in `build_query`. The way this is structured, it's not clear that `query.get_blocks is checkpoint_text_blocks`, but it looks like that is the case.

tyler-hoffman (changes requested):
Blocking with minor requested changes around the `build_query` method. Also left a question about checking for `ok` status.

tests/checkpoint/test_v1_checkpoint.py
cdkini: Ugh yeah I'm researching alternative ways to deal with this but I'm not sure

great_expectations/checkpoint/util.py
cdkini: yup I believe so - I've made the checks a bit more robust with `raise_for_status`

great_expectations/render/renderer/slack_renderer.py
```diff
@@ -1,14 +1,171 @@
...
+            "type": "context",
+            "elements": [
+                {
+                    "type": "mrkdwn",
+                    "text": f"Learn how to review validation results in Data Docs: {documentation_url}",  # noqa: E501
+                }
+            ],
+        }
+
+        all_blocks = [block for block in text_blocks] + [footer_section]
```
tyler-hoffman: nit: isn't this the same as `text_blocks + [footer_section]`?

tyler-hoffman (approved):

great_expectations/render/renderer/slack_renderer.py
cdkini: fixed!

Issue #9690, opened 2024-04-02
fluent_datasources does not reflect runtime datasource addition


ramananayak:
**Describe the bug**
I want to add fluent_datasource at runtime after a FileDataContext is already defined.
context.fluent_datasources  is of type dictionary. When I add a new fluent_datasource, it does not add to the existing dictionary.
Where as it works on datasource.

**To Reproduce**
```python
import great_expectations as gx
context = gx.data_context.FileDataContext(context_root_dir="my_context_dir")
connection_string = "postgresql+psycopg2://<user_name>:<password>@<host>:<port>/<database>"
runtime_datasource = gx.datasource.fluent.PostgresDatasource(name="ds_runtime", connection_string=connection_string,create_temp_table=True)

# Running below does not update the dictonary
context.fluent_datasources[runtime_datasource.name] = runtime_datasource

# where as, if I run below command then it will update properly. Also, it will also update fluent_datasources
context.datasources[runtime_datasource.name] = runtime_datasource
```

**Expected behavior**
context.fluent_datasources should show added runtime_datasource inside the dictonary

**Environment (please complete the following information):**
 - Operating System: MacOS
 - Great Expectations Version: 0.18.12
 - Data Source: Redshift
 - Cloud environment: AWS

**Additional context**

Kilo59:
@ramananayak 
Sorry for the confusion this is because the `context.fluent_datasources` property is just a dictionary comprehension of `context.datasources` with all non-fluent datasources filtered out.

Would could alter the return type annotation to be an Immutable `Mapping[str, FluentDatasource]` to help with this. But it wouldn't alter runtime behavior and you'd have to rely on a type-checker or IDE to warn about it being immutable.

https://github.com/great-expectations/great_expectations/blob/abcf67159ad6562478318e659771237561b107b4/great_expectations/data_context/data_context/abstract_data_context.py#L4375-L4381

The idiomatic way to add or update a datasource is by using one of the `context.sources.add_or_update_<DATASOURCE_TYPE>()` methods. This method also bootstraps the datasource with the components needed to do config substitution and connect certain datasources to things like s3/gcs/databricks etc.

```python
import great_expectations as gx
context = gx.data_context.FileDataContext(context_root_dir="my_context_dir")
connection_string = "postgresql+psycopg2://<user_name>:<password>@<host>:<port>/<database>"

runtime_datasource = context.sources.add_or_update_postgres(
    name="ds_runtime",
    connection_string=connection_string,
[...]

PR #9738, opened 2024-04-11
[MAINTENANCE] Export `great_expectations.data_context` types

Kilo59:
Set the return type of `context.fluent_datasources` to be an immutable `Mapping` to reflect that it should not be modified.

- See https://github.com/great-expectations/great_expectations/issues/9690

We also need to export the types for the `great_expectations.data_context` package so that type-checkers will actually raise warnings about improper usage.

- For more a more in-depth explanation see -> https://github.com/great-expectations/great_expectations/pull/7438

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -8,6 +8,7 @@ To show all available tasks `invoke --list`
 
 To show task help page `invoke <NAME> --help`
 """
+
 from __future__ import annotations
 
 import logging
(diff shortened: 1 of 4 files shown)


great_expectations/data_context/py.typed
Kilo59, line 1: The presence of the file is what controls which types get "exported".
We don't want to export types for all of great_expectations because a lot of our types are still "wrong".

Kilo59:
I can make an equivalent PR for `develop` but things like `context.fluent_datasources` don't really make sense for `V1` since everything will be fluent.

tyler-hoffman (approved):

tasks.py
```diff
@@ -277,7 +278,7 @@ def type_check(  # noqa: PLR0913, PLR0912
         print(f"  Clearing {mypy_cache} ... ", end="")
         try:
             shutil.rmtree(mypy_cache)
-            print("✅"),
+            (print("✅"),)
```
Kilo59, line 281:
```suggestion
            print("✅")
```

PR #9743, opened 2024-04-12
[MAINTENANCE] Rename Evaluation Parameter to Suite Parameter

joshua-stauffer:
This PR replaces Evaluation Parameter and closely related spellings with Suite Parameter.


great_expectations/expectations/expectation_configuration.py
```diff
@@ -20,12 +20,12 @@
 from typing_extensions import TypedDict
 
 from great_expectations.compatibility.typing_extensions import override
-from great_expectations.core.evaluation_parameters import (
-    _deduplicate_evaluation_parameter_dependencies,
-    build_evaluation_parameters,
-    find_evaluation_parameter_dependencies,
-)
 from great_expectations.core.metric_domain_types import MetricDomainTypes
+from great_expectations.core.suite_parameters import (
```
billdirks, line 24: I want to verify this doesn't break our integration work. I believe we're no longer serializing evaluation parameters.

joshua-stauffer: yeah, i'm currently working through the cloud integration - requires more workarounds than i hoped, but making progress.

billdirks (approved):
Thanks for doing this!
I looked through a lot of these files but not all. The pattern LGTM. I left 1 question as a one-off about making sure we aren't breaking any 1.0 cloud integration work.

PR #9756, opened 2024-04-15
[FEATURE] Add sort_ascending to BatchDefinition fluent API

joshua-stauffer:
Recent changes added a `sort_ascending: bool` parameter to Partitioners, so this PR allows defining that from the BatchDefinition fluent API.

### DIFF
--- a/tests/datasource/fluent/data_asset/test_sql_asset.py
+++ b/tests/datasource/fluent/data_asset/test_sql_asset.py
@@ -97,55 +97,76 @@ def test_add_batch_definition_fluent_sql__add_batch_definition_whole_table(datas
 
 
 @pytest.mark.unit
-def test_add_batch_definition_fluent_sql__add_batch_definition_yearly(datasource, asset):
-    # arrange
-    name = "batch_def_name"
-    column = "test_column"
-    expected_batch_definition = BatchDefinition(
-        name=name, partitioner=PartitionerYear(column_name=column), batching_regex=None
-    )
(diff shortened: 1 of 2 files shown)


tests/datasource/fluent/data_asset/test_sql_asset.py
```diff
@@ -97,55 +97,76 @@ def test_add_batch_definition_fluent_sql__add_batch_definition_whole_table(datas
...
+        ),
         batching_regex=None,
     )
     datasource.add_batch_definition.return_value = expected_batch_definition
 
     # act
-    batch_definition = asset.add_batch_definition_daily(name=name, column=column)
+    batch_definition = asset.add_batch_definition_daily(
+        name=name, column=column, sort_ascending=sort_ascending
+    )
```
cdkini, line 169: The output of this test doesn't change based on the new param right? Could we add a test where the order matters?

joshua-stauffer: thanks @cdkini! These methods just return a BatchDefinition, not a list of batches; that functionality is tested in `test_get_batch_list_from_batch_request__sort_ascending/descending`.  I created [a followup ticket](https://greatexpectations.atlassian.net/browse/V1-291) to add integration tests around the fluent batch def api.

billdirks: I will add a comment in that ticket that explicitly lists testing sorters as part of the AC.

billdirks (approved):

PR #9757, opened 2024-04-15
[MAINTENANCE] Migrate ci to enterprise-arc runners

nicgrayson:


tasks.py
```diff
@@ -1042,7 +1042,7 @@ def ci_tests(  # noqa: C901 - too complex (9)
     pytest_options = [f"--durations={slowest}", "-rEf"]
 
     if xdist:
-        pytest_options.append("-n auto")
+        pytest_options.append("-n 4")
```
billdirks, line 1088: Did you mean to hard code this to 4?

billdirks (approved):
Thanks! One question about the xdist number of workers.

tasks.py
nicgrayson: This was on purpose. Kubernetes exposes all cores of the host machine to the pod even if the pod is allocated fewer

Kilo59: Will logical core give the right number?
`4` might be good for CI, but it'll slow things down when testing locally.

Can you leave this as `auto` (or `logical`) and update the the `ci.yaml` to pass the hardcoded 4 value.

nicgrayson: no, should we make this something that changes when it detects that it's running in CI?

PR #9758, opened 2024-04-15
[MAINTENANCE] Enable `SIM101` + `SIM114` 

Kilo59:
Part 1 of multiple PR's to enable the `SIM` rules.
https://docs.astral.sh/ruff/rules/#flake8-simplify-sim

### Rules included
* SIM101- https://docs.astral.sh/ruff/rules/duplicate-isinstance-call/
* SIM114 - https://docs.astral.sh/ruff/rules/if-with-same-arms/

### DIFF
--- a/great_expectations/data_context/store/tuple_store_backend.py
+++ b/great_expectations/data_context/store/tuple_store_backend.py
@@ -76,6 +76,13 @@ class TupleStoreBackend(StoreBackend, metaclass=ABCMeta):
             self.verify_that_key_to_filepath_operation_is_reversible()
             self._fixed_length_key = True
 
+    def _is_missing_prefix_or_suffix(
+        self, filepath_prefix: str, filepath_suffix: str, key: str
+    ) -> bool:
+        missing_prefix = filepath_prefix and not key.startswith(filepath_prefix)
+        missing_suffix = filepath_suffix and not key.endswith(filepath_suffix)
+        return missing_prefix or missing_suffix
+
     @override
     def _validate_key(self, key) -> None:
         super()._validate_key(key)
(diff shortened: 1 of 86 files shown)


pyproject.toml
```diff
@@ -328,6 +330,26 @@ lint.ignore = [
...
+    "SIM105", # suppressible-exception
+    "SIM117", # multiple-with-statements
+    "SIM201", # negate-equal-op
+    "SIM110", # reimplemented-builtin 
+    "SIM910", # dict-get-with-none-default
+    "SIM401", # if-else-block-instead-of-dict-get
+    "SIM115", # open-file-with-context-handler
+    "SIM202", # negate-not-equal-op
+    "SIM210", # if-expr-with-true-false
+    "SIM112", # uncapitalized-environment-variables
```
Kilo59: This ones still need to be enabled in followup PRs

great_expectations/data_context/store/tuple_store_backend.py
```diff
@@ -76,6 +76,14 @@ def __init__(  # noqa: PLR0913
             self.verify_that_key_to_filepath_operation_is_reversible()
             self._fixed_length_key = True
 
+    @staticmethod
```
anthonyburdi, line 79: I'm ok if we remove the `@staticmethod` decorator or leave it. I don't have a strong argument either way.

great_expectations/datasource/fluent/snowflake_datasource.py
```diff
@@ -136,12 +136,16 @@ def _check_xor_input_args(cls, values: dict) -> dict:
...
+            )
             # Method 2 - individual args (account, user, and password are bare minimum)
-            elif isinstance(connection_string, ConnectionDetails) and bool(
+            has_min_connection_detail_values: bool = isinstance(
+                connection_string, ConnectionDetails
+            ) and bool(
                 connection_string.account and connection_string.user and connection_string.password
-            ):
+            )
+            if is_connection_string or has_min_connection_detail_values:
```
anthonyburdi, line 148: Nice!

anthonyburdi (approved):
LGTM! 
I think `tests/test_definitions/test_expectations_v3_api.py` could use some refactoring but I'm OK if that is not included here.

great_expectations/data_context/store/tuple_store_backend.py
Kilo59: If it isn't relying on instance or class attributes I think it's better as a static method.

PR #9774, opened 2024-04-16
[MAINTENANCE] Rename validations -> validation_results

tyler-hoffman:
Satisfies https://greatexpectations.atlassian.net/browse/V1-261.

There is no longer just a "validations_store", so this was just a giant find/replace, ignoring our changelog and versioned_docs/.

There are 2 files reviewers should give extra attention:
* `great_expectations/data_context/data_context/cloud_data_context.py`: updated the logic with some mappings from keys/values that will come back from cloud to what they need to be in 1.0
* `tests/data_context/test_cloud_data_context.py`: tests around ^


great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -274,41 +274,54 @@ def _prepare_v1_config(cls, config: dict) -> dict:
         # to prevent downstream issues
         config["fluent_datasources"] = _extract_fluent_datasources(config)
 
-        # V1 renamed EvaluationParameters to SuiteParameters, so this is a temporary patch
-        # until Cloud implements a V1 endpoint for DataContextConfig
-        config = cls._replace_evaluation_parameters_with_suite_parameters(config)
+        # V1 renamed EvaluationParameters to SuiteParameters, and Validations to ValidationResults
```
tyler-hoffman, line 280: Hey Reviewer! Please give this an extra set of eyes!

tests/data_context/test_cloud_data_context.py
```diff
@@ -0,0 +1,202 @@
...
+                "store_backend": {
+                    "base_directory": "expectations/",
+                    "class_name": "TupleFilesystemStoreBackend",
+                },
+            },
+        },
+        validation_results_store_name_key: "default_validation_results_store",
+    }
+
+
```
tyler-hoffman, line 152: Reviewer! Over here!

docs/docusaurus/versioned_docs/version-0.17/deployment_patterns/how_to_use_great_expectations_with_google_cloud_platform_and_bigquery.md
```diff
@@ -109,7 +109,7 @@ By default, Validations are stored in JSON format in the `uncommitted/validation
 ```YAML title="YAML" name="version-0.17.23 docs/docusaurus/versioned_docs/version-0.17.23/snippets/gcp_deployment_patterns_file_gcs.py expected_validations_store"
 ```
 
-Great Expectations can then be configured to use this new Validations Store, `validations_GCS_store`, by setting the `validations_store_name` value in the `great_expectations.yml` file.
```
cdkini: Could we revert changes made to `versioned_docs`?

tyler-hoffman: > Could we revert changes made to `versioned_docs`?

Good catch @cdkini ! backed those changes out!

great_expectations/data_context/data_context/cloud_data_context.py
cdkini: Could we use constants here for the names? I think we might have some of these as `DataContextVariables` or in `great_expectations/data_context/types/base.py`?

cdkini: Alternatively, should we change test code instead of source code here? I'm assuming we're making these changes to allow tests to pass right?

tyler-hoffman: AFAIK We need to do this mapping in the source code, at least for the fields not prefixed with "default", as the data coming back from cloud will have the old keys. But looking again, I think we can probably get rid of the renaming of "default_evaluation_parameter_store" and "default_validations_store", since I think those can be whatever.

I'll look into removing the stuff about "default_", and use the constants from `DataContextVariableSchema` for the v1 values (thanks for pointing me in that direction!).

I don't personally care too much about making constants for the old keys, since they are only used once, and I think their meaning should be pretty clear here. But if you feel strongly in the other direction, fell free to throw a blocker on here and make me do it 😄

cdkini (approved):

PR #9785, opened 2024-04-18
[MAINTENANCE] Generic type for Partitioner

joshua-stauffer:
Refactor Partitioner to be a generic type and allow DataAsset implementations to define it.


great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -422,7 +422,7 @@ def batch_parameters_to_batch_spec_kwarg_identifiers(
 ]
 
 
-class _SQLAsset(DataAsset):
+class _SQLAsset(DataAsset[Partitioner]):
```
tyler-hoffman: I have to imagine there's something we need to do here for the datasource type. Maybe just pass in `_DatasourceT`?

great_expectations/datasource/fluent/file_path_data_asset.py
```diff
@@ -79,7 +80,10 @@
 logger = logging.getLogger(__name__)
 
 
-class _FilePathDataAsset(DataAsset):
+_FilePathDatasource = TypeVar("_FilePathDatasource")
```
tyler-hoffman: This will need to be bound to something at least as specific as the TypeVar DataAsset uses. I'd probably make that public and use it here. But you could also just put `_FilePathDatasource = TypeVar("_FilePathDatasource", bound=MetaDatasource)` here.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -239,10 +240,10 @@ def _sorter_from_str(sort_key: str) -> Sorter:
 
 
 # It would be best to bind this to Datasource, but we can't now due to circular dependencies
-_DatasourceT = TypeVar("_DatasourceT", bound=MetaDatasource)
+DatasourceT = TypeVar("DatasourceT", bound=MetaDatasource)
 
 
-class DataAsset(FluentBaseModel, Generic[_DatasourceT]):
+class DataAsset(FluentBaseModel, Generic[DatasourceT, PartitionerT]):
```
Kilo59, line 246: The `ParitionerT` type here is a future annotation.

great_expectations/datasource/fluent/fluent_base_model.py
```diff
@@ -69,7 +66,265 @@ class Config:
...
     @classmethod
-    def parse_yaml(cls: Type[_Self], f: Union[pathlib.Path, str]) -> _Self:
+    def parse_yaml(cls: Type[Self], f: Union[pathlib.Path, str]) -> Self:
+        loaded = yaml.load(f)
+        logger.debug(f"loaded from yaml ->\n{pf(loaded, depth=3)}\n")
+        # noinspection PyArgumentList
+        config = cls(**loaded)
+        return config
+
+    @overload
```
tyler-hoffman: Non-blocking, but I don't love the duplication here. I don't like using multiple inheritance, but maybe the best option here is for `GenericBaseModel` to inherit from `FluentBaseModel` and `pydantic.GenericModel`? Looks like `pydantic.GenericModel inherits from `pydantic.BaseModel`, so I assume it'd work just fine.

great_expectations/datasource/fluent/fluent_base_model.py
```diff
@@ -321,7 +576,9 @@ def _recursively_set_config_value(  # noqa: C901 - too complex
                 _recursively_set_config_value(v, config_provider)
 
 
-def _update__fields_set__on_truthyness(model: FluentBaseModel, field_name: str) -> None:
```
tyler-hoffman: Benefit of the above suggestion: We don't need union types here.

tyler-hoffman (approved):
Just looked everything over and I think it looks solid, but maybe we get another set of eyes since I was pretty close to this work.

great_expectations/datasource/fluent/fluent_base_model.py
joshua-stauffer: if that works, that's great! added in [ef4651e](https://github.com/great-expectations/great_expectations/pull/9785/commits/ef4651e80793253928d7cf9e8755cd9b2c74debc)

great_expectations/datasource/fluent/fluent_base_model.py
joshua-stauffer: [ef4651e](https://github.com/great-expectations/great_expectations/pull/9785/commits/ef4651e80793253928d7cf9e8755cd9b2c74debc)

great_expectations/core/batch_definition.py
```diff
@@ -17,32 +17,36 @@
...
 
     # private attributes that must be set immediately after instantiation
-    _data_asset: DataAsset = pydantic.PrivateAttr()
+    # Note that we're using type Any, but the getter setter ensure the right types.
+    # If we actually specify DataAsset, pydantic errors out.
+    _data_asset: Any = pydantic.PrivateAttr()
 
     @property
-    def data_asset(self) -> DataAsset:
+    def data_asset(self) -> DataAsset[Any, PartitionerT]:
```
billdirks, line 41: Is is possible to change the `Any` to `DatasourceT` or does it have to be `Any` for thread reason you have `_data_asset: Any`?

great_expectations/core/batch_definition.py
```diff
@@ -17,32 +17,36 @@
     )
     from great_expectations.datasource.fluent.interfaces import Batch, DataAsset
 
+PartitionerT = TypeVar("PartitionerT", bound=Optional[Partitioner])
```
billdirks: Can you remove `Optional` from the `bound` argument? It looks like you often use `Optional[Partitioner]` when this can be `None`. I'm not sure if this works but maybe you can define `class BatchRequest(pydantic.GenericModel, Generic[Optional[PartitionerT]])` and `class BatchDefinition(pydantic.GenericModel, Generic[Optional[PartitionerT]])`?

great_expectations/datasource/fluent/file_path_data_asset.py
```diff
@@ -113,7 +116,7 @@ class _FilePathDataAsset(DataAsset):
     _data_connector: DataConnector = pydantic.PrivateAttr()
     # more specific `_test_connection_error_message` can be set inside `_build_data_connector()`
     _test_connection_error_message: str = pydantic.PrivateAttr("Could not connect to your asset")
-    _partitioner_implementation_map: dict[type[Partitioner], type[SparkPartitioner]] = (
+    _partitioner_implementation_map: Dict[Type[Partitioner], Type[SparkPartitioner]] = (
```
billdirks, line 119: We prefer the lowercase versions if they work.

great_expectations/datasource/fluent/fluent_base_model.py
```diff
@@ -41,10 +42,6 @@
 yaml.indent(mapping=2, sequence=4, offset=2)
 yaml.default_flow_style = False
 
-# TODO (kilo59): replace this with `typing_extensions.Self` once mypy supports it
```
billdirks, line 44: Hooray!

billdirks (commented):
Thanks for doing this!

great_expectations/datasource/fluent/file_path_data_asset.py
joshua-stauffer: good catch - changed this while trying to debug some opaque typing issues with pydantic. reverted in https://github.com/great-expectations/great_expectations/pull/9785/commits/d700bccc2a0c67d63aa925c78a52ea752f3146a6

great_expectations/datasource/fluent/fluent_base_model.py
joshua-stauffer: 🥳

great_expectations/core/batch_definition.py
joshua-stauffer: thanks @billdirks! I gave this some more thought, and have updated `PartitionerT` from a single upper bound of `Partitioner | None` to multiple upper bounds of `Partitioner` and `None` (and will soon update this to use the upper bounds of `SqlPartitioner`, `RegexPartitioner`, and `None`. Classes using the type can choose the upper bound that suites their use case, and then decide whether or not to make the field optional where its used. I guess its a subtle distinction, but hopefully make it clearer from reading the type that `PartitionerT` has two distinct valid types.

great_expectations/datasource/fluent/file_path_data_asset.py
joshua-stauffer: oops, actually this _was_ causing issues, so i had to go back to `Type` for now. This should go away when we drop support for 3.8.

PR #9792, opened 2024-04-22
[FEATURE] Add Regex Partitioner

joshua-stauffer:
This PR adds a new type of partitioner specific to file backed assets, which will replace the previously used partitioners.


great_expectations/core/batch_definition.py
```diff
@@ -1,13 +1,12 @@
...
 
-import re
 from typing import TYPE_CHECKING, Any, Generic, Optional, TypeVar
 
 from great_expectations.compatibility import pydantic
 
 # if we move this import into the TYPE_CHECKING block, we need to provide the
 # Partitioner class when we update forward refs, so we just import here.
-from great_expectations.core.partitioners import Partitioner
+from great_expectations.core.partitioners import Partitioner, RegexPartitioner
```
tyler-hoffman, line 9: non-blocking, but the naming here would make someone expect `RegexPartitioner` to be a specific subtype of `Partitioner`, but I'm pretty sure that isn't the case. Maybe a candidate for a mass find/replace in a subsequent PR?

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
```diff
@@ -215,7 +216,10 @@ def _get_unfiltered_batch_definition_list(
         batch_definition_list: list[LegacyBatchDefinition] = list()
         batch_definition_set = set()
         # if the batch request hasn't specified a batching_regex, fallback to a default
-        batching_regex = batch_request.batching_regex or self._batching_regex
+        if batch_request.partitioner:
+            batching_regex = batch_request.partitioner.regex
+        else:
+            batching_regex = self._batching_regex
```
tyler-hoffman, line 222: Is the else condition here an intermediate step that will get removed subsequently?

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -467,6 +460,15 @@ def sort_batches(self, batch_list: List[Batch], partitioner: PartitionerProtocol
                 ) from e
 
 
+class PartitionerI(Protocol):
```
tyler-hoffman: This file also contains `PartitionerProtocol`. Are we safe to delete that? their purpose is similar; this is just the more narrow definition.

tests/datasource/fluent/test_batch_request.py
```diff
@@ -113,7 +124,85 @@ def test_batch_request_config_serialization_round_trips(
...
+            id="Regex Yearly",
+        ),
+        pytest.param(
+            None,
+            BatchRequest[None],
+            id="None",
+        ),
+    ],
+)
+def test_batch_request_config_partitioner_round_trip_serialization(
```
tyler-hoffman: Having this functionality under test is great. But is this testing real-world use cases for deserialization? Will we always have the equivalent of `TypedBatchRequest` at deserialization time? My hunch is "no", but I'm guessing you've thought ahead on this.

Also obligatory nits:
* why not split this into 2 tests: one for dicts and one for json
* the constants make me think those strings will be important/reused, but they aren't.

joshua-stauffer: > is this testing real-world use cases for deserialization?

right, this is what i brought up at standup today. I wrote this test because i wanted to know if Pydantic needs the type to deserialize -- it doesn't, but mypy does. We're avoiding this right now by deserializing batch configs by magic in checkpoints. In the future, i would suggest the Asset/BatchDefinition should always be responsible for retrieving batch configs, and they will always have the correct type context.

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
joshua-stauffer: yes - this is will be removed once `batching_regex` is no longer a property of asset

great_expectations/core/batch_definition.py
joshua-stauffer: yeah - I would like to rename Partitioner to SqlPartitioner (or even better yet, just use the SqlPartitioner defined in the datasources package) but that's not currently possible due to circular imports.

tests/datasource/fluent/test_batch_request.py
joshua-stauffer: re: those nits - i copied the structure of the test above it 🙃 i'll update to address.

great_expectations/datasource/fluent/interfaces.py
joshua-stauffer: no; `PartitionerProtocol` is used by other code

tests/datasource/fluent/test_batch_request.py
joshua-stauffer: [001190e](https://github.com/great-expectations/great_expectations/pull/9792/commits/001190e4ecf3feeda131ba19efd2e84f60d77f42)

tests/datasource/fluent/test_batch_request.py
```diff
@@ -188,21 +188,84 @@ def test_batch_request_config_serialization_round_trips(
...
     batch_request_dict = batch_request.dict()
     assert TypedBatchRequest(**batch_request_dict) == batch_request
 
-    # json
+
+@pytest.mark.unit
+@pytest.mark.parametrize(
+    "partitioner,TypedBatchRequest",
+    [
+        pytest.param(
```
tyler-hoffman: I'm going to seem like a terrible person here, but we should DRY this up probably by making a helper function, assuming this list of `pytest.param` is identical to above.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -448,10 +440,11 @@ def _parse_order_by_sorters(
     ) -> List[Sorter]:
         return Datasource.parse_order_by_sorters(order_by=order_by)
 
-    def sort_batches(self, batch_list: List[Batch], partitioner: PartitionerProtocol) -> None:
+    def sort_batches(self, batch_list: List[Batch], partitioner: PartitionerI) -> None:
         """Sorts batch_list in place in the order configured in this DataAsset.
         Args:
             batch_list: The list of batches to sort in place.
+            partitioner:
```
billdirks: Please add a description for partitioner.

great_expectations/datasource/fluent/interfaces.py
billdirks: It feels odd to have 2 protocols for Partitioners where one is a superset of the other. I would either reuse the other protocol or make the other protocol extend this one.

billdirks: I would also rename this to `PartitionerSortingProtocol` or something more descriptive. Looking at the `I` suffix vs the `Protocol` suffix doesn't give me much info as a future developer of this code.

joshua-stauffer: thanks! updated in [526911a](https://github.com/great-expectations/great_expectations/pull/9792/commits/526911aacbbb025d04749d7a816c57330ece770d)

great_expectations/datasource/fluent/interfaces.py
joshua-stauffer: updated in [ca76ec2](https://github.com/great-expectations/great_expectations/pull/9792/commits/ca76ec27b37b81148fbfe84860311f30b49716fe)

tyler-hoffman (approved):

PR #9796, opened 2024-04-23
[MAINTENANCE] Remove Spark Partitioners

joshua-stauffer:
V1 supports batching file backed assets by file name, not by columns within files, so this functionality is no longer used.


billdirks (approved):
LGTM!

PR #9807, opened 2024-04-24
[MAINTENANCE] Remove `include_rendered_content` flag

cdkini:
Rather than granular control over this, we should default to always rendering content for Cloud and omitting it otherwise.


tests/data_context/test_get_data_context.py
```diff
@@ -272,24 +271,6 @@ def test_get_context_with_mode_equals_cloud_returns_cloud_data_context(
     assert isinstance(context, CloudDataContext)
 
 
-@pytest.mark.parametrize("ge_cloud_mode", [True, None])
-@pytest.mark.cloud
-def test_cloud_context_include_rendered_content(
```
billdirks, line 277: Do we need some cloud test to ensure we are getting rendered content?

billdirks (commented):
This lgtm. Is there a test that we include the rendered content when using cloud. It could be I missed it or is not in the diff since we already check for it and that check is unchanged.

tests/data_context/test_get_data_context.py
cdkini: This is just referring to the top level configuration (which doesn't exist anymore) - rendered content should exist on each object as attributes.

billdirks (approved):
Thank you!

PR #9816, opened 2024-04-25
[BUGFIX] Do not persist interactive batch defs

tyler-hoffman:
Updates the interactive validation flow (via `batch.validate`) to not add `BatchDefinition`s to the asset.

NOTE: The entire flow here is a bit weird, but I think that's a question for another day. Here it is:
* Create a batch definition
* Create a batch from that
* Call batch.validate
* That creates a short-lived batch definition to pass to the validator
The short-lived batch definition is hidden from the user in the interactive flow, so it's not currently the end of the world, but it feels weird. Potential additional work: add batch_definition to the batch (feels bad), or make validator take in batches. But I don't think it's worth doing that right now without really thinking about the model.

### DIFF
--- a/great_expectations/datasource/fluent/interfaces.py
+++ b/great_expectations/datasource/fluent/interfaces.py
@@ -1084,11 +1084,13 @@ class Batch:
                 "We can't validate batches that are attached to datasources without a data context"
             )
 
+        # note: batch definition is created but NOT added to the asset, as it should not persist
         batch_definition = BatchDefinition(
             name="-".join([self.datasource.name, self.data_asset.name, str(uuid.uuid4())]),
             partitioner=self.batch_request.partitioner,
-            batching_regex=self.batch_request.batching_regex,
         )
+        batch_definition.set_data_asset(self.data_asset)
+
         return V1Validator(
             batch_definition=batch_definition,
             batch_parameters=self.batch_request.options,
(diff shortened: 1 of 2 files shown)


tests/datasource/fluent/test_batch.py
```diff
@@ -156,3 +160,46 @@ def test_batch_validate_change_expectation_suite_result_format(
     assert boolean_result.success is True
     assert len(boolean_result.results) == 1
     assert len(boolean_result.results[0].result) == 0
+
+
+@pytest.mark.filesystem
```
tyler-hoffman, line 165: These tests were nice to have for the red/green cycle, but I'm not sure that tests asserting a lack of side effects are terribly meaningful. Reviewer: WDYT?

joshua-stauffer: since they're already written, i think its probably worthwhile to keep them - they do capture behavior we intend, and helps guard against someone accidentally changing the code to persist these in the future

joshua-stauffer (approved):
looks good, thanks!

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -1083,9 +1083,14 @@ def _validator(self) -> V1Validator:
             raise ValueError(  # noqa: TRY003
                 "We can't validate batches that are attached to datasources without a data context"
             )
-        batch_definition = self.data_asset.add_batch_definition(
-            name="-".join([self.datasource.name, self.data_asset.name, str(uuid.uuid4())])
+
+        # note: batch definition is created but NOT added to the asset, as it should not persist
+        batch_definition = BatchDefinition(
```
billdirks, line 1101: While it might feel strange to create a BatchDefinition outside of the fluent workflow, I think it demonstrates that this decoupling was a good way to implement this.

billdirks (approved):
LGTM!

PR #9817, opened 2024-04-25
[MAINTENANCE] Prework for implementing fluent batch definition api for file path assets

joshua-stauffer:
This PR adds core logic required for adding BatchDefinitions from file path assets.


great_expectations/datasource/fluent/data_asset/data_connector/regex_parser.py
```diff
@@ -16,55 +16,34 @@ def __init__(
...
-
-        return matches.groupdict()
-
-    def get_all_matched_group_values(self, target: str) -> List[str] | None:
-        # Check for `(?P<name>)` named group syntax
-        matches: Optional[Match[str]] = self.get_matches(target=target)
-        if matches is None:
-            return None
-
-        return list(matches.groups())
```
joshua-stauffer, line 52: unused code

great_expectations/datasource/fluent/data_asset/data_connector/regex_parser.py
```diff
@@ -120,38 +99,9 @@ def get_all_group_index_to_group_name_mapping(self) -> Dict[int, str]:
...
-    def get_group_index_to_group_value_mapping(
-        self,
-        target: str,
-    ) -> Dict[int, str]:
-        all_group_indexes: List[int] = self.get_all_group_indexes()
-        all_matched_group_values: List[str] = self.get_all_matched_group_values(target=target) or []
-        group_index_to_group_value_mapping: Dict[int, str] = dict(
-            zip(all_group_indexes, all_matched_group_values)
-        )
-        return group_index_to_group_value_mapping
```
joshua-stauffer, line 157: unused code

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
```diff
@@ -149,15 +149,15 @@ def get_data_reference_count(self) -> int:
 
     # Interface Method
     @override
-    def get_matched_data_references(self) -> List[str]:
+    def get_matched_data_references(self, regex: re.Pattern | None = None) -> List[str]:
```
joshua-stauffer, line 152: this change is required to test if a path exists using our current structure. In the world of BatchDefinitions it probably makes sense to update the API across the board to require a batch request here, but i wanted to minimize scope for this PR.

tests/datasource/fluent/data_asset/data_connector/test_regex_parser.py
```diff
@@ -104,18 +88,12 @@ def test_regex_pattern_first_named_group_second_common_group(
...
         regex_pattern=regex_pattern, unnamed_regex_group_prefix="batch_request_param_"
     )
 
     assert regex_parser.get_num_all_matched_group_values() == 2
-    assert regex_parser.get_num_named_matched_group_values() == 1
     assert regex_parser.get_named_group_name_to_group_index_mapping() == {"year": 1}
-    assert regex_parser.get_named_group_name_to_group_value_mapping(target=target) == {
-        "year": "2020"
-    }
-    assert regex_parser.get_all_matched_group_values(target=target) == ["2020", "03"]
```
joshua-stauffer, line 118: diff in this file strictly removes tests which exercised unused code paths.

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
billdirks: nit: maybe add an `Args` block. We don't do that consistently and this is pretty understandable so wanted to mention it but it is fine not to.

Why do we need to allow overriding `self._batching_regex`? That is, why isn't it already associated with the connector?

great_expectations/datasource/fluent/data_asset/data_connector/regex_parser.py
billdirks: Thanks for deleting and all the better names!

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
```diff
@@ -243,7 +243,7 @@ def _get_whole_directory_batch_definition_list(
         )
         return [batch_definition]
 
-    def _get_data_references(self, matched: bool) -> List[str]:
+    def _get_data_references(self, matched: bool, regex: re.Pattern | None = None) -> List[str]:
```
billdirks, line 246: I have a very small preference to make `regex` required and update the 1 call site where you don't pass in `regex` to pass in `self._batching_regex`.

great_expectations/datasource/fluent/file_path_data_asset.py
```diff
@@ -188,6 +228,102 @@ def build_batch_request(
...
+            raise AmbiguousPathError(path=path)
+        return self.add_batch_definition(
+            name=name,
+            partitioner=PartitionerPath(regex=regex),
+        )
+
+    def _add_batch_definition_yearly(self, name: str, regex: re.Pattern) -> BatchDefinition:
+        """Adds a BatchDefinition which defines yearly batches by file name.
+
+        Parameters:
```
billdirks, line 257: nit: Google style guide uses `Args` instead of `Parameters`: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

tests/datasource/fluent/data_asset/test_file_path_asset.py
```diff
@@ -87,3 +102,319 @@ def test_get_batch_list_from_batch_request__sort_descending(
...
+
+
+@pytest.fixture
+def file_path_data_connector(mocker):
+    return mocker.Mock(spec=FilePathDataConnector)
+
+
+@pytest.fixture
+def asset(request, datasource, file_path_data_connector) -> _FilePathDataAsset:
+    # asset = _FilePathDataAsset(name="test_asset", type="_sql_asset")
```
billdirks, line 119: Is this the only asset used now? It is strange that it's type is `_sql_asset` but it's a file path asset.

tests/datasource/fluent/data_asset/test_file_path_asset.py
```diff
@@ -87,3 +102,319 @@ def test_get_batch_list_from_batch_request__sort_descending(
...
+    return asset
+
+
+PATH_NAME = "data_2022-01.csv"
+
+
+def _asset_parameters():
+    return [
+        pytest.param(
+            _FilePathDataAsset(name="test_asset", type="_sql_asset"), id="FilePathDataAsset"
```
billdirks, line 135: Should this be a `_sql_asset` or a csv asset?

tests/datasource/fluent/data_asset/test_file_path_asset.py
```diff
@@ -87,3 +102,319 @@ def test_get_batch_list_from_batch_request__sort_descending(
...
+
+@pytest.mark.unit
+@pytest.mark.parametrize(
+    "path",
+    [
+        pytest.param(PATH_NAME, id="String Path"),
+        pytest.param(pathlib.Path(PATH_NAME), id="Pathlib Path"),
+    ],
+)
+@pytest.mark.parametrize("asset", _asset_parameters(), indirect=["asset"])
```
billdirks, line 148: I haven't seen this indirect parameterizing before. Interesting!

tests/datasource/fluent/data_asset/test_file_path_asset.py
```diff
@@ -87,3 +102,319 @@ def test_get_batch_list_from_batch_request__sort_descending(
...
+    name = "batch_def_name"
+    expected_regex = re.compile(str(path))
+    expected_batch_definition = BatchDefinition(
+        name=name, partitioner=PartitionerPath(regex=expected_regex)
+    )
+    datasource.add_batch_definition.return_value = expected_batch_definition
+    file_path_data_connector.get_matched_data_references.return_value = [PATH_NAME]
+
+    # act
+    batch_definition = asset._add_batch_definition_path(name=name, path=path)
```
billdirks, line 165: Why call the `_` prefix method instead of one without? That is, how do users call it?

tests/datasource/fluent/data_asset/test_file_path_asset.py
```diff
@@ -87,3 +102,319 @@ def test_get_batch_list_from_batch_request__sort_descending(
...
+
+    # assert
+    assert batch_definition == expected_batch_definition
+    datasource.add_batch_definition.assert_called_once_with(expected_batch_definition)
```
billdirks, line 169: We check this because we've mocked out the datasource? It seems like it would be testing behavior more/implementation less if we could retrieve the batch definitions from the asset and see it's what we want. I'm ok with this.

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
billdirks: I see below that you are doing an existence test with a passed in regex. I don't know the flow but that use case makes sense.

billdirks (approved):
Thanks for doing this!

I've left a number of small comments. My main curiosity is in tests asking about how users call `_add_batch_definition_path`. That is, is there an `add_batch_definition_path` we should using.

great_expectations/datasource/fluent/data_asset/data_connector/file_path_data_connector.py
joshua-stauffer: the new pattern is that the `DataConnector` gets the regex from the batch request. We haven't removed `batching_regex` from the asset init signature yet (which passes it through the DataConnector init signature as well), so we prefer the regex via args if possible, but if it isn't provided then we're in a legacy workflow and we default to the regex saved on the class. The class level regex should be removed soon!

great_expectations/datasource/fluent/file_path_data_asset.py
joshua-stauffer: oh great! thanks for flagging - i'll update in my followup PR

tests/datasource/fluent/data_asset/test_file_path_asset.py
joshua-stauffer: this is just laying the groundwork for adding the fluent API - the PR was getting kind of large but i wanted to include the cleanup work, so decided to post it as is. The user facing API will be implemented in child classes and call into these methods.

tests/datasource/fluent/data_asset/test_file_path_asset.py
joshua-stauffer: This asset will be replaced by a list of all the  child assets the API is implemented on - just have this here as a placeholder for now.

tests/datasource/fluent/data_asset/test_file_path_asset.py
joshua-stauffer: yeah, the `add_batch_definition` api is well-tested, so i just wanted to keep the scope of the test lean and just make sure we're calling into it correctly.

PR #9823, opened 2024-04-26
[FEATURE] Add fluent-style BatchDefinition API to file-backed DataAssets

joshua-stauffer:
This PR adds a fluent-style BatchDefinition API to file-backed DataAssets which can be batched by applying a regex to their path. It excludes Spark file-backed DataAssets which batch by an entire directory, such as `DirectoryCSVAsset`, `DirectoryParquetAsset`, etc.

```Python
class FilePathDataAsset:
    def add_batch_definition_path(self, name: str, path: PathStr) -> BatchDefinition: ...

    def add_batch_definition_yearly(
        self, name: str, regex: re.Pattern, sort_ascending: bool = True
    ) -> BatchDefinition: ...

    def add_batch_definition_monthly(
        self, name: str, regex: re.Pattern, sort_ascending: bool = True
    ) -> BatchDefinition: ...

    def add_batch_definition_daily(
        self, name: str, regex: re.Pattern, sort_ascending: bool = True
    ) -> BatchDefinition: ...


```


great_expectations/datasource/fluent/schemas/SparkAzureBlobStorageDatasource/CSVAsset.json
```diff
@@ -1,6 +1,6 @@
 {
     "title": "CSVAsset",
-    "description": "Base model for most fluent datasource related pydantic models.\n\nAdds yaml dumping and parsing methods.\n\nExtra fields are not allowed.\n\nSerialization methods default to `exclude_unset = True` to prevent serializing\nconfigs full of mostly unset default values.\nAlso prevents passing along unset kwargs to BatchSpec.\nhttps://docs.pydantic.dev/usage/exporting_models/",
+    "description": "Mixin encapsulating logic specific to regex-based FilePathDataAssets",
```
joshua-stauffer, line 3: these doc string changes are unfortunate, so i'll look into fixing it. they weren't correct in the first place, so at least we're not losing any data.

great_expectations/datasource/fluent/spark_file_path_datasource.py
```diff
@@ -328,8 +333,12 @@ def _get_reader_options_include(self) -> set[str]:
...
 
 
-class DirectoryORCAsset(_DirectoryDataAssetMixin, ORCAsset):
-    type: Literal["directory_orc"] = "directory_orc"  # type: ignore[assignment]
+class ORCAsset(_RegexDataAssetMixin, ORCAssetBase):
+    type: Literal["orc"] = "orc"
+
+
+class DirectoryORCAsset(_DirectoryDataAssetMixin, ORCAssetBase):
+    type: Literal["directory_orc"] = "directory_orc"
```
tyler-hoffman, line 341: non blocking, but I feel like we should consider not following the convention of `<X>Asset` and `Directory<X>Asset`, but instead do like `File<X>Asset` and `Directory<X>Asset`

tyler-hoffman (approved):

PR #9828, opened 2024-04-26
[MAINTENANCE] Update Pandas Types

joshua-stauffer:
This PR updates the Pandas type stub so that static type checkers can correctly discover new BatchDefinition APIs.


tyler-hoffman (approved):
We should add a type to the asset param of `test_add_batch_definition_fluent_file_path__add_batch_definition_monthly_success` to make mypy prove this works, but I've confirmed on my machine this is good.

PR #9832, opened 2024-04-29
[MAINTENANCE] Enable SIM211

anthonyburdi:
https://docs.astral.sh/ruff/rules/if-expr-with-false-true/

# if-expr-with-false-true (SIM211)

Fix is always available.

## What it does
Checks for `if` expressions that can be replaced by negating a given
condition.

## Why is this bad?
`if` expressions that evaluate to `False` for a truthy condition and `True`
for a falsey condition can be replaced with `not` operators, which are more
concise and readable.

## Example
```python
False if a else True
```

Use instead:
```python
not a
```

## References
- [Python documentation: Truth Value Testing](https://docs.python.org/3/library/stdtypes.html#truth-value-testing)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
       - id: detect-private-key
         exclude: tests/test_fixtures/database_key_test*
   - repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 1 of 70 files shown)


billdirks (approved):
LGTM

Kilo59 (approved):

PR #9835, opened 2024-04-29
[MAINTENANCE] Enable SIM201

anthonyburdi:
https://docs.astral.sh/ruff/rules/negate-equal-op/

# negate-equal-op (SIM201)

Fix is always available.

## What it does
Checks for negated `==` operators.

## Why is this bad?
Negated `==` operators are less readable than `!=` operators. When testing
for non-equality, it is more common to use `!=` than `==`.

## Example
```python
not a == b
```

Use instead:
```python
a != b
```

## References
- [Python documentation: Comparisons](https://docs.python.org/3/reference/expressions.html#comparisons)


NathanFarmer (approved):

billdirks (approved):
Thanks!

PR #9836, opened 2024-04-29
[MAINTENANCE] Enable SIM110

anthonyburdi:
https://docs.astral.sh/ruff/rules/reimplemented-builtin/

# reimplemented-builtin (SIM110)

Fix is sometimes available.

## What it does
Checks for `for` loops that can be replaced with a builtin function, like
`any` or `all`.

## Why is this bad?
Using a builtin function is more concise and readable. Builtins are also
more efficient than `for` loops.

## Example
```python
for item in iterable:
    if predicate(item):
        return True
return False
```

Use instead:
```python
return any(predicate(item) for item in iterable)
```

## References
- [Python documentation: `any`](https://docs.python.org/3/library/functions.html#any)
- [Python documentation: `all`](https://docs.python.org/3/library/functions.html#all)

### DIFF
--- a/great_expectations/datasource/fluent/databricks_sql_datasource.py
+++ b/great_expectations/datasource/fluent/databricks_sql_datasource.py
@@ -171,10 +171,7 @@ class DatabricksTableAsset(SqlTableAsset):
             True if the target string is bracketed by quotes.
         """
         # TODO: what todo with regular quotes? Error? Warn? "Fix"?
-        for quote in ["`"]:
-            if target.startswith(quote) and target.endswith(quote):
-                return True
-        return False
+        return any(target.startswith(quote) and target.endswith(quote) for quote in ["`"])
 
 
 @public_api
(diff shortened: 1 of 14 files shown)


great_expectations/datasource/fluent/databricks_sql_datasource.py
```diff
@@ -171,10 +171,7 @@ def _is_bracketed_by_quotes(target: str) -> bool:
             True if the target string is bracketed by quotes.
         """
         # TODO: what todo with regular quotes? Error? Warn? "Fix"?
-        for quote in ["`"]:
-            if target.startswith(quote) and target.endswith(quote):
-                return True
-        return False
+        return any(target.startswith(quote) and target.endswith(quote) for quote in ["`"])
```
billdirks: Remove loop altogether.

billdirks (approved):
LGTM!

Kilo59 (approved):

PR #9863, opened 2024-05-01
[BUGFIX] Set application query param for SnowflakeDatasource

roblim:
- Set application query param for SnowflakeDatasource when creating engine. This is to ensure GX gets attribution for the Snowflake Partner program
- This was attempted previously by setting the `SF_PARTNER` env var, which would typically be picked up by the Snowflake connecter to set the application query param. However, since GX uses snowflake-sqlalchemy to connect to Snowflake, the env var gets superseded, with attribution going to snowflake-sqlalchemy. 

### DIFF
--- a/great_expectations/datasource/fluent/snowflake_datasource.py
+++ b/great_expectations/datasource/fluent/snowflake_datasource.py
@@ -130,23 +130,22 @@ class SnowflakeDatasource(SQLDatasource):
             values["connection_string"] = connection_details
         return values
 
-    @pydantic.root_validator
-    def _check_xor_input_args(cls, values: dict) -> dict:
+    @pydantic.validator("connection_string")
+    def _check_connection_string(
+        cls, connection_string: ConnectionDetails | ConfigStr | SnowflakeDsn
+    ) -> ConnectionDetails | ConfigStr | SnowflakeDsn:
         # keeping this validator isn't strictly necessary, but it provides a better error message
-        connection_string: str | ConnectionDetails | None = values.get(
-            "connection_string"
-        )
-        if connection_string:
-            # Method 1 - connection string
-            if isinstance(connection_string, (str, ConfigStr)):
-                return values
-            # Method 2 - individual args (account, user, and password are bare minimum)
-            elif isinstance(connection_string, ConnectionDetails) and bool(
-                connection_string.account
-                and connection_string.user
-                and connection_string.password
-            ):
-                return values
+
+        # Method 1 - connection string
+        if isinstance(connection_string, (str, ConfigStr)):
+            return connection_string
+        # Method 2 - individual args (account, user, and password are bare minimum)
+        elif isinstance(connection_string, ConnectionDetails) and bool(
+            connection_string.account
+            and connection_string.user
+            and connection_string.password
+        ):
+            return connection_string
         raise ValueError(
             "Must provide either a connection string or a combination of account, user, and password."
         )
--- a/great_expectations/datasource/fluent/snowflake_datasource.py
+++ b/great_expectations/datasource/fluent/snowflake_datasource.py
@@ -129,7 +129,7 @@ class SnowflakeDatasource(SQLDatasource):
 
         if values.get("connection_string") and connection_details:
             raise ValueError(
-                "Cannot provide both a connection string or a combination of account, user, and password."
+                "Cannot provide both a connection string and a combination of account, user, and password."
             )
 
         if connection_details:
--- a/great_expectations/datasource/fluent/snowflake_datasource.py
+++ b/great_expectations/datasource/fluent/snowflake_datasource.py
@@ -213,11 +213,14 @@ class SnowflakeDatasource(SQLDatasource):
         return gx_exec_engine
 
     def _get_snowflake_partner_application(self) -> str:
+        """
+        This is used to set the application query parameter in the Snowflake connection URL, which provides attribution
+        to GX for the Snowflake Partner program.
+        """
+
         # This import is here to avoid a circular import
         from great_expectations.data_context import CloudDataContext
 
-        # This is used to set the application query parameter, which provides attribution to GX
-        # for the Snowflake Partner program.
         if isinstance(self._data_context, CloudDataContext):
             # TODO: Change this value to "great_expectations_cloud" once configured in Snowflake partner account
             return SNOWFLAKE_PARTNER_APPLICATION_CLOUD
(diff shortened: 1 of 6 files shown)


great_expectations/datasource/fluent/snowflake_datasource.py
```diff
@@ -218,10 +229,14 @@ def get_engine(self) -> sqlalchemy.Engine:
                 connection_string: str | dict = model_dict.pop("connection_string")
 
                 if isinstance(connection_string, str):
-                    self._engine = sa.create_engine(connection_string, **kwargs)
+                    self._engine = sa.create_engine(
+                        connection_string,
+                        connect_args={"application": snowflake_partner_application},
```
jcampbell: Just eyeballing this: does this conflict if the user has `connect_args` in their kwargs?

roblim: Ah great catch! Yeah, would probably break. Will update to merge if present.

roblim: Or actually, just noticed a test that suggests providing both str and connect args should yield ValueError. Lemme dig deeper and confirm expected behavior.

tests/datasource/fluent/test_snowflake_datasource.py
```diff
@@ -100,7 +100,9 @@ def test_conflicting_connection_string_and_args_raises_error(
     connection_string: ConfigStr | SnowflakeDsn | None | dict, connect_args: dict
 ):
     with pytest.raises(ValueError):
-        _ = SnowflakeDatasource(connection_string=connection_string, **connect_args)
+        _ = SnowflakeDatasource(
+            name="my_sf_ds", connection_string=connection_string, **connect_args
+        )
```
roblim, line 177: This test wasn't working properly. It was passing only because same error was being thrown for missing name, and when this was fixed, passing because improper url was also throwing same error. 

When test was fixed, it revealed that validation of connection_string XOR connection args wasn't happening.

great_expectations/datasource/fluent/snowflake_datasource.py
```diff
@@ -126,27 +126,32 @@ def _convert_root_connection_detail_fields(cls, values: dict) -> dict:
         for field_name in tuple(values.keys()):
             if field_name in connection_detail_fields:
                 connection_details[field_name] = values.pop(field_name)
+
+        if values.get("connection_string") and connection_details:
+            raise ValueError(
+                "Cannot provide both a connection string or a combination of account, user, and password."
+            )
+
```
roblim: `test_conflicting_connection_string_and_args_raises_error` was broken. After fixing, found that XOR check wasn't happening, so added this.

@Kilo59 - I assume we still wanted this XOR validation since we still had the above test, and there was a validator called `_check_xor_input_args` (but XOR check wasn't actually happening). LMK if you meant to remove XOR condition.

great_expectations/datasource/fluent/snowflake_datasource.py
```diff
@@ -126,27 +126,32 @@ def _convert_root_connection_detail_fields(cls, values: dict) -> dict:
...
         if connection_details:
             values["connection_string"] = connection_details
         return values
 
-    @pydantic.root_validator
-    def _check_xor_input_args(cls, values: dict) -> dict:
+    @pydantic.validator("connection_string")
+    def _check_connection_string(
```
roblim: Renamed since it wasn't actually doing XOR check and was just validating connection_string.

great_expectations/datasource/fluent/snowflake_datasource.py
roblim: Changed impl to add query param to url instead of using connect_args, mostly to allow for easier testing (once an engine is instantiated, there's no way to really check connect_args without actually connecting to Snowflake).

great_expectations/datasource/fluent/snowflake_datasource.py
```diff
@@ -218,10 +234,19 @@ def get_engine(self) -> sqlalchemy.Engine:
...
+                    self._engine = sa.create_engine(
+                        url,
+                        **kwargs,
+                    )
                 else:
                     self._engine = self._build_engine_with_connect_args(
-                        **connection_string
+                        application=snowflake_partner_application,
+                        **connection_string,
+                        **kwargs,
```
roblim, line 258: @Kilo59 - I noticed **kwargs was missing here. Assumed we still wanted it since we're still passing **kwargs to engine for str connection_string, and we were passing **kwargs here before. Is this correct?

Kilo59: I think the kwargs are passed elsewhere.

roblim: I thought they might be too, but they end up getting filtered out in `_get_connect_args`, since `kwargs` is defined in parent class:

```
    def _get_connect_args(self) -> dict[str, str | bool]:
        excluded_fields: set[str] = set(SQLDatasource.__fields__.keys())
        # dump as json dict to force serialization of things like AnyUrl
        return self._json_dict(exclude=excluded_fields, exclude_none=True)
```

great_expectations/datasource/fluent/snowflake_datasource.py
```diff
@@ -204,6 +210,17 @@ def get_execution_engine(self) -> SqlAlchemyExecutionEngine:
...
+        # This import is here to avoid a circular import
+        from great_expectations.data_context import CloudDataContext
+
+        # This is used to set the application query parameter, which provides attribution to GX
+        # for the Snowflake Partner program.
+        if isinstance(self._data_context, CloudDataContext):
+            # TODO: Change this value to ""great_expectations_cloud" once configured in Snowflake partner account
+            snowflake_partner_application = "great_expectations_oss"
+        else:
+            snowflake_partner_application = "great_expectations_oss"
```
cdkini: Can we make these constants?

cdkini: And is there a way to decouple Cloud logic here? I don't love leaking this logic here but we might need to given our current architecture

roblim: Sure. Added constants in https://github.com/great-expectations/great_expectations/pull/9863/commits/3f3dd3c4885bec2d3f2b8b55b4619129dc66b346. 

WRT second point, I don't think there's a cleaner way to decouple cloud logic, but I did factor it out into a separate util method so at least it's less exposed in the public method. https://github.com/great-expectations/great_expectations/pull/9863/commits/10066e7ba7b7b543a32bd7dfcb83a2471dc0242f

cdkini (approved):

PR #9872, opened 2024-05-02
[BUGFIX] fix sqlalchemy import

joshua-stauffer:
Due to [a recent change](https://github.com/great-expectations/great_expectations/pull/9781), `convert_to_json_serializable` errors if sqlalchemy is not installed.


great_expectations/core/util.py
```diff
@@ -64,6 +64,9 @@
 if not LegacyRow:
     LegacyRow = SQLALCHEMY_NOT_IMPORTED
 
+if not Row:
```
billdirks, line 67: This and the above `LegacyRow` conditional should not be necessary, [the compatibility file already does this](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/compatibility/sqlalchemy.py#L73) . `SQLALCHEMY_NOT_IMPORTED` evaluates to `False` so this should just be overwriting `Row` with the same thing.

Could you remove this and try with just your change below? You could also verify this here by checking if `Row` is already equal to `SQLALCHEMY_NOT_IMPORTED`

joshua-stauffer: that makes sense to me, but unfortunately it isn't the case. If we remove those null checks `SQLALCHEMY_NOT_IMPORTED` raises a `ModuleNotFoundError` because something is calling `__getattr__` on `SQLALCHEMY_NOT_IMPORTED`. Seems like something isn't quite working as expected, but not clear what, so i'll defer looking further and go with the established pattern.

tyler-hoffman (approved):

PR #9875, opened 2024-05-03
[MAINTENANCE] `ExpectationSuite` API cleanup

cdkini:
Delete any extraneous methods we don't need (citations, filtering, etc). Also, make any methods private if we can.


tests/core/test_expectation_suite_crud_methods.py
```diff
@@ -379,7 +379,14 @@ def test_add_expectation_configurations(
...
 
-    assert single_expectation_suite.isEquivalentTo(different_suite)
+    assert len(single_expectation_suite.expectations) == len(different_suite.expectations)
+    assert all(
+        mine.isEquivalentTo(theirs)
+        for (mine, theirs) in zip(
+            single_expectation_suite.expectation_configurations,
+            different_suite.expectation_configurations,
+        )
+    )
```
cdkini, line 324: Took the guts of the `isEquivalentTo` method and used it for the assertion here

great_expectations/core/expectation_suite.py
```diff
@@ -507,26 +389,7 @@ def remove_expectation(
...
-        self.expectations = [
-            expectation
-            for expectation in self.expectations
-            if expectation.expectation_type not in expectation_types
-        ]
-
-        return removed_expectations
-
-    def find_expectation_indexes(
+    def _find_expectation_indexes(
```
cdkini, line 392: This is used by other methods so it has to stay around but I don't think it needs to be public

great_expectations/core/expectation_suite.py
```diff
@@ -273,84 +264,6 @@ def expectation_configurations(self, value):
...
-            "batch_markers": batch_markers,
-            "batch_parameters": batch_parameters,
-            "profiler_config": profiler_config,
-            "comment": comment,
-        }
-        gx.util.filter_properties_dict(properties=citation, clean_falsy=True, inplace=True)
-        self.meta["citations"].append(citation)
-
-    # noinspection PyPep8Naming
-    def isEquivalentTo(self, other):
```
tyler-hoffman, line 320: I thought this one was too hard to pull out? Awesome if we can do it now!

great_expectations/core/expectation_suite.py
tyler-hoffman: I feel like we're misusing this. This method isn't private; it's deprecated. And it's already not in the public api. `typing-extensions` has a `deprecated` decorator. Thoughts on just using that?

tyler-hoffman (changes requested):
I want to at least discuss not using the psuedo-private `_` prefix for methods that are used outside of the class.

great_expectations/core/expectation_suite.py
tyler-hoffman: I suppose I care less about the distinction if the only external use case is in a test

tyler-hoffman (approved):

great_expectations/core/expectation_suite.py
cdkini: Just to close the loop on this, this method is only ever used within this class (at least after the other deletions in this file).

PR #9886, opened 2024-05-06
[MAINTENANCE] Remove URN support

tyler-hoffman:
## Overview

This removes anything urn-related, which trickled down into deleting a couple more methods that weren't obvious from their names / that they were urn-related. 

## Summary of non-obvious methods that were deleted (Please double check me!)
* suite.get_suite_parameter_dependencies: returned a dict built from configuration.get_suite_parameter_dependencies
* configuration.get_suite_parameter_dependencies : called find_suite_parameter_dependencies on all eval params, but then we only used urns from it
* find_suite_parameter_dependencies: only used by ^
* store_suite_parameters : only used by tests
_compile_suite_parameter_dependencies: used results of suite.get_suite_parameter_dependencies , which was just urn logic
* metric_store.get_bind_params: all keys were by urn 


great_expectations/core/suite_parameters.py
```diff
@@ -135,11 +133,8 @@ def get_parser(self):
             # or use provided pyparsing_common.number, but convert back to str:
             # fnumber = ppc.number().addParseAction(lambda t: str(t[0]))
             fnumber = Regex(r"[+-]?(?:\d+|\.\d+)(?:\.\d+)?(?:[eE][+-]?\d+)?")
-            ge_urn = Combine(
-                Literal("urn:great_expectations:") + Word(alphas, f"{alphanums}_$:?=%.&")
-            )
             variable = Word(alphas, f"{alphanums}_$")
-            ident = ge_urn | variable
+            ident = variable
```
cdkini: The naming here feels a little difficult to reason about - is there any way we can improve the readability? I realize this isn't a result of your PR so please feel free to ignore

tyler-hoffman: I think `ident` was supposed to be short for `identifier`. But no need to even keep it around. Just updated to use `variable` directly below. Thanks!

great_expectations/core/suite_parameters.py
```diff
@@ -413,30 +325,9 @@ def parse_suite_parameter(  # noqa: C901, PLR0912, PLR0915
 
     elif len(parse_results) == 0 or parse_results[0] != "Parse Failure":
         # we have a stack to evaluate and there was no parse failure.
-        # iterate through values and look for URNs pointing to a store:
         for i, ob in enumerate(EXPR.exprStack):
```
cdkini, line 328: Is this a `for/else` statement? Just confirming - is that intended or do we have a mis-indentation?

cdkini (approved):

great_expectations/core/suite_parameters.py
tyler-hoffman: Looks to me like this was not and is now not a `for/else` loop. So pretty sure this is as intended, and no change in behavior other than not doing url stuff. The `else` block looks to be matched in terms of logic with the `elif`, and also I'd expect to see a `break` in an `for/else`, otherwise the else isn't doing anything special. Going to merge, but if any of that rings false to you, LMK and I'll revert!

PR #9874, opened 2024-05-03
[FEATURE] Directory Asset BatchDefinition API

joshua-stauffer:
This PR is prework to add a fluent-style batch definition API to directory data assets. In order to support this, `_FilePathDataAsset` and direct descendents have been refactored in order to allow regex-based assets to use RegexPartitioners, and directory-based assets to use column partitioners. `SparkPartitioners` have been brought back as `DataframePartitioners`.

## other refactors
- `_FilePathDataAsset` has been renamed to `PathDataAsset`
- `great_expectations.datasource.fluent.data_asset.data_connector` package has been moved out of `data_asset` to the `fluent` package.
- concrete implementations of spark assets have been moved out of `spark_file_path_datasource.py` and into the `data_asset.spark` package
- concrete implementations of pandas assets have been moved out of `pandas_file_path_datasource.py` and into the `data_asset.pandas` package

PR #9888, opened 2024-05-07
[FEATURE] DirectoryAsset BatchDefinition API

joshua-stauffer:
This PR introduces a fluent-style BatchDefinition API to DirectoryAssets:
```Python
class DirectoryAsset:
    def add_batch_definition_daily(self, name: str, column: str) -> BatchDefinition: 
        ...

    def add_batch_definition_monthly(self, name: str, column: str) -> BatchDefinition:
        ...

    def add_batch_definition_yearly(self, name: str, column: str) -> BatchDefinition:
        ...

    def add_batch_definition_whole_directory(self, name: str) -> BatchDefinition:
        ...
```
Unlike other asset's BatchDefinition APIs, DirectoryAssets do not expose a sort parameter, since they require exact batch parameters and only ever return a single batch.
## Followup Work
This PR does not test that the column a user indicates exists; that work is captured in [V1-325](https://greatexpectations.atlassian.net/browse/V1-325).

## Pre Work
This PR was preceded by https://github.com/great-expectations/great_expectations/pull/9874, which refactored the inheritance hierarchy of path-based assets to allow adding this feature.

[V1-325]: https://greatexpectations.atlassian.net/browse/V1-325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


great_expectations/core/partitioners.py
```diff
@@ -36,6 +39,7 @@ class PartitionerDatetimePart(pydantic.BaseModel):
     column_name: str
     sort_ascending: bool = True
     method_name: Literal["partition_on_date_parts"] = "partition_on_date_parts"
+    param_names: List[str] = []
```
joshua-stauffer: these Partitioners are not supported will be removed soon; this attribute is a placeholder for mypy.

joshua-stauffer: found a cleaner solution that doesn't require changing these classes [d2cd4ae](https://github.com/great-expectations/great_expectations/pull/9888/commits/d2cd4aebb46922f7b25dbb8a4a2926b9289bbf36)

tests/datasource/fluent/test_batch_request.py
```diff
@@ -78,7 +77,7 @@ def test_batch_request_config_serialization_round_trips(
     batch_request_config: dict[str, Any] = {
         "datasource_name": datasource_name,
         "data_asset_name": data_asset_name,
-        "partitioner": PartitionerColumnValue(column_name="my_column"),
```
joshua-stauffer, line 81: unrelated to this PR, but we don't intend to support this going forward, so i replaced it with a time-based partitioner.

great_expectations/datasource/fluent/data_asset/path/directory_asset.py
```diff
@@ -66,10 +107,24 @@ def _get_batch_definition_list(
             )
         return batch_definition_list
 
-    @public_api
-    def add_batch_definition_whole_directory(self, name: str) -> BatchDefinition:
-        """Add a BatchDefinition which creates a single batch for the entire directory."""
-        return self.add_batch_definition(name=name, partitioner=None)
+    @singledispatchmethod
```
tyler-hoffman, line 110: TIL. This seems like it accomplishes the same as using `@overload`, but now we get really clean small methods, rather than one big one that handles all the cases?

great_expectations/datasource/fluent/data_asset/path/directory_asset.py
```diff
@@ -120,3 +175,54 @@ def build_batch_request(
...
+            ),
+        }
+
+        self._add_partitioner_batch_parameters(
+            batch_request=batch_request, parameters=batch_spec_options
+        )
+
+        return batch_spec_options
+
+    def _add_partitioner_batch_parameters(self, batch_request, parameters) -> dict:
```
tyler-hoffman: Can you add types to the params here? Also non-blocking, but pure functions tend to be simpler to reason about; I'd consider just returning the dict containing `partitioner_method` and `partitioner_kwargs` and merging in the caller.

great_expectations/datasource/fluent/data_asset/path/path_data_asset.py
```diff
@@ -239,3 +222,9 @@ def _get_reader_method(self) -> str:
     def _get_reader_options_include(self) -> set[str]:
         # subtypes control how reader options get serialized
         raise NotImplementedError
+
+    def _get_sortable_partitioner(
+        self, partitioner: Optional[PartitionerT]
+    ) -> Optional[PartitionerSortingProtocol]:
+        # allow subclasses to determine sorting configuration.
```
tyler-hoffman, line 230: Can this be an `abstractmethod`?

tests/datasource/fluent/test_spark_filesystem_datasource.py
```diff
@@ -1119,64 +1123,126 @@ def directory_asset(
...
 
 @pytest.fixture
-def column_value_partitioner():
-    return PartitionerColumnValue(column_name="passenger_count")
+def daily_partitioner():
+    return PartitionerYearAndMonthAndDay(column_name="pickup_datetime")
+
+
+@pytest.fixture
+def daily_batch_parameters_and_expected_result():
```
tyler-hoffman: I don't understand what these fixtures represent from looking at them. I'd probably rename something to the effect of `expected_result` -> `known_matching_row_count` or something along those lines. These tests that assert against real data are important, but it makes the reader have to guess where the vales came from. Or maybe even stop having them be fixtures, since it looks like only one of them is reused.

tests/datasource/fluent/test_spark_filesystem_datasource.py
```diff
@@ -1119,64 +1123,126 @@ def directory_asset(
...
+                    "path",
+                    "year",
+                ),
+                id="Yearly Partitioner",
+            ),
+            pytest.param(None, ("path",), id="No Partitioner"),
+        ],
+    )
+    def test_get_batch_parameters_keys_with_partitioner(
+        self, directory_asset, partitioner: Partitioner, expected_keys
```
tyler-hoffman: I miss these sometimes too, but probably good to get types on the test params.

tyler-hoffman (changes requested):
Nice! I left a couple smaller asks you can ignore if you disagree, but there's at least one method that is missing typing.

great_expectations/datasource/fluent/data_asset/path/directory_asset.py
joshua-stauffer: i like the syntax, but tend to avoid it because typing always gets wonky. not sure what is going on in the implementation but mypy has a very difficult time reasoning about what type is being returned. Also, it's not possible to use keyword args in the signature, which is unfortunate.

great_expectations/datasource/fluent/data_asset/path/directory_asset.py
joshua-stauffer: wild, great catch, i thought types were a hard requirement on new code. And love the suggestion of not mutating the dict in the helper method, updated in [c432636](https://github.com/great-expectations/great_expectations/pull/9888/commits/c4326368ae480abd7f6f10dee317d5edfa676000).

great_expectations/datasource/fluent/data_asset/path/path_data_asset.py
joshua-stauffer: In this case, yes ([e89cf57](https://github.com/great-expectations/great_expectations/pull/9888/commits/e89cf570efcd5e31783fa106f40a2bd71ea09345)) - but that makes me wonder why do we have a handful of other `NotImplemented` methods on this class, can we make them `abstractmethod`s as well? And the answer is no, we can't, because [Pandas dynamic assets](https://github.com/great-expectations/great_expectations/blob/86a58627cc09ec8ddde4028dd4d83024d82a8e25/great_expectations/datasource/fluent/data_asset/path/pandas/generated_assets.py#L35) are given the type `FileAsset`, but their implementations of these methods are dynamically generated, so mypy fails.

tests/datasource/fluent/test_spark_filesystem_datasource.py
joshua-stauffer: i'm all for in-lining fixtures, the more context in the test, the better [c89c7c6](https://github.com/great-expectations/great_expectations/pull/9888/commits/c89c7c660b3a23df35a9abfa70fdc0cf4affa177)

tests/datasource/fluent/test_spark_filesystem_datasource.py
joshua-stauffer: thanks! updated [c89c7c6](https://github.com/great-expectations/great_expectations/pull/9888/commits/c89c7c660b3a23df35a9abfa70fdc0cf4affa177)

great_expectations/datasource/fluent/data_asset/path/path_data_asset.py
tyler-hoffman: oh wow good to know, thanks for digging in!

tyler-hoffman (approved):

PR #9895, opened 2024-05-07
[MAINTENANCE] FileDataAsset BatchDefinition API accepts either `str` or `re.Pattern`

joshua-stauffer:
This PR updates the API on FileDataAssets to allow passing a string instead of a compiled regex as the `regex` parameter while building a BatchDefinition.


billdirks (approved):
LGTM, thanks!

tyler-hoffman (approved):
Thanks! This feels like a great QoL improvement!

PR #9898, opened 2024-05-07
[MAINTENANCE] Remove batching regex from FilePathDataConnector

joshua-stauffer:
This PR updates `FileDataAssets` and `DirectoryDataAssets` to not require a `batching_regex` parameter on instantiation.

## Changes
- batching_regex has been removed from the init signature of all Assets and DataConnectors

## Other things to note
- Implementation of `FilePathDataConnector` has been slightly refactored to make it clearer that its behavior differs significantly based on what sort of Asset it is serving.
- Some `FileDataAssets` did preprocessing of their batching regex inside their init method - that has now been moved to the `_preprocess_batching_regex` method.
- `test_connection` logic for `FileDataAsset` and `DirectoryDataAsset` is not that useful in V1, since the specific `batching_regex` used to identify files is no longer owned by the asset. The majority of those tests have been removed, and future work will add `test_connection` functionality to the `BatchDefinition`.
- several versions of `test_get_batch_list_from_fully_specified_batch_request` for specific datasource/assets were marked as xfail for reasons that proved not to be accurate. They've been removed, and we should verify that we have adequate test coverage for those code paths. They've all been annotated with comments in this PR.


tests/datasource/fluent/test_pandas_azure_blob_storage_datasource.py
```diff
@@ -369,99 +259,16 @@ def test_csv_asset_with_non_string_batching_regex_named_parameters(
...
-        "path": "alex_20200819_1300.csv",
-        "name": "alex",
-        "timestamp": "20200819",
-        "price": "1300",
-    }
-    assert batch.id == "pandas_abs_datasource-csv_asset-name_alex-timestamp_20200819-price_1300"
-
-    request = asset.build_batch_request({"name": "alex"})
-    batches = asset.get_batch_list_from_batch_request(request)
-    assert len(batches) == 2
```
joshua-stauffer, line 427: it looks like this test was never completed; i don't see any issue with "local credentials" as mentioned in the `pytest.mark.skip`, it just looks like the mock ABS backend was never finished/implemented correctly.

tests/datasource/fluent/test_pandas_google_cloud_storage_datasource.py
```diff
@@ -314,93 +206,12 @@ def test_csv_asset_with_non_string_batching_regex_named_parameters(
...
-        "name": "alex",
-        "timestamp": "20200819",
-        "price": "1300",
-    }
-    assert batch.id == "pandas_gcs_datasource-csv_asset-name_alex-timestamp_20200819-price_1300"
-
-    request = asset.build_batch_request({"name": "alex"})
-    batches = asset.get_batch_list_from_batch_request(request)
-    assert len(batches) == 2
-
```
joshua-stauffer, line 371: I didn't dig into this test as deeply as I did the pandas ABS test above, but `pytest.mark.skip` message is exactly the same, and the implementation looks almost identical, and it also doesn't pass, so it seems safe to assume that this test was also never completed.

tests/datasource/fluent/test_spark_azure_blob_storage_datasource.py
```diff
@@ -333,102 +243,16 @@ def test_csv_asset_with_non_string_batching_regex_named_parameters(
...
-        "name": "alex",
-        "timestamp": "20200819",
-        "price": "1300",
-        **asset_specified_metadata,
-    }
-    assert batch.id == "spark_abs_datasource-csv_asset-name_alex-timestamp_20200819-price_1300"
-
-    request = asset.build_batch_request({"name": "alex"})
-    batches = asset.get_batch_list_from_batch_request(request)
-    assert len(batches) == 2
```
joshua-stauffer, line 394: this test fails because the SparkExecutionEngine doesn't have the attribute that we're trying to patch:
```
>       monkeypatch.setattr(
            great_expectations.execution_engine.sparkdf_execution_engine.SparkDFExecutionEngine,
            "_instantiate_s3_client",
            instantiate_azure_client_spy,
            raising=True,
        )
E       AttributeError: <class 'great_expectations.execution_engine.sparkdf_execution_engine.SparkDFExecutionEngine'> has no attribute '_instantiate_s3_client'

test_spark_azure_blob_storage_datasource.py:354: AttributeError
```
As far as I can tell this test was never completed.

tests/datasource/fluent/test_spark_google_cloud_storage_datasource.py
```diff
@@ -311,96 +223,16 @@ def test_csv_asset_with_non_string_batching_regex_named_parameters(
...
-        "name": "alex",
-        "timestamp": "20200819",
-        "price": "1300",
-        **asset_specified_metadata,
-    }
-    assert batch.id == "spark_gcs_datasource-csv_asset-name_alex-timestamp_20200819-price_1300"
-
-    request = asset.build_batch_request({"name": "alex"})
-    batches = asset.get_batch_list_from_batch_request(request)
-    assert len(batches) == 2
```
joshua-stauffer, line 370: this test fails because the `SparkExecutionEngine` doesn't have the attribute that we're trying to patch:
```
>       monkeypatch.setattr(
            great_expectations.execution_engine.sparkdf_execution_engine.SparkDFExecutionEngine,
            "_instantiate_s3_client",
            instantiate_gcs_client_spy,
            raising=True,
        )
E       AttributeError: <class 'great_expectations.execution_engine.sparkdf_execution_engine.SparkDFExecutionEngine'> has no attribute '_instantiate_s3_client'
```
As far as I can tell this test was never completed.

joshua-stauffer: the execution engine also doesn't have `_instantiate_gcs_client`, which would make more sense than `_isntantiate_s3_client` as written.

tests/datasource/fluent/test_spark_s3_datasource.py
```diff
@@ -73,205 +65,68 @@ def spark_s3_datasource(s3_mock, s3_bucket: str) -> SparkS3Datasource:
...
-        "name": "alex",
-        "timestamp": "20200819",
-        "price": "1300",
-        **asset_specified_metadata,
-    }
-    assert batch.id == "spark_s3_datasource-csv_asset-name_alex-timestamp_20200819-price_1300"
 
-    request = asset.build_batch_request({"name": "alex"})
-    batches = asset.get_batch_list_from_batch_request(request)
-    assert len(batches) == 2
```
joshua-stauffer, line 239: this test fails with an opaque `Py4JJavaError` which makes me wonder if this datasource is actually functional:
```
        request = asset.build_batch_request({"name": "alex", "timestamp": "20200819", "price": "1300"})
>       batches = asset.get_batch_list_from_batch_request(request)

test_spark_s3_datasource.py:214: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../great_expectations/datasource/fluent/data_asset/path/path_data_asset.py:144: in get_batch_list_from_batch_request
    data, markers = execution_engine.get_batch_data_and_markers(batch_spec=batch_spec)
../../../great_expectations/execution_engine/sparkdf_execution_engine.py:515: in get_batch_data_and_markers
    batch_data = reader_fn(path)
../../../env/lib/python3.10/site-packages/pyspark/sql/readwriter.py:727: in csv
    return self._df(self._jreader.csv(self._spark._sc._jvm.PythonUtils.toSeq(path)))
../../../env/lib/python3.10/site-packages/py4j/java_gateway.py:1322: in __call__
    return_value = get_return_value(
../../../env/lib/python3.10/site-packages/pyspark/errors/exceptions/captured.py:169: in deco
    return f(*a, **kw)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

answer = 'xro29'
gateway_client = <py4j.clientserver.JavaClient object at 0x2898baef0>
target_id = 'o26', name = 'csv'

    def get_return_value(answer, gateway_client, target_id=None, name=None):
        """Converts an answer received from the Java gateway into a Python object.
    
        For example, string representation of integers are converted to Python
        integer, string representation of objects are converted to JavaObject
        instances, etc.
    
        :param answer: the string returned by the Java gateway
        :param gateway_client: the gateway client used to communicate with the Java
            Gateway. Only necessary if the answer is a reference (e.g., object,
            list, map)
        :param target_id: the name of the object from which the answer comes from
            (e.g., *object1* in `object1.hello()`). Optional.
        :param name: the name of the member from which the answer comes from
            (e.g., *hello* in `object1.hello()`). Optional.
        """
        if is_error(answer)[0]:
            if len(answer) > 1:
                type = answer[1]
                value = OUTPUT_CONVERTER[type](answer[2:], gateway_client)
                if answer[1] == REFERENCE_TYPE:
>                   raise Py4JJavaError(
                        "An error occurred while calling {0}{1}{2}.\n".
                        format(target_id, ".", name), value)
E                   py4j.protocol.Py4JJavaError: An error occurred while calling o26.csv.
E                   : java.lang.RuntimeException: java.lang.ClassNotFoundException: Class org.apache.hadoop.fs.s3a.S3AFileSystem not found
E                   	at org.apache.hadoop.conf.Configuration.getClass(Configuration.java:2688)
E                   	at org.apache.hadoop.fs.FileSystem.getFileSystemClass(FileSystem.java:3431)
E                   	at org.apache.hadoop.fs.FileSystem.createFileSystem(FileSystem.java:3466)
E                   	at org.apache.hadoop.fs.FileSystem.access$300(FileSystem.java:174)
E                   	at org.apache.hadoop.fs.FileSystem$Cache.getInternal(FileSystem.java:3574)
E                   	at org.apache.hadoop.fs.FileSystem$Cache.get(FileSystem.java:3521)
E                   	at org.apache.hadoop.fs.FileSystem.get(FileSystem.java:540)
E                   	at org.apache.hadoop.fs.Path.getFileSystem(Path.java:365)
E                   	at org.apache.spark.sql.execution.datasources.DataSource$.$anonfun$checkAndGlobPathIfNecessary$1(DataSource.scala:724)
E                   	at scala.collection.immutable.List.map(List.scala:293)
E                   	at org.apache.spark.sql.execution.datasources.DataSource$.checkAndGlobPathIfNecessary(DataSource.scala:722)
E                   	at org.apache.spark.sql.execution.datasources.DataSource.checkAndGlobPathIfNecessary(DataSource.scala:551)
E 
```

tests/datasource/fluent/test_pandas_azure_blob_storage_datasource.py
joshua-stauffer: it raises a `TestConnectionError` because the mock ABS client isn't returning anything:
```
        """Test the connection for the DataAsset.
    
        Raises:
            TestConnectionError: If the connection test fails.
        """
        try:
            if self._data_connector.test_connection():
                return None
        except Exception as e:
            raise TestConnectionError(  # noqa: TRY003
                f"Could not connect to asset using {type(self._data_connector).__name__}: Got {type(e).__name__}"  # noqa: E501
            ) from e
>       raise TestConnectionError(self._test_connection_error_message)
E       great_expectations.datasource.fluent.interfaces.TestConnectionError: No file in bucket "test_bucket" with prefix "" and recursive file discovery set to "False" matched regular expressions pattern "(?P<name>.+)_(?P<timestamp>.+)_(?P<price>\d{4})\.csv" using delimiter "/" for DataAsset "csv_asset".
```

great_expectations/datasource/fluent/data_connector/file_path_data_connector.py
```diff
@@ -212,7 +212,8 @@ def _get_unfiltered_batch_definition_list(
         batch_definition_set = set()
         # if the batch request hasn't specified a batching_regex, fallback to a default
         if batch_request.partitioner:
-            batching_regex = self._preprocess_batching_regex(batch_request.partitioner.regex)
+            # --- HEY JOSH! ---: Is this where we want to preprocess?
+            batching_regex = batch_request.partitioner.regex
```
cdkini: I've annotated every instance of `batching_regex` in this file (and removed all the preprocessing until we determine which ones are necessary). The issue we were seeing was that we'd accidentally preprocess too many times.

If we look at the usages of `batching_regex` within this file, it is evident that we may hit one, multiple, or none of these code paths based on usage. I don't think it is as simple as adding preprocessing to each instance.

I'll probably have a better idea after stepping away for a bit.

joshua-stauffer: resolved, thanks! 🥳

great_expectations/datasource/fluent/data_asset/path/file_asset.py
```diff
@@ -208,12 +191,21 @@ def _get_batch_definition_list(
         )
         return batch_definition_list
 
+    def _get_group_names(self, partitioner: Optional[FileNamePartitioner]) -> list[str]:
+        if not partitioner:
+            return []
+        regex_parser = RegExParser(
+            regex_pattern=partitioner.regex,
+            unnamed_regex_group_prefix=self._unnamed_regex_param_prefix,
```
billdirks, line 199: Do we need these `unamed_regex_param_prefix` anymore?

great_expectations/datasource/fluent/data_connector/file_path_data_connector.py
```diff
@@ -143,7 +136,10 @@ def build_batch_spec(self, batch_definition: LegacyBatchDefinition) -> PathBatch
     # Interface Method
     @override
     def get_data_reference_count(self) -> int:
-        data_references = self._get_data_references_cache(batching_regex=self._batching_regex)
+        # todo: in the world of BatchDefinition, this method must accept a BatchRequest.
```
billdirks, line 139: Can we file a ticket for this.

tests/datasource/fluent/test_spark_s3_datasource.py
billdirks: I believe one needs to install an AWS provided jar to get spark to be able to read from s3.

billdirks (approved):
I left a comment and a question but nothing that is blocking. Thanks!

PR #9906, opened 2024-05-09
[BUGFIX] fix snowflake schema introspection for `expect_column_values_to_be_in_type_list` and `expect_column_values_to_be_of_type`

dctalbot:


tests/expectations/core/test_expect_column_values_to_be_in_type_list.py
```diff
@@ -165,3 +202,73 @@ def test_expect_column_values_to_be_in_type_list_nullable_int():
...
+        (VARIANT(), ["VARIANT"], True, "VARIANT"),
+        # known name mismatches between snowflake and sqlalchemy
+        (TEXT(), ["VARCHAR"], False, "TEXT"),
+        (INTEGER(), ["INTEGER"], False, "NUMBER"),
+        (DECIMAL(), ["DECIMAL"], False, "NUMBER"),
+        # case-insensitive check, ignore precision
+        (BOOLEAN(), ["BoolEAN"], True, "BOOLEAN"),
+        (NUMBER(), ["number(42)"], True, "NUMBER"),
+        # unknown type
+        ("FOO", ["TEXT"], False, "FOO"),
```
NathanFarmer: ❤️

great_expectations/expectations/core/expect_column_values_to_be_of_type.py
```diff
@@ -362,6 +366,22 @@ def _validate_sqlalchemy(self, actual_column_type, expected_type, execution_engi
...
+                        expected_type
+                    )
+                    normalized_native_type_name = map_sfsqlalchemy_type_to_sf_type(
+                        normalized_sfsqlalchemy_type_name
+                    )
+                    return {
+                        "success": normalized_native_type_name
+                        == normalized_expected_type_name,
+                        "result": {"observed_value": normalized_native_type_name},
+                    }
```
NathanFarmer, line 383: I realize that you are following an existing pattern here, but I would have expected this logic to live at the metric-level (as opposed to the Expectation-level). New Expectations that leverage this metric will probably need to re-implement this logic (I don't see any reason why we would use the SQL Alchemy types in an Expectation).

NathanFarmer (changes requested):
If you moved the type logic into `great_expectations/expectations/metrics/table_metrics/table_column_types.py` that might fix your circular imports issue. It is not obvious to the unfamiliar, but `table.column_types` is a dependency metric of all Column Map Metrics (`ColumnValuesInTypeList` metric inherits from `ColumnMapMetricProvider`).

great_expectations/expectations/core/expect_column_values_to_be_of_type.py
dctalbot: Yes, I chose this approach because I was concerned about the potential downstream effects of modifying the metrics. The metrics utils also return SQLAlchemy type objects, but I want to work with named types as literal strings.

NathanFarmer: It is fair to think that changing the metric would be a breaking change for other Expectations. Especially since (for SQL) we are using the column metadata, and that metric is more widely used.

tests/expectations/core/test_expect_column_values_to_be_of_type.py
```diff
@@ -1,10 +1,48 @@
+from unittest.mock import Mock
+
 import pandas as pd
 import pytest
+from snowflake.sqlalchemy import (
```
NathanFarmer: I believe you should use the `great_expectations/compatibility/snowflake.py` import here.

tests/test_packaging.py
```diff
@@ -76,6 +76,7 @@ def test_requirements_files():
         | req_set_dict["requirements-dev-contrib.txt"]
         | req_set_dict["requirements-dev-lite.txt"]
         | req_set_dict["requirements-dev-api-docs-test.txt"]
+        | req_set_dict["requirements-dev-snowflake.txt"]
```
NathanFarmer: The only reason I think think of why Databricks would be failing is this new requirement. I don't think we should be changing this here.

NathanFarmer (approved):

reqs/requirements-dev-test.txt
```diff
@@ -2,3 +2,4 @@
 --requirement requirements-dev-lite.txt
 --requirement requirements-dev-contrib.txt
 --requirement requirements-dev-api-docs-test.txt
+--requirement requirements-dev-snowflake.txt
```
Kilo59:
```suggestion
```
I don't think we can do this as will cause all the tests to install the snowflake deps.

Kilo59 (approved):

PR #9908, opened 2024-05-09
[MAINTENANCE] Improve error message around instantiating and saving s…

tyler-hoffman:
…uites

## Main stuff
* Updates the error around saving a non-added suite to resolve https://greatexpectations.atlassian.net/browse/V1-240.
* Scope creep alert: also adds a test around, and cleans up, the exception that is raised when instantiating a suite with no context.

## Other stuff
* Updates `set_context` to allow `None` values. I think this is desirable to test specific cases like this, and it's not part of the public api, so it should only be used internally.
* There's a comment above the updated Exception about how we are clobbering errors from the store. I considered changing the logic so that we check in the suite if it has been persisted and raise there, but that introduces a race condition, and I think we mostly need to solve for the normal use case.

### DIFF
--- a/great_expectations/data_context/data_context/context_factory.py
+++ b/great_expectations/data_context/data_context/context_factory.py
@@ -17,6 +17,7 @@ from great_expectations._docs_decorators import public_api
 from great_expectations.exceptions import (
     GXCloudConfigurationError,
 )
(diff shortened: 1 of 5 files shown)


billdirks (approved):
LGTM

PR #9918, opened 2024-05-13
[MAINTENANCE] Ruff `0.4.4`

Kilo59:
Update version of `ruff`

1. https://github.com/astral-sh/ruff/releases/tag/v0.4.3
2. https://github.com/astral-sh/ruff/releases/tag/v0.4.4


tests/datasource/fluent/_fake_cloud_api.py
```diff
@@ -808,7 +808,8 @@ def gx_cloud_api_fake_ctx(
     LOGGER.info("Mocking the GX Cloud API")
 
     with responses.RequestsMock(
-        assert_all_requests_are_fired=assert_all_requests_are_fired
+        assert_all_requests_are_fired=assert_all_requests_are_fired,
+        passthru_prefixes=("/organizations",),
```
Kilo59, line 812:
```suggestion
        assert_all_requests_are_fired=assert_all_requests_are_fired
```

contrib/cli/requirements.txt
```diff
@@ -1,8 +1,8 @@
 Click>=7.1.2         # CLI tooling
 cookiecutter==2.1.1  # Project templating
-mypy==1.10.0            # Type checker
+mypy==1.10            # Type checker
```
Kilo59, line 3:
```suggestion
mypy==1.10.0            # Type checker
```

great_expectations/execution_engine/partition_and_sample/data_partitioner.py
```diff
@@ -73,7 +73,7 @@ def __hash__(self: PartitionerMethod):
         return hash(self.value)
 
 
-class DataPartitioner(abc.ABC):
+class DataPartitioner(abc.ABC):  # noqa: B024
```
Kilo59, line 76: I think the violation is understandable given we do need this class to be inherited and never instantied directly.

A `Protocal` with default methods COULD also work here (or having a distinct `Protocol` + a `DataPartitionerBase`), but these don't seem obviously better to me. 

I think this is the exception that proves the rule.

https://docs.astral.sh/ruff/rules/abstract-base-class-without-abstract-method/
# abstract-base-class-without-abstract-method (B024)

Derived from the **flake8-bugbear** linter.

## What it does
Checks for abstract classes without abstract methods.

## Why is this bad?
Abstract base classes are used to define interfaces. If an abstract base
class has no abstract methods, you may have forgotten to add an abstract
method to the class or omitted an `@abstractmethod` decorator.

If the class is _not_ meant to be used as an interface, consider removing
the `ABC` base class from the class definition.

## Example
```python
from abc import ABC


class Foo(ABC):
    def method(self):
        bar()
```

Use instead:
```python
from abc import ABC, abstractmethod


class Foo(ABC):
    @abstractmethod
    def method(self):
        bar()
```

## References
- [Python documentation: `abc`](https://docs.python.org/3/library/abc.html)

tyler-hoffman: I'm fine with it as is, even though ABCs without abstractmethods can be instantiated directly. We could also do something like suggested [here](https://stackoverflow.com/questions/7989042/preventing-a-class-from-direct-instantiation-in-python), but I think it should be apparent to readers that they shouldn't directly instantiate this, which is the most important thin IMO

tyler-hoffman (approved):

PR #9929, opened 2024-05-15
[MAINTENANCE] performance, avoid HTTP round trips 

dctalbot:
- Currently, when building the evaluation parameters, N round trips are made to the GX Cloud store for N Expectation Suites.
- This PR avoids this by implementing the `get_all` method on the ExpectationsStore's interface.
- I suggest using the "hide whitespace changes" option when reviewing the code diff

### DIFF
--- a/great_expectations/data_context/store/expectations_store.py
+++ b/great_expectations/data_context/store/expectations_store.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import random
 import uuid
-from typing import Dict
+from typing import TYPE_CHECKING, Dict
 
 import great_expectations.exceptions as gx_exceptions
 from great_expectations.compatibility.typing_extensions import override
@@ -23,6 +23,19 @@ from great_expectations.util import (
     filter_properties_dict,
     verify_dynamic_loading_support,
 )
+
+if TYPE_CHECKING:
+    from typing import Literal
+
+    from typing_extensions import TypedDict
+
+    class DataPayload(TypedDict):
+        id: str
+        attributes: dict
+        type: Literal["expectation_suite"]
+
+    class CloudResponsePayloadTD(TypedDict):
+        data: DataPayload | list[DataPayload]
 
 
 class ExpectationsStore(Store):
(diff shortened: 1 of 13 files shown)


great_expectations/data_context/store/expectations_store.py
```diff
@@ -300,3 +313,25 @@ def self_check(self, pretty_print):  # noqa: PLR0912
             print()
 
         return return_obj
+
+    @override
+    @staticmethod
+    def gx_cloud_response_json_to_object_collection(
```
dctalbot, line 319: see `great_expectations/data_context/store/datasource_store.py` for reference implementation

Kilo59 (commented):
Nice work! It's hard to overstate the impact this will have 😄 

Approach seems reasonable I would suggest using the `_fake_web_api` mock to assert that the expected number of calls are being made to `/expectations-suites` + `/expectations-suites?name=<SUITE-NAME>`

I think you'll need to take the PR off of draft mode to run the integration tests.

tyler-hoffman (commented):
Looks good to me. the sa `_get_all` is a mouthful, but looks consistent with the rest of the class. Happy to approve once tests for the other backends are ported over!

tests/datasource/fluent/test_contexts.py
```diff
@@ -514,5 +514,30 @@ def test_data_connectors_are_built_on_config_load(
...
+    validator = seeded_cloud_context.sources.pandas_default.read_csv(valid_file_path)
+    validator.expect_column_values_to_not_be_null("pickup_datetime")
+    validator.save_expectation_suite()
+    checkpoint = seeded_cloud_context.add_or_update_checkpoint(
+        name="my_quickstart_chekpoint",
+        validator=validator,
+    )
+
+    checkpoint.run()
+    # TODO assert GET expectation-suites called
```
Kilo59, line 539: Did you mean to do this? 

I'm working on this right now.

tyler-hoffman (approved):

tests/datasource/fluent/test_contexts.py
Kilo59: This is turning out to be much more of a pain to setup than I expected.
Let's do this as a followup.

Kilo59 (approved):

PR #9935, opened 2024-05-15
[MAINTENANCE] Ban future use of `convert_to_json_serializable`

Kilo59:
Use of `convert_to_json_serializable` is a sign that we aren't properly using pydantic models for objects that require json serialization.

Ban the import of the function and add `# noqa: TID25` for existing violations.

Also ban `ensure_json_serializable` (maybe we don't care as much about this one?)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -422,6 +422,7 @@ If you are working in a configuration file you may use the inline comment \
 "pydantic".msg = "Please do not import pydantic directly, import from great_expectations.compatibility.pydantic instead."
 "pkg_resources".msg = "pkg_resources module has been deprecated. Use importlib.resources or importlib.metada instead."
 "great_expectations.util.convert_to_json_serializable".msg = "Use pydantic for serialization."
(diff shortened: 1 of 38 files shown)


pyproject.toml
```diff
@@ -421,6 +421,8 @@ If you are working in a configuration file you may use the inline comment \
 # TODO: remove pydantic once our min version is pydantic v2
 "pydantic".msg = "Please do not import pydantic directly, import from great_expectations.compatibility.pydantic instead."
 "pkg_resources".msg = "pkg_resources module has been deprecated. Use importlib.resources or importlib.metada instead."
+"great_expectations.util.convert_to_json_serializable".msg = "Use pydantic for serialization."
+"great_expectations.util.ensure_json_serializable".msg = "Use pydantic for serialization."
```
Kilo59, line 425: Not sure if we care as much about `ensure_json_serializable` 🤷

jcampbell (approved):
Yes, this is what I had in mind!

PR #9941, opened 2024-05-17
[MAINTENANCE] Exclude patterns from codecov reports

Kilo59:
Ignore patterns in code that should don't require test coverage.

```
    # type-checking imports don't exist at runtime
    "if TYPE_CHECKING:",
    # Don't complain if tests don't hit NotImplementedErrors:
    "raise NotImplementedError",
    # Don't complain if non-runnable code isn't run:
    "if __name__ == .__main__.:",
    # Don't complain about abstract methods, they aren't run:
    "@(abc\\.)?abstractmethod",
```

https://coverage.readthedocs.io/en/7.5.1/config.html


tyler-hoffman (approved):
Thanks for keeping on top of this stuff!

PR #9950, opened 2024-05-21
[BUGFIX] Show FDS Asset name in DataDocs

### LINKED ISSUE #8790: data_asset_name can't be specified with fluent datasources for use in data docs
**Describe the bug**
When running a checkpoint and producing data docs from a batch request generated with the `build_batch_request()` method of a Fluent pandas filesystem datasource, the "Asset Name" column of the resulting data docs is never populated.

screenshot of data docs missing asset name

**To Reproduce**
[This repository contains a minimum reproduction](https://github.com/morphatic/gx-fluent-pandas-asset-name-bug-demo). 

* Here is [`great_expectations.yml`](https://github.com/morphatic/gx-fluent-pandas-asset-name-bug-demo/blob/main/gx/great_expectations.yml)
* Here are [the produced data docs online](https://morphatic.github.io/gx-fluent-pandas-asset-name-bug-demo/)

**Expected behavior**
The "Asset Name" column should be filled in the produced data docs.

**Possible Causes and Fixes**
There are two ways to approach this issue. It could be either or both of a problem in:

1. the API for the `BatchRequest` class implemented for Fluent datasources
2. how the default data docs renderer finds and processes metadata

When [I asked about this issue in the GX Slack channel](https://greatexpectationstalk.slack.com/archives/CUTCNHN82/p1695848627758059) it was suggested that I could specify a value for the `batch_spec_passthrough` parameter. I tried doing this in several ways:

```python
# I tried 2 ways of manually adding the property;
# Both resulted in: ValueError: "BatchRequest" object has no field "batch_spec_passthrough"
batch_request.batch_spec_passthrough = { "data_asset_name": "iris" }
batch_request["batch_spec_passthrough"] = { "data_asset_name": "people" }

# Tried adding `batch_spec_passthrough` during batch request creation
# Failed with: TypeError: _PandasDataAsset.build_batch_request() got an unexpected keyword "batch_spec_passthrough"
batch_request = asset.build_batch_request(batch_spec_passthrough = { ... })

# Tried adding it to the `validators` property when creating the checkpoint
# No error produced, but also didn't resolve the issue
checkpoint = context.add_or_update_checkpoint(
  name="demo_iris_expectations",
  run_name_template="%Y-%m-iris-demo",
  validations=[{
    "batch_request": batch_request,
    "expectation_suite_name": "iris_expectations",
    "batch_spec_passthrough": { "data_asset_name": "people" }
  }]
)
```
**Inconsistent API for `BatchRequest` classesses**
This led me to realize that the API for the `BatchRequest` class is NOT consistent across implementations:
* [`great_expectations.core.batch.BatchRequest`](https://github.com/great-expectations/great_expectations/blob/aac98f52e0e62276584a922c46089d8ee50b5783/great_expectations/core/batch.py#L479) is the "core" definition which defines and allows you to edit the `batch_spec_passthrough` property
* [`great_expectations.datasource.fluent.batch_request.BatchRequest`](https://github.com/great-expectations/great_expectations/blob/aac98f52e0e62276584a922c46089d8ee50b5783/great_expectations/datasource/fluent/batch_request.py#L47) is the "Fluent" version which has the same class name and has the same general function, but does NOT inherit the same properties as the core `BatchRequest` class. Instead it inherits from `pydantic.BaseModel`

It's not immediately clear to people who are relatively new to GX (like me) that these two `BatchReuquest` classes don't provide equivalent functionality.

**A "quick" fix in the Renderer**
The `data_asset_name` IS available in the metadata that gets passed to the renderer that creates the data docs. The [current renderer code does not access it](https://github.com/great-expectations/great_expectations/blob/35c6838c7a9733d26ea14c627db5e87717fe3d76/great_expectations/render/renderer/site_builder.py#L951-L968) however. I WAS able to get the "Asset Name" column in the data docs to populate correctly with the following change to that code:

```python
validation_success = validation.success
batch_kwargs = validation.meta.get("batch_kwargs", {})
batch_spec = validation.meta.get("batch_spec", {})
active_batch = validation.meta.get("active_batch_definition", {}) # <= ADD THIS

self.add_resource_info_to_index_links_dict(
  # ... other props
  asset_name=batch_kwargs.get("data_asset_name")
  or batch_spec.get("data_asset_name")
  or active_batch.get("data_asset_name"), # <= AND ADD THIS
  # ... other props
)
```

This "fixes" the immediate problem, but it feels kind of like a bandaid and that better solution would be to have a unified API for any class called `BatchRequest`.

 - Operating System: Windows
 - Great Expectations Version: 0.17.19
 - Data Source: Fluent Pandas filesystem datasource
 - Cloud environment: n/a

**Additional context**
[There's a thread about this issue in the GX Slack channel](https://greatexpectationstalk.slack.com/archives/CUTCNHN82/p1695848627758059).

Kilo59:
Fix missing `Asset Name` for Fluent Datasources in the Datadocs Home/index page.

See issue below for detailed description of the problem.
- https://github.com/great-expectations/great_expectations/issues/8790

Also updates some related type annotations.

### Before

image

### After

image

### DIFF
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -3880,6 +3880,8 @@ Generated, evaluated, and stored {total_expectations} Expectations during profil
         self,
         resource_identifier: ExpectationSuiteIdentifier
         | ValidationResultIdentifier
+        | GXCloudIdentifier
+        | str
         | None = None,
         site_name: Optional[str] = None,
         only_if_exists: bool = True,
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -4801,7 +4801,7 @@ Generated, evaluated, and stored {total_expectations} Expectations during profil
         validations_store_name=None,
         failed_only=False,
         include_rendered_content=None,
-    ) -> ExpectationValidationResult:
+    ) -> ExpectationValidationResult | dict:
         """Get validation results from a configured store.
 
         Args:
@@ -4839,7 +4839,7 @@ Generated, evaluated, and stored {total_expectations} Expectations during profil
             # run_id_set = set([key.run_id for key in filtered_key_list])
             if len(filtered_key_list) == 0:
                 logger.warning("No valid run_id values found.")
-                return {}  # type: ignore[return-value]
+                return {}
 
             filtered_key_list = sorted(filtered_key_list, key=lambda x: x.run_id)
 
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -169,7 +169,9 @@ if TYPE_CHECKING:
     )
     from great_expectations.data_context.store.validations_store import ValidationsStore
     from great_expectations.data_context.types.resource_identifiers import (
+        ExpectationSuiteIdentifier,
         GXCloudIdentifier,
+        ValidationResultIdentifier,
     )
     from great_expectations.datasource import LegacyDatasource
     from great_expectations.datasource.fluent.interfaces import (
@@ -3875,9 +3877,11 @@ Generated, evaluated, and stored {total_expectations} Expectations during profil
 
     def get_docs_sites_urls(
         self,
-        resource_identifier=None,
+        resource_identifier: ExpectationSuiteIdentifier
+        | ValidationResultIdentifier
+        | None = None,
         site_name: Optional[str] = None,
-        only_if_exists=True,
+        only_if_exists: bool = True,
         site_names: Optional[List[str]] = None,
     ) -> List[Dict[str, str]]:
         """
@@ -5272,11 +5276,13 @@ Generated, evaluated, and stored {total_expectations} Expectations during profil
     @public_api
     def build_data_docs(
         self,
-        site_names=None,
-        resource_identifiers=None,
-        dry_run=False,
+        site_names: list[str] | None = None,
+        resource_identifiers: list[ExpectationSuiteIdentifier]
+        | list[ValidationResultIdentifier]
+        | None = None,
+        dry_run: bool = False,
         build_index: bool = True,
-    ):
+    ) -> dict[str, str]:
         """Build Data Docs for your project.
 
         --Documentation--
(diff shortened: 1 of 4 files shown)


great_expectations/render/renderer/site_builder.py
```diff
@@ -971,6 +977,25 @@ def _add_validations_to_index_links(
...
 
 
+def _resolve_asset_name(validation_results: ExpectationValidationResult) -> str | None:
+    """
+    Resolve the asset name from the validation results meta data.
+    FDS does not store data_asset_name in batch_kwargs or batch_spec and it must be
+    pulled from the active batch definition.
+    """
+    batch_kwargs = validation_results.meta.get("batch_kwargs", {})
+    batch_spec = validation_results.meta.get("batch_spec", {})
```
Kilo59, line 987: One of both of these may be going away in V1 but `active_batch_definition` will still be present.

great_expectations/render/renderer/site_builder.py
```diff
@@ -971,6 +977,25 @@ def _add_validations_to_index_links(
                     logger.warning(error_msg)
 
 
+def _resolve_asset_name(validation_results: ExpectationValidationResult) -> str | None:
+    """
+    Resolve the asset name from the validation results meta data.
+    FDS does not store data_asset_name in batch_kwargs or batch_spec and it must be
+    pulled from the active batch definition.
+    """
```
Kilo59, line 985: @morphatic Thanks so much for the detailed issue description.
I ended up using exactly what you suggested.

Would you like me to make you a CoAuthor on this PR?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -4832,7 +4839,7 @@ def get_validation_result(  # noqa: PLR0913
             # run_id_set = set([key.run_id for key in filtered_key_list])
             if len(filtered_key_list) == 0:
                 logger.warning("No valid run_id values found.")
-                return {}
+                return {}  # type: ignore[return-value]
```
billdirks: Maybe make the return type a union with a dict so you don't have to ignore this here? This may cause you to assert at the call sites though.

billdirks (approved):
Thank you!

great_expectations/render/renderer/site_builder.py
morphatic: Glad it helped! I would not complain about being listed as a co-author. Thank you for asking!

Shinnnyshinshin (approved):

PR #9963, opened 2024-05-23
[MAINTENANCE] Ensure that validation definitions and checkpoints save before running

cdkini:
Ensure that `ValidationDefinition` and `Checkpoint` persist themselves to their respective stores at the start of their `run` methods (if they haven't already been saved). This will ensure that invariably, we will have references to them in the DB and ensure that any foreign keys generated in the result object will point to real data.

### DIFF
--- a/great_expectations/checkpoint/checkpoint.py
+++ b/great_expectations/checkpoint/checkpoint.py
@@ -146,7 +146,7 @@ class Checkpoint(BaseModel):
         run_id: RunIdentifier | None = None,
     ) -> CheckpointResult:
         if not self.id:
-            self._save_before_run()
+            self._add_to_store()
 
         run_id = run_id or RunIdentifier(run_time=dt.datetime.now(dt.timezone.utc))
         run_results = self._run_validation_definitions(
@@ -252,7 +252,12 @@ class Checkpoint(BaseModel):
 
         store.update(key=key, value=self)
 
-    def _save_before_run(self) -> None:
+    def _add_to_store(self) -> None:
+        """This is used to persist a checkpoint before we run it.
+
+        We need to persist a checkpoint before it can be run. If user calls runs but hasn't
+        persisted it we add it for them.
+        """
         from great_expectations import project_manager
 
         store = project_manager.get_checkpoints_store()
--- a/great_expectations/checkpoint/checkpoint.py
+++ b/great_expectations/checkpoint/checkpoint.py
@@ -145,6 +145,9 @@ class Checkpoint(BaseModel):
         expectation_parameters: Dict[str, Any] | None = None,
         run_id: RunIdentifier | None = None,
     ) -> CheckpointResult:
+        if not self.id:
+            self.save()
+
         run_id = run_id or RunIdentifier(run_time=dt.datetime.now(dt.timezone.utc))
         run_results = self._run_validation_definitions(
             batch_parameters=batch_parameters,
@@ -247,7 +250,10 @@ class Checkpoint(BaseModel):
         store = project_manager.get_checkpoints_store()
         key = store.get_key(name=self.name, id=self.id)
 
-        store.update(key=key, value=self)
+        if self.id:
+            store.update(key=key, value=self)
+        else:
+            store.add(key=key, value=self)
 
 
 class CheckpointResult(BaseModel):
--- a/great_expectations/checkpoint/checkpoint.py
+++ b/great_expectations/checkpoint/checkpoint.py
@@ -153,11 +153,7 @@ class Checkpoint(BaseModel):
             run_id=run_id,
         )
 
-        checkpoint_result = CheckpointResult(
-            run_id=run_id,
-            run_results=run_results,
-            checkpoint_config=self,
-        )
+        checkpoint_result = self._construct_result(run_id=run_id, run_results=run_results)
         self._run_actions(checkpoint_result=checkpoint_result)
 
         return checkpoint_result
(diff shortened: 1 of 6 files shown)


tests/core/test_validation_definition.py
```diff
@@ -189,6 +189,9 @@ def test_returns_expected_data(
                 "unsuccessful_expectations": 0,
                 "success_percent": 100.0,
             },
+            meta={
+                "validation_id": None,
```
billdirks: Can the validation_id be None? In the text_checkpoint test above you verify that it's a UUID.

tyler-hoffman (commented):
Is there a way we can create the full result object in one place? It looks like for both the checkpoint and validation id, we allow the result objects to be built, and then later tack on the ids. It feels like passing down a checkpoint/validation id into our run methods would make this stuff a bit more discoverable.

tests/core/test_validation_definition.py
cdkini: It will be populated if and only if we have persisted the object before running. Do we need to update the checkpoint/validation_definition to persist on run if it doesn't already exist in the store?

great_expectations/core/validation_definition.py
```diff
@@ -273,3 +277,15 @@ def identifier_bundle(self) -> _IdentifierBundle:
         self.suite.identifier_bundle()
 
         return _IdentifierBundle(name=self.name, id=self.id)
+
+    @public_api
+    def save(self) -> None:
```
cdkini, line 282: Surprised we missed this one - I'm imagining we want this to ensure parity across domain objects right?

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -231,6 +231,10 @@ def _send_get_request_to_api(self, url: str, params: dict | None = None) -> dict
             raise StoreBackendError(  # noqa: TRY003
                 f"Unable to get object in GX Cloud Store Backend: {get_user_friendly_error_message(http_err)}"  # noqa: E501
             )
+        except requests.ConnectionError as conn_err:
+            raise StoreBackendError(  # noqa: TRY003
+                f"Unable to get object in GX Cloud Store Backend: {conn_err}"
+            )
```
cdkini: Maybe not the best practice but the Cloud fake raises this error if we hit an endpoint that isn't implemented. The `Store.add` flow makes a GET to check existence before POSTing and when it misses, it manifests as this error in our fake.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -237,7 +250,10 @@ def save(self) -> None:
         store = project_manager.get_checkpoints_store()
         key = store.get_key(name=self.name, id=self.id)
 
-        store.update(key=key, value=self)
+        if self.id:
+            store.update(key=key, value=self)
+        else:
+            store.add(key=key, value=self)
```
cdkini: This enables a new flow:

```python
checkpoint = Checkpoint(...)
checkpoint.save() # Would have previously errored but now we can save directly
```

Are we okay with this?

great_expectations/checkpoint/checkpoint.py
```diff
@@ -239,6 +252,14 @@ def save(self) -> None:
 
         store.update(key=key, value=self)
 
+    def _save_before_run(self) -> None:
```
tyler-hoffman: I think @billdirks mentioned this in the first pass at this work, but why is this a separate method from `save`? I think the best move would be to remove this, but then comment the "why" when we call save.

Edit: I see - save updates, this adds. The name still feels off though. very non-blocking: the functionality just happens to be used before running the checkpoint, but isn't really about that. Maybe `_save_to_store` and a docstring on why we need to do this?

great_expectations/checkpoint/checkpoint.py
```diff
@@ -200,6 +199,20 @@ def _build_result_key(
             batch_identifier=batch_identifier,
         )
 
+    def _construct_result(
+        self,
+        run_id: RunIdentifier,
+        run_results: Dict[ValidationResultIdentifier, ExpectationSuiteValidationResult],
+    ) -> CheckpointResult:
+        for result in run_results.values():
+            result.meta["checkpoint_id"] = self.id
```
tyler-hoffman, line 208: Non-blocking, but I'm still in favor of passing checkpoint id down to where results are created and just fully create the result in one place.

great_expectations/core/validation_definition.py
tyler-hoffman: Oh weird. Yeah, definitely seems like we want it here.

tyler-hoffman (approved):
Approving, but with some notes.

I see you and @billdirks agreed to this approach last week, and I don't want to throw a wrench in anything. But I still find it surprising that we would persist a checkpoint upon running it. I don't think this is anywhere near as bad as the old validator/expectation-suite-authoring flow, since you couldn't tell what the outcome was going to be there, but there is still an element of surprise here that I don't love.

I think I would find things more clear if we just raised an exception if someone tried to run a checkpoint without first saving it. But this isn't a super strongly held belief.

great_expectations/checkpoint/checkpoint.py
billdirks: Since @cdkini is out I'll make this change so we can merge. Maybe I'll name to `_add_to_store` since `save` does and update.

great_expectations/core/validation_definition.py
```diff
@@ -273,3 +277,20 @@ def identifier_bundle(self) -> _IdentifierBundle:
...
+        from great_expectations import project_manager
+
+        store = project_manager.get_validation_definition_store()
+        key = store.get_key(name=self.name, id=self.id)
+
+        store.update(key=key, value=self)
+
+    def _save_before_run(self) -> None:
```
billdirks: Similarly, I'll rename this one also.

great_expectations/data_context/store/gx_cloud_store_backend.py
billdirks: I'll try adding a `from conn_err` here.

great_expectations/checkpoint/checkpoint.py
billdirks: Agreed that would be nicer. I'm going to leave this as is for now since @cdkini is out and I think we should merge and then get this sorted if we can.

tyler-hoffman: Sounds great

tyler-hoffman (approved):

PR #9967, opened 2024-05-24
[DOCS] core expectation model metadata

dctalbot:
- refactors core expectation model docstrings

### DIFF
--- a/great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
+++ b/great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
@@ -38,9 +38,17 @@ if TYPE_CHECKING:
     )
     from great_expectations.render.renderer_configuration import AddParamArgs
 
+EXPECTATION_SHORT_DESCRIPTION = (
+    "Expect the set of distinct column values to be contained by a given set."
+)
+COLUMN_ARG = "The column name."
+VALUE_SET_ARG = "A set of objects used for comparison."
+SUPPORTED_DATASOURCES = ["Snowflake", "PostgreSQL"]
+DATA_QUALITY_ISSUES = ["Sets"]
+
 
 class ExpectColumnDistinctValuesToBeInSet(ColumnAggregateExpectation):
-    """Expect the set of distinct column values to be contained by a given set.
+    __doc__ = f"""{EXPECTATION_SHORT_DESCRIPTION}
 
     expect_column_distinct_values_to_be_in_set is a \
     [Column Aggregate Expectation](https://docs.greatexpectations.io/docs/guides/expectations/creating_custom_expectations/how_to_create_custom_column_aggregate_expectations).
@@ -51,9 +59,9 @@ class ExpectColumnDistinctValuesToBeInSet(ColumnAggregateExpectation):
 
     Args:
         column (str): \
-            The column name.
+            {COLUMN_ARG}
         value_set (set-like): \
-            A set of objects used for comparison.
+            {VALUE_SET_ARG}
 
     Other Parameters:
         result_format (str or None): \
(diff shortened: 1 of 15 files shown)


great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
```diff
@@ -38,9 +38,17 @@
     )
     from great_expectations.render.renderer_configuration import AddParamArgs
 
+EXPECTATION_SHORT_DESCRIPTION = (
+    "Expect the set of distinct column values to be contained by a given set."
+)
+COLUMN_ARG = "The column name."
```
NathanFarmer: You probably should move this up into `expectation.py` and import it so you don't need to keep redefining it and can use it to add the `column` field `description`s up there.

great_expectations/expectations/core/expect_column_values_to_be_in_type_list.py
```diff
@@ -55,9 +55,23 @@
 
 logger = logging.getLogger(__name__)
 
+EXPECTATION_SHORT_DESCRIPTION = "Expect a column to contain values from a specified type list."
+COLUMN_ARG = "The column name."
+TYPE_LIST_ARG = """
+    A list of strings representing the data type that each column should have as entries. \
+    Valid types are defined by the current backend implementation and are dynamically loaded. \
```
NathanFarmer: This is going to end up creating a massive tooltip in the UI. Can you shorten this variable to only hold these first 2 sentences and then keep everything from "For example" on in the docstring only?

great_expectations/expectations/core/expect_column_values_to_be_of_type.py
```diff
@@ -74,9 +74,26 @@
     clickhouse_sqlalchemy = None
     ch_types = None
 
+EXPECTATION_SHORT_DESCRIPTION = "Expect a column to contain values of a specified data type."
+COLUMN_ARG = "The column name."
+TYPE__ARG = """
+    A string representing the data type that each column should have as entries. \
+    Valid types are defined by the current backend implementation and are dynamically loaded. \
```
NathanFarmer: Same comment here about shortening this.

NathanFarmer (changes requested):

great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
```diff
@@ -38,9 +39,16 @@
     )
     from great_expectations.render.renderer_configuration import AddParamArgs
 
+EXPECTATION_SHORT_DESCRIPTION = (
+    "Expect the set of distinct column values to be contained by a given set."
+)
+VALUE_SET_ARG = "A set of objects used for comparison."
```
NathanFarmer, line 45: Why are the other fields called "DESCRIPTION", but this one is called "ARG"?

great_expectations/expectations/core/expect_column_max_to_be_between.py
```diff
@@ -43,9 +44,25 @@
     )
     from great_expectations.render.renderer_configuration import AddParamArgs
 
+EXPECTATION_SHORT_DESCRIPTION = (
+    "Expect the column maximum to be between a minimum value and a maximum value."
+)
+MIN_VALUE_ARG = "The minimum value of the acceptable range for the column maximum."
```
NathanFarmer, line 50: Same question on naming conventions.

great_expectations/expectations/expectation.py
```diff
@@ -1771,7 +1772,7 @@ class ColumnAggregateExpectation(BatchExpectation, ABC):
         InvalidExpectationConfigurationError: If no `column` is specified
     """  # noqa: E501
 
-    column: str
+    column: str = Field(description=COLUMN_FIELD_DESCRIPTION)
```
NathanFarmer, line 1775: No problem adding this here, but you will need to do this with the rest of the fields next.

NathanFarmer (approved):
Approving with some non-blocking comments

great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
dctalbot: "description" refers to the pydantic, parameter name and "arg" refers to the docstring section.

NathanFarmer: Ok, but in the next PR you are going to be adding this "arg" to the pydantic description.

PR #9992, opened 2024-05-30
[FEATURE] Snowflake - use `Datasource` level `schema` when creating `TableAsset`

Kilo59:
The most significant change is to use the `Datasource` level `schema` when creating a `TableAsset` via `my_sf_datasource.add_table_asset(...)`.

In addition, several changes were made to make working with `Datasource` level `schema_` + `database` more accessible.

### Convienence properties 
Several properties were added to that can be accessed regardless of the underlying `connection_string` type (`ConnectionDetails`, `SnowflakeDsn` or `ConfigStr`).

1. `Snowflake`
  a. `Snowflake.database`
  b. `Snowflake.schema_`
2. `SnowflakeDsn`
  a. `SnowflakeDsn.schema_`
  b. `SnowflakeDsn.database`
  c. `SnowflakeDsn.account_identifier`
  d. `SnowflakeDsn.warehouse`
  e. `SnowflakeDsn.role`

## Deprecations
Providing a `schema_name` via `SnowflakeDatasource.add_table_asset()` has been deprecated.

### DIFF
--- a/great_expectations/datasource/fluent/snowflake_datasource.py
+++ b/great_expectations/datasource/fluent/snowflake_datasource.py
@@ -31,11 +31,17 @@ from great_expectations.datasource.fluent.sql_datasource import (
     FluentBaseModel,
     SQLDatasource,
     SQLDatasourceError,
+    TableAsset,
 )
 
 if TYPE_CHECKING:
     from great_expectations.compatibility import sqlalchemy
     from great_expectations.compatibility.pydantic.networks import Parts
+    from great_expectations.datasource.fluent.interfaces import (
+        BatchMetadata,
+        Sorter,
+        SortersDefinition,
+    )
     from great_expectations.execution_engine import SqlAlchemyExecutionEngine
 
 LOGGER: Final[logging.Logger] = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


great_expectations/datasource/fluent/snowflake_datasource.py
```diff
@@ -31,22 +37,77 @@
...
+    )
     from great_expectations.execution_engine import SqlAlchemyExecutionEngine
 
 LOGGER: Final[logging.Logger] = logging.getLogger(__name__)
 
 REQUIRED_QUERY_PARAMS: Final[
     Iterable[str]
 ] = {  # errors will be thrown if any of these are missing
-    "database",
-    "schema",
```
Kilo59, line 47: These aren't query params. 🙃 
Luckily, we never released this change.

tests/datasource/fluent/test_snowflake_datasource.py
```diff
@@ -9,62 +11,98 @@
...
+        id="connection_string ConfigStr with required query params",
+    ),
+    param(
+        {"connection_string": "${MY_CONN_STR_FULL}"},
+        id="connection_string ConfigStr - required params part of sub",
+    ),
+    param(
+        {"connection_string": "${MY_CONN_STR_MIN}?${MY_QUERY_PARAMS}"},
+        id="connection_string ConfigStr - dedicated query params sub",
+    ),
```
DrewHoo, line 43: We want to disallow these, right? else we give up the ability to say "hey if you edit that db name, these are all the configured assets that will no longer work"

Kilo59: Yeah, I think the best answer here would be to prevent config substitution in the `path` portion of the connection string.

But I think we should enforce that in the Cloud Backend rather than OSS. I am not 100% convinced of that yet, but that's my initial reaction.

DrewHoo: If we enforce it in the Cloud Backend and not here, we're creating space for migration pain for users who start using the OSS library and want to convert to a cloud account. Another way of asking this question: why is cloud migration pain an acceptable cost to pay to allow config substitution in connection string paths?

Kilo59: We'll enforce this in OSS with a followup.

DrewHoo (approved):

PR #9994, opened 2024-05-31
[MAINTENANCE] Add custom types to Expectation schemas

NathanFarmer:
Add validation models and rich field types to Expectation schemas


tests/test_definitions/column_map_expectations/expect_column_values_to_be_between.json
```diff
@@ -450,7 +450,7 @@
             "column": "x",
             "max_value": 20,
             "min_value": 0,
-            "mostly": 0.9
```
NathanFarmer, line 453: This test case was identical to the one above it functionally. Now it adds value by confirming that mostly can be set to higher precision.

great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
```diff
@@ -175,9 +179,7 @@ class ExpectColumnDistinctValuesToBeInSet(ColumnAggregateExpectation):
                 }}
     """  # noqa: E501
 
-    value_set: Union[list, set, SuiteParameterDict, None] = pydantic.Field(
-        default=[], description=VALUE_SET_DESCRIPTION
-    )
+    value_set: Optional[Union[SuiteParameterDict, ValueSet]]
```
dctalbot, line 182: breaking change?

great_expectations/expectations/expectation.py
```diff
@@ -2362,7 +2365,7 @@ class MulticolumnMapExpectation(BatchExpectation, ABC):
             the expectation.
     """  # noqa: E501
 
-    column_list: List[str]
+    column_list: List[StrictStr] = Field(min_items=1)
```
dctalbot: breaking change?

dctalbot (approved):

great_expectations/expectations/core/expect_column_distinct_values_to_be_in_set.py
NathanFarmer: The existing type didn't really make sense. Why would you allow `None` and then have a mutable empty `list` as default? There are a lot of tests in place for this expectation and they all pass. All documentation specifies this as a required field. Finally, this is only going into V1 which will have other breaking changes included.

great_expectations/expectations/expectation.py
NathanFarmer: I removed `Field(min_items=1)` because we don't know how this will always be implemented, and maybe an empty list is acceptable.

I left `StrictStr` in place of `str` because we shouldn't be allowing type coercion to `str` from other types for a column name. Column names that are numbers or come from some other serialized object don't make sense in practice.

PR #10000, opened 2024-06-04
[FEATURE] Add new `ConfigUri` type

Kilo59:
Added new `ConfigUri` type that behaves like a `ConfigStr` except config substitution is only allowed in the `user` + `password` section of a URL.

It also must be a valid URL as defined by [pydantic.AnyUrl](https://docs.pydantic.dev/1.10/usage/types/#urls).

### Example Usage

```python
from typing import Literal

import great_expectations as gx
from great_expectations.datasource.fluent import SqlDatasource

class MyCustomDatasource(SqlDatasource):
    type: Literal["custom_sql"] = "custom_sql"
    connection_string: ConfigUri


context = gx.get_context()

my_ds = context.sources.add_custom_sql(
    "my_datasource",
    connection_string="my_protocol+my_connector://user:${MY_PASSWORD}@host/a/b"
)
print(my_ds.scheme)
print(my_ds.path)
print(my_ds.password)
```
```sh
my_protocol+my_connector
a/b
${MY_PASSWORD}
```

### Invalid Usage Examples

```python
context.sources.add_custom_sql(
    "no_config_template_str",
    connection_string="my_protocol+my_connector://user:password@host/a/b"
)

context.sources.add_custom_sql(
    "config_sub_outside_of_user:password",
    connection_string="my_protocol+my_connector://user:password@${INVALID}"
)

context.sources.add_custom_sql(
    "must_seperate_config_templates_for_user_and_password",
    connection_string="my_protocol+my_connector://${INVALID}@host/a/b"
)

```

PR #10002, opened 2024-06-04
[FEATURE] Restrict substitutable sections for `Snowflake.connection_string`

Kilo59:
Disallow config substitution for `SnowflakeDatasource.connection_string` outside of the `user:password` sections.

- See #10000 for additional details


DrewHoo (approved):
LGTM

PR #10006, opened 2024-06-06
[MAINTENANCE] Upgrade to pytest 8

Kilo59:
Upgrade from pytest `7`. -> `8.2.2`

In addition to performance improvement and bug fixes.
The biggest change here is better diffs on failure and better error message for assertion failures other than simple equality checks.

We use `pytest-icdiff` for better error diffs, so we should evaluate whether we prefer the `pytest-icdiff` diff or the new native pytest diff.

https://pythontest.com/pytest/pytest-8-is-here/
https://docs.pytest.org/en/stable/changelog.html#pytest-8-2-2-2024-06-04

Also the pytest docs have been improved and they added a Dark Mode 😎.
https://docs.pytest.org/en/stable/index.html

## Other Changes
I've also updated the min versions of all of our pytest plugins.

We are temporarily increasing the unit test timeout we use from `1.5s` to `2.0s` to account for the arc runners seeming to run slower than the old public GitHub runners.


.github/workflows/ci.yml
```diff
@@ -198,7 +198,8 @@ jobs:
         run: pip install . -c constraints-dev.txt -r reqs/requirements-dev-test.txt
 
       - name: Run the unit tests
-        run: invoke ci-tests -m "unit" --xdist --slowest=10 --timeout=1.5 --reports
+        # TODO: revert the timeout back to 1.5 or lower after resolving arc issues 
+        run: invoke ci-tests -m "unit" --xdist --slowest=10 --timeout=2.0 --reports
```
Kilo59: @nicgrayson temporarily increasing the timeout here to see if it makes a difference

tests/test_packaging.py
```diff
@@ -248,7 +248,7 @@ def test_polish_and_ratchet_pins_and_upper_bounds():
         ("requirements-dev-test.txt", "docstring-parser", (("==", "0.15"),)),
         ("requirements-dev-test.txt", "ipykernel", (("<=", "6.17.1"),)),
         ("requirements-dev-test.txt", "moto", (("<", "3.0.0"), (">=", "2.0.0"))),
-        ("requirements-dev-test.txt", "snapshottest", (("==", "0.6.0"),)),
+        ("requirements-dev-test.txt", "snapshottest", (("==", "0.7.0"),)),
```
Kilo59, line 251:
```suggestion
        ("requirements-dev-test.txt", "snapshottest", (("==", "0.6.0"),)),
```

tyler-hoffman (approved):

NathanFarmer (approved):

PR #10026, opened 2024-06-12
[DOCS] Adds GX 1.0 preview docs for Run Validations topic

Rachel-Reverie:
## Description:
- Adds "Create a Validation Definition" page.
- Adds "Choose result format" page.
- Adds "Run a Validation Definition" page.

Code examples are in line and still need to be put under test.

## Definition of done


docs/docusaurus/docs/core/run_validations/_choose_result_format/_BASIC.md
```diff
@@ -0,0 +1,10 @@
+When the `result_format` is set to `"BASIC"` the Validation Results of each Expectation includes a `result` dictionary with information providing a basic justification for why it failed or succeeded. The format is intended for quick feedback and it works well in Jupyter Notebooks.
+
+You can check the [Validation Results reference tables](#validation-results-reference-tables) to see what information is provided in the `result` dictionary.
+
+To create a `"BASIC"` result format configuration use the following code:
+
+```python
+basic_rf_dict = {}
```
tyler-hoffman: nit: it's less code and probably more clear to just oneline this: `basic_rf_dict = {"result_format": "BASIC"}`. We also have enums for these values. I'd probably use those unless you're trying to minimize imports.

docs/docusaurus/docs/core/run_validations/_choose_result_format/_SUMMARY.md
```diff
@@ -0,0 +1,9 @@
+When the `result_format` key is set to `"SUMMARY"` the Validation Results of each Expectation includes a `result` dictionary with information that summarizes values to justify why it failed or succeeded.  This format is intended for more detailed exploratory work and includes additional information beyond what is included by `BASIC`.
```
tyler-hoffman, line 1: nit: Is "justify" the right word? Maybe "explain"?

docs/docusaurus/docs/core/run_validations/_choose_result_format/_result_verbosity_reference_table.md
```diff
@@ -0,0 +1,17 @@
+The following table lists the fields that can be found in the `result` dictionary of a Validation Result and the Result Format verbosity levels that return that field.
+
+| Fields within `result`                |BOOLEAN_ONLY    |BASIC           |SUMMARY         |COMPLETE        |
+----------------------------------------|----------------|----------------|----------------|-----------------
+|    element_count                      |no              |yes             |yes             |yes             |
```
tyler-hoffman, line 5: would using ✅ vs ❌ or something visual like that be easier than "yes" / "no"? I feel like that's the more common way I see these tables.

docs/docusaurus/docs/core/run_validations/_examples/choose_result_format.py
```diff
@@ -0,0 +1,78 @@
...
+
+import great_expectations as gx
+import great_expectations.expectations as gxe
+from great_expectations.core import ValidationDefinition
+from great_expectations.core.expectation_suite import ExpectationSuite
+
+context = gx.get_context()
+
+datasource_name = "all_csv_files"
+path_to_datasource_files = "/Users/rachelreverie/PycharmProjects/pythonProject/core_sandbox/2024_05_23_notebooks/data/taxi_yellow_tripdata_samples"
```
tyler-hoffman: Where are these example python files used? Are these going to be the source of snippets later? Either way, I'd probably use a path relative to the project. I'm guessing not a lot of readers of this have a `/Users/rachelreverie/` directory on their machine 😄

docs/docusaurus/docs/core/run_validations/create_a_validation_definition.md
```diff
@@ -0,0 +1,127 @@
...
+   suite_name = "<NAME OF AN EXISTING EXPECTATION SUITE>"
+   suite = context.get_expectation_suite(suite_name)
+   ```
+
+4. Retrieve the Batch Definition that describes the data to associate with the Expectation Suite.
+
+   Update the values of `data_source_name`, `data_asset_name`, and `batch_definition_name` in the following code with the names of your previously defined Data Source, one of its Data Assets, and a Batch Definition for that Data Asset.  Then execute the code to retrieve the Batch Definition:
+
+   ```python title="Python"
+   data_source_name = "my_datasource"
```
tyler-hoffman, line 59: Maybe a philosophical thing that's already answered, but I think names that tell a bit more of a story might help the concepts stick better, e.g.
```
data_source_name = "my_postgres_datasource"
data_asset_name = "taxi_trips_table"
batch_definition_name = "taxi_trips_by_day"
```
Or something like that

docs/docusaurus/docs/core/run_validations/_choose_result_format/_BASIC.md
Rachel-Reverie: We are trying to minimize imports, and I was advised to use the dictionaries as the "happy path" over enums or passing in strings.  The reason behind the format of the code currently is that I initially intended to show the various other valid keys for each Result Format configuration -- but decided to forgo the additional code in favor of outlining the options in a table.  I'll update these examples; since they aren't currently in python scripts under test they'll likely be revised when that takes place regardless.

docs/docusaurus/docs/core/run_validations/_choose_result_format/_SUMMARY.md
Rachel-Reverie: "Justify" is a hold over from the original language in the Result Format docs.  I think I also prefer "explain," and will update accordingly.

docs/docusaurus/docs/core/run_validations/_examples/choose_result_format.py
Rachel-Reverie: Yeah... If these were actually under test they'd be using a relative path based on the folder that gets generated in the test environment.  This is what happens when I copy code straight out of my Jupyter notebooks! :sweat_smile:

docs/docusaurus/docs/core/run_validations/create_a_validation_definition.md
Rachel-Reverie: I don't think it's been explicitly asked and answered, but my reasoning is: I try to leave the names as generic as possible to indicate that they can be substituted with anything the reader may have, instead.  In this case, the Data Source doesn't need to be a Postgres Data Source, or a File Data Source -- it can be any Data Source for any data format, and the request will work just the same.  Likewise for the Data Asset and the Batch Definition: it doesn't matter what the data in question is, it can be retrieved if you replace the generics with the names of your Data Asset and Batch Definition.

tyler-hoffman: Cool, that seems reasonable to me!

tyler-hoffman (approved):

docs/docusaurus/docs/core/run_validations/_examples/choose_result_format.py
Rachel-Reverie: To answer the first question: these were intended to be the scripts that were put under test, but since we're doing that in a second past for the moment they're just waiting until then.

PR #10027, opened 2024-06-12
[MAINTENANCE] Move `mostly` to correct Expectation classes 

NathanFarmer:

### DIFF
--- a/great_expectations/expectations/expectation.py
+++ b/great_expectations/expectations/expectation.py
@@ -1506,7 +1506,6 @@ class BatchExpectation(Expectation, ABC):
     batch_id: Union[str, None] = None
     row_condition: Union[str, None] = None
     condition_parser: Union[str, None] = None
-    mostly: Mostly = 1.0
 
     domain_keys: ClassVar[Tuple[str, ...]] = (
         "batch_id",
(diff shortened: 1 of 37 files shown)


great_expectations/expectations/expectation.py
```diff
@@ -1506,7 +1506,6 @@ class BatchExpectation(Expectation, ABC):
     batch_id: Union[str, None] = None
     row_condition: Union[str, None] = None
     condition_parser: Union[str, None] = None
-    mostly: Mostly = 1.0
```
NathanFarmer, line 1509: This was wrong. `mostly` does not apply to non-map Expectations, but those Expectations inherit from `BatchExpectation`.

tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -753,7 +753,6 @@ class TestColumnIdentifiers:
     @pytest.mark.parametrize(
         "expectation_type",
         [
-            "expect_column_to_exist",
```
josectobar: Why are we removing this expectation type?

NathanFarmer: `mostly` is not valid for this Expectation

josectobar (changes requested):
Is there a way we can fix the test without having to reduce the number of expectation types we're testing?

josectobar (approved):
Looks great thanks!

Issue #8294, opened 2023-07-12
Can't run checkpoints in parallel because great_expectations.yml is opened in write mode


Chr96er:
**Describe the bug**
(I have looked into the code to the best of my abilities without being able to debug because this happens when multiple threads run at the same time)
When multiple checkpoints are triggered simultaneously it sometimes happens that an exception is caused. In my case 5 checkpoints are triggered from Airflow and most of the time 2-4 of them fail with the exception below. Setting retries to 5 and the retry delay to 3 solves my particular problem, although the amount of checkpoints is likely to grow to 50-100 soon.

As far as I can tell this happens because Great Expectations opens great_expectations.yml in write mode in line 159 of `file_data_context.py`:
`with open(config_filepath, "w") as outfile:`

This was completely unexpected to me, since I don't think running a checkpoint requires any write operation on that file.

Airflow logs:
```

 scheduler | [2023-07-12T11:09:43.252-0500] {dagbag.py:541} INFO - Filling up the DagBag from airflow/dags/run-constraints-checkpoint.py
 scheduler | [2023-07-12T11:09:43.251-0500] {dagbag.py:541} INFO - Filling up the DagBag from airflow/dags/run-constraints-checkpoint.py
 scheduler | [2023-07-12T11:09:43.252-0500] {dagbag.py:541} INFO - Filling up the DagBag from airflow/dags/run-constraints-checkpoint.py
 scheduler | [2023-07-12T11:09:43.260-0500] {dagbag.py:541} INFO - Filling up the DagBag from airflow/dags/run-constraints-checkpoint.py
 scheduler | [2023-07-12T11:09:43.260-0500] {dagbag.py:541} INFO - Filling up the DagBag from airflow/dags/run-constraints-checkpoint.py
 scheduler | [2023-07-12T11:09:46.414-0500] {file_data_context.py:214} INFO - FileDataContext loading fluent config
 scheduler | [2023-07-12T11:09:46.426-0500] {config.py:185} INFO - Loading 'datasources' ->
 scheduler | [{'assets': [...],
 scheduler | 'REDACTED
 scheduler | 'create_temp_table': True,
 scheduler | 'name': 'REDACTED',
 scheduler | 'type': 'sql'}]
 scheduler | [2023-07-12T11:09:46.483-0500] {file_data_context.py:214} INFO - FileDataContext loading fluent config
[...]

PR #10038, opened 2024-06-14
[BUGFIX] Avoid writing to great_expectations.yml during init

jcampbell:
- Fixes #8294 by skipping a write that was happening during init.

Additional cleanup may be valuable

PR #10040, opened 2024-06-17
[BUGFIX] 0.18.x - Avoid writing to great_expectations.yml during init

Kilo59:
- Taken from https://github.com/great-expectations/great_expectations/pull/10038
- Resolves https://github.com/great-expectations/great_expectations/issues/8294

### DIFF
--- a/great_expectations/data_context/data_context/abstract_data_context.py
+++ b/great_expectations/data_context/data_context/abstract_data_context.py
@@ -763,7 +763,7 @@ class AbstractDataContext(ConfigPeer, ABC):
         return self._sources
 
     def _add_fluent_datasource(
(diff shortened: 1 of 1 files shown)


tyler-hoffman (approved):
Obligatory "you should backfill a test for this", but the change is extremely straight-forward, and maybe a headache to test, so I'm not terribly concerned about it.

Was this causing incorrect data to be persisted, or this was just some unneeded writes?

Kilo59:
@tyler-hoffman just unneeded writes.
Not adding tests, cause this is a quick fix to try to make the `0.18.6` release and was already added to `develop`.

This code path will be exercised as part of our current tests, even if we aren't explicitly testing for a specific number of writes.

PR #10048, opened 2024-06-20
[DOCS] GX 1.0 updated docs for Expectations

Rachel-Reverie:
## Description
- Adds topics for:
   - Create an Expectation
   - Test an Expectation
   - Organize Expectations into Expectation Suites

## Definition of done:


docs/docusaurus/docs/core/define_expectations/create_an_expectation.md
```diff
@@ -0,0 +1,178 @@
...
+
+   ```python title="Python"
+   from great_expectations import expectations as gxe
+   ```
+
+2. Determine the Expectation's evaluation parameters
+
+   To determine the parameters your Expectation uses to evaluate data, reference the Expectation's entry in the [Expectation Gallery](https://greatexpectations.io/expectations/).  Under the **Args** section you will find a list of parameters that are necessary for the Expectation to be evaluated, along with the a description of the value that should be provided.
+
+   Parameters with a default value do not need to be set.  Parameters that are set to `None` can be provided at runtime.  Values for all other parameters must be provided when the Expectation is created.
```
tyler-hoffman: > Parameters that are set to `None` can be provided at runtime
I don't believe this is true, and there are more references to it below.

tyler-hoffman (changes requested):
Still looking, but it looks like there's an error about the `None` values for expectations. I think that's supposed to be talking about expectation params with that `$PARAMETER` syntax. (Also that might be considered a more advanced use case that we should push to somewhere else)

docs/docusaurus/docs/core/define_expectations/create_an_expectation.md
```diff
@@ -0,0 +1,178 @@
...
+
+   For example, take the [**Args** section for `ExpectColumnMaxToBeBetween` in the Expectation Gallery](https://greatexpectations.io/expectations/expect_column_max_to_be_between#args).  In this section, you will see that for `ExpectColumnMaxToBeBetween`'s parameters:
+
+      - `column` must be a string.  Since it cannot be set to `None` its value must be provided when the Expectation is created.
+      - `min_value` and `max_value` can be either set when the Expectation is created, or be set to `None` when the Expectation is created and then have appropriate values passed in at runtime, instead.
+      - `strict_min` and `strict_max` both have default values that can be overriden when the Expectation is created, or left as is.  Because they cannot be set to `None` when the Expectation is created these parameters also cannot be passed in at runtime.
```
tyler-hoffman: super nit: `both have default values that can be overriden when the Expectation is created, or left as is` reads a bit weird to me. (also `overriden` is a typo). I'd think about something like `strict_min and strct_max are both optional fields; their defaults are documented with the expectation` or something. I don't know if you really "override" a default.

docs/docusaurus/docs/core/define_expectations/create_an_expectation.md
```diff
@@ -0,0 +1,178 @@
...
+
+2. Determine the Expectation's evaluation parameters
+
+   To determine the parameters your Expectation uses to evaluate data, reference the Expectation's entry in the [Expectation Gallery](https://greatexpectations.io/expectations/).  Under the **Args** section you will find a list of parameters that are necessary for the Expectation to be evaluated, along with the a description of the value that should be provided.
+
+   Parameters with a default value do not need to be set.  Parameters that are set to `None` can be provided at runtime.  Values for all other parameters must be provided when the Expectation is created.
+
+   For example, take the [**Args** section for `ExpectColumnMaxToBeBetween` in the Expectation Gallery](https://greatexpectations.io/expectations/expect_column_max_to_be_between#args).  In this section, you will see that for `ExpectColumnMaxToBeBetween`'s parameters:
+
+      - `column` must be a string.  Since it cannot be set to `None` its value must be provided when the Expectation is created.
```
tyler-hoffman: > Since it cannot be set to `None` its value must be provided when the Expectation is created.
I don't think this line of reasoning is 100% accurate. Its value must be provided when the expectation is created because there is no default value. We could (though I don't know if we currently do) have a field that can be `None`, but does not have a default, and therefore would need to be set to something when instantiating an expectation, even if it was to be set to None. The gallery is actually lying by omission a bit right now, because it doesn't show that `min_value` and `max_value` have a default. Both default to `None`.

docs/docusaurus/docs/core/define_expectations/create_an_expectation.md
```diff
@@ -0,0 +1,178 @@
...
+
+   GX comes with many built in Expectations to cover your data quality needs.  You can find a catalog of these Expectations in the [Expectation Gallery](https://greatexpectations.io/expectations/).  When browsing the Expectation Gallery you can filter the available Expectations by the data quality issue they address and by the Data Sources they support.  There is also a search bar that will let you filter Expectations by matching text in their name or docstring.
+
+   In your code, you will find the classes for Expectations in the `expectations` module:
+
+   ```python title="Python"
+   from great_expectations import expectations as gxe
+   ```
+
+2. Determine the Expectation's evaluation parameters
```
tyler-hoffman: "Evaluation parameters" is AFAIK not the right term here. That has basically been renamed "expectation parameters". I would just call these arguments/parameters, or something along those lines.

docs/docusaurus/docs/core/define_expectations/create_an_expectation.md
```diff
@@ -0,0 +1,178 @@
...
+       column="passenger_count",
+       min_value=None,
+       max_value=None
+   )
+   ```
+   
+   The runtime `expectation_parameters` dictionary for the above example would look like:
+
+   ```python title="Python"
+   runtime_expectation_parameters = {
```
tyler-hoffman: this variable is never used in sample code, and so it doesn't actually seem to do anything or be usable in the provided snippets. I think this whole section should be in an advanced section after running validations is covered, since this only makes sense in the context of running validations.

docs/docusaurus/docs/core/define_expectations/test_an_expectation.md
```diff
@@ -0,0 +1,189 @@
+---
```
tyler-hoffman, line 1: I think this file reads pretty well. I might consider shortening the first snippet and just assume the user has a batch definition, but potato patahto

docs/docusaurus/docs/core/define_expectations/organize_expectation_suites.md
```diff
@@ -0,0 +1,128 @@
...
+
+5. Add the Expectation to the Expectation Suite.
+
+   An Expectation Suite's `add_expectation(...)` method takes in an instance of an Expectation and adds it to the Expectation Suite's configuraton: 
+
+   ```python title="Python input"
+   suite.add_expectation(expectation)
+   ```
+   
+   You can also instantiate an Expectation at the same time as you add it to an Expectation Suite:
```
tyler-hoffman: nit: this is implied b the previous example.

tyler-hoffman (changes requested):
Left a couple more small nits after going through the whole thing

docs/docusaurus/docs/core/define_expectations/create_an_expectation.md
```diff
@@ -0,0 +1,178 @@
...
+
+   ```python title="Python"
+   preset_expectation = gxe.ExpectColumnMaxToBeBetween(
+       column="passenger_count",
+       min_value=4,
+       max_value=6
+   )
+   ```
+   
+   While in this example the same `ExpectColumnMaxToBeBetween` Expectation is created, but the values for `min_value` and `max_value` will be provided at runtime rather than set when the Expectation is created:
```
r34ctor:
```suggestion
   While in this example the same `ExpectColumnMaxToBeBetween` Expectation is created, the values for `min_value` and `max_value` will be provided at runtime rather than set when the Expectation is created:
```

docs/docusaurus/docs/core/define_expectations/create_an_expectation.md
```diff
@@ -0,0 +1,178 @@
...
+   The runtime `expectation_parameters` dictionary for the above example would look like:
+
+   ```python title="Python"
+   runtime_expectation_parameters = {
+      "min_value": 4,
+      "max_value": 6
+   }
+   ```
+
+   In some cases, you may have Expectations with the same parameters all of which are intended to be passed in at runtime.  When you create these Expectations you can provide instructions for GX to look up a specific parameter under a different name in the runtime `expectations_parameter` dictionary.  This is done by setting the value for the parameter as a dictionary containing the key `$PARAMETER` paired with the key to look for in the `expectations_parameter` dictionary.
```
r34ctor:
```suggestion
   In some cases, you may have Expectations with the same parameters, all of which are intended to be passed in at runtime.  When you create these Expectations you can provide instructions for GX to look up a specific parameter under a different name in the runtime `expectations_parameter` dictionary.  This is done by setting the value for the parameter as a dictionary containing the key `$PARAMETER` paired with the key to look for in the `expectations_parameter` dictionary.
```

docs/docusaurus/docs/core/define_expectations/organize_expectation_suites.md
```diff
@@ -0,0 +1,128 @@
...
+
+# Add a previously created Expectation to the Expectation Suite
+suite.add_expectation(expectation)
+
+# Add an Expectation to the Expectation Suite when it is created
+suite.add_expectation(gxe.ExpectColumnValuesToNotBeNull(column="pickup_datetime"))
+
+# Update the configuration of an Expectation, then push the changes to the Expectation Suite
+expectation.column = "pickup_location_id"
+expectation.save()
```
r34ctor, line 115: Does the Expectation Suite need to be added to the context's list of suites? If not, it would be worth mentioning this explicitly.

r34ctor (approved):
A couple of nits and an open question about Expectation Suites, but otherwise this looks great!

docs/docusaurus/docs/core/define_expectations/organize_expectation_suites.md
Rachel-Reverie: That is a good question.  I'll look into that and clarify in the doc.

tyler-hoffman (approved):

PR #10076, opened 2024-07-02
[FEATURE] Add atomic renderer for `ExpectMulticolumnSumToEqual`

NathanFarmer:
Add missing renderer using existing pattern

### DIFF
--- a/great_expectations/expectations/core/expect_multicolumn_sum_to_equal.py
+++ b/great_expectations/expectations/core/expect_multicolumn_sum_to_equal.py
@@ -10,6 +10,10 @@ from great_expectations.expectations.model_field_descriptions import MOSTLY_DESC
 from great_expectations.render import RenderedStringTemplateContent
 from great_expectations.render.components import LegacyRendererType
 from great_expectations.render.renderer.renderer import renderer
+from great_expectations.render.renderer_configuration import (
+    RendererConfiguration,
+    RendererValueType,
+)
 from great_expectations.render.util import (
     num_to_str,
     substitute_none_for_missing,
@@ -22,6 +26,7 @@ if TYPE_CHECKING:
     from great_expectations.expectations.expectation_configuration import (
         ExpectationConfiguration,
     )
+    from great_expectations.render.renderer_configuration import AddParamArgs
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


great_expectations/expectations/core/expect_multicolumn_sum_to_equal.py
```diff
@@ -227,6 +232,49 @@ def schema_extra(schema: Dict[str, Any], model: Type[ExpectMulticolumnSumToEqual
                 }
             )
 
+    @classmethod
+    def _prescriptive_template(
```
NathanFarmer, line 236: Code coverage is apparently poor here, but I believe this is now a legacy pattern anyway.

great_expectations/expectations/core/expect_multicolumn_sum_to_equal.py
```diff
@@ -227,6 +232,49 @@ def schema_extra(schema: Dict[str, Any], model: Type[ExpectMulticolumnSumToEqual
...
+                renderer_configuration=renderer_configuration,
+            )
+
+        if params.mostly and params.mostly.value < 1.0:
+            renderer_configuration = cls._add_mostly_pct_param(
+                renderer_configuration=renderer_configuration
+            )
+            template_str += " must be $sum_total, at least $mostly_pct % of the time."
+        else:
+            template_str += " must be $sum_total."
```
Kilo59, line 272: `non-blocking`
You could accumulate these items in a list and then `.join()` them at the end.
String operations in python are actually relatively fast though, so there might not actually be much benefit to this. 🤷 
```python
template_parts: list[str] = []
if params.column_list:
  array_param_name = "column_list"
  param_prefix = "column_list_"
  renderer_configuration = cls._add_array_params(
    array_param_name=array_param_name,
    param_prefix=param_prefix,
    renderer_configuration=renderer_configuration,
  )
  template_parts += "Sum across columns "
  template_parts += cls._get_array_string(
    array_param_name=array_param_name,
    param_prefix=param_prefix,
    renderer_configuration=renderer_configuration,
  )

...

  renderer_configuration.template_str = "".join(template_parts)
  return renderer_configuration
```

Kilo59 (approved):
LGTM.

~Do we usually test these?~ From your comment, it looks like no.

great_expectations/expectations/core/expect_multicolumn_sum_to_equal.py
NathanFarmer: Prefer to keep it the same as it is on every other core Expectation. My understanding is there is a new (simpler) rendering pattern, and we may be re-evaluating the entire rendering arch.

great_expectations/expectations/core/expect_multicolumn_sum_to_equal.py
NathanFarmer: There are tests for the end result, but it looks like we missed unit tests for the helpers.

Issue #9210, opened 2024-01-09
relax ruamel.yaml pin


Redoubts:
Can the version pin for `ruamel.yaml` be relaxed? The issue linked in the requirements file is closed.

Kilo59:
@ Redoubts I suggest just opening a PR with the desired change and explaining why you want to see the change.

Kilo59:
@Redoubts I suggest just opening a PR with the desired change and explaining why you want to see the change.

Issue #4152, opened 2022-02-09
Documentation using deprecated function of ruamel.yaml


AvdN:
https://github.com/great-expectations/great_expectations/blob/341a9a5f238dc2c05be7c2b3c3cc4e42440230c5/tests/integration/docusaurus/connecting_to_your_data/database/snowflake_python_example.py#L42

I came across the line above, through a question on stack-overflow ( https://stackoverflow.com/questions/71029745/unable-to-initialize-snowflake-data-source/71045925#71045925 ) which was tagged with ruamel.yaml ( I am the author of that package).
The line uses the function `ruamel.yaml.yaml` and that is deprecated, the correct thing to do would be to method `.dump()` on a `ruamel.yaml.YAML()` instance together with the `transform` parameter:
```
from ruamel.yaml import YAML

yaml = YAML()
yaml.dump(datasource_config, sys.stdout, transform=context.test_yaml_config)
```

After answering that question I looked at the source code for great_expectations and noticed `YAML()` is used in the source code, but that there are many entries doing `from ruamel import yaml` in the docusaurus subdirs. 

Fortunately you pin ruamel.yaml ( `>=0.16,<0.17.18`) but IMO you should update the documentation ASAP so people will no longer create new usage of deprecation. 

I am willing to make a PR fixing all these, but I am not sure how to test the documentation, so I would need some assistance if that is required. I rather not make a PR for ~75  files and get it rejected.

austiezr:
Hey @AvdN ! Thank you for reaching out about this; we'd love to work with you on a contribution here, but we need to have an internal conversation around legacy support here before moving forward. We'll review internally and get back to you!

jamesmyatt:
Any update on this? Upper bound constraints make dependency resolution much harder and slower.

BwL1289:
also interested in an update?

AvdN:
> also interested in an update?

Why the question mark? Are you asking me?

This seems to be closed, I never got feedback to go ahead with a PR.

BwL1289:
@AvdN that was directed towards @austiezr. Apologies for the confusion and thanks for your contributions.

Kilo59:
@AvdN In what version is this drepecated function removed?
I am wanting to relax the pinned version of `ruamel.yaml`

nicnic65:
Bumping this to ask the same, our dependency upgrades are being blocked by this pin on `ruamel.yaml`.

AvdN:
[...]

PR #10081, opened 2024-07-03
[MAINTENANCE] loosen ruamel pin

Kilo59:
Loosen the pinned `ruamel.yaml` (`<0.18`) dependency to make dependency resolving easier with `great_expectations` and increase compatibility with other packages.

The deprecated methods `great_expectations` relies on are not removed until `0.18`, see https://github.com/great-expectations/great_expectations/issues/4152#issuecomment-2222958105.

- Resolves https://github.com/great-expectations/great_expectations/issues/9210

We may be able to loosen this even more for V `1.0`.


Kilo59:
Thought we needed this for the agent (but we don't).

requirements.txt
```diff
@@ -25,7 +25,7 @@ pyparsing>=2.4
 python-dateutil>=2.8.1
 pytz>=2021.3
 requests>=2.20
-ruamel.yaml>=0.16,<0.17.18  # See https://github.com/great-expectations/great_expectations/issues/4152
+ruamel.yaml>=0.16,<0.18  # See https://github.com/great-expectations/great_expectations/issues/4152
```
Kilo59, line 28: The deprecated method we currently use is still present up until `0.18`.
We don't need to be quite as strict with our pin.

anthonyburdi (approved):

PR #10084, opened 2024-07-05
[BUGFIX] Z-score renderer when `double_sided`

NathanFarmer:
This Expectation renders the incorrect text when `double_sided` is True


great_expectations/expectations/core/expect_column_value_z_scores_to_be_less_than.py
```diff
@@ -252,8 +252,15 @@ def _prescriptive_template(
         else:
             template_str = "Value z-scores must be "
 
-        if params.double_sided is True:
-            template_str += "greater than -$threshold and less than $threshold"
+        if params.double_sided.value is True:
```
NathanFarmer, line 255: The main bug was that I forgot to use `.value` here. The rest is just an improvement that:
1. ensures the "-" sign is part of the gray styling and
2. uses the correct order for the range when the user passes a negative value.

dctalbot (approved):

PR #10091, opened 2024-07-09
[MAINTENANCE] 0.18.x - mypy - `possibly-undefined`

Kilo59:
Enable the `possible-undefined` optional mypy rule.

>## Warn about variables that are defined only in some execution paths
>
>mypy generates an error if it cannot verify that a variable will be defined in all execution >paths. This includes situations when a variable definition appears in a loop, in a >conditional branch, in an except handler, etc. For example:

```python
from typing import Iterable

def test(values: Iterable[int], flag: bool) -> None:
    if flag:
        a = 1
    z = a + 1  # Error: Name "a" may be undefined [possibly-undefined]

    for v in values:
        b = v
    z = b + 1  # Error: Name "b" may be undefined [possibly-undefined]
```

https://mypy.readthedocs.io/en/stable/error_code_list2.html#warn-about-variables-that-are-defined-only-in-some-execution-paths-possibly-undefined

### See also
- [playground](https://mypy-play.net/?mypy=latest&python=3.12&gist=ce99a37c2c2d71e7af2f2852c4b67941) that @tyler-hoffman totally didn't steal from [stackoverflow](https://stackoverflow.com/questions/78191580/how-to-catch-potentially-undefined-variables-with-pylint-or-mypy)
- https://github.com/great-expectations/cloud/pull/332
- https://github.com/great-expectations/great_expectations/pull/10092


Kilo59:
Before any fixes applied
```
Found 13 errors in 12 files (checked 551 source files)
```

Kilo59:
Before any fixes applied
```
Found 13 errors in 12 files (checked 551 source files)
```

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -230,7 +230,10 @@ def __init__(  # noqa: PLR0913
         filter_properties_dict(properties=self._config, inplace=True)
 
     @override
-    def _get(self, key: Tuple[GXCloudRESTResource, str | None, str | None]) -> ResponsePayload:  # type: ignore[override]
+    def _get(  # type: ignore[override]
+        self,
+        key: Tuple[GXCloudRESTResource, str | None, str | None],
+    ) -> ResponsePayload:
```
Kilo59, line 236:
```suggestion
    def _get(self, key: Tuple[GXCloudRESTResource, str | None, str | None]) -> ResponsePayload: # type: ignore[override]
```

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -195,7 +195,9 @@ def configure_validator(self, validator) -> None:
 
     @override
     def load_batch_data(
-        self, batch_id: str, batch_data: Union[PandasBatchData, pd.DataFrame]  # type: ignore[override]
+        self,
+        batch_id: str,
+        batch_data: Union[PandasBatchData, pd.DataFrame],  # type: ignore[override]
```
Kilo59, line 200:
```suggestion
        self, batch_id: str, batch_data: Union[PandasBatchData, pd.DataFrame] # type: ignore[override]
```

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -230,7 +230,7 @@ def __init__(  # noqa: PLR0913
         filter_properties_dict(properties=self._config, inplace=True)
 
     @override
-    def _get(self, key: Tuple[GXCloudRESTResource, str | None, str | None]) -> ResponsePayload:  # type: ignore[override]
+    def _get(self, key: Tuple[GXCloudRESTResource, str | None, str | None]) -> ResponsePayload: # type: ignore[override]
```
Kilo59, line 233:
```suggestion
    def _get(self, key: Tuple[GXCloudRESTResource, str | None, str | None]) -> ResponsePayload:  # type: ignore[override]
```

tyler-hoffman (approved):
LGTM thanks!

great_expectations/expectations/core/expect_column_median_to_be_between.py
```diff
@@ -298,6 +298,8 @@ def _prescriptive_renderer(  # type: ignore[override] # TODO: Fix this type igno
                 template_str = f"median must be {at_most_str} $max_value."
             elif params["max_value"] is None:
                 template_str = f"median must be {at_least_str} $min_value."
+            else:
+                raise ValueError("unresolvable template_str")
```
Kilo59, line 302: This is the biggest runtime change I've made.
We now raise this `ValueError` if this `template_str` is not defined.
We would have hit an error anyways, but this does it earlier.

I've copied this pattern to 3 other expectations in this PR.

PR #10092, opened 2024-07-09
[MAINTENANCE] mypy - `possibly-undefined`

Kilo59:
Enable the `possible-undefined` optional mypy rule.

>## Warn about variables that are defined only in some execution paths
>
>mypy generates an error if it cannot verify that a variable will be defined in all execution >paths. This includes situations when a variable definition appears in a loop, in a >conditional branch, in an except handler, etc. For example:

```python
from typing import Iterable

def test(values: Iterable[int], flag: bool) -> None:
    if flag:
        a = 1
    z = a + 1  # Error: Name "a" may be undefined [possibly-undefined]

    for v in values:
        b = v
    z = b + 1  # Error: Name "b" may be undefined [possibly-undefined]
```

https://mypy.readthedocs.io/en/stable/error_code_list2.html#warn-about-variables-that-are-defined-only-in-some-execution-paths-possibly-undefined

### See also
- [playground](https://mypy-play.net/?mypy=latest&python=3.12&gist=ce99a37c2c2d71e7af2f2852c4b67941) that @tyler-hoffman totally didn't steal from [stackoverflow](https://stackoverflow.com/questions/78191580/how-to-catch-potentially-undefined-variables-with-pylint-or-mypy)
- https://github.com/great-expectations/cloud/pull/332
- https://github.com/great-expectations/great_expectations/pull/10091


Kilo59:
Before any fixes applied
```
Found 15 errors in 12 files (checked 610 source files)
```

great_expectations/expectations/core/expect_column_sum_to_be_between.py
```diff
@@ -338,16 +340,11 @@ def _prescriptive_renderer(
...
-                **{  # type: ignore[arg-type]
-                    "content_block_type": "string_template",
-                    "string_template": {
-                        "template": template_str,
-                        "params": params,
-                        "styling": runtime_configuration.get("styling"),
-                    },
-                }
+                content_block_type="string_template",
+                string_template={"template": template_str, "params": params, "styling": styling},
```
Kilo59, line 347: @NathanFarmer Can `styling` be `None`?

contrib/cli/requirements.txt
```diff
@@ -1,6 +1,6 @@
 Click>=7.1.2         # CLI tooling
 cookiecutter==2.1.1  # Project templating
-mypy==1.10.0            # Type checker
+mypy==1.10.1            # Type checker
```
Kilo59, line 3:
```suggestion
mypy==1.11.0            # Type checker
```

great_expectations/execution_engine/sqlalchemy_batch_data.py
```diff
@@ -264,28 +264,39 @@ def _create_temporary_table(  # noqa: C901, PLR0912, PLR0915
...
+            stmt_1 = (
+                f"CREATE PRIVATE TEMPORARY TABLE {temp_table_name}"
+                f" ON COMMIT PRESERVE DEFINITION AS {query}"
+            )
+            # prior to oracle 18c only GLOBAL temp tables existed and only the data is transient
+            # this means an empty table will persist after the db session
+            stmt_2 = (
+                f"CREATE GLOBAL TEMPORARY TABLE {temp_table_name}"
+                f" ON COMMIT PRESERVE ROWS AS {query}"
+            )
```
joshua-stauffer, line 299: this looks like a feature unrelated to the mypy change - if this is going to be included, we should test it

joshua-stauffer (changes requested):

great_expectations/execution_engine/sqlalchemy_batch_data.py
Kilo59: It is because of the mypy rule change.
The only thing not related are the formatting changes.

You can see the same change here (in 0.18.x) without the extra formatting diff.
Are you seeing a behavior change that I missed?
https://github.com/great-expectations/great_expectations/pull/10091/files#diff-80bb3fa3777f40e701e2d78e2ab270456bf1f7f88ec61c331fcaff00f2219ee7

joshua-stauffer: thanks! looked net new, but i see that its just shuffled around

joshua-stauffer (approved):
👍

PR #10105, opened 2024-07-12
[MAINTENANCE] Parametrize `TestConnectionError`

Kilo59:
Update `TestConnectionError` exception to make usage more consistent and bring it in-line with suggestions from `TRYceratops`.

In particular, `TRY003`
https://docs.astral.sh/ruff/rules/raise-vanilla-args/

Also updates `TestConnectionError` to inherit from `ConnectionError` instead the base `Exception`.

---

# raise-vanilla-args (TRY003)

Derived from the **tryceratops** linter.

## What it does
Checks for long exception messages that are not defined in the exception
class itself.

## Why is this bad?
By formatting an exception message at the `raise` site, the exception class
becomes less reusable, and may now raise inconsistent messages depending on
where it is raised.

If the exception message is instead defined within the exception class, it
will be consistent across all `raise` invocations.

This rule is not enforced for some built-in exceptions that are commonly
raised with a message and would be unusual to subclass, such as
`NotImplementedError`.

## Example
```python
class CantBeNegative(Exception):
    pass


def foo(x):
    if x < 0:
        raise CantBeNegative(f"{x} is negative")
```

Use instead:
```python
class CantBeNegative(Exception):
    def __init__(self, number):
        super().__init__(f"{number} is negative")


def foo(x):
    if x < 0:
        raise CantBeNegative(x)
```

---

## Related

- https://guicommits.com/handling-exceptions-in-python-like-a-pro/
- https://github.com/guilatrova/tryceratops

### DIFF
--- a/great_expectations/datasource/fluent/interfaces.py
+++ b/great_expectations/datasource/fluent/interfaces.py
@@ -171,7 +171,28 @@ class PartitionerProtocol(PartitionerSortingProtocol, Protocol):
 
 
 class TestConnectionError(Exception):
-    pass
+    """
+    Raised if `.test_connection()` fails to connect to the datasource.
+    """
+
+    def __init__(
+        self,
+        message: str = "Attempt to connect to datasource failed",
+        cause: Exception | None = None,
+        addendum: str | None = None,
+    ):
+        """
+        Args:
+            `message` base of the error message to be provided to the user.
+            `addendum` is optional additional information that can be added to the error message.
+            `cause` is the original exception that caused the error, the repr of which will be added
+                to the error message.
+        """
+        if cause:
+            message += f": due to {cause!r}"
+        if addendum:
+            message += f": {addendum}"
+        super().__init__(message)
 
 
 class GxDatasourceWarning(UserWarning):
--- a/great_expectations/datasource/fluent/interfaces.py
+++ b/great_expectations/datasource/fluent/interfaces.py
@@ -178,15 +178,16 @@ class TestConnectionError(ConnectionError):
     def __init__(
         self,
         message: str = "Attempt to connect to datasource failed",
+        *,
         cause: Exception | None = None,
         addendum: str | None = None,
     ):
         """
         Args:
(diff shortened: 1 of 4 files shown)


great_expectations/datasource/fluent/interfaces.py
```diff
@@ -170,8 +170,31 @@ def batch_parameters_to_batch_spec_kwarg_identifiers(
...
+    def __init__(
+        self,
+        message: str = "Attempt to connect to datasource failed",
+        cause: Exception | None = None,
+        addendum: str | None = None,
+    ):
+        """
+        Args:
+            `message` base of the error message to be provided to the user.
+            `addendum` is optional additional information that can be added to the error message.
```
tyler-hoffman: Looks like the last 2 args are out of order in the docstring. I could let this slide if there was a `*` between `message` and the other args.

tyler-hoffman (approved):

great_expectations/datasource/fluent/interfaces.py
Kilo59: Thanks, I will fix both issues.
Fix the order and make them keyword only.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -170,8 +170,32 @@ def batch_parameters_to_batch_spec_kwarg_identifiers(
...
+        Args:
+            `message` base of the error message to be provided to the user.
+            `cause` is the original exception that caused the error, the repr of which will be added
+                to the error message.
+            `addendum` is optional additional information that can be added to the error message.
+        """
+        self.cause = cause  # not guaranteed to be the same as `self.__cause__`
+        self.addendum = addendum
+        if cause:
+            message += f": due to {cause!r}"
```
Kilo59, line 195:
```suggestion
            message += f": due to {cause.__class__.__name__}:{e!s}"
```
Alternatively, we could do this, which might look slightly better in our app.

PR #10115, opened 2024-07-17
[DOCS] 1.0 connect to filesystem data guides

Rachel-Reverie:
## Description
Adds guides for:
- Creating a filesystem Data Source
- Creating a file or directory Data Asset on that Data Source
- Creating a path, whole folder, or partitioned yearly/monthly/daily Batch Definition on that Data Asset

For the environments:
- Local and networked
- S3
- GCS
- ABS

Note: example scripts are not under test; that will be done in a second pass PR.

## Definition of done:


docs/docusaurus/docs/core/connect_to_data/connect_to_data.md
```diff
@@ -10,7 +10,7 @@ import LinkCard from '@site/src/components/LinkCard';
 import OverviewCard from '@site/src/components/OverviewCard';
 
 <OverviewCard title={frontMatter.title}>
-  Connect to data in various source formats.  Organize that data into Batches for validation.
+      {frontMatter.description}
```
tyler-hoffman, line 13: oh does this just pull from the description at the top?!

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_examples/_directory_partitioned_daily.py
```diff
@@ -0,0 +1,27 @@
...
+from great_expectations import gx
+
+context = gx.get_context()
+
+data_source_name = "my_filesystem_data_source"
+data_asset_name = "my_file_data_asset"
+file_data_asset = context.get_data_source(data_source_name).get_asset(data_asset_name)
+
+# <snippet name="docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_examples/_directory_partitioned_daily.py - define Batch Definition parameters">
+batch_definition_name = "yellow_tripdata_sample_daily"
```
tyler-hoffman, line 11: very nit: I'd probably merge this with the next snippet, since they don't make sense without eachother.

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_examples/_file_partitioned_daily.py
```diff
@@ -0,0 +1,31 @@
+# <snippet name="docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_examples/_file_partitioned_daily.py - full_example">
+from great_expectations import gx
+
```
tyler-hoffman, line 3: just noting - this looks basically the same as the directory partitioned daily file. maybe that's fine. but it seems like the actual distinction is in the datasource/asset setup that's outside of this file.

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_tab-directory_batch_definition.md
```diff
@@ -0,0 +1,180 @@
...
+
+2. Add a Batch Definition to the Data Asset.
+
+   A whole directory Batch Definition returns all of the records in a data file as a single Batch.  A partitioned directory Batch Definition will subdivide the records according to a datetime field and return those records that match a specified year, month, or day.
+
+   <Tabs queryString="batch_definition" groupId="batch_definition" defaultValue='whole_directory'>
+
+   <TabItem value="whole_directory" label="Whole directory">
+   
+   Because a whole directory Batch Definition returns the records from all of the files it can read in the Data Asset you only need to provide one addditional piece of information to define one:
```
tyler-hoffman, line 31: I dig this kind of explanation.

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_tab-directory_batch_definition.md
```diff
@@ -0,0 +1,180 @@
...
+   </TabItem>
+
+   <TabItem value="daily" label="Daily">
+   
+   Update the `batch_definition_name` and `batch_definition_column` variables in the following code, then execute it to create a Batch Definition that subdivides the records in a directory by day:
+
+   ```python title="Python" name="docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_examples/_directory_partitioned_daily.py - define Batch Definition parameters"
+   ```
+
+   </TabItem>
```
tyler-hoffman, line 78: super nit: looking at indentation here, it's hard to tell what's nested under what. Was there a `Tabs` inside a `TabsItem` inside a `Tabs`?

tyler-hoffman (approved):
There's a lot here, but it looks good to me! Mostly nits at code level, and then cross-referenced what you have here with some of the scripts referenced in https://greatexpectations.atlassian.net/wiki/spaces/SUP/pages/1090486302/Integration+Testing+against+GCS.

Thanks!

docs/docusaurus/docs/core/connect_to_data/connect_to_data.md
Rachel-Reverie: Yep!

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_tab-directory_batch_definition.md
Rachel-Reverie: Indeed there was.  Sometimes Docusaurus can be finicky with indentation, though.  There's a certain level of indentation which is also a short cut for "make this a code block" which can be a bit painful.  I often do my reviews in the netlify or local site preview so I can be sure that tabs and nested tab groups are rendering properly, and then search for the content that needs corrected in the actual files only after I've spotted something in the rendered site.

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_examples/_file_partitioned_daily.py
Rachel-Reverie: Most of the setup -- retrieving the Data Source and Asset -- are the same.  The major difference in the code is whether or not the Batch Definition takes in a `column` parameter or a `regex` parameter, and the corresponding values.  The explanation surrounding the directory vs the file Batch Definition is also different since each returns records on a different basis, so it seemed reasonable to maintain a separate script for each.  Since these are going to be put under test eventually, I wanted each to stand on their own (and to have the full process so that they could be put in the 'full example' tab as a full process without having to have two code blocks pulling from two files). Finally, having each under test would ensure that if a breaking change was made in either the File or Directory workflow without a similar change being made in the other, it would be caught when that particular version of the script ran in CI.

But yes: A lot of the time when we're dealing with examples toward the end of a process the bulk of the script is identical to previous ones with only a method or two and the corresponding parameters at the end having been changed.

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_examples/_directory_partitioned_daily.py
Rachel-Reverie: Good catch.  Those were supposed to be merged -- they are for all the other Batch Definition sections in this document, but apparently I overlooked doing that in the directory Data Asset partitioned Batch Definition examples.

PR #10124, opened 2024-07-18
[MAINTENANCE] Ruff `0.5.3`

Kilo59:
Update to `ruff` `0.5.3` and enable Github Actions annotations on linter errors.

https://docs.astral.sh/ruff/integrations/#github-actions

image

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -29,6 +29,7 @@ from docs.sphinx_api_docs_source.build_sphinx_api_docs import SphinxInvokeDocsBu
 
 if TYPE_CHECKING:
     from invoke.context import Context
+    from typing_extensions import Literal
 
 
 LOGGER = logging.getLogger(__name__)
@@ -138,6 +139,7 @@ def lint(
     fmt_: bool = True,
     fix: bool = False,
     unsafe_fixes: bool = False,
+    output_format: Literal["full", "concise", "github"] | None = None,
     watch: bool = False,
     pty: bool = True,
 ):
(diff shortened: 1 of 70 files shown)


tasks.py
```diff
@@ -153,6 +153,8 @@ def lint(
         cmds.append("--unsafe-fixes")
     if watch:
         cmds.append("--watch")
+    if os.getenv("GITHUB_ACTIONS"):
```
billdirks: Is this always set in github? It allows integration with their UI?

tests/datasource/fluent/test_invalid_datasource.py
```diff
@@ -50,29 +50,28 @@
...
 
     @pytest.mark.parametrize("base_ds_method_name", DATASOURCE_PUBLIC_METHODS)
     def test_datasource_methods(self, base_ds_method_name: str):
         """Ensure that InvalidDatasource overrides the applicable Datasource methods."""
-        for base_ds_method_name in DATASOURCE_PUBLIC_METHODS:
-            method = getattr(InvalidDatasource, base_ds_method_name, None)
-            assert method, f"Expected {base_ds_method_name} to be defined on InvalidDatasource"
-            with pytest.raises(TypeError):
-                method()
+        method = getattr(InvalidDatasource, base_ds_method_name, None)
```
billdirks, line 63: Why do we remove the loop over `DATASOURCE_PUBLIC_METHODS` here? Oh just saw the paramterization above and I'm guessing that's the reason? Nice find!

billdirks (approved):

tests/datasource/fluent/test_invalid_datasource.py
Kilo59: Yeah, one of the new ruff rules that caught it and it still took me a bit to see it.

tasks.py
Kilo59: Yes exactly.
https://docs.astral.sh/ruff/integrations/#github-actions

PR #10126, opened 2024-07-19
[FEATURE] Misc Data Source cleanup

joshua-stauffer:
1. This PR adds logic to drop `DataAsset.order_by` on serialization. The field has been replaced by BatchDefinitions, but it has not yet been removed, and the cloud API will raise a 422 error if it is present.
2. Error messages from 422 responses are now also provided to the user within the exception message.
3. Default serialization of data sources is with `exclude_unset=True`. This PR adds explicit `method_name` params to partitioners constructed in the batch definition API to ensure they are serialized.


great_expectations/data_context/store/datasource_store.py
```diff
@@ -85,7 +86,20 @@ def serialize(self, value: FluentDatasource) -> dict:
         """
         See parent 'Store.serialize()' for more information
         """
-        return value._json_dict()
+        # DataAsset.order_by is not supported by v1, but has not yet been removed,
+        # so we drop it before serialization and alert the user with a warning.
```
Kilo59, line 90: Why do we need this warning period before removing it?

joshua-stauffer: this will not be removed prior to release, but we're not serializing it, so this alerts the user

great_expectations/data_context/store/gx_cloud_store_backend.py
```diff
@@ -62,6 +62,8 @@ def get_user_friendly_error_message(
             errors = error_json.get("errors")
         if errors:
             support_message.append(json.dumps(errors))
+        else:
+            support_message.append(json.dumps(error_json))
```
tyler-hoffman, line 66: The flow here is somewhat confusing. If I'm reading this correctly, there are 3 ways we get errors?
* error_json is a list, so assign errors to error_json
* error_json is a dict with an errors list, so we pull them out
* error_json is the error itself, so we use that?
Why not make the first if/else clear about those flows (i.e. replace the else with an elif isinstance(dict), then add another else to say `errors = error_json`

great_expectations/datasource/fluent/data_asset/path/directory_asset.py
```diff
@@ -47,23 +47,29 @@ def add_batch_definition_daily(self, name: str, column: str) -> BatchDefinition:
         # todo: test column
         return self.add_batch_definition(
             name=name,
-            partitioner=ColumnPartitionerDaily(column_name=column),
+            partitioner=ColumnPartitionerDaily(
```
tyler-hoffman, line 50: These `method_name`s are the default values, but just added so that pydantic serializes correctly, right?

If so, I think our serialization relies on way too much knowledge on how things are initialized, and should change (in another PR)

tyler-hoffman (approved):
I think updating that branching logic would make sense before merging, but not blocking

great_expectations/datasource/fluent/data_asset/path/directory_asset.py
Kilo59: Sounds like we need to do what we did for the `type` field, and update the `self.__fields_set__` attribute before serialization.

https://github.com/great-expectations/great_expectations/blob/dc46c682a66402f6c9b79ad7e67b51d59557366c/great_expectations/datasource/fluent/fluent_base_model.py#L176

https://github.com/great-expectations/great_expectations/blob/dc46c682a66402f6c9b79ad7e67b51d59557366c/great_expectations/datasource/fluent/constants.py#L9-L11

great_expectations/data_context/store/gx_cloud_store_backend.py
joshua-stauffer: that was my first approach, but they've been typed in an artificial way (using `TypedDict`), and it would be more effort to rewrite. I think its pretty high priority to get these API errors to users, so just went with the most direct approach.

Kilo59: Yes that TypedDict isn't accurate anymore.
Updating and standardizing the format of our DomainErrors is one of my current priorities.

Until that's done I don't think it's worth spending too much time on this, but we do need to make sure users can see some details about the errors.

PR #10146, opened 2024-07-26
[FEATURE] Clean up import structure for gx

billdirks:
Cleanup dependencies to `dir(gx)`, `dir(gx.package)`, etc so all the imports are in the correct place. For example all the key domain objects can be accessed off of `gx`, such as `gx.ExpectationSuite`, `gx.Checkpoint`, etc.

We still have extra symbols appearing. To update that will require refactoring the codebase since they used internally when importing subpackages/submodules/symbols.


great_expectations/checkpoint/checkpoint.py
```diff
@@ -107,7 +107,7 @@ class Config:
     def _validate_validation_definitions(
         cls, validation_definitions: list[ValidationDefinition] | list[dict]
     ) -> list[ValidationDefinition]:
-        from great_expectations import project_manager
+        from great_expectations.data_context.data_context.context_factory import project_manager
```
tyler-hoffman, line 110: Not loving the `data_context.data_context.` duplication, but I'm assuming there's a good reason for it, and more importantly it's not part of the public API.

great_expectations/exceptions/__init__.py
```diff
@@ -1 +1,79 @@
-from .exceptions import *
+from .exceptions import (
```
tyler-hoffman, line 1: For consideration at a later time: this and the next file are going to be a pain to maintain (and maybe surprising for folks adding exceptions/expectations). Using magic should be a last resort, but maybe it's warranted here. Do we want to consider using some cleverness with the `inspect` module to dynamically infer these? That'd also make life easier if we wanted to start exposing `__all__` in these.

tyler-hoffman (approved):
LGTM!

great_expectations/exceptions/__init__.py
billdirks: I will look into this for followup work.

tyler-hoffman: totally non-urgent!

great_expectations/checkpoint/checkpoint.py
billdirks: I tried to revert this in a bunch and this caused circular dependency issues. I'm guessing we don't always have to do this and we might be able to resolve but I'm going to check it in like this for now and we can investigate later.

PR #10150, opened 2024-07-30
[DOCS] 1.0 preview docs: Configure project Stores

Rachel-Reverie:
## Description:
- Adds the 1.0 guide for configuring project Stores in a File Data Context

Note: Code is not under test; this will be done in a followup PR.

## Definition of Done:


docs/docusaurus/docs/core/configure_project_settings/configure_metadata_stores/configure_metadata_stores.md
```diff
@@ -0,0 +1,118 @@
...
+
+   Once a Data Context's `variables` have been updated, the changes need to be saved to the Data Context's configuration file so that they will persist when the Data Context is loaded in the future.  This is done with:
+
+   ```python title="Python"
+   context.variables.save()
+   ```
+
+5. Re-initialize the File Data Context.
+
+   Because Store configurations are loaded when the Data Context is initialized, you will need to re-initialize your Data Context before your changes will take effect.  This is done by loading the Data Context again, exactly as when it was loaded the first time:
```
billdirks, line 85: I didn't know this was true. I will file a ticket so we don't have to do this.

billdirks (approved):
This LGTM, thanks!

docs/docusaurus/docs/core/configure_project_settings/configure_metadata_stores/configure_metadata_stores.md
billdirks: https://greatexpectations.atlassian.net/browse/V1-433

docs/docusaurus/docs/core/configure_project_settings/configure_metadata_stores/configure_metadata_stores.md
```diff
@@ -0,0 +1,118 @@
...
+# Access the Stores through the Data Context's `variables` attribute:
+print(context.variables.stores['expectations_store'])
+print(context.variables.stores['validation_definition_store'])
+print(context.variables.stores['checkpoint_store'])
+print(context.variables.stores['suite_parameter_store'])
+print(context.variables.stores['validation_results_store'])
+
+# Update the path of the Data Context's Expectations Store:
+expectation_store_directory = "my_expectations_store/"
+context.variables.stores['expectations_store']['store_backend']['base_directory'] = expectation_store_directory
```
billdirks: I just tried running this and this isn't updating. I will update will the real assignment statement here.

billdirks: It looks like `context.variables.stores` always returns a copy of stores with any variable substitution substituted (eg `${my_var}` will resolve to it's actual value). To edit a field you need to do access it through `context.variables.config` so the assignment will look like:
```
context.variables.config.stores['expectations_store']['store_backend']['base_directory'] = expectation_store_directory
```

docs/docusaurus/docs/core/configure_project_settings/configure_metadata_stores/configure_metadata_stores.md
```diff
@@ -0,0 +1,118 @@
...
+
+# Save changes to the Data Context's configuration:
+context.variables.save()
+
+# Re-initialize the Data Context:
+context = gx.get_context(mode="file")
```
billdirks, line 116: I actually have to restart the ipython session. After updating the config as I describe earlier this always fails with:
```
context = gx.get_context(mode="file")
Invalid store configuration: Please check the configuration of your TupleFilesystemStoreBackend named expectations_store. Exception was: 
 'UUID' object has no attribute 'replace'
```

billdirks: This is actually quite an subtle bug. I've described it here: https://greatexpectations.atlassian.net/browse/V1-434

billdirks (changes requested):
Thanks for writing this. It has led to the discovery of a rather bad bug. I think we should hold off on merging until we discuss our path forward.

docs/docusaurus/docs/core/configure_project_settings/configure_metadata_stores/configure_metadata_stores.md
Rachel-Reverie: I did some digging -- this might be a mistake on my part.  The notes I finally turned up state that when toggling analytics through `context.variables` you need to re-initialize your Data Context for it to take effect, not when you update Stores.

docs/docusaurus/docs/core/configure_project_settings/configure_metadata_stores/configure_metadata_stores.md
billdirks: I'll work on fixing this bug so we can merge the docs.

docs/docusaurus/docs/core/configure_project_settings/configure_metadata_stores/configure_metadata_stores.md
```diff
@@ -0,0 +1,118 @@
...
+   
+   When updating a Store configuration you will use the same key to access it from your Data Context's `variables.stores` attribute as was used to print it in the above example.
+   
+3. Update the `base_path` of the Store to change.
+
+   Each Store has a `store_backend` configuration that determines how and where the Store accesses and saves information.  To change the location of a Store, you will update the `store_backend` configuration's `base_directory` value.  For instance, to change the location of an Expectation Store, you would update the `expectation_store_directory` variable in the following code and execute it:
+
+   ```pyhton title="Python"
+   expectation_store_directory = "my_expectations_store/"
+   context.variables.stores['expectations_store']['store_backend']['base_directory'] = expectation_store_directory
```
billdirks: This needs to be updated to:
```suggestion
   context.variables.config.stores['expectations_store']['store_backend']['base_directory'] = expectation_store_directory
```

docs/docusaurus/docs/core/configure_project_settings/configure_metadata_stores/configure_metadata_stores.md
billdirks: I believe this will be true for this right now. We have a bug tracking toggling analytics but the underlying issue will cause the same thing behavior here.

docs/docusaurus/docs/core/configure_project_settings/configure_metadata_stores/configure_metadata_stores.md
billdirks: This needs to be updated to:
```suggestion
context.variables.config.stores['expectations_store']['store_backend']['base_directory'] = expectation_store_directory
```

billdirks (commented):
I've added 2 changes (really the same 1 in 2 places) that need to happen. Otherwise, LGTM. I'll work on fixing the outstanding bug.

docs/docusaurus/docs/core/configure_project_settings/configure_metadata_stores/configure_metadata_stores.md
billdirks: This is new fixed and merged.

docs/docusaurus/docs/core/configure_project_settings/configure_metadata_stores/configure_metadata_stores.md
```diff
@@ -0,0 +1,118 @@
...
+
+   ```python title="Python"
+   import great_expectations as gx
+   
+   context = gx.get_context(mode="file")
+   ```
+
+2. Determine the Store to update.
+
+   GX utilizes 5 Stores for different types of data and metadata.  These Stores are the Expectations Store, Validation Definitions Store, Checkpoint Store, Validation Results Store, and the Suite Parameter Store.  All Stores can be accessed by passing a corresponding key to a Data Context's `variables.config.stores` attribute.  The following code shows how to print the configuration for each of these stores:
```
billdirks: If you don't use `.config` you will see the resolved store name after variable substitution. If you use `.config` you will see the raw name without substitution. Do we want to make that distinction here?

billdirks (approved):
Added a comment out the distinction between using `.config` and not when accessing the store variables. Not sure if we want to call out the difference? But it might be nice so people can see what the real store path is.

billdirks (approved):
Thanks for the update. That sounds great!

PR #10165, opened 2024-08-01
[FEATURE] Update dataframe batch.validate workflow

billdirks:
The shape of the API are implementing is described is:
```
# Initial data asset setup
context = gx.get_context()
datasource = context.data_sources.add_spark(name="myspark")
asset = datasource.add_dataframe_asset(name="spark_asset")

# Validating a batch
bd = asset.add_batch_definition_whole_dataframe(name="bd")
batch = bd.get_batch(batch_parameters={"dataframe": df})
# my_expectation is an Expectation
batch.validate(my_expectation)

# Validating a batch definition
bd = asset.add_batch_definition_whole_dataframe(name="another_bd")
# es is an ExpectationSuite
validation_definition = context.add_validation_definition(batch_definition=bd, expectation_suite=es)
validation_definition.validate(batch_parameters={"dataframe": df})
```


great_expectations/datasource/datasource_dict.py
```diff
@@ -129,13 +129,7 @@ def _init_fluent_datasource(self, name: str, ds: FluentDatasource) -> FluentData
                         datasource_name=name,
                         data_asset_name=asset.name,
                     )
-                    cached_data_asset = self._in_memory_data_assets.get(in_memory_asset_name)
```
billdirks, line 132: This code block is removed because we no longer store the dataframe on the asset.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -477,7 +479,11 @@ def _get_batch_metadata_from_batch_request(self, batch_request: BatchRequest) ->
         batch_metadata = _ConfigurationSubstitutor().substitute_all_config_variables(
             data=batch_metadata, replace_variables_dict=config_variables
         )
-        batch_metadata.update(copy.deepcopy(batch_request.options))
+        batch_metadata.update(
```
billdirks, line 482: This was added since the dataframe now is always passed in in options and we don't want to copy that to the metadata (spark will actual die if we try). Previously, for runtime dataframes we forced `batch_request.options` to be `{}` and added a special argument for the dataframe which broke Liskov and arguably the Interface Segragation principles of SOLID.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -994,7 +1000,7 @@ def __init__(  # noqa: PLR0913
     def _create_id(self) -> str:
         options_list = []
         for key, value in self.batch_request.options.items():
-            if key != "path":
+            if key not in ("path", "dataframe"):
```
billdirks, line 1003: We don't want the `dataframe` to be part of the Batch id. We should probably have this configurable at the asset/batch definition level since it seems strange for the top level batch to know specifics about different assets/batch definitions that produce them, but I am following the pattern here and we can refactor later.

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -356,21 +355,13 @@ def _short_id() -> str:
     return str(uuid.uuid4()).replace("-", "")[:11]
 
 
-class DataFrameAsset(_PandasDataAsset, Generic[_PandasDataFrameT]):
+class DataFrameAsset(_PandasDataAsset):
```
billdirks, line 358: We remove the dataframe from the asset since it now needs to be passed in when making the batch definition. This let's us remove all the dataframe logic and this Generic from the asset.

tests/datasource/fluent/integration/test_integration_datasource.py
```diff
@@ -483,15 +487,20 @@ def test_pandas_data_adding_dataframe_in_file_reloaded_context(
...
+    assert batch.data.dataframe.equals(df)
 
+    # Reload the asset and see that we can re-add the df to the batch definition
     context = gx.get_context(context_root_dir=context.root_directory, cloud_mode=False)
     dataframe_asset = context.get_datasource(datasource_name="fluent_pandas_datasource").get_asset(
         asset_name="my_df_asset"
     )
-    _ = dataframe_asset.build_batch_request(dataframe=df)
-    assert dataframe_asset.dataframe.equals(df)  # type: ignore[attr-defined] # _PandasDataFrameT
+    reloaded_batch_def = dataframe_asset.get_batch_definition(batch_definition_name="bd")
```
billdirks, line 506: We should change the argument to `name` instead of `batch_definition_name`. I can do that in a separate PR.

billdirks: I actually see both patterns in the code depending on domain object. I'd prefer these to all be name since we know the type of the object from the method name, eg `get_<domain_object_type>`.

tests/datasource/fluent/integration/test_integration_datasource.py
```diff
@@ -525,19 +541,124 @@ def test_spark_data_adding_dataframe_in_file_reloaded_context(
...
+        context.get_datasource(datasource_name="fluent_spark_datasource")
+        .get_asset(asset_name="my_df_asset")
+        .get_batch_definition(batch_definition_name="bd")
+    )
+    new_batch = retrieved_bd.get_batch(batch_parameters={"dataframe": spark_df})
+    assert isinstance(new_batch.data, SparkDFBatchData)
+    assert new_batch.data.dataframe.toPandas().equals(df)
+
+
+@dataclass
```
billdirks, line 563: From here to the end of the file are new e2e tests that verify the validation workflows for batches and validation_defintions.

docs/docusaurus/docs/snippets/get_existing_data_asset_from_existing_datasource_pandas_filesystem_example.py
```diff
@@ -44,7 +44,7 @@
 
 # Python
 # <snippet name="docs/docusaurus/docs/snippets/get_existing_data_asset_from_existing_datasource_pandas_filesystem_example.py build_batch_request_with_dataframe">
-my_batch_request = my_asset.build_batch_request(dataframe=dataframe)
+my_batch_request = my_asset.build_batch_request(options={"dataframe": dataframe})
```
tyler-hoffman, line 47: First impression here: does this interface put us in a place where things are less predictable? Should there just be a different method for building a batch request via dataframe?

tests/datasource/fluent/integration/test_integration_datasource.py
```diff
@@ -525,19 +542,124 @@ def test_spark_data_adding_dataframe_in_file_reloaded_context(
...
+class ContextSparkDataSourceAndFrame:
+    context: EphemeralDataContext
+    datasource: SparkDatasource
+    dataframe: SparkDataFrame
+
+
+def _validate_whole_dataframe_batch_definition(
```
tyler-hoffman, line 625: nit: it feels weird to have this method defined in between its uses. Obviously non-blocking.

tyler-hoffman: Slightly more of a nit, but also not blocking: this pattern of putting the asserts and some other other non-setup stuff in a helper function makes the test somewhat harder to reason about, and likely invites overloading even more logic into here. I guess some things I'd prefer in tests:
* `assert`s live in the test itself, not a helper method
* I don't want to have to read the helper implementation to know what expectation my data is tested on.

tyler-hoffman (approved):
This looks good to me!

docs/docusaurus/docs/snippets/get_existing_data_asset_from_existing_datasource_pandas_filesystem_example.py
cdkini: Hmm I like the unified method call perhaps we can add a specialized method based on need?

tests/datasource/fluent/integration/test_integration_datasource.py
billdirks: Yes, this tests should really be parameterized and the helper functions should be inline. I can follow up with that fix.

docs/docusaurus/docs/snippets/get_existing_data_asset_from_existing_datasource_pandas_filesystem_example.py
billdirks: We need to have a better runtime story. This is a stopgap for the short term. I think currently `options` and other parameters exist and this throws if you pass them. This adds consistency. It also  prevents you from adding dataframes onto the asset which had some weird consequences.
But overall, I agree that this isn't ideal.

PR #10176, opened 2024-08-06
[MAINTENANCE]  Revert "[MAINTENANCE] Temporarily update Pandas pins to unblock V1 prerelease"

billdirks:
Reverts great-expectations/great_expectations#10175


billdirks:
I'm going to close this. We need to do some other work to undo this pin.

PR #10203, opened 2024-08-13
[DOCS] Puts 1.0 connect to SQL data code snippets in the documentation under test.

Rachel-Reverie:
## Description:
- puts code for "connect to SQL data" documentation under test.

## Definition of done:


docs/docusaurus/docs/core/connect_to_data/sql_data/_create_a_data_asset/create_a_data_asset.py
```diff
@@ -6,16 +6,15 @@
...
-    csv_path="./data/yellow_tripdata_sample_2019-01.csv",
-    connection_string="postgresql+psycopg2://postgres:@localhost/test_ci",
+# The following import and setup method sets up the data and environment used to test this example.
+# It can be disregarded by anyone using referencing this script as an example of a
+# GX workflow.
+from docs.docusaurus.docs.components._testing.utility_scripts.postgres_data_setup import (
+    setup,
 )
 
+setup()
```
billdirks, line 16: This validates that the snippet runs without dying. You may want to put in an assert after each snippet ends to so some validation that we are seeing what we expect.

For example after the `retrieve data source` snippet you could add `assert data_source.name == datasource_name`

billdirks (approved):
Thanks! Added a comment about adding an asserts after the end of each snippet. See the example inline. It can be very basic and adds a little more safety than "we didn't crash".

PR #10204, opened 2024-08-13
[MAINTENANCE] Ensure GXCloudStoreBackend session is closed

anthonyburdi:
To avoid memory leaks, we want to make sure that the session created in `GXCloudStoreBackend` is closed when the object is no longer needed and then garbage collected. We use `weakref.finalize` [python docs](https://docs.python.org/3.11/library/weakref.html#weakref.finalize) to add a callback that is executed when the `GXCloudStoreBackend` is garbage collected. This callback closes the connection.

Using a context manager (`with`) statement here is not possible since the session is created in the `GXCloudStoreBackend` constructor (`__init__`) and is needed for the lifetime of the `GXCloudStoreBackend`. We also considered making `GXCloudStoreBackend` a context manager but that would be a much larger and potentially breaking code change. We also considered using a custom `__del__` method (not recommended) or dependency injection (larger code changes, potentially breaking).


shiplet (approved):
🚀 🚀

PR #10206, opened 2024-08-13
[MAINTENANCE] Ensure CloudDataStore session is closed

anthonyburdi:
To avoid memory leaks, we want to make sure that the session created in CloudDataStore is closed when the object is no longer needed and then garbage collected. We use weakref.finalize [python docs](https://docs.python.org/3.11/library/weakref.html#weakref.finalize) to add a callback that is executed when the CloudDataStore is garbage collected. This callback closes the connection.

Using a context manager (with) statement here is not possible since the session is created in the CloudDataStore constructor (__init__) and is needed for the lifetime of the CloudDataStore. We also considered making CloudDataStore a context manager but that would be a much larger and potentially breaking code change. We also considered using a custom __del__ method (not recommended) or dependency injection (larger code changes, potentially breaking).

NOTE: This PR also changes the GXCloudStoreBackend finalizer to use an unbound method which is recommended in the python docs (see link in the code).

### DIFF
--- a/great_expectations/experimental/metric_repository/cloud_data_store.py
+++ b/great_expectations/experimental/metric_repository/cloud_data_store.py
@@ -68,13 +68,13 @@ class CloudDataStore(DataStore[StorableTypes]):
         super().__init__(context=context)
         assert context.ge_cloud_config is not None
         assert self._context.ge_cloud_config is not None
-        self.session = create_session(
+        self._session = create_session(
             access_token=context.ge_cloud_config.access_token,
             retry_count=0,  # Do not retry on authentication errors
         )
         # Finalizer to close the session when the object is garbage collected.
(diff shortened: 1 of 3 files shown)


great_expectations/experimental/metric_repository/cloud_data_store.py
```diff
@@ -70,6 +72,9 @@ def __init__(self, context: CloudDataContext):
             access_token=context.ge_cloud_config.access_token,
             retry_count=0,  # Do not retry on authentication errors
         )
+        # Finalizer to close the session when the object is garbage collected.
```
Shinnnyshinshin, line 75: Thank you for these notes. They are really helpful 🙇🏼

shiplet (approved):

PR #10211, opened 2024-08-14
[MAINTENANCE] Enable Codecov Test Result Reporting

Kilo59:
Start uploading test result reports to Codecov.
This allows us to more easily track flaky and slow tests.

image

- See also https://github.com/great-expectations/cloud/pull/394


Kilo59:
The only thing I'm not sure about is if these uploads are counted per the `after_n_builds` `codecov.yaml` directive.

tyler-hoffman (approved):

PR #10212, opened 2024-08-14
[DOCS] Transition feedback Jira tickets to kanban board

JessSaavedra:
## Description
Jira tickets created from feedback modal were first located in the intake board and had to be manually moved to "to do" list so they appear in the docs kanban board. This PR is to automatically move them by changing their status once they're created so this doesn't have to be done manually

## Video
[Grabación de pantalla desde 14-08-24 14:52:47.webm](https://github.com/user-attachments/assets/1862efaa-8ff9-4120-aac1-52a4ee944608)

### DIFF
--- a/docs/docusaurus/netlify/functions/createJiraTicketInDocsBoard.js
+++ b/docs/docusaurus/netlify/functions/createJiraTicketInDocsBoard.js
@@ -48,11 +48,15 @@ exports.handler = async (req, context) => {
         const result = await response.json()
 
         if (result.id) {
-            await httpPostRequest(`${JIRA_ISSUE_ENDPOINT_URL}/${result.id}/transitions`, {
-                "transition": {
-                    "id": TO_DO_STATE_ID
-                }
-            })
+            try {
+                await httpPostRequest(`${JIRA_ISSUE_ENDPOINT_URL}/${result.id}/transitions`, {
+                    "transition": {
+                        "id": TO_DO_STATE_ID
+                    }
+                })
+            } catch (error) {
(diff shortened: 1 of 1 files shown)


docs/docusaurus/netlify/functions/createJiraTicketInDocsBoard.js
```diff
@@ -14,34 +15,44 @@ const fullDescription = (description, name, email, selectedValue) => {
...
+                },
+                "labels": [
+                    "feedback_modal"
+                ],
+            }
         })
 
         const result = await response.json()
 
+        await httpPostRequest(`${JIRA_ISSUE_ENDPOINT_URL}/${result.id}/transitions`, {
```
josectobar: Let's confirm that the result contains an ID before we make the second request. Otherwise, we should throw an error. I think there could be cases when we can get a response with no ID if we hit some different request code like 400

josectobar (changes requested):
It looks great, just a small change request

docs/docusaurus/netlify/functions/createJiraTicketInDocsBoard.js
JessSaavedra: If we get a request code that's not "ok" like 400, we're already throwing that error by returning the `response.status` in the return of the function so I don't think we need some extra logic to do that :thinking: I'm definitely adding the check of having `result.id` to make the second request though, as not to make an invalid request to the API.

josectobar (approved):

PR #10213, opened 2024-08-14
[FEATURE] Update domain object `save()` methods to upsert instead of updating

cdkini:
We should cascade saves from parents down to their children

### DIFF
--- a/great_expectations/checkpoint/checkpoint.py
+++ b/great_expectations/checkpoint/checkpoint.py
@@ -271,7 +271,7 @@ class Checkpoint(BaseModel):
                 ValidationDefinitionRelatedResourceSaveError,
             ) as e:
                 raise CheckpointRelatedResourceSaveError(
-                    name=self.name,
+                    checkpoint_name=self.name,
                     related_resource_type=ValidationDefinition.__class__.__name__,
                     related_resource_name=validation.name,
                 ) from e
--- a/great_expectations/checkpoint/checkpoint.py
+++ b/great_expectations/checkpoint/checkpoint.py
@@ -29,7 +29,10 @@ from great_expectations.data_context.types.resource_identifiers import (
     ExpectationSuiteIdentifier,
     ValidationResultIdentifier,
 )
-from great_expectations.exceptions.exceptions import CheckpointRunWithoutValidationDefinitionError
+from great_expectations.exceptions.exceptions import (
+    CheckpointRunWithoutValidationDefinitionError,
+    StoreBackendError,
+)
 from great_expectations.render.renderer.renderer import Renderer
 
 if TYPE_CHECKING:
@@ -256,10 +259,24 @@ class Checkpoint(BaseModel):
     def save(self) -> None:
         from great_expectations.data_context.data_context.context_factory import project_manager
 
+        for validation in self.validation_definitions:
+            try:
+                validation.save()
+            except ValueError as e:
+                raise ValueError(
+                    f"Could not save Checkpoint due to failure to save child ValidationDefinition '{validation.name}'"
+                ) from e
+
         store = project_manager.get_checkpoints_store()
         key = store.get_key(name=self.name, id=self.id)
 
-        store.update(key=key, value=self)
+        try:
+            if store.has_key(key):
+                store.update(key=key, value=self)
+            else:
+                store.add(key=key, value=self)
+        except (ValueError, StoreBackendError) as e:
+            raise ValueError(f"Could not save Checkpoint '{self.name}'") from e
 
     def _add_to_store(self) -> None:
         """This is used to persist a checkpoint before we run it.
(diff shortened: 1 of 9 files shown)


great_expectations/exceptions/exceptions.py
```diff
@@ -48,6 +48,11 @@ class ExpectationSuiteError(DataContextError):
     pass
 
 
+class ExpectationSuiteSaveError(ExpectationSuiteError):
+    def __init__(self, name: str) -> None:
+        super().__init__(message=f"Could not save ExpectationSuite '{name}'")
```
cdkini, line 53: I don't know if I love these error types - could use suggestions on naming! Are they too granular?

great_expectations/checkpoint/checkpoint.py
```diff
@@ -154,7 +161,7 @@ def run(
             raise CheckpointRunWithoutValidationDefinitionError()
 
         if not self.id:
-            self._add_to_store()
+            self.save()
```
cdkini: Should we save each time no matter what? Is there a possibility of persisting a checkpoint and not having the children persisted? I think so right?

great_expectations/checkpoint/checkpoint.py
```diff
@@ -256,23 +263,32 @@ def _sort_actions(self) -> List[CheckpointAction]:
...
-        store.add(key=key, value=self)
+        try:
+            if store.has_key(key):
+                store.update(key=key, value=self)
+            else:
+                store.add(key=key, value=self)
+        except (
+            StoreBackendError,  # Generic error raised by store backends (namely GXCloudStoreBackend) # noqa: E501
+            ValueError,  # Generic error by certain stores/store backends
+        ) as e:
```
cdkini: I hate having to do this but different stores and store backends raise different things and we don't have uniformity around error handling or types

great_expectations/checkpoint/checkpoint.py
billdirks: Yes, we should persist each time.

great_expectations/core/validation_definition.py
```diff
@@ -201,7 +208,7 @@ def run(
         run_id: RunIdentifier | None = None,
     ) -> ExpectationSuiteValidationResult:
         if not self.id:
-            self._add_to_store()
+            self.save()
```
cdkini: Same note here

great_expectations/checkpoint/checkpoint.py
```diff
@@ -256,23 +263,32 @@ def _sort_actions(self) -> List[CheckpointAction]:
...
-        store.update(key=key, value=self)
-
-    def _add_to_store(self) -> None:
-        """This is used to persist a checkpoint before we run it.
-
-        We need to persist a checkpoint before it can be run. If user calls runs but hasn't
-        persisted it we add it for them.
-        """
-        from great_expectations.data_context.data_context.context_factory import project_manager
+        for validation in self.validation_definitions:
```
billdirks, line 265: If we want to be careful about rollbacks and not leaving us in an inbetween state, we'd need to keep track of current state so we can restore if we hit an exception part way through.

great_expectations/core/batch_definition.py
```diff
@@ -73,7 +74,18 @@ def get_batch(self, batch_parameters: Optional[BatchParameters] = None) -> Batch
...
 
     def save(self) -> None:
-        self.data_asset._save_batch_definition(self)
+        from great_expectations.data_context.data_context.context_factory import project_manager
+
+        datasource = self.data_asset.datasource
+        datasource_dict = project_manager.get_datasources()
+
+        try:
+            datasource_dict[datasource.name] = datasource
```
billdirks: Does this work because we have a `DatasourceDict` which overrides `__setitem__`?
This seems strange since everywhere else we set things up in memory and then call save. But we can't change this now.
If we have access to the context, we could call `context.add_or_update_datasource` I believe.

great_expectations/exceptions/exceptions.py
billdirks: We are suppose to err on the side of granularity and let people doing the error aggregation cloud feature group as needed.

great_expectations/core/validation_definition.py
```diff
@@ -285,20 +292,34 @@ def identifier_bundle(self) -> _IdentifierBundle:
...
-        key = store.get_key(name=self.name, id=self.id)
-
-        store.update(key=key, value=self)
-
-    def _add_to_store(self) -> None:
-        """This is used to persist a validation_definition before we run it.
-
-        We need to persist a validation_definition before it can be run. If user calls runs but
-        hasn't persisted it we add it for them."""
+        try:
```
billdirks, line 294: Similar issue about rollbacks here if there is an error. Let's chat sync.

PR #10214, opened 2024-08-14
[MAINTENANCE] Add required keys to SuiteValidationResult.meta

joshua-stauffer:
Early in the V1 work, we removed most fields from SuiteValidationResult.meta. We've since discovered that gx cloud depends on several of them, so this PR adds them back.

### DIFF
--- a/tests/core/test_validation_definition.py
+++ b/tests/core/test_validation_definition.py
@@ -9,6 +9,7 @@ import pytest
 
 import great_expectations as gx
 import great_expectations.expectations as gxe
+from great_expectations import __version__ as GX_VERSION
 from great_expectations.core.batch_definition import BatchDefinition
 from great_expectations.core.expectation_suite import ExpectationSuite
 from great_expectations.core.expectation_validation_result import (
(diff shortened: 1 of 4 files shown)


tests/core/test_validation_definition.py
```diff
@@ -214,6 +236,10 @@ def test_adds_requisite_ids(
         assert output.meta == {
             "validation_id": validation_definition.id,
             "checkpoint_id": checkpoint_id,
+            "batch_spec": ACTIVE_BATCH_SPEC,
```
billdirks, line 239: It looks like this test uses a `mock_validator`. Does this just insert these values? Do we need a test where we run a real validation and verify we see these fields?

billdirks (commented):
Thanks for tracking this down! I have 1 question about tests.

tests/core/test_validation_definition.py
joshua-stauffer: the test in `test_checkpoint.py` is a full integration test that runs a real validation. i could break it out into its own test, if that's clearer.

billdirks (approved):
LGTM, thanks!

PR #10215, opened 2024-08-14
[MAINTENANCE] Get tests around rendered content passing

tyler-hoffman:
This breaks up an old xfailed test, and gets it passing. The pre-existing test tested the following cases:
* When rendered content shouldn't show (which is now in non-cloud contexts)
* When rendered content should show (which is not in cloud contexts)
* When there were errors in the renderer (had to slightly update the custom expectation for this, as well as a bit of internal logic)
* Legacy expectation renderers. No longer supported, and we don't believe they were used in cloud.

The implementation changes were largely driven by seeing that we were now passing through empty lists, where we once passed None, and removing unneeded logic along with that.


great_expectations/render/renderer/inline_renderer.py
```diff
@@ -215,30 +215,3 @@ def get_rendered_content(
         )
 
         return self._get_atomic_rendered_content_for_object(render_object=render_object)
-
-    @staticmethod
-    def replace_or_keep_existing_rendered_content(
```
tyler-hoffman, line 220: Talked to @NathanFarmer and we agreed this should be safe to remove. This function was largely just to abstract away replacing missing rendered_content with new rendered content, plus a bit of extra logic about the failed renderer_types that seem to have been preempting a situation that isn't driven by needs of the business logic.

tests/data_context/test_data_context.py
```diff
@@ -691,156 +693,107 @@ def _validate(  # type: ignore[override]
...
+        )
+        expectation_suite = empty_data_context.suites.get(name=suite_name)
 
-    # Once we include_rendered_content, we get rendered_content on each ExpectationConfiguration in the ExpectationSuite.  # noqa: E501
-    expectation_suite = context.suites.get(name=expectation_suite_name)
-    for expectation_configuration in expectation_suite.expectation_configurations:
         assert all(
-            isinstance(rendered_content_block, RenderedAtomicContent)
-            for rendered_content_block in expectation_configuration.rendered_content
+            expectation.rendered_content is None for expectation in expectation_suite.expectations
```
NathanFarmer, line 709: It sounds like this wasn't part of this PR, but does the `rendered_content` attribute even need to exist in non-cloud contexts?

NathanFarmer (approved):

PR #10217, opened 2024-08-15
[MAINTENANCE] Raise informative errors if child objects are not persisted before parent

cdkini:


great_expectations/checkpoint/checkpoint.py
```diff
@@ -153,8 +153,9 @@ def run(
         if not self.validation_definitions:
             raise CheckpointRunWithoutValidationDefinitionError()
 
-        if not self.id:
-            self._add_to_store()
+        saved, errs = self.is_saved()
+        if not saved:
+            raise ValueError(errs)
```
cdkini: TODO: Use a custom type that aggregates errors and does a good job of displaying them in a readable manner

great_expectations/checkpoint/checkpoint.py
```diff
@@ -252,6 +253,22 @@ def _sort_actions(self) -> List[CheckpointAction]:
...
+
+        all_validations_saved = True
+        for validation_definition in self.validation_definitions:
+            validation_saved, validation_errs = validation_definition.is_saved()
+            errs.extend(validation_errs)
+            all_validations_saved = validation_saved and all_validations_saved
+
+        self_saved = self.id is not None
+        if not self_saved:
+            errs.append(f"Please save Checkpoint '{self.name}' before continuing.")
```
cdkini: TODO: Add helpful API tips to errors so users know how to remedy their code

great_expectations/checkpoint/checkpoint.py
```diff
@@ -153,8 +158,13 @@ def run(
         if not self.validation_definitions:
             raise CheckpointRunWithoutValidationDefinitionError()
 
-        if not self.id:
-            self._add_to_store()
+        added, errors = self.is_added()
+        if not added:
+            # The checkpoint itself is not added but all children are - we can add it for the user
+            if len(errors) == 1 and isinstance(errors[0], CheckpointNotAddedError):
+                self._add_to_store()
```
cdkini, line 165: Feels a lil ugly but the core logic seems sound

great_expectations/exceptions/exceptions.py
```diff
@@ -34,29 +34,33 @@ def __str__(self) -> str:
         return self.message
 
 
-class SuiteEditNotebookCustomTemplateModuleNotFoundError(ModuleNotFoundError):
-    def __init__(self, custom_module) -> None:
-        message = f"The custom module '{custom_module}' could not be found"
-        super().__init__(message)
```
cdkini, line 40: Legacy remnant I thought to remove while I was here

great_expectations/exceptions/exceptions.py
```diff
@@ -90,6 +103,14 @@ def __init__(self) -> None:
         )
 
 
+class CheckpointRelatedResourcesNotAddedError(ResourcesNotAddedError):
```
cdkini, line 106: Maybe some of this naming could be cleaned up

great_expectations/checkpoint/checkpoint.py
```diff
@@ -153,8 +158,13 @@ def run(
         if not self.validation_definitions:
             raise CheckpointRunWithoutValidationDefinitionError()
 
-        if not self.id:
-            self._add_to_store()
+        added, errors = self.is_added()
+        if not added:
+            # The checkpoint itself is not added but all children are - we can add it for the user
+            if len(errors) == 1 and isinstance(errors[0], CheckpointNotAddedError):
```
billdirks, line 164: I think `errors == [CheckpointNotAddedError]` is a simpler check

great_expectations/checkpoint/checkpoint.py
```diff
@@ -252,6 +262,21 @@ def _sort_actions(self) -> List[CheckpointAction]:
...
         return priority_actions + secondary_actions
 
+    def is_added(self) -> tuple[bool, list[ResourceNotAddedError]]:
+        errs: list[ResourceNotAddedError] = []
+
+        validations_added: bool = True
+        for validation_definition in self.validation_definitions:
+            validation_added, validation_errs = validation_definition.is_added()
+            errs.extend(validation_errs)
+            validations_added = validation_added and validations_added
```
billdirks, line 272: You could factor the top part of this function into it's own one: `are_dependencies_added` and call that here and above. Then you wouldn't need to check for a list with only `CheckpointNotAddedError` above.

great_expectations/core/validation_definition.py
```diff
@@ -114,6 +118,21 @@ def asset(self) -> DataAsset:
     def data_source(self) -> Datasource:
         return self.asset.datasource
 
+    def is_added(self) -> tuple[bool, list[ResourceNotAddedError]]:
```
billdirks, line 121: Similar comment about refactoring the child/dependencies part into it's own function that this calls and updating the check in `run`.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -252,6 +262,21 @@ def _sort_actions(self) -> List[CheckpointAction]:
...
+
+        self_added = self.id is not None
+        if not self_added:
+            errs.append(CheckpointNotAddedError(name=self.name))
+
+        return (validations_added and self_added, errs)
```
billdirks, line 278: I am fine with returning this as a tuple. Other alternatives include defining a class or dataclass.

billdirks (commented):
Thanks for doing this. A small request about making a separate `are_deps_added` method or updating the check to be `== [Error]` instead of a len and isinstance check.

great_expectations/checkpoint/checkpoint.py
cdkini: I was wary of adding too much public surface area since a user working with these objects can just tab complete and see everything on it - perhaps a minor issue but I'd vote to keep it like this for now and clean up later if we desire?

great_expectations/checkpoint/checkpoint.py
cdkini: I think I'm going to put it in a dataclass just so the signature has some context - what's a good name 🤔

great_expectations/checkpoint/checkpoint.py
cdkini: Actually if we have `are_dependencies_added`, do we even need `is_added`?

cdkini: My thought here is to has `is_added` return a custom dataclass and then have a property or method on the dataclass that determines whether children are persisted

billdirks: For checkpoints no, but for validation_definitions yes. In terms of autocomplete, I'd just prefix with a `_` since we should only be calling the `_are_dependencies_added` internal to this class.

tests/checkpoint/test_checkpoint.py
```diff
@@ -419,16 +427,20 @@ class TestCheckpointResult:
...
         suite = mocker.Mock(spec=ExpectationSuite)
         suite.name = self.suite_name
+        suite.is_added.return_value = (True, [])
         return suite
 
     @pytest.fixture
     def mock_batch_def(self, mocker: MockerFixture):
-        return mocker.Mock(spec=BatchDefinition)
+        bd = mocker.Mock(spec=BatchDefinition)
+        bd._copy_and_set_values().is_added.return_value = (True, [])
```
cdkini, line 436: Unfortunately need to mock this in this manner due to Pydantic - not sure how to get around it

great_expectations/checkpoint/checkpoint.py
cdkini: I think this strategy makes error aggregation a little annoying though - if dependencies are not added and neither is the parent, we need to take the output of this function and manually append the `CheckpointNotAddedError`.

```python
deps_added, errors = self._are_dependencies_added()
if not deps_added:
    if not self.id:
       ... # Add additional error
    raise 
```

tests/datasource/fluent/_fake_cloud_api.py
```diff
@@ -401,6 +403,15 @@ def put_datasource_cb(request: PreparedRequest) -> CallbackResult:
     parsed_url = urllib.parse.urlparse(request.url)
     datasource_id = parsed_url.path.split("/")[-1]
 
+    # Ensure that assets and batch definitions get IDs
+    # Note that this logic should also happen in POST but is not implemented for our fake
```
cdkini, line 407: We may want to add this to our POST cb (even though assets are never added with the initial creation in our test suite)

great_expectations/checkpoint/checkpoint.py
cdkini: Just keeping it as a tuple for simplicity's sake

great_expectations/checkpoint/checkpoint.py
cdkini: `self.is_added()` already compiles those errors

billdirks: So you're saying it's equivalent since at the call site now you subtract that error when doing the check, ie `if len(errors) == 1 and isinstance(errors[0], CheckpointNotAddedError)`. I guess that's fair.

great_expectations/checkpoint/checkpoint.py
billdirks: I see, this equality isn't quite right since you have an instance of this Error class.

tests/datasource/fluent/_fake_cloud_api.py
```diff
@@ -384,7 +386,7 @@ def post_datasources_cb(
         )
 
 
-def put_datasource_cb(request: PreparedRequest) -> CallbackResult:
+def put_datasource_cb(request: PreparedRequest) -> CallbackResult:  # noqa: C901
```
billdirks, line 389: nit: You can avoid the noqa by putting the new functionality into a helper function.

billdirks (approved):
LGTM, thanks for doing this. I had 1 nit comment about avoiding a noqa in tests.

PR #10219, opened 2024-08-15
[FEATURE] Filter bad validation definitions and checkpoints coming back through stores.

billdirks:

### DIFF
--- a/great_expectations/data_context/store/store.py
+++ b/great_expectations/data_context/store/store.py
@@ -13,9 +13,11 @@ from typing import (
     Type,
 )
 
+from marshmallow import ValidationError as MarshmallowValidationError
 from typing_extensions import TypedDict
 
 import great_expectations.exceptions as gx_exceptions
+from great_expectations.compatibility.pydantic import ValidationError as PydanticValidationError
 from great_expectations.core.data_context_key import DataContextKey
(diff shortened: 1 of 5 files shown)


great_expectations/data_context/store/store.py
```diff
@@ -233,7 +233,18 @@ def get_all(self) -> list[Any]:
         if self.cloud_mode:
             objs = self.gx_cloud_response_json_to_object_collection(objs)
 
-        return list(map(self.deserialize, objs))
+        deserializable_objs: list[Any] = []
+        bad_objs: list[Any] = []
+        for obj in objs:
+            try:
+                deserializable_objs.append(self.deserialize(obj))
+            except Exception:  # We want to filter out all undeserializable configs
```
joshua-stauffer: i know that this is in the top level store, so deserialize will have an arbitrary implementation, but i think we want to actually just catch `ValidationError` here. Is there a known use case in which we need to catch a non-pydantic error?

billdirks: Thanks! I am looking at ExpectationSuites and they use Marshmallow so this will be a marshmallow `ValidationError`. I will restrict to those 2 types.

joshua-stauffer (approved):
thanks, looks good!

PR #10221, opened 2024-08-16
[BUGFIX] Ensure that data source and nested objects obtain IDs on add for all environments

cdkini:
Datasources/assets/batch definitions do not gets ids when added with file or ephemeral contexts - this change should fix that

### DIFF
--- a/great_expectations/core/batch_definition.py
+++ b/great_expectations/core/batch_definition.py
@@ -87,7 +87,7 @@ class BatchDefinition(pydantic.GenericModel, Generic[PartitionerT]):
         )
         batch_definition_bundle = _IdentifierBundle(
             name=self.name,
-            id=self.id,
+            id=str(self.id) if self.id else None,
         )
 
         return _EncodedValidationData(
--- a/great_expectations/datasource/datasource_dict.py
+++ b/great_expectations/datasource/datasource_dict.py
@@ -171,6 +171,8 @@ class CacheableDatasourceDict(DatasourceDict):
         return ds
 
     def _add_ids(self, ds: FluentDatasource) -> FluentDatasource:
+        # File and ephemeral contexts do not use the store, so we need to add IDs here.
+        # Note that this is used for both `add` and `update` operations.
         if ds.id is None:
             ds.id = uuid.uuid4()
         for asset in ds.assets:
--- a/great_expectations/datasource/datasource_dict.py
+++ b/great_expectations/datasource/datasource_dict.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import uuid
 from collections import UserDict
 from typing import TYPE_CHECKING, Protocol, TypeVar, runtime_checkable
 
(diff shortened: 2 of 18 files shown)


great_expectations/datasource/datasource_dict.py
```diff
@@ -70,10 +70,7 @@ def data(self) -> dict[str, FluentDatasource]:  # type: ignore[override] # `data
 
         configs = self._datasource_store.get_all()
         for config in configs:
-            if isinstance(config, FluentDatasource):
-                name = config.name
```
cdkini, line 74: These checks are just confusing at this point - removing as we should only have FDS

tests/datasource/fluent/test_viral_snippets.py
```diff
@@ -29,6 +29,7 @@
 logger = logging.getLogger(__file__)
 
 
+@pytest.mark.xfail(reason="YAML file shares state with other tests")
```
cdkini: I don't love this but I'd ask to come back to this once I wrap up the `save` work - we add ids from another test I believe and then hit an assertion error here because we don't expect those values

tests/datasource/fluent/test_metadatasource.py
```diff
@@ -403,7 +403,22 @@ def assert_fluent_datasource_content(
...
+    assert isinstance(config_from_gx_yaml, dict)
+    config_from_gx_yaml_without_ids = _remove_ids(config_from_gx_yaml)
+    assert config_from_gx_yaml_without_ids == fluent_datasource_config
+
+
+def _remove_ids(config: dict) -> dict:
+    for data_source in config.values():
+        data_source.pop("id")
+        for asset in data_source.get("assets", {}).values():
+            print(asset)
```
billdirks: stray print?

tests/datasource/fluent/test_viral_snippets.py
billdirks: We discussed a path forward here.

great_expectations/datasource/datasource_dict.py
```diff
@@ -170,20 +167,26 @@ def __contains__(self, name: object) -> bool:
 
     @override
     def set_datasource(self, name: str, ds: FluentDatasource) -> FluentDatasource | None:
-        self.data[name] = ds
+        self.data[name] = self._add_ids(ds)
+        return ds
+
+    def _add_ids(self, ds: FluentDatasource) -> FluentDatasource:
+        # File and ephemeral contexts do not use the store, so we need to add IDs here.
+        # Note that this is used for both `add` and `update` operations.
```
billdirks, line 175: We should have ids when we update going forward but we need to do this on update now since we may have configs stored without ids?

billdirks (commented):
Thanks! Looks good overall. I have a question inline and would rather not skip that test.

great_expectations/core/batch_definition.py
```diff
@@ -87,7 +87,7 @@ def identifier_bundle(self) -> _EncodedValidationData:
         )
         batch_definition_bundle = _IdentifierBundle(
             name=self.name,
-            id=self.id,
+            id=str(self.id) if self.id else None,
```
billdirks, line 90: Ugh, looks like for everything else we use `str` or `str | UUID` (but must always be str) for an `id`.

great_expectations/datasource/datasource_dict.py
cdkini: If I update a datasource, it may already have an id but any newly added children or grandchildren may not

billdirks: That makes sense, thanks!

billdirks (approved):
Thanks! I wish we were consistent with our id types, but that falls outside the scope of this. LGTM.

great_expectations/core/batch_definition.py
cdkini: Ugh yeah we could update the identifier bundle to use UUIDs but this is the path of least resistance

PR #10228, opened 2024-08-17
[DOCS] Puts scripts for 1.0 "create a Data Context" docs under test

Rachel-Reverie:
## Description
- Puts scripts for 1.0 "create a Data Context" docs under test
- Adds instructions for instantiating a File Data Context from a specific folder.

## Definition of done


docs/docusaurus/docs/core/set_up_a_gx_environment/_create_a_data_context/cloud_data_context.py
```diff
@@ -20,3 +20,5 @@
 print(context)
 # </snippet>
 # </snippet>
+
+assert type(context).__name__ == "CloudDataContext"
```
tyler-hoffman, line 24: not that it matters, but I'd usually expect this to just be `assert isinstance(context, CloudDataContext)`

tyler-hoffman (approved):

docs/docusaurus/docs/core/set_up_a_gx_environment/_create_a_data_context/cloud_data_context.py
Rachel-Reverie: It probably should be, but we also suggested using `print(type(context).__name__)` to verify what kind of Data Context when `gx.get_context()` is used without specifying a mode, and I think I just copied it from there.

PR #10240, opened 2024-08-20
[DOCS] Puts 1.0 example code for how to configure project settings under test

Rachel-Reverie:
## Description
- Puts the 1.0 docs examples under test for:
   - How to configure Metadata Stores
   - How to toggle Analytics Events
- Pulls the example of using string substitution for a PostgreSQL connection string from a connect to data example script that is currently under test.

## Definition of done


docs/docusaurus/docs/core/configure_project_settings/_examples/configure_metadata_stores.py
```diff
@@ -0,0 +1,56 @@
...
+
+# Load a File Data Context
+context = gx.get_context(mode="file")
+# </snippet>
+# Hide this
+set_up_context_for_example(context)
+
+# Access the Stores through the Data Context's `variables` attribute:
+# <snippet name="docs/docusaurus/docs/core/configure_project_settings/_examples/configure_metadata_stores.py - access Metadata Store configurations">
+print(context.variables.config.stores["expectations_store"])
```
tyler-hoffman, line 26: I wonder if we should use `stores.get("expectations_store")` etc for these. I know your code is for file contexts, but I first tried this line with a cloud context and it blew up. We have some store configs, but not all in cloud.

tyler-hoffman (approved):

tyler-hoffman (approved):

docs/docusaurus/docs/core/configure_project_settings/_examples/configure_metadata_stores.py
Rachel-Reverie: When I wrote this initially I was given the impression that for GX Cloud we expect people to use the GX Cloud stores -- after all, part of the benefit to GX Cloud is that all of the backend storage is handled for you.

tyler-hoffman: Yeah, I think that's the correct assumption. I was just surprised to run that line of code and see it raise. But I think you're fine to keep it as is.

PR #10243, opened 2024-08-20
[DOCS] Puts 1.0 doc examples for how to configure Data Docs into scripts under test

Rachel-Reverie:
## Description
- Puts the docs examples for how to configure Data Docs into python scripts.
- Puts the local filesystem version under test.

## Definition of done:


docs/docusaurus/docs/core/configure_project_settings/configure_data_docs/_backends/_s3.md
```diff
@@ -0,0 +1,21 @@
+An Amazon S3 Data Doc site requires the following `store_backend` information:
+
+- `class_name`: The name of the class to implement for accessing the target environment.  For Amazon S3 this will be `TupleS3StoreBackend`.
+- `bucket`: The name of the Amazon S3 bucket that will host the Data Docs site.
+- `prefix`: The path of the folder that will contain the Data Docs pages relative to the root of the Amazon S3 bucket.  The combination of `container` and `bucket` must be unique accross all Stores used by a Data Context.
+- `boto3_options`: The credentials for accessing your Amazon S3 account.  Amazon S3 supports multiple methods of providing credentials, such as use of an endpoint url, access key, or role assignment.  For more information on how to configure your Amazon S3 credentials, see Amazon's documentation for how to [Configure the AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html).
```
tyler-hoffman, line 6: I think as long as we link to aws docs, we're good here. I haven't tried too many options with boto3_options though.

docs/docusaurus/docs/core/configure_project_settings/configure_data_docs/_examples/data_docs_abs.py
```diff
@@ -0,0 +1,100 @@
...
+    "site_index_builder": {"class_name": "DefaultSiteIndexBuilder"},
+}
+
+# Add a Store backend configuration to the Data Docs configuration
+# <snippet name="docs/docusaurus/docs/core/configure_project_settings/configure_data_docs/_examples/data_docs_abs.py - add store backend">
+container = "my_abs_container"
+prefix = "data_docs/"
+connection_string = "${AZURE_STORAGE_CONNECTION_STRING}"  # This uses string substitution to get the actual connection string from an environment variable or config file.
+
+site_config["store_backend"] = {
```
tyler-hoffman: Tyler's preference alert: I find it way harder to reason about when we mutate objects, e.g. site_config is only partially built above and then we add another key to it here. I want one place to grok the shape of it.

tyler-hoffman (approved):
To the best of my knowledge, this is all correct. I might add something about prefix being optional (I think that's true in all of these).

docs/docusaurus/docs/core/configure_project_settings/configure_data_docs/_examples/data_docs_abs.py
Rachel-Reverie: That's fair.  I originally split it because the store_backend is the only part that differs between environments, but I can merge the two sections back together easily enough now that I have sample scripts for each environment to pull snippets from.

PR #10245, opened 2024-08-21
[BUGFIX] Use `is_added` checks in `identifier_bundle` serialization logic

cdkini:

### DIFF
--- a/great_expectations/core/batch_definition.py
+++ b/great_expectations/core/batch_definition.py
@@ -76,16 +76,16 @@ class BatchDefinition(pydantic.GenericModel, Generic[PartitionerT]):
 
         return batch_list[-1]
 
-    def is_added(self) -> tuple[bool, list[ResourceNotAddedError]]:
+    def is_added(self) -> tuple[bool, ResourceNotAddedError | None]:
         if self.id:
             return True, []
-        return False, [BatchDefinitionNotAddedError(name=self.name)]
+        return False, BatchDefinitionNotAddedError(name=self.name)
 
     def identifier_bundle(self) -> _EncodedValidationData:
         # Utilized as a custom json_encoder
-        added, errors = self.is_added()
+        added, error = self.is_added()
         if not added:
-            raise errors[0]
+            raise error
 
         asset = self.data_asset
         data_source = asset.datasource
--- a/great_expectations/core/batch_definition.py
+++ b/great_expectations/core/batch_definition.py
@@ -76,16 +76,16 @@ class BatchDefinition(pydantic.GenericModel, Generic[PartitionerT]):
 
         return batch_list[-1]
 
-    def is_added(self) -> tuple[bool, ResourceNotAddedError | None]:
+    def is_added(self) -> tuple[bool, list[ResourceNotAddedError]]:
         if self.id:
             return True, []
-        return False, BatchDefinitionNotAddedError(name=self.name)
+        return False, [BatchDefinitionNotAddedError(name=self.name)]
 
     def identifier_bundle(self) -> _EncodedValidationData:
         # Utilized as a custom json_encoder
-        added, error = self.is_added()
+        added, errors = self.is_added()
         if not added:
-            raise error
+            raise errors[0]  # Should only be one as this is a leaf node
 
         asset = self.data_asset
         data_source = asset.datasource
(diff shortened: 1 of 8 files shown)


great_expectations/core/batch_definition.py
```diff
@@ -76,13 +76,17 @@ def get_batch(self, batch_parameters: Optional[BatchParameters] = None) -> Batch
...
             return True, []
-        return False, [BatchDefinitionNotAddedError(name=self.name)]
+        return False, BatchDefinitionNotAddedError(name=self.name)
 
     def identifier_bundle(self) -> _EncodedValidationData:
         # Utilized as a custom json_encoder
+        added, error = self.is_added()
+        if not added:
+            raise error
+
```
cdkini: I was missing these checks in the `identifier_bundle` custom JSON encoder func that we use when serializing - previously I only had these checks in `run()` workflows

great_expectations/core/batch_definition.py
```diff
@@ -76,13 +76,17 @@ def get_batch(self, batch_parameters: Optional[BatchParameters] = None) -> Batch
 
         return batch_list[-1]
 
-    def is_added(self) -> tuple[bool, list[ResourceNotAddedError]]:
+    def is_added(self) -> tuple[bool, ResourceNotAddedError | None]:
```
cdkini: Deciding to make leaf nodes just return a single instance instance of a list to make it easier to raise (alternative would have been `raise errors[0]` and that feels bad)

billdirks: I prefer to leave this as a list so we can define a simpler protocol. If we change this, then a call to `foo.is_added` will need to know whether foo is an internal node or a leaf node and will have to process the results differently even though the implementer doesn't really care if they are different.  For example, below, you now use `append` instead of `extend`, but it is `extend` in other places like in checkpoints. If we really wanted to make this symmetric everywhere, I think we would always wrap the errors returned instead of returning `errors[0]`. But I think that might be unnecessary at this time, so I think `errors[0]` is ok.

billdirks (commented):
Thanks for doing this so quickly! One comment inline.

billdirks (approved):
This is an improvement. Thank you! 
It recurses now into an error but doesn't show all errors. Here is an example:
```
import pandas as pd

mode="ephemeral"
cnt=1

context = gx.get_context(mode=mode)

expectation = gx.expectations.ExpectColumnValuesToBeBetween(column="passenger_count", min_value=1, max_value=6)
datasource = context.data_sources.add_pandas(name=f"my_pandas_ds_{cnt}")
asset = datasource.add_dataframe_asset(name=f"my_pandas_asset_{cnt}")

suite = gx.ExpectationSuite(name=f"my_suite_{cnt}")
suite.add_expectation(expectation)

bd = asset.add_batch_definition_whole_dataframe(name=f"my_bd_{cnt}")
vd = gx.ValidationDefinition(name=f"my_vd_{cnt}", data=bd, suite=suite)
vd2 = gx.ValidationDefinition(name=f"another_vd_{cnt}", data=bd, suite=suite)

# All the recursive errors in vd are present but there are no vd2 errors present.
cp = context.checkpoints.add(gx.Checkpoint(name=f"my_cp_{cnt}", validation_definitions=[vd, vd2]))
#ValidationDefinitionRelatedResourcesNotAddedError:
#	ExpectationSuite 'my_suite_1' must be added to the DataContext before it can be updated. Please call `context.suites.add(<SUITE_OBJECT>)`, then try your action again.
#	ValidationDefinition 'my_vd_1' must be added to the DataContext before it can be updated. Please call `context.validation_definitions.add(<VALIDATION_DEFINITION_OBJECT>)`, then try your action again.
```

tests/core/test_batch_definition.py
```diff
@@ -127,20 +125,34 @@ def test_get_batch_raises_error_with_empty_batch_list(mocker: pytest_mock.MockFi
...
+    ds = context.data_sources.add_pandas("pandas_datasource")
     asset = ds.add_csv_asset("my_asset", "data.csv")
     batch_definition = asset.add_batch_definition("my_batch_definition")
 
-    assert batch_definition.identifier_bundle() == _EncodedValidationData(
-        datasource=_IdentifierBundle(name="pandas_datasource", id=None),
-        asset=_IdentifierBundle(name="my_asset", id=None),
-        batch_definition=_IdentifierBundle(name="my_batch_definition", id=None),
+    result = batch_definition.identifier_bundle()
+    assert result.datasource.name == "pandas_datasource" and result.datasource.id is not None
```
tyler-hoffman, line 135: I have no strong opinions here, but why not just make these separate asserts?

tyler-hoffman (approved):

PR #10249, opened 2024-08-21
[MAINTENANCE] Add `AddedDiagnostics` helper class to `is_added` flows

cdkini:


great_expectations/core/added_diagnostics.py
```diff
@@ -0,0 +1,112 @@
...
+
+    errors: list[ResourceNotAddedError]
+
+    @property
+    def is_added(self) -> bool:
+        return len(self.errors) == 0
+
+    @abstractmethod
+    def raise_for_error(self) -> None:
+        """
```
tyler-hoffman, line 42: So this will either raise or potentially return None?

cdkini: Yup akin to `requests.Response.raise_for_status`

cdkini: could do something like:
```
if not diagnostics.is_added:
    raise diagnostics.exception()
```
but the conditional check will be the same with every use

great_expectations/checkpoint/checkpoint.py
```diff
@@ -158,13 +166,13 @@ def run(
...
-        if not added:
+        diagnostics = self.is_added()
+        if not diagnostics.is_added:
             # The checkpoint itself is not added but all children are - we can add it for the user
-            if len(errors) == 1 and isinstance(errors[0], CheckpointNotAddedError):
+            if not diagnostics.parent_added and diagnostics.children_added:
                 self._add_to_store()
             else:
-                raise CheckpointRelatedResourcesNotAddedError(errors=errors)
+                diagnostics.raise_for_error()
```
tyler-hoffman, line 175: @cdkini this line being in a conditional is the reason I was surprised below that it didn't always raise. Not saying anything needs to change. Just thinking aloud.

tyler-hoffman (approved):
I think this looks reasonable. One thing that we'll maybe have to think about in the future (or maybe you already thought about this) is that something can exist and have an id, but have been deleted. I think it's a weird state we don't have to solve for today.

PR #10277, opened 2024-08-26
[MAINTENANCE] Check that `BatchDefinitions` and `ExpectationSuites` are up-to-date before saving

cdkini:


great_expectations/core/batch_definition.py
```diff
@@ -79,9 +81,24 @@ def get_batch(self, batch_parameters: Optional[BatchParameters] = None) -> Batch
...
         )
+        if diagnostics.is_added:
+            diagnostics = self._is_fresh()
+        return diagnostics
+
+    def _is_fresh(self) -> BatchDefinitionAddedDiagnostics:
+        datasource_dict = project_manager.get_datasources()
+        datasource = datasource_dict[self.data_asset.datasource.name]
+        asset = datasource.get_asset(self.data_asset.name)
+        batch_def = asset.get_batch_definition(self.name)
```
cdkini: Error handling here? Unlikely but what happens if the parent/asset/batch_def doesn't exist? Should we return the same error, a different one, or perhaps panic?

great_expectations/core/batch_definition.py
```diff
@@ -79,9 +81,24 @@ def get_batch(self, batch_parameters: Optional[BatchParameters] = None) -> Batch
         return batch_list[-1]
 
     def is_added(self) -> BatchDefinitionAddedDiagnostics:
-        return BatchDefinitionAddedDiagnostics(
+        diagnostics = BatchDefinitionAddedDiagnostics(
             errors=[] if self.id else [BatchDefinitionNotAddedError(name=self.name)]
         )
+        if diagnostics.is_added:
```
cdkini: Is `is_added` the best name still?

great_expectations/expectations/expectation.py
```diff
@@ -337,6 +337,20 @@ def schema_extra(schema: Dict[str, Any], model: Type[Expectation]) -> None:
...
+            return False
+
+        self_dict = self.dict()
+        other_dict = other.dict()
+
+        for attr in ("notes", "meta"):
+            self_dict.pop(attr, None)
+            other_dict.pop(attr, None)
+
+        return self_dict == other_dict
```
cdkini: This feels bad but I'm seeing some kind of mismatch with `meta` being persisted inconsistently as `None`/`{}`

great_expectations/core/expectation_suite.py
```diff
@@ -249,9 +250,20 @@ def save(self) -> None:
...
             errors=[] if self.id else [ExpectationSuiteNotAddedError(name=self.name)]
         )
+        if not diagnostics.is_added:
+            return diagnostics
+        return self._is_fresh()
+
+    def _is_fresh(self) -> ExpectationSuiteAddedDiagnostics:
+        key = self._store.get_key(name=self.name, id=self.id)
+        suite_dict = self._store.get(key=key)
+        suite = ExpectationSuite(**suite_dict)
```
cdkini: The store should be returning this rich type but it doesn't 😕 

I can follow up with this change

great_expectations/core/expectation_suite.py
```diff
@@ -105,7 +107,9 @@ def __init__(  # noqa: PLR0913
         self.meta = meta
         self.notes = notes
 
-        self._store = project_manager.get_expectations_store()
+    @property
+    def _store(self) -> ExpectationsStore:
+        return project_manager.get_expectations_store()
```
cdkini, line 116: We do some deepcopying in the validator that doesn't actually copy this attr over (latent bug) - we can remedy this by grabbing it just-in-time through a property

great_expectations/core/batch_definition.py
```diff
@@ -83,9 +85,24 @@ def get_batch(self, batch_parameters: Optional[BatchParameters] = None) -> Batch
         return batch_list[-1]
 
     def is_added(self) -> BatchDefinitionAddedDiagnostics:
-        return BatchDefinitionAddedDiagnostics(
+        diagnostics = BatchDefinitionAddedDiagnostics(
             errors=[] if self.id else [BatchDefinitionNotAddedError(name=self.name)]
         )
+        if not diagnostics.is_added:
+            return diagnostics
+        return self._is_fresh()
```
billdirks: I think we should reverse `_is_fresh` and `is_added`. That is, the fresh check should verify that it's added and that it is fresh.

great_expectations/core/batch_definition.py
billdirks: If the fresh/added is reversed, our `diagnostics` should be expanded to include this error.

great_expectations/core/batch_definition.py
```diff
@@ -83,9 +85,24 @@ def get_batch(self, batch_parameters: Optional[BatchParameters] = None) -> Batch
...
+
+    def _is_fresh(self) -> BatchDefinitionAddedDiagnostics:
+        datasource_dict = project_manager.get_datasources()
+        datasource = datasource_dict[self.data_asset.datasource.name]
+        asset = datasource.get_asset(self.data_asset.name)
+        batch_def = asset.get_batch_definition(self.name)
+
+        return BatchDefinitionAddedDiagnostics(
+            errors=[]
+            if self == batch_def
```
billdirks: Does this equality check work or do we have to define `__eq__`?

great_expectations/core/expectation_suite.py
billdirks: There is a comment about this in `suite_factory.py.get()`. There are errors that can happen here because, even though the suite, has been validated with marshmallow, the expectations inside have not been validated. Doing this will instantiate the expectations and do the pydantic validation.
So, if we need to do this here, we need to be a bit careful and handle `PydanticValidationError`s that can occur. See https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/core/factory/suite_factory.py#L115

great_expectations/core/expectation_suite.py
```diff
@@ -249,9 +253,20 @@ def save(self) -> None:
...
     def is_added(self) -> ExpectationSuiteAddedDiagnostics:
-        return ExpectationSuiteAddedDiagnostics(
+        diagnostics = ExpectationSuiteAddedDiagnostics(
             errors=[] if self.id else [ExpectationSuiteNotAddedError(name=self.name)]
         )
+        if not diagnostics.is_added:
+            return diagnostics
+        return self._is_fresh()
+
+    def _is_fresh(self) -> ExpectationSuiteAddedDiagnostics:
```
billdirks: Don't we need to add the corresponding changes to `ValidationDefinition` and `Checkpoint`?

great_expectations/expectations/expectation.py
```diff
@@ -336,6 +336,20 @@ def schema_extra(schema: Dict[str, Any], model: Type[Expectation]) -> None:
...
 
+    @override
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Expectation):
+            return False
+
+        self_dict = self.dict()
+        other_dict = other.dict()
+
+        for attr in ("notes", "meta"):
```
billdirks: This will change the objects in place. We'll need to make copies if you do this, eg
```
almost_equal = dict(self_dict, {"notes": None, "meta": None}) == dict(other_dict, {"notes": None, "meta": None})
```
I would do follow up checks on `notes` and `meta` to check their equality allowing for the equivalence of `None` and `{}`.

great_expectations/expectations/expectation.py
billdirks: See comment above.

great_expectations/core/batch_definition.py
billdirks: Does the `datasource_dict` reach out to cloud to get the latest?

billdirks (commented):
Thanks for doing this. I left questions inline.

great_expectations/core/batch_definition.py
billdirks: pydantic does this for us.

great_expectations/core/expectation_suite.py
billdirks: This will happen in a follow up PR.

great_expectations/core/batch_definition.py
billdirks: It does reach out to cloud.

great_expectations/core/batch_definition.py
```diff
@@ -82,14 +84,32 @@ def get_batch(self, batch_parameters: Optional[BatchParameters] = None) -> Batch
...
             errors=[] if self.id else [BatchDefinitionNotAddedError(name=self.name)]
         )
 
+    def _is_fresh(self) -> BatchDefinitionFreshnessDiagnostics:
+        datasource_dict = project_manager.get_datasources()
+
+        # TODO: Add error handling
+        datasource = datasource_dict[self.data_asset.datasource.name]
+        asset = datasource.get_asset(self.data_asset.name)
+        batch_def = asset.get_batch_definition(self.name)
```
cdkini: NOTE - Add try/except logic and aggregate errors here

great_expectations/core/expectation_suite.py
```diff
@@ -248,11 +252,27 @@ def save(self) -> None:
...
+    def _is_added(self) -> ExpectationSuiteFreshnessDiagnostics:
+        return ExpectationSuiteFreshnessDiagnostics(
             errors=[] if self.id else [ExpectationSuiteNotAddedError(name=self.name)]
         )
 
+    def _is_fresh(self) -> ExpectationSuiteFreshnessDiagnostics:
+        # TODO: Add error handling here
+        key = self._store.get_key(name=self.name, id=self.id)
+        suite_dict = self._store.get(key=key)
+        suite = ExpectationSuite(**suite_dict)
```
cdkini: NOTE - Add try/except logic and aggregate errors here

cdkini: NOTE - Use logic from suite factory to proper deserialize here

great_expectations/exceptions/exceptions.py
```diff
@@ -42,14 +52,8 @@ class ResourceNotAddedError(DataContextError):
...
-class ResourcesNotAddedError(ValueError):
-    def __init__(self, errors: list[ResourceNotAddedError]) -> None:
-        self._errors = errors
-        super().__init__("\n\t" + "\n\t".join(str(e) for e in errors))
-
-    @property
-    def errors(self) -> list[ResourceNotAddedError]:
-        return self._errors
+class ResourcesNotAddedError(GreatExpectationsAggregateError):
+    pass
```
cdkini, line 56: Going to rename all these `NotAdded` errors to `NotFresh` in a separate PR - I'd like to be pretty thoughtful about the error hierarchy we have here it feels like it can get pretty unwieldy quick

great_expectations/exceptions/exceptions.py
```diff
@@ -65,6 +69,14 @@ def __init__(self, name: str) -> None:
         )
 
 
+class ExpectationSuiteNotFreshError(ResourceNotAddedError):
```
cdkini, line 72: Each of suite, batch def, validation def, and checkpoint should have two error types:
- not added
- not fresh

They should all be derived from the same base class (probably `ResourceNotFreshError`?)

great_expectations/core/freshness_diagnostics.py
```diff
@@ -18,59 +17,52 @@
 
 
 @dataclass
-class AddedDiagnostics:
+class FreshnessDiagnostics:
```
cdkini, line 20: "Freshness" is more appropriate moving forward as it also implicitly encapsulates "addition" of an object to a store

great_expectations/data_context/store/expectations_store.py
```diff
@@ -341,6 +341,12 @@ def deserialize(self, value):  # type: ignore[explicit-override] # FIXME
         else:
             raise TypeError(f"Cannot deserialize value of unknown type: {type(value)}")  # noqa: TRY003
 
+    def deserialize_suite_dict(self, suite_dict: dict) -> ExpectationSuite:
+        suite = ExpectationSuite(**suite_dict)
+        if suite._include_rendered_content:
+            suite.render()
+        return suite
```
cdkini, line 348: Small step towards moving this logic here - if I override `deserialize`, things seem to break? This seems like a viable solution for now but I don't like having multiple "deserialize" methods here

great_expectations/core/batch_definition.py
```diff
@@ -76,14 +82,70 @@ def get_batch(self, batch_parameters: Optional[BatchParameters] = None) -> Batch
...
         )
 
+    def _is_fresh(self) -> BatchDefinitionFreshnessDiagnostics:
+        datasource_dict = project_manager.get_datasources()
+
+        datasource: Datasource | None = None
+        try:
+            datasource = datasource_dict[self.data_asset.datasource.name]
+        except KeyError:
+            pass
```
cdkini: This logic needs tests - does it make sense to capture all these errors in aggregate or should we panic and alert the user upon hitting the first one?

great_expectations/core/batch_definition.py
```diff
@@ -76,14 +81,62 @@ def get_batch(self, batch_parameters: Optional[BatchParameters] = None) -> Batch
...
         )
 
+    def _is_fresh(self) -> BatchDefinitionFreshnessDiagnostics:
+        datasource_dict = project_manager.get_datasources()
+
+        datasource: Datasource | None
+        try:
+            datasource = datasource_dict[self.data_asset.datasource.name]
+        except KeyError:
+            datasource = None
```
billdirks, line 102: You could change this try/except to `datasource = datasource_dict.get(self.data_asset.datasource.name, None)` since this is a dict. This way is consistent with the other calls though. 🤷

great_expectations/core/expectation_suite.py
```diff
@@ -248,11 +256,44 @@ def save(self) -> None:
...
+            suite_dict = self._store.get(key=key)
+        except StoreBackendError:
+            suite_dict = None
+        if not suite_dict:
+            return ExpectationSuiteFreshnessDiagnostics(
+                errors=[ExpectationSuiteNotFoundError(name=self.name)]
+            )
+
+        suite: ExpectationSuite | None
+        if suite_dict:
```
billdirks: You don't need this `if` since you will have returned on line 278 if we don't have a `suite_dict`.

great_expectations/core/freshness_diagnostics.py
```diff
@@ -10,67 +9,67 @@
...
-@dataclass
-class _ParentAddedDiagnostics(AddedDiagnostics):
-    parent_error_class: ClassVar[Type[ResourceNotAddedError]]
-    children_error_classes: ClassVar[Tuple[Type[ResourceNotAddedError], ...]]
-    raise_for_error_class: ClassVar[Type[ResourcesNotAddedError]]
+    parent_error_class: ClassVar[Type[GreatExpectationsError]]
+    children_error_classes: ClassVar[Tuple[Type[GreatExpectationsError], ...]]
 
-    def update_with_children(self, *children_diagnostics: AddedDiagnostics) -> None:
+    def update_with_children(self, *children_diagnostics: FreshnessDiagnostics) -> None:
```
billdirks, line 72: This comment isn't requesting a change but is a thought I wanted to share. The `+` operator on lists creates a new list each time. These lists are small so it's likely fine. Another way to do this would be to reverse the self.errors, append the child errors, and then reverse this.

great_expectations/data_context/store/expectations_store.py
billdirks: Can you file a ticket to combine these? Or is there already a ticket to move the logic out of the factory and into the store?

great_expectations/exceptions/exceptions.py
billdirks: Since there is a lot of them, I would move all the exceptions related to freshness into their own file. I would also include a module level docstring (eg a docstring at the top of the file) that has an explanation of the hierarchy (ie the comment you made in slack).

tests/checkpoint/test_checkpoint.py
```diff
@@ -1087,27 +1089,39 @@ def test_checkpoint_run_adds_requisite_ids(self, tmp_path: pathlib.Path):
     ],
 )
 @pytest.mark.unit
-def test_is_added(
+def test_is_fresh(
```
billdirks, line 1092: There are 2 ways a domain object can be stale:

1. A user has grabbed the domain object but hasn't saved.
2. A user has grabbed the domain object and hasn't changed it (or has and saved it) but then it changed out from under them by someone else.

Ultimately we should tests both cases. It seems like we can't for an ephemeral but we could for a filesystem context (by changing the file outside of gx api calls) or a cloud context.

If we don't add tests for case 2 now, we should track and add them as a follow up.

tests/core/test_expectation_suite.py
```diff
@@ -93,14 +93,6 @@ def suite_with_single_expectation(
 class TestInit:
     """Tests related to ExpectationSuite.__init__()"""
 
-    @pytest.mark.unit
```
billdirks, line 96: Why remove this test?

billdirks (approved):
This looks good! I have a few comments. One is about breaking up these exceptions into their own module and one is about tests. They could be deferred to a future PR or added to this one.

tests/checkpoint/test_checkpoint.py
cdkini: yup I've added these tests for the leaf nodes (suites and batch defs) and will do so for checkpoints and validation definitions when I get there!

tests/core/test_expectation_suite.py
cdkini: we no longer grab the store in the `__init__` so we won't hit this error - this is a change in behavior but it more in line with how the rest of the domain objects work (you'll hit the data context dependency error when trying to use the object rather than at instantiation time)

PR #10286, opened 2024-08-27
[DOCS] Updates method for checking the version of the installed GX Core library

Rachel-Reverie:
## Description
- See title
- Puts example under test

## Definition of done


rachhouse:
@Rachel-Reverie _Try GX Core_ is still using the `<ReleaseVersionBox/>` component (of `import ReleaseVersionBox from '../../components/versions/_gx_version_code_box.mdx'`) to display the expected output from the version check.

I see that _Install GX_ now states for output "If GX was installed correctly, the version number of the installed GX library will be printed.", which I think is better than the current `<ReleaseVersionBox/>` output.

* Do we still need the `<ReleaseVersionBox/>` component?
* Would you like to update _Try GX Core_ to match the new _Install GX_ text as part of this PR, or would you like me to do it with another small PR once this is merged (I'm fine with either).

Thanks!

Rachel-Reverie:
> @Rachel-Reverie _Try GX Core_ is still using the `<ReleaseVersionBox/>` component (of `import ReleaseVersionBox from '../../components/versions/_gx_version_code_box.mdx'`) to display the expected output from the version check.
> 
> I see that _Install GX_ now states for output "If GX was installed correctly, the version number of the installed GX library will be printed.", which I think is better than the current `<ReleaseVersionBox/>` output.
> 
> * Do we still need the `<ReleaseVersionBox/>` component?
> * Would you like to update _Try GX Core_ to match the new _Install GX_ text as part of this PR, or would you like me to do it with another small PR once this is merged (I'm fine with either).
> 
> Thanks!

I think it's fine for us to drop the <ReleaseVersionBox/> component -- it was only particularly useful in that it allowed us to insert a version number into a code block dynamically.  I'll update the _Try GX Core_ text in this PR, no need for us to make another one.

rachhouse (approved):
LGTM!

PR #10295, opened 2024-08-27
[FEATURE] Replace get_batch_list_from_batch_request with get_batch and get_batch_identifiers_list

tyler-hoffman:
## Goals
For pandas filesystem datasources, we've had a long-standing performance issue where we read all of the data for batches, even though we'll only ever use one. This ticket resolves that, but replacing `get_batch_list_from_batch_request` with just `get_batch`, The interface is essentially equivalent, but rather than returning a list of Batches, and letting the caller pick the last one, we now just only fetch what would be the last batch.

Some users called `get_batch_list_from_batch_request` to see the available batch metadata, so we don't want to break that behavior. To replace that workflow, we're exposing `get_batch_identifiers_list`, which returns the corresponding metadata.

## Notes on the PR
DataAsset is now an ABC; before it was treated like one, but wasn't actually. This meant updating a couple dummy assets used in tests to "implement" the new methods.

Lots of docs used the old method. These have been updated to use `get_batch`. Q: Are those docs actually used right now? They are somewhat misleading, since these are not in the public api.

Lots of tests were updated. I tried to use reasonable discretion in how to proceed. A bunch of the tests were making assertions about batches and batch specs. In some places, I just updated the the existing assertions and rather than just call `get_batch_list_from_batch_request`, I called both `get_batch` and `get_batch_identifiers_list` and made reasonable assertions. Probably not how I'd greenfield the tests, but in the interest of time, I figured it wasn't worth breaking things up more

There's some logic around sorting that was pretty gnarly before this PR. We want to preserve that sorting, but now apply it to both batch metadata and legacy batch defs, so I've now added an additional layer to make it work in both cases. Open to feedback if there's a more readable approach.


great_expectations/core/batch_definition.py
```diff
@@ -69,14 +69,8 @@ def get_batch(self, batch_parameters: Optional[BatchParameters] = None) -> Batch
...
             A Batch of data.
         """
-        batch_list = self.data_asset.get_batch_list_from_batch_request(
-            self.build_batch_request(batch_parameters=batch_parameters)
-        )
-
-        if len(batch_list) == 0:
-            raise ValueError("No batch found")  # noqa: TRY003
-
-        return batch_list[-1]
```
tyler-hoffman, line 83: The asset is already doing the sorting, so moving the logic about picking the last batch to the asset as well.

great_expectations/experimental/datasource/fabric.py
```diff
@@ -117,7 +116,7 @@ def get_batch_list_from_batch_request(self, batch_request: BatchRequest) -> list
             batch_request=batch_request
         )
 
-        batch_list.append(
+        return [
```
tyler-hoffman: Same as above - just making this more clear that we only return one element.

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -153,7 +152,7 @@ def get_batch_list_from_batch_request(self, batch_request: BatchRequest) -> list
             batch_request=batch_request, ignore_options=("dataframe",)
         )
 
-        batch_list.append(
+        return [
```
tyler-hoffman: This function only ever returns one batch, so updated the method setup to reflect that.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -509,53 +518,86 @@ def _order_by_validator(
...
-                return -1
+def _sort_batch_identifiers_with_none_metadata_values(
+    get_val: Callable[[_T], Any],
+) -> Callable[[_T, _T], int]:
+    def _compare_function(a: _T, b: _T) -> int:
+        a_val = get_val(a)
+        b_val = get_val(b)
 
-            if a.metadata[key] > b.metadata[key]:
+        if a_val is not None and b_val is not None:
```
tyler-hoffman, line 586: I wish we only supported versions of python with match case!

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -1324,7 +1324,7 @@ def get_batch_list(  # noqa: PLR0913
         batch_spec_passthrough: Optional[dict] = None,
         batch_parameters: Optional[Union[dict, BatchParameters]] = None,
         **kwargs: Optional[dict],
-    ) -> List[Batch]:
+    ) -> Batch:
```
billdirks, line 1330: Can we rename this function and the `_` version to `get_batch` since it no longer returns a list?

great_expectations/datasource/fluent/data_asset/path/path_data_asset.py
```diff
@@ -116,55 +116,62 @@ def _validate_batch_request(self, batch_request: BatchRequest) -> None:
             )
 
     @override
-    def get_batch_list_from_batch_request(self, batch_request: BatchRequest) -> List[Batch]:
-        """A list of batches that match the BatchRequest.
+    def get_batch_identifiers_list(self, batch_request: BatchRequest) -> List[dict]:
```
billdirks, line 120: Anything with a batch request is internal gx api only. Can we make a public method that returns all the batch identifiers? eg `get_batch_identifiers(self)` which calls this with an empty batch request? Or maybe the public method should be on the batch_definition so we can get sorting there?
This can be done in a followup PR but we should think about how best to make this public.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -509,53 +515,86 @@ def _order_by_validator(
...
         """
+
+        def get_value(key: str) -> Callable[[Batch], Any]:
+            return lambda bd: bd.metadata[key]
+
+        return self._sort_batch_data_list(batch_list, partitioner, get_value)
+
+    def sort_legacy_batch_definitions(
+        self,
+        batch_definition_list: List[LegacyBatchDefinition],
```
billdirks: Can we rename parameter to `legacy_batch_definition_list` to not get it confused with our new batch definitions.

great_expectations/data_context/data_context/abstract_data_context.py
billdirks: All these `get_batch` methods really return the last batch if the batch request is underspecified. Maybe `get_last_batch` is a more appropriate name?

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -580,76 +580,87 @@ def _fully_specified_batch_requests(self, batch_request: BatchRequest) -> List[B
...
 
         Args:
             batch_request: A batch request for this asset. Usually obtained by calling
                 build_batch_request on the asset.
 
         Returns:
-            A list of batches that match the options specified in the batch request.
+            A list Batch that matches the options specified in the batch request.
         """
+        from great_expectations.core import IDDict
```
billdirks: Can we import these at the top of the file?

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -580,76 +580,87 @@ def _fully_specified_batch_requests(self, batch_request: BatchRequest) -> List[B
...
-                datasource_name=self.datasource.name,
-                data_connector_name=_DATA_CONNECTOR_NAME,
-                data_asset_name=self.name,
-                batch_identifiers=IDDict(batch_spec["batch_identifiers"]),
-                batch_spec_passthrough=None,
+        else:
+            sorted_metadata_dicts = unsorted_metadata_dicts
+        sorted_metadata_dicts = sorted_metadata_dicts[batch_request.batch_slice]
+        batch_metadata = sorted_metadata_dicts[-1]
+        request_index = unsorted_metadata_dicts.index(batch_metadata)
```
billdirks, line 638: Can you add a comment that requests and unsorted_metadata_dicts or in the same order which is why we can get the correct request_index from this lookup.

tests/core/test_batch_definition.py
```diff
@@ -64,64 +79,15 @@ def test_get_batch_retrieves_only_batch(mocker: pytest_mock.MockFixture):
...
 
     # Assert
     assert batch == mock_batch
-    mock_get_batch_list_from_batch_request.assert_called_once_with(
-        batch_definition.build_batch_request()
-    )
-
-
-@pytest.mark.unit
-def test_get_batch_retrieves_last_batch(mocker: pytest_mock.MockFixture):
```
billdirks, line 83: Is there another test that asserts we get the last batch when the batch_request is underspecified?

tests/core/test_batch_definition.py
```diff
@@ -64,64 +79,15 @@ def test_get_batch_retrieves_only_batch(mocker: pytest_mock.MockFixture):
...
-
-    # Assert
-    assert batch == batch_b
-    mock_get_batch_list_from_batch_request.assert_called_once_with(
-        batch_definition.build_batch_request()
-    )
-
-
-@pytest.mark.unit
-def test_get_batch_raises_error_with_empty_batch_list(mocker: pytest_mock.MockFixture):
```
billdirks, line 107: We no longer need this test?

billdirks (commented):
It looks good. I had a few questions I wanted to discuss before approving.

Either in this PR or in a followup we should decide what we want our `@public_api` method to be that exposes this to a user.

great_expectations/datasource/fluent/data_asset/path/path_data_asset.py
tyler-hoffman: Yeah, I was thinking about those options as well. My gut is that we should put it on batch definitions. I'll follow up on that.

tyler-hoffman: https://greatexpectations.atlassian.net/jira/software/projects/V1/boards/76?selectedIssue=V1-505

tests/core/test_batch_definition.py
tyler-hoffman: Yeah, as far as I can tell, we should be good there. A lot of the existing tests that used to test `get_batch_list_from_batch_request`were updated to test this. And just to be sure I added a net new `test_get_batch_partitioner__sort_ascending_respected` test mostly because I wanted something targeted to prove I didn't mess up that request index piece you asked for a comment on.

tests/core/test_batch_definition.py
tyler-hoffman: The setup here (mocking an empty list returned no longer made sense), but your comment made me realized I should have backfilled tests around this at the asset level. Done in. https://github.com/great-expectations/great_expectations/pull/10295/commits/2774a349510a29500fe437bd75a164d43cc112fe.

billdirks (approved):
LGTM, thanks!

PR #10299, opened 2024-08-28
[DOCS] Indicate named parameters in reference table by including the value types

Rachel-Reverie:
## Description
- See title

## Definition of done

### DIFF
--- a/docs/docusaurus/docs/core/connect_to_data/sql_data/_create_a_data_source/_datasource_method_reference_table.md
+++ b/docs/docusaurus/docs/core/connect_to_data/sql_data/_create_a_data_source/_datasource_method_reference_table.md
@@ -1,7 +1,7 @@
-| Database type    | Data Context method                                                |
-|------------------|--------------------------------------------------------------------|
-| PostgreSQL       | `context.data_sources.add_postgres(name, connection_string)`       |
-| SQLite | `context.data_sources.add_sqlite(name, connection_string)`         |
-| DataBricks SQL   | `context.data_sources.add_databricks_sql(name, connection_string)` |
-| Snowflake        | `context.data_sources.add_snowflake(name, connection_string)`      |
-| Other SQL        | `context.data_sources.add_sql(name, connection_string)`            |
+| Database type    | Data Context method                                                          |
+|------------------|------------------------------------------------------------------------------|
+| PostgreSQL       | `context.data_sources.add_postgres(name: str, connection_string: str)`       |
+| SQLite | `context.data_sources.add_sqlite(name: str, connection_string: str)`         |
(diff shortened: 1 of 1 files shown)


docs/docusaurus/docs/core/connect_to_data/sql_data/_create_a_data_source/_datasource_method_reference_table.md
```diff
@@ -1,7 +1,7 @@
...
-| SQLite | `context.data_sources.add_sqlite(name, connection_string)`         |
-| DataBricks SQL   | `context.data_sources.add_databricks_sql(name, connection_string)` |
-| Snowflake        | `context.data_sources.add_snowflake(name, connection_string)`      |
-| Other SQL        | `context.data_sources.add_sql(name, connection_string)`            |
\ No newline at end of file
+| Database type    | Data Context method                                                          |
+|------------------|------------------------------------------------------------------------------|
+| PostgreSQL       | `context.data_sources.add_postgres(name: str, connection_string: str)`       |
+| SQLite | `context.data_sources.add_sqlite(name: str, connection_string: str)`         |
+| DataBricks SQL   | `context.data_sources.add_databricks_sql(name: str, connection_string: str)` |
```
rachhouse, line 5: Minor suggestion - due to Docusaurus width settings, the Databricks line breaks awkwardly at max width. You might want to consider updating it (possibly the others, for consistency) to use a code box and reduce line break weirdness at various page widths.
```
context.data_sources.add_databricks_sql(
    name: str,
    connection_string: str
)
```

rachhouse (approved):
@Rachel-Reverie Looks good to me! Just one small suggestion on proactive taming of weird line breaks, feel free to use it at your discretion.

docs/docusaurus/docs/core/connect_to_data/sql_data/_create_a_data_source/_datasource_method_reference_table.md
Rachel-Reverie: I'm going to merge this as-is and create a follow up ticket for improving this.  Unfortunately, getting a code box into a markdown table isn't supported (markdown doesn't allow multiline elements in a table cell other than using `<br/>` to split a single line of standard text into multiple lines when rendered)  or rebuilding a markdown table in html and maintaining consistent use of the CSS classes docusaurus automatically applies when formatting markdown can be a bit troublesome.  At that point it would be simpler to replace the table with a tab group that has a single tab for each Data Source type which contains the corresponding method signature -- and at _that_ point I would prefer to use an instance of example code for each method instead of just listing the signature, which would require a bit more lift to put under test.

PR #10318, opened 2024-08-30
[BUGFIX] FabricPowerBIDatasource 1.0 - TypeError: _PowerBIAsset.build_batch_request() got an unexpected keyword argument 'options'

tyler-hoffman:
Fix issue caused by `FabricPowerBIDatasource` not accepting any keyword arguments a part of `build_batch_request()`.

The PowerBI asset types do not support batching or partitioning so these passed in values will be `None` at runtime.


Kilo59 (approved):

great_expectations/experimental/datasource/fabric.py
```diff
@@ -114,7 +118,7 @@ def get_batch_list_from_batch_request(self, batch_request: BatchRequest) -> list
         )
 
         batch_metadata: BatchMetadata = self._get_batch_metadata_from_batch_request(
-            batch_request=batch_request
+            batch_request=batch_request, ignore_options=("dataframe",)
```
Kilo59, line 121: I don't think this is strictly necessary but I also don't see this causing any problems.

PR #10347, opened 2024-09-05
[MAINTENANCE] Add `description` field to `ExpectationConfigurationSchema` and render always `description` over `template_str`

NathanFarmer:
- If `description` is set on the Expectation, always render it in GX Cloud instead of the default renderer
- Don't serialize `description` in `ExpectationSuite` if the field as not been set
- Add a few serialization snapshots for `Expectation`
--------


great_expectations/expectations/core/unexpected_rows_expectation.py
```diff
@@ -122,15 +121,8 @@ def _prescriptive_template(
...
     ) -> RendererConfiguration:
-        add_param_args: AddParamArgs = (
-            ("description", RendererValueType.STRING),
-            ("unexpected_rows_query", RendererValueType.STRING),
-        )
-        for name, param_type in add_param_args:
-            renderer_configuration.add_param(name=name, param_type=param_type)
-
-        renderer_configuration.template_str = (
-            "$description" if renderer_configuration.params.description else ""
```
NathanFarmer, line 133: The individual renderers no longer need to add this logic. It is handled by `RendererConfiguration` automatically (if a `description` is set, that is what gets rendered).

great_expectations/render/renderer_configuration.py
```diff
@@ -503,7 +509,13 @@ def _get_row_condition_string(row_condition_str: str) -> str:
 
     @validator("template_str")
     def _set_template_str(cls, v: str, values: dict) -> str:
-        if values.get("_row_condition"):
+        if values["configuration"] and values["configuration"].description:
```
dctalbot, line 512:
```suggestion
        if values.get("configuration") and values["configuration"].description:
```
this may be a safer check to avoid KeyErrors at run time

dctalbot (approved):

great_expectations/render/renderer_configuration.py
NathanFarmer: Oops I missed this. Good call. I will create a small PR for this.

PR #10365, opened 2024-09-09
[MAINTENANCE] Add "freshness" checks for ValidationDefinitions and Checkpoints

cdkini:


tests/checkpoint/test_checkpoint.py
```diff
@@ -1150,37 +1140,50 @@ def test_checkpoint_run_with_slack_action_no_page_links(self, tmp_path: pathlib.
 @pytest.mark.unit
 def test_is_fresh(
     in_memory_runtime_context,
-    id: str | None,
-    validation_def_id: str | None,
-    suite_id: str | None,
-    batch_def_id: str | None,
-    is_fresh: bool,
+    id: bool,
```
tyler-hoffman: The names/types don't make sense to me here. `has_id`, etc maybe?

tyler-hoffman: Looking at the changes below, I think what existed before was more readable. Does that just not work with the new implementation?

tyler-hoffman (approved):
Some questions about the test changes, but nothing blocking. Otherwise makes sense to me!

tests/checkpoint/test_checkpoint.py
cdkini: Unfortunately not because overwriting an ID with a new UUID causes things to be out of sync (which is now important because we're retrieving with stores)

tyler-hoffman: Ah, gotcha. Makes sense!

PR #10385, opened 2024-09-12
[FEATURE] Add context.enable_analytics method

tyler-hoffman:
**WARNING** This PR requires changes in mercury to make analytics_enabled nullable.

Adds a method to explicitly set if analytics are enabled in the project config.

This changes behavior so that the value in the project config, if present, takes precedence over the environment variable. The rationale here is that it would be confusing to explicitly set this this variable, but then have the environment variable (if set to `False`) to override this variable. I think this is reasonable; you can have a global environment variable, but then have project-specific settings.

We've talked about solving all of the ways we need to get the current context in sync without an extra call to `get_context`, but I'm not sure there's a 1-size-fits-all solution without a pretty large lift, e.g. reworking stores. FWIW the main issue here was that the setup of our analytics was done in a call from the constructor, and didn't end up making sense to do in a property, as I'd hoped.


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -333,6 +334,17 @@ def _save_project_config(self) -> None:
...
+    @public_api
+    def enable_analytics(self, enable: Optional[bool]) -> None:
+        """
+        Enable or disable analytics for this DataContext.
+        With non-ephemeral contexts, this can be preserved via context.variables.save().
+
+        If set to None, the `GX_ANALYTICS_ENABLED` environment variable will be used.
+        """
+        self.config.analytics_enabled = enable
+        self._init_analytics()
```
cdkini, line 346: The rest of the variables are set with `context.variables.xyz` and `context.variables.save()` - do we care that we're deviating here?

great_expectations/data_context/templates.py
```diff
@@ -130,17 +130,5 @@ def dump(self, data, stream=None, **kw):  # type: ignore[explicit-override] # FI
 """
 )
 
-USAGE_STATISTICS_ENABLED = """
-analytics_enabled: True
-"""
-
-USAGE_STATISTICS_DISABLED = """
-analytics_enabled: False
-"""
```
cdkini, line 139: Why this removal?

tyler-hoffman: This field is what will show up in the config by default. The approach in this PR is for both the env var and the config variable to be optional bools, so this variable would not be present unless a user specifically sets it to True or False. Now the default value is handled directly in `_determine_analytics_enabled`.

It's been a few weeks since I looked at this, but I think there was also some additional complexity if we continued to set a default here in terms of implementation details. But the main thing is just philosophically treating this as None OR what a user explicitly selects.

great_expectations/data_context/data_context/abstract_data_context.py
tyler-hoffman: I originally wasn't too worried about deviating from the `context.variables.xyz = ` pattern, but this will still require a `context.variables.save()`, which is non-obvious. Maybe worth thinking about how to make it work with a property setter on `self.config.analytics_enabled`

cdkini: Could we just add the `context.variables.save()` call here (and perhaps do that for all variables)?

tyler-hoffman: Seems reasonable to me.

tyler-hoffman: Though it does differ from what we are doing with other models.

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -273,18 +273,19 @@ def _init_analytics(self) -> None:
...
-        config_file_enabled = self.config.analytics_enabled
-        if config_file_enabled is None:
-            config_file_enabled = True
+        config_enabled = self.config.analytics_enabled
         env_var_enabled = ENV_CONFIG.posthog_enabled
-        return config_file_enabled and env_var_enabled
+        if config_enabled is not None:
+            return config_enabled
+        elif env_var_enabled is not None:
+            return env_var_enabled
```
cdkini, line 286: This will always be non-null due to Pydantic coercion I think?

tyler-hoffman: Looks like it's still marshmallow, and is optional: https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/data_context/types/base.py#L1116

tyler-hoffman: oh I see. That's for the environment variable. Yeah, I guess that'd have to change as well if we otherwise want the behavior implied in this method.

tyler-hoffman: Updated.

cdkini (approved):

PR #10386, opened 2024-09-12
[MAINTENANCE] Ensure that all nested validation definition diagnostics are emitted from a parent checkpoint

cdkini:


tests/checkpoint/test_checkpoint.py
```diff
@@ -603,7 +603,6 @@ def test_checkpoint_run_no_actions(self, validation_definition: ValidationDefini
         assert validation_result.success is True
         assert len(validation_result.results) == 1 and validation_result.results[0].success is True
 
-        assert result.checkpoint_config == checkpoint
```
cdkini, line 615: This check breaks due to a store retrieval - not sure how much value it is to check that the checkpoint is part of the result but I can rework this if desired (currently hitting a mock so we have a "freshness" error)

joshua-stauffer (commented):
are there tests that specifically validate this logic?

tests/checkpoint/test_checkpoint.py
```diff
@@ -1307,3 +1306,145 @@ def test_is_fresh_raises_error_when_child_deps_not_found(in_memory_runtime_conte
...
+        )
+        return context.checkpoints.add(
+            Checkpoint(name=self.checkpoint_name, validation_definitions=[vd])
+        )
+
+    def _build_checkpoint_without_factories(self) -> Checkpoint:
+        bd = BatchDefinition(name=self.batch_definition_name)
+        suite = ExpectationSuite(name=self.suite_name)
+        vd = ValidationDefinition(name=self.validation_definition_name, data=bd, suite=suite)
+        return Checkpoint(name=self.checkpoint_name, validation_definitions=[vd])
```
Kilo59, line 1372: If your goal is to test checkpoints with and without factories, you can turn these a parametrized fixture, and then you can have a single test that automatically gets run with both types of checkpoints.
https://docs.pytest.org/en/latest/how-to/fixtures.html#fixture-parametrize

Kilo59 (approved):
Looks good to me.
Made a suggestion WRT the test helpers.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -126,6 +131,102 @@ class Config:
...
+    ) -> str:
+        """
+        Override the default json method to enable proper diagnostics around validation_definitions.
+
+        NOTE: This should be removed in favor of a field/model serializer when we upgrade to
+              Pydantic 2.
+        """
+        if not exclude:
+            exclude = set()
+        exclude = exclude.union({"validation_definitions"})
```
billdirks: I don't think a Mapping has a `union` operation.

tests/checkpoint/test_checkpoint.py
billdirks: If we don't have a test for the checkpoint_config being part of the result elsewhere, I think this is worthwhile.

billdirks (commented):
Thanks for doing this! I haven't looked at your new tests yet but left a few comments on the parts I have read.

great_expectations/checkpoint/checkpoint.py
Kilo59:
```suggestion
        if isinstance(exclude, Set):
           exclude = exclude.add("validation_definitions")
        else:
           exclude["__all__"] = "validation_definitions"
```
We need to account for the mapping.

Although there are still a few edge cases this leaves out like...
What if the `exclude` mapping object already has `__all__` key set.

https://docs.pydantic.dev/1.10/usage/exporting_models/#advanced-include-and-exclude

great_expectations/checkpoint/checkpoint.py
```diff
@@ -126,6 +131,107 @@ class Config:
...
+        diagnostics = CheckpointFreshnessDiagnostics(errors=[])
+        for validation_definition in self.validation_definitions:
+            try:
+                identifier_bundle = validation_definition.identifier_bundle()
+                data["validation_definitions"].append(identifier_bundle.dict())
+            except ResourceFreshnessAggregateError as e:
+                diagnostics.errors.extend(e.errors)
+
+        diagnostics.raise_for_error()
+        return data
```
Kilo59: nit: Neither of these uses `self`, so they could be `class` or `static` methods.

cdkini: I believe the latter one does but good call on the former!

PR #10392, opened 2024-09-13
[BUGFIX] Using `{batch}` keyword in `UnexpectedRowsQuery`

NathanFarmer:
- Fix a bug where using `{batch}` fails with Postgres because the dialect requires subquery aliases to be used on columns in SELECT and WHERE clauses
- Fix a bug where using `{batch}` with a splitter fails with all backends because compiling without `literal_binds` caused active batch splitter bind params to be variables instead of literal parameter values
--------------


great_expectations/validator/validator.py
```diff
@@ -414,7 +414,9 @@ def __getattr__(self, name):
         if self.active_batch is None:
             raise TypeError("active_batch cannot be None")  # noqa: TRY003
         name = name.lower()
-        if name.startswith("expect_") and get_expectation_impl(name):
+        if (
+            name.startswith("expect_") or name == "unexpected_rows_expectation"
+        ) and get_expectation_impl(name):
```
NathanFarmer, line 419: This is only needed for the test runner as calling `Validator.unexpected_rows_expectation` is no longer part of the public API.

ashmortar (approved):

great_expectations/expectations/metrics/query_metric_provider.py
```diff
@@ -42,3 +60,45 @@ def _get_query_from_metric_value_kwargs(cls, metric_value_kwargs: dict) -> str:
...
+            # Add a new WHERE condition
+            query = unfiltered_query.replace("WHERE", f"WHERE {batch_filter} AND")
+        elif "GROUP BY" in query.upper():
+            # If there is no existing WHERE clause, but there is a GROUP BY clause
+            # add the WHERE clause before the GROUP BY clause
+            query = unfiltered_query.replace("GROUP BY", f"WHERE {batch_filter} GROUP BY")
+        elif "ORDER BY" in query.upper():
+            # If there is no existing WHERE clause, but there is an ORDER BY clause
+            # add the WHERE clause before the ORDER BY clause
+            query = unfiltered_query.replace("ORDER BY", f"WHERE {batch_filter} ORDER BY")
```
dctalbot:
```suggestion
            query = unfiltered_query.replace("WHERE", f"WHERE {batch_filter} AND")
        elif "GROUP BY" in query.upper():
            query = unfiltered_query.replace("GROUP BY", f"WHERE {batch_filter} GROUP BY")
        elif "ORDER BY" in query.upper():
            query = unfiltered_query.replace("ORDER BY", f"WHERE {batch_filter} ORDER BY")
```

generally best to reserve comments for "why" questions, not "what". If there is uncertainty in the code, we should add tests. In this case, the code is clear enough IMO.

PR #10393, opened 2024-09-13
[BUGFIX] Fix action equality

tyler-hoffman:
This bug prevented running checkpoints with slack or email actions because their renderers didn't evaluate to being equal. Note: I'm only addressing the actions we support per https://greatexpectations.atlassian.net/wiki/spaces/DVRL/pages/744849422/Current+state+of+support+and+testing+for+GX+integrations

### DIFF
--- a/great_expectations/render/renderer/email_renderer.py
+++ b/great_expectations/render/renderer/email_renderer.py
@@ -80,3 +80,7 @@ class EmailRenderer(Renderer):
         else:
             logger.warning("No docs link found. Skipping data docs link in the email message.")
         return report_element
+
+    @override
+    def __eq__(self, other: object) -> bool:
+        return isinstance(other, EmailRenderer)
--- a/tests/actions/test_core_actions.py
+++ b/tests/actions/test_core_actions.py
@@ -810,6 +810,14 @@ class TestV1ActionRun:
         assert "Successfully posted results" in result["result"]
 
     @pytest.mark.unit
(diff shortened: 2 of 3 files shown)


great_expectations/render/renderer/email_renderer.py
```diff
@@ -80,3 +80,7 @@ def _get_report_element(self, docs_link):
         else:
             logger.warning("No docs link found. Skipping data docs link in the email message.")
         return report_element
+
+    @override
+    def __eq__(self, other: object) -> bool:
+        return isinstance(other, EmailRenderer)
```
cdkini, line 86: Should we override the base renderer and just assert that the types are the same?

tests/actions/test_core_actions.py
```diff
@@ -380,6 +380,23 @@ def test_APINotificationAction_run(self, checkpoint_result: CheckpointResult):
             ],
         )
 
+    def test_EmailAction_equality(self):
+        """I know, this one seems silly. But this was a bug."""
```
cdkini, line 385: 😭

cdkini (approved):

great_expectations/render/renderer/email_renderer.py
tyler-hoffman: I figured it might be simplest to just implement this here in case we ended up wanting any other branching down the line. Seems unlikely that we would though. IDK, would you prefer that approach? If so I'll update

PR #10402, opened 2024-09-16
[FEATURE] Add expectation windows for dynamic parameters

rreinoldsc:
# What

Adds a new optional field to Expectation that allows configuration of temporal window definitions with dynamic parameters

## Serialization:
for dict, empty `windows` key is omitted, intentional
for json, empty `windows` key is present with `null` value, default

# Why

To support dynamic parameters

# Reference
[Ticket](https://greatexpectations.atlassian.net/browse/ZELDA-970)


great_expectations/expectations/expectation.py
```diff
@@ -75,6 +75,7 @@
 from great_expectations.expectations.sql_tokens_and_types import (
     valid_sql_tokens_and_types,
 )
+from great_expectations.expectations.window import Window  # noqa: TCH001
```
billdirks: Instead of ignoring TCH001, you can move this to the type checking block.

great_expectations/expectations/window.py
```diff
@@ -0,0 +1,13 @@
+from great_expectations.compatibility import pydantic
+
+
+class PercentOffset(pydantic.BaseModel):
```
billdirks: Maybe name this ProportionalOffset instead of percentage if you expect a ratio.

tests/expectations/test_expectation.py
```diff
@@ -225,6 +226,40 @@ def test_expectation_configuration_property():
     )
 
 
+@pytest.mark.unit
+def test_expectation_configuration_window():
+    expectation = gxe.ExpectColumnMaxToBeBetween(
+        column="foo",
+        min_value=0,
+        max_value=10,
+        windows=[
```
billdirks, line 235: With respect to your optional comment above, you can have tests where you don't set `windows` and one where you set it to `None` and see serialization works correctly.

billdirks (commented):
Thanks for doing this. Before this goes in, you should verify that mercury and mercury-v1 don't die when it sees they new field.

great_expectations/expectations/window.py
rreinoldsc: I'm leaning towards 'Offset' and its corresponding key to 'offset', so that we don't narrow ourselves in so early in this feature lifecycle. We can later define different 'Offset' subclasses

great_expectations/expectations/window.py
```diff
@@ -0,0 +1,13 @@
...
+    positive: float
+    negative: float
+
+
+class Window(pydantic.BaseModel):
+    constraint_fn: str
+    parameter_name: str
+    range: int
+    percent_offset: PercentOffset
```
rreinoldsc: Suggest naming to 'offset'

tests/expectations/test_expectation.py
rreinoldsc: o nice, yeah

great_expectations/expectations/expectation.py
rreinoldsc: This was causing a pydantic inheritance issue, I'll repro and put error here

great_expectations/expectations/window.py
rreinoldsc: cc @shiplet lets do a design review

great_expectations/expectations/window.py
rreinoldsc: cc @shiplet for design review

great_expectations/expectations/expectation.py
billdirks: Oh, yes. Pydantic and static type checking are sometimes at odds since pydantic is very dynamic by nature. It could by mypy thinks this is only used in type checking but pydantic uses that info at runtime for validations.

great_expectations/expectations/window.py
rreinoldsc: outcome: to change key to `offset`

great_expectations/expectations/window.py
rreinoldsc: outcome: to change key to `offset`

great_expectations/expectations/window.py
rreinoldsc: renamed key in 9e413e1e33eda30a28116d978f0869e11338f1c9
renamed class in ba4dc3006

rreinoldsc: I know the impl was not exactly your rec, but I think its an improvement - thoughts @billdirks ?

tests/expectations/test_expectation.py
rreinoldsc: Strangely, `None` succeeds an input output, but `[]` on input is being ignored on output. Looking into it

rreinoldsc:
~Seeing a timeout failure of 2.0s on a test that I marked as `slow`, which should have a 10s timeout.~ resolved

tldr; I thought `slowest=10` sets `slow` marker tests to limit of 10s, but it means show the slowest 10 tests. Needed `filesystem tag`

great_expectations/expectations/window.py
billdirks: Yes, this is great, thanks!

tests/expectations/core/test_core_model_schemas.py
```diff
@@ -22,6 +22,7 @@ def test_all_core_model_schemas_are_serializable():
         model.schema_json()
 
 
+@pytest.mark.slow  # 4.2s
```
billdirks: `unit` will force this to have an upper bound of 2s. If this test takes longer you should use a different marker from [this list](https://github.com/great-expectations/great_expectations/blob/develop/tests/conftest.py#L112). Every test must have exactly 1 of these markers. If there is a reason it's slow, like it touches the filesystem (I think this test does?) use `pytest.mark.filesystem`. If there is no reason lining up with a particular marker you can use `big`.

billdirks: I don't think if the `slow` marker actually does anything anymore?

rreinoldsc: Ah filesystem worked, ty

shiplet (approved):
Would prob want an additional approval beyond mine, but to my knowledge this looks good! 🚀

great_expectations/expectations/expectation.py
rreinoldsc: I'm importing the `Window` pydantic class, but only use it for typing. This would normally just go under TYPE_CHECKING, which is what `ruff` is asking for.

However, with `Window` import in type checking block, ruff is happy but test itself fails with:
`pydantic.v1.errors.ConfigError: field "windows" not yet prepared so type is still a ForwardRef, you might need to call ExpectColumnMaxToBeBetween.update_forward_refs()`

rreinoldsc: @Kilo59 we talked about a fix for base classes today at BE CoP, but this seems like a different issue, because its not a base class, its import of a model for typing. Any ideas?

tests/expectations/test_expectation.py
rreinoldsc: update: It was the `dict(exclude_defaults=True)`, updated default and getting expected results

great_expectations/expectations/expectation.py
Kilo59: It's still essentially the same issue. Did you try adding this model (or its parent) to the ruff list of runtime evaluated base classes?

rreinoldsc: Adding it to runtime evaluated base classes didn't fix it, I'll take another look at it tomorrow

Kilo59: Can you commit and push your attempted fix (tomorrow)?

It could be that we need the gx pydantic compatibility class in here, or there could be some issue with Metaclasses, but Fluent datasources use metaclasses too and so I doubt that's it. 
https://github.com/great-expectations/great_expectations/blob/dec5dcede7a562659af2d62977e36d9596ea2848/pyproject.toml#L394-L401

rreinoldsc: Forced commit 7eb605a41. I think its not working as expect because my Window class is not a base class, its just an import.

Here is a [demo](https://gxlabs.zoom.us/clips/share/zYsDKPfxkhYg5jKnq65t6V04cY0WPYX7feuMiN0Q31BdZuEJwQCg2jrGXqbOX0jJ0lblW_RWgsj1lwCWoOXn1zew.od5ECZ0pAnpaWrQ4) of the errors

pyproject.toml
```diff
@@ -398,6 +398,7 @@ runtime-evaluated-base-classes = [
     "great_expectations.datasource.fluent.fluent_base_model.FluentBaseModel",
     "great_expectations.datasource.fluent.interfaces.Datasource",
     "great_expectations.datasource.fluent.sql_datasource.SQLDatasource",
+    "great_expectations.expectations.window.Window"
 ]
```
Kilo59:
```suggestion
    "great_expectations.compatibility.pydantic.BaseModel",
    "pydantic.v1.BaseModel",
    "great_expectations.expectations.Expectation",
]
```
Try this
It's possible only the first 2 are needed but I would commit all 3 first.

great_expectations/expectations/expectation.py
Kilo59: The class that needs to be recognized as needing runtime annotations is `Expectation` not `Window`.

Made a suggestion here.
https://github.com/great-expectations/great_expectations/pull/10402#discussion_r1767102792

TrangPham (approved):
Looks good to me once you have tests and checks passing

great_expectations/expectations/expectation.py
rreinoldsc: Ah I see I had the wrong class. I also applied your suggestion, which worked, then I trimmed it down to one net additional class:

 ```
great_expectations.compatibility.pydantic.BaseModel
```
47a6bbf39

All good now - thank you for the support

File: pyproject.toml (hunk 6 above)
Comment (rreinoldsc) [reply pyproject.toml]: worked. ty

**Comment by rreinoldsc** [comment]
> you should verify that mercury and mercury-v1 don't die when it sees they new field.

fyi @billdirks this PR will be merge coordinated with the [BE PR](https://github.com/greatexpectationslabs/mercury/pull/3820)

File: great_expectations/checkpoint/checkpoint.py [hunk 7]
```diff
@@ -34,7 +34,7 @@
 )
 from great_expectations.compatibility.typing_extensions import override
 from great_expectations.core.expectation_validation_result import (
-    ExpectationSuiteValidationResult,  # noqa: TCH001
+    ExpectationSuiteValidationResult,
```
rreinoldsc, line 37: a positive side effect of updating the pydantic runtime base classes in pyproject.toml

great_expectations/expectations/expectation.py
```diff
@@ -63,8 +63,9 @@
     COLUMN_DESCRIPTION,
     COLUMN_LIST_DESCRIPTION,
     MOSTLY_DESCRIPTION,
+    WINDOWS_DESCRIPTION,
 )
-from great_expectations.expectations.model_field_types import (  # noqa: TCH001  # types needed for pydantic deser
+from great_expectations.expectations.model_field_types import (  # types needed for pydantic deser
```
rreinoldsc: same as above, a positive side effect of updating the pydantic runtime base classes in pyproject.toml

great_expectations/expectations/expectation.py
```diff
@@ -1281,6 +1284,7 @@ def configuration(self) -> ExpectationConfiguration:
         description = kwargs.pop("description", None)
         id = kwargs.pop("id", None)
         rendered_content = kwargs.pop("rendered_content", None)
+        # We might need the windows key here
```
rreinoldsc:
```suggestion
```

great_expectations/expectations/model_field_descriptions.py
```diff
@@ -7,3 +7,4 @@
     "If specified, sets the condition on which a given row is to be ignored."
 )
 VALUE_SET_DESCRIPTION = "A set of objects used for comparison."
+WINDOWS_DESCRIPTION = "Definition(s) for evaluation of time-series windows"
```
rreinoldsc:
```suggestion
WINDOWS_DESCRIPTION = "Definition(s) for evaluation of temporal windows"
```

tests/integration/cloud/rest_contracts/conftest.py
```diff
@@ -9,7 +9,7 @@
 
 import pact
 import pytest
-from typing_extensions import Annotated, TypeAlias  # noqa: TCH002
+from typing_extensions import Annotated, TypeAlias
```
rreinoldsc, line 12: same as above, a positive side effect of updating the pydantic runtime base classes in pyproject.toml

tests/expectations/test_expectation.py
```diff
@@ -225,6 +226,59 @@ def test_expectation_configuration_property():
     )
 
 
+@pytest.mark.unit
+def test_expectation_configuration_window():
```
Shinnnyshinshin, line 230: 🚀

Shinnnyshinshin (approved):
LGTM too :) thanks for this @rreinold

great_expectations/expectations/expectation.py
```suggestion
from great_expectations.expectations.model_field_types import (
```

great_expectations/expectations/window.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Window(pydantic.BaseModel):
+    """
+    A definition for a temporal window across <`range`> number of previous invocations
+    """
+
+    constraint_fn: str
+    parameter_name: str
+    range: int
+    offset: Offset
```
Kilo59: I would suggest we set the `extra` ignore behavior to forbid for both of these models.
Otherwise it can be easy to start providing extra arguments to this class without noticing.

> `extra`
>whether to ignore, allow, or forbid extra attributes during model initialization. Accepts the string values of 'ignore', 'allow', or 'forbid', or values of the Extra enum (default: Extra.ignore). 'forbid' will cause validation to fail if extra attributes are included, 'ignore' will silently ignore any extra attributes, and 'allow' will assign the attributes to the model.


https://docs.pydantic.dev/1.10/usage/model_config/#options

```python
from great_expectations.compatibility import pydantic, Extra

class Offset(pydantic.BaseModel):
    """
    A threshold in which a metric will be considered passable
    """

    positive: float
    negative: float

    class Config:
        extra = Extra.forbid

class Window(pydantic.BaseModel):
    """
    A definition for a temporal window across <`range`> number of previous invocations
    """

    constraint_fn: str
    parameter_name: str
    range: int
    offset: Offset

    class Config:
        extra = Extra.forbid
```

Kilo59 (approved):

great_expectations/expectations/window.py
rreinoldsc: Yep great call, added in a63abe335, tweaked imports in 24edd2f2aa8042e97a69c246f74495fbc7c57bc0

great_expectations/expectations/expectation.py
rreinoldsc: 453d864df

PR #10406, opened 2024-09-16
[BUGFIX] Fix Databricks SQL Regex and Like based Expectations

Kilo59:
Regex based Expectations do not currently work for DatabricksSQL.

Added 2 expectations to the SQL datasource integration tests. Both just matching on wildcards. Even with wildcards, both of these failed previously for Databricks SQL.
1. `expect_column_values_to_match_regex`
2. `expect_column_values_to_match_like_pattern`


tests/datasource/fluent/integration/test_sql_datasources.py
```diff
@@ -764,16 +772,34 @@ def _raw_query_check_column_exists(
         return True
 
 
+MAPPING: Mapping[type[Expectation], object] = {  # TODO: do something with this
+    ExpectColumnValuesToMatchRegex: {},
+    ExpectColumnToExist: {},
+    ExpectColumnValuesToNotBeNull: {},
+    ExpectColumnValuesToMatchLikePattern: {},
+}
```
Kilo59, line 780:
```suggestion
```

great_expectations/expectations/metrics/util.py
```diff
@@ -80,11 +88,14 @@
 
 
 def get_dialect_regex_expression(  # noqa: C901, PLR0911, PLR0912, PLR0915
-    column, regex, dialect, positive=True
-):
+    column: sa.Column,
+    regex: str,
+    dialect: ModuleType | Type[sa.Dialect] | sa.Dialect,
```
Kilo59, line 111: The types the `dialect` object can be causes me pain 😬

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -378,6 +378,8 @@ def __init__(  # noqa: C901, PLR0912, PLR0913, PLR0915
             self.dialect_module = import_library_module(
                 module_name="clickhouse_sqlalchemy.drivers.base"
             )
+        elif self.dialect_name == GXSqlDialect.DATABRICKS:
+            self.dialect_module = import_library_module("databricks.sqlalchemy.base")
```
Kilo59: This change didn't seem to be necessary to fix the expectations.

great_expectations/expectations/metrics/util.py
anthonyburdi: Non-blocking: Is it worth making an alias?

anthonyburdi (approved):

great_expectations/expectations/metrics/util.py
Kilo59: I'd be afriad that will lock-in or encourage the behavior of using `dialect` in all these different ways.

I'd like to spend some effort eliminating or reducing the multiple types of "dialects" we provide here.
It's possible that we don't actually need to be this flexible here, but I haven't done a deep dive on that.

PR #10412, opened 2024-09-17
[BUGFIX] Ignore unsupported `INTERVAL` type as part of CDM

Kilo59:
Currently CDM and Schema introspection fails for any Databricks SQL table asset that has an `INTERVAL` column.

This change should prevent the presence of an `INTERVAL` column from breaking CDM for all other columns.

https://docs.databricks.com/en/sql/language-manual/data-types/interval-type.html

### DIFF
--- a/great_expectations/experimental/metric_repository/metric_retriever.py
+++ b/great_expectations/experimental/metric_repository/metric_retriever.py
@@ -137,7 +137,7 @@ class MetricRetriever(abc.ABC):
         Excludes columns that are unsupported or missing type metadata
         """
         columns_to_skip: List[str] = []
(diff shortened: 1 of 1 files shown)


anthonyburdi (approved):
Non blocking but prob worth adding a test to either/both in a followup:

- `tests/experimental/metric_repository/test_metric_list_metric_retriever_integration.py`
- `tests/experimental/metric_repository/test_metric_list_metric_retriever.py`

Kilo59:
@anthonyburdi 
I'll add tests before merging this version, but I was planning to leave them out of the `0.18.x` one.

Kilo59:
If I recall, I think we determined that this doesn't actually fix the issue so I'm closing this.

PR #10416, opened 2024-09-17
[docs] add information about credential management

adeola-ak:
updating deployment patterns and architecture page to have information about how to manage/store credentials

### DIFF
--- a/docs/docusaurus/docs/cloud/about_gx.md
+++ b/docs/docusaurus/docs/cloud/about_gx.md
@@ -80,6 +80,20 @@ For example, you might opt to run and interact with the GX Core Python library l
 
 GX provides flexible, robust products that allow your organization to quickly deploy GX Cloud or GX Core to fit your unique requirements.
 
+## Credential management 
+
+To connect to your database in GX Cloud, there are two methods for managing credentials:
+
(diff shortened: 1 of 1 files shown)


docs/docusaurus/docs/cloud/about_gx.md
```diff
@@ -80,6 +80,20 @@ For example, you might opt to run and interact with the GX Core Python library l
 
 GX provides flexible, robust products that allow your organization to quickly deploy GX Cloud or GX Core to fit your unique requirements.
 
+## Credential management
```
rachhouse, line 83: The About GX page is currently being overhauled (by me) in a separate PR and won't be a suitable place for this information after the reorganization is complete. I'd suggest we add this information to the *Manage Data Assets* page instead, since it's relevant to Data Source creation.

rachhouse: Looking at the content in https://github.com/great-expectations/great_expectations/pull/10417/files, I think it'd be helpful to combine the two together into a _Data Source credential management_ section (within the _Manage Data Assets_ page) that describes both Direct input and Environment variable substitution - that way, all the relevant information is in one place.

PR #10419, opened 2024-09-18
[MAINTENANCE] Move FabricPowerBIDatasource out of experimental dir

Kilo59:
Move the `FabricPowerBIDatasource` out of the `great_expectations.experimental` sub-package.

It's not clear what a having a datasource in this directory actually means and it causes confusion with our contributor [`great_expectations_experimental` package](https://github.com/great-expectations/great_expectations/tree/develop/contrib/experimental).


great_expectations/experimental/datasource/__init__.py
Kilo59, line 1: We could keep this module here, and re-export from the new location.

tyler-hoffman (commented):
I have no issues with this. The one thing that is kind of weird here is that [fabric is officially unsupported by core](https://greatexpectations.atlassian.net/wiki/spaces/DVRL/pages/744849422/Current+state+of+support+and+testing+for+GX+integrations). Does this move signal otherwise? (I don't think it does - @joshzzheng has been pretty clear that documentation == supported

tyler-hoffman (approved):
For other folks: @Kilo59 looped me in on discussions to promote this to a community-supported datasource.

joshzzheng (approved):

PR #10420, opened 2024-09-18
[BUGFIX] Support Spark connect dataframes

tyler-hoffman:
We previously did isinstance checks on spark dataframes and only allowed non-connect dataframes. This fixes that.

NOTE: This does *not* work if the spark session was created via the `SparkConnectSession`'s factory methods. This is documented in an xfailed test.


tests/integration/spark/test_spark_connect.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@pytest.mark.xfail(
+    reason="If we use the factory method on SparkConnectSession to create the session, we fail",
+    strict=False,
+)
+def test_spark_connect_with_bad_factory_method(
+    spark_validation_definition: ValidationDefinition,
+):
+    """The purpose of this test is to document an issue with using SparkConnectSession's
```
billdirks: If we need to run this test in isolation since it uses spark connect and the other tests don't we could do the following:
1. Add a `spark_connect` marker to this test/conftest.py  [REQUIRED_MARKERS](https://github.com/great-expectations/great_expectations/blob/develop/tests/conftest.py#L112) list.
2. I think you'll need to a key/value to tasks.py [MARKER_DEPENDENCY_MAP](https://github.com/great-expectations/great_expectations/blob/develop/tasks.py#L931) with the key `spark_connect` whose value is the same as `spark`.
3. Update ci.yml to add `spark_connect` to the [marker list](https://github.com/great-expectations/great_expectations/blob/develop/.github/workflows/ci.yml#L425C9-L425C16).

We'll then need to make it a required marker in the github UI.

tyler-hoffman: Do you it's worth it? Happy to do those updates if you want, and I had thought about it earlier, but I decided against it because:

* If anything, the spark_connect marker would probably apply to all tests in this file, so we'd probably want another to just be spark_connect_without_a_session or something. We could do it, but it seems potentially confusing to me to have a marker that would really just apply to the one file
* I mostly just added the test because it was more clear (IMO) that adding a comment in the description.
So I personally think it's fine to keep it as is, but if you lean toward making sure it runs and fails, I can add another marker.
* In this case, I don't see much value to being alerted in the unlikely case that this started passing

But if you lean toward it, I'm happy to do it!

tyler-hoffman: Well, ended up needing a separate flag for `spark_connect` to just support running these tests (otherwise CI errored saying we already had a regular spark session, so couldn't get a spark connect one). So added that. There were a couple other changes that needed to be made as well. But @billdirks I think I'll have to bug you about making it required in the GH UI.

billdirks: Does this test always fail now? Wondering if we can update the `strict` value and the docstring?

billdirks (approved):
Thanks for getting to the bottom of this!

tests/integration/spark/test_spark_connect.py
tyler-hoffman: Turns out it now passes with this mark. I was very confident it was not when I was running both tests with the other mark. This was the fundamental problem I was hitting when testing this lasts week too. That doesn't make sense, I'm not yet sure on the specifics of what is different, but I'm no longer xfailing the test. I'll mess around with this a bit tomorrow.

PR #10424, opened 2024-09-19
[BUGFIX] Handle DatabricksSQL attribute error and update dependency

Kilo59:
Depending on the version and extra depenencies the `databricks` package may not always have a `.sqlalchemy` sub-module.

This PR handles the potential `AttributeError` and bumps the minimum version of `databricks-sql-connector`

### DIFF
--- a/great_expectations/expectations/metrics/util.py
+++ b/great_expectations/expectations/metrics/util.py
@@ -68,7 +68,7 @@ except ImportError:
 
 try:
     import databricks.sqlalchemy as sqla_databricks
-except ImportError:
+except (ImportError, AttributeError):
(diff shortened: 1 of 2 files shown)


NathanFarmer (approved):

great_expectations/expectations/metrics/util.py
```diff
@@ -68,7 +68,7 @@
 
 try:
     import databricks.sqlalchemy as sqla_databricks
-except ImportError:
+except (ImportError, AttributeError):
     sqla_databricks = None  # type: ignore[assignment]
```
Kilo59, line 73: I think this module should probably use our `NotImported` pattern. But I don't want to do that as part of this PR.

PR #10441, opened 2024-09-30
[MAINTENANCE] Remove Python 3.8 Support (EOL)

Kilo59:
As of October 2024 Python 3.8 has reached End of Life.

https://devguide.python.org/versions/
https://endoflife.date/python

This PR removes Python 3.8 as a supported version and removes all CI related checks for it.
Going forward Python 3.9 is the minimum supported version of Python.

NOTE: Netlify still only supports 3.8 (and 2.7 lol), so its config is still set to 3.8. Sorry.


great_expectations/compatibility/typing_extensions.py
```diff
@@ -7,7 +7,7 @@
         Annotated,
     )
 except ImportError:
-    from typing_extensions import Annotated
```
Kilo59, line 10: Removing this

pyproject.toml
```diff
@@ -317,6 +317,7 @@ lint.ignore = [
     # https://beta.ruff.rs/docs/rules/#flake8-pyi-pyi
     "PYI053", # string-or-bytes-too-long - causes mypy to fail on some of our type stubs
     "PYI054", # numeric-literal-too-long - causes mypy to fail on some of our type stubs
+    "UP035", # TODO: remove once min version of pydantic supports using collections.abc types
```
Kilo59, line 320: This caused issues with our min version of pydantic when using things like `collections.abc.Mapping` instead of `typing.Mapping`.

I'll do a followup to in-line ignore them instead, I didn't want to do it in this PR because that will cause changes to a heck-ton of files (400+).

setup.py
Kilo59, line 1: The changes in `setup.py` are what actually prevent installs of (new) gx versions with python 3.8.

great_expectations/expectations/expectation.py
```diff
@@ -1826,7 +1826,7 @@ class ColumnMapExpectation(BatchExpectation, ABC):
     """  # noqa: E501
 
     column: StrictStr = Field(min_length=1, description=COLUMN_DESCRIPTION)
-    mostly: Mostly = 1.0
+    mostly: Mostly = 1.0  # type: ignore[assignment] # FIXME
```
Kilo59: ~I haven't had time to dig into what's going here with `Mostly`.~
Changing the mypy target to Python 3.9 surfaces this issue where the `Mostly = Annotated[_Mostly, float]` no longer sees these `float` types as valid. I suspect it never did but it's only being surfaced now.

My initial assumption is that it's because `Annotated` wasn't being used quite right but I don't know for sure. I tried updating our compatibility layer to first use the `typing_extensions` version before falling back to the stdlib version since `typing_extensions` can include bug fixes or enhancements. But that didn't seem to do anything.

## Edit

Changing the order of the types fixed it.
```python
Mostly = Annotated[_Mostly, float] # this causes typing issues
Mostly = Annotated[float, _Mostly] # this works as expected
```


## References
https://peps.python.org/pep-0593/
https://docs.python.org/3/library/typing.html#typing.Annotated

tests/checkpoint/test_checkpoint_id_pk.py
```diff
@@ -81,13 +81,17 @@ def expected_sql_query_output() -> str:
 
 @pytest.fixture
 def expect_column_values_to_be_in_set() -> gxe.ExpectColumnValuesToBeInSet:
-    return gxe.ExpectColumnValuesToBeInSet(column="animals", value_set=["cat", "fish", "dog"])
+    return gxe.ExpectColumnValuesToBeInSet(
+        column="animals",
+        value_set=["cat", "fish", "dog"],  # type: ignore[arg-type] # FIXME
+    )
```
Kilo59: Seeing a similiar issue with the `ValueSet` as we have with `Mostly`

```python
ValueSet = Annotated[_ValueSet, Union[list, set]]
```

See 👇 for more details
https://github.com/great-expectations/great_expectations/pull/10441/files#r1782964972

Kilo59: Order of the types needed to be reversed.

great_expectations/expectations/model_field_types.py
```diff
@@ -48,7 +48,7 @@ def __modify_schema__(cls, field_schema: Dict[str, Any], field: Union[fields.Mod
             field_schema["multipleOf"] = 0.01
 
 
-Mostly = Annotated[_Mostly, float]
+Mostly = Annotated[float, _Mostly]
```
Kilo59: see https://github.com/great-expectations/great_expectations/pull/10441#discussion_r1782964972

tests/validator/test_validation_graph.py
```diff
@@ -439,23 +439,27 @@ class DummyExecutionEngine:
     execution_engine = cast(ExecutionEngine, DummyExecutionEngine)
 
     # ValidationGraph is a complex object that requires len > 3 to not trigger tqdm
-    with mock.patch(
-        "great_expectations.validator.validation_graph.ValidationGraph._parse",
-        return_value=(
-            {},
-            {},
+    with (
```
tyler-hoffman, line 442: is this formatting change intended to be part of this PR?

tyler-hoffman (approved):

tests/validator/test_validation_graph.py
Kilo59: Yes it's a syntax/formatting change made possible by bumping our min version of python to 3.9 and updating the `ruff` "python version target".

great_expectations/datasource/fluent/sources.py
```diff
@@ -128,7 +128,7 @@ def register_datasource(cls, ds_type: Type[Datasource]) -> None:
             )
 
         # rollback type registrations if exception occurs
-        with cls.type_lookup.transaction() as ds_type_lookup, ds_type._type_lookup.transaction() as asset_type_lookup:  # noqa: E501
+        with cls.type_lookup.transaction() as ds_type_lookup, ds_type._type_lookup.transaction() as asset_type_lookup:  # fmt: skip # noqa: E501
```
tyler-hoffman, line 131: Had to add `# fmt: skip` to get the formatter to chill here 🤔

Kilo59: Disabling `SIM117` didn't do the trick?

https://docs.astral.sh/ruff/rules/multiple-with-statements/#multiple-with-statements-sim117

tyler-hoffman: That's already ignored, but that rule is about multiple nested context managers, not about the formatting with parens. Unless I'm misreading something.

tests/analytics/test_analytics.py
```diff
@@ -78,9 +78,12 @@ def test_event_identifiers(analytics_config):
 def test_ephemeral_context_init(monkeypatch):
     monkeypatch.setattr(ENV_CONFIG, "gx_analytics_enabled", True)  # Enable usage stats
 
-    with mock.patch(
-        "great_expectations.data_context.data_context.abstract_data_context.init_analytics"
-    ) as mock_init, mock.patch("posthog.capture") as mock_submit:
+    with (
```
billdirks, line 81: Is this just formatting changes?

billdirks (approved):
LGTM, Thank you @tyler-hoffman and @Kilo59!

PR #10443, opened 2024-09-30
[BUGFIX] Enable `SlackNotificationAction` variable substitution

cdkini:
We should allow for `${VAR}` syntax with this action


great_expectations/checkpoint/actions.py
```diff
@@ -213,9 +215,9 @@ class SlackNotificationAction(DataDocsAction):
 
     type: Literal["slack"] = "slack"
 
-    slack_webhook: Optional[str] = None
-    slack_token: Optional[str] = None
-    slack_channel: Optional[str] = None
+    slack_webhook: Optional[Union[ConfigStr, str]] = None
+    slack_token: Optional[Union[ConfigStr, str]] = None
+    slack_channel: Optional[Union[ConfigStr, str]] = None
```
cdkini, line 220: Should a channel be a secret str? Felt less likely than the rest but I think consistency makes sense here

great_expectations/checkpoint/actions.py
```diff
@@ -234,20 +236,32 @@ def _validate_renderer(cls, renderer: dict | SlackRenderer) -> SlackRenderer:
 
     @root_validator
     def _root_validate_slack_params(cls, values: dict) -> dict:
-        slack_token = values["slack_token"]
-        slack_channel = values["slack_channel"]
-        slack_webhook = values["slack_webhook"]
+        from great_expectations.data_context.data_context.context_factory import project_manager
+
+        config_provider = project_manager.get_config_provider()
```
Kilo59, line 241: `non-blocking`

Just thinking out loud... but now that we can pull the config provider from the `project_manager` we don't actually need to pass it to `ConfiStr.get_config_value(config_provider)`. 

We could always have this method `ConfiStr.get_config_value()` pull from the project manager.

https://github.com/great-expectations/great_expectations/blob/ef46703c4ee92f5c3b131546292646a2d4ec4810/great_expectations/datasource/fluent/config_str.py#L47-L53

I would do this separately from this PR.

great_expectations/checkpoint/actions.py
Kilo59: I would go ahead and keep it substituable, although not because it needs to be secret.
It adds flexibility without further complicating anything.

Kilo59 (approved):

leodrivera:
Hello @cdkini 

I've tested your implementation, and it seems to be working fine. However, the checkpoint json files are being saved with the values substituted, not the placeholders. As this file is intended to be committed alongside the other gx config files, I believe this is not an intended behavior.

I think that the expected behavior should just be like the `connection_string` from the *add_or_update_postgres* method, where the placeholders are saved in the `great_expectations.yml` file.

great_expectations/checkpoint/actions.py
```diff
@@ -234,20 +236,32 @@ def _validate_renderer(cls, renderer: dict | SlackRenderer) -> SlackRenderer:
...
+        for credential in ("slack_webhook", "slack_token", "slack_channel"):
+            if isinstance(values[credential], ConfigStr):
+                values[credential] = cls._substitute_slack_credential(
+                    slack_credential=values[credential],
+                    config_provider=config_provider,
+                )
```
Kilo59, line 247: @cdkini sorry I missed this in the review.
The values shouldn't be updated in the validator method. They should only be substituted when they are actually needed which looks like that would be inside the `.run()` method.

- https://github.com/great-expectations/great_expectations/issues/10473

PR #10492, opened 2024-10-09
[MAINTENANCE] Prevent deserializing all data sources when adding a new one

cdkini:
To check for existence, we look into `context.data_sources.all()` - we only really need to check the key/name here. Let's use `self._datasources` `DatasourceDict` to do this.


great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -535,7 +535,7 @@ def _add_fluent_datasource(
             )
 
         # We currently don't allow one to overwrite a datasource with this internal method
-        if datasource_name in self.data_sources.all():
+        if datasource_name in self._datasources:
```
cdkini, line 538: Nit - we should better differentiate `self._data_sources` and `self._datasources` - we should combine the two or perhaps make the latter `self._data_source_dict`

TrangPham:
Thanks for your quick action here. There are other places which would have a similar issue. For example:
`AbstractDataContext.get_datasource` also calls `.data_sources.all()[name]`

As much as possible when an action pertains to one entity, the code shouldn't be verifying all other entities.

great_expectations/data_context/data_context/abstract_data_context.py
TrangPham: May I suggest `_raw_data_sources`

rreinoldsc (approved):
Also seeing this bug, thanks for fixing

PR #10517, opened 2024-10-15
[CONTRIB] Exclude null values no days missing metric

itaise:
Hi, 
Here is a PR to exclude null values from a metric - ColumnDistinctDates, which is used in a test we contributed and are actively using. 
If the column checked contains null values, they are failing the test with an exception (as None does not have strftime method). 
Can you please check?

Thanks a lot!


itaise:
Can someone please review? thanks!

NathanFarmer (changes requested):
Hi @itaise! Sorry that this review seems to have fallen through the cracks. Based on your description this seems to be a bug. Can you please add a test to `tests/integration/datasources_and_expectations/expectations` that demonstrates the issue with the Expectation you described? You can follow the same pattern shown for other Expectations in that directory.

wookasz:
Closing for lack of activity. Please don't hesitate to reopen if there are additional changes or if you feel this was done in error.

PR #10524, opened 2024-10-15
[MAINTENANCE] Update gx row_condition parser

josectobar:
- Remove "experimental" status from `great_expectation` parser.    
- Update related docs
- Add `==` condition to replace logic in `renderer_configuration`


great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -666,7 +666,7 @@ def get_domain_records(  # noqa: C901, PLR0912, PLR0915
             condition_parser = domain_kwargs.get("condition_parser", None)
             if condition_parser == "spark":
                 data = data.filter(row_condition)
-            elif condition_parser == "great_expectations__experimental__":
+            elif condition_parser == "great_expectations":
```
NathanFarmer: Can you use the [ConditionParser](https://github.com/great-expectations/great_expectations/blob/26b118a44a54dfc1b11917b8b39d4db2db560ef8/great_expectations/expectations/model_field_types.py#L117) enum I just created in these 2 branches (`spark` and `great_expectations`)?

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -653,7 +653,7 @@ def get_domain_records(  # noqa: C901, PLR0912, PLR0915
         # Filtering by row condition.
         if "row_condition" in domain_kwargs and domain_kwargs["row_condition"] is not None:
             condition_parser = domain_kwargs["condition_parser"]
-            if condition_parser == "great_expectations__experimental__":
+            if condition_parser == "great_expectations":
```
NathanFarmer: Same comment. Please use Can you use the [ConditionParser](https://github.com/great-expectations/great_expectations/blob/26b118a44a54dfc1b11917b8b39d4db2db560ef8/great_expectations/expectations/model_field_types.py#L117) enum I just created in these 2 branches (`spark` and `great_expectations`)? enum here also.

Also, can you look for this condition in `pandas_execution_engine.py` and use it there as well?

NathanFarmer (changes requested):
Looking pretty good! In addition to the other requested changes, we need to update `ConditionParser` to use the new string (`great_expectations`). Also, I think we should have a new member of `ConditionParser` called `ConditionParser.GX_DEPRECATED` that references `great_expectations__experimental__` and allow users to pass that string as well. I am concerned that this breaking change will actually affect some users otherwise.

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -525,15 +526,12 @@ def get_domain_records(  # noqa: C901, PLR0912
         if row_condition:
             condition_parser = domain_kwargs.get("condition_parser", None)
 
-            # Ensuring proper condition parser has been provided
-            if condition_parser not in ["python", "pandas"]:
```
josectobar, line 529: I removed the `python` condition since it seemed related to `filter_conditions`.

josectobar:
> Looking pretty good! In addition to the other requested changes, we need to update `ConditionParser` to use the new string (`great_expectations`). Also, I think we should have a new member of `ConditionParser` called `ConditionParser.GX_DEPRECATED` that references `great_expectations__experimental__` and allow users to pass that string as well. I am concerned that this breaking change will actually affect some users otherwise.

Is the idea that users will use this `GX_DEPRECATED` flag? I made the change. I'm just not sure how this would work..

great_expectations/execution_engine/sparkdf_execution_engine.py
```diff
@@ -664,9 +665,9 @@ def get_domain_records(  # noqa: C901, PLR0912, PLR0915
         row_condition = domain_kwargs.get("row_condition", None)
         if row_condition:
             condition_parser = domain_kwargs.get("condition_parser", None)
-            if condition_parser == "spark":
+            if condition_parser == ConditionParser.SPARK:
                 data = data.filter(row_condition)
-            elif condition_parser == "great_expectations__experimental__":
+            elif condition_parser == ConditionParser.GX:
```
NathanFarmer: My idea was to also allow `ConditionParser.GX_DEPRECATED` here, so users can continue to pass the old string without breaking.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -653,7 +654,7 @@ def get_domain_records(  # noqa: C901, PLR0912, PLR0915
         # Filtering by row condition.
         if "row_condition" in domain_kwargs and domain_kwargs["row_condition"] is not None:
             condition_parser = domain_kwargs["condition_parser"]
-            if condition_parser == "great_expectations__experimental__":
+            if condition_parser == ConditionParser.GX:
```
NathanFarmer: `if condition_parser in (ConditionParser.GX, ConditionParser.GX_DEPRECATED):`

NathanFarmer (changes requested):

great_expectations/execution_engine/sparkdf_execution_engine.py
josectobar: Got it, that makes sense 😬

josectobar: Do we have a deprecation strategy? How long do we keep these types of paths around?

NathanFarmer: We follow semver, so it wouldn't be until GX 2.0 in theory.

NathanFarmer (approved):

NathanFarmer (approved):

PR #10529, opened 2024-10-16
[MAINTENANCE] Add test to public api to ensure that decorated methods have decorated parent classes

cdkini:


great_expectations/_docs_decorators.py
```diff
@@ -89,10 +90,28 @@ def my_method(some_argument):
...
-    return func
+    public_api_introspector.add(func_or_cls)
+    existing_docstring = func_or_cls.__doc__ if func_or_cls.__doc__ else ""
+    func_or_cls.__doc__ = WHITELISTED_TAG + existing_docstring
+
+    _populate_public_api_registry(func_or_cls)
+    return func_or_cls
+
+
+_public_api_registry = defaultdict(set)
```
cdkini: Don't love globals but on import, this gets populated (see `public_api_registry.json` for output)

great_expectations/_docs_decorators.py
```diff
@@ -89,10 +90,28 @@ def my_method(some_argument):
...
+
+def _populate_public_api_registry(func_or_cls: F) -> None:
+    if isinstance(func_or_cls, type):
+        key = f"{func_or_cls.__module__}.{func_or_cls.__qualname__}"
+        _public_api_registry[key].add("__init__")
+    else:
+        parts = func_or_cls.__qualname__.split(".")
+        if len(parts) >= 2:  # noqa: PLR2004
+            cls = parts[0]
+            func = parts[1]
```
cdkini: Actual impl could be cleaned up but this works

public_api_registry.json
```diff
@@ -0,0 +1,380 @@
+{
+  "great_expectations.render.components.RenderedContent": [
+    "to_json_dict"
+  ],
```
cdkini: Anything without `__init__` will not be rendered in docs because the parent is not rendered.

We should do one of two things:
1. Add `@public_api` to the parent
2. Remove `@public_api` from the method

cdkini: A test around this that checks these conditions would be easy to write given the registry code above

public_api_registry.json
```diff
@@ -0,0 +1,380 @@
...
+    "add",
+    "all",
+    "delete",
+    "get"
+  ],
+  "great_expectations.data_context.data_context_variables.DataContextVariables": [
+    "__init__",
+    "save"
+  ],
+  "great_expectations.data_context.data_context.abstract_data_context.AbstractDataContext": [
```
cdkini: This is a big one - we don't have docs around this but we've decorated a bunch of methods

public_api_registry.json
```diff
@@ -0,0 +1,380 @@
...
+    "list_data_docs_sites",
+    "suites",
+    "update_data_docs_site",
+    "update_project_config",
+    "validation_definitions",
+    "view_validation_result"
+  ],
+  "great_expectations.data_context.data_context.cloud_data_context.CloudDataContext": [
+    "__init__"
+  ],
```
cdkini: In an ideal world, should child classes that don't override their parents have duplicate docs around their methods?

public_api_registry.json
donaldheppner: do we have to check this in?

cdkini: nope just showcasing what the current public API looks like

script.py
```diff
@@ -0,0 +1,60 @@
...
+
+    with open("public_api_registry.json", "w") as f:
+        json.dump(registry, f, indent=2)
+
+
+if __name__ == "__main__":
+    classes = find_classes_that_need_public_api_decorator()
+    pprint(classes)
+    """
+    great_expectations.checkpoint.checkpoint.CheckpointResult
```
donaldheppner: why this list of classes?

great_expectations/_docs_decorators.py
cdkini: Could actually leverage existing `_PublicApiIntrospector`

script.py
cdkini: These are classes that have `public_api` methods without actually being decorated themselves (resulting in them not showing up in docs)

great_expectations/datasource/fluent/pandas_datasource.py
```diff
@@ -83,6 +83,7 @@ class PandasDatasourceError(Exception):
     pass
 
 
+@public_api
 class _PandasDataAsset(DataAsset):
```
cdkini, line 87: Do we promote this one to `PandasDataAsset`?

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -484,6 +484,7 @@ def batch_parameters_to_batch_spec_kwarg_identifiers(
 ]
 
 
+@public_api
 class _SQLAsset(DataAsset[DatasourceT, ColumnPartitioner], Generic[DatasourceT]):
```
cdkini, line 488: Do we promote this one to `SQLAsset`?

great_expectations/render/components.py
```diff
@@ -125,6 +125,7 @@ class LegacyDescriptiveRendererType(str, Enum):
     VALUE_COUNTS_BAR_CHART = ".".join([LegacyRendererType.DESCRIPTIVE, "value_counts_bar_chart"])
 
 
+@public_api
 class RenderedContent:
```
cdkini: Idk how much we care about renderer stuff being public but we already had some classes/methods in our docs so I added decorators to any violations

great_expectations/_docs_decorators.py
```diff
@@ -60,6 +69,23 @@ def add(self, func: F) -> None:
             logger.exception(f"Could not add this function to the public API list: {func}")
             raise
 
+    def _add_to_class_registry(self, func: F) -> None:
+        if isinstance(func, type):
+            key = f"{func.__module__}.{func.__qualname__}"
+            self._class_registry[key].add(self.CLASS_DEFINITION)
```
billdirks: How do we know this is a class and not a top level function? I haven't yet looked at how this is used, so does that matter?

great_expectations/_docs_decorators.py
```diff
@@ -60,6 +69,23 @@ def add(self, func: F) -> None:
...
+        else:
+            parts = func.__qualname__.split(".")
+            if len(parts) >= 2:  # noqa: PLR2004
```
billdirks: Could you add a reason for the noqa?

Also, this assumes that a dotted name will be a method. It could be a nested class or a class variable/constant? It seems like we could look at the type of func?

great_expectations/data_context/data_context/abstract_data_context.py
```diff
@@ -134,6 +134,7 @@
 T = TypeVar("T", dict, list, str)
 
 
+@public_api
```
billdirks, line 135: It makes me a bit sad that we have to make an abstract class part of our public api to get our docs to render, but so be it.

great_expectations/data_context/types/base.py
```diff
@@ -153,7 +153,6 @@ def to_yaml_str(self) -> str:
         """
         return object_to_yaml_str(obj=self.commented_map)
 
-    @public_api
```
billdirks, line 156: I am ok with removing these for now. It is unclear to me if that is the right decision. It seems like people may want to serialize these out and this provides a means to do that.

great_expectations/datasource/fluent/pandas_datasource.py
billdirks: I don't think this should ever be directly instantiated and only exists to subclass. The reason we need to add the public api decorator is for docs to show up. We should probably name it `AbstractPandasDataAsset` like we do for `AbstractDataContext`. Leaving it like this is ok as well.

great_expectations/datasource/fluent/sql_datasource.py
billdirks: Similar to the pandas comment. Probably prepend Abstract to the name or leave as is. We want to communicate that it shouldn't be instantiated directly.

great_expectations/render/components.py
billdirks: @NathanFarmer What are your thoughts here? Should this be public or is this internal to our codebase?

billdirks (commented):
Thanks for doing this!

I have some comments about the robustness of the new instrumentation you add for tests. Happy to chat about it sync.

I also tagged @NathanFarmer about the rendering question.

great_expectations/render/components.py
NathanFarmer: I don't think this is public any longer now that we have removed support for custom Expectations. Users won't be writing their own renderers with `UnexpectedRowsExpectation`.

great_expectations/datasource/fluent/pandas_datasource.py
cdkini: I'd vote to rename in a follow-up if that's okay!

great_expectations/datasource/fluent/sql_datasource.py
cdkini: I'd vote to rename in a follow-up if that's okay!

great_expectations/_docs_decorators.py
cdkini: `type` denotes a class here

great_expectations/_docs_decorators.py
```diff
@@ -60,6 +71,24 @@ def add(self, func: F) -> None:
...
+    def _add_to_class_registry(self, func: F) -> None:
+        if isinstance(func, type):
+            key = f"{func.__module__}.{func.__qualname__}"
+            self._class_registry[key].add(self.CLASS_DEFINITION)
+        else:
+            parts = func.__qualname__.split(".")
+            METHOD_PARTS_LENGTH = 2  # Standalone functions will have a length of 1
+            if len(parts) >= METHOD_PARTS_LENGTH:
+                cls = parts[0]
+                method = parts[1]
```
billdirks: When len(parts) > 2 do we still want to set the method to parts[1]? Is it ever > 2? If not, I think we should raise when this is > 2 which an error message to update this method to handle that case.

billdirks (commented):
I don't want to block this check. I think a quick conversation can resolve my question and we will be able to merge. Thanks for writing up this check!

great_expectations/_docs_decorators.py
cdkini: An example is: `DataSourceManager._register_add_datasource.<locals>.crud_method_info`

This happens when we have closures

billdirks (approved):
We've discussed in person and Chetan will add some comments for the > 2 case.

Thanks!

PR #10530, opened 2024-10-17
[BUGFIX] fix rendering performance

dctalbot:


tests/expectations/core/test_expect_column_values_to_be_of_type.py
```diff
@@ -96,3 +102,16 @@ def test_expect_column_values_to_be_of_type_string_dialect_sqlite(sa):
...
+def test_expect_column_values_to_be_in_set_render_performance():
+    x = ExpectColumnValuesToBeInSet(
+        column="foo_column_name", value_set=[random.randint(0, 10) for _ in range(1000)]
+    )
+
+    start = time.time()
+    x.render()
+    end = time.time()
+    duration_s = end - start
+    assert duration_s < 1.0, f"Rendering took {duration_s} seconds"
```
dctalbot, line 117: currently takes ~1 minute locally

great_expectations/render/renderer_configuration.py
```diff
@@ -75,9 +74,6 @@ class Config:
         validate_assignment = True
         arbitrary_types_allowed = True
 
-    def __len__(self) -> int:
-        return len(self.__fields__)
```
dctalbot, line 79: `__fields__` is a class member and should not be accessed on class instances 
e.g. 
```py
model_instance = MyModel(name="Alice", age=30)

# This is correct (class-level attribute)
print(MyModel.__fields__)

# This is incorrect (instance-level, __fields__ doesn't exist here)
print(model_instance.__fields__)  # This will be empty or not exist
```

In my case, it was empty at a certain point in the model lifecycle and causing a root validator to have unexpected behavior

Kilo59: Typically there isn't actually any problem with accessing class methods/attributes from an instance method. I suspect this is because of some dynamic programing pydantic is doing.

But based on your example you should be able to rewrite `__len__` like this and have it always work...

```python
    def __len__(self) -> int:
        return len(self.__class__.__fields__)
```

great_expectations/render/renderer_configuration.py
```diff
@@ -16,7 +16,6 @@
     Type,
     TypeVar,
     Union,
-    cast,
```
Kilo59, line 19: 👍 
IMO we should almost never use `cast`, its like a worse type-ignore because it doesn't say anything about "why" the type is wrong. However, with a type-ignore, we actually have to provide an error code.

great_expectations/render/renderer_configuration.py
```diff
@@ -590,34 +577,22 @@ def add_param(
...
+        assert isinstance(param_type, RendererValueType)
+
+        # if we already moved the suite parameter raw_kwargs to a param,
+        # we need to combine the param passed to add_param() with those existing raw_kwargs
+        if name in self.params.__dict__ and self.params.__dict__[name]["suite_parameter"]:
+            self.params.__dict__[name] = renderer_param(
+                schema=RendererSchema(type=param_type),
+                value=value,
+                suite_parameter=self.params.__dict__[name]["suite_parameter"],
+            )
```
Kilo59, line 593: Can you do this with standard dict merging rather than checking for the presence of a key?
```python
original = {"a": 1, "b": 2}
update = {"a": 2, "c": 3}
merged = {**original, **update}
```

Note: there are several different ways to merge dictionaries, ☝️ this method is generally the best and is what I use most of the time.
However in Python `3.9` which OSS is finally on, you should also be able to use the union operator `|`.
```python
original = {"a": 1, "b": 2}
update = {"a": 2, "c": 3}
merged = original | update
```

https://treyhunner.com/2016/02/how-to-merge-dictionaries-in-python/

NathanFarmer (approved):
Wow!

great_expectations/render/renderer_configuration.py
dctalbot: It seems like this would perform worse. I don't think there is any need to allocate a second dictionary and then iterate over the two in order to allocate a third. Would this be for readability or something else?

Kilo59: I'm using a generic example of merging dictionaries.
I didn't mean to suggest that you actually create 3 new dictionaries.

You would only need to consider this if the general problem you are trying to solve is merging dictionaries, which it looks like at a glance.
You have more context than me WRT to whether this is appropriate or not.

PR #10535, opened 2024-10-18
[MAINTENANCE] Bump `ruff` to `0.7.0`

cdkini:
https://astral.sh/blog/ruff-v0.7.0


Kilo59 (approved):
I know that's not the point of this PR, but all of these `TRY203/TRY302` violations should be fixed (not necessarily in this PR).
And by fixed I mean we should remove the "useless" try/except.
But in the schema of things, low priority.
https://docs.astral.sh/ruff/rules/useless-try-except/

PR #10543, opened 2024-10-21
[BUGFIX] File Path Batch Definitions serialize incorrectly

joshua-stauffer:


tests/datasource/fluent/test_schemas.py
```diff
@@ -114,8 +116,12 @@ def _sort_lists(schema_as_dict: dict) -> None:
     if "Pandas" in str(schema_dir) and _PANDAS_SCHEMA_VERSION != PANDAS_VERSION:
         pytest.xfail(reason=f"schema generated with pandas {_PANDAS_SCHEMA_VERSION}")
 
+    if pydantic.VERSION < MIN_PYDANTIC_VERSION:
```
tyler-hoffman: Is my understanding correct that 3.9 in CI is using pydantic v1, and so now we are basically just allowing those failures?

Non blocking since you aren't introducing this pattern, but I really don't think we should be dynamically xfailing. It makes it pretty opaque if this is even running at all just looking at it. I guess we're doing the same thing for pandas, so if we do keep this pattern, we should at least update the docstring (it has a note about pandas version)

My proposal: only make this run on the latest version of python we support in support, and do that with a decorator (I think our best option to implement that is `pytest.mark.skipif`, right?). It's still not great IMO, but at least we don't bury an "I guess it's okay for this to test" inside the test definition.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -110,7 +110,7 @@ class PartitionerSortingProtocol(Protocol):
     sort_ascending: bool
 
     @property
-    def param_names(self) -> list[str]:
+    def param_names(self) -> Union[list[str], tuple[str]]:
```
tyler-hoffman: This type is incorrect; `tuple[str]` indicates a tuple with exactly one string. I think you want `tuple[str, ...]`. (that doesn't solve your type ignore below, though, unfortunately)

tyler-hoffman (approved):
Approving because we need to get this fixed. But I think if you don't do it now, we need to circle back on the xfail situation in this test.

great_expectations/datasource/fluent/interfaces.py
joshua-stauffer: thanks, that's correct

joshua-stauffer: updated in [update return type](https://github.com/great-expectations/great_expectations/pull/10543/commits/b2ea48d4a8b8c575270d0d2ac736609d014ccf01)

joshua-stauffer: this actually did solve the type ignore, thanks! 🥳

tests/datasource/fluent/test_schemas.py
joshua-stauffer: thanks for taking a closer look, this was buggy.

there were two problems originally:
- i annotated the partitioners with `tuple`, and that silently failed on python 3.9.
- i misunderstood how our pydantic compatibility layer works. It essentially downgrades us to use pydantic V1, even if we have pydantic v2 installed.

There are breaking changes in schema generation between pydantic V1 and V2, so i was also encountering those when running code gen outside of the compatibility layer locally. But that was a red herring for the real problem, which was the first bullet point above.

PR #10554, opened 2024-10-23
[MAINTENANCE] Basic expectation testing framework

tyler-hoffman:
This only provides support for pandas data frame data sources.

See https://github.com/great-expectations/great_expectations/pull/10505/files#diff-c87c5173e88bb0f60b00107f4586304f4869a976a4f998425efe3481c0db9f56 for the PoC; this should look the same except a) I removed an unneeded `description` field, and b) there's some semblance of file organization now.

### Notes on file structure
I'm colocating each pair of `DataSourceConfig` and `BatchSetup`. These have circular references, at least in terms of typing, and collocating makes this simpler. Plus we should always expect the two to be added together, so I think it makes sense to colocate.

### Suggested order of reading for reviewers
1. The tests
2. `data_source_config/base.py`
3. `data_source_config/pandas_data_frame.py`
4. The conftest


tests/integration/test_utils/data_source_config/base.py
```diff
@@ -0,0 +1,72 @@
...
+from great_expectations.data_context.data_context.abstract_data_context import AbstractDataContext
+from great_expectations.datasource.fluent.interfaces import Batch
+
+if TYPE_CHECKING:
+    import pandas as pd
+    import pytest
+
+
+@dataclass(frozen=True)
+class DataSourceConfig(ABC):
```
cdkini: Nit - could we update this name and the one below to signify this is a config for a test? Just to differentiate between an actual DS config

tyler-hoffman: Yeah, I like that!

tests/integration/test_utils/data_source_config/base.py
```diff
@@ -0,0 +1,72 @@
...
+    @abstractmethod
+    def label(self) -> str:
+        """Label that will show up in test name."""
+        ...
+
+    @property
+    @abstractmethod
+    def pytest_mark(self) -> pytest.MarkDecorator:
+        """Mark for pytest"""
+        ...
```
cdkini, line 33: Is this the right place to configure the mark? Are we saying that all tests using a particular datasource are of a singular test classification? My inclination here is to make it a constructor arg to allow for updating if we happen to use it in a more complex situation?

tyler-hoffman: My assumption was that for a given data source / asset type, we would always want it to be the same. Is there a counter example you're thinking of?

tests/integration/test_utils/data_source_config/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .base import BatchSetup, DataSourceConfig
+from .pandas_data_frame import PandasDataFrameBatchSetup, PandasDataFrameDatasourceConfig
+
+__all__ = [
+    "BatchSetup",
+    "DataSourceConfig",
+    "PandasDataFrameBatchSetup",
+    "PandasDataFrameDatasourceConfig",
+]
```
cdkini: Why this?

tests/integration/data_sources_and_expectations/test_canonical_expectations.py
```diff
@@ -0,0 +1,31 @@
...
+)
+
+
+@parameterize_batch_for_data_sources(
+    data_source_configs=[
+        PandasDataFrameDatasourceConfig(),
+    ],
+    data=pd.DataFrame({"a": [1, 2]}),
+)
+def test_min(batch_for_datasource) -> None:
```
joshua-stauffer: super nit, but since we're setting the pattern - why not simplify param to just `batch`?

tests/integration/test_utils/data_source_config/base.py
tyler-hoffman: Done! lmk if you think there's something better than what I changed it to!

tests/integration/test_utils/data_source_config/__init__.py
tyler-hoffman: I've seen some linters yell about not doing this, and just added it out of habit. But there's no need. Removing it since it's just more stuff to maintain.

tests/integration/test_utils/data_source_config/base.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @abstractmethod
+    def teardown(self) -> None: ...
+
+    @staticmethod
+    def _random_resource_name() -> str:
+        return "".join(random.choices(string.ascii_lowercase, k=10))
+
+    @cached_property
+    def _context(self) -> AbstractDataContext:
```
joshua-stauffer, line 71: is there a reason we would return anything other than an ephemeral context here?

tests/integration/test_utils/data_source_config/base.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @property
+    @abstractmethod
+    def pytest_mark(self) -> pytest.MarkDecorator:
+        """Mark for pytest"""
+        ...
+
+    @abstractmethod
+    def create_batch_setup(self, data: pd.DataFrame) -> BatchTestSetup:
+        """Create a batch setup object for this data source."""
```
joshua-stauffer, line 37: this is really the critical feature of this class. I wonder if we could make that clearer from the naming convention we're using?

joshua-stauffer: calling it `...Config` is fine i think, just took me a couple readthroughs to appreciate that it wasn't a passive config object. Calling it something like `BatchTestSetupFactory` might have made that process shorter.

tests/integration/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+                (data, t),
+                id=t.test_id,
+                marks=[t.pytest_mark],
+            )
+            for t in data_source_configs
+        ]
+        parameterize_decorator = pytest.mark.parametrize(
+            batch_for_datasource.__name__,
+            pytest_params,
+            indirect=True,
```
joshua-stauffer, line 41: why do we need `indirect=True` here?

joshua-stauffer: _...proceeds to read doc string immediately below_ 😂

joshua-stauffer (approved):
this is really great! only serious suggestions i have are around naming, but i'm very happy to ship this as is and iterate on it as we go. 👏

tests/integration/test_utils/data_source_config/base.py
cdkini: Eh no worst case we refactor if we get there

cdkini (approved):

tests/integration/data_sources_and_expectations/test_canonical_expectations.py
```diff
@@ -0,0 +1,31 @@
...
+    assert result.success
+
+
+@parameterize_batch_for_data_sources(
+    data_source_configs=[
+        PandasDataFrameDatasourceTestConfig(),
+    ],
+    data=pd.DataFrame({"a": [1, 2], "b": [3, 4]}),
+)
+def test_max(batch_for_datasource) -> None:
```
cdkini: Nit - could we expand this name to be `test_expect_column_max_to_be_between` or something more explicit in case we have other "max" related tests and this file grows significantly

tests/integration/data_sources_and_expectations/test_canonical_expectations.py
tyler-hoffman: My main argument in terms of making it something less generic that just `batch` is that this is a bit of magic, and I wanted the name to be unique enough to be easily searchable. Maybe that isn't compelling, though, idk. what do you think? fwiw go-to-definition works just fine with this in my editor, so it should still be discoverable if we rename it to something like `batch`

tests/integration/test_utils/data_source_config/base.py
tyler-hoffman: Nope. I left the broader return type to denote that this isn't really about the specific type of context - it's an implementation detail that we return ephemeral.

PR #10555, opened 2024-10-23
[MAINTENANCE] Reenable xfailed e2e tests

joshua-stauffer:
We xfailed a number of e2e tests which required server-side integration work. This PR turns those tests back on and updates them to V1 syntax. 

The test design itself is problematic; there are many `asserts` within fixtures, and state easily gets stale due to the way that fixtures are scoped.


tests/integration/cloud/end_to_end/conftest.py
```diff
@@ -57,54 +56,37 @@ def datasource_name(
...
+    """This ExpectationSuite is shared by each E2E test, so its expected that the data
+    used by each test follows the same shape."""
     expectation_suite_name = f"es_{uuid.uuid4().hex}"
     expectation_suite = context.suites.add(ExpectationSuite(name=expectation_suite_name))
     assert len(expectation_suite.expectations) == 0
-    yield expectation_suite
-    expectation_suite = context.suites.get(name=expectation_suite_name)
-    assert len(expectation_suite.expectations) > 0
+    expectation_suite.add_expectation(ExpectColumnValuesToNotBeNull(column="name", mostly=1))  # type: ignore[arg-type]  # todo: fix in core-412
+    assert len(expectation_suite.expectations) == 1
```
tyler-hoffman: nit: while you're here, I'd probably delete these asserts. That functionality should be covered in unit tests. I can see an argument for the `with pytest.raises` staying, though, since that is more of an end-to-end check.

great_expectations/datasource/fluent/data_asset/path/file_asset.py
```diff
@@ -91,7 +91,10 @@ def add_batch_definition_path(self, name: str, path: PathStr) -> BatchDefinition
             raise AmbiguousPathError(path=path)
         return self.add_batch_definition(
             name=name,
-            partitioner=FileNamePartitionerPath(regex=regex),
+            partitioner=FileNamePartitionerPath(
+                regex=regex,
+                param_names=(),
```
tyler-hoffman, line 96: question/nit: we don't really have to pass in `param_names` do we?

tests/integration/cloud/end_to_end/test_pandas_datasource.py
```diff
@@ -67,90 +63,60 @@ def datasource(
...
+    return checkpoint
 
 
-@pytest.mark.xfail(reason="Fails due to a V1 change in the Checkpoint shape.", strict=True)
 @pytest.mark.cloud
-def test_checkpoint_run(checkpoint: Checkpoint):
+def test_checkpoint_run(checkpoint: Checkpoint, pandas_test_df):
     """Test running a Checkpoint that was created using the entities defined in this module."""
-    checkpoint_result: CheckpointResult = checkpoint.run()
+    checkpoint_result = checkpoint.run(batch_parameters={"dataframe": pandas_test_df})
```
tyler-hoffman: I'm confused why this works. When I dig deep into fixtures, it looks like we have this expectation: `ExpectColumnValuesToNotBeNull(column="name", mostly=1)` for a a dataframe with no `name` column.

tests/integration/cloud/end_to_end/test_pandas_filesystem_datasource.py
```diff
@@ -2,41 +2,49 @@
...
     from great_expectations.datasource.fluent.data_asset.path.pandas.generated_assets import (
         CSVAsset,
         ParquetAsset,
     )
-    from great_expectations.validator.validator import Validator
 
 
 @pytest.fixture(scope="module")
-def base_dir(tmp_path: pathlib.Path) -> pathlib.Path:
+def base_dir(tmp_path: pathlib.Path, csv_path, parquet_path) -> pathlib.Path:
```
tyler-hoffman: Obligatory public shaming nit for not including types on these

tyler-hoffman (commented):
Wow, thanks for doing all this. Left some minor feedback, but looks good to me. Once tests are passing, I'm happy to give it a quick look and a 👍

tests/integration/cloud/end_to_end/test_pandas_datasource.py
tyler-hoffman: Update: I see it's failing CI for this reason 😄

great_expectations/datasource/fluent/data_asset/path/file_asset.py
joshua-stauffer: we do, unfortunately, because [the FluentDatasource base model overrides pydantics default](https://github.com/great-expectations/great_expectations/blob/84cfd3d6b25aa5b9998e27bcb87e31b2b123a72a/great_expectations/datasource/fluent/fluent_base_model.py#L159) of `exclude_unset=False`. If we don't explicitly pass this, it gets dropped.

tests/integration/cloud/end_to_end/conftest.py
joshua-stauffer: great suggestion, will update 👍

tests/integration/cloud/end_to_end/test_pandas_datasource.py
joshua-stauffer: indeed

tyler-hoffman (approved):

PR #10556, opened 2024-10-24
[MAINTENANCE] Filesystem expectation testing

tyler-hoffman:
*TODO*: Change target to develop after the current target branch is merged!

This adds support for integration testing pandas filesystem csv assets. This necessitated a small update in the framework: `create_batch_setup` now needs access to `requests` for cases when it needs to do pytest-y stuff. This gets injected in from the `batch_for_datasource` fixture. The idea is that we then get what we need out of it as quick as we can (tmp_path in this case), and we don't pass `requests` deeper into the BatchSetup object.


tests/integration/test_utils/data_source_config/pandas_filesystem_csv.py
```diff
@@ -0,0 +1,71 @@
...
+    @property
+    @override
+    def pytest_mark(self) -> pytest.MarkDecorator:
+        return pytest.mark.filesystem
+
+    @override
+    def create_batch_setup(
+        self, data: pd.DataFrame, request: pytest.FixtureRequest
+    ) -> BatchTestSetup:
+        tmp_path = request.getfixturevalue("tmp_path")
```
cdkini, line 29: Hmm this feels a little magical but I guess there's no way around it?

tests/integration/test_utils/data_source_config/pandas_filesystem_csv.py
```diff
@@ -0,0 +1,71 @@
...
+            base_dir=tmp_path,
+        )
+
+
+class PandasFilesystemCsvBatchTestSetup(BatchTestSetup[PandasFilesystemCsvDatasourceTestConfig]):
+    def __init__(
+        self,
+        config: PandasFilesystemCsvDatasourceTestConfig,
+        data: pd.DataFrame,
+        base_dir,
```
cdkini: Nit - could we also annotate this?

tests/integration/test_utils/data_source_config/pandas_filesystem_csv.py
```diff
@@ -0,0 +1,71 @@
...
+            .get_batch()
+        )
+
+    @override
+    def setup(self) -> None:
+        file_path = self._base_dir / self.csv_path
+        self.data.to_csv(file_path, index=False)
+
+    @override
+    def teardown(self) -> None: ...
```
cdkini, line 67: Blank because `tmp_path` does its own teardown? If so, could we add a comment here about that?

cdkini: What is stopping me from passing any old pathlib.Path here? Should we do a `if self.csv_path.exists(): self.csv_path.unlink()`?

cdkini (approved):

tests/integration/test_utils/data_source_config/pandas_filesystem_csv.py
tyler-hoffman: Yeah, I don't love it either, but it's the best I could come up with. Alternatives I considered:
* have `batch_for_datasource` (the pytest.fixture) have `tmp_path` as one of the fixtures in its parameters, and then pass that to `create_batch_setup`: I decided against this because it's not needed for many data sources, and introducing that pattern could lead us to later passing a bunch of stuff into `create_batch_setup`
* not using the fixture, and just doing the temp path generation ourselves. Seems like the additional code wasn't worth it to me

tests/integration/test_utils/data_source_config/pandas_filesystem_csv.py
tyler-hoffman: Oh shame on me! absolutely.

tests/integration/test_utils/data_source_config/pandas_filesystem_csv.py
tyler-hoffman: Yeah, I'll throw in a comment. We shouldn't ever be invoking `PandasFilesystemCsvBatchTestSetup` directly, so I'd rather have 2 lines less of code and not try to clean up for weird cases like that. But LMK if you disagree.

tyler-hoffman (commented):
Updated that type annotation. LMK if you want to push back on any of my other non-changes 😄

PR #10562, opened 2024-10-24
[BUGFIX] `ExpectationSuite` equality should ignore expectation ordering

cdkini:
The fact that a store can return expectations in different orders results in a latent bug here - we sometimes fail "freshness" checks due to inconsistency. We should ignore order when determining equality and it is not relevant to how a suite functions or operates.


great_expectations/data_context/store/expectations_store.py
```diff
@@ -267,12 +267,12 @@ def _add_ids_on_create(self, suite: ExpectationSuite) -> ExpectationSuite:
         """
         suite["id"] = str(uuid.uuid4())
         if isinstance(suite, ExpectationSuite):
-            key = "expectation_configurations"
```
cdkini, line 270: Latent bug I think - if we update config (which are read-only properties), we don't actually update anything.

tyler-hoffman (approved):
LGTM. Thanks especially for the tests around expectation sorting!

PR #10571, opened 2024-10-25
[BUGFIX] Fix typing on `mostly` and `value_set` fields

tyler-hoffman:
## Context
We were previously using `Annotated` incorrectly. The correct way to use it is `Annotated[T, x]`, where `T` is the base case, and `x` is the metadata. We had those type args backward. Pydantic uses the first argument for validation, and schema generation, so schemas were generated as expected, but mypy was upset because e.g. `mostly=1` does not satisfy the type hint that mostly must be an instance of `_Mostly`. So we had to change a bunch of stuff

## About our validation vs schema.
Our validation was fairly lax, so I kept it that way in this PR. e.g. `multiple_of` was not enforced, but was just there for the schema. Pydantic actually has support out of the box for most of what we were doing, including `multiple_of`, but since we didn't enforce this before, I didn't want to start enforcing it now.

## The problem
From experimenting, it looks like constraints on nested annotations are not supported in schema generation. See https://github.com/great-expectations/great_expectations/pull/10577 for how I'd like to have written the solution, and how it fails in CI.

## Implementation details
* We're using the correct ordering in `Annotated` type args
* I'm taking advantage of `pydantic.Field` allowing for arbitrary fields to be passed in AND the `schema_extra` override on the base Expectation class. Basically during `schema_extra`, we pop out a custom field, `schema_overrides`, and write the contents of it to whatever property it was on.

I just copy/pasted what was in the snapshots of our expectation schema json for `value-set`. For `mostly`, I used out of the box functionality where I could, and only passed the `multiple_of` in `schema_overrides`. The change in how these fields are added to the schema results in a bunch of noise in the PR due to ordering changes.


great_expectations/expectations/core/schemas/ExpectColumnMostCommonValueToBeInSet.json
```diff
@@ -81,7 +81,7 @@
         },
         "value_set": {
             "title": "Value Set",
-            "description": "A list of potential values to match.",
```
tyler-hoffman, line 84: This was the only instance where we had a different description for the value set. I think it makes sense to make it the same as the others.

great_expectations/expectations/core/schemas/ExpectColumnPairValuesAToBeGreaterThanB.json
```diff
@@ -80,12 +80,12 @@
         },
         "mostly": {
             "title": "Mostly",
-            "default": 1.0,
```
tyler-hoffman, line 83: Lotta noise on these. It's functionally the same, just a different ordering for `mostly`.

great_expectations/expectations/expectation.py
```diff
@@ -312,6 +312,15 @@ def schema_extra(schema: Dict[str, Any], model: Type[Expectation]) -> None:
...
             }
 
+            # Add extra fields to schema from custom schema_overrides
+            # schema_overrides is not a pydantic concept, but pydantic.Field allows
+            # us to pass through arbitrary fields.
+            for prop in schema["properties"].values():
+                if overrides := prop.pop("schema_overrides", None):
+                    assert isinstance(overrides, dict)
+                    for key, value in overrides.items():
+                        prop[key] = value
```
joshua-stauffer: This is fine, but cleaner might be:
```suggestion
                    prop.update(overrides)
```

joshua-stauffer (approved):
looks good - thanks for fixing this

great_expectations/expectations/expectation.py
tyler-hoffman: Oh yeah, much better.

PR #10582, opened 2024-10-28
[FEATURE] PostgreSQL Expectation Tests

joshua-stauffer:
This PR adds PostgreSQL to the list of backends we can test against using the expectation integration test framework.


tests/integration/data_sources_and_expectations/test_canonical_expectations.py
```diff
@@ -25,6 +28,7 @@ def test_expect_column_min_to_be_between(batch_for_datasource) -> None:
     data_source_configs=[
         PandasDataFrameDatasourceTestConfig(),
         PandasFilesystemCsvDatasourceTestConfig(),
+        PostgreSQLDatasourceTestConfig(column_types={"a": POSTGRESQL_TYPES.INTEGER}),
```
tyler-hoffman, line 31: Not intended as a leading question: Is there a benefit to using the dialect type vs the generic, e.g. `sa.INTEGER()`?

Also wondering if it should be an instance of the type. At least the generics, you can encode a bit more info with the instance, like `sa.DateTime(timezone=False)`

tests/integration/test_utils/data_source_config/base.py
```diff
@@ -17,9 +17,13 @@
...
 
 
+_ColumnTypes = TypeVar("_ColumnTypes")
+
+
 @dataclass(frozen=True)
-class DataSourceTestConfig(ABC):
+class DataSourceTestConfig(ABC, Generic[_ColumnTypes]):
     name: Optional[str] = None
+    column_types: Union[Dict[str, _ColumnTypes], None] = None
```
tyler-hoffman, line 26: This field means nothing for non-sql types, right? I don't like to overdo it with inheritance trees, but possibly worth subclassing a `SQLDataSourceTestConfig(DataSourceTestConfig)` with this field.

tests/integration/data_sources_and_expectations/test_canonical_expectations.py
```diff
@@ -1,17 +1,20 @@
 import pandas as pd
+import sqlalchemy.dialects.postgresql as POSTGRESQL_TYPES
 
 import great_expectations.expectations as gxe
 from tests.integration.conftest import parameterize_batch_for_data_sources
 from tests.integration.test_utils.data_source_config import (
     PandasDataFrameDatasourceTestConfig,
     PandasFilesystemCsvDatasourceTestConfig,
 )
+from tests.integration.test_utils.data_source_config.postgres import PostgreSQLDatasourceTestConfig
```
tyler-hoffman: Obviously non-blocking, but I think `PostgreSQLDatasourceTestConfig` should be included in the `__init__.py` file so it can be imported alongside the other 2 config classes above.

tests/integration/test_utils/data_source_config/postgres.py
```diff
@@ -0,0 +1,195 @@
...
+        )
+
+
+class PostgresBatchTestSetup(BatchTestSetup[PostgreSQLDatasourceTestConfig]):
+    def __init__(
+        self,
+        config: PostgreSQLDatasourceTestConfig,
+        data: pd.DataFrame,
+    ) -> None:
+        self.table_name = f"postgres_expectation_test_table_{randint(0, 1000000)}"
```
tyler-hoffman, line 149: fwiw the parent class also has `_random_resource_name` if we want to standardize how we gen random stuff.

tests/integration/test_utils/data_source_config/postgres.py
```diff
@@ -0,0 +1,195 @@
...
+        columns = [Column(name, type) for name, type in self.get_column_types().items()]
+        self.table = Table(self.table_name, self.metadata, *columns, schema=self.schema)
+        self.metadata.create_all(self.engine)
+        with self.engine.connect() as conn:
+            # pd.DataFrame(...).to_dict("index") returns a dictionary where the keys are the row
+            # index and the values are a dict of column names mapped to column values.
+            # Then we pass that list of dicts in as parameters to our insert statement.
+            #   INSERT INTO test_table (my_int_column, my_str_column) VALUES (?, ?)
+            #   [...] [('1', 'foo'), ('2', 'bar')]
+            conn.execute(insert(self.table), list(self.data.to_dict("index").values()))
```
tyler-hoffman, line 184: I'm getting deprecation warnings on this line. Nesting it under `with conn.begin():` resolves it. If we update to use that, it looks like on exiting a begin context manager, the transaction is committed for us.

tyler-hoffman (approved):

tests/integration/test_utils/data_source_config/base.py
joshua-stauffer: I think it would apply to spark as well, right? At least data frames, not sure if there would be any difference writing it out to disk for a file backed system.

tests/integration/data_sources_and_expectations/test_canonical_expectations.py
joshua-stauffer: thanks, missed this

tests/integration/data_sources_and_expectations/test_canonical_expectations.py
joshua-stauffer: > Is there a benefit to using the dialect type vs the generic

The main benefit i see is that it provides the caller with a clear set of valid types for that dialect. Under the hood they're the same object. For instance, `Integer` seems like a pretty standard type, but snowflake doesn't have it - it uses `Number` instead. Although i think in the dialect `Integer` is aliased to `Number`, so you could probably get away with using the SqlAlchemy `Integer` object. But from my perspective, its a clearer and more informative test if we use the types specific to that dialect.

tests/integration/test_utils/data_source_config/postgres.py
joshua-stauffer: thanks! i'm torn about this - having something descriptive in the table name sure helps if things start going wrong, or if it doesn't get cleaned up or something.

tests/integration/test_utils/data_source_config/postgres.py
joshua-stauffer: yeah that's because you're on sqla 1.4. Our tests run on 2.0 or this would fail (because of the warning), but i'll look into tweaking it to be compatible.

joshua-stauffer: fixed in [f13a6ba](https://github.com/great-expectations/great_expectations/pull/10582/commits/f13a6ba9f87718cc80dda1cc6149f2c72a928ec2)

tyler-hoffman: thanks!

PR #10590, opened 2024-10-29
[FEATURE] Add check for valid column type when calling add_batch_def in a sql asset

billdirks:


great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -930,7 +987,7 @@ def as_selectable(self) -> sqlalchemy.Selectable:
 
         This can be used in a from clause for a query against this data.
         """
-        return sa.text(self.qualified_name)  # type: ignore[return-value]
+        return sa.table(self.table_name, schema=self.schema_name)
```
billdirks, line 1022: The return type didn't match the annotated return type. I actually need a selectable for my new check to work.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -712,6 +723,71 @@ def build_batch_request(
...
+
+        This isn't meant to be called directly. This is called internally when a batch definition
+         is added. Data asset implementers can override this for their specific data asset.
+
+        Raises:
+            SqlAddBatchDefinitionError: The specified column to partition on is not of
+                a permissible type for batching (ie date or datetime) or no data is
+                present in this column.
+        """
+        # We only support certain partitioners for using as batch definitions.
```
billdirks, line 743: I think this class's types need to be overhauled in terms of partitioners since we only support a subset how the classes typing is setup with generics would make this a big change which I didn't want to take on atm.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -712,6 +723,71 @@ def build_batch_request(
...
+        assert isinstance(
+            partitioner,
+            (
+                ColumnPartitionerYearly,
+                ColumnPartitionerMonthly,
+                ColumnPartitionerDaily,
+            ),
+        )
+        # A _SQLAsset must have a SQLDatasource
```
billdirks, line 752: I'm not sure how to enforce since with how we currently have typing setup in this class.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -81,7 +81,6 @@
 if TYPE_CHECKING:
     from sqlalchemy.sql import quoted_name  # noqa: TID251 # type-checking only
 
-    from great_expectations.compatibility import sqlalchemy
```
billdirks, line 84: We currently  import sqlalchemy in the non-typing section aliased to `sa` so removing this is cleanup.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -804,7 +880,7 @@ def _create_batch_spec(self, batch_spec_kwargs: dict) -> BatchSpec:
         """
         raise NotImplementedError
 
-    def as_selectable(self) -> sqlalchemy.Selectable:
+    def as_selectable(self) -> sa.Selectable:
```
billdirks: We had imported `sqlalchemy` as `sqlalchemy` and as `sa` so all these changes are cleanup.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -876,10 +952,8 @@ def _default_table_name(cls, table_name: str, values: dict, **kwargs) -> str:
     def _resolve_quoted_name(cls, table_name: str) -> str | quoted_name:
         table_name_is_quoted: bool = cls._is_bracketed_by_quotes(table_name)
 
-        from great_expectations.compatibility import sqlalchemy
```
billdirks, line 879: We've already imported this in the module so ruff removed this.

tests/datasource/fluent/conftest.py
```diff
@@ -483,3 +486,24 @@ def _source(
 @pytest.fixture
 def create_source() -> ContextManager:
     return _source  # type: ignore[return-value]
+
+
+@pytest.fixture
+def sql_datasource_table_asset_test_connection_noop(
```
billdirks, line 492: This fixture was moved from `test_sql_datasources.py` to `conftest.py` so I could reuse it in `test_sql_asset.py`. Both these test files test functionality in `sql_datasource.py`.

tests/datasource/fluent/data_asset/test_sql_asset.py
```diff
@@ -68,23 +77,75 @@ def test_get_batch_identifiers_list__sort_descending(postgres_asset):
         batches[i]["year"] = year
 
 
+class FakeResult:
```
billdirks, line 79: Since we actually make a request to the database now to validate the batch definition column is valid, I now have to mock out the db interaction.

tests/datasource/fluent/data_asset/test_sql_asset.py
```diff
@@ -68,23 +77,75 @@ def test_get_batch_identifiers_list__sort_descending(postgres_asset):
...
+
 @pytest.fixture
-def datasource(mocker):
+def datasource(mocker) -> SQLDatasource:
     return mocker.Mock(spec=SQLDatasource)
 
 
 @pytest.fixture
-def asset(datasource) -> _SQLAsset:
+def datasource_with_mocked_get_engine(datasource) -> SQLDatasource:
```
billdirks, line 131: Current tests die since we now make a db query when validating a batch definition. The current tests now use this fixture instead of `datasource` to mock the db query.

great_expectations/datasource/fluent/sql_datasource.py
billdirks: This is actually necessary. Added a comment here and where we import sqlalchemy in typing block.

great_expectations/datasource/fluent/sql_datasource.py
billdirks: See comments added to code. I no longer remove this line.

great_expectations/datasource/fluent/sqlite_datasource.py
```diff
@@ -109,6 +110,13 @@ def __init__(self, **kwargs):
 
     type: Literal["table"] = "table"
 
+    @override
+    def validate_batch_definition(self, partitioner: ColumnPartitioner) -> None:
```
billdirks, line 114: These sqlite assets should be public but aren't. I will mark them public and put a public api marker on these methods in a different PR.

There is already another issue about follow up sqlite work, CORE-254.

tests/datasource/fluent/data_asset/test_sql_asset.py
```diff
@@ -168,3 +232,83 @@ def test_add_batch_definition_fluent_sql__add_batch_definition_daily(
...
+        table_name="my_table",
+        schema_name="my_schema",
+    )
+
+
+def add_query_asset(datasource: SQLDatasource):
+    return datasource.add_query_asset(name="my_query_asset", query="select * from my_table")
+
+
+@pytest.mark.unit
```
billdirks, line 263: This is the test that exercises this new check. The other test changes are updating previous tests now that we may actually connect to a db.

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -484,6 +486,11 @@ def batch_parameters_to_batch_spec_kwarg_identifiers(
 ]
 
 
+class SqlAddBatchDefinitionError(Exception):
```
tyler-hoffman: Should this go in `great_expectations/exceptions/exceptions.py`?

great_expectations/datasource/fluent/sql_datasource.py
```diff
@@ -712,6 +719,73 @@ def build_batch_request(
...
+                    sa.select(column, selectable).limit(1)  # type: ignore[call-overload]  # sqlalchemy typing is missing variants
+                )
+            except Exception as query_error:
+                raise SqlAddBatchDefinitionError(
+                    msg=f"Attempt to read an example non-null '{column}' value from '{selectable}'"
+                    " failed so column type can't be verified to be a date or datetime."
+                ) from query_error
+
+            r = row.first()
+            if not r or not isinstance(r[0], (datetime, date)):
```
tyler-hoffman: Non-blocking, but `r[0]` works because we only select one column, right? Is there a way to make that more clear here, e.g. `r[partitioner.column_name]`? Not the end of the world, I just needed to think about it for an extra second.

tests/datasource/fluent/data_asset/test_sql_asset.py
```diff
@@ -168,3 +232,83 @@ def test_add_batch_definition_fluent_sql__add_batch_definition_daily(
...
+@pytest.mark.parametrize(
+    "queried_row, expect_error_or_no",
+    [
+        pytest.param(
+            [datetime(2024, 10, 28, 0, 0, 0, tzinfo=timezone.utc)],
+            lambda: ExpectNoError(),
```
tyler-hoffman: Non-blocking, but I'd like this to be considered: I find this test setup with the no-op context manager kinda confusing. What do you think about making 2 tests: one for all the cases where an error is raised, and one for the happy path?

great_expectations/datasource/fluent/sql_datasource.py
tyler-hoffman: Yeah, I think in general we could make better use of generics so this wasn't necessary, but agreed that it's needed in the current state.

tyler-hoffman (approved):

PR #10604, opened 2024-10-30
[BUGFIX] Incorrect row count when `UnexpectedRowsExpectation` returns more than 200 rows

NathanFarmer:
Bugfix for `observed_value` "X unexpected rows" when `UnexpectedRowsExpectation` returns more than 200 rows.

- Add `query.row_count` metric
- Add `unexpected_rows_query.row_count` metric, which defines the name of the Expectation parameter that contains the user's query string
- Refactor `QueryMetricProvider` helper methods to allow for query string code re-use in `query.row_count` metric
- Add query metric tests

This will increase runtimes for `UnexpectedRowsExpectation`, but computing a row count metric on large datasets is something we already do in all `ColumnMapExpectations`, as well as other Expectations.


great_expectations/expectations/metrics/query_metrics/query_row_count.py
```diff
@@ -0,0 +1,76 @@
...
+if TYPE_CHECKING:
+    from great_expectations.compatibility import pyspark
+
+
+class QueryRowCount(QueryMetricProvider):
+    metric_name = "query.row_count"
+    value_keys = ("query",)
+
+    @metric_value(engine=SqlAlchemyExecutionEngine)
+    def _sqlalchemy(
```
NathanFarmer, line 27: `UnexpectedRowsQuery` only supports SQL and Spark (see instruction 2 [here](https://docs.greatexpectations.io/docs/core/customize_expectations/use_sql_to_define_a_custom_expectation#procedure)). These two execution engines are all that has been implemented for `query.table` as well.

tests/expectations/core/test_unexpected_rows_expectation.py
```diff
@@ -60,54 +61,38 @@ def test_unexpected_rows_expectation_invalid_query_info_message(query: str, capl
...
-                    "trip_distance": 0.0,
-                }
-            ],
+            1,
             id="failure",
         ),
+        pytest.param(
+            "SELECT * FROM {batch} WHERE passenger_count > 0",
+            False,
+            "97853 unexpected rows",
```
NathanFarmer, line 84: Bugfix assertion

great_expectations/expectations/core/unexpected_rows_expectation.py
```diff
@@ -138,7 +141,7 @@ def _validate(
         execution_engine: ExecutionEngine | None = None,
     ) -> Union[ExpectationValidationResult, dict]:
         metric_value = metrics["unexpected_rows_query.table"]
-        unexpected_row_count = len(metric_value)
+        unexpected_row_count = metrics["unexpected_rows_query.row_count"]
```
dctalbot, line 144: I wonder if these magic strings should be stored in variables?

great_expectations/expectations/metrics/query_metrics/query_row_count.py
```diff
@@ -0,0 +1,76 @@
...
+                execution_engine=execution_engine,
+            )
+        )
+        count_column_name = "unexpected_row_count"
+        row_count_query = (
+            f"SELECT COUNT(*) as {count_column_name} FROM "
+            f"({substituted_batch_subquery}) AS substituted_batch_subquery"
+        )
+        result: Union[Sequence[sa.Row[Any]], Any] = execution_engine.execute_query(
+            sa.text(row_count_query)  # type: ignore[arg-type]
```
dctalbot: Can you add a brief justification for the type ignore as part of the code comment?

dctalbot (approved):

great_expectations/expectations/core/unexpected_rows_expectation.py
NathanFarmer: I could refer to `UnexpectedRowsQueryRowCount.metric_name`. It would be a new pattern to use everywhere. We have plans to re-write the metrics engine entirely, so I'm not sure it's worth the effort right now.

great_expectations/expectations/metrics/query_metrics/query_row_count.py
NathanFarmer: I was actually able to fix it (in multiple places) 🥳

PR #10611, opened 2024-10-31
[BUGFIX] Parse spaces in row condition column name

ashmortar:
Use `QuotedString` from pyparsing to parse column_name from submitted string to allow for spaces in column names.  Reported in [this thread on discourse](https://discourse.greatexpectations.io/t/expectations-with-the-row-condition-throws-an-exception-when-either-column-or-value-have-space/1247)

### DIFF
--- a/great_expectations/expectations/row_conditions.py
+++ b/great_expectations/expectations/row_conditions.py
@@ -11,11 +11,11 @@ from pyparsing import (
     Combine,
     Literal,
     ParseException,
+    QuotedString,
     Regex,
     Suppress,
     Word,
     alphanums,
-    alphas,
 )
 
 import great_expectations.exceptions as gx_exceptions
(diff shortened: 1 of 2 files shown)


great_expectations/expectations/row_conditions.py
```diff
@@ -34,11 +34,7 @@ def _set_notnull(s, l, t) -> None:  # noqa: E741 # ambiguous name `l`
 
 
 WHITESPACE_CHARS = " \t"
-column_name = Combine(
-    Suppress(Literal('col("'))
```
NathanFarmer, line 38: Do you understand why removing `Suppress` allows this to continue working? I don't see us referencing `column_name.column` anywhere. Why not leave `Suppress` in there? I'm not very familiar with `pyparsing`.

ashmortar: When you use `setResultsName` on a portion of a combined matcher it only uses the portion that you are calling that method on.

So the full expression basically _matches_ for `col("<anytext>")` but will only set `<anytext>` under the key `column` in the results dict.

I.E. the code on develop would work exactly the same if you remove `Supress(` completely from `column_name`.  This can be verified by removing it and running the unit tests again.

NathanFarmer (approved):

PR #10613, opened 2024-11-01
[BUGFIX] Batch Expectations correctly handle `date` min and max values

joshua-stauffer:

### DIFF
--- a/tests/integration/data_sources_and_expectations/test_canonical_expectations.py
+++ b/tests/integration/data_sources_and_expectations/test_canonical_expectations.py
@@ -1,10 +1,10 @@
 from datetime import datetime, timezone
 
 import pandas as pd
-import sqlalchemy.dialects.postgresql as POSTGRESQL_TYPES
 
 import great_expectations.expectations as gxe
 from great_expectations.compatibility.snowflake import SNOWFLAKE_TYPES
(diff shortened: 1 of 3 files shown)


tests/integration/data_sources_and_expectations/test_canonical_expectations.py
```diff
@@ -1,8 +1,10 @@
+from datetime import datetime, timezone
+
 import pandas as pd
-import sqlalchemy.dialects.postgresql as POSTGRESQL_TYPES
```
joshua-stauffer, line 2: we shouldn't import directly from sqlalchemy

tyler-hoffman: Note to non-Josh folks: the postgresql types were just imported from the sqla types, so they are equivalent. Which explains mypy being fine with this change.

tyler-hoffman (approved):

tyler-hoffman:
Why'd this break now (or I'm assuming as of v1?)

joshua-stauffer:
it seems like this broke because V1 now deserializes the `min_value` and `max_value` fields on `BatchExpectations` as `datetime.date` rather than `datetime.datetime`, if they are a date rather than a timestamp.

PR #10618, opened 2024-11-01
[FEATURE][WIP] Update metrics.column_types to use the DB type and return string value

TrangPham:


great_expectations/expectations/core/expect_column_values_to_be_of_type.py
```diff
@@ -407,6 +407,43 @@ def _validate_pandas(
         }
 
     def _validate_sqlalchemy(self, actual_column_type, expected_type, execution_engine):
+        success = False
+        if expected_type is None:
+            success = True
+
+        if (
+            isinstance(actual_column_type, str)
+            and actual_column_type.lower() == expected_type.lower()
```
TrangPham, line 387:
```suggestion
            and isinstance(expected_type, str)
            and actual_column_type.lower() == expected_type.lower()
```

TrangPham:
Marking as ready for review so that the rest of the CI test suite runs

great_expectations/expectations/metrics/util.py
```diff
@@ -30,8 +30,8 @@
 )
 from great_expectations.compatibility.typing_extensions import override
 from great_expectations.execution_engine import (
-    PandasExecutionEngine,  # noqa: TCH001
-    SqlAlchemyExecutionEngine,  # noqa: TCH001
+    PandasExecutionEngine,
```
Shinnnyshinshin, line 33: looks like `static-analysis` is complaining. Should be pretty simple to move the import statement to a different block

tests/integration/data_sources_and_expectations/expectations/test_expect_column_values_to_be_of_type.py
```diff
@@ -66,6 +65,18 @@ def test_success_for_type__INTEGER(batch_for_datasource: Batch) -> None:
...
 
 
+@parameterize_batch_for_data_sources(
+    data_source_configs=[
+        BigQueryDatasourceTestConfig(),
+    ],
+    data=DATA,
+)
+def test_success_for_type__INT64(batch_for_datasource: Batch) -> None:
+    expectation = gxe.ExpectColumnValuesToBeOfType(column=INTEGER_COLUMN, type_="INT64")
```
Shinnnyshinshin, line 75: 🚀

docs/docusaurus/docs/reference/learn/data_quality_use_cases/schema_resources/schema_strict_and_relaxed.py
```diff
@@ -85,7 +85,7 @@
 
 relaxed_suite.add_expectation(
     gxe.ExpectColumnValuesToBeInTypeList(
-        column="transfer_amount", type_list=["DOUBLE_PRECISION", "STRING"]
+        column="transfer_amount", type_list=["DOUBLE PRECISION", "STRING"]
```
Shinnnyshinshin, line 88: @TrangPham This is a simple question (but just to confirm my understanding), but could you describe why this is changing from `DOUBLE_PRECISION` to `DOUBLE PRECISION`?

https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#postgresql-data-types

Shinnnyshinshin: My understanding : SQALCHEMY calls it `DOUBLE_PRECISION` and POSTGRES actually calls it `DOUBLE PRECISION` (with no underscore). We are going with what postgres calls it.

tests/test_definitions/column_map_expectations/expect_column_values_to_be_in_type_list.json
```diff
@@ -110,12 +138,438 @@
...
+            "success": false,
+            "unexpected_list": ["hello", "jello", "1"],
+            "unexpected_index_list": [{"s": "hello", "pk_index": 0}, {"s": "jello", "pk_index": 1}, {"s": "1", "pk_index": 2}]
+          },
+          "only_for": [
+            "pandas"
+          ]
+        },
+                {
+          "title": "negative_test_string_and_int_values_complete_output",
```
Shinnnyshinshin, line 491: A nit, but in cases where a test is only for a particular backend or execution engine, would it be possible to add it to the `title`?  So this would become `negative_test_string_and_int_values_complete_output_pandas` or something similar.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -186,7 +193,20 @@ def _get_dialect_type_module(dialect):  # noqa: C901
             and teradatatypes is not None
         ):
             return teradatatypes
-    except (TypeError, AttributeError):
+    except (TypeError, AttributeError, ModuleNotFoundError):
+        pass
+
+    # Clickhouse types module
+    try:
```
Shinnnyshinshin, line 200: Confirming understanding here. Were we not correctly checking whether the `Clickhouse` module was imported successfully before?

tests/integration/data_sources_and_expectations/test_misconfigured_expectations.py
```diff
@@ -67,7 +67,7 @@ def test_pandas(self, batch_for_datasource) -> None:
     def test_bigquery(self, batch_for_datasource) -> None:
         self._assert_misconfiguration(
             batch_for_datasource=batch_for_datasource,
-            exception_message="No matching signature for operator * for argument types: FLOAT64, STRING",  # noqa: E501
```
Shinnnyshinshin, line 70: Could you help me understand this change better? Would we still be missing the operator for `*` for `STRING`?

tests/expectations/core/test_expect_column_values_to_be_of_type.py
```diff
@@ -31,15 +31,15 @@ def test_expect_column_values_to_be_of_type_string_dialect_pyathena(sa):
     # Monkey-patch dialect for testing purposes.
     validator.execution_engine.dialect_module = aws.sqlalchemy_athena
 
-    result = validator.expect_column_values_to_be_of_type("col", type_="string")
+    result = validator.expect_column_values_to_be_of_type("col", type_="TEXT")
```
Shinnnyshinshin, line 34: `TEXT` is `sqlite` type, without being translated into `sqlalchemy`

great_expectations/util.py
```diff
@@ -979,45 +978,6 @@ def import_make_url():
     return make_url
 
 
-def get_clickhouse_sqlalchemy_potential_type(type_module, type_) -> Any:
```
Shinnnyshinshin, line 982: moved to `expectations/util.py`

docs/docusaurus/docs/reference/learn/data_quality_use_cases/schema_resources/schema_strict_and_relaxed.py
TrangPham: Yes that's correct. We are using SQLAlchemy's compile function to compile the SQLAlchemy type to the underlying DB type. 

This allows us to do a string compare later instead of needing to find the SQLAlchemy class for the type on the engine. IE no more get engine module and find attribute on the module.

tests/test_definitions/column_map_expectations/expect_column_values_to_be_in_type_list.json
TrangPham: Sounds good.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
TrangPham: That's correct. Previously there were two functions in different places which did the same/similar thing. I combined them and made sure all the checks are in place.

tests/integration/data_sources_and_expectations/test_misconfigured_expectations.py
TrangPham: This was erroring with a slightly different error message for me. Which leads me to believe the error message was updated in the underlying code and there's a difference based on version of the underlying library used. 

In my CI test it was showing with similar message but after the * it had a new line then the types.

great_expectations/expectations/metrics/util.py
```diff
@@ -461,13 +461,28 @@ def get_sqlalchemy_column_metadata(
...
-        return columns
+            return case_insensitive_columns
+        else:
+            # WARNING: Do not edit columns in place. It will affect the underlying inspector object.
+            columns_copy = [column.copy() for column in columns]
+
+            # Note: Column reflection fallback does not always return types
+            for column in columns_copy:
+                if "type" in column:
+                    column["type"] = column["type"].compile(dialect=engine.dialect)
```
Shinnnyshinshin, line 483: confirming my understanding : this is converting the column type (a `str`) to the equivalent sqlalchemy type?

great_expectations/expectations/metrics/util.py
```diff
@@ -461,13 +461,28 @@ def get_sqlalchemy_column_metadata(
...
+
+            return columns_copy
```
alena-hutchinson, line 485: Can this `if, else` block be simplified? It seems like the only difference is in what we are defining as `columns` here, and the `for` loop is the same in both cases. A comment explaining why we want `columns` to be different depending on the dialect would also be helpful.

alena-hutchinson (changes requested):
Is it possible to break this into two separate PRs? It seems like a good chunk of the diffs are from cleaning up/moving code but not altering functionality. The cleanup is awesome but makes it a little harder to hone in on reviewing the pieces of code where functionality is being changed, especially since this PR has a larger number of files.

tests/test_definitions/column_map_expectations/expect_column_values_to_be_of_type.json
```diff
@@ -33,6 +33,12 @@
         },
         "redshift": {
           "x": "INTEGER"
+        },
+        "bigquery": {
+          "x": "INT64"
+        },
+        "databricks": {
+          "x": "INT"
```
Shinnnyshinshin, line 41: @TrangPham could we confirm that these new schemas are actually running? I see a `only_for: ["pandas"]` below on (line 56) and suspect it might be skipping most of them.

tests/test_definitions/column_map_expectations/expect_column_values_to_be_in_type_list.json
```diff
@@ -110,12 +138,438 @@
...
+              "BOOLEAN"
+            ]
+          },
+          "out": {
+            "success": true
+          },
+          "only_for": [
+            "sqlalchemy"
+          ],
+          "suppress_test_for": ["mysql", "mssql"]
```
Shinnnyshinshin, line 402: nit: are we missing a `bool` test for `mysql`?

Shinnnyshinshin: n/m it's a `TINYINT` for `mysql`. Interesting :)

docs/docusaurus/docs/reference/learn/data_quality_use_cases/schema_resources/schema_strict_and_relaxed.py
tyler-hoffman: How do we message this change? Is this a breaking change or a bugfix? Does this have any other affect other than `"_"` -> `" "`, e.g. are there any types/dbs where it will map it do something else, like `DOUBLE_PERCISION` -> `DOUBLE`? I think it might be worth getting more eyes on this, but I also don't think we need to overthink it.

great_expectations/execution_engine/sqlalchemy_execution_engine.py
```diff
@@ -144,11 +144,18 @@
     teradatasqlalchemy = None
     teradatatypes = None
 
+try:
+    import clickhouse_sqlalchemy
```
tyler-hoffman, line 148: I think this should get moved to our `compatibility` directory and follow the patterns there. But do we care about clickhouse?

great_expectations/expectations/core/expect_column_values_to_be_in_type_list.py
```diff
@@ -457,61 +463,51 @@ def _validate_pandas(  # noqa: C901, PLR0912
...
-                        types.append(potential_type)
-                except AttributeError:
-                    logger.debug(f"Unrecognized type: {type_}")
+        elif (
+            isinstance(actual_column_type, str)
+            # Note: expected_types_list is a list of strings in the configuration of the expectation
+            and actual_column_type.lower()
+            in [expected_type.lower() for expected_type in expected_types_list]
+        ):
+            # In newer versions of GX, metric.column_type returns a string
```
tyler-hoffman, line 476: I'm not sure this helps. By "newer versions", are you referring to the version this goes out in, and anything going forward? I think it might be more clear to just remove the comment and make it clear via type hints.

great_expectations/expectations/core/expect_column_values_to_be_in_type_list.py
```diff
@@ -457,61 +463,51 @@ def _validate_pandas(  # noqa: C901, PLR0912
...
+            # of the DB type, so we can compare directly.
+            success = True
 
-            if len(types) == 0:
-                logger.warning("No recognized sqlalchemy types in type_list for current dialect.")
-            types = tuple(types)
-            success = isinstance(actual_column_type, types)
+        if not success:
+            # Fallback to the old method of validating types.
+            success = self._deprecated_validate_sqlalchemy(
```
tyler-hoffman, line 482: I assume this is related to my above question?

When do we fall back to this? Should we know beforehand which branch will be relevant? This kind of logic makes me concerned that we don't trust the system.

great_expectations/expectations/core/expect_column_values_to_be_in_type_list.py
```diff
@@ -457,61 +463,51 @@ def _validate_pandas(  # noqa: C901, PLR0912
...
+        ):
+            # In newer versions of GX, metric.column_type returns a string
+            # of the DB type, so we can compare directly.
+            success = True
 
-            if len(types) == 0:
-                logger.warning("No recognized sqlalchemy types in type_list for current dialect.")
-            types = tuple(types)
-            success = isinstance(actual_column_type, types)
+        if not success:
```
tyler-hoffman, line 480: This can just be an `else` block, right? I think that would make the flow a bit easier to follow.

great_expectations/expectations/metrics/util.py
```diff
@@ -461,13 +461,28 @@ def get_sqlalchemy_column_metadata(
...
+                if "type" in col:
+                    col["type"] = col["type"].compile(dialect=engine.dialect)
 
-        return columns
+            return case_insensitive_columns
+        else:
+            # WARNING: Do not edit columns in place. It will affect the underlying inspector object.
+            columns_copy = [column.copy() for column in columns]
+
+            # Note: Column reflection fallback does not always return types
```
tyler-hoffman, line 480: The types we currently have here don't make it clear what's what, and I'm unsure how to do with the info in this comment. What else can it return? Is this comment saying we might have non-strings at this point, or after the call to `compile`?  Either way, seems like the types of `columns_copy` before and after it is mutated are different, and it would likely be simpler to understand if they were 2 separate dicts with good types.

great_expectations/expectations/util.py
```diff
@@ -0,0 +1,105 @@
...
+
+from packaging import version
+
+from great_expectations.compatibility.bigquery import BIGQUERY_GEO_SUPPORT
+from great_expectations.execution_engine.sqlalchemy_dialect import GXSqlDialect
+
+logger = logging.getLogger(__name__)
+
+
+def get_clickhouse_sqlalchemy_potential_type(type_module, type_) -> Any:
```
tyler-hoffman, line 16: Note to self: this file looks to just be utils we moved from other places.

tests/expectations/core/test_expect_column_values_to_be_in_type_list.py
```diff
@@ -30,7 +30,7 @@ def test_expect_column_values_to_be_in_type_list_dialect_pyathena_string(sa):
     validator.execution_engine.dialect_module = aws.sqlalchemy_athena
 
     result = validator.expect_column_values_to_be_in_type_list(
-        "col", type_list=["string", "boolean"]
+        "col", type_list=["string", "boolean", "TEXT"]
```
tyler-hoffman, line 33: This addition is because the call to `compile` now maps the class `TEXT` to the string `"TEXT"`?

tests/integration/data_sources_and_expectations/expectations/test_expect_column_values_to_be_in_type_list.py
```diff
@@ -36,38 +36,61 @@
...
+            STRING_COLUMN,
+            [
+                "str",
+                "string",
+                "StringType",
+                "VARCHAR",
+                "STRING",
+                "VARCHAR(255)"  # PostgreSQL,
+                "VARCHAR(16777216)",  # Snowflake
+                "VARCHAR(max) COLLATE SQL_Latin1_General_CP1_CI_AS",  # SQL Server
```
tyler-hoffman, line 73: that's quite the type

tests/integration/data_sources_and_expectations/expectations/test_expect_column_values_to_be_of_type.py
tyler-hoffman: Is this a bugfix to the bigquery bigint bug I've heard about?

tests/integration/data_sources_and_expectations/test_misconfigured_expectations.py
tyler-hoffman: Yeah, we started hitting this elsewhere. Updated in https://github.com/great-expectations/great_expectations/pull/10750.

tests/test_definitions/column_map_expectations/expect_column_values_to_be_of_type.json
```diff
@@ -33,6 +33,12 @@
         },
         "redshift": {
           "x": "INTEGER"
+        },
+        "bigquery": {
```
tyler-hoffman, line 37: We shouldn't rely on these json files any more. I've been trying to just leave behind one test in these (the existence of these tests is used by some test about rendered content where it counts the number of expectations, and I haven't had bandwidth to replace that). But I would definitely not add to this file.

tyler-hoffman (changes requested):
The changes look good to me, assuming we want to move from sqla types to raw strings. Is there a convo on this I can follow somewhere? My understanding is it's related to FE changes?

Repeating some inline comments, but things to consider before merging:
* Smaller: no additions to the `json` test files
* Bigger: It's unclear to me how to frame this in terms of bugfix/maintenance/feature and how that plays into semver. I'm concerned about this being a breaking change that technically requires a major version bump (which I don't think is acceptable from a product perspective). Leaving a blocker on here until that is resolved.

tests/integration/data_sources_and_expectations/expectations/test_expect_column_values_to_be_of_type.py
TrangPham: Yes it would fix that one too I believe, though have not specifically tested with bigquery E2E

docs/docusaurus/docs/reference/learn/data_quality_use_cases/schema_resources/schema_strict_and_relaxed.py
TrangPham: Thanks for raising the concern. After discussing this with others, we will consider this a bug fix and we will limit the scope of the change based on the dialect as well as still maintaining the fallback for expectation evaluation. The fallback will ensure that existing expectations will continue to pass if they have been passing.

great_expectations/expectations/metrics/util.py
TrangPham: What is happening here is 

`column["type"]` holds an instance of a SQLAlchemy type class. When we `compile` it, that returns a String which is the actual DB type. 

So for example:

```
x = sqlalchemy.types.BigInteger
x.compile(dialect=databricks)
print(x)
```

Would output "BIGINT"

great_expectations/expectations/metrics/util.py
TrangPham: I tried that but the mypy type checking really hated me

TrangPham:
Closed in favour of smaller PRs

PR #10619, opened 2024-11-02
[MAINTENANCE] Prevent unneeded test setup/teardown

tyler-hoffman:
See the comment toward the bottom of `conftest.py` for details on the approach. But a couple notes on some of the changes here:
* We now have a mapping of `TestConfig` -> `BatchTestSetup` that is shared across test runs so that we can use them when appropriate and only run setup/teardown once
* For this behavior to work, we need `TestConfig` to be hashable, so I implemented both `__hash__` and `__eq__`.
* Because of ^, it was easiest just to push `extra_data` down to the base `BatchTestSetup`. I'm interested in reworking this a bit in a subsequent refactor.
* Because gx doesn't support use of multiple concurrent contexts, there's now a call to `set_context` in our fixture before we run yield to the test. Without this, tests were able to run with different tests' contexts, resulting in errors around not finding the datasource referenced by batches
* Because of ^, `BatchTestSetup._context` was made public as `BatchTestSetup.context`


tests/integration/data_sources_and_expectations/test_canonical_expectations.py
```diff
@@ -148,29 +148,29 @@ class TestExpectTableRowCountToEqualOtherTable:
         data_source_configs=[
             PostgreSQLDatasourceTestConfig(
                 column_types={"col_a": sqltypes.INTEGER},
-                extra_assets={"test_table_two": {"col_b": sqltypes.VARCHAR}},
+                extra_assets={"test_table_a": {"col_b": sqltypes.VARCHAR}},
```
tyler-hoffman: TODO in a subsequent task: figure out how to generate these names ourselves like we do for the main asset. https://greatexpectations.atlassian.net/browse/CORE-586

tests/integration/conftest.py
```diff
@@ -66,20 +103,42 @@ def decorator(func: _F) -> _F:
     return decorator
 
 
+# NOTE on performance setup/teardown:
+# When we get equivalent TestConfigs, we only instantiate one BatchTestSetup for all of them, and
+# only perform its setup/teardown once. batch_for_datasource instantiate the BatchTestSetup
+# immediately before the first test that needs it and store it in cached_test_configs. Subsequetn
```
billdirks: s/Subseqetn/Subsequent/

tests/integration/conftest.py
```diff
@@ -4,18 +4,52 @@
...
+                ),
+            )
+        )
+
+    @override
+    def __eq__(self, value: object) -> bool:
+        # We need to implement this ourselves to call `.equals` on dataframes.`
+        if not isinstance(value, TestConfig):
+            return False
+        return all(
```
billdirks, line 44: Can we do an equality check on the hashes? Not sure if that's better but this seems to very similar to that and then we'd only have 1 block of code to update if this changed.

tests/integration/data_sources_and_expectations/test_test_performance.py
```diff
@@ -0,0 +1,104 @@
...
+)
+def test_caching_within_a_test(batch_for_datasource) -> None:
+    # This should fail in setup or teardown if the setup and teardown are not being cached
+    ...
+
+
+@parameterize_batch_for_data_sources(
+    data_source_configs=[
+        DummyTestConfig(column_types={"a": POSTGRESQL_TYPES.INTEGER}),
+        DummyTestConfig(column_types={"a": POSTGRESQL_TYPES.INTEGER}),
```
billdirks: Do we need both of these in this test? We assume this test is running after the first one?

tests/integration/test_utils/data_source_config/base.py
```diff
@@ -53,6 +55,33 @@ def test_id(self) -> str:
...
+            [
+                super().__eq__(value),
+                self.label == value.label,
+                self.pytest_mark == value.pytest_mark,
+            ]
+        )
+
+    @override
+    def __hash__(self) -> int:
+        hashable_col_types = dict_to_tuple(self.column_types) if self.column_types else None
```
billdirks, line 72: If 2 objects have the same value of `__eq__` are they guaranteed to have the same hash? This seems to be a requirement of hash (eg so putting them in dictionaries works as expected): [https://docs.python.org/3/reference/datamodel.html#object.\_\_hash\_\_](https://docs.python.org/3/reference/datamodel.html#object.__hash__)

tests/integration/conftest.py
tyler-hoffman: I want to push back a bit here. I considered similar approaches. I went with this because, while more tedious, it saves us from the (very unlikely) case of hash collisions. But the likelihood of that is sooo small, so I'm totally happy to change to a hash comparison, but wanted to lay out my thoughts. LMK.

tests/integration/data_sources_and_expectations/test_test_performance.py
tyler-hoffman: Good call. Even if this doesn't run after the first one, I think it still proves the same thing.

tests/integration/test_utils/data_source_config/base.py
tyler-hoffman: >If 2 objects have the same value of __eq__ are they guaranteed to have the same hash

If they have the same value of __eq__ AND implement __hash__, it must be the same, but most (probably all well defined?) mutable objects do not implement hash.
```
>>> hash({"foo": "bar"})
TypeError: unhashable type: 'dict'
```
The trick I've always seen is to use tuples, but maybe there's something cleaner?

tests/integration/conftest.py
tyler-hoffman: Let the record show that I need to request changes to your request of changes: `s/Subseqetn/Subsequetn/` 😛 

But thanks, fixed!

tests/integration/conftest.py
```diff
@@ -66,20 +103,52 @@ def decorator(func: _F) -> _F:
...
+# session, _cleanup will clean up all the BatchTestSetups.
+
+
+@pytest.fixture(scope="session")
+def _cached_test_configs() -> dict[TestConfig, BatchTestSetup]:
+    """Fixture to hold cached test configurations across tests."""
+    cached_test_configs: dict[TestConfig, BatchTestSetup] = {}
+
+    return cached_test_configs
+
```
joshua-stauffer, line 120: thanks for putting this in a fixture!

joshua-stauffer (approved):
✅

tests/integration/conftest.py
billdirks: You are right, this is better.

tests/integration/test_utils/data_source_config/base.py
billdirks: I don't know if we are talking about the same thing? Using tuples is fine. I'm asking if `a == b` are we guaranteed that `a.__hash__() == b.__hash__()`?  Maybe it's true but it wasn't apparent when I read this code.

My link didn't work because the last characters got removed from the url (fixed now) but I'm referring to this from docs:
> The only required property is that objects which compare equal have the same hash value; it is advised to mix together the hash values of the components of the object that also play a part in comparison of objects by packing them into a tuple and hashing the tuple.

If we use this for a key in a dict, and this isn't true, the lookup can fail since both hash and equality are used in the lookup.

tyler-hoffman: Oh yeah, I misunderstood your question. My intention was definitely the implement this so if a == b are we guaranteed that a.__hash__() == b.__hash__(). I'm pretty confident it does this. But LMK if you aren't convinced (or if I'm still missing something 😄 )

PR #10624, opened 2024-11-04
[DOCS] remove unsupported actions

klavavej:
resolves https://greatexpectations.atlassian.net/browse/DOC-844 - removes unsupported actions from the API docs (opsgenie, pagerduty, SNS)

I'm leaving `UpdateDataDocsAction` and `ValidationAction` in place as I assume those are supported and the ticket was about third party integrations that aren’t currently supported. Let me know if I’m wrong though


great_expectations/checkpoint/actions.py
```diff
@@ -352,7 +352,6 @@ def _send_slack_notification(self, payload: dict) -> dict:
         return {"slack_notification_result": slack_notif_result}
 
 
-@public_api
```
billdirks, line 355: Are these actions no longer GX support or no longer community supported either?

That is, do we want community supported actions in our public docs and labeled as such or not appearing at all.

If these aren't community supported, can we delete the actions altogether?

klavavej: > Are these actions no longer GX support or no longer community supported either?

These actions are unsupported by GX _and_ unsupported by the community (source of truth for integration support is https://docs.greatexpectations.io/docs/application_integration_support#integrations) 

> If these aren't community supported, can we delete the actions altogether?

Yes, but that's outside the scope of this docs PR. I filed https://greatexpectations.atlassian.net/browse/CORE-600 for this and product/eng can then prioritize as y'all see fit (I'm not sure how impactful the tech debt is / if it's worth your time to clean it up or easier to just leave it in place in case product decides we should support these integrations in the future)

billdirks: We should always remove dead code.

docs/docusaurus/static/_redirects
```diff
@@ -73,7 +73,7 @@ http://legacy.016.docs.greatexpectations.io/* http://docs.greatexpectations.io 3
 /en/latest/guides/how_to_guides/miscellaneous/how_to_template.html https://legacy.docs.greatexpectations.io/en/latest/guides/how_to_guides/miscellaneous/how_to_template.html
 /en/latest/guides/how_to_guides/miscellaneous/how_to_add_and_test_a_new_sqlalchemydataset_class.html https://legacy.docs.greatexpectations.io/en/latest/guides/how_to_guides/miscellaneous/how_to_add_and_test_a_new_sqlalchemydataset_class.html
 /en/latest/guides/how_to_guides/miscellaneous/how_to_use_official_docker_images.html https://legacy.docs.greatexpectations.io/en/latest/guides/how_to_guides/miscellaneous/how_to_use_official_docker_images.html
-/en/latest/guides/how_to_guides/miscellaneous/how_to_setup_opsgenie_alert_notifications.html https://legacy.docs.greatexpectations.io/en/latest/guides/how_to_guides/miscellaneous/how_to_setup_opsgenie_alert_notifications.html
+/en/latest/guides/how_to_guides/miscellaneous/how_to_setup_opsgenie_alert_notifications.html /docs/0.18/reference/api/checkpoint/opsgeniealertaction_class/
```
billdirks: Why aren't we completely removing opgenie here if it is not supported? Is it because the code still exists? When we remove that code, will we have to remove this code and update this file again?

tests/checkpoint/test_checkpoint.py
```diff
@@ -651,12 +649,8 @@ def test_checkpoint_sorts_actions(self, validation_definition: ValidationDefinit
         Pydantics and mocks.
         Ideally, this would be tested through the public `run()` method.
         """
-        pd_action = PagerdutyAlertAction(
-            name="my_pagerduty_action", api_key="api_key", routing_key="routing_key"
-        )
-        og_action = OpsgenieAlertAction(name="my_opsgenie_action", api_key="api_key")
         data_docs_action = UpdateDataDocsAction(name="my_docs_action")
-        actions: List[CheckpointAction] = [pd_action, og_action, data_docs_action]
+        actions: List[CheckpointAction] = [data_docs_action]
```
billdirks: This test is suppose to verify sorting actions. If there is only 1 action, this test is no longer verifying this. Could you add actions to replace the ones you removed?

docs/sphinx_api_docs_source/public_api_excludes.py
```diff
@@ -827,4 +827,19 @@
         name="get_batch_parameters_keys",
         filepath=pathlib.Path("great_expectations/datasource/fluent/sql_datasource.py"),
     ),
+    IncludeExcludeDefinition(
```
billdirks, line 830: If we were to remove the unsupported actions would we have to undo this change and remove these?

billdirks (changes requested):
Thanks for working towards removing unsupported docs. It looks like the updated test no longer tests what the test name claims it does.

docs/docusaurus/static/_redirects
billdirks: That is, if we need to remove it later, could you add this to the ticket you filed, CORE-600?

docs/sphinx_api_docs_source/public_api_excludes.py
billdirks: If so, could you add it as a task to do in CORE-600?

docs/docusaurus/static/_redirects
klavavej: hmmm good point! I forgot this part of our support posture -https://docs.greatexpectations.io/docs/application_integration_support#gx-versions - I'll revise this redirect

tests/checkpoint/test_checkpoint.py
klavavej: I'll try! But I'm definitely out of my comfort zone here so if the update I make doesn't do the trick I'd love to pair on this

docs/sphinx_api_docs_source/public_api_excludes.py
klavavej: > If we were to remove the unsupported actions would we have to undo this change and remove these?

I assume yes, but I'm not 100% confident in my understanding of how the api docs get built

> If so, could you add it as a task to do in CORE-600?

yes, will do!

great_expectations/checkpoint/actions.py
billdirks: Thinking about this more, maybe we need to leave the code in for semantic versioning. Removing the public api decorator will necessitate a minor version bump.

tests/checkpoint/test_checkpoint.py
billdirks: Thanks for doing this! Happy to pair.

billdirks (approved):
LGTM, thanks!

docs/sphinx_api_docs_source/public_api_excludes.py
tyler-hoffman: @klavavej the api docs build is complex. I'm happy to chat about it.

PR #10630, opened 2024-11-05
[MAINTENANCE] dynamically generate extra table names in tests

tyler-hoffman:
Resolves CORE-586. We now randomly generate table names.

**The bad news**
There is now a semantic change change to the api, and I don't love it. The keys in `extra_data` and `extra_column_types` are now labels for correlation, NOT the table name. I suspect this will bite someone. I originally wanted to use lists for each and just do away with the labels/names, but since we will typically infer table types, we can't guarantee that `extra_data` and `extra_column_types` are the same length.

**The good news**
To make this work, I created a new fixture ,`batch_setup_for_datasource`, that is used by the other fixtures. The other fixtures can pick off just the stuff they want. I think this opens up some really cool doors for expanding the framework down the road. E.g. we could really easily make another fixture that rather than getting a batch, gets a checkpoint.


tests/integration/test_utils/data_source_config/sql.py
```diff
@@ -71,8 +72,12 @@ def setup(self) -> None:
             name=self.table_name, df=self.data, column_types=self.config.column_types or {}
         )
         extra_table_data = [
-            TableData(name=name, df=df, column_types=self.config.extra_column_types.get(name, {}))
-            for name, df in self.extra_data.items()
+            TableData(
```
tyler-hoffman, line 75: I think the changes here do the right thing, but they're done in a kinda ugly way. I plan to circle back and do a slight refactor here after this PR merges.

tests/integration/data_sources_and_expectations/test_canonical_expectations.py
```diff
@@ -146,23 +147,35 @@ class TestExpectTableRowCountToEqualOtherTable:
...
         data=pd.DataFrame({"a": [1, 2, 3, 4]}),
-        extra_data={"test_table_a": pd.DataFrame({"col_b": ["a", "b", "c", "d"]})},
+        extra_data={"other_table": pd.DataFrame({"col_b": ["a", "b", "c", "d"]})},
     )
-    def test_success(self, batch_for_datasource):
-        expectation = gxe.ExpectTableRowCountToEqualOtherTable(other_table_name="test_table_a")
+    def test_success(
+        self,
+        batch_for_datasource: Batch,
+        extra_table_names_for_datasource: list[str],
```
tyler-hoffman, line 158: We discussed making this a mapping. I think that's definitely better, but plan to do it in a followup. I want to simplify a bit of the logic around how tables are set up, and that will make converting this to a map simpler.

joshua-stauffer (approved):
✅

PR #10632, opened 2024-11-06
[BUGFIX] Missing renderer params when `row_condition` is used

NathanFarmer:
- Fix regression where Expectation `kwargs` were not getting added as renderer `params`
- Revert logic changing "==" to "is" - this is a GX Cloud concern
---------

### DIFF
--- a/great_expectations/render/renderer_configuration.py
+++ b/great_expectations/render/renderer_configuration.py
@@ -433,9 +433,9 @@ class RendererConfiguration(pydantic_generics.GenericModel, Generic[RendererPara
         _params: Optional[Dict[str, Dict[str, Union[str, Dict[str, RendererValueType]]]]] = (
             values.get("_params")
         )
-        if not values["params"]:
-            values["params"] = _RendererValueBase()
-        if _params and _params != values["params"]:
+        # this logic loads the params that exist
+        # upon RendererConfiguration instantiation, e.g. row_condition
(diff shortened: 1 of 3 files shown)


great_expectations/render/renderer_configuration.py
```diff
@@ -479,7 +477,6 @@ def _parse_row_condition_str(row_condition_str: str) -> str:
             .replace(" or ", " OR ")
             .replace("~", " NOT ")
             .replace(" not ", " NOT ")
-            .replace("==", " is ")
```
NathanFarmer, line 482: Decided we should handle this logic in GX Cloud since there are other rendering substitutions like this that need to be done and are strictly related to GX Cloud-only configuration options.

great_expectations/render/renderer_configuration.py
```diff
@@ -461,7 +459,7 @@ def _get_row_conditions_list_from_row_condition_str(
         row_condition_str: str,
     ) -> List[str]:
         # divide the whole condition into smaller parts
-        row_conditions_list = re.split(r"AND|OR|NOT(?! in)|\(|\)", row_condition_str)
```
NathanFarmer, line 464: We were treating the keyword `col` as a variable when it is not. This change treats it as a delimiter, and then it is retained as part of the template string that doesn't get substituted.

ashmortar (approved):

PR #10643, opened 2024-11-08
[MAINTENANCE]: standardize _atomic_diagnostic_observed_value

ashmortar:
- Updates the Expectation class `_atomic_diagnostic_observed_value` method to use a template string and params like other atomic renderers
- Updates the ExpectationConfiguration class to better infer observed_value type
- Adds tests to assert proper type inference

### DIFF
--- a/great_expectations/render/renderer_configuration.py
+++ b/great_expectations/render/renderer_configuration.py
@@ -1,15 +1,15 @@
 from __future__ import annotations
 
 import re
-from datetime import datetime
+from datetime import date, datetime
 from enum import Enum
 from numbers import Number
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     Dict,
     Generic,
-    Iterable,
     List,
     Optional,
     Tuple,
@@ -19,7 +19,6 @@ from typing import (
 )
 
 import dateutil
-from dateutil.parser import ParserError
 from typing_extensions import TypeAlias, TypedDict
 
 from great_expectations.compatibility.pydantic import (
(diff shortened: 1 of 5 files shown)


great_expectations/render/renderer_configuration.py
```diff
@@ -155,6 +154,66 @@ class CodeBlock(TypedDict):
...
+        return False
+
+
+PARAM_TYPE_TESTERS: Dict[
+    RendererValueType,
+    Tuple[
+        Callable[[Any], bool],
+        Callable[[RendererValueType, Any], str],
+    ],
+] = {
```
NathanFarmer, line 192: My understanding is that since we dropped support for python 3.8, you should be able to use built-in types here (e.g. `dict` and `tuple`), but maybe I'm wrong about that. I do think this uses pydantic v1, which supports python 3.7.

tests/expectations/test_expectation_atomic_renderers.py
```diff
@@ -2202,9 +2203,107 @@ def test_atomic_diagnostic_observed_value_with_empty_result(get_diagnostic_rende
...
+        "type": "expect_table_row_count_to_equal",
+        "kwargs": {},
+    }
+    update_dict = {
+        "expectation_config": ExpectationConfiguration(**expectation_config),
+        "result": result,
+    }
+    rendered_content = get_diagnostic_rendered_content(update_dict)
+    res = rendered_content.to_json_dict()
+    pprint(res)
```
NathanFarmer, line 2300: Do we want this stdout? If the test fails we usually get a diff on the assertion line.

NathanFarmer (approved):

tests/expectations/test_expectation_atomic_renderers.py
ashmortar: ~ah, oops that was in there for debugging, I will remove it~

nevermind I was just following the pattern of the existing tests.  Should we remove all of them?

PR #10645, opened 2024-11-08
[BUGFIX] ExpectationSuiteValidationResult should contain checkpoint_name

TrangPham:
In V1, a checkpoint cannot be looked up by ID and ran, instead you need to use the name. 

ExpectationSuiteValidationResultMeta shows that checkpoint_name should be available but it is not. This PR rectifies that.
https://github.com/great-expectations/great_expectations/blob/d4b8c189121dc565c4aeb50c6c58f5ce4345baa3/great_expectations/core/expectation_validation_result.py#L399-L405

A separate discussion needs to be had if checkpoint_id should be used for looking up and running checkpoints instead of checkpoint_name.

Here are some related PRs for context:
- https://github.com/great-expectations/great_expectations/pull/10169

### DIFF
--- a/great_expectations/checkpoint/checkpoint.py
+++ b/great_expectations/checkpoint/checkpoint.py
@@ -356,6 +356,7 @@ class Checkpoint(BaseModel):
     ) -> CheckpointResult:
         for result in run_results.values():
             result.meta["checkpoint_id"] = self.id
+            result.meta["checkpoint_name"] = self.name
 
         return CheckpointResult(
             run_id=run_id,
--- a/great_expectations/core/validation_definition.py
+++ b/great_expectations/core/validation_definition.py
@@ -241,9 +241,10 @@ class ValidationDefinition(BaseModel):
         return batch_definition
 
     @public_api
-    def run(
+    def run(  # noqa: PLR0913
         self,
         *,
+        checkpoint_name: Optional[str] = None,
         checkpoint_id: Optional[str] = None,
         batch_parameters: Optional[BatchParameters] = None,
         expectation_parameters: Optional[SuiteParameterDict] = None,
@@ -289,6 +290,8 @@ class ValidationDefinition(BaseModel):
         results = validator.validate_expectation_suite(self.suite, expectation_parameters)
         results.meta["validation_id"] = self.id
         results.meta["checkpoint_id"] = checkpoint_id
+        if checkpoint_name:
+            results.meta["checkpoint_name"] = checkpoint_name
 
         # NOTE: We should promote this to a top-level field of the result.
         #       Meta should be reserved for user-defined information.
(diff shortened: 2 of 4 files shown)


great_expectations/checkpoint/checkpoint.py
```diff
@@ -356,6 +356,7 @@ def _construct_result(
     ) -> CheckpointResult:
         for result in run_results.values():
             result.meta["checkpoint_id"] = self.id
+            result.meta["checkpoint_name"] = self.name
```
TrangPham, line 359: Note that within another PR it was noted that perhaps the logic here should be removed. https://github.com/great-expectations/great_expectations/pull/10169 in favor of passing checkpoint_id down to the location which creates the result such the creation is all done in one place.

great_expectations/core/validation_definition.py
```diff
@@ -241,9 +241,10 @@ def _decode_data(cls, data_dict: dict) -> BatchDefinition:
         return batch_definition
 
     @public_api
-    def run(
+    def run(  # noqa: PLR0913
         self,
         *,
+        checkpoint_name: Optional[str] = None,
```
anthonyburdi, line 247: Do we need to do a minor version bump for changing the public api or is it ok to do in a patch release because it is optional?

anthonyburdi (approved):
LGTM - One non-blocking question.

great_expectations/core/validation_definition.py
TrangPham: @billdirks what do you think?

To me, this does not require it as we aren't change the ExpectationSuiteValidationResultMeta class. Checkpoint_name was already a field on it.

joshua-stauffer (changes requested):
I'm putting a block on this so timber can have a chance to review at our sync today. The implementation looks fine, but blocking because we explicitly decided to remove the `checkpoint_name` parameter from the `Checkpoint.run` signature and `meta`. I do think there may be other approaches to solve this problem, which might actually end up being better for users in the long run.

joshua-stauffer:
hey @TrangPham , i brought this to the team, and there's consensus that updating the public API of `ValidationResult.run` and putting `checkpoint_name` back into `meta` isn't in line with our longer term goals. As such, I'd recommend closing this PR. If you have questions, disagree, or want to discuss more, i'm happy to put some time on the calendar to sync 👍 

Per our conversations in [jira](https://greatexpectations.atlassian.net/browse/ZELDA-1086?focusedCommentId=21828) and [slack](https://greatexpectationslabs.slack.com/archives/C05V0M18TEJ/p1730811080213879?thread_ts=1730765983.400089&cid=C05V0M18TEJ), I'd recommend you get the checkpoint name at the graphql layer. Happy to pair if you decide to go with that approach and have questions!

PR #10647, opened 2024-11-08
[MAINTENANCE] Standardize result object from Checkpoint action runs

cdkini:
We should have a consistent API for `ValidationAction.run` - this new model is a lightweight wrapper around a status and the existing payload but should allow us to be better prepared for extensibility that users want.

### DIFF
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -151,7 +151,7 @@ class ValidationAction(BaseModel):
             data_docs_results = action_context.filter_results(class_=UpdateDataDocsAction)
             data_docs_pages = {}
             for result in data_docs_results:
-                data_docs_pages.update(result)
+                data_docs_pages.update(result.run_info)
             return data_docs_pages
 
         return None
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -86,7 +86,7 @@ def _build_renderer(config: dict) -> Renderer:
 
 
 class ValidationActionResult(BaseModel):
-    success: bool | None
+    success: bool | None # None if not run at all but could be an enum
     run_info: dict
 
 
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -319,7 +319,9 @@ class SlackNotificationAction(DataDocsAction):
             if response is not None
             else ValidationActionRunStatus.FAILURE
         )
-        return ValidationActionResult(status=status, result={"slack_notification_result": response})
+        return ValidationActionResult(
+            status=status, run_info={"slack_notification_result": response}
+        )
 
     def _render_validation_result(
         self,
@@ -523,13 +525,13 @@ class MicrosoftTeamsNotificationAction(ValidationAction):
             )
         )
 
-        success = (
+        status = (
             ValidationActionRunStatus.SUCCESS
             if teams_notif_result is not None
             else ValidationActionRunStatus.FAILURE
         )
-        return ValidationActionRunStatus(
-            success=success, run_info={"microsoft_teams_notification_result": teams_notif_result}
+        return ValidationActionResult(
+            status=status, run_info={"microsoft_teams_notification_result": teams_notif_result}
         )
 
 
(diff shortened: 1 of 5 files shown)


great_expectations/checkpoint/actions.py
```diff
@@ -85,23 +86,42 @@ def _build_renderer(config: dict) -> Renderer:
...
+class ValidationActionResult(BaseModel):
+    """
+    The result of running a ValidationAction.
+
+    Args:
+        status: The status of the action run. If the action was not run, the status is "not_run".
+        run_info: A dictionary containing information about the run.
+    """
+
+    status: ValidationActionRunStatus
```
billdirks, line 104: Should we include the action name (eg the class name) in this result object or will it be obvious since this is embedded inside something? Eg if I see `not_run` will I know that is associated with a pager duty action?

great_expectations/checkpoint/actions.py
```diff
@@ -85,23 +86,42 @@ def _build_renderer(config: dict) -> Renderer:
...
+    run_info: dict
```
billdirks, line 105: Does the run_info vary from action to action or can we be more concrete about this?

great_expectations/checkpoint/actions.py
```diff
@@ -409,9 +438,15 @@ def _run_pypd_alert(self, dedup_key: str, message: str, success: bool):
...
-            return {"pagerduty_alert_result": "success"}
+            return ValidationActionResult(
+                status=ValidationActionRunStatus.SUCCESS,
+                run_info={"pagerduty_alert_result": "success"},
+            )
 
-        return {"pagerduty_alert_result": "none sent"}
+        return ValidationActionResult(
+            status=ValidationActionRunStatus.NOT_RUN,
+            run_info={"pagerduty_alert_result": "none sent"},
```
billdirks, line 448: For ms teams, the equivalent value is `None` instead of `"none sent"`. Can these be the same thing?

great_expectations/checkpoint/actions.py
billdirks: It seems like the run_info dict has a key which is a string like `<action_type>_result`. If we put the action type in it's own field, it would make that info easier to parse out.

tests/actions/test_core_actions.py
```diff
@@ -427,6 +432,9 @@ def test_run_integration_success(
         self,
         checkpoint_result: CheckpointResult,
     ):
+        if not os.environ.get("GX_MS_TEAMS_WEBHOOK"):
```
billdirks, line 435: I don't like programmatically skipping tests. What has happened in the past is someone makes a change and we start skipping this in CI without realizing it. I'd rather just fail and require this env variable or have this moved to a different marker where we require this.

billdirks (commented):
Thanks for standardizing this more. I left a few comments inline.

great_expectations/checkpoint/actions.py
cdkini: It'll be obvious since this is only ever used in `Checkpoint.run` - my thinking is we can take the aggregated results (of these actions) and either log or append to the checkpoint result object.

Don't want to make too many changes here until we can plan better.

great_expectations/checkpoint/actions.py
cdkini: Ah yeah I don't love our current dicts but I'm trying to minimize changes. They're all pretty much `{"action_type_in_snake_case": "success_or_failure"}` so they could be simplified with a simple bool but I think I want to keep it this way for a few reasons:
1. Minimize changes (for now)
2. Allow for more complex `run_info` (`UpdateDataDocs` leverages this and future custom actions might want to include more details?)

great_expectations/checkpoint/actions.py
cdkini: Same note as above - don't love these dicts but doing what I can to minimize changes. If we can determine a consistent standard, I'd be happy to make changes across the board.

great_expectations/checkpoint/checkpoint.py
```diff
@@ -297,7 +298,10 @@ def run(
         )
 
         checkpoint_result = self._construct_result(run_id=run_id, run_results=run_results)
-        self._run_actions(checkpoint_result=checkpoint_result)
+
+        action_results = self._run_actions(checkpoint_result=checkpoint_result)
+        # TODO(cdkini): Figure out how to handle action results
+        print(action_results)  # We could add these to our checkpoint result or logger.warning?
```
cdkini, line 304: This is where I'm unsure but the goal with this whole effort is to make action results more transparent - we currently silently fail or skip

cdkini: If we log, there's still a chance can lose track of it.
I'd lean towards updating the checkpoint result but we need to be careful about what we add there - don't want it becoming as cumbersome as its V0 counterpart.

PR #10658, opened 2024-11-12
[MAINTENANCE] Use random schema in integration tests

tyler-hoffman:
* replaces the abstract `schema` with an abstract bool `use_schema`
* implements a new `schema` at the parent class to randomize the schema name
* If `use_schema` is `True`, creates a random schema
NOTE: This allows us to basically revert a lot of https://github.com/great-expectations/great_expectations/pull/10630. I'll get that in a follow-up PR.

### DIFF
--- a/tests/integration/test_utils/data_source_config/postgres.py
+++ b/tests/integration/test_utils/data_source_config/postgres.py
@@ -1,4 +1,4 @@
-from typing import Mapping, Union
+from typing import Mapping
 
 import pandas as pd
 import pytest
@@ -43,10 +43,10 @@ class PostgresBatchTestSetup(SQLBatchTestSetup[PostgreSQLDatasourceTestConfig]):
     def connection_string(self) -> str:
         return "postgresql+psycopg2://postgres@localhost:5432/test_ci"
 
-    @override
     @property
-    def schema(self) -> Union[str, None]:
-        return "public"
+    @override
+    def use_schema(self) -> bool:
+        return False
 
     @override
     def make_batch(self) -> Batch:
--- a/tests/integration/test_utils/data_source_config/snowflake.py
+++ b/tests/integration/test_utils/data_source_config/snowflake.py
@@ -48,15 +48,13 @@ class SnowflakeConnectionConfig(BaseSettings):
     SNOWFLAKE_PW: str
     SNOWFLAKE_ACCOUNT: str
     SNOWFLAKE_DATABASE: str
-    SNOWFLAKE_SCHEMA: str
     SNOWFLAKE_WAREHOUSE: str
     SNOWFLAKE_ROLE: str = "PUBLIC"
 
-    @property
-    def connection_string(self) -> str:
+    def connection_string(self, schema: str) -> str:
         return (
             f"snowflake://{self.SNOWFLAKE_USER}:{self.SNOWFLAKE_PW}"
-            f"@{self.SNOWFLAKE_ACCOUNT}/{self.SNOWFLAKE_DATABASE}/{self.SNOWFLAKE_SCHEMA}"
+            f"@{self.SNOWFLAKE_ACCOUNT}/{self.SNOWFLAKE_DATABASE}/{schema}"
             f"?warehouse={self.SNOWFLAKE_WAREHOUSE}&role={self.SNOWFLAKE_ROLE}"
         )
 
@@ -65,7 +63,9 @@ class SnowflakeBatchTestSetup(SQLBatchTestSetup[SnowflakeDatasourceTestConfig]):
     @property
     @override
     def connection_string(self) -> str:
-        return self.snowflake_connection_config.connection_string
+        schema = self.schema
+        assert schema
+        return self.snowflake_connection_config.connection_string(schema)
 
     @property
     @override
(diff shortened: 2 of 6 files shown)


tests/integration/test_utils/data_source_config/postgres.py
```diff
@@ -43,10 +43,10 @@ class PostgresBatchTestSetup(SQLBatchTestSetup[PostgreSQLDatasourceTestConfig]):
     def connection_string(self) -> str:
         return "postgresql+psycopg2://postgres@localhost:5432/test_ci"
 
-    @override
     @property
-    def schema(self) -> Union[str, None]:
-        return "public"
```
tyler-hoffman, line 49: Looks like we didn't need the schema on this one. (also I switched the decorator ordering while I was here because pyright was upset about it. I didn't look into it, but usually it's just a bit more sensitive than our mypy settings).

tests/integration/test_utils/data_source_config/snowflake.py
```diff
@@ -48,29 +48,29 @@ class SnowflakeConnectionConfig(BaseSettings):
     SNOWFLAKE_PW: str
     SNOWFLAKE_ACCOUNT: str
     SNOWFLAKE_DATABASE: str
-    SNOWFLAKE_SCHEMA: str
     SNOWFLAKE_WAREHOUSE: str
     SNOWFLAKE_ROLE: str = "PUBLIC"
 
-    @property
-    def connection_string(self) -> str:
+    def connection_string(self, schema: str) -> str:
```
tyler-hoffman: This is dynamic now

tests/integration/test_utils/data_source_config/snowflake.py
```diff
@@ -48,29 +48,29 @@ class SnowflakeConnectionConfig(BaseSettings):
...
 
 
 class SnowflakeBatchTestSetup(SQLBatchTestSetup[SnowflakeDatasourceTestConfig]):
-    @override
     @property
+    @override
     def connection_string(self) -> str:
-        return self.snowflake_connection_config.connection_string
+        schema = self.schema
+        assert schema
```
tyler-hoffman: These checks are kinda obnoxious, and we can solve them with generics, but I want to kick the can on that.

tests/integration/test_utils/data_source_config/snowflake.py
```diff
@@ -48,29 +48,30 @@ class SnowflakeConnectionConfig(BaseSettings):
...
-    SNOWFLAKE_SCHEMA: str
     SNOWFLAKE_WAREHOUSE: str
     SNOWFLAKE_ROLE: str = "PUBLIC"
 
     @property
     def connection_string(self) -> str:
         return (
             f"snowflake://{self.SNOWFLAKE_USER}:{self.SNOWFLAKE_PW}"
-            f"@{self.SNOWFLAKE_ACCOUNT}/{self.SNOWFLAKE_DATABASE}/{self.SNOWFLAKE_SCHEMA}"
+            f"@{self.SNOWFLAKE_ACCOUNT}/{self.SNOWFLAKE_DATABASE}"
```
billdirks, line 60: I might add a comment about how this is used and why it's ok to not have the connection string here. Namely, [we require it when configuring a snowflake datasource](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/datasource/fluent/snowflake_datasource.py#L106) so I had to read more to understand why we can not have it here.

tests/integration/test_utils/data_source_config/snowflake.py
```diff
@@ -48,29 +48,30 @@ class SnowflakeConnectionConfig(BaseSettings):
...
         )
 
 
 class SnowflakeBatchTestSetup(SQLBatchTestSetup[SnowflakeDatasourceTestConfig]):
-    @override
     @property
+    @override
     def connection_string(self) -> str:
+        schema = self.schema
+        assert schema
```
billdirks: Why do you need to assert the schema here? It doesn't look like it is returned in the connection string on the line below.

tests/integration/test_utils/data_source_config/sql.py
```diff
@@ -113,6 +130,11 @@ def setup(self) -> None:
     def teardown(self) -> None:
         for table in self.tables:
             table.drop(self.engine)
+        with self.engine.connect() as conn, conn.begin():
```
billdirks: pytest question: Are we guaranteed teardown will get called, even if the test errors out? I want to make sure we tear down the schema.

billdirks (commented):
Thanks for doing this! I have some questions around the snowflake schema.

tests/integration/test_utils/data_source_config/snowflake.py
tyler-hoffman: Good call. Updated.

tests/integration/test_utils/data_source_config/snowflake.py
tyler-hoffman: ah thanks. That's leftover. Removing now!

tests/integration/test_utils/data_source_config/sql.py
tyler-hoffman: As long as the test errors out, rather than something in the fixtures. There are ways we can pollute things, but that's been the case for a while with other tests as well. I think we should be covered for most cases. I'll share out some more details in slack.

billdirks (approved):
Thank you!

PR #10662, opened 2024-11-13
[MAINTENANCE] Improve experience around expectation deletion with Cloud-backed suites

cdkini:
A few relevant changes here:
- Expectation equality checks should be done with the actual object (and not `.configuration`)
  - This is due to the fact that `__eq__` has specific logic around rendered content and other fields that could go awry
  - I've decided that both `notes` and `meta` should be excluded from `__eq__`; they are too volatile due to misc updates from Cloud and represent "metadata" as opposed to actual state in my opinion
- We should ignore `id` when determining uniqueness with `suite.add_expectation()` (users can very easily add duplicates without realizing)
```python
for _ in range(10):
    e = gxe.ExpectColumnValuesToBeBetween(...)
    suite.add_expectation(e)
print(len(suite.expectations)) # 1 if this `id` change but 10 otherwise - we want the former
```

- Update docstring to point to using suite indexing when deleting

---


great_expectations/core/expectation_suite.py
```diff
@@ -145,19 +145,13 @@ def add_expectation(self, expectation: _TExpectation) -> _TExpectation:
             )
         should_save_expectation = self._has_been_saved()
         expectation_is_unique = all(
-            expectation.configuration != existing_expectation.configuration
-            for existing_expectation in self.expectations
+            expectation != existing_expectation for existing_expectation in self.expectations
```
cdkini: Equality checks should utilize `Expectation.__eq__` rather than `configuration`

great_expectations/core/expectation_suite.py
```diff
@@ -145,19 +145,13 @@ def add_expectation(self, expectation: _TExpectation) -> _TExpectation:
...
         if expectation_is_unique:
             # suite is a set-like collection, so don't add if it not unique
-            self.expectations.append(expectation)
             if should_save_expectation:
-                try:
-                    expectation = self._store.add_expectation(suite=self, expectation=expectation)
-                    self.expectations[-1].id = expectation.id
-                except Exception as exc:
-                    self.expectations.pop()
-                    raise exc  # noqa: TRY201
```
cdkini, line 160: We can just wait to append until we succeed with persistence - also, the `id` is already attached to the input expectation

great_expectations/expectations/expectation.py
```diff
@@ -368,7 +368,7 @@ def __eq__(self, other: object) -> bool:
 
         # rendered_content is derived from the rest of the expectation, and can/should
         # be excluded from equality checks
-        exclude: set[str] = {"rendered_content"}
+        exclude: set[str] = {"rendered_content", "id"}
```
cdkini: Legacy equality checks (i.e. using `configuration`) exclude `id`

tests/expectations/test_expectation.py
```diff
@@ -362,67 +362,103 @@ def test_expectation_with_duplicate_suite_parameters(self):
...
+            pytest.param(None, None, id="both_none"),
+            pytest.param(None, "aaae648e-0a43-431b-81a0-04e68f1473ae", id="missing_id"),
+            pytest.param(
+                "bbbe648e-0a43-431b-81a0-04e68f1473ae",
+                "aaae648e-0a43-431b-81a0-04e68f1473ae",
+                id="different_ids",
+            ),
+        ],
     )
+    def test_expectation_equality_with_id(self, id_a: str | None, id_b: str | None):
```
cdkini: This is the new test

great_expectations/core/expectation_suite.py
```diff
@@ -145,19 +145,13 @@ def add_expectation(self, expectation: _TExpectation) -> _TExpectation:
...
+                expectation = self._store.add_expectation(suite=self, expectation=expectation)
```
tyler-hoffman, line 154: well this seems much more clear!

great_expectations/core/expectation_suite.py
```diff
@@ -218,13 +212,15 @@ def _process_expectation(
     @public_api
     def delete_expectation(self, expectation: Expectation) -> Expectation:
         """Delete an Expectation from the collection.
+        The input Expectation must be in the suite and referenced by index.
```
tyler-hoffman: I'm going to throw a blocker on here. The second part of this seems wrong. How could it matter how you retrieve the expectation?

great_expectations/expectations/expectation.py
tyler-hoffman: Do we want to maintain that? Does still comparing the id break anything? Would it be a bug if the ids don't match? If we need this logic, maybe add it to the comment above about rendered_content?

tyler-hoffman (changes requested):
Left a blocker on 1.5 smallish things - LMK if I'm just confused

great_expectations/core/expectation_suite.py
cdkini: Hmm this is fair - I want to encourage users to delete by using `suite.expectations[idx]` instead of `gxe.ExpectWhateverWhatever`. I think the example here is good enough.

tyler-hoffman: Yeah, the example is fine with me. FWIW I'd personally probs find it by like `next(e for e in suite.expectations where isinstance(e, MyExpectation))`

great_expectations/expectations/expectation.py
cdkini: I don't love it but if we look at the script I mentioned in Slack, we get different behavior based on this:

```python
suite = context.suites.add(gx.ExpectationSuite("chetan-test-2024-11-13-v21"))

for i in range(10):
    e = gxe.ExpectColumnValuesToBeBetween(column="age", min_value=0, max_value=100)
    suite.add_expectation(e)

print(len(suite.expectations)) # How many should we see?
```

We historically have gotten 1 from this but making `id` part of the `__eq__` check gets us 10.

tyler-hoffman (approved):
I have mixed feelings about equality being true when both ids exist and are different, but I think we can table that discussion and merge this. Thanks for fixing this!

great_expectations/core/expectation_suite.py
```diff
@@ -144,27 +145,27 @@ def add_expectation(self, expectation: _TExpectation) -> _TExpectation:
...
         expectation.register_save_callback(save_callback=self._save_expectation)
 
         self._submit_expectation_created_event(expectation=expectation)
 
         return expectation
 
+    def _determine_if_expectation_is_unique(self, expectation: Expectation) -> bool:
+        # Expectation is deemed unique if it is not already in the suite
+        # We do not consider the id of the expectation in this check
+        expectation_copies = copy.deepcopy(self.expectations)
```
tyler-hoffman: I'm guessing we also want to account for `rendered_content`, right?

Also does this get expensive when we have rendered content? If we want to save ourselves the deep copy (and mutation), you could do something like `model1.dict(exclude={"id"}) == model2.dict(exclude={"id"})` in a loop, or that's what chatgpt tells me at least 😄

tests/expectations/test_expectation.py
```diff
@@ -398,31 +398,29 @@ def test_expectation_equality_with_notes(
...
-        pytest.param({}, None, True, id="both_falsy"),
-        pytest.param({"author": "Bob Dylan"}, None, False, id="missing_meta"),
-        pytest.param(
-            {"author": "Bob Dylan"}, {"author": "John Lennon"}, False, id="different_meta"
-        ),
-        pytest.param({"author": "Bob Dylan"}, {"author": "Bob Dylan"}, True, id="equivalent_meta"),
+        pytest.param(None, None, id="both_none"),
+        pytest.param({}, None, id="both_falsy"),
+        pytest.param({"author": "Bob Dylan"}, None, id="missing_meta"),
+        pytest.param({"author": "Bob Dylan"}, {"author": "John Lennon"}, id="different_meta"),
```
tyler-hoffman: Potentially a huge blocker for me: did these guys actually write books?

great_expectations/expectations/expectation.py
```diff
@@ -368,18 +368,12 @@ def __eq__(self, other: object) -> bool:
 
         # rendered_content is derived from the rest of the expectation, and can/should
         # be excluded from equality checks
-        exclude: set[str] = {"rendered_content"}
+        # notes and meta are simple metadata and should be excluded from equality checks
+        exclude: set[str] = {"rendered_content", "notes", "meta"}
```
tyler-hoffman: Should we really exclude these for equality? Why not just handle these kind of weird situations in the helper method above?

great_expectations/core/expectation_suite.py
```diff
@@ -219,12 +220,13 @@ def _process_expectation(
...
+        Example:
+            >>> suite.delete_expectation(suite.expectations[0])
+
         Raises:
             KeyError: Expectation not found in suite.
         """
-        remaining_expectations = [
-            exp for exp in self.expectations if exp.configuration != expectation.configuration
-        ]
+        remaining_expectations = [exp for exp in self.expectations if exp != expectation]
```
tyler-hoffman: Should this be using the same kind of logic as in `_determine_if_expectation_is_unique` wrt ignoring ids?

tyler-hoffman (commented):
Just thinking aloud: I wonder if some of this gets easier if we implement something like an `is_equivalent` or `is_equalish` on expectations? Then we can let the `__eq__` method have fewer exceptions, and the logic will be in a somewhat findable/reusable place?

great_expectations/core/expectation_suite.py
```diff
@@ -144,27 +144,42 @@ def add_expectation(self, expectation: _TExpectation) -> _TExpectation:
...
+        Helper method to determine if an expectation is already in the suite.
+
+        Note that this check is less stringent than Expectation.__eq__ and excludes
+        a few fields that are not relevant for uniqueness in the suite.
+
+        Args:
+            expectation: The expectation to check for.
+
+        Returns:
+            A tuple of a boolean indicating whether the expectation is already in the suite
```
tyler-hoffman: Non-blocking: the method return type feels like a bit of a smell, and designed for a very specific use case. Would it be simpler if we took in 2 expectations, and just returned a bool, then let `delete_expectation` be responsible for looping?

Don't feel like you need to over-index on this - we've already had a ton of back and forth, and I appreciate your patience!!

tyler-hoffman (approved):

PR #10663, opened 2024-11-13
[MAINTENANCE] Reduce test duration of flakey test

tyler-hoffman:
This test fails with some regularity in CI, but not in the way the test expects. See [this example](https://github.com/great-expectations/great_expectations/actions/runs/11824674689/job/32946716375). It fails with `Failed: Timeout >2.0s` because we currently invoke pytest with `--timeout=2`, so we would would never fail the assertion as written.

We could also throw a `mark.pytest.slow` on this instead of cutting the time in half, but I'm not sure if that matters.

The fix here:
* Reduce the expected time and `large_number` by half so we can see the expected error if appropriate. I'm assuming complexity is O(n) with `large_number`.
* Then reduce `large_number` by another 25% to avoid the flakiness.

### DIFF
--- a/tests/expectations/core/test_expect_column_values_to_be_of_type.py
+++ b/tests/expectations/core/test_expect_column_values_to_be_of_type.py
@@ -109,11 +109,11 @@ def test_expect_column_values_to_be_in_set_render_performance():
     This test prevents a historical bug in which rendering took ~10 seconds to render 400 items.
     """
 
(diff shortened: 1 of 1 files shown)


dctalbot (approved):
If this test continues to be a problem, we can just remove it. It's shoehorning a benchmark into a test, and really we could use a more holistic performance assessment setup in CI.

PR #10667, opened 2024-11-14
[MAINTENANCE] Condition parser typing and row conditions testing

NathanFarmer:
- Update `condition_parser` api to not raise type errors when using a string where the `ConditionParser` enum used to be expected
- Update tests due to different error message
- Add `row_condition` tests with new Expectation testing framework
-------------


great_expectations/expectations/core/schemas/ExpectColumnDistinctValuesToBeInSet.json
```diff
@@ -280,17 +285,6 @@
...
             "additionalProperties": false
-        },
-        "ConditionParser": {
-            "title": "ConditionParser",
-            "description": "Type of parser to be used to interpret a Row Condition.",
-            "enum": [
-                "great_expectations",
-                "great_expectations__experimental__",
-                "pandas",
-                "spark"
```
NathanFarmer, line 291: The strings `great_expectations__experimental__` and `spark` will continue to work at runtime, but they will raise type errors with static analysis tools.

great_expectations/expectations/model_field_types.py
```diff
@@ -69,10 +68,6 @@
...
-class ConditionParser(str, Enum):
-    """Type of parser to be used to interpret a Row Condition."""
-
-    GX = "great_expectations"
-    GX_DEPRECATED = "great_expectations__experimental__"
-    PANDAS = "pandas"
-    SPARK = "spark"
+ConditionParser = Literal[
+    "great_expectations", "great_expectations__experimental__", "pandas", "spark"
+]
```
NathanFarmer: Without this change we are seeing type errors at the API-level when a user passes a string instead of an enum (e.g. `pandas` insetead of `ConditionParser.PANDAS`)

tests/integration/data_sources_and_expectations/test_expectation_conditions.py
```diff
@@ -0,0 +1,112 @@
...
+            id="number - lt",
+        ),
+        pytest.param(
+            "quantity==1",
+            id="number - eq",
+        ),
+    ],
+)
+def test_expect_column_min_to_be_between__pandas_row_condition(
+    batch_for_datasource, row_condition
```
tyler-hoffman: Non-blocking, and I should probably write something up and get buy in, but I think we should add types for test parameters.

tests/integration/data_sources_and_expectations/test_expectation_conditions.py
```diff
@@ -0,0 +1,112 @@
...
+        row_condition=row_condition,
+        condition_parser="pandas",
+    )
+    result = batch_for_datasource.validate(expectation)
+    assert result.success
+
+
+@parameterize_batch_for_data_sources(
+    data_source_configs=[
+        SparkFilesystemCsvDatasourceTestConfig(),
```
tyler-hoffman, line 74: I just merged a change so that you don't have to specify column types for dates, so you can leave these constructors empty now.

great_expectations/expectations/model_field_types.py
dctalbot: Why can't the API handle both types? Is this a breaking change?

great_expectations/execution_engine/pandas_execution_engine.py
```diff
@@ -526,7 +525,7 @@ def get_domain_records(  # noqa: C901, PLR0912
         if row_condition:
             condition_parser = domain_kwargs.get("condition_parser", None)
 
-            if condition_parser == ConditionParser.PANDAS:
+            if condition_parser == "pandas":
```
dctalbot: We should probably still import this as a string literal from `model_field_types.py`. It's more maintainable in the event that you need to refactor e.g. you can trace the imports rather than grep code for these hard-coded strings. `"pandas"` is also quite generic, so finding all instances of it that are in the context of row conditions would be laborious.

great_expectations/expectations/model_field_types.py
NathanFarmer: This isn't a breaking change. The user has always been expected to pass a string in the API. The problem before was if you passed a string and ran a type checker you would have to add a type ignore on the line because you passed a string instead of the enum (it would still work at runtime). We don't want to encourage users to use the enum at the API-level, but that's what the type checker would tell you to do.

great_expectations/execution_engine/pandas_execution_engine.py
NathanFarmer: I agree that it's more maintainable, but I'm not sure how that would work. I guess we could retain the enum and only use it internally. Then we would need to keep the enum and the string literal type in sync.

NathanFarmer: Or is there some way to reference a string literal type member that I'm not aware of?

dctalbot: completely untested, but something like this?

```py
CONDITION_PARSER_GX: Literal["great_expectations"] = "great_expectations"
CONDITION_PARSER_GX_EXPERIMENTAL: Literal["great_expectations__experimental__"] = "great_expectations__experimental__"
CONDITION_PARSER_PANDAS: Literal["pandas"] = "pandas"
CONDITION_PARSER_SPARK: Literal["spark"] = "spark"

ConditionParser = Union[
    CONDITION_PARSER_GX, 
    CONDITION_PARSER_GX_EXPERIMENTAL,
    CONDITION_PARSER_PANDAS,
    CONDITION_PARSER_SPARK
]
```

dctalbot (approved):

great_expectations/execution_engine/pandas_execution_engine.py
NathanFarmer: Your version doesn't work, but I figured it out:
```python
from typing import Literal, get_args

ConditionParser = Literal[
    "great_expectations", "great_expectations__experimental__", "pandas", "spark"
]

(
    CONDITION_PARSER_GREAT_EXPECTATIONS,
    CONDITION_PARSER_GREAT_EXPECTATIONS_DEPRECATED,
    CONDITION_PARSER_PANDAS,
    CONDITION_PARSER_SPARK,
) = get_args(ConditionParser)
```

PR #10673, opened 2024-11-15
[MAINTENANCE] Fixture to get assets for test framework

tyler-hoffman:
The functional change here is that we have a new fixture to get assets for the test framework.

Doing this required exposing an asset property on the `BatchTestSetup` object, so I had to touch all the existing subclasses of it.

This will unblock the work for CORE-254, but I think it's also a good example of how we can expand the framework. I can imagine us doing something similar for the datasource, but for now that seems like YAGNI


tests/integration/test_utils/data_source_config/base.py
```diff
@@ -84,15 +84,27 @@ def __hash__(self) -> int:
...
-class BatchTestSetup(ABC, Generic[_ConfigT]):
+class BatchTestSetup(ABC, Generic[_ConfigT, _AssetT]):
     """ABC for classes that set up and tear down batches."""
 
     def __init__(self, config: _ConfigT, data: pd.DataFrame) -> None:
         self.config = config
         self.data = data
 
+    @cached_property
+    def name(self) -> str:
```
tyler-hoffman: Before we created one in the `make_batch` methods, but I moved it to a cached property for reuse. It doesn't really matter that we use the same name across datasource/asset/batch def, but I figure why not?

tests/integration/test_utils/data_source_config/databricks.py
```diff
@@ -68,21 +68,16 @@ def inferrable_types_lookup(self) -> dict[type, TypeEngine]:
     def _databrics_connection_config(self) -> DatabricksConnectionConfig:
         return DatabricksConnectionConfig()  # type: ignore[call-arg]  # retrieves env vars
 
+    @cached_property
```
tyler-hoffman, line 71: Note to reviewer: `make_batch` is now defined in the parent class in `sql.py`

tests/integration/conftest.py
```diff
@@ -164,6 +164,17 @@ def batch_for_datasource(
...
 
 
+@pytest.fixture
+def asset_for_datasource(
+    _batch_setup_for_datasource: BatchTestSetup,
+) -> Generator[DataAsset, None, None]:
+    """Fixture that yields an asset for a specific data source type.
+    This must be used in conjunction with `indirect=True` to defer execution
+    """
+    set_context(_batch_setup_for_datasource.context)
```
cdkini, line 174: Why is this necessary?

tests/integration/test_utils/data_source_config/databricks.py
cdkini, line 71: Does this being a cached property have significant performance implications? Is there any risk around caching here?

tests/integration/test_utils/data_source_config/databricks.py
```diff
@@ -68,21 +68,16 @@ def inferrable_types_lookup(self) -> dict[type, TypeEngine]:
...
-            .add_batch_definition_whole_table(name=name)
-            .get_batch()
+    def asset(self) -> TableAsset:
+        return self.context.data_sources.add_databricks_sql(
+            name=self._random_resource_name(),
+            connection_string=self.connection_string,
+        ).add_table_asset(
+            name=self._random_resource_name(),
+            table_name=self.table_name,
+            schema_name=self.schema,
```
cdkini, line 80: Should we use the same random name for everything in the hierarchy? Could make debugging easier?

cdkini (approved):

tests/integration/conftest.py
tyler-hoffman: Because we cache the setup objects and may reuse them across tests, we might have created/set a different context since this context was created/set.

tests/integration/test_utils/data_source_config/databricks.py
tyler-hoffman: I originally thought we would want to use the same asset if we access it multiple times. I'm not concerned about performance here at all TBH. I'm not totally sure we need it to be consistent across calls though. Maybe we end up making it just a `make_asset` method. I want to kick the can on that change to see what our needs are, but happy to do it now if you have opinions.

tests/integration/test_utils/data_source_config/databricks.py
tyler-hoffman: I wanted to originally, but it ended causing issues if/when we called `make_batch` multiple times on the same asset. And I couldn't really figure out a case where that would help with debugging. But like above, happy to revisit

PR #10697, opened 2024-11-21
[MAINTENANCE] Add atomic diagnostic renderer to `ExpectColumnMostCommonValueToBeInSet`

NathanFarmer:
- Observed value renderer for `Expect column most common value to be in set`
- Adds new a `render_state` property to `RenderedContent` params for rendering set item states


reqs/requirements-dev-snowflake.txt
```diff
@@ -1,4 +1,4 @@
 pandas<2.2.0; python_version >= "3.9"
-snowflake-connector-python>=2.5.0; python_version < "3.11"
+snowflake-connector-python>=2.5.0,<3.12.3; python_version < "3.11"
```
NathanFarmer, line 2: CI is currently broken due to static-analysis failing on a new version of snowflake connector on python 3.9

reqs/requirements-dev-snowflake.txt
```diff
@@ -1,4 +1,5 @@
 pandas<2.2.0; python_version >= "3.9"
 snowflake-connector-python>=2.5.0; python_version < "3.11"
 snowflake-connector-python>2.9.0; python_version >= "3.11" # earlier versions fail to build on 3.11
-snowflake-sqlalchemy>=1.2.3,!=1.5.2 # https://github.com/snowflakedb/snowflake-sqlalchemy/issues/485
+snowflake-sqlalchemy>=1.2.3,!=1.5.2,<1.7.0; python_version < "3.10" # https://github.com/snowflakedb/snowflake-sqlalchemy/issues/485
+snowflake-sqlalchemy>=1.2.3,!=1.5.2; python_version >= "3.10" # https://github.com/snowflakedb/snowflake-sqlalchemy/issues/485
```
NathanFarmer, line 5: CI is currently broken due to static-analysis failing on a new version of snowflake-sqlalchemy on python 3.9

great_expectations/expectations/core/expect_column_most_common_value_to_be_in_set.py
```diff
@@ -329,6 +336,66 @@ def _prescriptive_renderer(
...
+            param_prefix=ov_param_prefix,
+            renderer_configuration=renderer_configuration,
+        )
+
+        params = renderer_configuration.params
+        for _, param in params:
+            if param not in (
+                params.value_set,
+                params.observed_value,
+            ):
```
dctalbot: Is this meant to be equivalent to `if _.startswith(ov_param_prefix)` ? I think filtering what to include may be more robust than filtering what to exclude

great_expectations/render/renderer_configuration.py
```diff
@@ -274,27 +275,30 @@ class _RendererParamBase(_RendererValueBase):
...
             """
-            param_type: RendererValueType = values["schema"]["type"]
-            value: Any = values["value"]
-
-            (tester, error) = PARAM_TYPE_TESTERS[param_type]
-            if not tester(value):
-                raise RendererConfigurationError(error(param_type, value))
+            # if render_state is in values, this root_validator is being called
+            # on assignment rather than during instantiation
+            if not values.get("render_state"):
```
dctalbot, line 295: This looks potentially brittle? what happens if we exclude this check?

dctalbot (approved):

great_expectations/expectations/core/expect_column_most_common_value_to_be_in_set.py
NathanFarmer: Yeah I was annoyed I didn't have a use for the first value in that tuple and this is the perfect use. Much better.

great_expectations/render/renderer_configuration.py
NathanFarmer: It's a very annoying pydanic v1 issue where the `root_validator` is run on assignment. Because we have the aliased attribute `renderer_schema` -> `schema` it gets passed in as `schema` upon instantiation and then internally it becomes `renderer_schema`. Then this `root_validator` runs again when we assign to `render_state` and the value in this dict is now `renderer_schema` and there is a `KeyError` that `schema` cannot be found. The alternative would be to check for `schema` or `renderer_schema`, but I felt like it shouldn't run any of this logic again regardless. I think the issue is fixed in pydantic v2 and the analog to `root_validator` isn't called after instantiation.

PR #10743, opened 2024-12-05
[FEATURE] Enable custom actions in V1

cdkini:
Dynamic registry and metaclass will allow users to define their own actions and have them successfully serialize/deserialize without issue.

### DIFF
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -135,7 +135,7 @@ class ValidationActionRegistry:
 
         self._registered_actions[action_type] = action_class
 
-    def get(self, action_type: str) -> Type[ValidationAction]:
+    def get(self, action_type: str | None) -> Type[ValidationAction]:
         """
         Return a ValidationAction class based on its type.
         Used when instantiating actions from a checkpoint configuration.
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -134,13 +134,13 @@ class ValidationActionRegistry:
 
         self._registered_actions[action_type] = action_class
 
-    def get_from_config(self, action_config: dict) -> Type[ValidationAction]:
+    def get(self, action_type: str) -> Type[ValidationAction]:
         """
-        Return a ValidationAction class based on its configuration.
+        Return a ValidationAction class based on its type.
         Used when instantiating actions from a checkpoint configuration.
 
         Args:
-            action_config: The configuration of the action to instantiate.
+            action_type: The 'type' key from the action configuration.
 
         Returns:
             The ValidationAction class corresponding to the configuration.
@@ -148,7 +148,6 @@ class ValidationActionRegistry:
         Raises:
             ValidationActionRegistryRetrievalError: If the action type is not registered.
         """
-        action_type = action_config.get("type")
         if action_type not in self._registered_actions:
             raise ValidationActionRegistryRetrievalError(action_type)
 
@@ -158,7 +157,7 @@ class ValidationActionRegistry:
 _VALIDATION_ACTION_REGISTRY = ValidationActionRegistry()
 
 
-@dataclass_transform(kw_only_default=True, field_specifiers=(Field,))
+@dataclass_transform(kw_only_default=True, field_specifiers=(Field,))  # Enables type hinting
 class MetaValidationAction(ModelMetaclass):
     """MetaValidationAction registers ValidationAction as they are defined, adding them to
     the registry.
--- a/great_expectations/checkpoint/actions.py
+++ b/great_expectations/checkpoint/actions.py
@@ -23,7 +23,7 @@ from typing import (
 )
 
 import requests
-from typing_extensions import Annotated
+from typing_extensions import dataclass_transform
 
 from great_expectations._docs_decorators import public_api
 from great_expectations.analytics.client import submit as submit_event
@@ -35,6 +35,7 @@ from great_expectations.compatibility.pydantic import (
     BaseModel,
     Extra,
     Field,
+    ModelMetaclass,
     root_validator,
     validator,
 )
@@ -50,6 +51,10 @@ from great_expectations.data_context.types.resource_identifiers import (
 from great_expectations.data_context.util import instantiate_class_from_config
 from great_expectations.datasource.fluent.config_str import ConfigStr
 from great_expectations.exceptions import ClassInstantiationError
+from great_expectations.exceptions.exceptions import (
+    ValidationActionAlreadyRegisteredError,
+    ValidationActionRegistryRetrievalError,
+)
 from great_expectations.render.renderer import (
     EmailRenderer,
     MicrosoftTeamsRenderer,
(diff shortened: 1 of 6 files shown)


great_expectations/checkpoint/actions.py
```diff
@@ -103,8 +108,79 @@ def filter_results(self, class_: Type[ValidationAction]) -> list[dict]:
...
+    def __new__(cls, clsname, bases, attrs):
+        newclass = super().__new__(cls, clsname, bases, attrs)
+
+        action_type = newclass.__fields__.get("type")
+        if action_type and action_type.default:  # Excludes base classes
+            _VALIDATION_ACTION_REGISTRY.register(
+                action_type=action_type.default, action_class=newclass
+            )
+
+        return newclass
```
cdkini, line 179: Can this be done cleaner? I want to avoid `ValidationAction` and `DocsAction` base classes that don't have `type`

great_expectations/checkpoint/actions.py
```diff
@@ -50,6 +51,10 @@
 from great_expectations.data_context.util import instantiate_class_from_config
 from great_expectations.datasource.fluent.config_str import ConfigStr
 from great_expectations.exceptions import ClassInstantiationError
+from great_expectations.exceptions.exceptions import (
+    ValidationActionAlreadyRegisteredError,
+    ValidationActionRegistryRetrievalError,
```
cdkini, line 56: Do we need custom error types? Ruff says so with our rule we've opted in and it's a low lift.

great_expectations/checkpoint/actions.py
```diff
@@ -103,8 +108,79 @@ def filter_results(self, class_: Type[ValidationAction]) -> list[dict]:
...
+            ValidationActionRegistryRetrievalError: If the action type is not registered.
+        """
+        action_type = action_config.get("type")
+        if action_type not in self._registered_actions:
+            raise ValidationActionRegistryRetrievalError(action_type)
+
+        return self._registered_actions[action_type]
+
+
+_VALIDATION_ACTION_REGISTRY = ValidationActionRegistry()
```
billdirks, line 158: Since this isn't a constant I would lowercase it.

tests/checkpoint/test_checkpoint.py
```diff
@@ -465,15 +467,18 @@ def test_checkpoint_deserialization_failure(
 
         assert expected_error in str(e.value)
 
-    @pytest.mark.unit
-    def test_checkpoint_deserialization_with_actions(self, mocker: MockerFixture):
-        # Arrange
+    def _set_context(self, mocker: MockerFixture):
```
billdirks: This is what we were doing before but it looks dangerous since it sets some global variable in `set_context(context)` and never undoes it. All our tests should be setting this themselves if needed so maybe this is fine but it feels like we should do this as a context manager and undo it to it's previous state.

great_expectations/checkpoint/actions.py
```diff
@@ -103,8 +108,79 @@ def filter_results(self, class_: Type[ValidationAction]) -> list[dict]:
...
+
+
+@dataclass_transform(kw_only_default=True, field_specifiers=(Field,))
```
billdirks: I wonder if we need to specify this this decorator is also on pydantic's `ModelMetaclass`? It's fine if we need it but it would be good if we don't need to be aware of that detail in pydantic.

billdirks (commented):
Thanks! This looks good. I left a few comments for consideration.

great_expectations/checkpoint/actions.py
tyler-hoffman: I don't feel strongly, but I think it's best practice. Either way, we should probably do what lint says, or otherwise discuss what rules to add/drop external to PRs.

great_expectations/checkpoint/actions.py
```diff
@@ -103,8 +108,79 @@ def filter_results(self, class_: Type[ValidationAction]) -> list[dict]:
         return [action_result for action, action_result in self._data if isinstance(action, class_)]
 
 
+class ValidationActionRegistry:
+    """
+    Registers both GX-defined and user-defined ValidationActions.
```
tyler-hoffman: > both GX-defined and user-defined ValidationActions
Is this a meaningful distinction?

great_expectations/checkpoint/actions.py
```diff
@@ -103,8 +108,79 @@ def filter_results(self, class_: Type[ValidationAction]) -> list[dict]:
...
+
+        Raises:
+            ValidationActionAlreadyRegisteredError: If the action type is already registered.
+        """
+        if action_type in self._registered_actions:
+            raise ValidationActionAlreadyRegisteredError(action_type)
+
+        self._registered_actions[action_type] = action_class
+
+    def get_from_config(self, action_config: dict) -> Type[ValidationAction]:
```
tyler-hoffman: Why pass in the whole config if we only use its `type` field? We should just pass in the `type`. This is really small, but i think it's the correct ™️  thing to do from an API perspective.

great_expectations/checkpoint/actions.py
tyler-hoffman: Isn't it though? The object mutable, but the reference is constant.

great_expectations/checkpoint/actions.py
tyler-hoffman: I remember having to do the same thing on expectation metaclasses. I don't recall the details, but I think this is needed.

great_expectations/checkpoint/actions.py
```diff
@@ -103,8 +108,79 @@ def filter_results(self, class_: Type[ValidationAction]) -> list[dict]:
...
+    """
+    Registers both GX-defined and user-defined ValidationActions.
+
+    Enables the instantiation of ValidationActions based on their configuration.
+    """
+
+    def __init__(self):
+        self._registered_actions: dict[str, Type[ValidationAction]] = {}
+
+    def register(self, action_type: str, action_class: Type[ValidationAction]) -> None:
```
tyler-hoffman, line 122: it'd be nice if `type` were a `ClassVar` on the action class and we could just pass in the class. Looks like that's not the case though?

tests/checkpoint/test_checkpoint.py
tyler-hoffman: Making it a generator fixture would mean you could do both the setup teardown, plus you wouldn't have to explicitly pass in `mocker`, which is a slight ergonomics benefit.

tests/checkpoint/test_checkpoint.py
```diff
@@ -496,6 +501,69 @@ def test_checkpoint_deserialization_with_actions(self, mocker: MockerFixture):
...
+            "name": "my_checkpoint",
+            "validation_definitions": [
+                {"id": "3fb9ce09-a8fb-44d6-8abd-7d699443f6a1", "name": "my_validation_def"}
+            ],
+        }
+
+        checkpoint = Checkpoint.parse_obj(serialized_checkpoint)
+
+        # Assert
+        assert checkpoint.actions[0].type == "custom"
```
tyler-hoffman: I think we know this to be the case, but an isinstance check probably gives us a slight bit more confidence here.

tyler-hoffman (commented):
Looks great to me! I'd love to see the slight tweak to `get_from_config`'s signature, but that's not blocking. The only reason I'm not approving is because AFAIK this will 422 against cloud, so we need to get the backend changes in first. LMK if I'm mistaken!

tyler-hoffman (changes requested):
Actually I'm going to throw a blocker on here just to make sure we don't accidentally release something that breaks against cloud. I'll approve once cloud is ready for this!

great_expectations/checkpoint/actions.py
cdkini: It's what enables extensibility so I thought to call it out - happy to remove the distinction if we don't see it adding value.

tyler-hoffman: My thought is that it makes an arbitrary distinction, which could cause confusion. Your call on if you agree or not. Or maybe get another opinion. IDK, it's totally non-blocking

great_expectations/checkpoint/actions.py
cdkini: Ah no and I think that changes some serialization behavior? I'd vote to not update that here

great_expectations/checkpoint/actions.py
cdkini: I was thinking the internals around `type` should be encapsulated within the action file and related classes but I agree, this feels a little inelegant.

great_expectations/checkpoint/actions.py
cdkini: I could go either way but my vote is to leave it as is

great_expectations/checkpoint/actions.py
cdkini: Enables type hinting (breaks otherwise) - added a comment

tests/checkpoint/test_checkpoint.py
cdkini: Made a fixture

great_expectations/checkpoint/actions.py
billdirks: When I say a constant object I mean an immutable object; eg what you would get in a data class when you set `frozen=True`. I think that makes it easier to reason about a variable thats all caps means.

great_expectations/checkpoint/actions.py
billdirks: I agree. I think it's clearer to write "Registers actions" or something. Then a reader doesn't even need to think if there is a difference between the two (and there isn't).

great_expectations/checkpoint/actions.py
```diff
@@ -103,8 +108,79 @@ def filter_results(self, class_: Type[ValidationAction]) -> list[dict]:
...
+
+        Raises:
+            ValidationActionAlreadyRegisteredError: If the action type is already registered.
+        """
+        if action_type in self._registered_actions:
+            raise ValidationActionAlreadyRegisteredError(action_type)
+
+        self._registered_actions[action_type] = action_class
+
+    def get(self, action_type: str | None) -> Type[ValidationAction]:
```
tyler-hoffman, line 138: Doesn't this syntax not work for 3.9?

cdkini: `str | None` vs `Optional[str]`? I think that's fine?

tyler-hoffman (approved):
Looks great!

great_expectations/checkpoint/actions.py
tyler-hoffman: I guess you're right! I have no idea what I'm doing!

PR #10745, opened 2024-12-06
[BUGFIX] Fix like pattern expectations to work without optional SQL deps

joshua-stauffer:
Previously, `LikePattern` expectations would fail if run in an environment without MySQL, MsSQL, or Postgresql Sqlalchemy libs due to a faulty attribute check. This PR updates the logic to be resilient in case those optional deps are not installed.


great_expectations/expectations/metrics/util.py
```diff
@@ -902,16 +902,26 @@ def get_dialect_like_pattern_expression(  # noqa: C901, PLR0912, PLR0915
...
-            dialect.dialect,
-            (
-                sa.dialects.sqlite.dialect,
-                sa.dialects.postgresql.dialect,
-                sa.dialects.mysql.dialect,
-                sa.dialects.mssql.dialect,
-            ),
-        ):
-            dialect_supported = True
+        try:
```
tyler-hoffman, line 905: It doesn't matter, and this isn't on you, but this is just reads so badly / isn't DRY when it feels like it should be. Trying to think if there's a way to dry this up. The only thing that comes to mind is to have a list of functions that get the dialect, something like
```
dialect_getters = [lambda: sa.dialects.sqlite.dialect, lambda: a.dialects.mysql.dialect, ...]
for get_dialect in dialect_getters:
  try:
    if issubclass(dialect.dialect, get_dialect()):
      ..
   except:
   ..
```

tests/integration/data_sources_and_expectations/test_canonical_expectations.py
```diff
@@ -154,3 +154,14 @@ def test_expect_column_mean_to_be_between(batch_for_datasource):
     expectation = gxe.ExpectColumnMeanToBeBetween(column="a", min_value=2, max_value=3)
     result = batch_for_datasource.validate(expectation)
     assert result.success
+
+
+@parameterize_batch_for_data_sources(
+    data_source_configs=SQL_DATA_SOURCES,
+    data=pd.DataFrame({"a": ["Chicago", "chicago", "CHICAGO"]}),
+)
+def test_expect_column_value_to_match_like_pattern(batch_for_datasource):
```
tyler-hoffman: We have a dedicated file for testing this type of expectation now. So this should be moved over there. But does this prove anything those tests didn't prove? Looks like I only had a subset of data sources running, so maybe that's the difference?

It's hard to tell that this works based on the integration test since the issue is so low-level. I'd consider also adding a unit test around the method you updated that mocks out the existence of those deps.

tyler-hoffman (commented):
I'm refraining from approving/blocking for now. I don't understand how this is different than the existing tests that didn't catch this.

tests/integration/data_sources_and_expectations/test_canonical_expectations.py
joshua-stauffer: yes, that's a great idea - unit test added here: [e290b2e](https://github.com/great-expectations/great_expectations/pull/10745/commits/e290b2ecd5aec3809e4c9029a14e15292485a304)

great_expectations/expectations/metrics/util.py
joshua-stauffer: 😂 i did think about this for approximately 5 seconds, but IMO the repetition is way more legible and clear than any shorter option i could think of, outside of rewriting the whole way this function is being used. In any case, not going to make that change here.

tyler-hoffman: Cool, works for me!

tyler-hoffman (approved):
Sweet! Confirmed the test fails on develop!

PR #10774, opened 2024-12-12
[MAINTENANCE] Add EventBridge Scheduler service to `cloud-tests`

NathanFarmer:
Add EventBridge Scheduler service to `cloud-tests`


assets/docker/mercury/docker-compose.yml
```diff
@@ -100,15 +110,27 @@ services:
...
       db-provisioner:
         condition: service_completed_successfully
+      localstack:
+        condition: service_healthy
     entrypoint:
       - /bin/bash
       - -c
       - |
+        cd /app/services/ge_cloud
+        ./scripts/localstack_init.sh
```
billdirks, line 133: I think you can remove the 2 alembic lines in this entrypoint since they are about setting the db up and the v0 mercury service already does that.

tests/datasource/fluent/test_pandas_datasource.py
```diff
@@ -486,6 +493,8 @@ def test_cloud_get_csv_asset_not_in_memory(valid_file_path: pathlib.Path):
 
     assert csv_asset_name not in context.data_sources.all()._in_memory_data_assets
 
+    context.data_sources.delete(name=datasource_name)
```
billdirks: If this is meant to clean up we should do it in a `try/finally` block. For example, if the assert fails in the line above, this will never be run.

billdirks (changes requested):
Thanks for fixing this! I have 2 small cleanup asks inline.

tests/datasource/fluent/test_pandas_datasource.py
```diff
@@ -474,17 +475,26 @@ def test_read_dataframe(empty_data_context: AbstractDataContext, test_df_pandas:
...
+        cloud_base_url=os.environ.get("GX_CLOUD_BASE_URL"),
+        cloud_organization_id=os.environ.get("GX_CLOUD_ORGANIZATION_ID"),
+        cloud_access_token=os.environ.get("GX_CLOUD_ACCESS_TOKEN"),
     )
-    csv_asset = datasource.get_asset(name=csv_asset_name)
-    csv_asset.build_batch_request()
+    datasource_name = f"DS_{uuid.uuid4().hex}"
+    csv_asset_name = f"DA_{uuid.uuid4().hex}"
+    try:
+        datasource = context.data_sources.add_pandas(name=datasource_name)
```
billdirks: nit: The add_pandas line should be pulled out of the try block because you can't delete it if it doesn't get successfully created.

billdirks (approved):
One small comment but otherwise LGTM. Thanks!

tests/datasource/fluent/test_pandas_datasource.py
NathanFarmer: Right

PR #10818, opened 2025-01-03
[MAINTENANCE] Add `context.validation_definitions.add_or_update` support

cdkini:
Bring back add or update functionality through the validation definition factory


great_expectations/core/factory/validation_definition_factory.py
```diff
@@ -103,3 +104,24 @@ def get(self, name: str) -> ValidationDefinition:
...
+        """Add or update an ValidationDefinition by name.
+
+        If an ValidationDefinition with the same name exists, overwrite it, otherwise
+        create a new ValidationDefinition.
+
+        Args:
+            validation: ValidationDefinition to add or update
+        """
+        try:
+            _ = self.get(name=validation.name)
```
joshua-stauffer: i think we'll need this value to add the `id` to the validation

great_expectations/data_context/data_context/context_factory.py
```diff
@@ -111,6 +112,9 @@ def get_datasources(self) -> DatasourceDict:
     def get_validator(self, batch_request: BatchRequest) -> Validator:
         return self._project.get_validator(batch_request=batch_request)
 
+    def get_suite_factory(self) -> SuiteFactory:
+        return self._project.suites
```
joshua-stauffer, line 116: i think our latest pattern here is to access the `__project` directly. but there's some inconsistency, so i'm not 100% sure.

great_expectations/core/factory/validation_definition_factory.py
```diff
@@ -103,3 +104,24 @@ def get(self, name: str) -> ValidationDefinition:
...
+        except DataContextError:
+            return self.add(validation=validation)
+
+        suite_factory = project_manager.get_suite_factory()
+        validation.suite = suite_factory.add_or_update(suite=validation.suite)
+        validation.save()
```
joshua-stauffer, line 133: how does this work with the batch definition?

great_expectations/core/factory/validation_definition_factory.py
```diff
@@ -103,3 +104,44 @@ def get(self, name: str) -> ValidationDefinition:
...
+
+        batch_definition = validation.batch_definition
+        asset = batch_definition.data_asset
+        data_source = asset.datasource
+
+        updated_data_source = data_source_manager.add_or_update(data_source=data_source)
+        updated_asset = updated_data_source.get_asset(asset.name)
+        updated_batch_definition = updated_asset.get_batch_definition(batch_definition.name)
+
+        batch_definition.id = updated_batch_definition
```
joshua-stauffer: this a bug?

great_expectations/core/factory/validation_definition_factory.py
```diff
@@ -103,3 +104,43 @@ def get(self, name: str) -> ValidationDefinition:
...
+        validation.save()
+
+        return validation
+
+    def _add_or_update_batch_definition(self, validation: ValidationDefinition) -> None:
+        data_source_manager = project_manager.get_data_source_manager()
+
+        batch_definition = validation.batch_definition
+        asset = batch_definition.data_asset
+        data_source = asset.datasource
```
joshua-stauffer: how do we expect a user to create these relationships?

great_expectations/data_context/data_context/context_factory.py
cdkini: Hmm is it? I'm seeing a lot of `_project` usage since it is a property that will panic if the context isn't set?

great_expectations/core/factory/validation_definition_factory.py
cdkini: I don't think we do - we get these from the fluent API

great_expectations/core/factory/validation_definition_factory.py
cdkini: what specifically? let me know if this is still present in the current iteration

great_expectations/core/batch_definition.py
```diff
@@ -69,6 +69,15 @@ def build_batch_request(
             partitioner=self.partitioner,
         )
 
+    @public_api
+    def save(self) -> None:
```
tyler-hoffman, line 73: We didn't have this already?!

great_expectations/core/factory/validation_definition_factory.py
```diff
@@ -103,3 +104,32 @@ def get(self, name: str) -> ValidationDefinition:
...
+        Args:
+            validation: ValidationDefinition to add or update
+        """
+        # Always add or update underlying suite to avoid freshness issues
+        suite_factory = project_manager.get_suite_factory()
+        validation.suite = suite_factory.add_or_update(suite=validation.suite)
+
+        try:
+            existing_validation = self.get(name=validation.name)
+            existing_batch_definition = existing_validation.data
```
tyler-hoffman, line 124: why are we getting the batch def but not the suite? Don't we need the same logic for both child objects?

tests/core/test_batch_definition.py
```diff
@@ -259,3 +259,27 @@ def test_is_fresh_fails_on_batch_definition_retrieval(in_memory_runtime_context)
     assert diagnostics.success is False
     assert len(diagnostics.errors) == 1
     assert isinstance(diagnostics.errors[0], BatchDefinitionNotFoundError)
+
+
+@pytest.mark.unit
+def test_save(in_memory_runtime_context):
```
tyler-hoffman: nit: Can you throw a type hint on the fixture? Makes it slightly nicer for autocomplete/type errors in the test body.

tests/core/factory/test_validation_definition_factory.py
```diff
@@ -466,6 +466,152 @@ def test_validation_definition_factory_round_trip(
...
+        batch_def = self._build_batch_definition(context)
+        suite = self._build_suite()
+        vd = ValidationDefinition(
+            name=vd_name,
+            data=batch_def,
+            suite=suite,
+        )
+
+        # act
+        created_vd = context.validation_definitions.add_or_update(validation=vd)
```
tyler-hoffman, line 548: Maybe I'm missing it, but I don't see how this is different than `_test_add_new_validation`. I guess I assumed from than name that we'd persist a validation, then do an upsert with a new suite.

tests/core/factory/test_validation_definition_factory.py
```diff
@@ -466,6 +466,152 @@ def test_validation_definition_factory_round_trip(
...
+        suite = self._build_suite()
+        vd = ValidationDefinition(
+            name=vd_name,
+            data=batch_def,
+            suite=context.suites.add(suite),
+        )
+        existing_vd = context.validation_definitions.add(validation=vd)
+
+        # act
+        vd.suite.expectations = [
```
tyler-hoffman, line 581: We should probably also have a test where we don't mutate the VD, but instead create a new one.

tyler-hoffman (changes requested):
Left some questions, but blocking because I think we need a bit more test coverage around:
* Creating new VDs, rather than mutating
* Add/update using new BatchDefs.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -734,11 +734,10 @@ def add_batch_definition(
...
 
-        output = updated_datasource.get_asset(asset_name).get_batch_definition(
-            batch_definition.name
-        )
-        output.set_data_asset(batch_definition.data_asset)
-        return output
+        updated_asset = updated_datasource.get_asset(asset_name)
+        updated_batch_definition = updated_asset.get_batch_definition(batch_definition.name)
+
+        return updated_batch_definition
```
cdkini, line 740: Cloud tests seem to fail when we call `set_data_asset` on the old `batch_definition.data_asset` 😕 - still determining why this fixes it and if this is problematic

great_expectations/core/batch_definition.py
cdkini: Nope think we pushed it

great_expectations/core/factory/validation_definition_factory.py
cdkini: So batch definitions will always be created before (invariant based on our fluent API) - thus, this will always be an update (`save()`). If there's a way to make this clearer, let's discuss!

tests/core/factory/test_validation_definition_factory.py
cdkini: The suite here is not yet persisted (the other test already has the suite saved to the store)

tests/core/factory/test_validation_definition_factory.py
cdkini: Added a test for this

tests/core/factory/test_validation_definition_factory.py
tyler-hoffman: Oh I see now, thanks!

tyler-hoffman (approved):

PR #10843, opened 2025-01-10
[MAINTENANCE] Remove xfailed test

NathanFarmer:

### DIFF
--- a/tests/integration/data_sources_and_expectations/expectations/test_expect_column_values_to_be_in_type_list.py
+++ b/tests/integration/data_sources_and_expectations/expectations/test_expect_column_values_to_be_in_type_list.py
@@ -76,13 +76,11 @@ def test_success_complete(batch_for_datasource: Batch) -> None:
     assert result_dict["observed_value"] in type_list
 
 
-@pytest.mark.xfail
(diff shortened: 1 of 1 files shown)


tests/integration/data_sources_and_expectations/expectations/test_expect_column_values_to_be_in_type_list.py
```diff
@@ -76,31 +76,6 @@ def test_success_complete(batch_for_datasource: Batch) -> None:
     assert result_dict["observed_value"] in type_list
 
 
-@pytest.mark.xfail
-@parameterize_batch_for_data_sources(
-    data_source_configs=[DatabricksDatasourceTestConfig()],
-    data=DATA,
-)
-def test_success_complete_errors(batch_for_datasource: Batch) -> None:
```
NathanFarmer, line 84: This test could be passing by adding `INT` to the array, but we already have a "success" test below this for Databricks. It's not clear how this is meant to test "errors". This is the only xfailed test left for the type/type list Expectations.

tyler-hoffman (approved):
I don't recall what the original point here was either, though looks like it's been updated since the test was created. Either way, seems safe to remove.

PR #10861, opened 2025-01-16
[MAINTENANCE] Handle `aws-chunked` encoding type data in `TupleS3StoreBackend`

NathanFarmer:
Handle a `boto3` `ContentEncoding` string that takes the form of a list 😃 


great_expectations/data_context/store/tuple_store_backend.py
```diff
@@ -557,19 +557,35 @@ def _get_all(self) -> list[Any]:
...
+        # encodings included with `aws-chunked` we will remove the `aws-chunked` string from
+        # the list. We do not intend to add support for reading in chunks at this time.
+        # Calling botocore.response.StreamingBody.read() without arguments
+        # will read the entire stream.
+        encodings: list[str] = content_encoding.split(",")
+        if "aws-chunked" in encodings:
+            encodings.remove("aws-chunked")
+        data = s3_response_object["Body"].read()
+        # Maybe it's theoretically possible to encode an entire file with multiple encodings.
+        # It's an odd choice of format for ContentEncoding to allow multiple encoding strings.
```
tyler-hoffman, line 583: I'm a little worried about what "as far as we know...", but this seems like it moves us forward.

tyler-hoffman (approved):

great_expectations/data_context/store/tuple_store_backend.py
NathanFarmer: If `ContentEncoding` is empty or contains only `aws-chunked`, we will use `utf-8` like before. If there are multiple encodings other than `aws-chunked` it will likely cause an error here, but that may mean the intent of this field is something other than what the description in the API docs says. At the end of the day, this is just unfortunate.

NathanFarmer: I edited that comment language on line 584 slightly to make it clear.

PR #10869, opened 2025-01-16
[MAINTENANCE] Add user-agent-str to analytics events

tyler-hoffman:
Parameterize events (and contexts) with `user_agent_str`. This will be used by gx operators (and can be used by the agent/runner as well).

This touches more files than I would have expected, but the high level is:
* 4 files are just passing the field through context constructors
* analytics / event classes / wiring it up are a few more
* Minor updates to and added test cases.

Should replace https://github.com/great-expectations/great_expectations/pull/10651.


great_expectations/analytics/client.py
```diff
@@ -56,7 +57,13 @@ def init(  # noqa: PLR0913 # FIXME CoP
         conf["organization_id"] = organization_id
     if oss_id:
         conf["oss_id"] = oss_id
-    update_config(config=Config(cloud_mode=cloud_mode, **conf))
+    update_config(
+        config=Config(
+            cloud_mode=cloud_mode,
+            user_agent_str=user_agent_str,
```
tyler-hoffman, line 63: mypy was upset when I put the string in `conf`. I think we can probably clean this method up at some point - there's a bunch of logic that doesn't really impact behavior.

joshua-stauffer (approved):
looks good, thanks! i think we'll also need an API to set this, for the case in which we don't control the initialization of the context but are still running actions against it. but no reason to include it here.

PR #10892, opened 2025-01-27
[MAINTENANCE] Add `row_condition` datetime testing for Pandas and Spark

NathanFarmer:
- Gets x-failed pandas and spark row condition tests passing
- Adds support for `column_types` to spark in Expectation testing framework
- Adds support for pandas/spark i/o options
- Fix latent bug due to spark writing partitions with the string `filename.csv` in them, but those filenames don't END in that string. We have a limitation for `FileDataAsset` that you can only read in one file. The `DirectoryAsset` is used for reading multiple partitions.
- Add `README` for running spark tests locally

-------------


tests/integration/conftest.py
```diff
@@ -29,7 +29,6 @@ def __hash__(self) -> int:
         return hash(
             (
                 self.__class__,
-                self.data_source_config,
```
NathanFarmer, line 32: This hashing function started failing once I added the read/write options to the Pandas/Spark configs.

tests/integration/test_utils/data_source_config/pandas_filesystem_csv.py
```diff
@@ -13,7 +14,13 @@
 )
 
 
+@dataclass(frozen=True)
 class PandasFilesystemCsvDatasourceTestConfig(DataSourceTestConfig):
+    # see https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html for options
+    read_options: dict[str, Any] = field(default_factory=dict)
+    # see https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html for options
+    write_options: dict[str, Any] = field(default_factory=dict)
```
NathanFarmer: I only ended up using these options for pandas, but I added them to spark as well. I needed them during debugging for spark and I think that will also happen in the future.

tests/integration/test_utils/data_source_config/spark_filesystem_csv.py
```diff
@@ -72,7 +113,9 @@ def make_batch(self) -> Batch:
     @override
     def setup(self) -> None:
         file_path = self._base_dir / self.csv_path
-        self.data.to_csv(file_path, index=False)
+        self._spark_data.write.format("csv").option("header", True).options(
+            **self.config.write_options
+        ).save(str(file_path))
```
NathanFarmer, line 118: We are writing csv files using spark now. This symmetrical behavior was important because I was having trouble debugging reading datetimes that were written by pandas.

tests/integration/test_utils/data_source_config/spark_filesystem_csv.py
```diff
@@ -55,11 +67,40 @@ def __init__(
...
+    @property
+    def _spark_schema(self) -> Union["pyspark_types.StructType", None]:
+        from great_expectations.compatibility.pyspark import types as pyspark_types
+
+        column_types = self.config.column_types or {}
+        struct_fields = [
+            pyspark_types.StructField(column_name, column_type())
+            for column_name, column_type in column_types.items()
+        ]
+        return pyspark_types.StructType(struct_fields) if struct_fields else None
```
NathanFarmer, line 83: Again I only ended up adding this for debugging, but I think it's valuable. Now you can pass `column_types` to Spark tests and it will convert that to a spark schema.

tests/integration/test_utils/data_source_config/spark_filesystem_csv.py
NathanFarmer: It's likely this is actually the required behavior to get the tests passing. There were issues before because the datetimes were being interpreted as strings and could not be compared with lt/gt.

tests/integration/data_sources_and_expectations/expectations/test_expect_compound_column_values_to_be_unique.py
```diff
@@ -28,7 +40,10 @@
 
 @parameterize_batch_for_data_sources(data_source_configs=ALL_DATA_SOURCES, data=DATA)
 def test_golden_path(batch_for_datasource: Batch) -> None:
-    expectation = gxe.ExpectCompoundColumnsToBeUnique(column_list=[STRING_COL, INT_COL, INT_COL_2])
+    expectation = gxe.ExpectCompoundColumnsToBeUnique(
+        column_list=[STRING_COL, INT_COL, INT_COL_2],
+        ignore_row_if="any_value_is_missing",
```
NathanFarmer, line 45: This test should not have been passing without this flag set. It seems like there is still a bug that allows it to pass with SQL and pandas from my understanding. The set [`None`, `None`, `None`] is not unique.

great_expectations/expectations/row_conditions.py
```diff
@@ -119,35 +119,21 @@ def _parse_great_expectations_condition(row_condition: str):
...
-            return F.col(column) > num
-        elif op == "<":
-            return F.col(column) < num
-        elif op == ">=":
-            return F.col(column) >= num
-        elif op == "<=":
-            return F.col(column) <= num
-        elif op == "==":
-            return F.col(column) == num
+        return generate_condition_by_operator(F.col(column), parsed["op"], F.lit(num))
```
tyler-hoffman, line 136: well this seems nicer!

tests/integration/conftest.py
tyler-hoffman: What was the failure? If we don't include this in the hash, I would expect to run into issues with thinking that 2 TestConfigs with different data source types are equivalent. I'm going to throw a blocker on until we've discussed this.

tests/integration/data_sources_and_expectations/expectations/test_expect_column_values_to_be_of_type.py
```diff
@@ -37,6 +37,19 @@
...
+try:
+    from great_expectations.compatibility.pyspark import types as PYSPARK_TYPES
+
+    SPARK_COLUMN_TYPES = {
+        INTEGER_COLUMN: PYSPARK_TYPES.IntegerType,
+        INTEGER_AND_NULL_COLUMN: PYSPARK_TYPES.IntegerType,
+        STRING_COLUMN: PYSPARK_TYPES.StringType,
+        NULL_COLUMN: PYSPARK_TYPES.IntegerType,
+    }
+except ModuleNotFoundError:
```
tyler-hoffman, line 49: bummer we have to do this, but I think it seems reasonable.

tyler-hoffman (changes requested):
I want to discuss the changes to the hash method before this merges. I suspect that it introduces a bug (which may not yet be caught by CI)

tests/integration/conftest.py
NathanFarmer: Something like "unable to hash dictionary". Wouldn't the hash of `data` always be unique?

tyler-hoffman: The original thinking was that when we hash, it should be unique to the data AND the datasource. Like we want to differentiate between MY_DATA for postgres vs MY_DATA for snowflake, but I'm realizing now that this is fine; even on hash collisions, we still store both in the set and it should work. FWIW, I think excluding the new read/write fields from the hash might be the most ideal state (I think you can just pass in `hash=False` to those `field`s. But I think we're fine either way.

tyler-hoffman (approved):

tests/integration/conftest.py
NathanFarmer: I'll try the `hash=False` solution and then if that doesn't work I'll try to use `json.dumps`.

tyler-hoffman: I trust your judgement on it, but I'd probably use what you currently have over using json.dumps, in case equivalent data serialized differently, e.g. keys in a dictionary showing up in a different order. Maybe we'd never hit that since order is preserved with python 3, but idk. Either way, I think if it works, we should ship it!

NathanFarmer: I do think you can do something like `json.dumps(d, sort_keys=True)` when your keys are all strings.

PR #10906, opened 2025-01-31
[MAINTENANCE] Add `strict` to Window type

NathanFarmer:
Add `strict` to Window type so it can be used with dynamic parameters Expectations

### DIFF
--- a/great_expectations/expectations/window.py
+++ b/great_expectations/expectations/window.py
@@ -23,7 +23,7 @@ class Window(pydantic.BaseModel):
     parameter_name: str
     range: int
     offset: Offset
-    strict: bool
+    strict: bool = False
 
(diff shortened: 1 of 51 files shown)


great_expectations/expectations/window.py
```diff
@@ -23,6 +23,7 @@ class Window(pydantic.BaseModel):
     parameter_name: str
     range: int
     offset: Offset
+    strict: bool
```
NathanFarmer: @shiplet is adding a new required parameter here a breaking change? I don't see windows anywhere in the docs or gallery, which tells me it isn't really part of the public API. I think we are probably fine to make this required, but I could add a default if there is risk.

shiplet (approved):

PR #10916, opened 2025-02-06
[MAINTENANCE] Ensure clear error if cloud mode requested and env vars aren't found

joshua-stauffer:
Previously, there was a bug where the legacy param `cloud_mode` would always be `None` in `ProjectManager._get_cloud_context`. As a result, the error intended for users who set `mode="cloud"` was never reached, and users received an opaque message about their DataContext being `None`. This PR clarifies the logic, and maintains the same behavior as before:
- if cloud envs are set, the user gets a CloudDataContext
- if cloud envs are set but the user specifies `cloud_mode=False`, the user gets an EphemeralDataContext
- if the user sets `mode=...`, it takes precedence over `cloud_mode`

Example of users hitting this bug: https://discourse.greatexpectations.io/t/valueerror-when-trying-to-create-clouddatacontext/2024

### DIFF
--- a/great_expectations/data_context/data_context/context_factory.py
+++ b/great_expectations/data_context/data_context/context_factory.py
@@ -351,7 +351,7 @@ class ProjectManager:
             )
         else:
             raise GXCloudConfigurationError(  # noqa: TRY003 # one time exception
-                "Unable to create a CloudDataContext due to one or more missing environment variables: " 
+                "Unable to create a CloudDataContext due to one or more missing environment variables: "
                 "GX_CLOUD_ORGANIZATION_ID, GX_CLOUD_ACCESS_TOKEN"
             )
 
--- a/great_expectations/data_context/data_context/context_factory.py
+++ b/great_expectations/data_context/data_context/context_factory.py
@@ -206,6 +206,7 @@ class ProjectManager:
                 cloud_organization_id=cloud_organization_id,
                 user_agent_str=user_agent_str,
                 cloud_mode=cloud_mode,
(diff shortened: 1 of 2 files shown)


great_expectations/data_context/data_context/context_factory.py
```diff
@@ -326,7 +329,11 @@ def _get_cloud_context(  # noqa: PLR0913 # FIXME CoP
         cloud_organization_id: str | None = None,
         user_agent_str: str | None = None,
         cloud_mode: bool | None = None,
+        mode: ContextModes | None = None,
     ) -> CloudDataContext | None:
+        if cloud_mode is False:
```
tyler-hoffman, line 334: Completely non-blocking, just complaining. Why is this even a parameter? this is equivalent to calling `do_the_thing(actually_do_not_do_the_thing=True)`. I'm assuming it's because of upstream stuff that bled down to this layer?

tyler-hoffman (approved):

great_expectations/data_context/data_context/context_factory.py
joshua-stauffer: yeah, `cloud_mode` used to be the canonical way to get a CloudDataContext. And when we switched to the `mode` enum, we left it to avoid breaking changes. And we missed removing it in v1...

joshua-stauffer: it just means our API has two ways to do the same thing, like pretty much everywhere else 😄 bonus points for consistency?

PR #10920, opened 2025-02-07
[MAINTENANCE] Add `Metric` and `Domain` base classes

NathanFarmer:
This PR adds new `Metric` and `Domain` base classes for defining and instantiating metrics.

Example metric definition:
```
class ColumnValuesBetween(Metric, ColumnValues):
    min_value: Optional[Comparable] = None
    max_value: Optional[Comparable] = None
    strict_min: bool = False
    strict_max: bool = False
```

Example metric instantiation:
```
from great_expectations.metrics import ColumnValuesBetween

column_values_between = ColumnValuesBetween(
    batch_id=...,
    table=...,
    column=...,
    min_value=...,
    max_value=...,
)
```
-----------


great_expectations/metrics/column_map/between.py
```diff
@@ -0,0 +1,12 @@
...
+from great_expectations.core.types import Comparable
+from great_expectations.metrics import Metric
+from great_expectations.metrics.domain import ColumnMap
+
+
+class Between(Metric, ColumnMap):
+    min_value: Optional[Comparable] = None
+    max_value: Optional[Comparable] = None
+    strict_min: bool = False
+    strict_max: bool = False
```
NathanFarmer: We still need to add a docstring and tests for this. I only put it in the PR as a demonstration of how it will work.

great_expectations/metrics/__init__.py
```diff
@@ -0,0 +1 @@
+from .metric import Metric
```
billdirks: Do we want to have all metrics available at `gx.metrics.*`. That way people don't need to reach into `column_map` to find their metric.

great_expectations/metrics/registry.py
```diff
@@ -0,0 +1,11 @@
+from typing import Final
+
+from great_expectations.metrics.domain import ColumnMap, Domain
+
+DOMAIN_NAMES: Final[dict[type[Domain], str]] = {
```
billdirks: It would be nice if we didn't have to manually edit this file every time we added a domain or metric. Instead of making these `Final` could we add an add method that we call after the domain/metric definition or automatically gets added in the metaclass by looking at a class variable?

great_expectations/metrics/metric.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+
+    def __new__(cls, *args, **kwargs):
+        if cls is Metric:
+            raise AbstractClassInstantiationError(cls.__name__)
+        return super().__new__(cls)
+
+    @property
+    def id(self) -> tuple[str, str, str]:
+        return self.to_config().id
```
billdirks: Ultimately, I'd rather not have to call `to_config` but have it computed at instantiation time and fixed. Maybe this is done in a followup ticket due to that other bug?

billdirks (approved):
I've left a few comments that we can discuss but that don't block this PR. Thanks for doing this!

great_expectations/metrics/registry.py
```diff
@@ -0,0 +1,11 @@
...
+    ColumnMap: "column_values",
+}
+
+METRIC_REGISTRY: Final[dict[str, set[str]]] = {
+    DOMAIN_NAMES[ColumnMap]: {"between"},
+}
```
joshua-stauffer: what value do we get out of having a global registry? I see two uses here:
- a place to store the name of a domain class
- check that contributors are correctly wiring up metrics to domains

For the first case, i think we're better served by just having a `name` attribute on the domain class. I'm not convinced the second use case is actually adding any value, since a) the check is aimed at library authors, not users, and b) the check can be easily circumvented by just adding a magic string to this file.

It's possible that i'm missing a path that makes this pattern valuable, so help me fill in the gaps if that's the case!

joshua-stauffer (dismissed):
this looks really good! i have one question around the pattern though, so blocking for more conversation.

great_expectations/metrics/__init__.py
NathanFarmer: With the current syntax, the issue with that is that it might not be clear what the metric represents from the import:

```
from great_expectations.metrics.column_map import Between
```
vs the proposed:
```
from great_expectations.metrics import Between
```

I'm pretty sure based on our current metric names there wouldn't be any namespace collisions though (all imports `from great_expectations.metrics` would be distinct in the proposed import.

great_expectations/metrics/metric.py
NathanFarmer: Good call. I think now that `id`s are immutable we can do that.

great_expectations/metrics/__init__.py
NathanFarmer: We discussed this offline

great_expectations/metrics/metric.py
NathanFarmer: I went ahead and did this, but it has the side effect of making all metric attributes immutable. Otherwise, we need to be able to call `to_config` on attribute assignment to update the config.

great_expectations/metrics/registry.py
NathanFarmer: I went ahead and nuked the registry. It will matter how we name classes now, but the existing metrics system will complain if that is done incorrectly.

great_expectations/metrics/registry.py
NathanFarmer: I removed the registry entirely. There is already a registry in the existing metrics system and the metric names are inferred entirely from the class names.

great_expectations/metrics/registry.py
NathanFarmer: If we don't like the existing names as class names we can add a `name` override string

great_expectations/metrics/metric.py
```diff
@@ -0,0 +1,149 @@
...
+
+        A metric for a single table row count value:
+
+        >>> class RowCount(Metric, Table):
+        ...     ...
+
+    Notes:
+        - The Metric class cannot be instantiated directly - it must be subclassed.
+        - Subclasses must specify a Domain type as a mixin.
+        - The subclass name and specified Domain type must be registered in the METRIC_REGISTRY.
```
billdirks: I think this no longer applies since you've removed the registry?

NathanFarmer: Nice catch!

PR #10929, opened 2025-02-11
[BUGFIX] Make `MetricConfiguration.id` immutable

NathanFarmer:
- Makes `MetricConfiguration.id` immutable, and updates references to mutated `id`s
- Narrows types by adding `MetricConfigurationID` and `IDDictID` and update references

There is buggy behavior where `MetricConfiguration.id`s are changing in the middle of metric computation, due to changes in `metric_domain_kwargs` and `metric_value_kwargs`, which are referenced in the `id` property.

---------

### DIFF
--- a/great_expectations/core/id_dict.py
+++ b/great_expectations/core/id_dict.py
@@ -4,16 +4,21 @@ import hashlib
 import json
 from typing import Any, Set, TypeVar, Union
 
+from typing_extensions import Never
+
 from great_expectations.compatibility.typing_extensions import override
 from great_expectations.util import convert_to_json_serializable  # noqa: TID251 # FIXME CoP
 
 T = TypeVar("T")
 
 
+IDDictID = Union[str, tuple[Never, ...]]
+
+
 class IDDict(dict):
     _id_ignore_keys: Set[str] = set()
 
-    def to_id(self, id_keys=None, id_ignore_keys=None):
+    def to_id(self, id_keys=None, id_ignore_keys=None) -> IDDictID:
         if id_keys is None:
             id_keys = self.keys()
         if id_ignore_keys is None:
--- a/great_expectations/core/id_dict.py
+++ b/great_expectations/core/id_dict.py
@@ -4,15 +4,13 @@ import hashlib
 import json
 from typing import Any, Set, TypeVar, Union
 
-from typing_extensions import Never
-
 from great_expectations.compatibility.typing_extensions import override
 from great_expectations.util import convert_to_json_serializable  # noqa: TID251 # FIXME CoP
 
 T = TypeVar("T")
 
 
-IDDictID = Union[str, tuple[Never, ...]]
+IDDictID = Union[str, tuple[()]]
 
 
 class IDDict(dict):
@@ -25,7 +23,7 @@ class IDDict(dict):
             id_ignore_keys = self._id_ignore_keys
         id_keys = set(id_keys) - set(id_ignore_keys)
         if len(id_keys) == 0:
-            return tuple()
+            return ()
         elif len(id_keys) == 1:
             key = list(id_keys)[0]
             return f"{key}={self[key]!s}"
--- a/great_expectations/expectations/core/expect_table_row_count_to_equal_other_table.py
+++ b/great_expectations/expectations/core/expect_table_row_count_to_equal_other_table.py
@@ -300,7 +300,7 @@ class ExpectTableRowCountToEqualOtherTable(BatchExpectation):
         # computes the same metric over more than one domain
         # ValidationDependencies does not allow duplicate metric names
         # and the registry is checked to ensure the metric name is registered
-        # As a work-around, after the regsitry check
+        # As a work-around, after the registry check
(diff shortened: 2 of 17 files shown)


great_expectations/experimental/metric_repository/metric_retriever.py
```diff
@@ -293,7 +294,7 @@ def _get_table_columns(self, batch_request: BatchRequest) -> Metric:
     def _get_table_column_types(self, batch_request: BatchRequest) -> Metric:
         metric_name = MetricTypes.TABLE_COLUMN_TYPES
 
-        metric_lookup_key: _MetricKey = (metric_name, tuple(), "include_nested=True")
+        metric_lookup_key = MetricConfigurationID(metric_name, tuple(), tuple())
```
NathanFarmer: This `include_nested=True` is the only reference to a mutated `MetricConfiguration.id` I could find in the code. Everywhere else this argument gets updated is in test mock responses. At instantiation time the 3rd argument to this id is an empty tuple, but later `metric_value_kwargs` gets updated and that reference mutates the id. Now the id is no longer mutated and we needed to update this lookup value.

great_expectations/validator/validation_graph.py
```diff
@@ -274,8 +276,8 @@ def _resolve(  # noqa: C901, PLR0912, PLR0915 # FIXME CoP
                 # Access "ExecutionEngine.resolve_metrics()" method, to resolve missing "MetricConfiguration" objects.  # noqa: E501 # FIXME CoP
                 metrics.update(
                     self._execution_engine.resolve_metrics(
-                        metrics_to_resolve=computable_metrics,  # type: ignore[arg-type]  # Metric typing needs further refinement.
-                        metrics=metrics,  # type: ignore[arg-type]  # Metric typing needs further refinement.
```
joshua-stauffer, line 278: 🥳

great_expectations/expectations/core/expect_table_row_count_to_equal_other_table.py
```diff
@@ -296,23 +296,29 @@ def get_validation_dependencies(
         kwargs = configuration.kwargs if configuration else {}
         other_table_name = kwargs.get("other_table_name")
 
-        # create copy of table.row_count metric and modify "table" metric domain kwarg to be other table name  # noqa: E501 # FIXME CoP
-        table_row_count_metric_config_other: Optional[MetricConfiguration] = deepcopy(
+        # At this time, this is the only Expectation that
+        # computes the same metric over more than one domain
+        # ValidationDependencies does not allow duplicate metric names
+        # and the registry is checked to ensure the metric name is registered
```
joshua-stauffer, line 302: looks like you got this to work, so not blocking. But i am curious, where does this check happen? i looked for it, and it wasn't obvious. Seems like a more robust solution would solve this outside of the expectation.

joshua-stauffer (approved):
thanks for fixing this - left one non-blocking question that i would be curious to follow up on when you get the chance.

great_expectations/core/id_dict.py
```diff
@@ -4,16 +4,21 @@
...
 
+from typing_extensions import Never
+
 from great_expectations.compatibility.typing_extensions import override
 from great_expectations.util import convert_to_json_serializable  # noqa: TID251 # FIXME CoP
 
 T = TypeVar("T")
 
 
+IDDictID = Union[str, tuple[Never, ...]]
```
joshua-stauffer: i'm having trouble when using this type with pydantic. did you mean `IDDictID = Union[str, tuple[()]]`? Or if that isn't what you meant, what does `Never` stand for here?

NathanFarmer: Interesting, I used `tuple[()]` at first and mypy complained locally, but I think that's actually the correct type. I will try that and see if it works in CI.

great_expectations/expectations/core/expect_table_row_count_to_equal_other_table.py
NathanFarmer: I agree that we could solve this in another way, but it represents a pretty big change in the metrics system that I don't think we want to take on in this PR.

NathanFarmer: There is a super `BatchExpectation.get_validation_dependencies()` call that calls [this function](https://github.com/great-expectations/great_expectations/blob/710167a370d2d9932e19c133af265b3d20e08b1b/great_expectations/expectations/registry.py#L345), which checks if the metric exists in the registry. I will add that to the comment.

great_expectations/core/id_dict.py
NathanFarmer: @joshua-stauffer I was able to get the static analysis to work with `Union[str, tuple[()]]` if I updated `tuple()` calls to `()` when instantiating `MetricConfigurationID`. If the tests pass I will merge it like this. I think `tuple[Never, ...]` means something like "a tuple that will never have anything in it" which is semantically the same. Both `tuple()` and `()` create an empty tuple, so if this works better for your PR I will merge it this way.

PR #10932, opened 2025-02-12
[MAINTENANCE] Metric Result types

joshua-stauffer:
This PR outlines an approach for creating a general set of MetricResult types that cover the majority of metrics we expect users to request.


great_expectations/metrics/metric_results.py
```diff
@@ -0,0 +1,94 @@
...
+IDDictID = Union[str, tuple[()]]
+
+
+class MetricConfigurationID(NamedTuple):
+    metric_name: str
+    metric_domain_kwargs_id: IDDictID
+    metric_value_kwargs_id: IDDictID
+
+
+class _MetricResult(BaseModel):
```
billdirks: This is basically the interface. I think you could add `value: Any` here and have the subclasses restrain this `value` appropriately. I think all implementers have the `value` field.

If you did this maybe you could get rid of the `MetricResult` union?

great_expectations/metrics/metric_results.py
```diff
@@ -0,0 +1,94 @@
...
+    def __init__(self, expected_metric: str, actual_metric: str):
+        super().__init__(
+            f"Invalid metric: expected {expected_metric} but received {actual_metric}."
+        )
+
+
+class TableColumns(_SuccessfulMetricResult):
+    @validator("id")
+    def validate_id(cls, v):
+        if v[0] != "table.columns":
```
billdirks: It seems like you could pull this logic up to `_SuccessfulMetricResult` and instead have a class attribute `name` which here would be set to `table.columns`. That way you wouldn't need to repeat this logic.

great_expectations/metrics/metric_results.py
```diff
@@ -0,0 +1,94 @@
...
+
+    value: int
+
+
+class ErrorMetricResult(_MetricResult):
+    value: dict[str, Union[int, dict, str]]
+    error: Literal[True] = True
+
+
+MetricResult = Union[
```
billdirks: Do you need this Union. I don't like having a list that we expect to grow. If this list is small and going to stay small it is ok, but I think we can get what we need by pulling value into `_MetricResult`.

great_expectations/metrics/metric_results.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class InvalidMetricError(TypeError):
+    def __init__(self, expected_metric: str, actual_metric: str):
+        super().__init__(
+            f"Invalid metric: expected {expected_metric} but received {actual_metric}."
+        )
+
+
+class TableColumns(_SuccessfulMetricResult):
```
NathanFarmer: The names of these subclasses mirror the names of the metrics themselves. I think we need a different convention for the result classes if we even need the subclasses at all.

great_expectations/metrics/metric_results.py
```diff
@@ -0,0 +1,37 @@
...
+    value: _MetricResultValue
+
+
+class MetricErrorResult(MetricResult[dict[str, Union[int, dict, str]]]): ...
+
+
+class TableColumnsResult(MetricResult[list[str]]): ...
+
+
+class _ColumnType(BaseModel):
```
billdirks: I'd consider making this public since it's a field type in a real return type.

billdirks (approved):
LGTM, thanks!

PR #10938, opened 2025-02-13
[MAINTENANCE] Make `Metric.config` un-instantiable and excluded from auto-complete

NathanFarmer:
This makes `Metric.config` un-instantiable and removes it from auto-complete

### DIFF
--- a/great_expectations/metrics/metric.py
+++ b/great_expectations/metrics/metric.py
@@ -1,4 +1,5 @@
 import re
+from functools import cache
 from typing import ClassVar, Final
 
 from typing_extensions import dataclass_transform
@@ -78,7 +79,9 @@ class Metric(BaseModel, metaclass=MetaMetric):
 
     @property
     def config(self) -> MetricConfiguration:
-        return self._to_config()
+        return Metric._metric_to_config(
+            instance_class=self.__class__, metric_value_set=frozenset(self.dict().items())
+        )
 
     @staticmethod
     def _pascal_to_snake(class_name: str) -> str:
@@ -114,11 +117,15 @@ class Metric(BaseModel, metaclass=MetaMetric):
         # that a Domain exists in __bases__ should have been confirmed in MetaMetric.__new__
         raise MixinTypeError(cls.__name__, "Domain")
 
-    def _to_config(self) -> MetricConfiguration:
+    @staticmethod
+    @cache
+    def _metric_to_config(
+        instance_class: type["Metric"], metric_value_set: frozenset
+    ) -> MetricConfiguration:
         """Returns a MetricConfiguration instance for this Metric."""
-        instance_class = self.__class__
         metric_domain_kwargs = {}
         metric_value_kwargs = {}
+        metric_values = dict(metric_value_set)
         for base_type in instance_class.__bases__:
             if issubclass(base_type, Domain):
                 domain_fields = base_type.__fields__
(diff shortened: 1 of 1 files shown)


great_expectations/metrics/metric.py
```diff
@@ -118,31 +118,35 @@ def _get_metric_name(cls) -> str:
...
+    @staticmethod
+    @cache
+    def _to_config(
+        instance_class: type["Metric"], metric_value_set: frozenset
+    ) -> MetricConfiguration:
         """Returns a MetricConfiguration instance for this Metric."""
         metric_domain_kwargs = {}
         metric_value_kwargs = {}
-        for base_type in cls.__bases__:
+        metric_values = dict(metric_value_set)
```
billdirks, line 129: I don't understand this line of code. Usually you can't change a `set` to a `dict`. What is an example `metric_value_set` here?

billdirks (commented):
I don't understand the type of the `_to_config` method argument `metric_value_set`. Previously `model_values` was a dict and you haven't updated tests so I presume it is still being passed in as a dict? I don't understand why mypy isn't complaining though if that is the case. It seems like we would die if it was really a set?

NathanFarmer:
> I don't understand the type of the `_to_config` method argument `metric_value_set`. Previously `model_values` was a dict and you haven't updated tests so I presume it is still being passed in as a dict? I don't understand why mypy isn't complaining though if that is the case. It seems like we would die if it was really a set?

@billdirks It gets converted back to a dict in the method. It's required because a dict isn't hashable for the cache argument check.

great_expectations/metrics/metric.py
billdirks: I see, above it's always called with a frozenset of tuples. Can you update the signature; I'll leave a comment above.

great_expectations/metrics/metric.py
```diff
@@ -118,31 +118,35 @@ def _get_metric_name(cls) -> str:
         # that a Domain exists in __bases__ should have been confirmed in MetaMetric.__new__
         raise MixinTypeError(cls.__name__, "Domain")
 
-    @classmethod
-    def _to_config(cls, model_values: dict) -> MetricConfiguration:
+    @staticmethod
+    @cache
+    def _to_config(
+        instance_class: type["Metric"], metric_value_set: frozenset
```
billdirks:
```suggestion
        instance_class: type["Metric"], metric_value_set: frozenset[tuple]
```

billdirks:
@NathanFarmer I see it's always a frozenset of tuples. I added a suggstion about updating the type in the method signature. In general, `dict(frozen_set_var)` will not work eg `dict(frozenset([1,2,3])` will crash.

great_expectations/metrics/metric.py
```diff
@@ -73,17 +73,17 @@ def __new__(cls, *args, **kwargs):
...
-    def _set_config(cls, values) -> dict:
-        if "config" not in values or values["config"] is None:
-            values["config"] = cls._to_config(values)
-        return values
-
     @property
     def id(self) -> MetricConfigurationID:
         return self.config.id
 
+    @property
```
billdirks, line 80: You could cache here also? That makes it clear that is fixed per instance.

billdirks (approved):
LGTM, thanks! Added one comment about additional cacheing.

great_expectations/metrics/metric.py
NathanFarmer: That's what I tried at first, but you actually shouldn't do it according to [this ruff rule we have](https://docs.astral.sh/ruff/rules/cached-instance-method/). That's why I had to make it a `staticmethod` and pass the class instead of using `self`.

PR #10944, opened 2025-02-14
[FEATURE] BatchRowCount Metric

joshua-stauffer:
This PR introduces the first Batch level metric, known internally as table metrics.


great_expectations/metrics/domain.py
```diff
@@ -32,6 +39,11 @@ def __new__(cls, *args, **kwargs):
             raise AbstractClassInstantiationError(cls.__name__)
         return super().__new__(cls)
 
+    @classmethod
+    @override
+    def internal_name(cls):
```
NathanFarmer: Can we rename this to just `name`?

great_expectations/metrics/domain.py
```diff
@@ -32,6 +39,11 @@ def __new__(cls, *args, **kwargs):
...
+        return "map"
```
NathanFarmer: I think the name for this is actually `values` isn't it?

tests/metrics/test_metric.py
```diff
@@ -13,7 +13,7 @@
 TABLE = "my_table"
 COLUMN = "my_column"
 
-FULLY_QUALIFIED_METRIC_NAME = "column_values.above"
```
NathanFarmer: This is meant to look like [this](https://github.com/great-expectations/great_expectations/blob/6c3f8de396495746a12d655aecc31c62d40c75d1/great_expectations/expectations/metrics/column_map_metrics/column_values_between.py#L23)

NathanFarmer (changes requested):

great_expectations/metrics/domain.py
joshua-stauffer: metric names are the connection point to a metric resolver internally, so the classes we've renamed in this API need to be mapped back to the internal names when we generate the name used for the metric configuration.

tests/metrics/test_metric.py
joshua-stauffer: same point as above, the fully qualified metric name is meant for consumption by core internals

great_expectations/metrics/domain.py
joshua-stauffer: yeah, i started with that. My concern is that it would be surprising that the `name` field returns something apparently unrelated, which is an implementation detail.

great_expectations/metrics/domain.py
NathanFarmer: Where do you see a metric name with the word "map" in it?

great_expectations/metrics/domain.py
NathanFarmer: This class itself is an implementation detail right?

great_expectations/metrics/domain.py
joshua-stauffer: indeed 🤦‍♂️

great_expectations/metrics/domain.py
```diff
@@ -58,3 +70,26 @@ class ColumnValues(Values):
...
 
     column: NonEmptyString
+
+    @classmethod
+    @override
+    def internal_name(cls):
+        return "column_map"
+
+
+class Batch(Domain):
```
NathanFarmer, line 64: @billdirks What do you think about the `Batch` vs `Table` naming convention? Crazy idea...what if this class should be called `Table` and the class above it should be called `Batch` (the only attribute on that class is `batch_id`).

NathanFarmer: Or what if all of these inherit from the existing `Batch` class!

great_expectations/metrics/domain.py
joshua-stauffer: if `Batch` -> `Table` is the only change we need to make, maybe we should just leave the current magic string transformation in place

great_expectations/metrics/domain.py
joshua-stauffer: yes, definitely an implementation detail, but also will be available on the object we're giving to consumers, and we already have a property called `name`, which returns a different value. Looks like this actually raises a warning from pydantic: 
```
/Users/josh/code/gx/env/lib/python3.12/site-packages/pydantic/_internal/_fields.py:192: UserWarning: Field name "name" in "B" shadows an attribute in parent "A"
```

NathanFarmer: Oh right this will show up on a metric 🤦🏻 

I think if we need this it should be called `domain_name` then, but I would prefer to rename `Batch` to `Table` and keep the class name inference.

joshua-stauffer: we could also just move everything back into the internals of `metric`. I opted for this because i thought we would need to update more than one class string. On the other hand, I do think its generally clearer to have these Domain classes be responsible for knowing what their called, rather than the `Metric` base class.

NathanFarmer: I am considering adding a private `_domain: DomainT` generic attribute on `Metric` for `batch.compute_metrics()` and swapping the mixin for `class TableRowCount(Metric[Table]): ...`. We can add a name override attribute if we need it later. I think the names for multi-column metrics are not consistent, so we might want to change those.

joshua-stauffer: the benefit now is that the domain params are automatically available, right? how that work with a generic, would you just need to re-implement all the domain-specific fields on the metric implementation?

great_expectations/metrics/domain.py
joshua-stauffer: `Table` seems off to me, since this domain applies to lots of things which aren't tables. And even when it does apply to an actual table, row conditions and partitioners mean we're not actually validating the entire table. Seems like a more accurate description of the domain is "Whole Batch".

great_expectations/metrics/domain.py
NathanFarmer: I want to turn this
```
batch_row_count = BatchRowCount(
    batch_id=...
    table=...
    min_value=...
)
batch.compute_metrics(batch_row_count)
```
into this
```
batch_row_count = BatchRowCount(min_value=...)
batch.compute_metrics(batch_row_count)
```
I think we can infer the domain arguments from the `Batch` in `batch.compute_metrics()`.

NathanFarmer: We can't infer all domain arguments, but we can infer those first 2 I think.

billdirks: Since I'm doing childcare I can't read too closely at the moment. I prefer that we rename this from `internal_name` since, as an internal developer, it doesn't give me a sense of how this is used. I prefer `domain_name` or `domain_id` since the one usage I do see is `domain_identifier = base_type.internal_name()`.

great_expectations/metrics/domain.py
joshua-stauffer: i reverted this in [0f1e811](https://github.com/great-expectations/great_expectations/pull/10944/commits/0f1e811fa781257221f16c67b4410153328bae54)

tests/metrics/test_metric.py
joshua-stauffer: updated in https://github.com/great-expectations/great_expectations/pull/10944/commits/5db04b5f03629207643932425088ac2cce5dfd39

great_expectations/metrics/domain.py
joshua-stauffer: thanks all for the discussion. I've removed these `internal_name` methods for now in favor of the current magic string transformation in the `Metric` base class. That logic depends on our metric and domain class names following a convention that matches GX internals, which i find confusing. But if we want to change that, let's do it as a separate piece of work.

great_expectations/metrics/domain.py
```diff
@@ -58,3 +59,16 @@ class ColumnValues(Values):
...
+    """A domain type for metrics that compute over an entire batch.
+
+    The Batch domain type is used to define metrics that evaluate conditions or compute
+    values for an entire table. This class is intended to be used as a mixin
+    with the Metric class when defining a new Metric.
+    """
+
+    table: Optional[NonEmptyString] = None
```
billdirks, line 72: When is `table` `None`? When computing over a dataframe what is `table`?

tests/metrics/batch/test_batch.py
```diff
@@ -0,0 +1,93 @@
...
+            # todo: replace with batch.compute_metrics when available
+            metrics_calculator = MetricsCalculator(
+                execution_engine=batch.data.execution_engine,
+                show_progress_bars=False,
+            )
+            batch.data.execution_engine.batch_manager.load_batch_list(batch_list=[batch])
+            metric = BatchRowCount(
+                batch_id=batch.id,
+            )
+            result = BatchRowCountResult(
```
billdirks, line 47: This makes me want update the api so that I can do:
```
result = metric.compute()
```

The compute metric would look like:
```
def compute(self) -> T:
    return T(id=self.id, value=metrics_calculator.get_metric(metric=self.config))
```

where T in this case would be `BatchRowCountResult`. Not exactly sure of all the mechanics but maybe we could make a metric be a generic that takes a class bounded by MetricResult. Actually my questions are:

1. Will this complication will go away when we implement `batch.compute_metric(BatchRowCount())`? 
2. Is `batch.compute_metric(BatchRowCount())` what the api will look like since the domain kwargs will be going away?

Tagging @NathanFarmer also.

billdirks (approved):
I think the issue I pointed out in comments will be resolved with an upcoming PR. Let me know if I'm misunderstanding and we can chat more tomorrow. Going to give it the 🚢 since, while we need to resolve these, we can iterated on this and overall it lgtm.

tests/metrics/batch/test_batch.py
NathanFarmer: @billdirks I was just running up against this issue

> but maybe we could make a metric be a generic that takes a class bounded by MetricResult

in the `batch.compute_metrics()` PR and I was going down the road of doing that exact thing.

> Will this complication will go away when we implement batch.compute_metric(BatchRowCount())

Right now the goal is to make it go away without implementing `metric.compute()`. We could decide to add it, but the `batch.compute_metrics()` API allows for 1 metric to be passed and 1 result to be returned right now.

It's not so straight forward though, because it turns out we have metric dependencies getting computed and I think we need to make a decision about whether to also return those. When you ask for `BatchRowCount`, multiple metrics always get computed.

> Is batch.compute_metric(BatchRowCount()) what the api will look like since the domain kwargs will be going away?

That's what the API will look like for this one metric, but I also realized all domain kwargs cannot be inferred from `Batch` (e.g. `column`).

NathanFarmer (approved):

great_expectations/metrics/domain.py
joshua-stauffer: i know this is weird, but i think the better question is, when is table _not_ `None`? I tested against a couple backends, and it appears to be unused. It's also causing the huge amount of logs reported in `gx-398`. I'm working on removing it altogether in https://github.com/great-expectations/great_expectations/pull/10951.

tests/metrics/batch/test_batch.py
joshua-stauffer: Yeah, i was also considering if/how we should return all of the generated metrics, rather than just the requested metric. I think we should consider the supporting metrics to be an implementation detail, because in the future we would like the flexibility to implement backend-specific execution engines which may or may not require the same set of supporting metrics.

The best solution I see is to store the supporting metrics, so the user can request them if needed without needing to recompute anything.

NathanFarmer: The more I think about it, the more I agree. We should only return the metrics that were requested. The fact that other metrics are computed as a result is an implementation detail. It can be thought of as an optimization if related metrics are requested.

great_expectations/metrics/domain.py
NathanFarmer: I actually had to do this same thing in my PR due to testing with pandas. If we can remove it altogether that would be great! I'm not sure that we can though because I haven't looked at all of the legacy metric implementations.

PR #10950, opened 2025-02-19
[FEATURE] `Batch.compute_metrics()`

NathanFarmer:
- Add `Batch.compute_metrics()` method
- Add `Metric` generic type for defining result type
- Add `MetaMetric` registry for `Metric` type lookups
- Improve `MetricErrorResult` generic type and update references
- Update existing tests to use `Batch.compute_metrics()`
---------------


great_expectations/metrics/column_values/between.py
```diff
@@ -1,11 +1,20 @@
...
 from great_expectations.metrics.domain import ColumnValues
 from great_expectations.metrics.metric import Metric
+from great_expectations.metrics.metric_results import MetricResult
+
+
+class ColumnValuesBetweenResult(MetricResult[tuple[pd.Series, IDDict, dict]]): ...
+
 
+class ColumnValuesBetween(Metric[ColumnValuesBetweenResult], ColumnValues):
+    condition_metric = True
```
NathanFarmer: This was required because in the legacy system the name for this metric gets transformed from `column_values.between` into `column_values.between.condition` 😮‍💨 

Maybe instead of doing this, the best approach is to add the `name` class variable to every concrete metric and dispense with the inference stuff.

joshua-stauffer: yes, i think a more explicit solution will be easier to maintain long term.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -1243,3 +1249,70 @@ def _create_validator(self, *, result_format: ResultFormatUnion) -> V1Validator:
...
+    def _metrics_calculator_result_to_metric_result(
+        metrics_calculator_result: tuple[_MetricsDict, _AbortedMetricsInfoDict],
+    ) -> MetricResult | list[MetricResult]:
+        raw_metrics = metrics_calculator_result[0]
+        raw_metrics_errors = metrics_calculator_result[1]
+        if len(raw_metrics) + len(raw_metrics_errors) == 1:
+            metric_result_dict = raw_metrics or raw_metrics_errors
+            metric_configuration_id = next(iter(metric_result_dict))
+            value = metric_result_dict[metric_configuration_id]
+            MetricResultType = Batch._get_metric_result_type_from_metric_name(
```
joshua-stauffer: style comment: i'd prefer to not access a private member of the class. Can we make this a class method? i believe that's also in line with the google style guide recommendation to avoid staticmethod.

great_expectations/metrics/metric_results.py
```diff
@@ -10,6 +10,9 @@ class MetricResult(GenericModel, Generic[_MetricResultValue]):
     id: MetricConfigurationID
     value: _MetricResultValue
 
+    class Config:
+        arbitrary_types_allowed = True
```
NathanFarmer, line 18: I had to add this once I added a `MetricResult` that returns `pd.Series`. It still does an `isinstance` check, just not anything else (which I guess pydantic does with built-in types).

great_expectations/datasource/fluent/interfaces.py
NathanFarmer: I may not be up to date about our recent Core conventions. I would typically do this to indicate that this is for internal use only, but I guess we do have `@public_api` for that.

great_expectations/metrics/column_values/between.py
NathanFarmer: I made the change to explicit `name` here: https://github.com/great-expectations/great_expectations/pull/10953

great_expectations/metrics/__init__.py
```diff
@@ -1 +1,3 @@
+from .batch.row_count import BatchRowCount
 from .column_values.between import ColumnValuesBetween
+from .metric import Metric
```
NathanFarmer, line 3: Since it looks like in some cases users will be needing the type hint `list[Metric]`, I simplified the `Metric` import.

great_expectations/metrics/batch/row_count.py
```diff
@@ -3,8 +3,8 @@
 from great_expectations.metrics.metric_results import MetricResult
```
NathanFarmer, line 5: I renamed this file, because I think we will be adding much more logic to each of these classes over time. Happy to discuss if we want to do that now.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -1243,3 +1257,123 @@ def _create_validator(self, *, result_format: ResultFormatUnion) -> V1Validator:
...
+            inferred by the static type checker as list[Domain]. You can work around
+            this by adding an explicit annotation (e.g. metrics: list[Metric] = ...)
+            https://github.com/python/mypy/issues/18712
+        """
+        if isinstance(metrics, Metric):
+            metrics = [metrics]
+
+        execution_engine = self.data.execution_engine
+        execution_engine.batch_manager.load_batch_list(batch_list=[self])
+        metrics_calculator = MetricsCalculator(execution_engine=execution_engine)
```
billdirks: This preamble necessary to setup the metrics calculator is used 2 other places in this class and takes real domain knowledge to understand that one needs to do this. Can you put this in a private helper method, eg `_get_metrics_calculator()` and use it here and the other 2 places.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -1243,3 +1257,123 @@ def _create_validator(self, *, result_format: ResultFormatUnion) -> V1Validator:
...
+                    metric_name=metric_name,
+                    metric_types=metric_types,
+                )
+                metric_results.append(
+                    MetricResultType(
+                        id=raw_metrics[metric_name][0],
+                        value=raw_metrics[metric_name][1],
+                    )
+                )
+            if metric_name in raw_metrics_errors:
```
billdirks: Can a metric return both a result and an error? If not, I'd make this an `elif`.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -1243,3 +1257,123 @@ def _create_validator(self, *, result_format: ResultFormatUnion) -> V1Validator:
...
+        """Given a metric name and a list of Metric types gets the
+        MetricResult type for that Metric. Returns the base
+        MetricResult class if no MetricResult subclass is defined."""
+        for metric_type in metric_types:
+            if metric_type.name == metric_name:
+                return metric_type.get_metric_result_type()
+        return MetricResult
+
+    @classmethod
+    def get_all_metric_classes(cls) -> list[type[Metric]]:
```
billdirks: I'd prefer a registry over this type of introspection. Advantages of a registry:
* Contributors can register their metrics and don't have to add their metrics to this package. Note, we could get around this by using `Metric.__subclasses__()`. If we want to go this introspection route, I would prefer that to this. It also allows us to get rid of `METRICS_PACKAGE`.
* Registry lookup is faster since we don't need to do this traversal every time we call this method.

great_expectations/datasource/fluent/interfaces.py
NathanFarmer: I'm pretty sure from reading [this try/except](https://github.com/great-expectations/great_expectations/blob/b0ca52840f6438f33259db447528cb2ebe78dc9c/great_expectations/validator/validation_graph.py#L275) the answer is no, it's one or the other. It's not the easiest to interpret given it all being in a while loop, and the successful metrics graph is resolved and updated in-place. I will update to `elif`.

tests/datasource/fluent/test_batch.py
```diff
@@ -290,3 +294,57 @@ def test_batch_validate_expectation_suite_does_not_persist_a_batch_definition(
...
     assert len(asset.batch_definitions) == 0
+
+
+@pytest.mark.filesystem
+def test_batch_compute_metrics_single_metric_success(
+    pandas_setup: Tuple[AbstractDataContext, Batch],
+):
+    _, batch = pandas_setup
+    metric = ColumnValuesBetween(
+        batch_id=batch.id,
```
billdirks, line 305: I'm wondering if we need to pass in `batch.id` since we could infer that from the `batch` when we call `batch.compute_metrics(...)`.

billdirks (commented):
Thanks for doing this! I added 1 more comment around the api.

tests/datasource/fluent/test_batch.py
NathanFarmer: Yes, I have been planning to take `batch_id` private, but I am going to do it in it's own PR.

great_expectations/datasource/fluent/interfaces.py
NathanFarmer: Thanks for the tip. I added a runtime registry to `MetaMetric.__new__`.

great_expectations/metrics/metric.py
```diff
@@ -20,34 +21,62 @@ def __init__(self, class_name: str, mixin_superclass_name: str) -> None:
...
+                metric_name = attrs["name"]
+            except KeyError:
+                raise MissingAttributeError(name, "name")
+            MetaMetric._registry[metric_name] = register_cls
+        return register_cls
+
+    @classmethod
+    def get_registered_metric_class_from_metric_name(cls, metric_name: str) -> type["Metric"]:
+        """Returns the registered Metric class for a given metric name."""
+        return cls._registry[metric_name]
```
billdirks: This can throw a `KeyError` which might not produce a good error message. Maybe try/catch to produce a better error message, eg: `No metric named metric_name`.

billdirks (approved):
Thank you! I left 1 small comment about improving an error message. LGTM.

PR #10951, opened 2025-02-19
[BUGFIX] remove unused domain key

joshua-stauffer:
The `table` domain key is unused, and causes many log messages when attempting to resolve metrics.


tyler-hoffman (approved):
LGTM as long as we're passing tests with this change!

NathanFarmer (approved):
🚀 🚀 🚀

PR #10953, opened 2025-02-19
[MAINTENANCE] Require `Metric.name` instead of using `name` inference

NathanFarmer:
The inference step is going to be too hard to maintain due to inconsistencies in naming conventions in the legacy system

### DIFF
--- a/great_expectations/metrics/metric.py
+++ b/great_expectations/metrics/metric.py
@@ -20,11 +20,6 @@ class MixinTypeError(TypeError):
         )
 
 
-class MissingAttributeError(AttributeError):
-    def __init__(self, class_name: str, attribute_name: str) -> None:
-        super().__init__(f"`{class_name}` must define the `{attribute_name}` attribute.")
-
-
 @dataclass_transform()
 class MetaMetric(ModelMetaclass):
     def __new__(cls, name, bases, attrs):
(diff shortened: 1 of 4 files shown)


great_expectations/metrics/metric.py
```diff
@@ -30,6 +34,8 @@ def __new__(cls, name, bases, attrs):
             or not any(issubclass(base_type, Domain) for base_type in bases)
         ):
             raise MixinTypeError(name, "Domain")
+        if name != "Metric" and not attrs.get("name"):
+            raise MissingAttributeError(name, "name")
```
joshua-stauffer: is this necessary? i would expect pydantic to handle this for us.

great_expectations/metrics/metric.py
```diff
@@ -84,41 +89,6 @@ def config(self) -> MetricConfiguration:
...
-                    [
-                        translated_domain_name,
-                        metric_class_snake_case.replace(domain_class_snake_case, "").strip("_"),
-                    ]
-                )
-
-        # this should never be reached
-        # that a Domain exists in __bases__ should have been confirmed in MetaMetric.__new__
-        raise MixinTypeError(cls.__name__, "Domain")
-
```
joshua-stauffer, line 121: 🕺

great_expectations/metrics/metric.py
joshua-stauffer: i see that it doesn't. However, I think we can still remove this. We don't expect users to encounter this error, just developers who are working on the library.

NathanFarmer: This is an open source library and I would hope one day we have external contributors defining metrics. Even if only internal developers encounter it, I don't see any harm in this.

NathanFarmer: Ok you talked me into removing this. I think we would prefer in the future to not have to define this. It's just required because of the legacy system. Not raising an error in `Metric` makes it clear that only use it because we have to.

joshua-stauffer (approved):
looks great, thanks!

PR #10959, opened 2025-02-21
[FEATURE] `ColumnValuesNonNull` and `ColumnValuesNonNullCount` metrics

NathanFarmer:
- Implement `ColumnValuesNonNull` and `ColumnValuesNonNullCount` metrics
- Add special case for return type of Metrics with names that end in `condition` so they don't also return domain and value kwargs
- Use new backend-specific testing pattern established by Expectation integration testing framework
- Move integration test files into `tests/integration/metrics` like Expectation integration testing framework
------------


great_expectations/metrics/metric_name.py
```diff
@@ -0,0 +1,11 @@
+from enum import Enum
+
+
+class DomainName(str, Enum):
+    BATCH = "table"
+    COLUMN_VALUES = "column_values"
+
+
+class MetricNameSuffix(str, Enum):
+    CONDITION = "condition"
```
NathanFarmer, line 10: I added this because `condition` gets special treatment. Hopefully we won't find any more special cases like that and we can pare this file down.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -1324,10 +1325,14 @@ def metrics_calculator_result_to_metric_result(
             if metric_name in raw_metrics:
                 MetricType = MetaMetric.get_registered_metric_class_from_metric_name(metric_name)
                 MetricResultType = MetricType.get_metric_result_type()
+                metric_configuration_id = raw_metrics[metric_name][0]
+                value = raw_metrics[metric_name][1]
+                if metric_name.endswith(MetricNameSuffix.CONDITION) and isinstance(value, tuple):
+                    value = value[0]
```
joshua-stauffer: what are we discarding here?

NathanFarmer: Good question! I added this comment:
```
  # "condition" metrics return the domain and value kwargs
  # we just want the value, which is the first item in the tuple
```

great_expectations/metrics/metric_results.py
```diff
@@ -27,6 +31,16 @@ class MetricErrorResultValue(TypedDict):
...
 
 
+if pyspark and sqlalchemy:
+    ConditionValues = Union[pd.Series, pyspark.sql.Column, BinaryExpression]
+elif pyspark:
+    ConditionValues = Union[pd.Series, pyspark.sql.Column]  # type: ignore[misc]  # can't find type to satisfy "<typing special form>"
+elif sqlalchemy:
+    ConditionValues = Union[pd.Series, BinaryExpression]  # type: ignore[misc]  # can't find type to satisfy "<typing special form>"
+else:
+    ConditionValues = pd.Series  # type: ignore[misc]  # can't find type to satisfy "<typing special form>"
```
joshua-stauffer: I strongly suggest we simplify this to `Union[pd.Series, pyspark.sql.Column, BinaryExpression]`. As is, i think this is a fancy `Any`, at least from the perspective of static type checkers. The error i get from mypy locally is useful:
```
Cannot assign multiple types to name "ConditionValues" without an explicit "Type[...]" annotation
```

joshua-stauffer: so if conditional types like this are required in order to get pydantic to work correctly, we need to preface it with something like `ConditionValues: Union[pd.Series, pyspark.sql.Column, BinaryExpression]`, so that static type checkers have a single type definition to use.

NathanFarmer: What would that code look like? I wasted a lot of time on this. The issue is that the user might not have the right dependencies installed to do the `isinstance` check when instantiating the result object.

joshua-stauffer: ah right, and including `NotImportedError` in the union causes a runtime error?

NathanFarmer: Yeah exactly. And even before running the tests, in some CI stages I was seeing import errors.

joshua-stauffer: i've also lost a significant amount of time on this in the past. I wonder if we could add a new compatibility type that is either the type or `None`, so we can use it in unions. The type would still be wrong, since `None` would be included in some situations, but at least it would read correctly, and would be usable for static type checkers.

NathanFarmer: I think that used to be the old pattern before we introduced `NotImportedError` 🤣

joshua-stauffer: yeah. i can understand why we arrived at the current solution for runtime objects, but its a disaster in terms of typing.

great_expectations/datasource/fluent/interfaces.py
joshua-stauffer: great, thank you!

great_expectations/metrics/metric_results.py
NathanFarmer: I discussed this with Bill and we came up with a better solution that fixes the issue with losing the static typing information. Let me know what you think!

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -1324,10 +1325,16 @@ def metrics_calculator_result_to_metric_result(
             if metric_name in raw_metrics:
                 MetricType = MetaMetric.get_registered_metric_class_from_metric_name(metric_name)
                 MetricResultType = MetricType.get_metric_result_type()
+                metric_configuration_id = raw_metrics[metric_name][0]
+                value = raw_metrics[metric_name][1]
+                # "condition" metrics return the domain and value kwargs
+                # we just want the value, which is the first item in the tuple
+                if metric_name.endswith(MetricNameSuffix.CONDITION) and isinstance(value, tuple):
+                    value = value[0]
```
billdirks: nit: I'd move this logic into a helper method, eg `parse_metric_value`.

I make make an equivalent one for `parse_metric_config_id`.

These seem like lower level details than this method reads.

great_expectations/metrics/__init__.py
```diff
@@ -1,3 +1,3 @@
 from .batch.row_count import BatchRowCount
-from .column_values.between import ColumnValuesBetween
+from .column_values.non_null import ColumnValuesNonNull, ColumnValuesNonNullCount
```
billdirks, line 2: Why are we deleting between? Isn't this a new metric in addition to between?

great_expectations/metrics/metric_results.py
```diff
@@ -27,6 +33,56 @@ class MetricErrorResultValue(TypedDict):
...
+        except (ImportError, AttributeError):
+            pass
+
+        raise ConditionValuesValueError(type(value))
+
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate_value_type
+
+    def equals(self, other: pd.Series) -> bool:
```
billdirks: Are `equals` and `compares` methods and how they only support certain operations a result of how these metrics are computed now?

billdirks (commented):
Thanks for doing this! Overall it looks good. Do we need to delete between? That is, can we have the null metrics in addition to between?

great_expectations/metrics/__init__.py
NathanFarmer: Between was there to demonstrate how it would all look, but then I learned we needed to add this (different) metric to support another epic. Between was never tested, which is really the bulk of the remaining work at this point.

great_expectations/metrics/metric_results.py
NathanFarmer: I really did this to make the type checker happy in my tests. We could just remove these equality-checking methods and ignore the type errors.

great_expectations/datasource/fluent/interfaces.py
NathanFarmer: I think I understand the ask here. I refactored some of this into 2 helper methods and added `NamedTuples` to make it clearer what's going on.

great_expectations/metrics/metric_results.py
NathanFarmer: I removed these and added an assertion to the tests to handle it

great_expectations/metrics/metric_results.py
```diff
@@ -27,6 +33,44 @@ class MetricErrorResultValue(TypedDict):
...
+            if isinstance(value, BinaryExpression):
+                return value
+        except (ImportError, AttributeError):
+            pass
+
+        raise ConditionValuesValueError(type(value))
+
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate_value_type
```
joshua-stauffer, line 71: 👏 nice workaround!

joshua-stauffer (approved):
looks good, thanks for fixing that gnarly type

tests/integration/metrics/column_values/test_non_null.py
```diff
@@ -0,0 +1,131 @@
...
+]
+
+SQL_DATA_SOURCES: Sequence[DataSourceTestConfig] = [
+    BigQueryDatasourceTestConfig(),
+    DatabricksDatasourceTestConfig(),
+    MSSQLDatasourceTestConfig(),
+    PostgreSQLDatasourceTestConfig(),
+    SnowflakeDatasourceTestConfig(),
+    SqliteDatasourceTestConfig(),
+]
```
joshua-stauffer, line 50: i suggest we move these to a common `conftest` module so we can test every metric against a standard set of backends.

We should probably restrict this list to only test against data sources which are officially supported by core. As far as I know, MSSQL is not on that list.

tests/integration/metrics/column_values/test_non_null.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class TestColumnValuesNonNull:
+    @parameterize_batch_for_data_sources(
+        data_source_configs=PANDAS_DATA_SOURCES,
+        data=DATA_FRAME,
+    )
+    def test_success_pandas(self, batch_for_datasource) -> None:
```
joshua-stauffer, line 58: can we type these test params?

PR #10960, opened 2025-02-21
[BUGFIX] Validation Definition factory add_or_update doesnt allow updating batch definition

joshua-stauffer:
`ValidationDefinitionFactory.add_or_update` overwrites the ID of the incoming batch definition with the ID of the existing batch definition, and then calls `BatchDefinition.save`. This is ok if the BatchDefinition belongs to the same data source, but in the case that it does not, this errors unexpectedly. 

This PR updates the `ValidationDefinitionFactory` to simply call `BatchDefinition.save`, then update the `ValidationDefinition`


tyler-hoffman (changes requested):
Looks totally legit, but I think this should have an accompanying test. LMK if that's prohibitively difficult

tests/core/factory/test_validation_definition_factory.py
```diff
@@ -634,6 +639,105 @@ def _test_overwrite_existing_validation(self, context: AbstractDataContext):
         assert updated_vd.id == existing_vd.id
         context.validation_definitions.get(vd_name)
 
+    @pytest.mark.filesystem
+    @pytest.mark.xfail(
```
tyler-hoffman, line 643: Probably good to reference `GX-481` in the reason for xfailing these.

tests/core/factory/test_validation_definition_factory.py
```diff
@@ -634,6 +639,105 @@ def _test_overwrite_existing_validation(self, context: AbstractDataContext):
...
+            name=vd_name,
+            data=new_batch_def,
+            suite=suite,
+        )
+        updated_vd = context.validation_definitions.add_or_update(validation=new_vd)
+
+        # assert
+        assert updated_vd.suite.id == existing_vd.suite.id
+        assert (
+            updated_vd.data.id != existing_vd.data.id
```
tyler-hoffman, line 735: I'd probably also assert that the id is non-None

tyler-hoffman (approved):
Left minor thoughts, but this looks great.

tests/core/factory/test_validation_definition_factory.py
joshua-stauffer: [673a28e](https://github.com/great-expectations/great_expectations/pull/10960/commits/673a28e99c15f453ee82856d0cb16bd493ac9292)

tests/core/factory/test_validation_definition_factory.py
joshua-stauffer: [9af4ef1](https://github.com/great-expectations/great_expectations/pull/10960/commits/9af4ef1fc2dd3f9c5d0754cf85e02aeff476df43)

PR #10966, opened 2025-02-25
[MAINTENANCE] Remove `Domain` mixin from Metrics API

NathanFarmer:
Remove the `Domain` mixin from the metrics API in favor of `Metric` subclasses


great_expectations/metrics/metric.py
```diff
@@ -40,20 +46,15 @@ class MetaMetric(ModelMetaclass):
...
-                issubclass(base_type, Domain) for base_type in bases
-            ):
-                raise MixinTypeError(name, "Domain")
-            try:
-                metric_name = attrs["name"]
-            except KeyError:
-                raise MissingAttributeError(name, "name")
-            MetaMetric._registry[metric_name] = register_cls
+            metric_name = attrs.get("name")
+            # Some subclasses of metric may not have a name
```
billdirks, line 54: I'm curious why we'd define a metric and then not register it. What's the use case?

billdirks (approved):
Thanks! This seems simpler.

great_expectations/metrics/metric.py
NathanFarmer: New subclasses like BatchMetric, ColumnMetric, etc

PR #10969, opened 2025-02-25
[FEATURE] ColumnPairValueInSet metric

joshua-stauffer:
This PR adds and tests the first ColumnPairValue metric, ColumnPairValuesInSetUnexpectedCount.


great_expectations/metrics/metric.py
```diff
@@ -118,7 +117,7 @@ def id(self) -> MetricConfigurationID:
     def config(self) -> MetricConfiguration:
         return Metric._to_config(
             instance_class=self.__class__,
-            metric_value_set=frozenset(self.dict().items()),
+            metric_value_set=list(self.dict().items()),
```
joshua-stauffer, line 130: this change was required because the metric value from users is a set

billdirks: Could you `frozenset(list(self.dict().items()))`? I'm not sure it actually matters. I also remove the cache decorator in [my PR](https://github.com/great-expectations/great_expectations/pull/10971) which is a replacement for an previous equivalent PR that I didn't merge.

tests/metrics/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+    DataSourceTestConfig,
+    PandasDataFrameDatasourceTestConfig,
+    PandasFilesystemCsvDatasourceTestConfig,
+    PostgreSQLDatasourceTestConfig,
+    SnowflakeDatasourceTestConfig,
+    SparkFilesystemCsvDatasourceTestConfig,
+    SqliteDatasourceTestConfig,
+)
+
+PANDAS_DATA_SOURCES: list[DataSourceTestConfig] = [
```
billdirks, line 13: Thanks for these constants!

tests/integration/metrics/column_pair_values/test_in_set.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+        result = batch.compute_metrics(metric)
+        assert isinstance(result, ColumnPairValuesInSetUnexpectedCountResult)
+        assert result.value == FAILURE_RESULT_COUNT
+
+    @pytest.mark.parametrize(
+        "ignore_row_if,unexpected_count",
+        [
+            ("either_value_is_missing", 1),  # this is expected
+            ("both_values_are_missing", 1),  # shouldn't this be 3?
```
billdirks: I don't understand these questions.

billdirks (commented):
Functional code LGTM! I didn't look at the tests really and would like to.

tests/integration/metrics/column_pair_values/test_in_set.py
joshua-stauffer: yeah, i was seeing some unexpected behavior but hadn't yet pinned it down. The bug is that SQL data sources treat the `neither` condition the same as the `both_values_are_missing` condition. I reworked these tests to make the divergent behavior clear.

great_expectations/metrics/metric.py
joshua-stauffer: no - the issue is that we're asking users to pass in a `set` as a value input to the metric. if we really need to keep this caching behavior, we would need to mutate that to be a frozenset.

NathanFarmer: I think the problem with removing the caching is that each time `Metric.config` is called the `MetricConfiguration` will have a new id. Can we save the `MetricConfiguration` to a private `_config` and return that if it is available?

NathanFarmer (dismissed):

great_expectations/metrics/metric.py
joshua-stauffer: why would it have a new ID? are we expecting a metric to be mutated? i thought that just happens in the metrics internals, not in our API classes.

joshua-stauffer: if that is a concern, it would be very useful to have that captured in a test

NathanFarmer: Ok, I re-read the `MetricConfiguration` code and now see that the id should be the same. I tried and failed to test the caching. We should be able to easily test that the id doesn't change though. If you are going to remove the cache can you please add a test to `tests/metrics/test_metric.py` `TestMetricConfig` to assert that calling twice results in the same id?

billdirks: @NathanFarmer, I removed the `id` field from this class in my PR: https://github.com/great-expectations/great_expectations/pull/10971/files#diff-c1f77a9f3411e39ae24afca33308d6b78ea4b4d636f7db0b5424c395bfdaed95L114 since it wasn't being used anywhere. Is this problematic?

tests/integration/metrics/column_pair_values/test_in_set.py
```diff
@@ -0,0 +1,149 @@
...
+            ignore_row_if=ignore_row_if,
+        )
+        result = batch.compute_metrics(metric)
+        assert result.value == unexpected_count
+
+    @pytest.mark.parametrize(
+        "ignore_row_if,unexpected_count",
+        [
+            ("either_value_is_missing", 2),
+            ("both_values_are_missing", 6),
```
billdirks: Are these 2 cases buggy? I would expect this to be the same as pandas. If so, could you change this to:
```
    @pytest.mark.parametrize(
        "ignore_row_if,unexpected_count",
        [
            pytest.param("either_value_is_missing", 1,  marks.pytest.mark.xfail),  # returns 2
            pytest.param("both_values_are_missing", 3, marks.pytest.mark.xfail),  # returns 6
            pytest.param("neither", 6),
        ],
    )
```
If this is caused by expected spark backend behavior, can you keep the test as is but add a comment you do for sql:
```
("either_value_is_missing", 2), # Spark ...
```

billdirks (commented):
I left a comment in tests that should be addressed but otherwise LGTM. Thanks!

great_expectations/metrics/metric.py
NathanFarmer: No, I guess this isn't a problem for the consumers of this API. 🚢

great_expectations/metrics/column_pair_values/in_set.py
```diff
@@ -0,0 +1,15 @@
...
+from great_expectations.metrics.metric_results import MetricResult
+
+
+class ColumnPairValuesInSetUnexpectedCountResult(MetricResult[int]): ...
+
+
+class ColumnPairValuesInSetUnexpectedCount(
+    Metric[ColumnPairValuesInSetUnexpectedCountResult], ColumnPairValues
+):
+    name = "column_pair_values.in_set.unexpected_count"
```
NathanFarmer: Can you add the string `unexpected_count` to the enum `MetrixNameSuffix` and use it here? We have one (`condition`) that is a special case and I have been trying to keep an inventory of them just in case they have some meaning. I think they get used more generally in the legacy system metric "provider" superclasses.

great_expectations/metrics/domain.py
```diff
@@ -59,6 +59,21 @@ class ColumnValues(Values):
...
+    with the Metric class when defining a new Metric.
+    """
+
+    column_A: NonEmptyString
+    column_B: NonEmptyString
+    ignore_row_if: Literal["both_values_are_missing", "either_value_is_missing", "neither"] = (
+        "both_values_are_missing"
+    )
+
+
```
NathanFarmer: You will need to add `column_A`, `column_B`, and `ignore_row_if` to
```
class ColumnPairMetric(BatchMetric[_MetricResult]): ...
```
in `great_expectations/metrics/column_pair/column_pair.py` and add the class to the `great_expectations/metrics/column_pair` package to shorten the import.

great_expectations/metrics/column_pair_values/in_set.py
```diff
@@ -0,0 +1,15 @@
...
+from great_expectations.metrics.metric import Metric
+from great_expectations.metrics.metric_results import MetricResult
+
+
+class ColumnPairValuesInSetUnexpectedCountResult(MetricResult[int]): ...
+
+
+class ColumnPairValuesInSetUnexpectedCount(
+    Metric[ColumnPairValuesInSetUnexpectedCountResult], ColumnPairValues
+):
```
NathanFarmer: This should become
```
class ColumnPairValuesInSetUnexpectedCount(ColumnPairMetric[ColumnPairValuesInSetUnexpectedCountResult]):
```
and this file should be named `great_expectations/metrics/column_pair/values_in_set.py` and added to the `great_expectations/metrics` package.

NathanFarmer (changes requested):
Sorry, my cleanup PR broke this one. I hope my comments can get you pretty far.

tests/integration/metrics/column_pair_values/test_in_set.py
joshua-stauffer: sure - they seem to be buggy, but i'm not sure what's causing it. i'll add the xfails 👍

great_expectations/metrics/column_pair_values/in_set.py
joshua-stauffer: ok - i've updated this, but i don't totally understand the utility. imo, this makes the code significantly less readable, so if there isn't an immediate benefit here, i suggest we just revert to plain strings.

tests/integration/metrics/batch/test_row_count.py
```diff
@@ -1,39 +1,8 @@
...
-
-SQL_DATA_SOURCES: Sequence[DataSourceTestConfig] = [
-    BigQueryDatasourceTestConfig(),
-    DatabricksDatasourceTestConfig(),
-    MSSQLDatasourceTestConfig(),
-    PostgreSQLDatasourceTestConfig(),
-    SnowflakeDatasourceTestConfig(),
-    SqliteDatasourceTestConfig(),
-]
+from tests.metrics.conftest import PANDAS_DATA_SOURCES, SPARK_DATA_SOURCES, SQL_DATA_SOURCES
```
NathanFarmer, line 5: Thanks for doing this!

tests/integration/metrics/column_pair_values/test_in_set.py
```diff
@@ -0,0 +1,159 @@
...
+        result = batch.compute_metrics(metric)
+        assert result.value == unexpected_count
+
+    @pytest.mark.parametrize(
+        "ignore_row_if,unexpected_count",
+        [
+            pytest.param(
+                "either_value_is_missing",
+                1,
+                marks=pytest.mark.xfail(reason="returns 2", strict=True),
```
NathanFarmer, line 98: I suspect this has something to do with the wrong type being inferred for Spark and if you pass `column_types` it will pass. I'm not going to block on this though.

NathanFarmer (approved):
Thanks for doing this!

tests/integration/metrics/column_pair_values/test_in_set.py
NathanFarmer: I'm guessing Spark is inferring the wrong column type due to the `None`s, but I'm not 100% sure. Pretty sure I've encountered this in the past.

tests/integration/metrics/column_pair_values/test_in_set.py
joshua-stauffer: yeah, i wondered the same thing - i have a similar issue on the remaining PR, so will try it there, and update this if it works

tests/integration/metrics/column_pair_values/test_in_set.py
```diff
@@ -0,0 +1,159 @@
+import pandas as pd
```
NathanFarmer, line 1: Can you rename this directory to `column_pair` and the file to `test_values_in_set.py`?

NathanFarmer (changes requested):

tests/integration/metrics/column_pair_values/test_in_set.py
joshua-stauffer: 😬

joshua-stauffer: yeah, can make a followup PR

PR #10973, opened 2025-02-26
[FEATURE] MultiColumnValues metric

joshua-stauffer:
This PR adds the first multi column values metric, `MultiColumnSumEqualUnexpectedCount`


tests/integration/metrics/multicolumn_values/test_sum_equal.py
```diff
@@ -0,0 +1,48 @@
...
+COL_A_WITH_NULLS = "A_WITH_NULLS"
+COL_B_WITH_NULLS = "B_WITH_NULLS"
+DATA_FRAME = pd.DataFrame(
+    {
+        COL_A: [1, 2, 3, 4],
+        COL_B: [4, 3, 2, 1],
+    },
+)
+SUCCESS_SUM_TOTAL = 5
+DATA_FRAME_WITH_NULLS = pd.DataFrame(
```
billdirks: This isn't used. Maybe there is a test commit that didn't get pushed up for testing failures? Also it would be nice to test that we catch it when 2 rows don't equal to 5 but neither is `None`.

billdirks (commented):
Thanks for doing this. I'd like to see a test for 2 columns not adding up to the target. Also, the None entries aren't being used and it could nice to see if/when they are ignored but maybe that is redundant with the tests in the other metric since they hopefully use the same code path.

tests/integration/metrics/multicolumn_values/test_sum_equal.py
joshua-stauffer: yes - i hadn't written the `ignore_row_if` tests yet. While it shares a name with the other `ignore_row_if` condition, it has completely different values, and a separate implementation, so i want to test it.

tests/integration/metrics/multicolumn_values/test_sum_equal.py
```diff
@@ -0,0 +1,147 @@
...
+        "ignore_row_if,unexpected_count",
+        [
+            pytest.param(
+                "any_value_is_missing", 1, marks=pytest.mark.xfail(reason="returns 6", strict=True)
+            ),
+            pytest.param(
+                "all_values_are_missing",
+                3,
+                marks=pytest.mark.xfail(reason="returns 6", strict=True),
+            ),
```
NathanFarmer: More examples where I think passing `column_types` might help

tests/integration/metrics/multicolumn_values/test_sum_equal.py
```diff
@@ -0,0 +1,147 @@
...
+    @pytest.mark.parametrize(
+        "ignore_row_if,unexpected_count",
+        [
+            pytest.param("any_value_is_missing", 1),
+            pytest.param(
+                "all_values_are_missing",
+                3,
+                marks=pytest.mark.xfail(reason="returns 1", strict=True),
+            ),
+            pytest.param("never", 6, marks=pytest.mark.xfail(reason="returns 1", strict=True)),
```
NathanFarmer: Or maybe since this fails on SQL this is a real bug

great_expectations/metrics/multi_column_values/multi_column_values.py
```diff
@@ -0,0 +1,15 @@
+from typing import Literal, Optional
```
NathanFarmer: Can you rename this directory and file `multi_column` to follow the existing pattern

NathanFarmer: Also the `__init__.py` in this directory should import `from .multi_column import MultiColumnMetric`

great_expectations/metrics/multi_column_values/sum.py
```diff
@@ -0,0 +1,14 @@
+from typing import Union
```
NathanFarmer: Can you rename this file `sum_equal.py` like in the tests?

tests/integration/metrics/multicolumn_values/test_sum_equal.py
```diff
@@ -0,0 +1,155 @@
+import pandas as pd
```
NathanFarmer: Can you rename this directory `multi_column`?

NathanFarmer (changes requested):
I'm not blocking on the xfails, just some requests for re-organization. Also, can you add this metric to the `great_expectations.metrics` package?

tests/integration/metrics/multicolumn_values/test_sum_equal.py
joshua-stauffer: right, hard to say - i've pushed the explicit types to see if that helps

tests/integration/metrics/multi_column/test_sum_equal.py
```diff
@@ -0,0 +1,156 @@
...
+    {
+        COL_A: [1, 2, 3, 4],
+        COL_B: [4, 3, 2, 1],
+    },
+)
+SUCCESS_SUM_TOTAL = 5
+SUCCESS_COUNT = 0
+FAILURE_SUM_TOTAL = 0
+FAILURE_COUNT = 4
+DATA_FRAME_WITH_NULLS = pd.DataFrame(
```
billdirks, line 30: Can you add a case here where both are not None but don't sum to 5?

billdirks (approved):
This LGTM, thanks! I have requested you add a test for the failure when the sums do no sum to the expected value and neither are None.

tests/integration/metrics/multi_column/test_sum_equal.py
joshua-stauffer: i believe the `failure` test case below covers this, but let me know in case there's something else you're looking for

tests/integration/metrics/multicolumn_values/test_sum_equal.py
joshua-stauffer: we can't use explicit types as pytest parameters, because they belong to the Spark compatibility layer and need to exist in the top level namespace, and so blow up in the unit tests.

NathanFarmer: I had to do [this](https://github.com/great-expectations/great_expectations/blob/325133d85920cb7bd271d4c65422a8069e560140/tests/integration/data_sources_and_expectations/test_expectation_conditions.py#L26) in the past

NathanFarmer (approved):

great_expectations/metrics/multi_column/multi_column.py
```diff
@@ -0,0 +1,14 @@
+from typing import Literal, Optional
+
+from great_expectations.compatibility.pydantic import StrictStr
+from great_expectations.expectations.model_field_types import ConditionParser
+from great_expectations.metrics.metric import Metric, _MetricResult
+
+
+class MultiColumnMetric(Metric[_MetricResult], kw_only=True):
```
NathanFarmer, line 8: I think this should inherit from `BatchMetric`. A `MultiColumnMetric` will always be a subset of a batch. Then you won't need to add `row_condition`. Also, the goal is to infer `condition_parser` eventually, but so far we aren't testing conditions anywhere so I don't care if we include `condition_parser` or not.

NathanFarmer (changes requested):

great_expectations/metrics/multi_column/multi_column.py
joshua-stauffer: i don't think we can infer `condition_parser`, since for instance the `Spark` allows both the stock GX parser and custom spark queries to be passed through.

NathanFarmer (approved):

PR #10976, opened 2025-02-27
[DOCS] Adds script to update links when new docs version is cut

JessSaavedra:
## Description
As the API Reference doc files are generated with sphinx and later manipulated with beautiful soup, links for a newly cut version of docusaurus will not be updated correctly automatically. The purpose of this script is to be able to update all the links of api reference mdx files inside `versioned_docs` folder.

## How to use

1. Cut new version of Docusaurus first
2. Run the script passing as a param the last archived version (so for example if old version was 1.1.0 and you release version 1.2.0 after cut, you should pass 1.1.0 as a parameter)
`python3 explicit_docusaurus_version_in_api_reference_links.py 1.1.0`
This will result on the internal links on version 1.1.0 folder to have the version explicitly set in the links so they won't break when the page is up 

---

### DIFF
--- a/scripts/explicit_docusaurus_version_in_api_reference_links.py
+++ b/scripts/explicit_docusaurus_version_in_api_reference_links.py
@@ -1,0 +1,20 @@
+from pathlib import Path
+
+import sys
+
+if __name__ == "__main__":
+    version_to_update = sys.argv[1]
+    folder_path = f"../docs/docusaurus/versioned_docs/version-{version_to_update}/reference/api"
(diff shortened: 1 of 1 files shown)


tyler-hoffman (approved):
Bummer we have to do this. I thought all the versioning issues were ironed out. But it's been a while since I've looked at this stuff, and the code here makes sense. Can you add a docstring with some of the info you included in the PR description (e.g. how/when to call), and then I think there's a readme about releasing that should be updated.

Also thanks for doing this!

PR #10979, opened 2025-02-28
[BUGFIX] Fix 3 batch.compute_metrics bugs and enforce len(metrics) == len(results)

billdirks:
Fixes 3 batch.compute_metrics() bugs:
1. We no longer overwrite the metric result if metric names are identical.
2. The config id is different between batches. Previously, we set the batch_id after the hashing to create the id was computed.
3. Return a list of results if a list is passed into compute_metrics(). Previously a list of 1 would return the result not a list containing only the result.

We also enforce the invariant the the length of metrics passed in will equal the length of the results by adding a catchall `else` branch.

### DIFF
--- a/great_expectations/datasource/fluent/interfaces.py
+++ b/great_expectations/datasource/fluent/interfaces.py
@@ -21,7 +21,6 @@ from typing import (
     Mapping,
     MutableMapping,
     MutableSequence,
-    NamedTuple,
     Optional,
     Protocol,
     Sequence,
@@ -59,16 +58,12 @@ from great_expectations.exceptions.exceptions import (
     MissingDataContextError,
 )
 from great_expectations.metrics.metric import MetaMetric, Metric
-from great_expectations.metrics.metric_name import MetricNameSuffix
 from great_expectations.metrics.metric_results import (
     MetricErrorResult,
-    MetricErrorResultValue,
     MetricResult,
 )
 from great_expectations.validator.metrics_calculator import (
     MetricsCalculator,
-    _AbortedMetricsInfoDict,
-    _MetricsDict,
 )
 
 logger = logging.getLogger(__name__)
@@ -109,8 +104,6 @@ if TYPE_CHECKING:
         TypeLookup,
     )
     from great_expectations.expectations.expectation import Expectation
-    from great_expectations.validator.computed_metric import MetricValue
-    from great_expectations.validator.metric_configuration import MetricConfigurationID
     from great_expectations.validator.v1_validator import (
         Validator as V1Validator,
     )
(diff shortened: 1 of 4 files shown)


great_expectations/datasource/fluent/interfaces.py
```diff
@@ -1302,88 +1287,40 @@ def compute_metrics(self, metrics: Metric | list[Metric]) -> MetricResult | list
             this by adding an explicit annotation (e.g. metrics: list[Metric] = ...)
             https://github.com/python/mypy/issues/18712
         """
+        is_single_metric = False
```
billdirks, line 1292: I keep track of this so I can tell whether `[metric]` or just `metric` was passed in so we can return `[metric_result]` or `metric_result`. Previously we just checked for the length of the input which doesn't distinguish these.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -1302,88 +1287,40 @@ def compute_metrics(self, metrics: Metric | list[Metric]) -> MetricResult | list
...
-                metric_results.append(
-                    MetricErrorResult(
-                        id=raw_metrics_errors[metric_name].metric_configuration_id,
-                        value=raw_metrics_errors[metric_name].value,
+                value = metrics_calculator_results[metric_id_for_batch]
+                results.append(MetricResultType(id=metric_id_for_batch, value=value))
+            elif metric_id_for_batch in metrics_calculator_errors:
+                value = metrics_calculator_errors[metric_id_for_batch]
+                results.append(MetricErrorResult(id=metric_id_for_batch, value=value))
+            else:
```
billdirks, line 1317: Added an `else` to enforce the one output per input. I saw this fail when the metric named changed for the spark mean failure (eg bug in gx).

great_expectations/metrics/metric.py
```diff
@@ -120,16 +122,20 @@ def __new__(cls, *args, **kwargs):
...
+
     def config(self, batch_id: str) -> MetricConfiguration:
         # This class is frozen so Metric._to_config will always return the same value
         # when the same batch_id is passed in.
         if not batch_id:
             raise EmptyStrError("batch_id")
+
         config = Metric._to_config(
             instance_class=self.__class__,
+            batch_id=batch_id,
```
billdirks, line 139: We compute the config id in the `__init__` method so it is invariant. That means when I set the batch_id after instantiation, it didn't update the id. This was a bug since that means different batches would have the same id.

tests/metrics/test_metric.py
```diff
@@ -19,6 +22,13 @@
 class ColumnValuesAboveResult(MetricResult[bool]): ...
 
 
+class ColumnValuesAbove(ColumnMetric[ColumnValuesAboveResult]):
```
billdirks, line 25: Since we defined this in every test class I made this top level. In the definition tests, I renamed this to `MyColumnValuesAbove`.

great_expectations/metrics/metric.py
```diff
@@ -104,6 +105,7 @@ class Metric(Generic[_MetricResult], BaseModel, metaclass=MetaMetric):
     name: ClassVar[StrictStr]
 
     _domain_fields: ClassVar[tuple[str, ...]] = (
+        "batch_id",
```
NathanFarmer: Why did you need to add this?

NathanFarmer (approved):
One question, but other than that, simpler AND more test cases!

great_expectations/metrics/metric.py
billdirks: I was going to comment on this. This wasn't necessary, I just added it because I thought it belonged as a domain field. We're treating it differently though so I will remove.

PR #10984, opened 2025-02-28
[MAINTENANCE] Analytics event for validation definition runs

tyler-hoffman:
I'm bundling a couple things in here that I think would be nice to have for analytics:
* events when we run a validation definition. Note that we already emit an event when running a checkpoint, but you can run a validation definition without running a checkpoint. I think both are useful, even if there is overlap.
* Adds a `mode` field to all events to distinguish from ephemeral, file, and cloud. We also had "cloud_mode", but that is just a boolean. We should probably remove it, but that will take coordination.


great_expectations/analytics/events.py
```diff
@@ -255,6 +256,31 @@ def __init__(self, validation_definition_id: str | None = None):
...
+        }
+
+    def __init__(
+        self,
+        *,
+        validation_definition_id: str | None = None,
+        checkpoint_id: str | None = None,
+    ):
+        self.checkpoint_id = checkpoint_id
+        super().__init__(
```
tyler-hoffman, line 278: Coming back to this, it's a bummer that we are implementing our own constructors for these events. dataclasses should do that for us. Seems like there's a lot of low-priority cleanup around here that we could get to once we finish implementing all the features 😛 .

great_expectations/analytics/actions.py
```diff
@@ -14,4 +14,5 @@
 EXPECTATION_SUITE_EXPECTATION_UPDATED: Action = Action("expectation_suite.expectation_updated")
 VALIDATION_DEFINITION_CREATED: Action = Action("validation_definition.created")
 VALIDATION_DEFINITION_DELETED: Action = Action("validation_definition.deleted")
+VALIDATION_DEFINITION_DELETED: Action = Action("validation_definition.ran")
```
sujensen: Is that supposed to be VALIDATION_DEFINITION_RAN ?

tyler-hoffman: Hah yeah!

great_expectations/analytics/client.py
```diff
@@ -46,6 +46,7 @@ def init(  # noqa: PLR0913 # FIXME CoP
     oss_id: Optional[UUID] = None,
     cloud_mode: bool = False,
     user_agent_str: Optional[str] = None,
+    mode: Literal["cloud", "ephemeral", "file", None] = None,
```
joshua-stauffer: i assume there's a valid case, but want to ask anyway - in what case can this be `None`?

joshua-stauffer (approved):
looks great, thank you!

great_expectations/analytics/client.py
tyler-hoffman: I did it to follow the pattern of everything being optional, but there's no reason to make this field optional other than to make mistakes easier to ship.

PR #10995, opened 2025-03-04
[BUGFIX] Raise error when ExpectColumnValuesToBeBetween is run against an unsupported column

joshua-stauffer:
We enforce that max_value and min_value for ExpectColumnValuesToBeBetween must be a numeric or datetime type, but we don't check the underlying data. In SQL contexts where the underlying data is something else such as VARCHAR, users receive opaque exceptions generated by their database, and it isn't clear what they should do to resolve the error. Worse, because of how we bundle metric computations into a single SQL query, every expectation computed in the same run will likely fail with the same opaque error.
- OSS report: https://greatexpectationstalk.slack.com/archives/CUTCNHN82/p1739807314212519
- internal bug report: https://greatexpectations.atlassian.net/browse/GX-306

This PR updates the ColumnValuesBetween metric to make a basic check against the column type prior to building the query.


josectobar (approved):
Looks great!

PR #10997, opened 2025-03-05
[MAINTENANCE] refactor to use data_context fixture

joshua-stauffer:
This PR updates a number of tests to use a new `data_context` fixture that previously were written as separate tests.


tyler-hoffman (approved):
Thanks! Much less code smell 😉

PR #11000, opened 2025-03-06
[BUGFIX] on update data source don't delete assets

joshua-stauffer:
This PR updates `context.data_sources.add_or_update...` to only update the Datasource config itself, and not remove existing assets, which is current behavior. While as far as I can see this was a bug, changing it will break existing behavior. The issue is that we have not implemented `Datasource.update_asset`. Assets can never be updated, only created or deleted. We probably need to ship that change along with this one, so users who were relying on the buggy implementation have an alternative.


joshua-stauffer:
POC, no intent to merge right now.

PR #11011, opened 2025-03-08
[FEATURE] Add initial RedShift datasource

billdirks:


great_expectations/self_check/util.py
```diff
@@ -1307,6 +1307,9 @@ def candidate_test_is_on_temporary_notimplemented_list_v3_api(context, expectati
     return False
 
 
+test_connections = set()
```
billdirks: This isn't strictly required. I noticed redshift connections being slow and that we check if we are able to connect many times. This caches our check so we only make the connection once.

.github/workflows/ci.yml
```diff
@@ -574,6 +582,7 @@ jobs:
           - mssql
           - mysql
           - postgresql
+          - redshift
```
billdirks, line 585: The redshift tests won't actually run until after this PR is merged.

great_expectations/self_check/util.py
```diff
@@ -1317,6 +1317,11 @@ def candidate_test_is_on_temporary_notimplemented_list_v3_api(context, expectati
     return False
 
 
+# We cache whether we can successfully connect to a test backend so we don't have to
+# validate we can connect over and over again.
+_successful_backend_connection = set()
```
billdirks, line 1322: I haven't plumbed this up to all the connections which means we still do some extra work. I can do this in a followup PR. The redshift connection takes a while to establish so I added it there for now.

great_expectations/datasource/fluent/redshift_datasource.py
```diff
@@ -0,0 +1,29 @@
...
+    Args:
+        name: The name of this redshift datasource.
+        connection_string: The SQLAlchemy connection string used to connect to the redshift
+            database. This will use a redshift with psycopg2. For example:
+            "redshift+psycopg2://username@host.amazonaws.com:5439/database"
+        assets: An optional dictionary whose keys are TableAsset or QueryAsset names and whose
+            values are TableAsset or QueryAsset objects.
+    """
+
+    type: Literal["redshift"] = "redshift"  # type: ignore[assignment] # This is a hardcoded constant
```
NathanFarmer, line 28: Out of scope for this PR since it applies to all Data Sources, but I think this should be annotated with `ClassVar` and then added to the schema [this way](https://github.com/great-expectations/great_expectations/blob/65e4ce7ab194a981babba6765ed99c851614f3e0/great_expectations/expectations/core/expect_column_max_to_be_between.py#L226) if we really need it in the schema.

great_expectations/self_check/util.py
```diff
@@ -1513,9 +1518,11 @@ def build_test_backends_list(  # noqa: C901, PLR0912, PLR0913, PLR0915 # FIXME C
...
             # noinspection PyUnresolvedReferences
             try:
-                engine = _create_redshift_engine()
-                conn = engine.connect()
-                conn.close()
+                if "redshift" not in _successful_backend_connection:
+                    engine = _create_redshift_engine()
+                    conn = engine.connect()
+                    conn.close()
+                    _successful_backend_connection.add("redshift")
```
NathanFarmer, line 1525: I thought this only applied to legacy testing, and I don't see any changes to legacy tests. Does this code get run in production? I'm just now realizing that I don't understand why this entire file isn't in the `tests` directory.

NathanFarmer (approved):

great_expectations/self_check/util.py
billdirks: I think this file should be in the tests directory. Historically I think we had some tools that let people check their yaml strings that lived in this file. After talking with product today, I pulled in a ticket to delete some legacy tests and hopefully that will allow be to delete this file though I haven't tracked down everywhere this file is used yet.

great_expectations/datasource/fluent/redshift_datasource.py
billdirks: That sounds good. I don't like this type ignore. I'll have to see how making this a classvar interacts with deserializing since we use this field for pydantic type discrimination to create the right datasource class from the configuration.

great_expectations/self_check/util.py
billdirks: Oh in terms of why this is here. redshift is in the backend list so it now tries to connect everytime this function is called. It takes seconds to connect to redshift and this function actually gets called multiple times during tests.

PR #11019, opened 2025-03-12
[DOCS] fix error in filesystem data source docs

joshua-stauffer:
This PR corrects an error that a user pointed out in our filesystem data source docs: https://discourse.greatexpectations.io/t/regex-compile-error-when-trying-to-add-batch-definition-path/2138


docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_data_source/_local_or_networked/_local_or_networked.md
```diff
@@ -38,7 +38,7 @@ import PandasDefault from './_pandas_default.md'
    - `name`: A descriptive name used to reference the Data Source.  This should be unique within the Data Context.
    - `base_directory`: The path to the folder that contains the data files, or the root folder of the directory hierarchy that contains the data files.
    
-   If you are using a File Data Context, you can provide a path that is relative to the Data Context's `base_directory`.  Otherwise, you should provide the absolute path to the folder that contains your data.
+   If you are using a Filesystem Data Source, provide a path that is relative to the Data Source's `base_directory`.  Otherwise, provide an absolute path to the folder that contains your data.
```
klavavej: I find it confusing that the Data Source `base_directory` parameter should be "relative to the Data Source's `base_directory`". This feels like a circular definition or a chicken and egg problem. 
Would it be accurate to instead say the following?

```suggestion
   If you are using a Filesystem Data Source, provide a path that is relative to the Data Source's root directory.  Otherwise, provide an absolute path to the folder that contains your data.
```

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_data_source/_local_or_networked/_pandas.py
```diff
@@ -4,7 +4,7 @@
 context = gx.get_context()
 
 # Define the Data Source's parameters:
-# This path is relative to the `base_directory` of the Data Context.
+# This path is relative to the `base_directory` of the Data Source.
```
klavavej: Similar to another note, it feels like a circular definition to say that the Data Source's `base_directory` parameter is "relative to the `base_directory` of the Data Source"

Would it be accurate to say 
```suggestion
# This path is relative to the root directory of the Data Source.
```

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_data_source/_local_or_networked/_spark.py
```diff
@@ -4,7 +4,7 @@
 context = gx.get_context()
 
 # Define the Data Source's parameters:
-# This path is relative to the `base_directory` of the Data Context.
+# This path is relative to the `base_directory` of the Data Source.
```
klavavej: again wondering if it would be more accurate/clear to say

```suggestion
# This path is relative to the root directory of the Data Source.
```

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_data_source/_local_or_networked/_local_or_networked.md
klavavej: There's some seemingly related content at https://docs.greatexpectations.io/docs/core/define_expectations/retrieve_a_batch_of_test_data?method=from_pandas_default#procedure-pandas-default and I wonder if it should get the same kind of update (bold used only to highlight the phrase of interest)

> The simplest method is to provide an absolute path to the datafile that you will retrieve records from. However, if you are using a **File Data Context** you can also provide a path relative to the **Data Context's** `base_directory`.

Should this be changed to 

> The simplest method is to provide an absolute path to the datafile that you will retrieve records from. However, if you are using a **Filesystem Data Source** you can also provide a path relative to the **Data Source**'s `base_directory`.

klavavej (changes requested):
requesting changes because I'm confused about one aspect of the update

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_data_source/_local_or_networked/_local_or_networked.md
joshua-stauffer: hm, i think it's clearer to refer directly to the `base_directory` parameter. Calling it something else like "root directory" means readers need to figure out what that is on their own. To clarify, it isn't the `base_directory` that is relative to the `base_directory`, it is the BatchDefinition's `regex` that is relative to the `base_directory` parameter of the data source.

joshua-stauffer: While that doc shares the exact phrase i corrected, it is actually correct! In that case, the user doesn't have a data source (yet). They're just pointing to data -- which can be referenced as an absolute path or one relative to the `base_directory` parameter of their GX config (data context) -- and receiving a batch in return.

klavavej: >To clarify, it isn't the base_directory that is relative to the base_directory, it is the BatchDefinition's regex that is relative to the base_directory parameter of the data source.

huh, now I'm even more confused

I read the word "path" in the updated paragraph as referring to the same thing as the word "path" in the immediately preceding definition of `base_directory`.

Screenshot 2025-03-12 at 12 38 23 PM

This update appears in the "Create a Data Source" section of the page. Should it have been made in the "Create a Batch Definition" section instead?

joshua-stauffer: yes! nice catch, and thanks for double checking.

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_tab-file_batch_definition.md
```diff
@@ -31,7 +31,7 @@ Batch Definitions for File Data Assets can be configured to return the content o
    To define a path Batch Definition you need to provide the following information:
 
    - `name`: A name by which you can reference the Batch Definition in the future.  This should be unique within the Data Asset.
-   - `path`: The path within the Data Asset of the data file containing the records to return.
+   - `path`: The path within the Data Asset of the data file containing the records to return. When using a Filesystem Data Source, `path` is relative to Data Source param `base_directory`.
```
klavavej: suggestion to use the full word "parameter" for clarity. also adds a missing article

```suggestion
   - `path`: The path within the Data Asset of the data file containing the records to return. When using a Filesystem Data Source, `path` is relative to the Data Source parameter `base_directory`.
```

docs/docusaurus/docs/core/connect_to_data/filesystem_data/_create_a_batch_definition/_tab-file_batch_definition.md
```diff
@@ -42,7 +42,10 @@ Batch Definitions for File Data Assets can be configured to return the content o
 
    <TabItem value="partitioned" label="Partitioned">
    
-   GX Core currently supports partitioning File Data Assets based on dates.  The files can be returned by year, month, or day.
+   GX Core currently supports partitioning File Data Assets based on dates.  The files can be returned by year, month, or day. To define a Batch Definition, you need to provide the following information: 
+
+    - `name`: A name by which you can reference the Batch Definition in the future.  This should be unique within the Data Asset.
+    - `regex`: A regular expression used to match against file names. When using a Filesystem Data Source, `regex` is relative to Data Source param `base_directory`.
```
klavavej: suggestion to use the full word "parameter" for clarity. also adds a missing article

```suggestion
    - `regex`: A regular expression used to match against file names. When using a Filesystem Data Source, `regex` is relative to the Data Source parameter `base_directory`.
```

klavavej (approved):
thanks! this makes a lot more sense. Left a couple suggestions for small grammar tweaks

PR #11027, opened 2025-03-14
[FEATURE] Add the ability to run checkpoints with cloud windowed expectations

billdirks:

### DIFF
--- a/great_expectations/data_context/data_context/cloud_data_context.py
+++ b/great_expectations/data_context/data_context/cloud_data_context.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import logging
 import os
-from urllib.parse import urljoin
 import uuid
 from typing import (
     TYPE_CHECKING,
@@ -13,6 +12,7 @@ from typing import (
     Optional,
     Union,
 )
+from urllib.parse import urljoin
 
 from requests import HTTPError, Response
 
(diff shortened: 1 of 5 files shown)


great_expectations/data_context/data_context/cloud_data_context.py
```diff
@@ -654,3 +657,57 @@ def _add_datasource(
...
+        try:
+            overlapping_keys = set(expectation_parameters.keys()) & set(
+                data["data"]["expectation_parameters"].keys()
+            )
+            if overlapping_keys:
+                logger.warning(
+                    "Passed in expectation_parameters also found in GX Cloud. Overwriting "
+                    f"passed in values with GX Cloud values for keys: {overlapping_keys}"
+                )
+            expectation_parameters.update(data["data"]["expectation_parameters"])
```
shiplet, line 700: Okay so this always overrides locally-defined expectation params with the ones found in cloud, is that right?

shiplet (approved):
lgtm! only one non-blocking question

great_expectations/data_context/data_context/cloud_data_context.py
billdirks: Yes. If there is both a local and a cloud value for a parameter the cloud one wins. I have a test that sets the same parameter locally and asserts it gets overridden with the cloud value. A warning will be logged when this occurs that we are overriding with the cloud value.

If there are parameters are set locally but not in cloud, these will keep their local setting.

shiplet: Awesome, thanks for clarifying!

PR #11047, opened 2025-03-26
[DOCS] Add key pair authentication as a limitation of ExpectAI in manage_expectations.md

adeola-ak:
add key pair authentication as a limitation of ExpectAI


klavavej (approved):
looks great - thank you!

PR #11055, opened 2025-04-01
[BUGFIX] Prevent sqlite metric impl from overwriting default sqlalchemy

billdirks:
Note, this bug was never release onto pypi.

### DIFF
--- a/great_expectations/execution_engine/sqlite_execution_engine.py
+++ b/great_expectations/execution_engine/sqlite_execution_engine.py
@@ -27,7 +27,7 @@ class ColumnStandardDeviation(BaseColumnStandardDeviation):
     """MetricProvider Class for Aggregate Standard Deviation metric for SQLite databases."""
 
     # We should change this decorator to compute this metric a completely new way
-    @column_aggregate_partial(engine=SqlAlchemyExecutionEngine)
+    @column_aggregate_partial(engine=SqliteExecutionEngine)
     def _sqlalchemy(cls, column, _dialect, _metrics, **kwargs):
         """Sqlite Standard Deviation implementation"""
         mean = _metrics["column.mean"]
--- a/great_expectations/expectations/metrics/column_aggregate_metric_provider.py
+++ b/great_expectations/expectations/metrics/column_aggregate_metric_provider.py
@@ -60,7 +60,7 @@ def column_aggregate_value(
     if issubclass(engine, PandasExecutionEngine):
 
         def wrapper(metric_fn: Callable):
-            @metric_value(engine=PandasExecutionEngine)
+            @metric_value(engine=engine)
             @wraps(metric_fn)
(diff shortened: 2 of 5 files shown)


great_expectations/execution_engine/sqlite_execution_engine.py
```diff
@@ -27,7 +27,7 @@ class ColumnStandardDeviation(BaseColumnStandardDeviation):
     """MetricProvider Class for Aggregate Standard Deviation metric for SQLite databases."""
 
     # We should change this decorator to compute this metric a completely new way
-    @column_aggregate_partial(engine=SqlAlchemyExecutionEngine)
+    @column_aggregate_partial(engine=SqliteExecutionEngine)
```
billdirks, line 30: This bug had 2 parts. The first part was not setting the engine correctly here.

great_expectations/expectations/metrics/column_aggregate_metric_provider.py
```diff
@@ -60,7 +60,7 @@ def column_aggregate_value(
     if issubclass(engine, PandasExecutionEngine):
 
         def wrapper(metric_fn: Callable):
-            @metric_value(engine=PandasExecutionEngine)
+            @metric_value(engine=engine)
```
billdirks, line 63: The second part of the bug was hard coding the engine names in the metric decorators. I've grepped all the metric provider files and this file is the only one I saw containing hard coded values.

If we ever need to add other engines, we will need to update these decorators since they switch on subclass. We should refactor to avoid this open/close principle violation if we add a new engine.

great_expectations/expectations/registry.py
```diff
@@ -307,16 +307,30 @@ def register_metric(  # noqa: PLR0913 # FIXME CoP
...
-def get_metric_provider(
-    metric_name: str, execution_engine: ExecutionEngine
+def _get_metric_definition(metric_name: str) -> dict:
+    try:
+        return _registered_metrics[metric_name]
+    except KeyError:
+        raise gx_exceptions.MetricProviderError(f"No definition found for {metric_name}")  # noqa: TRY003 # FIXME CoP
+
+
+def get_sqlalchemy_metric_provider(
```
billdirks, line 317: I added this helper method because I didn't want to reach into the _registered_metrics from tests. I also didn't want to need a real sqlalchemy execution engine in tests because we don't have sqlalchemy connectors in unit tests and I didn't want to add that dependency. This felt ok because this is the default providers.

great_expectations/expectations/metrics/column_aggregate_metric_provider.py
NathanFarmer: It sure doesn't seem like refactoring this would be hard. The logic is almost identical at this point.

NathanFarmer (approved):

PR #11062, opened 2025-04-03
[FEATURE] Fix ExpectColumnValuesToBeOfType bug and also work with sqla2

billdirks:


great_expectations/execution_engine/redshift_execution_engine.py
```diff
@@ -0,0 +1,179 @@
...
+from great_expectations.execution_engine.sqlalchemy_execution_engine import (
+    SqlAlchemyExecutionEngine,
+)
+from great_expectations.expectations.metrics.metric_provider import metric_value
+from great_expectations.expectations.metrics.table_metrics.table_column_types import (
+    ColumnTypes as BaseColumnTypes,
+)
+
+
+class RedshiftExecutionEngine(SqlAlchemyExecutionEngine):
```
billdirks, line 21: I add a redshift specific execution engine so I can write metrics specific to redshift and isolate this from the rest of our sqlalchemy datasources. This is to prevent conditional logic in our metrics code that switches based on dialect.

great_expectations/execution_engine/redshift_execution_engine.py
```diff
@@ -0,0 +1,179 @@
...
+    def _sqlalchemy(
+        cls,
+        execution_engine: SqlAlchemyExecutionEngine,
+        metric_domain_kwargs: dict,
+        metric_value_kwargs: dict,
+        metrics: dict[str, Any],
+        runtime_configuration: dict,
+    ):
+        # For sqlalchemy version < 2 fallback to default implementation
+        if sa.__version__[0] != "2":
```
billdirks, line 72: The internals of the postgres dialect, on which redshift is based, changes between sqlalchemy 1 and 2. It was easier to separate the logic based on the sqlalchemy version since there appear to be real differences. We fallback to our default implementation for sqlalchemy 1.

great_expectations/execution_engine/redshift_execution_engine.py
```diff
@@ -0,0 +1,179 @@
...
+                runtime_configuration=runtime_configuration,
+            )
+
+        # For sqlalchemy 2 use this new implementation which avoids incompatible parts of dialect
+        # Get the table information
+        assert isinstance(execution_engine, RedshiftExecutionEngine)
+        table_name, schema_name = cls._get_table_schema(execution_engine, metric_domain_kwargs)
+        full_table_name = f"{schema_name}.{table_name}" if schema_name else table_name
+        # Query for the column information
+        query = sa.text(f"""
```
billdirks, line 88: Instead of calling `get_columns` we execute this query. `get_columns` fails for sqlalchemy 2. This query will succeed but returns a string so we have to process the returned string which is all the string manipulation below. While string manipulation can be brittle, the return string is ultimately pretty constrained so I'm ok with this solution though it is a bit ugly.

Note, I tracked down the issued with calling `get_columns`. There is a private method `_load_domains` which the redshift dialect inherits from the postgres one. This basically runs a query to get some type information from the database. We could override this but there are some complications:
1. We need to have our own version of the redshift dialect. We do have a fork but we haven't started using it yet.
2. The information returned in sqlalchemy 2 is richer than in sqlalchemy 1. This will take some investigation to figure out how to do this which I didn't think was worth it at the moment.

I will file a followup ticket and we can pursue the overloading `_load_domains` approach if we decide that is worth it and we want to fix our redshift fork more completely.

tests/integration/data_sources_and_expectations/expectations/test_expect_column_values_to_be_in_type_list.py
```diff
@@ -666,3 +668,113 @@ def test_success_complete_databricks_double_type_only(
         assert isinstance(result_dict["observed_value"], str)
         assert isinstance(expectation.type_list, list)
         assert result_dict["observed_value"] in expectation.type_list
+
+
+# Redshift will lowercase column names so we make the column names here all lowercase
+# These means we aren't case insensitive in this expectation.
```
billdirks, line 674: I can file a followup ticket about this if we want to pursue it.

great_expectations/execution_engine/redshift_execution_engine.py
```diff
@@ -0,0 +1,179 @@
...
+        "boolean": aws.redshiftdialect.BOOLEAN,
+        "smallint": aws.redshiftdialect.SMALLINT,
+        "integer": aws.redshiftdialect.INTEGER,
+        "bigint": aws.redshiftdialect.BIGINT,
+        "real": aws.redshiftdialect.REAL,
+        "double precision": aws.redshiftdialect.DOUBLE_PRECISION,
+        # Numeric is odd since the data returns numeric as a type to pg_get_cols
+        # In sqlalchemy 1.4 we get the type sa.sql.sqltypes.NUMERIC returned.
+        # However, the sqlalchemy redshift dialect only has DECIMAL. The official
+        # redshift docs say 'DECIMAL' is the correct type and NUMER is an alias:
```
billdirks, line 40: This DECIMAL/NUMERIC inconsistency within the dialect is a little strange and we can fix up our fork once that is up if we want to follow up on this.

great_expectations/execution_engine/redshift_execution_engine.py
```diff
@@ -0,0 +1,179 @@
...
+        "time with time zone": aws.redshiftdialect.TIMETZ,
+        "timestamp without time zone": aws.redshiftdialect.TIMESTAMP,
+        "timestamp with time zone": aws.redshiftdialect.TIMESTAMPTZ,
+    }
+    if aws.redshiftdialect
+    else {}
+)
+
+
+def split_by_comma_outside_quotes(text: str) -> list[str]:
```
dctalbot: It would be nice to cover this with a unit test. I'm pretty good with regex and I still have a hard time parsing the implementation here.

great_expectations/execution_engine/redshift_execution_engine.py
```diff
@@ -0,0 +1,179 @@
...
+            )
+
+        # For sqlalchemy 2 use this new implementation which avoids incompatible parts of dialect
+        # Get the table information
+        assert isinstance(execution_engine, RedshiftExecutionEngine)
+        table_name, schema_name = cls._get_table_schema(execution_engine, metric_domain_kwargs)
+        full_table_name = f"{schema_name}.{table_name}" if schema_name else table_name
+        # Query for the column information
+        query = sa.text(f"""
+            select pg_get_cols('{full_table_name}');
```
dctalbot: I think you might have a better time parsing with `select * from pg_get_cols('{full_table_name}') `

See the second example here: https://docs.aws.amazon.com/redshift/latest/dg/PG_GET_COLS.html#PG_GET_COLS-example

great_expectations/execution_engine/redshift_execution_engine.py
```diff
@@ -0,0 +1,179 @@
...
+            # If the type is parameterized, we must removed the arguments to the types to do a
+            # string lookup from the type string to the actual type.
+            result = r[0].strip("()")
+            _schema, _table, column, raw_column_type, _column_num = split_by_comma_outside_quotes(
+                result
+            )
+            column_type_str_parts = raw_column_type.strip('"').split("(")
+            column_base_type = REDSHIFT_TYPES.get(column_type_str_parts[0])
+            if column_base_type is None:
+                raise GreatExpectationsError(  # noqa: TRY003
```
dctalbot: I wonder if we should turn off TRY003. Otherwise, I'd appreciate some inline justification for ignoring the rule in each instance

great_expectations/execution_engine/redshift_execution_engine.py
dctalbot: actually see my comment below, maybe we can avoid this completely

dctalbot (changes requested):
blocking for discussion

great_expectations/execution_engine/redshift_execution_engine.py
billdirks: deleted, thanks!

great_expectations/execution_engine/redshift_execution_engine.py
billdirks: This is so much better!

great_expectations/execution_engine/redshift_execution_engine.py
billdirks: I added a new error type so I can not add the noqa. I don't think this error message has much value and will see if we want to remove in general.

dctalbot (approved):

PR #11082, opened 2025-04-09
[BUGFIX] Expectations fail validation when using sets/tuples as value_set due to JSON serialization

gyermich:
This PR fixes an issue where using sets or tuples as `value_set` in expectations like `ExpectColumnValuesToBeInSet` would fail validation. The failure occurred because JSON does not natively support sets or tuples, leading to these types being converted to lists during serialization. This mismatch caused validation issues. The fix ensures that Pydantic coerces set-like inputs to list before validation (excluding string and bytes).

**Example**
Before this fix, the following would fail:
```python
COUNTRIES = ("UNITED STATES", "CHINA", "SPAIN")  # tuple
expectation = ExpectColumnValuesToBeInSet(
    value_set=COUNTRIES,
    column="country_name_en"
)
```

After the fix, both tuples and sets work as expected:
```python
# Both of these now work:
COUNTRIES_TUPLE = ("UNITED STATES", "CHINA", "SPAIN")
COUNTRIES_SET = {"UNITED STATES", "CHINA", "SPAIN"}
```

**Testing**
- Added test cases for various input types 
- Verified that `SuiteParameterDict` handling remains unchanged
- Ran the script referenced in the JIRA issue description locally - it now works as expected

### DIFF
--- a/great_expectations/expectations/model_field_types.py
+++ b/great_expectations/expectations/model_field_types.py
@@ -1,6 +1,6 @@
-from typing import Literal, Sequence, Union, get_args
+from typing import Any, List, Literal, Sequence, Union, get_args
 
-from great_expectations.compatibility import pydantic
+from great_expectations.compatibility.pydantic import BaseModel, Field, validator
 from great_expectations.compatibility.typing_extensions import Annotated
 from great_expectations.core.suite_parameters import (
     SuiteParameterDict,  # used in pydantic validation
@@ -10,9 +10,31 @@ from great_expectations.expectations.model_field_descriptions import (
     VALUE_SET_DESCRIPTION,
 )
 
+
+class ValueSetModel(BaseModel):
+    value: Union[List, SuiteParameterDict]
+
+    @validator("value", pre=True)
+    def cast_to_list(cls, value: Any) -> Any:
+        if isinstance(value, (set, Sequence)) and not isinstance(value, (str, bytes)):
+            return list(value)
+        return value
+
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, value: Any) -> Union[List, SuiteParameterDict]:
+        if isinstance(value, SuiteParameterDict):
+            return value
+        model = cls(value=value)
+        return model.value
+
+
 MostlyField = Annotated[
     Union[float, SuiteParameterDict],
-    pydantic.Field(
+    Field(
         description=MOSTLY_DESCRIPTION,
         ge=0.0,
         le=1.0,
@@ -22,8 +44,8 @@ MostlyField = Annotated[
 ]
 
 ValueSetField = Annotated[
-    Union[Sequence, set, SuiteParameterDict],
-    pydantic.Field(
+    ValueSetModel,
+    Field(
         title="Value Set",
         description=VALUE_SET_DESCRIPTION,
         schema_overrides={
--- a/great_expectations/expectations/model_field_types.py
+++ b/great_expectations/expectations/model_field_types.py
@@ -1,6 +1,6 @@
-from typing import Any, List, Literal, Sequence, Union, get_args
+from typing import List, Literal, Union, get_args
 
-from great_expectations.compatibility.pydantic import BaseModel, Field, validator
+from great_expectations.compatibility.pydantic import Field
 from great_expectations.compatibility.typing_extensions import Annotated
 from great_expectations.core.suite_parameters import (
     SuiteParameterDict,  # used in pydantic validation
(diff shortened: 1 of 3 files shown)


great_expectations/expectations/model_field_types.py
```diff
@@ -22,10 +22,14 @@
...
-    Union[Sequence, set, SuiteParameterDict],
-    pydantic.Field(
+    Union[List, SuiteParameterDict],
+    Field(
         title="Value Set",
         description=VALUE_SET_DESCRIPTION,
+        # Add type constraint to exclude strings and bytes
+        json_schema_extra={
+            "not": {"anyOf": [{"type": "string"}, {"type": "string", "format": "binary"}]}
+        },
```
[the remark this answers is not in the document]
gyermich: I am also not sure if we want to update schemas - can remove this addition and undo schema changes

great_expectations/expectations/model_field_types.py
```diff
@@ -22,10 +22,12 @@
...
 
 ValueSetField = Annotated[
-    Union[Sequence, set, SuiteParameterDict],
-    pydantic.Field(
+    Union[List, SuiteParameterDict],
+    Field(
         title="Value Set",
         description=VALUE_SET_DESCRIPTION,
+        # Add type constraint to exclude strings and bytes
+        not_={"anyOf": [{"type": "string"}, {"type": "string", "format": "binary"}]},
```
gyermich, line 30: for reviewers - this prevents string and bytestring from coercing to a list. Should we skip this change and instead have an expectation validation error if someone passes a string instead of expected iterables (which is unlikely but possible)?
I am also not sure if we want to update schemas - can remove this addition and undo schema changes

great_expectations/expectations/model_field_types.py
```diff
@@ -22,11 +22,15 @@
...
-    pydantic.Field(
+    Union[List, SuiteParameterDict],
+    Field(
         title="Value Set",
         description=VALUE_SET_DESCRIPTION,
+        # Add type constraint to exclude strings and bytes
         schema_overrides={
+            "not": {
+                "anyOf": [{"type": "string"}, {"type": "string", "format": "binary"}]
+            },
```
gyermich, line 33: for reviewers - this prevents string and bytestring from coercing to a list. Should we skip this change and instead have an expectation validation error if someone passes a string instead of expected iterables (which is unlikely but possible)?
I am also not sure if we want to update schemas - can remove this addition and undo schema changes

great_expectations/expectations/model_field_types.py
```diff
@@ -22,11 +22,13 @@
...
 ValueSetField = Annotated[
-    Union[Sequence, set, SuiteParameterDict],
-    pydantic.Field(
+    Union[List, SuiteParameterDict],
+    Field(
         title="Value Set",
         description=VALUE_SET_DESCRIPTION,
+        # Add type constraint to exclude strings and bytes
         schema_overrides={
+            "not": {"anyOf": [{"type": "string"}, {"type": "string", "format": "binary"}]},
```
gyermich: for reviewers - this prevents string and bytestring from coercing to a list. Should we skip this change and instead have an expectation validation error if someone passes a string instead of expected iterables (which is unlikely but possible)?
I am also not sure if we want to update schemas - can remove this addition and undo schema changes

NathanFarmer: Does this actually prevent the type coercion when using the API, or does it just change the schema? If it just changes the schema, I don't think it's necessary. You could raise an error in the validator if the wrong type is passed. This could be a more widespread problem so I wouldn't spend too much time thinking about it.

gyermich: By "prevent from coercing" I meant it would prevent processing the validation since this schema update would raise a validation error:
```
pydantic.v1.error_wrappers.ValidationError: 2 validation errors for ExpectColumnValuesToBeInSet
value_set
  value is not a valid list (type=type_error.list)
value_set
  value is not a valid dict (type=type_error.dict)
```

gyermich: Actually, I think we may not need it - I just tested without the schema changes locally and it still throws this validation error. I think it was coersing the string when we had `Sequence` which would allow string and bytes, but with `List` it's no longer the case. Going to undo this change!

NathanFarmer: I guess that's what my question was. I wouldn't expect it to raise a validation error if you are just updating the schema. Are you sure if you remove the schema changes that it will fail to raise the error?

gyermich: This is now reverted - updated the diff

NathanFarmer (approved):

PR #11087, opened 2025-04-10
[MAINTENANCE] Improve PostHog pageview tracking in docs

ashmortar:
Disable automatic pageview capture to prevent duplicates
Add custom NavigationTracker component for SPA navigation

### DIFF
--- a/docs/docusaurus/src/components/NavigationTracker/index.js
+++ b/docs/docusaurus/src/components/NavigationTracker/index.js
@@ -1,0 +1,14 @@
+import { useEffect } from 'react';
+import { useLocation } from '@docusaurus/router';
+
+export default function NavigationTracker() {
+  const location = useLocation();
+
+  useEffect(() => {
+    if (typeof window !== 'undefined' && window.posthog) {
+      window.posthog.capture('$pageview');
+    }
+  }, [location]);
+
+  return null;
+} 
--- a/docs/docusaurus/src/theme/Root/index.js
+++ b/docs/docusaurus/src/theme/Root/index.js
@@ -1,0 +1,14 @@
+import React from 'react';
+import BrowserOnly from '@docusaurus/BrowserOnly';
+import NavigationTracker from '../../components/NavigationTracker';
+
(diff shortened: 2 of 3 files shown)


wookasz (approved):

docs/docusaurus/src/theme/Root/index.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import BrowserOnly from '@docusaurus/BrowserOnly';
+import NavigationTracker from '../../components/NavigationTracker';
+
+export default function Root({children}) {
+  return (
+    <>
+      {children}
+      <BrowserOnly>
+        {() => <NavigationTracker />}
```
josectobar, line 10: why do we need an arrow function?

docs/docusaurus/src/components/NavigationTracker/index.js
```diff
@@ -0,0 +1,14 @@
+import { useEffect } from 'react';
+import { useLocation } from '@docusaurus/router';
+
+export default function NavigationTracker() {
```
josectobar, line 4: Why not make this a hook instead?

docs/docusaurus/src/theme/Root/index.js
ashmortar: It is how the [BrowserOnly](https://docusaurus.io/docs/docusaurus-core#browseronly) component works in docusaurus.

docs/docusaurus/src/components/NavigationTracker/index.js
ashmortar: This is also to simplify its usage in the root layout BrowseOnly component.  If I extract the logic into a hook we end up with:

```
export default function Root({children}) {
  return (
    <>
      {children}
      <BrowserOnly>
        {() => {
          useNavigationTracking();
          return null;
        }}
      </BrowserOnly>
    </>
  );
} 
```

Which is functionally the same thing.  Basically the logic has to happen all client side which means being wrapped in browser only and at the top level of the app.

PR #11098, opened 2025-04-14
[MAINTENANCE] Remove test_expectations_v3_api.py

billdirks:


tests/analytics/test_analytics.py
```diff
@@ -79,6 +79,10 @@ def test_event_identifiers(analytics_config):
         assert "organization_id" not in properties
 
 
+@pytest.mark.xfail(
```
billdirks, line 82: This test was only passing because other tests ran first setting up the context. This test always fails when run in isolation. I will write a ticket to followup fixing this.

tests/integration/common_workflows/expect_column_values_to_be_between__evaluation_parameters.json
```diff
@@ -0,0 +1,92 @@
+{
```
billdirks, line 1: I deleted all the json files but this one. It turns out we have some postgresql tests that depend on this table being created. It was previously created in the delete tests setup. I've updated this file to be postgresql only. I've deleted all the test info and just left the data. I moved the file to be next to the test module that uses it.

tests/integration/common_workflows/test_sql_asset_workflows.py
```diff
@@ -30,6 +42,30 @@
 VALUES_ON_MY_FAVORITE_DAY = [8]
 
 
+@pytest.fixture(scope="module", autouse=True)
+def setup_module():
+    # setup table
```
billdirks, line 47: These tests rely on a table that used to be created and populated in the test_expectations_v3_api.py file. They would only pass if run after those tests and would fail if run in isolation. I've moved the table creation logic to this module as well as an edited version of that 1 json file that defined this table.

tests/render/test_render_BulletListContentBlock.py
```diff
@@ -66,78 +54,3 @@ def test_parse_row_condition_string_pandas_engine():
             "row_condition__3": "PClass != '1st'",
         },
     )
-
-
-@pytest.mark.filesystem
-def test_all_expectations_using_test_definitions():
```
billdirks, line 72: I will look closer at this expectation.

tests/render/test_render_BulletListContentBlock.py
```diff
@@ -69,75 +64,18 @@ def test_parse_row_condition_string_pandas_engine():
...
-    with open(
-        file_relative_path(
-            __file__,
-            "./output/test_render_bullet_list_content_block.json",
-            strict=False,
-        ),
-        "w",
-    ) as f:
-        json.dump(test_results, f, indent=2)
+def test_expectations_using_expectation_definitions():
```
billdirks, line 67: I believe the purpose of this test was to a specific render for all expectations that were defined using `test_expectations_v3_api.py`. Since that testing strategy was hard to maintained and didn't contain all the expectations, this strategy isn't ideal. For the purpose of this PR, I extracted all the expectations that were under test and moved them into `BulletListContentBlock.json`. This let me remove all the filtering and I can not iterate through them and run the asserts in this test. This should be equivalent to what was being run before.

tests/render/test_render_BulletListContentBlock.py
```diff
@@ -69,75 +64,18 @@ def test_parse_row_condition_string_pandas_engine():
...
-
-                test_results[test_definitions["expectation_type"]].append(
-                    {
-                        test["title"]: render_result,
-                        # "rendered_template":rendered_template
-                    }
-                )
-
-    # TODO: accommodate case where multiple datasets exist within one expectation test definition
-    with open(
```
billdirks, line 135: I've removed this since it doesn't seem to be testing anything. Maybe it tests that the results are json serializable? Every result should be json serializable because we have called `to_json_dict` above and unit tests for that behavior should go there. Also, we don't need to actually write a file to verify this.

tests/render/test_render_BulletListContentBlock.py
dctalbot: How do we ensure `tests/render/BulletListContentBlock.json` is always up to date?

> and I can not iterate through them

s/not/now ?

dctalbot (changes requested):
I like the refactor to `test_expectations_using_expectation_definitions`, but I'm curious if we have a way to generate `tests/render/BulletListContentBlock.json` and ensure it is always current.

tests/render/test_render_BulletListContentBlock.py
billdirks: It's already out of date since we've already moved away from this json testing framework. I think we should eventually delete this method and incorporate this test in our new expectation testing framework. I will file a followup ticket.
I think it's outside the scope of this work since it's currently falling out of date and this doesn't make it worse. I think it's a good callout to fix it.

dctalbot (approved):

tests/render/test_render_BulletListContentBlock.py
billdirks: The ticket is: https://greatexpectations.atlassian.net/browse/GX-850

billdirks:
I've added https://greatexpectations.atlassian.net/browse/GX-850 to track the block render test work. Adding this to the merge queue.

PR #11103, opened 2025-04-15
[FEATURE] Add `ColumnValuesNotMatchRegexCount` metric

NathanFarmer:


great_expectations/metrics/column/column_values_not_match_regex_count.py
```diff
@@ -0,0 +1,12 @@
...
+from great_expectations.metrics.metric_results import MetricResult
+
+
+class ColumnValuesNotMatchRegexCountResult(MetricResult[int]): ...
+
+
+class ColumnValuesNotMatchRegexCount(ColumnMetric[ColumnValuesNotMatchRegexCountResult]):
+    name = "column_values.not_match_regex.count"
+
+    regex: StrictStr
```
tyler-hoffman, line 12: ooh I don't think we did this for the other ones did we? What's the extra validation that comes with it?

tyler-hoffman (approved):

great_expectations/metrics/column/column_values_not_match_regex_count.py
NathanFarmer: I did it for the last one before this. It rejects anything than can be coerced into a string. I also created this [NonEmptyString](https://github.com/great-expectations/great_expectations/blob/7f0ee6f53828cc6a420f2d316d4a8f760f1107f2/great_expectations/metrics/metric.py#L14) type, but I don't think it applies in this case.

PR #11105, opened 2025-04-15
[FEATURE] Allow user to provide connection details to connect to Redshift

alena-hutchinson:
In addition to a `connection_string`, we want to allow users to connect to Redshift by providing `connection_details`. 

### DIFF
--- a/tests/datasource/fluent/test_redshift_datasource.py
+++ b/tests/datasource/fluent/test_redshift_datasource.py
@@ -4,9 +4,13 @@ import pytest
 from pytest_mock import MockerFixture
 
 from great_expectations.data_context import EphemeralDataContext
-from great_expectations.datasource.fluent.redshift_datasource import RedshiftDsn, RedshiftConnectionDetails
+from great_expectations.datasource.fluent.redshift_datasource import (
+    RedshiftConnectionDetails,
+    RedshiftDsn,
+)
 
 LOGGER = logging.getLogger(__name__)
+
 
 def build_connection_string(scheme, username, password, host, port, database, ssl_mode):
     return f"{scheme}://{username}:{password}@{host}:{port}/{database}?sslmode={ssl_mode}"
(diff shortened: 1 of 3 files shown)


tests/datasource/fluent/test_redshift_datasource.py
```diff
@@ -0,0 +1,136 @@
...
+    password = "password"
+    host = "host"
+    port = 1234
+    database = "database"
+    ssl_mode = "allow"
+    connection_string = build_connection_string(
+        scheme, username, password, host, port, database, ssl_mode
+    )
+
+    expected_kwargs = RedshiftDsn(
```
billdirks, line 50: nit: I could move this right before the assert call. That would make this test structured in the "Arrange, Act, Assert" way where the arrange is setting up the connection string, the action is adding the datasource and calling get_engine, and the assert is asserting we call the spy with these arguments.

tests/datasource/fluent/test_redshift_datasource.py
```diff
@@ -0,0 +1,136 @@
...
+        port=port,
+        path=f"/{database}",
+        query=f"sslmode={ssl_mode}",
+    )
+
+    context = ephemeral_context_with_defaults
+    data_source = context.data_sources.add_redshift(
+        name="redshift_test", connection_string=connection_string
+    )
+    data_source.get_engine()
```
billdirks: Since get_engine doesn't return anything and you are using for a side effect I'd add a comment here that says that you are making this call so we can see the correct connection details being used when getting the engine.

tests/datasource/fluent/test_redshift_datasource.py
```diff
@@ -0,0 +1,136 @@
...
+    host = "host"
+    port = 1234
+    database = "database"
+    ssl_mode = "allow"
+    connection_string = build_connection_string(
+        scheme, username, password, host, port, database, ssl_mode
+    )
+
+    expected_kwargs = RedshiftDsn(
+        connection_string,
```
billdirks, line 51: Why do we need to use both the connection_string and all the other arguments when instantiating the `RedshiftDsn`? If we just specify the connection string, will the others be inferred? Maybe we also have to pass scheme (I seem to recall that is necessary)?

tests/datasource/fluent/test_redshift_datasource.py
```diff
@@ -0,0 +1,136 @@
...
+):
+    create_engine_spy = mocker.patch.object(sa, "create_engine")
+
+    username = "username"
+    password = "password"
+    host = "host"
+    port = 1234
+    database = "database"
+    ssl_mode = RedshiftSSLModes.ALLOW
+    connection_string = build_connection_string(
```
billdirks: I would move this line and the expect_kwargs definition by the assert since they are only used there.

billdirks (commented):
Thanks for doing this! Overall it looks good. I have some small comments about test structure related to the "Arrange Act Assert" organization strategy.

tests/datasource/fluent/test_redshift_datasource.py
alena-hutchinson: You're right. We don't need these, they do get inferred from the `connection_string`.

billdirks (approved):
LGTM, thanks!

PR #11108, opened 2025-04-17
[FEATURE] Add `ColumnDescriptiveStats` metric

NathanFarmer:

### DIFF
--- a/great_expectations/expectations/metrics/column_aggregate_metrics/column_descriptive_stats.py
+++ b/great_expectations/expectations/metrics/column_aggregate_metrics/column_descriptive_stats.py
@@ -79,8 +79,6 @@ class ColumnDescriptiveStats(ColumnAggregateMetricProvider):
         execution_engine: Optional[ExecutionEngine] = None,
         runtime_configuration: Optional[dict] = None,
     ):
-        """Returns a dictionary of given metric names and their corresponding configuration,
-        specifying the metric types and their respective domains"""
         dependencies: dict = super()._get_evaluation_dependencies(
             metric=metric,
             configuration=configuration,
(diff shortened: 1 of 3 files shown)


great_expectations/expectations/metrics/column_aggregate_metrics/column_descriptive_stats.py
```diff
@@ -0,0 +1,108 @@
...
+from great_expectations.expectations.metrics.metric_provider import metric_value
+from great_expectations.validator.metric_configuration import MetricConfiguration
+
+if TYPE_CHECKING:
+    from great_expectations.expectations.expectation_configuration import (
+        ExpectationConfiguration,
+    )
+
+
+class DescriptiveStats(NamedTuple):
```
tyler-hoffman: Not a blocker, but why a named tuple? I guess I would have expected a dataclass, since using this as a tuple seems like an easy way to mix up which value is which.

tyler-hoffman (approved):
Question about the data type, but it looks good!

great_expectations/expectations/metrics/column_aggregate_metrics/column_descriptive_stats.py
NathanFarmer: I guess at first I was just going to return 4 values, which by default is a tuple, and then I named those values. I actually see now that the consumer uses a `BaseModel` so I think I'll just do that. It should functionally work the same regardless though...

NathanFarmer: > since using this as a tuple seems like an easy way to mix up which value is which

Why doesn't adding the names solve that problem? I guess because you can still use indexing?

tyler-hoffman: Yeah, I just feel like if it's used as a tuple, 4 values of the same type makes it a bit easier to e.g. confuse mean with standard dev.

PR #11119, opened 2025-04-22
[FEATURE] Changes to Redshift `connection_string` validator to support a dict type

alena-hutchinson:
Updates to the `RedshiftDatasource` validator to support when the `connection_string` is a `dict` type. 

### DIFF
--- a/great_expectations/datasource/fluent/redshift_datasource.py
+++ b/great_expectations/datasource/fluent/redshift_datasource.py
@@ -84,7 +84,6 @@ class RedshiftDatasource(SQLDatasource):
         connection_string = f"redshift+psycopg2://{connection_details.user}:{connection_details.password}@{connection_details.host}:{connection_details.port}/{connection_details.database}?sslmode={connection_details.sslmode.value}"
         return connection_string
 
-
     @property
     @override
     def execution_engine_type(self) -> Type[SqlAlchemyExecutionEngine]:
--- a/great_expectations/datasource/fluent/redshift_datasource.py
+++ b/great_expectations/datasource/fluent/redshift_datasource.py
@@ -3,11 +3,10 @@ from __future__ import annotations
 from enum import Enum
 from typing import TYPE_CHECKING, Literal, Type, Union
 
-from pydantic.v1 import validator
-
 from great_expectations.compatibility.pydantic import (
     AnyUrl,
     BaseModel,
+    validator,
 )
 from great_expectations.compatibility.typing_extensions import override
 from great_expectations.datasource.fluent.config_str import ConfigStr
--- a/great_expectations/datasource/fluent/redshift_datasource.py
+++ b/great_expectations/datasource/fluent/redshift_datasource.py
@@ -56,8 +56,8 @@ class RedshiftDatasource(SQLDatasource):
         name: The name of this Redshift datasource.
         connection_string: The SQLAlchemy connection string used to connect to the Redshift
             database. This will use a redshift with psycopg2. For example:
-            "redshift+psycopg2://usern@host.amazonaws.com:5439/database". Alternatively,
-             a RedshiftConnectionDetails object can be used.
+            "redshift+psycopg2://user:password@host.amazonaws.com:5439/database?sslmode=sslmode".
+            Alternatively, a RedshiftConnectionDetails object can be used.
         If connection_details is used, connection_string cannot also be provided.
         assets: An optional dictionary whose keys are TableAsset or QueryAsset names and whose
             values are TableAsset or QueryAsset objects.
--- a/great_expectations/datasource/fluent/redshift_datasource.py
+++ b/great_expectations/datasource/fluent/redshift_datasource.py
@@ -37,7 +37,6 @@ class RedshiftSSLModes(Enum):
 class RedshiftConnectionDetails(BaseModel):
     """
     Information needed to connect to a Redshift database.
-    Alternative to a connection string.
     """
 
     user: str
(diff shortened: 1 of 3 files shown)


great_expectations/datasource/fluent/schemas/RedshiftDatasource.json
```diff
@@ -1,6 +1,6 @@
...
+    "description": "Adds a Redshift datasource to the data context using psycopg2.\n\nArgs:\n    name: The name of this Redshift datasource.\n    connection_string: The SQLAlchemy connection string used to connect to the Redshift\n        database. This will use a redshift with psycopg2. For example:\n        \"redshift+psycopg2://user:password@host.amazonaws.com:5439/database?sslmode=sslmode\".\n        Alternatively, a RedshiftConnectionDetails object can be used.\n    If connection_details is used, connection_string cannot also be provided.\n    assets: An optional dictionary whose keys are TableAsset or QueryAsset names and whose\n        values are TableAsset or QueryAsset objects.",
```
klavavej: I don't think this is the right place to apply this change (this looks like generated json) but there are some stray words in this description compared to the similar page for [postgres](https://deploy-preview-11097.docs.greatexpectations.io/docs/reference/api/datasource/fluent/postgresdatasource_class/). This sentence also has grammar issues so I suggest removing it.

```suggestion
    "description": "Adds a Redshift datasource to the data context using psycopg2.\n\nArgs:\n    name: The name of this Redshift datasource.\n    connection_string: The SQLAlchemy connection string used to connect to the Redshift\n        database. For example:\n        \"redshift+psycopg2://user:password@host.amazonaws.com:5439/database?sslmode=sslmode\".\n        Alternatively, a RedshiftConnectionDetails object can be used.\n    If connection_details is used, connection_string cannot also be provided.\n    assets: An optional dictionary whose keys are TableAsset or QueryAsset names and whose\n        values are TableAsset or QueryAsset objects.",
```

great_expectations/datasource/fluent/redshift_datasource.py
```diff
@@ -57,32 +55,34 @@ class RedshiftDatasource(SQLDatasource):
         name: The name of this Redshift datasource.
         connection_string: The SQLAlchemy connection string used to connect to the Redshift
             database. This will use a redshift with psycopg2. For example:
-            "redshift+psycopg2://username@host.amazonaws.com:5439/database"
-        connection_details: A RedshiftConnectionDetails object defining the connection details.
+            "redshift+psycopg2://user:password@host.amazonaws.com:5439/database?sslmode=sslmode".
+            Alternatively, a RedshiftConnectionDetails object can be used.
         If connection_details is used, connection_string cannot also be provided.
```
billdirks, line 58: You can remove this line since there is no more `connection_details` argument. It looks like we also accept a dictionary whose keys are the connection detail attributes.

great_expectations/datasource/fluent/redshift_datasource.py
```diff
@@ -57,32 +55,34 @@ class RedshiftDatasource(SQLDatasource):
...
         assets: An optional dictionary whose keys are TableAsset or QueryAsset names and whose
             values are TableAsset or QueryAsset objects.
     """
 
     type: Literal["redshift"] = "redshift"  # type: ignore[assignment] # This is a hardcoded constant
-    connection_string: Union[ConfigStr, RedshiftDsn]
+    connection_string: Union[RedshiftConnectionDetails, ConfigStr, RedshiftDsn]  # type: ignore[assignment] # Deviation from parent class as individual args are supported for connection
```
billdirks: I think you can remove the `RedshiftConnectionDetails` since you always coerce to a `str`.
Pros: Someone who goes `redshift_datasource.connection_string` doesn't have to handle the different union types.
Cons: Autocomplete will only give the string argument. One will have to do type ignores when passing in these non-str types.
We don't annotate the `dict` argument.

I've gone back and forth. I think we can leave this with the type union. I'd add the dict type, probably as a TypedDict with the right keys.

great_expectations/datasource/fluent/schemas/RedshiftDatasource.json
billdirks: This is derived from the docstring in `redshift_datasource.py`.

tests/datasource/fluent/test_redshift_datasource.py
```diff
@@ -56,38 +56,38 @@ def test_create_engine_is_called_with_expected_kwargs_using_connection_string(
...
 
-    username = "username"
+    user = "user"
     password = "password"
     host = "host"
     port = 1234
     database = "database"
-    ssl_mode = RedshiftSSLModes.ALLOW
+    sslmode = RedshiftSSLModes.ALLOW
     connection_details = RedshiftConnectionDetails(
```
billdirks: It would be good to add 1 more test identical to this one where we use a dict instead.
You might be able to write a parameterized test with the inputs the connection_string argument and the expected arguments you assert against.

billdirks (commented):
Thanks for doing this! I had a few small comments. The main one is adding a dict type as a connection_string parameter since it looks like we take that and need to for the agent. Also adding a test for `dict` input would be good.

great_expectations/datasource/fluent/schemas/RedshiftDatasource.json
alena-hutchinson: Removed from `redshift_datasource.py` and regenerated schema

billdirks (approved):
LGTM, thanks!

PR #11127, opened 2025-04-23
[MAINTENANCE] Improve `MetricErrorResult` and `Batch.compute_metrics()` API typing

NathanFarmer:
- Update `Batch.compute_metrics()` API typing to include `MetricErrorResult`
- Simplify `MetricErrorResult`
- Consolidate error types
--------

### DIFF
--- a/great_expectations/datasource/fluent/interfaces.py
+++ b/great_expectations/datasource/fluent/interfaces.py
@@ -62,8 +62,6 @@ from great_expectations.metrics.metric_name import MetricNameSuffix
 from great_expectations.metrics.metric_results import (
     MetricErrorResult,
     MetricErrorResultValue,
-    MetricInternalErrorResult,
-    MetricInternalErrorResultValue,
     MetricResult,
 )
 from great_expectations.validator.metrics_calculator import (
@@ -1331,12 +1329,12 @@ class Batch:
                 results.append(MetricErrorResult(id=metric_id_for_batch, value=value))
             else:
                 results.append(
-                    MetricInternalErrorResult(
+                    MetricErrorResult(
                         id=metric_id_for_batch,
-                        value=MetricInternalErrorResultValue(
-                            msg=f"Metric {metric.name} not found in results: "
-                            "{list(metrics_calculator_results.keys())} or errors: "
-                            "{list(metrics_calculator_errors.keys())}"
+                        value=MetricErrorResultValue(
+                            exception_message=f"Metric {metric.name} not found in results: "
+                            f"{list(metrics_calculator_results.keys())} or errors: "
+                            f"{list(metrics_calculator_errors.keys())}",
                         ),
                     )
                 )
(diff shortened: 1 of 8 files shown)


great_expectations/datasource/fluent/interfaces.py
```diff
@@ -1317,16 +1319,22 @@ def compute_metrics(
...
+                value = MetricErrorResultValue(
+                    exception_message=metrics_calculator_error["exception_info"].exception_message,
+                    exception_traceback=metrics_calculator_error[
+                        "exception_info"
+                    ].exception_traceback,
+                )
                 results.append(MetricErrorResult(id=metric_id_for_batch, value=value))
             else:
                 results.append(
-                    MetricInternalErrorResult(
```
NathanFarmer, line 1324: Now that this type is being added to the API, I decided to consolidate it with `MetricErrorResult`.

great_expectations/datasource/fluent/interfaces.py
```diff
@@ -1257,14 +1256,16 @@ def _create_validator(self, *, result_format: ResultFormatUnion) -> V1Validator:
         )
 
     @overload
-    def compute_metrics(self, metrics: Metric[_MetricResultT]) -> _MetricResultT: ...
+    def compute_metrics(
+        self, metrics: Metric[_MetricResultT]
+    ) -> _MetricResultT | MetricErrorResult: ...
```
billdirks, line 1261: A `MetricErrorResult` is a `_MetricResultT` since it's defined as:
```
class MetricErrorResult(MetricResult[MetricErrorResultValue]): ...
```

tyler-hoffman (approved):
Thanks!

great_expectations/datasource/fluent/interfaces.py
NathanFarmer: It won't match the type var type though right? It will necessarily be different than the input type.

great_expectations/metrics/metric_results.py
```diff
@@ -23,20 +22,14 @@ class Config:
         arbitrary_types_allowed = True
 
 
-class MetricErrorResultValue(TypedDict):
-    exception_info: ExceptionInfo
+class MetricErrorResultValue(BaseModel):
```
billdirks, line 25: The MetricErrorResultValue isn't public or returned by anything public so it is ok to change. If you wanted to make it backwards compatible you could do 
```
class MetricErrorResultValue(BaseModel):
  exception_info: ExceptionInfo
  exception_message: str
  exception_traceback: Union[str, None] = None
```

NathanFarmer: This is returned by the API. This is what you can find in `MetricErrorResult.value`. We had it the way you show in the first incarnation of `MetricErrorResult`, but its shape is confusing. `ExpectionInfo` includes duplicate `exception_message` and `exception_traceback` 😮‍💨

great_expectations/datasource/fluent/interfaces.py
billdirks: Oh, I see. You are right.

billdirks (approved):
Thanks!

